PR #517, opened 2019-11-24
Added Portis support.

mimafogeus2:
- Users can now log in with their Portis accounts.
- Instead of a single login button, there are now two buttons, one for each supported provider.


NoahZinsmeister:
Thanks for the PR @mimafogeus2, but we'll be adding different wallet providers on our own timeline/as we deem they are necessary.

PR #521, opened 2019-11-27
Replace networkId with chainId

PaulRBerg:
It seems web3-react v6 uses `chainId` instead of `networkId`, but the `useFactoryContract` Hook is still deconstructing the old field. Here's a [CodeSandbox](https://codesandbox.io/s/web3-react-v6-gv56z) that proves it (see the logs).

Capture d’écran 2019-11-27 à 16 26 50


NoahZinsmeister:
good catch, thanks @PaulRBerg !

PR #522, opened 2019-11-27
Import useWeb3React from the hooks folder

PaulRBerg:
Not sure if this is a bug, but I noticed that in all components `useWeb3React` is imported from the local hooks folder, not from the `@web3-react/core` package, except for `AccountDetails`.


NoahZinsmeister:
This is intentional! We use the library's version wherever we need to deal with the multi-root stuff (network + account contexts), and the version from hooks whenever we just want an active connection.

NoahZinsmeister:
Ah good catch again! Thanks @PaulRBerg , yes we're using the local version from hooks to abstract away the multi-context.

PR #533, opened 2019-12-05
Adds min-width to custom slippage input.

MicahZoltu:
On Firefox (maybe other browsers as well) the lack of min-width results in the default `input` element min-width which is ~92px.  This results in the flex container overflowing the bounding element.  By setting min-width to 20px, the box will instead grow to fill the available space, which is defined by its parent that has the pink border.  This avoids the overflow issue (as long as the parent doesn't get too small, which it won't at the moment since it is fixed size).

Without `min-width` set in Firefox:
image

With `min-width` set in Firefox:
image

### DIFF
--- a/src/components/TransactionDetails/index.js
+++ b/src/components/TransactionDetails/index.js
@@ -169,6 +169,7 @@ const Input = styled.input`
   background: ${({ theme }) => theme.inputBackground};
   flex-grow: 1;
   font-size: 12px;
+  min-width: 20px;
 
   outline: none;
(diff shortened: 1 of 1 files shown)


MicahZoltu:
I tested this fix via Firefox debug tools and I'm reasonably confident that this is the right place to put the fix.  Someone should verify the fix works when building from source before merging.

MicahZoltu:
I was able to verify the fix worked with the deploy preview.

PR #534, opened 2019-12-05
block some countries

NoahZinsmeister:


banteg:
For the love of crypto, remove this bullshit.

vavanu4:
Did the fking Binance number two appear for us in Belarus?

illlefr4u:
It's disgusting! Start with a great idea, get development grants, gain weight, raise a round of financing and start banning people in countries. 
The usual story. @haydenadams change your mind, please. Dont want to belive that this is true!

ArtemBernatskyy:
uniswap isn't Defi anymore, RIP

tshabs:
Hi,

While it was great to meet you in Berlin over the summer I have been continually disappointed with Uniswap since.

For example, I don't get the obsession with socks while people are trying to build real products (such as swap-box) and in the meantime communication has been zero (just as with this pull request offering zero explanation for users).

I am originally from one of the countries in your blacklist which adds to the sting as I know people who need tools like this more than any of us (and already go through great lengths and risk to get uncensored internet connectivity, which can be dangerous).

I understand you need to be legally compliant but the failure here is in communication to end users (also linking people to a wiki page explaining internet censorship in general wouldn't hurt).

Hopefully this doesn't stem from a sort of naivety where you are self censoring / self policing because some over paranoid lawyer told you to without getting several opinions and finally making your own determination.

Finally, seeing as you are blocking to cover your own ass, you should NOT have the blocks be included in master branch (and it wouldn;t hurt to provide better deployment instructions).

Other than these complaints I hope you are doing well.

T.

davekaj:
I don't see this as a bad thing. It is the world we live in. 

Uniswap the protocol needs to be developed by some group of people. That group of people should not put themselves at risk for running an illegal front end.

The code is open source, anyone can fork the front end and remove the blocks and put it on IPFS. That is what is so great about open source software! Different groups around the world can work together to ensure everyone has access to Uniswap. 

In this case I put myself in their shoes. If I were building the Uniswap protocol and a lawyer who understands the law much better than me told me to block these countries, I would, because law is not my specialty and I will take my lawyers advice.

Then, I would post the non blocked version myself on IPFS, and secretly let people know about it.  This is strategic survival. Not putting yourself at risk, while continuing to work on the protocol, and anonymously hosting a completely unbanned front end.

naumovlab:
use openswap.eth

marsrobertson:
Related.

See the commit: https://github.com/steemit/condenser/commit/cdf56507e11f367f6decc42a6c059c2d30db9cf4#diff-60c2218cad55eefbe2060ac6aa46a4e0

Since then the list grew more: https://github.com/steemit/condenser/blob/master/src/app/utils/GDPRUserList.js

Surely you can clone the repo, run on localhost or IPFS?

amiut:
It's not DeFi anymore it's CeFi

PR #543, opened 2019-12-09
Allow to set custom deadline

### LINKED ISSUE #232: Adjustable deadline
The UI currently sets a pretty short default deadline, which causes failed transactions. Users are not aware of there being a deadline for their transaction since this is not indicated on the UI, so they might set a low gas price that is bound to trigger the deadline. They also can't see the reason why the TX failed on Etherscan.

The UI should display what the deadline is for the transaction. It should also allow the user to adjust it.

xuhcc:
Added deadline input field to 'Advanced Details' dropdown section.

Resolves #232


NoahZinsmeister:
LGTM, thanks @xuhcc !

PR #544, opened 2019-12-09
Balance and USD Refactor

ianlapham:
- move all balance fetching logic into balance context 

- update USD logic to use balance data for reserve information 

- stagger network requests on initial balance fetching

- minor change to wallet display for injected icon

**Open questions**
 - Do we want to handle ethRate fetching within the balance context (right now the calculation is done in the components by getting eth and token reserves for a given exchange, then doing ethBalance/tokenBalance and accounting for decimals)?  <-- see CurrencySelectModal 

- Right now we have a hook to get all balance data (which includes exchange balances and account balances. Do we want to have a separate hook that only exposes account balances (safeAccess(...state, [chainid, account]))?


NoahZinsmeister:
LGTM! pushed some small changes that we can discuss offline.

one quick thing that might be nice to look into before we merge is the performance of the token modal when it opens. right now it's really slow :( there might be some free wins here by improving the sorting code, memoizing, making some components pure, etc. probably not worth spending _too_ much time on, but maybe there's low-hanging fruit.

NoahZinsmeister:
oh, and 1 other thing to consider would be to first fetch all token balances, and only then fetch the eth + token balances for tokens *if* the user has a non-zero balance in that token. this would prevent some network spam, with the primary downside that we'd introduce a data request waterfall (and there might be some trickiness in keeping those now-not-fetched eth + token balances up to date if we ever think that the user might have a balance in the token, though i think that may be handled by our existing setup).

PR #558, opened 2019-12-15
Remove stale "handleNetworkChange" function

PaulRBerg:
I was rebasing the latest commits and I noticed this stale function.


PaulRBerg:
Whoops, I'm tired. I missed the "networkChanged" PromiEvent from the side effect cleanup.

PR #574, opened 2019-12-22
fix ordering of tokens

TomAFrench:
Currently DAI is shown at the bottom of the list of tokens after ZRX. As the rest of the list is in (mostly) alphabetical order, I've popped it further up in its proper place.


TomAFrench:
Hmm, I remember seeing DAI being shown at the end of the list but now when I go back to the frontend it's now being properly sorted. Closing this as it's unnecessary.

NoahZinsmeister:
thanks @TomAFrench , good catch!

TomAFrench:
Cheers!

PR #577, opened 2019-12-27
Update GA

ianlapham:
- Update Google analytics 

- Regroup into main categories (advanced interactions, transactions, wallet interactions)

- Add transaction values in USD for swaps, sends, add and remove 

- Fix bug which causes "transaction details" event to trigger many times on user input (notice on analytics dashboard "open" action occurs an unrealistic magnitude more than other events)

- edit styling on details hover popup


NoahZinsmeister:
LGTM!

PR #589, opened 2020-01-07
add warning for broken ERC20

ianlapham:
- Add warning for new broken token 0xc3761EB917CD790B30dAD99f6Cc5b4Ff93C4F9eA

- rearrange some imports  

- show warning on all pages, not just add liquidity


rphansen91:
What constitutes being added to the broken token list? I am specifically asking about the reason for including token contract `0x55296f69f40ea6d20e478533c15a6b08b654e758`. Thank you!

PR #594, opened 2020-01-12
Add Gitpod config

nordicenergy:


NoahZinsmeister:
i'm not too inclined to support gitpod @nordicenergy , could you elaborate on why we'd want this?

PR #601, opened 2020-01-15
update market rate calculation and input parsing for 0 decimal tokens

ianlapham:
- fix bugs when making txs with 0 decimal tokens 

- fix parsing issues for 0 decimal tokens on 1st liquidity add 

- bump styled components version (


NoahZinsmeister:
looks like the styled components change messed up the styling.

mind reverting the changes to `src/contexts/Tokens.js`? or at least checksumming the addresses if it's a better Rinkeby DAI.

ianlapham:
The DAI that was there before is probably better (more usage) so just reverted back.

PR #603, opened 2020-01-16
Save user's wallet preference

cwhinfrey:
This pull request adds the ability to store the user's wallet preference (the last wallet they used). When a user refreshes the page or visits Uniswap again, their preferred wallet will reconnect when possible. The following supported wallets can reconnect:
* Injected wallets
* Portis
* Fortmatic (currently a rate limit error is shown when reconnecting. This error must be solved first.)

Also, when a browser blocks pop ups, wallets like Authereum will fallback to redirecting to the wallet's domain for authentication and then redirecting back to Uniswap. This PR will ensure the user does not have to click "Connect a wallet" a second time to connect after being redirected.

To try functionality:
* Log in with Portis or Fortmatic
* Refresh the page
* You will still be logged in to the selected wallet or will be asked to authenticate if you have logged out of that wallet

### DIFF
--- a/src/components/AccountDetails/index.js
+++ b/src/components/AccountDetails/index.js
@@ -4,7 +4,7 @@ import { useWeb3React } from '../../hooks'
 import { isMobile } from 'react-device-detect'
 import Copy from './Copy'
 import Transaction from './Transaction'
-import { SUPPORTED_WALLETS } from '../../constants'
+import { SUPPORTED_WALLETS, WALLET_PREFERENCE_KEY } from '../../constants'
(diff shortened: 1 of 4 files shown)


cwhinfrey:
Check out this functionality here:
[Mainnet](https://uniswap.authereum.org/)
[Kovan](https://uniswap.demo.authereum.org/)

I also, updated it to not block the UI when loading the user's preferred wallet but those changes aren't on this PR right now. If this PR is being considered, let me know and I'll add the changes to not block the UI.

PR #610, opened 2020-01-24
Token check

ianlapham:
- add warning card for tokens pasted in our linked to through URL


ianlapham:
@callil The one thing we want to check is where to align the "added by user" text in the modal

PR #622, opened 2020-01-31
Add Authereum 🛡️❤️🦄

cwhinfrey:
Add Authereum as a wallet option.


NoahZinsmeister:
@cwhinfrey should this be using `@web3-react/authereum-connector@^6.0.8`, the version that includes the changes you made in your recent PR to `web3-react`? also, why are you calling `window.open`, shouldn't this be `enable` on the authereum provider?

cwhinfrey:
@NoahZinsmeister I just bumped the version to 6.0.8.

> also, why are you calling window.open, shouldn't this be enable on the Authereum provider?

This is to link out to the user's Account page when they click "show Authereum". Let me know if you would expect different behavior and I can update it.

PR #629, opened 2020-02-09
Make token address copyable in token selection screen

KadenZipfel:
Positioning may not be ideal and may cause issues on mobile layout. Open to suggestions.

### DIFF
--- a/src/components/CurrencyInputPanel/Copy.js
+++ b/src/components/CurrencyInputPanel/Copy.js
@@ -25,14 +25,14 @@ const TransactionStatusText = styled.span`
 
 export default function CopyHelper({ toCopy }) {
   const [isCopied, setCopied] = useCopyClipboard()
-  
+
   function handleClick(e) {
-    setCopied(toCopy);
-    e.stopPropagation();
+    setCopied(toCopy)
+    e.stopPropagation()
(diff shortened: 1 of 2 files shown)


callil:
I like this idea! I think it can be a really good UX improvement - few things to fix:

The layout of token names is a bit messed up now - they should stay on one line. The copy should say "Token Address Copied" 

As far as the copy button - I think the better location of the copy button would be next to the token name as seen in this screenshot, alternatively it could also go next to the ticker name:

Screen Shot 2020-02-10 at 11 00 01 AM

KadenZipfel:
@callil Some notes:
- I removed the copy button for ETH since it would just copy 'ETH'
- Initially I lined up the copy button the way you have in that screenshot, but then 'Token Address Copied' is misaligned when you click it so I decided to align that instead
- I'm a little bit concerned that 'Token Address Copied' might be too long in some circumstances, maybe it would be worth reducing the font size here?

NoahZinsmeister:
`Token Address Copied` causes a pretty bad reflow on mobile (and also the icon looks a little funny on mobile as well, sometimes it collides with the `-` used to indicate no balance).

other than that, this LGTM!

PR #631, opened 2020-02-12
Prioritize exact token symbol matches

GregTheGreek:
This adds an opinionated sort to the `filteredTokenList` to prioritize `DAI` when a user inputs `DAI` into the token search.

Previous:
Screen Shot 2020-02-12 at 10 06 42

New:
Screen Shot 2020-02-12 at 11 10 01

### DIFF
--- a/src/components/CurrencyInputPanel/index.js
+++ b/src/components/CurrencyInputPanel/index.js
@@ -568,7 +568,12 @@ function CurrencySelectModal({ isOpen, onDismiss, onTokenSelect, urlAddedTokens 
       })
       return regexMatches.some(m => m)
     })
-    return searchQuery.toUpperCase() === 'DAI' && list.sort((a, b) => { return a.symbol === 'DAI' ? -1 : 1 })
+    return (
+      searchQuery.toUpperCase() === 'DAI' &&
+      list.sort((a, b) => {
+        return a.symbol === 'DAI' ? -1 : 1
+      })
+    )
   }, [tokenList, searchQuery])
 
   function _onTokenSelect(address) {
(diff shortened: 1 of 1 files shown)


NoahZinsmeister:
thanks for this @GregTheGreek 

a few comments:

1. the sort should happen when the filtered list is created, [here](https://github.com/Uniswap/uniswap-frontend/blob/bab5a1a5c6b9bc1b2c011bf168c8f8c1a983c665/src/components/CurrencyInputPanel/index.js#L570)

2. the same logic should apply to all exact symbol matches against `searchQuery`, and should be case-insensitive (see code earlier in that file for lowercased logic)

happy to merge once these two fixes are addressed!

GregTheGreek:
@NoahZinsmeister I made some changes, let me know if thats what you wanted

NoahZinsmeister:
i meant something more along the lines of:

```javascript
return (
  list.sort((a, b) => {
    return a.symbol.toLowerCase() === searchQuery.toLowerCase() ? -1 : 1
  })
)
```

also are we sure that sorting the list in this way isn't going to change the existing order?

ChristophSiegenthaler:
Love the idea! When I apply the changes I get "exchange not found" on all querries except "dai", however.

NoahZinsmeister:
@ChristophSiegenthaler yeah it's because the list isn't being returned unless the `searchQuery` matches DAI atm, which is wrong

GregTheGreek:
> ```js
> return (
>   list.sort((a, b) => {
>     return a.symbol.toLowerCase() === searchQuery.toLowerCase() ? -1 : 1
>   })
> )
> ```

Right this makes sense, but to your point about the existing order, it breaks it. Thats why i prioritized DAI (should have returned everything otherwise). Personally, I feel like we can get away with isolating DAI since its one of the few things out there that have a lot of wrapped versions, otherwise just return the original list

GregTheGreek:
(BTW i was AFK this weekend, so sorry for not responding)
Actually, here is an alternative approach:

```
if (searchQuery === "") return list;

return (
    list.sort((a, b) => {
        return a.symbol.toLowerCase() === searchQuery.toLowerCase() ? -1 : 1
    })
)
```

NoahZinsmeister:
LGTM, thanks @GregTheGreek !

PR #634, opened 2020-02-13
update logic for txn eth value on GA

ianlapham:
- update to include correct ETH txn size for GA events on swap and send

### DIFF
--- a/src/components/ExchangePage/index.jsx
+++ b/src/components/ExchangePage/index.jsx
@@ -629,12 +629,11 @@ export default function ExchangePage({ initialCurrency, sending = false, params 
     if (inputCurrency === 'ETH') {
       ethTransactionSize = inputValueFormatted
     } else if (outputCurrency === 'ETH') {
-      // we want input value * exchangeRate
       ethTransactionSize = inputValueFormatted * amountFormatter(exchangeRate, 18, 6, false)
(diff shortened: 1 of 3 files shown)


src/components/ExchangePage/index.jsx
```diff
@@ -623,13 +624,18 @@ export default function ExchangePage({ initialCurrency, sending = false, params
...
-      inputEthPerToken = inputReserveToken && inputReserveETH ? inputReserveETH / inputReserveToken : null
+    if (inputCurrency === 'ETH') {
+      ethTransactionSize = inputValueFormatted
+    } else if (outputCurrency === 'ETH') {
+      ethTransactionSize = inputValueFormatted * amountFormatter(exchangeRate, 18, 6, false)
+    } else {
+      const tokenBalance = inputReserveToken && new BigNumber(inputReserveToken.toString())
+      const ethBalance = inputReserveETH && new BigNumber(inputReserveETH.toString())
+      let ethRate = ethBalance && tokenBalance && ethBalance.div(tokenBalance)
+      ethTransactionSize = inputValueFormatted * ethRate
```
NoahZinsmeister: is this right?? feel like we should probably never be using `*`, and isn't `inputValueFormatted` a string?

PR #671, opened 2020-04-09
add Monarch Token (MT)

mmbower1:


mmbower1:
Should this be done in the "Uniswap Front End Token Listing Request" section from the uniswap docs instead?

PR #682, opened 2020-04-17
Add price bar and spacing tweaks

callil:


haydenadams:
@callil seems the price shown is the market price and not the execution price of a trade (as it is in the current interface)

callil:
Good catch - @ianlapham I used the midPrice like it was before - whats the correct way to get price in this case?

ianlapham:
@callil We can get the average execution price from the "trade" instance. I can update when i pull this in

PR #706, opened 2020-05-02
Update local env setup documentation

KartikTalwar:
Hey all, was setting up the frontend locally and noticed that the variable change in the `.env.local.example` (from #621) renamed `REACT_APP_NETWORK_ID` to `REACT_APP_CHAIN_ID` but the README still references the old variable name so addressing that documentation fix here


ianlapham:
Looks good thanks!

PR #731, opened 2020-05-08
Use best trade to get the best route

moodysalem:


src/contexts/Transactions.tsx
```diff
@@ -16,7 +16,11 @@ const ADD = 'ADD'
 const CHECK = 'CHECK'
 const FINALIZE = 'FINALIZE'
 
-const TransactionsContext = createContext()
+interface TransactionState {
+
```
moodysalem, line 20: todo....

PR #733, opened 2020-05-08
Type safety for themed components

moodysalem:


src/components/AccountDetails/index.tsx
```diff
@@ -217,24 +217,25 @@ function renderTransactions(transactions, pending) {
...
-  confirmedTransactions,
-  ENSName,
-  openOptions
-}) {
+                                         toggleWalletModal,
+                                         pendingTransactions,
+                                         confirmedTransactions,
+                                         ENSName,
+                                         openOptions
+                                       }) {
```
NoahZinsmeister, line 232: is this really what prettier does? is your IDE picking up the repo's version + config?

moodysalem: No, just IntelliJ. Once we have linters working for TS + a github action for it, we can reformat everything all at once

PR #734, opened 2020-05-08
Cypress

moodysalem:


.github/workflows/tests.yaml
```diff
@@ -0,0 +1,34 @@
+name: Tests
+on:
+  push:
+    branches:
+      - v2
+  pull_request:
+    branches:
+      - v2
+jobs:
```
NoahZinsmeister, line 9: we should probably add a job to replicate what travis was doing

moodysalem: the travis job only check linting on js files right now, which isn't useful as i'm moving them to TS, so added a todo for this at the bottom of the file

PR #739, opened 2020-05-08
noah changes

NoahZinsmeister:
see commits for more detailed comments


src/components/ExchangePage/index.tsx
```diff
@@ -311,48 +314,39 @@ function ExchangePage({ sendingInput = false, history, params }: ExchangePagePro
...
-        ? parsedAmounts[Field.OUTPUT]
-        : calculateSlippageAmount(parsedAmounts[Field.OUTPUT])?.[0] &&
-          new TokenAmount(tokens[Field.INPUT], calculateSlippageAmount(parsedAmounts[Field.OUTPUT])?.[0])
+  const slippageAdjustedAmounts: { [field: number]: TokenAmount } = {
+    [independentField]: parsedAmounts[independentField],
+    [dependentField]: parsedAmounts[dependentField]
+      ? tradeType === TradeType.EXACT_INPUT
+        ? new TokenAmount(
+            tokens[dependentField],
+            calculateSlippageAmount(parsedAmounts[dependentField], allowedSlippage)[0]
```
moodysalem, line 341: Can we add a unit test to this? Feel like it should be in SDK, something like `Trade#minimumAmountOut(allowedSlippage: Percent): TokenAmount` and `Trade#maximumAmountIn(allowedSlippage: Percent): TokenAmount`

For exact out/in trades the above methods respectively would just return output/input

src/components/ExchangePage/index.tsx
```diff
@@ -468,7 +460,7 @@ function ExchangePage({ sendingInput = false, history, params }: ExchangePagePro
           sending ? recipient : account,
           deadlineFromNow
         ]
-        value = hex(slippageAdjustedAmounts[Field.INPUT].raw)
+        value = BigNumber.from(slippageAdjustedAmounts[Field.INPUT].raw.toString())
```
moodysalem, line 463: I think we should extract this transaction construction logic into a unit testable thing since we definitely need to verify the arguments in the tx we construct (todo)

src/components/ExchangePage/index.tsx
```diff
@@ -643,14 +640,15 @@ function ExchangePage({ sendingInput = false, history, params }: ExchangePagePro
...
+    account
   ])
 
   // warnings on slippage
   const warningLow: boolean =
     slippageFromTrade &&
-    parseFloat(slippageFromTrade.toFixed(4)) < ALLOWED_SLIPPAGE_MEDIUM / 100 &&
-    parseFloat(slippageFromTrade.toFixed(4)) > 0
+    parseFloat(slippageFromTrade.toFixed(4)) > 0 &&
+    parseFloat(slippageFromTrade.toFixed(4)) < ALLOWED_SLIPPAGE_MEDIUM / 100
```
moodysalem, line 651: opportunity to use !greaterThan if you want to convert the constants into percents which should be more reliable

src/components/ExchangePage/index.tsx
```diff
@@ -827,16 +825,8 @@ function ExchangePage({ sendingInput = false, history, params }: ExchangePagePro
...
-                      ? slippageAdjustedAmounts[Field.OUTPUT]?.toFixed(5) === '0.00000'
-                        ? '<0.00001'
-                        : slippageAdjustedAmounts[Field.OUTPUT]?.toFixed(5)
-                      : '-'
-                    : slippageAdjustedAmounts[Field.INPUT]
-                    ? slippageAdjustedAmounts[Field.INPUT]?.toFixed(5) === '0.00000'
-                      ? '<0.00001'
-                      : slippageAdjustedAmounts[Field.INPUT]?.toFixed(5)
-                    : '-'}
+                    ? slippageAdjustedAmounts[Field.OUTPUT]?.toSignificant(4) ?? '-'
```
moodysalem, line 828: I ran into an issue with operator order before, does this need parentheses?

src/components/ExchangePage/index.tsx
```diff
@@ -1304,11 +1266,7 @@ function ExchangePage({ sendingInput = false, history, params }: ExchangePagePro
                     warningMedium={warningMedium}
                     warningHigh={warningHigh}
                   >
-                    {priceSlippage
-                      ? priceSlippage.toFixed(4) === '0.0000'
-                        ? '<0.0001%'
-                        : priceSlippage.toFixed(4) + '%'
-                      : '-'}
+                    {priceSlippage?.toFixed(2) === '0.00' ? '<0.01%' : `${priceSlippage?.toFixed(2)}%` ?? '-'}
```
moodysalem, line 1269: weird expression, wouldn't 0.009 round to 0.01 or does it truncate? prefer .lessThan(Percent(1,100))

src/components/PoolFinder/index.tsx
```diff
@@ -15,71 +14,40 @@ import { AutoColumn, ColumnCenter } from '../Column'
...
-  const position: TokenAmount = useAddressBalance(account, pair?.liquidityToken)
-
-  const newPair: boolean = pair && JSBI.equal(pair.reserve0.raw, JSBI.BigInt(0))
-  const allowImport: boolean = position && JSBI.greaterThan(position.raw, JSBI.BigInt(0))
-
-  const allPairs = useAllPairs()
-  const pairCount = Object.keys(allPairs)?.length
-  const pairCountPrevious = usePrevious(pairCount)
-  const [newLiquidity, setNewLiquidity] = useState<boolean>(false) // check for unimported pair
+  const pair: Pair = useReserves(token0, token1)
```
moodysalem, line 37: `useReserves` is an odd name for a function that returns a pair

src/contexts/Tokens.ts
```diff
@@ -8,20 +8,31 @@ export const ALL_TOKENS = [
...
 
   // Ropsten Tokens
   new Token(ChainId.ROPSTEN, '0xaD6D458402F60fD3Bd25163575031ACDce07538D', 18, 'DAI', 'Dai Stablecoin')
 
   // Goerli Tokens
 ]
+  // remap WETH to ETH
+  .map(token => {
+    if (token.equals(WETH[token.chainId])) {
+      ;(token as any).symbol = 'ETH'
```
moodysalem, line 31: 😢

src/hooks/Trades.ts
```diff
@@ -1,26 +1,38 @@
...
+  const aToDAI = useReserves(inputToken, chainId === ChainId.MAINNET ? DAI : null)
+  const bToDAI = useReserves(outputToken, chainId === ChainId.MAINNET ? DAI : null)
+
+  // get token<->USDC pairs
+  const aToUSDC = useReserves(inputToken, chainId === ChainId.MAINNET ? USDC : null)
+  const bToUSDC = useReserves(outputToken, chainId === ChainId.MAINNET ? USDC : null)
 
   return useMemo(() => {
-    const allPairs = [pairBetween, aToETH, bToETH].filter(p => !!p)
+    const allPairs = [pairBetween, aToETH, bToETH, aToDAI, bToDAI, aToUSDC, bToUSDC].filter(p => !!p)
```
moodysalem, line 33: optionally, usdc to dai? dai to eth? usdc to eth?

src/components/TokenLogo/index.tsx
```diff
@@ -7,9 +7,7 @@ import { WETH } from '@uniswap/sdk'
 import { ReactComponent as EthereumLogo } from '../../assets/images/ethereum-logo.svg'
 
 const TOKEN_ICON_API = address =>
-  `https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/${isAddress(
-    address
-  )}/logo.png`
+  `https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/${address}/logo.png`
```
ianlapham, line 10: the addresses we store are always checksumed right?

src/components/ExchangePage/index.tsx
moodysalem: https://github.com/Uniswap/uniswap-sdk/pull/23

NoahZinsmeister: added unit tests, agreed that it should be in the SDK!

src/components/ExchangePage/index.tsx
NoahZinsmeister: want to get this merged, so going to resolve, but i agree

src/components/ExchangePage/index.tsx
NoahZinsmeister: yup, done

src/components/ExchangePage/index.tsx
NoahZinsmeister: i tried adding parens but prettier ate them. fwiw `??` should have higher precedence than the ternary, so it should be ok (although looking at it again, i guess it's weird that we're specifying the `'-'` fallback separately for each case, could just do it once...)

src/components/ExchangePage/index.tsx
NoahZinsmeister: yeah it would round to 0.01 by default (thought his could be changed by passing additional arguments to toFixed). i was just replicating existing logic but agreed its a bad check, fixed

src/contexts/Tokens.ts
NoahZinsmeister: i know i know. i have a `getTokenDisplayValue` function or something in hypertext, imo that's a better option for display overrides (logic overrides are a different beast)

src/hooks/Trades.ts
NoahZinsmeister: A -> USDC -> DAI -> B seems like a bad option, but in theory yes we could include them. perhaps depends on observational data on which pairs have high liquidity.

src/components/PoolFinder/index.tsx
NoahZinsmeister: agreed, relic of old code, changed

src/components/TokenLogo/index.tsx
NoahZinsmeister: fixed a case where they weren't but otherwise yes, should be, the SDK guarantees it

PR #743, opened 2020-05-09
Create some integration tests

moodysalem:


moodysalem:
Going to merge since these are just more integration tests

PR #747, opened 2020-05-12
Ui changes

ianlapham:
- copy changes and some feature tweaks


src/components/AddressInputPanel/index.tsx
```diff
@@ -190,16 +190,15 @@ export default function AddressInputPanel({
...
-                <Link href={getEtherscanLink(chainId, data.name, 'address')} style={{ fontSize: '14px' }}>
-                  (View on Etherscan)
-                </Link>
-              ) : (
-                <Link href={getEtherscanLink(chainId, data.address, 'address')} style={{ fontSize: '14px' }}>
-                  (View on Etherscan)
-                </Link>
-              )}
+              {data.name ||
+                (data.address && (
```
NoahZinsmeister, line 194: the parens are wrong here btw, let's try to be more careful. i'll fix in in my PR, so no need to do anything.

moodysalem:
Was removing the light mode link in this PR from the menu intentional?

PR #750, opened 2020-05-12
fix bugs on non existent vs no reserves, fix search scanning, small UI changes

ianlapham:
- for usePair: return undefined if loading, null if no pair created, and pair otherwise
- fix QR code scanning on dark mode

### DIFF
--- a/src/state/transactions/actions.ts
+++ b/src/state/transactions/actions.ts
@@ -1,5 +1,15 @@
-import { TransactionReceipt } from '@ethersproject/providers'
 import { createAction } from '@reduxjs/toolkit'
+
+export interface SerializableTransactionReceipt {
+  to: string
+  from: string
+  contractAddress: string
+  transactionIndex: number
+  blockHash: string
+  transactionHash: string
+  blockNumber: number
+  status?: number
+}
 
 export const addTransaction = createAction<{
   chainId: number
@@ -8,6 +18,8 @@ export const addTransaction = createAction<{
   summary?: string
 }>('addTransaction')
 export const checkTransaction = createAction<{ chainId: number; hash: string; blockNumber: number }>('checkTransaction')
-export const finalizeTransaction = createAction<{ chainId: number; hash: string; receipt: TransactionReceipt }>(
-  'finalizeTransaction'
-)
+export const finalizeTransaction = createAction<{
+  chainId: number
+  hash: string
+  receipt: SerializableTransactionReceipt
+}>('finalizeTransaction')
--- a/src/state/transactions/hooks.tsx
+++ b/src/state/transactions/hooks.tsx
@@ -36,7 +36,22 @@ export function Updater() {
                 if (!receipt) {
                   dispatch(checkTransaction({ chainId, hash, blockNumber: globalBlockNumber }))
                 } else {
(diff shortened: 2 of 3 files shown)


src/components/CreatePool/index.tsx
```diff
@@ -35,16 +35,14 @@ function CreatePool({ history }) {
   const [step, setStep] = useState<number>(1)
 
   const pair = usePair(token0, token1)
-  const pairExists = // used to detect new exchange
-    pair && JSBI.notEqual(pair.reserve0.raw, JSBI.BigInt(0)) && JSBI.notEqual(pair.reserve1.raw, JSBI.BigInt(0))
 
   useEffect(() => {
-    if (token0Address && token1Address && pair && !pairExists) {
+    if (token0Address && token1Address && !pair) {
```
NoahZinsmeister, line 40: hm, what's the logic here? a bit hard to parse what the steps represent, a few comments here and there might helpful

src/components/SearchModal/index.tsx
```diff
@@ -184,6 +184,19 @@ function SearchModal({
...
   const [temporaryToken, setTemporaryToken] = useState<Token | null>()
+
+  // filters for ordering
+  const [activeFilter, setActiveFilter] = useState(FILTERS.BALANCES)
+
+  // toggle specific token import view
+  const [showTokenImport, setShowTokenImport] = useState(false)
+
+  // for sorting
+  const escapeStringRegexp = string => string
```
NoahZinsmeister, line 195: this should be: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Escaping

NoahZinsmeister: let's just make that a util function, make sure we're not doing this anywhere else, and remove the `escape-string-regexp` dependency

src/components/CreatePool/index.tsx
ianlapham: Yep replaced this with more clear steps for the views 

- one thing im considering is just linking straight to the regular add liquidity page. The only reason we keep the AddLiquidity component within this flow is so on the nav we can say "Create Pool" instead of 'Add Liquidity" on the last step (based on the route)

src/components/SearchModal/index.tsx
ianlapham: Yep adding this to utils

src/components/CreatePool/index.tsx
ianlapham: Another thing is - at some point we discussed that this flow is probably best to let them pick tokens and then add - but should there ever be an option to just create the pool without adding? currently you cant do that in the UI

haydenadams: > but should there ever be an option to just create the pool without adding? currently you cant do that in the UI

prob? kinda nice to cover the gas cost for someone else I guess if its like $5

ianlapham: for new pools we could maybe popup a smaller option that says "create without liquidity" or something (on the add page)

ianlapham: idk if the added confusion is worth helping the few people who want to create and not add though

PR #754, opened 2020-05-12
Rewrite tx store

moodysalem:


src/components/ExchangePage/index.tsx
```diff
@@ -373,20 +374,20 @@ function ExchangePage({ sendingInput = false, history, params }: ExchangePagePro
     const signer = await getProviderOrSigner(library, account)
     // get token contract if needed
     let estimate: Function, method: Function, args
-    if (tokens[Field.INPUT].equals(WETH[chainId])) {
-      ;(signer as any)
+    if (tokens[Field.INPUT].equals(WETH[chainId]) && signer instanceof JsonRpcSigner) {
```
NoahZinsmeister, line 377: is there ever a case where this isn't true?

moodysalem: I fixed it to always just call `getSigner` so the type check isn't necessary anymore

PR #755, opened 2020-05-12
Balance context overhaul

NoahZinsmeister:
replace `contexts/Balances.ts` with `data/BatchAccountBalances.tsx`

the basic idea is to mount a component for each token and pair that the app is aware of, and use that component to fetch the current user's balances in that token/pair. this works pretty well from the fetching side of things, but we also need to read/access this data, and that's tricky because our token/pair lists are dynamic. since our version of SWR doesn't allow direct cache access, and the version that does has [a bug](https://github.com/zeit/swr/issues/361) (the fix to which has been merged but not published), i had to use a big hack: call hooks in a `map`. to ensure that react doesn't yell at us for that, i keyed the parent component with a unique identifier of the thing we're mapping over, so that the # of hooks called per component instance is always the same


NoahZinsmeister:
another thing to note is that this pr arguably contains a regression, in that it doesn't stagger the initial calls (perhaps leading to more network spam on startup, although in practice SWR caching might lead to better outcomes here). it also has one of the same fundamental problems that the old solution had, which is that it doesn't batch calls, but that's out of scope for this PR.

PR #761, opened 2020-05-13
Balances rewrite

moodysalem:
This PR does balance fetching with the minimum number of fetches, and also the state is easily inspected via redux (can see how many listeners are attached to each address). In practice, it seems to work pretty well, as balances appear to update almost instantly when a TX is confirmed.

I'm not sure we need both SWR and redux. It's basically syntactic sugar for what's happening here, but moves the cache outside the store and we have less control and visibility over how we trigger updates/batch/invalidate the cache (e.g. we could parse the swap TX and optimistically update the balances when a tx is confirmed, or pessimistically clear balances for the involved tokens)


src/state/wallet/hooks.ts
```diff
@@ -0,0 +1,137 @@
...
+ * Returns a map of token addresses to their eventually consistent token balances for a single account.
+ */
+export function useTokenBalances(
+  address?: string,
+  tokens?: (Token | undefined)[]
+): { [tokenAddress: string]: TokenAmount } {
+  const dispatch = useDispatch<AppDispatch>()
+  const { chainId } = useWeb3React()
+
+  const validTokens: Token[] = useMemo(() => tokens?.filter(t => t && t.address && isAddress(t.address)) ?? [], [
```
NoahZinsmeister, line 59: `.filter(t => isAddress(t?.address)` should be ok

src/state/wallet/hooks.ts
```diff
@@ -0,0 +1,137 @@
...
+    return () => dispatch(stopListeningForBalance({ addresses }))
+  }, [addresses, dispatch])
+
+  const rawBalanceMap = useSelector<AppState>(({ wallet: { balances } }) => balances)
+
+  return useMemo(() => {
+    return addresses.reduce<{ [address: string]: JSBI }>((map, address) => {
+      const key = balanceKey({ address, chainId })
+      const { value } = rawBalanceMap[key] ?? {}
+      map[address] = JSBI.BigInt(value ?? 0)
```
NoahZinsmeister: shouldn't return 0 per offline discussion

src/state/wallet/hooks.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  return useMemo(() => {
+    if (!address || validTokens.length === 0) {
+      return {}
+    }
+    return (
+      validTokens.reduce<{ [address: string]: TokenAmount }>((map, token) => {
+        const key = balanceKey({ address, chainId, tokenAddress: token.address })
+        const { value } = rawBalanceMap[key] ?? {}
+        map[token.address] = new TokenAmount(token, JSBI.BigInt(value ?? 0))
```
NoahZinsmeister: shouldn't return 0 per offline discussion

src/state/wallet/hooks.ts
```diff
@@ -0,0 +1,137 @@
...
+export function useTokenBalanceTreatingWETHasETH(account?: string, token?: Token): TokenAmount | undefined {
+  const balances = useTokenBalancesTreatWETHAsETH(account, [token])
+  return token && token.address && balances?.[token.address]
+}
+
+// mimics useAllBalances
+export function useAllTokenBalancesTreatingWETHasETH(): { [account: string]: { [tokenAddress: string]: TokenAmount } } {
+  const { account } = useWeb3React()
+  const allTokens = useAllTokens()
+  const allTokensArray = useMemo(() => Object.values(allTokens ?? {}), [allTokens])
```
NoahZinsmeister, line 140: shouldn't this be `?? []`?

src/state/wallet/updater.tsx
```diff
@@ -0,0 +1,102 @@
...
+        }) ?? []
+      if (needsUpdate.length > 0) {
+        map[address] = needsUpdate
+      }
+      return map
+    }, {})
+  }, [activeTokenBalanceListeners, allBalances, chainId, lastBlockNumber])
+
+  useEffect(() => {
+    if (ethBalancesNeedUpdate.length === 0) return
```
NoahZinsmeister, line 65: do we need to add a `|| !!!library` condition here (and in the token updater) to trigger the early return if there's no library for whatever reason? or are we ok with seeing possible errors due to that in the `catch`?

PR #762, opened 2020-05-13
Changes to useReserves, escapreRegex moved to utils, code cleanup

ianlapham:


src/components/CreatePool/index.tsx
```diff
@@ -32,108 +38,109 @@ function CreatePool({ history }) {
...
 
   const pair = usePair(token0, token1)
-  const pairExists = // used to detect new exchange
-    pair && JSBI.notEqual(pair.reserve0.raw, JSBI.BigInt(0)) && JSBI.notEqual(pair.reserve1.raw, JSBI.BigInt(0))
 
+  // if both tokens selected but pair doesnt exist, enable button to create pair
   useEffect(() => {
-    if (token0Address && token1Address && pair && !pairExists) {
-      setStep(2)
+    if (token0Address && token1Address && !pair) {
```
NoahZinsmeister, line 47: should this be `pair === null`? if pair is undefined it's just loading right?

ianlapham: yep good catch - should only enable if its specifically null

PR #763, opened 2020-05-13
v2 slippage fix

NoahZinsmeister:


src/components/ExchangePage/index.tsx
```diff
@@ -219,12 +219,25 @@ function ExchangePage({ sendingInput = false, history, params }: ExchangePagePro
...
+  // for each hop in our trade, take away the "innate" price impact from 0.3% fees
+  // e.g. for 3 tokens/2 hops: 1 - ((1 - .03) * (1-.03))
+  const baseFee = basisPointsToPercent(10000 - 30)
+  const realizedFee = !trade
+    ? undefined
+    : trade.route.path.length === 2
+    ? baseFee
+    : basisPointsToPercent(10000).subtract(
+        new Array(trade.route.path.length - 2)
+          .fill(0)
```
moodysalem, line 231: is fill needed?

NoahZinsmeister: yeah, i tested it with an array of empties and it didn't trigger the reduce, sort of strange

moodysalem: guess we could have just used `trade.route.pairs`

NoahZinsmeister: yeah, might have been cleaner

PR #777, opened 2020-05-16
Dark theme tweaks and reworked helper text

callil:


haydenadams:
seems like my changes made this build fail

PR #792, opened 2020-05-18
Strict typescript

moodysalem:


src/state/application/hooks.ts
```diff
@@ -4,17 +4,17 @@ import { addPopup, PopupContent, removePopup, toggleWalletModal } from './action
 import { useSelector, useDispatch } from 'react-redux'
 import { AppState } from '../index'
 
-export function useBlockNumber() {
+export function useBlockNumber(): number | undefined {
   const { chainId } = useWeb3React()
 
-  return useSelector((state: AppState) => state.application.blockNumber[chainId])
+  return useSelector((state: AppState) => state.application.blockNumber[chainId ?? -1])
```
NoahZinsmeister, line 10: hm the other changes make sense, but i don't see why this one is necessary.

moodysalem: chainId can be `undefined`, and undefined can't be used to index into the object with numeric keys

NoahZinsmeister: yeah i see after reading more, just feels a little funky to have the `-1` in there

NoahZinsmeister: anyway it's fine, LGTM

PR #806, opened 2020-05-19
manual approve fallback for remove liquidity

### LINKED ISSUE #795: Pool/Remove Liquidity/Remove/Sign doesn't work with Ledger+MetaMask or AlphaWallet
**Bug Description**
Pool/Remove Liquidity/Remove/Sign doesn't work with Ledger+MetaMask or AlphaWallet (iOS and Andriod)

**Steps to Reproduce**
1. Go to Pool/Remove Liquidity/Remove/Sign
2. Click on Sign
Ledger+MetaMask: MetaMask popup signing page, but won't ask for signing on Ledger
AlphaWallet: nothing happens

BTW, what is this Signing for?

NoahZinsmeister:
closes #795


src/pages/Pool/RemoveLiquidity.tsx
```diff
@@ -128,6 +123,8 @@ export default function RemoveLiquidity({ token0, token1 }: { token0: string; to
   const pair = usePair(inputToken, outputToken)
   const pairContract: Contract = usePairContract(pair?.liquidityToken.address)
 
+  console.log(pair?.liquidityToken?.address)
```
cappaberra, line 126: Just curious if you meant to leave this logging statement in here?

src/components/WalletModal/index.tsx
```diff
@@ -2,8 +2,7 @@ import React, { useState, useEffect } from 'react'
 import ReactGA from 'react-ga'
 import styled from 'styled-components'
 import { isMobile } from 'react-device-detect'
-import { UnsupportedChainIdError } from '@web3-react/core'
-import { useWeb3React } from '../../hooks'
+import { UnsupportedChainIdError, useWeb3React } from '@web3-react/core'
```
moodysalem, line 5: huh, what did this break?

moodysalem: && should we rename our version to not overlap?

src/pages/Pool/RemoveLiquidity.tsx
NoahZinsmeister: nope, fixed

src/hooks/useApproveCallback.ts
```diff
@@ -9,53 +9,63 @@ import { computeSlippageAdjustedAmounts } from '../utils/prices'
 import { calculateGasMargin } from '../utils'
 import { useTokenContract, useWeb3React } from './index'
 
-// returns a function to approve the amount required to execute a trade if necessary, otherwise null
-export function useApproveCallback(trade?: Trade, allowedSlippage = 0): [undefined | boolean, () => Promise<void>] {
+// returns a boolean indicating whether approval is necessary, and a function to approve if it is
```
moodysalem, line 12: what does undefined mean for the boolean? can we be more explicit `false | [boolean, fn]`

src/pages/Pool/RemoveLiquidity.tsx
```diff
@@ -389,68 +400,109 @@ export default function RemoveLiquidity({ token0, token1 }: { token0: string; to
...
-        sigInputs[1],
-        sigInputs[2]
-      ]
+
+    const token0IsETH = tokens[Field.TOKEN0].equals(WETH[chainId])
+    const oneTokenIsETH = token0IsETH || tokens[Field.TOKEN1].equals(WETH[chainId])
+
+    let estimate, method, args
+    // we have approval, use normal remove liquidity
+    if (mustApprove === false) {
```
moodysalem, line 438: false for undefined, but seems intentional

src/components/WalletModal/index.tsx
NoahZinsmeister: it was sometimes activating account connectors under the network key, which messed some stuff up. yeah, i'm down to rename, will do

moodysalem: this explains the issues i was running into ln. thanks!

src/hooks/useApproveCallback.ts
NoahZinsmeister: `undefined` means the current approval amount is yet to be fetched. `false | [boolean, fn]` wouldn't play nicely with destructuring. if we want to be more explicit we could make the first arg an enum?

src/pages/Pool/RemoveLiquidity.tsx
NoahZinsmeister: yeah this is a bit of a weird one, our disabled states should prevent `mustApprove` from ever being `undefined` at this line

src/hooks/useApproveCallback.ts
moodysalem: yeah that's better, but i'm also ok with what it is now too

NoahZinsmeister: for the record, made it an enum and as a bonus we now get pending state from this hook too

PR #808, opened 2020-05-19
add route visualization

### LINKED ISSUE #793: Show swap route
UI should reflect the swap route in advanced or confirmation modal.

NoahZinsmeister:
closes #793


moodysalem:
code lgtm, stylewise I think there was a border style and also should we only show if route > 2 pairs?

PR #815, opened 2020-05-20
replace commas with periods

NoahZinsmeister:
fixes a language localization problem where some mobile users were unable to type period as their decimal separator


src/components/NumericalInput/index.tsx
```diff
@@ -65,8 +62,10 @@ export const Input = React.memo(function InnerInput({
       {...rest}
       value={value}
       onChange={event => {
-        enforcer(event.target.value)
+        // replace commas with periods, because uniswap exclusively uses period as the decimal separator
+        enforcer(event.target.value.replace(/,/g, '.'))
```
moodysalem, line 66: it could also help to have a pattern attribute on the input
https://www.w3schools.com/tags/att_input_pattern.asp

PR #817, opened 2020-05-20
Make the hooks directory strict

moodysalem:


src/hooks/Tokens.ts
```diff
@@ -52,7 +52,7 @@ export function useAllTokens(): { [address: string]: Token } {
             tokenMap[token.address] = token
             return tokenMap
           },
-          { ...ALL_TOKENS[chainId] }
+          { ...ALL_TOKENS[chainId as ChainId] }
```
NoahZinsmeister, line 55: oh this can actually just be `ALL_TOKENS[chainId as ChainId]` i believe. the way it is now is a relic of when we used check tokens against membership in `ALL_TOKENS` to determine whether or not they were user added. that was flawed because our code modifies `ALL_TOKENS`, and this was just a debugging attempt which doesn't logically change anything.

NoahZinsmeister:
pending CI and my comment if you want to address it, this can be merged

src/hooks/Tokens.ts
moodysalem: It can't because in the `reduce` we are modifying the initial value (meaning the ALL_TOKENS map) and we don't want to make a copy on every iteration of reduce. This actually fixed an issue where you couldn't remove tokens from the search modal.

Added a comment to explain

PR #828, opened 2020-05-22
feat(expert mode): Add expert mode

callil:
Start of expert mode.

Todo:
- [x] use browser alert box to force user to type "confirm" when turning on
- [x] Add a cool icon to your account in the header to indicate you are an expert 🥇 
- [x] Fire a "transaction submitted" notification when successful only if confirm modal is bypassed
- [x] Decide where the toggle should go
- [x] Decide whether we need or want this - or if a "cleaner" solution is possible


moodysalem:
ui comments on first look
- can't tell toggle is off in dark mode with the subtle background color
- help text seems more specific than the name 'expert mode', is there anything else we want this to do?

moodysalem:
This feels really good

haydenadams:
I'm getting a custom slippage: NAN bug. I would rename it to "Slippage Tolerance" and default to 0.5%

Screen Shot 2020-06-11 at 3 42 12 PM

src/components/Settings/index.tsx
```diff
@@ -0,0 +1,253 @@
...
+              <StyledCloseIcon onClick={() => setShowConfirmation(false)} />
+            </RowBetween>
+            <Break />
+            <AutoColumn gap="lg" style={{ padding: '0 2rem' }}>
+              <Text fontWeight={500} fontSize={20}>
+                Expert mode turns off the confirm transaction prompt and allows high slippage trades that often result
+                in bad rates and lost funds.
+              </Text>
+              <Text fontWeight={600} fontSize={20}>
+                ONLY USE THIS MODE IF YOU KNOW WHAT YOU ARE DOING.
```
moodysalem, line 176: this copy feels a little extra, not sure it's going to get the message across any more than the confirmation modal and prompt

src/components/Settings/index.tsx
```diff
@@ -0,0 +1,253 @@
...
+            </AutoColumn>
+          </AutoColumn>
+        </ModalContentWrapper>
+      </Modal>
+      <StyledMenuButton onClick={toggle}>
+        <StyledMenuIcon />
+        {expertMode && (
+          <EmojiWrapper>
+            <span role="img" aria-label="wizard-icon">
+              🧙
```
moodysalem, line 201: this is cute but we can't change the color--can we do something that matches our color scheme a little better?

haydenadams: 🐳 ?

src/components/Settings/index.tsx
callil: Really don't want normies to blow past here and the all caps is a bit of an eye catcher. We could change the copy though.

src/components/Settings/index.tsx
moodysalem: or if we want to be able to change the color with the theme
https://feathericons.com/?query=coffee
https://openmoji.org/library/#search=wizard

src/components/Settings/index.tsx
moodysalem: make them type 
"I know what I'm doing"

callil: I still like the yelling in your face modal - the system stuff feels way too default and non threatening. As long as we keep the toggle->modal->type to confirm flow I'll be happy.

src/components/SlippageTabs/index.tsx
```diff
@@ -146,15 +142,14 @@ export default function SlippageTabs({ rawSlippage, setRawSlippage, deadline, se
...
-      </RowFixed>
-
-      <SlippageSelector>
+    <AutoColumn gap="md">
+      <AutoColumn gap="sm">
+        <RowFixed>
+          <TYPE.black fontWeight={400} fontSize={14} color={theme.text2}>
+            Set slippage tolerance
+          </TYPE.black>
+          <QuestionHelper text="Your transaction will revert if the price changes unfavorably by more than this percentage. Default is 0.5%." />
```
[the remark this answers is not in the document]
ianlapham: Yep removed it

src/components/Settings/index.tsx
ianlapham: Down to use a feather icon or a neutral colored emoji - the skull is pretty mono

src/state/user/reducer.ts
```diff
@@ -50,13 +62,13 @@ function pairKey(token0Address: string, token1Address: string) {
 
 const initialState: UserState = {
   lastVersion: '',
-
   userDarkMode: null,
   matchesDarkMode: false,
-
+  userExpertMode: false,
+  userSlippageTolerance: INITIAL_ALLOWED_SLIPPAGE,
+  userDeadline: DEFAULT_DEADLINE_FROM_NOW,
```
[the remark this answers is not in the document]
ianlapham: Added a check on new version to set to defaults

src/components/swap/PriceSlippageWarningCard.tsx
```diff
@@ -1,30 +0,0 @@
-import { Percent } from '@uniswap/sdk'
```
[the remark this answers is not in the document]
ianlapham: Yeah decided the rest of the stuff was enough

src/components/Header/index.tsx
```diff
@@ -204,13 +209,18 @@ export default function Header() {
...
+            <BalanceWrapper>
+              {account && userEthBalance ? (
+                <Text style={{ flexShrink: 0 }} pl="0.75rem" pr="0.5rem" fontWeight={500}>
+                  {userEthBalance?.toSignificant(4)} ETH
+                </Text>
+              ) : null}
+            </BalanceWrapper>
             <Web3Status />
           </AccountElement>
+          <div style={{ pointerEvents: 'auto' }}>
```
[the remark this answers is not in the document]
ianlapham: Is this in the latest version?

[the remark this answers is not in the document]
ianlapham: Got it - yep can delete

src/state/user/reducer.ts
```diff
@@ -75,6 +87,16 @@ export default createReducer(initialState, builder =>
             delete state.tokens[ChainId.MAINNET][WETH[ChainId.MAINNET].address]
           }
         }
```
moodysalem, line 89: can we delete this block? it's no longer relevant (and also wrong)

src/state/user/reducer.ts
```diff
@@ -75,6 +87,16 @@ export default createReducer(initialState, builder =>
...
+
+        // slippage isnt being tracked in local storage, reset to default
+        if (!!!state.userSlippageTolerance) {
```
moodysalem, line 92: what if the user sets to 0? we should probably do `if (typeof state.userSlippageTolerance !== 'number')`

src/state/user/reducer.ts
```diff
@@ -75,6 +87,16 @@ export default createReducer(initialState, builder =>
...
+          state.userSlippageTolerance = INITIAL_ALLOWED_SLIPPAGE
+        }
+
+        // deadline isnt being tracked in local storage, reset to default
+        if (!!!state.userDeadline) {
```
moodysalem, line 97: typeof check preferred here, even though 0 deadline won't ever work

src/utils/prices.ts
```diff
@@ -42,7 +43,7 @@ export function computeTradePriceBreakdown(
 // computes the minimum amount out and maximum amount in for a trade given a user specified allowed slippage in bips
 export function computeSlippageAdjustedAmounts(
   trade: Trade,
-  allowedSlippage: number
+  allowedSlippage: number = INITIAL_ALLOWED_SLIPPAGE
```
moodysalem, line 46: do we need a default? seems like we'd be hiding a bug

ianlapham: Can remove this - had this in before adding the reset in the reducer

PR #841, opened 2020-05-28
Style updates, Approve UI updates

ianlapham:
- Pull in style tweaks on wallet modal and txns
- Refactor approve flow to be two buttons


src/components/AccountDetails/index.tsx
```diff
@@ -94,34 +94,13 @@ const YourAccount = styled.div`
...
-  align-items: center;
-`
-
 const LowerSection = styled.div`
   ${({ theme }) => theme.flexColumnNoWrap}
   padding: 1.5rem;
   flex-grow: 1;
   overflow: auto;
   background-color: ${({ theme }) => theme.bg2};
+  /* border-top: 1px solid ${({ theme }) => theme.bg2}; */
```
moodysalem, line 103: just delete

src/components/Modal/index.tsx
```diff
@@ -167,7 +171,7 @@ export default function Modal({
                         hidden={true}
                         minHeight={minHeight}
                         maxHeight={maxHeight}
-                        mobile={isMobile}
+                        mobile={isMobile ?? undefined}
```
moodysalem, line 174: why is this ever undefined?

src/components/Modal/index.tsx
```diff
@@ -191,9 +195,15 @@ export default function Modal({
                 style={props}
                 onDismiss={onDismiss}
                 initialFocusRef={initialFocusRef}
-                mobile={false}
+                mobile={undefined}
```
moodysalem, line 198: also why this

src/components/Modal/index.tsx
```diff
@@ -191,9 +195,15 @@ export default function Modal({
...
               >
-                <StyledDialogContent hidden={true} minHeight={minHeight} maxHeight={maxHeight} isOpen={isOpen}>
+                <StyledDialogContent
+                  hidden={true}
+                  minHeight={minHeight}
+                  maxHeight={maxHeight}
+                  isOpen={isOpen}
+                  aria-label=""
```
moodysalem, line 205: is this label required with attrs above? can we do just one?

src/pages/Send/index.tsx
```diff
@@ -95,6 +95,15 @@ export default function Send({ location: { search } }: RouteComponentProps) {
   // check whether the user has approved the router on the input token
   const [approval, approveCallback] = useApproveCallbackFromTrade(bestTrade, allowedSlippage)
 
+  // check if user has gone through approval process, used to show two step buttons, reset on token change
+  const [approvalSubmitted, setApprovalSubmitted] = useState<boolean>(false)
+  // mark when a user has submitted an approval, reset onTokenSelection for input field
+  useEffect(() => {
+    if (approval === ApprovalState.PENDING) {
+      setApprovalSubmitted(true)
```
moodysalem, line 103: is this a state variable so they can swap right after they submit the approve?

src/components/Modal/index.tsx
ianlapham: I did this because warning was firing when isMobile was false, but was weird because it should be expecting a boolean - it may have to do with the way teh styled components is passing props down ill take another look

src/components/Modal/index.tsx
ianlapham: same as above

src/components/Modal/index.tsx
ianlapham: Nope can remove this

src/pages/Send/index.tsx
ianlapham: Yeah we only need this to keep the two step UI around when theyre in this flow - otherwise when the approval changes it will go back to just one swap button (but we want the two step to show up instead here)

src/components/Modal/index.tsx
moodysalem: If you do a pull, I touched the same thing in this file to address the same problem earlier, did it slightly differently though by just not passing through the mobile prop

moodysalem:
this looks a little strange
image

PR #855, opened 2020-06-03
use single call to get token information

moodysalem:
~i think bytes32 returndata if decoded as a string will probably work fine~


moodysalem:
> imo we should only include `name` and `symbol`
done

PR #866, opened 2020-06-05
support proposed new FoT-capable methods

### LINKED ISSUE #835: Support of fee-on-transfer tokens
The Uniswap v2 front-end doesn't support deflationary tokens. (tokens that burn a % of each transaction, like the one I'm contributing to, "BSOV" (https://bsov.io).
BSOV Token Request: https://github.com/Uniswap/uniswap-frontend/issues/836
Deflation is very important for the deflationist community, and many of us wish that BSOV keeps on burning, while trading on-chain.

Possibly like @NoahZinsmeister described to me in the Discord: A feature in the Uniswap v2 front-end that lets the user "set their slippage to whatever % deflation they expect from their token".

**Describe alternatives you've considered**
When there is a token request of a deflationary token on Github, then there could be a custom parameter that could be prefixed for easier implementation of deflationary tokens in the Uniswap v2 front-end, so the user would not need to set their slippage for **every** trade, to match the % deflation.

**Additional context**
While I don't know how to develop these things, then there are several things I can help you with, like paying you some BSOV or doing tasks revolving Media & Communication (like design and videos).

NoahZinsmeister:
closes #835

### DIFF
--- a/src/hooks/useSwapCallback.ts
+++ b/src/hooks/useSwapCallback.ts
@@ -90,11 +90,13 @@ export function useSwapCallback(
         chainId as ChainId
       )
 
-      let estimate, method: Function, args: Array<string | string[] | number>, value: BigNumber | null
+      // let estimate: Function, method: Function,
+      let methodNames: string[],
+        args: Array<string | string[] | number>,
+        value: BigNumber | null = null
       switch (swapType) {
         case SwapType.EXACT_TOKENS_FOR_TOKENS:
-          estimate = routerContract.estimateGas.swapExactTokensForTokens
-          method = routerContract.swapExactTokensForTokens
+          methodNames = ['swapExactTokensForTokens', 'swapExactTokensForTokensSupportingFeeOnTransferTokens']
           args = [
             slippageAdjustedInput.raw.toString(),
             slippageAdjustedOutput.raw.toString(),
@@ -102,11 +104,9 @@ export function useSwapCallback(
             recipient,
             deadlineFromNow
           ]
-          value = null
           break
         case SwapType.TOKENS_FOR_EXACT_TOKENS:
-          estimate = routerContract.estimateGas.swapTokensForExactTokens
-          method = routerContract.swapTokensForExactTokens
+          methodNames = ['swapTokensForExactTokens']
           args = [
             slippageAdjustedOutput.raw.toString(),
             slippageAdjustedInput.raw.toString(),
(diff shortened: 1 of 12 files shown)


src/hooks/useSwapCallback.ts
```diff
@@ -90,112 +90,148 @@ export function useSwapCallback(
...
-          }
+        )
+      )
 
-          return response.hash
-        })
-        .catch(error => {
-          console.error(`Swap or gas estimate failed`, error)
-          throw error
+      const indexOfSuccessfulEstimation = safeGasEstimates.findIndex(safeGasEstimate =>
```
moodysalem, line 160: https://caniuse.com/#search=findindex

src/hooks/useSwapCallback.ts
```diff
@@ -90,112 +90,148 @@ export function useSwapCallback(
...
-                slippageAdjustedInput.toSignificant(3) +
-                ' ' +
-                trade.inputAmount.token.symbol +
-                ' for ' +
-                slippageAdjustedOutput.toSignificant(3) +
-                ' ' +
-                trade.outputAmount.token.symbol
+      const safeGasEstimates = await Promise.all(
+        methodNames.map(methodName =>
+          routerContract.estimateGas[methodName](...args, value ? { value } : {})
```
moodysalem, line 152: a hook for `useGasEstimate` would be generally useful, but would require constructing all these args on every render, which is probably ok for the improved UX (we can disable the button and show an error if we can't get a gas estimate from either swap)

src/hooks/useSwapCallback.ts
NoahZinsmeister: oof, will change this

src/hooks/useSwapCallback.ts
NoahZinsmeister: yeah i like it, i'll start to refactor

src/hooks/useSwapCallback.ts
NoahZinsmeister: wait you linked TypedArray, Array is fine: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex

we're using findIndex in 2 other places in the codebase as well

moodysalem: thought i'd seen it elsewhere, sgtm.

src/hooks/useSwapCallback.ts
NoahZinsmeister: going to track this in #872

PR #868, opened 2020-06-05
Create iw.json

CryptoCatVC:

Hebrew locale


NoahZinsmeister:
what's the difference between this and #867?

CryptoCatVC:
> what's the difference between this and #867?

@NoahZinsmeister in #867
there is a missing " on line 45
on yourPoolShare

PR #883, opened 2020-06-10
feat(v1-support): Enable executing swaps on v1 or v2

moodysalem:


src/state/swap/hooks.ts
```diff
@@ -90,76 +93,121 @@ export function useDerivedSwapInfo(): {
...
-    error = error ?? 'Enter an amount'
+  if (!tokens[Field.INPUT] || !tokens[Field.OUTPUT]) {
+    error = error ?? 'Select a token'
   }
 
-  const [balanceIn, amountIn] = [tokenBalances[Field.INPUT], parsedAmounts[Field.INPUT]]
-  if (balanceIn && amountIn && balanceIn.lessThan(amountIn)) {
-    error = 'Insufficient ' + amountIn.token.symbol + ' balance'
-  }
+  // this check is incorrect, it should check against the maximum amount in
```
moodysalem, line 135: this check was basically deleted, users can now approve/swap with insufficient balance, which i think is an ok degradation (not important to solve since this check was wrong).

PR #885, opened 2020-06-10
feat(migration): improve v1 migration flow and add remove

NoahZinsmeister:


NoahZinsmeister:
UX improvements for handling V1 -> V2 liquidity migration, and direct V1 removal.

to add liquidity on rinkeby, use the link from: https://github.com/Uniswap/uniswap-frontend/releases/tag/v1.0.0. I use rinkeby DAI for testing: 0xc7AD46e0b8a400Bb3C915120d284AafbA8fc4735 (note that this isn't the default in the V1 rinkeby interface).

some screenshots:

Screen Shot 2020-06-11 at 1 13 47 PM

Screen Shot 2020-06-11 at 1 15 03 PM

moodysalem:
I like the old arrows in the links better than these long ones
image

wrapping issue
image

do we need to say v2 here?
image

src/pages/MigrateV1/index.tsx
```diff
@@ -17,14 +17,27 @@ import { EmptyState } from './EmptyState'
...
 
 export default function MigrateV1({ history }: RouteComponentProps) {
   const theme = useContext(ThemeContext)
   const { account, chainId } = useActiveWeb3React()
 
   const [tokenSearch, setTokenSearch] = useState<string>('')
   const handleTokenSearchChange = useCallback(e => setTokenSearch(e.target.value), [setTokenSearch])
-  useToken(tokenSearch)
+
+  // automatically add the search token
```
moodysalem, line 31: why do we automatically add here?

src/hooks/useIsDefaultToken.ts
```diff
@@ -0,0 +1,11 @@
+import { Token } from '@uniswap/sdk'
+
+import { ALL_TOKENS } from '../constants/tokens'
+
+export function isDefaultToken(token?: Token): boolean {
+  return Boolean(token && ALL_TOKENS[token.chainId]?.[token.address])
+}
+
+export default function useIsDefaultToken(token?: Token): boolean {
```
moodysalem, line 9: I don't love wrapping this in a hook, it makes it seem like you have to call it in order (like with other hooks)

src/pages/MigrateV1/index.tsx
NoahZinsmeister: otherwise if they refresh or page forward/backward or try to add multiple custom tokens the entry disappears, which feels a bit weirder than automatically adding them imo.

PR #887, opened 2020-06-11
fix approval bugs

NoahZinsmeister:
clear signature on input in remove


NoahZinsmeister:
fyi @ianlapham little bug with that PR earlier, this fixes it (and also clears signatures when users change the input amount)

PR #903, opened 2020-06-22
upgrade to walletconnect v1

NoahZinsmeister:
deprecates #901


moodysalem:
@NoahZinsmeister  if you add a prefix a la `feat(walletconnect): ` to the commit message, it will trigger a minor version bump and add it to release notes

PR #905, opened 2020-06-24
fix(send page): support swap + send query parameters on send page

ianlapham:
- currently on send page we force a token selection of null when sendingWithSwap isnt true (this overwrites output token info from url queries) 
- changes the way we remove the swap with send so this doesn't happen
- adds an override options to query param hook so we dont autopopulate the output with WETH if they want to link to a normal 1 token send


moodysalem:
the query parameters for send are really unintuitive, feels more and more like we should remove swapping from the send page and add a link to add a recipient field in the swap page and have a recipient query parameter

something like

`(+ Add a recipient)`

under the to field in the swap @callil 

Then for send page we can just have token=x, amount=y, recipient=z and it's never confusing

moodysalem:
i also would not be opposed to an intermediate state where we do not support send+swap at all and have intuitive query params in send, but i know this is not a popular opinion

moodysalem:
this got closed automatically when the v2 branch was deleted (in favor of master)

PR #915, opened 2020-07-01
Renames REP to REPv1.

MicahZoltu:
https://www.augur.net/blog/v2-launch/

TL;DR: Augur v2 launch is coming up and will introduce a new REP token.  FF has requested all exchanges rename REP to REPv1 to avoid confusion.

Going forward, REP tokens will contain versioning in their name/symbol on chain so this should be a one-time "fix" for Augur v1 REP.


MicahZoltu:
Preview looks good, and as a bonus you can now find REP by typing `Augur` in the search box.

MicahZoltu:
cc @NoahZinsmeister since we discussed in Discord briefly.

PR #945, opened 2020-07-10
Update mainnet.ts

pgebheim:
Keep REP Ticker / Description for v1 during/after v2 Launch.


moodysalem:
@MicahZoltu

MicahZoltu:
Against all advice I have given, Forecast Foundation believes that going back on the original plan and naming REPv1 just REP is a good idea.  So while I wholeheartedly disagree with the decisions that lead up to this PR, it *is* the current narrative from FF (published as of a few hours ago).

moodysalem:
This should go in https://github.com/Uniswap/default-token-list

PR #947, opened 2020-07-11
feat(weth): support WETH across the site and use sdk 3.0

moodysalem:


callil:
re: removing liquidity when one token is WETH perhaps we can do something quite simple for now like this: 
Screen Shot 2020-07-15 at 10 09 34 AM

callil:
Small visual but on the account: 
Screen Shot 2020-07-15 at 11 12 13 AM

Height should match the other buttons

PR #967, opened 2020-07-16
disable VersionSwitch

PokimLee:
limits users to use uniswap in version2


moodysalem:
We no longer provide access to Uniswap V1 interface, so we want to give people the opportunity to swap on V1 when the price is better (and always notify/warn them when it's not)

PR #979, opened 2020-07-20
set default tokens with Router Query params

PokimLee:
Now, we can set the default tokens(From/To) with Router Query params

just open uniswap like this:
https://app.uniswap.org/#/swap/?from=0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2&to=0xdAC17F958D2ee523a2206206994597C13D831ec7

and you will see:
image

### DIFF
--- a/src/pages/Swap/util.js
+++ b/src/pages/Swap/util.js
@@ -1,0 +1,46 @@
+// 获取路由中某个参数
+const getQueryVariable = variable => {
+  const query = window.location.href.split('?')[1]
+  if (!query) return false
+  const vars = query.split('&')
+  for (let i = 0; i < vars.length; i++) {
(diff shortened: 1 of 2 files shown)


moodysalem:
This is already possible using `inputCurrency` and `outputCurrency`
https://uniswap.org/docs/v2/interface-integration/custom-interface-linking/#swap-page

src/pages/Swap/util.js
```diff
@@ -0,0 +1,46 @@
+// 获取路由中某个参数
+const getQueryVariable = variable => {
+  const query = window.location.href.split('?')[1]
```
moodysalem, line 3: we never access window.location directly, we always use the react router provided location

PR #983, opened 2020-07-20
feat(token lists): implement the uniswap default list as a token list

moodysalem:


moodysalem:
> generally LGTM so far
> 
> ideally we'd be able to avoid using `unpkg` for the default list, and instead be able to resolve the list over IPFS via a contenthash record on an ENS name or directly via IPNS (+DNSLink), but it's not necessarily a blocker

agreed, https://github.com/Uniswap/default-token-list/issues/2

moodysalem:
forgot imported tokens...

PR #1005, opened 2020-07-27
improvement(token warnings): show better warnings for imported tokens

ianlapham:
- more extreme warning UI for imported tokens

### DIFF
--- a/cypress/integration/token-warning.ts
+++ b/cypress/integration/token-warning.ts
@@ -1,13 +1,19 @@
 describe('Warning', () => {
-  beforeEach(() => cy.visit('/swap?outputCurrency=0x0a40f26d74274b7f22b28556a27b35d97ce08e0a'))
+  beforeEach(() => {
+    cy.clearLocalStorage()
+    cy.visit('/swap?outputCurrency=0x0a40f26d74274b7f22b28556a27b35d97ce08e0a')
+  })
   it('Check that warning is displayed', () => {
     cy.get('.token-warning-container').should('be.visible')
   })
   it('Check that warning hides after button dismissal.', () => {
-    cy.get('.token-dismiss-button').click({ force: true })
+    cy.get('.token-dismiss-button').click()
     cy.get('.token-warning-container').should('not.be.visible')
   })
-  it('Check that supression persists across sessions.', () => {
+  it('Check supression persists across sessions.', () => {
+    cy.get('.token-warning-container').should('be.visible')
+    cy.get('.token-dismiss-button').click()
+    cy.reload()
     cy.get('.token-warning-container').should('not.be.visible')
   })
 })
--- a/cypress/integration/token-warning.ts
+++ b/cypress/integration/token-warning.ts
@@ -1,0 +1,13 @@
+describe('Warning', () => {
+  beforeEach(() => cy.visit('/swap?outputCurrency=0x0a40f26d74274b7f22b28556a27b35d97ce08e0a'))
+  it('Check that warning is displayed', () => {
+    cy.get('.token-warning-container').should('be.visible')
+  })
+  it('Check that warning hides after button dismissal.', () => {
+    cy.get('.token-dismiss-button').click({ force: true })
+    cy.get('.token-warning-container').should('not.be.visible')
+  })
+  it('Check that supression persists across sessions.', () => {
+    cy.get('.token-warning-container').should('not.be.visible')
+  })
+})
--- a/src/components/TokenWarningCard/index.tsx
+++ b/src/components/TokenWarningCard/index.tsx
@@ -14,6 +14,7 @@ import { AutoRow, RowBetween } from '../Row'
 import { AutoColumn } from '../Column'
 import { AlertTriangle } from 'react-feather'
 import { ButtonError } from '../Button'
+import { useTokenWarningDismissal } from '../../state/user/hooks'
 
 const Wrapper = styled.div<{ error: boolean }>`
   background: ${({ theme }) => transparentize(0.6, theme.white)};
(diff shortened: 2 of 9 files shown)


src/state/user/hooks.tsx
```diff
@@ -165,7 +165,7 @@ export function usePairAdder(): (pair: Pair) => void {
  * Returns whether a token warning has been dismissed and a callback to dismiss it,
  * iff it has not already been dismissed and is a valid token.
  */
-export function useTokenWarningDismissal(chainId?: number, token?: Token): [boolean, null | (() => void)] {
+export function useTokenWarningDismissal(chainId?: number, token?: Token | Currency): [boolean, null | (() => void)] {
```
moodysalem, line 168: just `Currency` for the type is fine, `Token` extends `Currency`

src/components/TokenWarningCard/index.tsx
```diff
@@ -94,48 +63,82 @@ export default function TokenWarningCard({ token, ...rest }: TokenWarningCardPro
...
+            padding="0.5rem 1rem"
+            style={{
+              borderRadius: '10px'
+            }}
+            onClick={() => {
+              dismissToken0 && dismissToken0()
+              dismissToken1 && dismissToken1()
+            }}
+          >
+            <TYPE.body color="white">I understand</TYPE.body>
```
moodysalem, line 137: should this be like text1 or something?

src/components/TokenWarningCard/index.tsx
```diff
@@ -94,48 +63,82 @@ export default function TokenWarningCard({ token, ...rest }: TokenWarningCardPro
...
+      <AutoColumn gap="lg">
+        <AutoRow gap="6px">
+          <AlertTriangle stroke="#F82D3A" />
+          <TYPE.main color={'#F82D3A'}>Token imported</TYPE.main>
+        </AutoRow>
+        <TYPE.body color={'#F82D3A'}>
+          Anyone can create and name any ERC20 token on Ethereum, including creating fake versions of existing tokens
+          and tokens that claim to represent projects that do not have a token.
+        </TYPE.body>
+        <TYPE.body color={'#F82D3A'}>
```
moodysalem, line 110: hard coded colors here

src/components/TokenWarningCard/index.tsx
```diff
@@ -2,80 +2,49 @@ import { Currency, Token } from '@uniswap/sdk'
...
 
-const DUPLICATE_NAME_HELP_TEXT = `${HELP_TEXT} This token has the same name or symbol as another token in your list.`
-
 interface TokenWarningCardProps extends PropsOfExcluding<typeof Wrapper, 'error'> {
   token?: Token
+  dismissed: boolean
 }
 
-export default function TokenWarningCard({ token, ...rest }: TokenWarningCardProps) {
+export default function TokenWarningCard({ token, dismissed, ...rest }: TokenWarningCardProps) {
```
moodysalem, line 40: prefer not to pass props to components that determine whether they are rendered, but rather make that determination in the parent component (better performance is one reason, but this is better for readability imo)

src/components/TokenWarningCard/index.tsx
```diff
@@ -94,48 +63,82 @@ export default function TokenWarningCard({ token, ...rest }: TokenWarningCardPro
...
-export function TokenWarningCards({ currencies }: { currencies: { [field in Field]?: Currency } }) {
+export function TokenWarningCards({
+  currencies,
+  dismissedToken0,
+  dismissedToken1,
+  dismissToken0,
+  dismissToken1
+}: {
+  currencies: { [field in Field]?: Currency }
+  dismissedToken0: boolean
```
moodysalem, line 94: should this component just look up the state itself?

haydenadams:
On deploy preview I get a warning no matter what token I select

src/components/TokenWarningCard/index.tsx
ianlapham: I use white because we actually want it to be white for both light and dark mode (text colors switch from black to white)

src/components/TokenWarningCard/index.tsx
ianlapham: the reason i did it in swap page was so that we could tell the AppBody if its disabled or not, i guess if we looked up state in here could pass it back up to parent with a callback, but thats kinda weird right

ianlapham:
> On deploy preview I get a warning no matter what token I select

Fixed now - question is do we want to compare to our default list or the currently loaded list

ianlapham:
> sorry i ended up making multiple comments
> 
> it would be great if we could add an integration test and/or unit tests for the reducer state

So test that the dismissal works for this?

moodysalem:
> So test that the dismissal works for this?

yeah an integration test like this would be good if you don't want it to break
- load page with token not on default list
- check dismissal shows
- click dismiss, check it's gone
- visit again, check it's not rendered (persisted)

ianlapham:
^^^
Cool i can give those tests a shot

cypress/integration/token-warning.ts
```diff
@@ -0,0 +1,13 @@
+describe('Warning', () => {
+  beforeEach(() => cy.visit('/swap?outputCurrency=0x0a40f26d74274b7f22b28556a27b35d97ce08e0a'))
+  it('Check that warning is displayed', () => {
+    cy.get('.token-warning-container').should('be.visible')
+  })
+  it('Check that warning hides after button dismissal.', () => {
+    cy.get('.token-dismiss-button').click({ force: true })
+    cy.get('.token-warning-container').should('not.be.visible')
+  })
+  it('Check that supression persists across sessions.', () => {
```
moodysalem, line 10: a good practice is that tests should not depend on execution order (i.e. you should be able to run this test alone and have it pass), you could clear local storage between visits in the `beforeEach`

src/components/TokenWarningCard/index.tsx
moodysalem: def don't want to pass anything back to the callback, just prefer calling `useSelector` and `useDispatch` in components that need it, over passing it via props

if we're worried about performance, we can `React.memo` this component

cypress/integration/token-warning.ts
```diff
@@ -0,0 +1,13 @@
+describe('Warning', () => {
+  beforeEach(() => cy.visit('/swap?outputCurrency=0x0a40f26d74274b7f22b28556a27b35d97ce08e0a'))
+  it('Check that warning is displayed', () => {
+    cy.get('.token-warning-container').should('be.visible')
+  })
+  it('Check that warning hides after button dismissal.', () => {
+    cy.get('.token-dismiss-button').click({ force: true })
```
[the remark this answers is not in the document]
ianlapham: Nope no reason to

cypress/integration/token-warning.ts
ianlapham: Cool added a change for this

src/components/TokenWarningCard/index.tsx
ianlapham: just removed these as props

PR #1015, opened 2020-07-30
improvement(swap): Better swap errors for FoT

### LINKED ISSUE #701: Self-destructing eth into an exchange with no liquidity breaks frontend
**Bug Description**
You can self-destruct a contract into an exchange with no liquidity leaving it with a non-zero ETH balance. This does not affect the `addLiquidity` function, but it breaks the "Add Liquidity" feature of the frontend.

**Steps to Reproduce**
1. Create a new exchange
2. Write a contract that self-destructs and sends some ether to the exchange contract.
3. Try adding liquidity on the frontend, it doesn't work.

**Expected Behavior**
It should work.

**Additional Context**
This has happened to UMA in the wild. Here is the [transaction](https://etherscan.io/tx/0x3e1a5bc5903fb82fc08259debb1e6c427321186f49bbe295af3c61b21a183fb9) that caused it.
Screen Shot 2020-04-29 at 22 52 43

These [checks](https://github.com/Uniswap/uniswap-frontend/blob/1fe266025e5e2f7188fc5450954c080f3201566d/src/pages/Pool/AddLiquidity.js#L243) seem to be relevant to the problem. They should be aligned with the assumptions of the contract and check for liquidity token supply instead.

### LINKED ISSUE #820: Wallet modal small ux improvements
Copy change in the wallet modal:

Connect To A Wallet -> Connect to a wallet

Also the link under "Learn more about wallets goes to an outdated link. It should go to: 

https://ethereum.org/wallets/

### LINKED ISSUE #871: improve UI for failing swaps (+ add/remove liquidities)

### LINKED ISSUE #988: Alert users about price changes in the confirm modal
Right now if the user rests on the confirm modal for too long a block confirmation can trigger a price update. Right now it can be hard to catch visually. Let's add a warning inline and disable the button until the user accepts the new price.

Initial concept:
Screen Shot 2020-07-21 at 3 49 09 PM

It's possible that forcing another click is not the best UX so if needed it could just be the alert inline. If we go down that route, we should disable the button for 1-2 seconds after the price updates so as to avoid a click that occurs a millisecond after.

moodysalem:
fixes #820 
fixes #988 
fixes #871 for swaps (create new issue for add/remove liquidity)
fixes #701


src/connectors/NetworkConnector.ts
```diff
@@ -22,19 +22,83 @@ class RequestError extends Error {
   }
 }
 
+interface BatchItem {
+  request: { jsonrpc: '2.0'; id: number; method: string; params: unknown }
+  resolve: (result: any) => void
+  reject: (error: Error) => void
+}
+
 class MiniRpcProvider implements AsyncSendable {
```
moodysalem, line 31: batches eth_estimateGas, eth_gas and potentially eth_blockNumber though that is usually sequential with the other two

src/hooks/useSwapCallback.ts
```diff
@@ -1,136 +1,185 @@
...
+            recipient,
+            ttl: deadline
+          })
+        )
+
+        if (trade.tradeType === TradeType.EXACT_INPUT) {
           swapMethods.push(
             Router.swapCallParameters(trade, {
-              feeOnTransfer: false,
+              feeOnTransfer: true,
```
moodysalem, line 74: this diff came out real funny

PR #1023, opened 2020-08-06
feat(lists): allow selecting and adding token lists

### LINKED ISSUE #1021: Display token warning in modal
The new token warning is very verbose (a good thing) but takes up a huge amount of space in the UI. We might want to instead display this in a modal that you click through to proceed instead?

moodysalem:
fixes #1021


callil:
@moodysalem Lists selection screen now shows list source under the name. This is the section that will need that code to strip the extraneous url stuff (see coingecko as an example).

There is also a new menu component there to manage more elements. This section has a `View list` link that currently is broken. I would like this link to either be a link to the tokenlists.org website using the "full" list url (as in not stripped of the `.link`)

The update button still appears inline when the update is needed, but I wasn't able to test this (didn't know how to trigger `pending`).

Otherwise tested on mobile and in dark mode.

moodysalem:
> @moodysalem Lists selection screen now shows list source under the name. This is the section that will need that code to strip the extraneous url stuff (see coingecko as an example).

done

> There is also a new menu component there to manage more elements. This section has a `View list` link that currently is broken. I would like this to be a link to the tokenlists.org website using the "full" list url (as in not stripped of the `.link`)

i think we should fix this in the token list site, either by redirecting to the .link url or just handling .eth urls by fetching the .link

> The update button still appears inline when the update is needed, but I wasn't able to test this (didn't know how to trigger `pending`).

tested, where should this go?
image

the way i test this is i pull down the default list, i build it, i use serve to serve it locally, and then i import that localhost list url

`serve -l 3001 build --cors`

PR #1026, opened 2020-08-07
improvement(i18n): Optimize i18n key and add some zh-CN i18n

0xashu:
- Optimize i18n key:  If there is no translation, then fall back i18n key display ❤️.
- Add some zh-CN i18n: Many Chinese users like to use Uniswap, but unfortunately there is no translation 😔.
- Test passed ✅.


moodysalem:
Is it a best practice to use the english translation as the key? Also it seems we need to figure out interpolation, e.g. for the 'Approve X SYMBOL' message so we can show the best translation

Aaaaaashu:
> Is it a best practice to use the english translation as the key?

Yes, Yes, clear and easy to maintain.
On the one hand, we no longer have to worry about i18n key naming, on the other hand, i18n is created when you coding.

> for the 'Approve X Token' message so we can show the best translation
```js
i18n Key：{t('Approve {{symbol}} Token', { symbol: "ETH"})}
en-US: Approve ETH Token // No need to write en-US i18n file
zh-CN: "Approve {{symbol}} Token": "授权 {{symbol}} 代币"
ja-JP: "Approve {{symbol}} Token": "{{symbol}}トークンを承認"
ko-KR: "Approve {{symbol}} Token": "{{symbol}} 토큰 승인"
```

src/components/CurrencyInputPanel/index.tsx
```diff
@@ -179,7 +179,7 @@ export default function CurrencyInputPanel({
                     style={{ display: 'inline' }}
                   >
                     {!hideBalance && !!currency && selectedCurrencyBalance
-                      ? 'Balance: ' + selectedCurrencyBalance?.toSignificant(6)
+                      ? `${t('Balance')}: ${selectedCurrencyBalance?.toSignificant(6)}`
```
moodysalem, line 182: can we use interpolation here

src/components/swap/TradePrice.tsx
```diff
@@ -27,8 +27,8 @@ export default function TradePrice({
 
   const show = Boolean(inputCurrency && outputCurrency)
   const label = showInverted
-    ? `${outputCurrency?.symbol} per ${inputCurrency?.symbol}`
-    : `${inputCurrency?.symbol} per ${outputCurrency?.symbol}`
+    ? `${outputCurrency?.symbol} / ${inputCurrency?.symbol}`
```
moodysalem, line 30: here as well? will the / make sense in every language?

src/pages/AddLiquidity/index.tsx
```diff
@@ -405,9 +407,11 @@ export default function AddLiquidity({
                           width={approvalB !== ApprovalState.APPROVED ? '48%' : '100%'}
                         >
                           {approvalA === ApprovalState.PENDING ? (
-                            <Dots>Approving {currencies[Field.CURRENCY_A]?.symbol}</Dots>
+                            <Dots>
+                              {t('Approving')} {currencies[Field.CURRENCY_A]?.symbol}
+                            </Dots>
                           ) : (
-                            'Approve ' + currencies[Field.CURRENCY_A]?.symbol
+                            `${t('Approve')} ${currencies[Field.CURRENCY_A]?.symbol}`
```
moodysalem, line 414: and here

src/pages/AddLiquidity/index.tsx
```diff
@@ -418,9 +422,11 @@ export default function AddLiquidity({
                           width={approvalA !== ApprovalState.APPROVED ? '48%' : '100%'}
                         >
                           {approvalB === ApprovalState.PENDING ? (
-                            <Dots>Approving {currencies[Field.CURRENCY_B]?.symbol}</Dots>
+                            <Dots>
+                              {t('Approving')} {currencies[Field.CURRENCY_B]?.symbol}
+                            </Dots>
                           ) : (
-                            'Approve ' + currencies[Field.CURRENCY_B]?.symbol
+                            `${t('Approve')} ${currencies[Field.CURRENCY_B]?.symbol}`
```
moodysalem, line 429: and here

src/pages/AddLiquidity/index.tsx
```diff
@@ -418,9 +422,11 @@ export default function AddLiquidity({
                           width={approvalA !== ApprovalState.APPROVED ? '48%' : '100%'}
                         >
                           {approvalB === ApprovalState.PENDING ? (
-                            <Dots>Approving {currencies[Field.CURRENCY_B]?.symbol}</Dots>
+                            <Dots>
+                              {t('Approving')} {currencies[Field.CURRENCY_B]?.symbol}
```
moodysalem, line 426: and here

src/pages/AddLiquidity/index.tsx
```diff
@@ -405,9 +407,11 @@ export default function AddLiquidity({
                           width={approvalB !== ApprovalState.APPROVED ? '48%' : '100%'}
```
moodysalem, line 407: this button should probably be pulled out into a component

moodysalem:
> On the one hand, we no longer have to worry about i18n key naming, on the other hand, i18n is created when you coding.

One other con is that the entire english translation is encoded in every translation file, meaning the size of these translation files is larger than it could be with arbitrary keys.

src/components/CurrencyInputPanel/index.tsx
Aaaaaashu: I think both are okay, interpolation is easier to handle spaces.

src/components/swap/TradePrice.tsx
Aaaaaashu: I think mathematics should be more general than words, how do you think about it?

src/pages/AddLiquidity/index.tsx
Aaaaaashu: This is mainly to complete yarn lint

Aaaaaashu:
> One other con is that the entire english translation is encoded in every translation file, meaning the size of these translation files is larger than it could be with arbitrary keys.

Indeed, this is not a perfect solution, but it should solve most of the translation problems.
If the problem occurs as you mentioned, we can try the abbreviation i18n Key.

src/pages/AddLiquidity/index.tsx
Aaaaaashu: I haven't read this code carefully. Is there anything I need to help optimize?

src/components/CurrencyInputPanel/index.tsx
moodysalem: sorry, i mean in the translation as discussed above, e.g.  `t('Balance: {{balance}}', {balance: selectedCurrencyBalance?.toSignificant(6)})`

Aaaaaashu: Aha, good idea

Aaaaaashu: The code has been improved, but because the clone is used as i18n-next "char to split namespace from key", so I removed ":"

https://github.com/i18next/react-i18next/issues/985

moodysalem: This seems to be another reason to use arbitrary keys instead of the english translation (which includes the colon). I would strongly prefer that. Would you be able to make that change?

Aaaaaashu: These separator options are configurable. 
I still prefer to use English directly as the key to making everything simple.

Aaaaaashu: In fact, there is no problem with using arbitrary keys. It mainly depends on core developers to maintain i18n keys so that the community can participate in translation easily. But now it is very difficult to participate in translation.

There are many Chinese Fork Uniswap projects appearing in the Chinese community because the official Uniswap i18n is difficult to translate directly.

NavidGoalpure:
Hi
I like add Persian language to uniswap-interface
i am react developer but it is my first contribute experience in a global open source project.
can someone guide me step by step on what I should do?
thanks

NavidGoalpure:
there are very hard coded English words in project.
Is it a good idea that i change them to load from the i18next Namespaces?

Aaaaaashu:
> there are very hard coded English words in project.
> Is it a good idea that i change them to load from the i18next Namespaces?

Not sure, whether @moodysalem accepts this PR.
Otherwise, now many hard code English words are difficult to translate

moodysalem:
@Aaaaaashu 
I would strongly prefer to use arbitrary short keys, instead of the english translation, as the keys in translation files.

One other problem with using english translation as keys, is if we wish to change the english text we have to change all other translation files.

I understand this makes it harder for the community to participate in translation since they need to match the key to its location in the interface, but I think we should solve this with other tooling that allows you to translate via the interface (e.g. locize)

Would you mind changing the keys to be unrelated to the english translation, and then adding the english translation to the en.json file?

Aaaaaashu:
> @Aaaaaashu
> I would strongly prefer to use arbitrary short keys, instead of the english translation, as the keys in translation files.
> 
> One other problem with using english translation as keys, is if we wish to change the english text we have to change all other translation files.
> 
> I understand this makes it harder for the community to participate in translation since they need to match the key to its location in the interface, but I think we should solve this with other tooling that allows you to translate via the interface (e.g. locize)
> 
> Would you mind changing the keys to be unrelated to the english translation, and then adding the english translation to the en.json file?

I understand what you think, I can try to use arbitrary short keys, but some texts that are too long are not good abbreviations and may need your help.

TomasFlow (approved):

TomasFlow (approved):

TomasFlow (approved):

KeKo6988:
``

KeKo6988:
Tfffd

KeKo6988:
[]() 

> ``

ChristophSiegenthaler:
@moodysalem: Is this still the prefered way forward: 
>I would strongly prefer to use arbitrary short keys, instead of the english translation, as the keys in translation files.

Thinking about updating the german i18n file soon.

ChristophSiegenthaler:
@moodysalem: i18next supports nested keys if one disables the "keySeparator" line in the i18n init. This would allow for a structured locale json, i.e. de.json: 
```
{
  "wallet": {
      "connect": "Wallet verbinden",
      "connect_long": "Mit einer Wallet verbinden",
      "no": "Keine Wallet gefunden"
  },
  "token": {
      "select": "Token auswählen"
  }
}
```
and to call it:
`{t('wallet.connect')}`
IMO this would be a very elegant and flexible solution. Is this the path I should aim for?

ChristophSiegenthaler:
https://github.com/ChristophSiegenthaler/uniswap-frontend/commit/73e1c5f000da7c0a5a9d49e6f728574cfbec9ac1

And sorry that I hijacked this PR discussion. I'll open a proper PR for the refactoring alone if this is welcome ...

justindomingue:
@ChristophSiegenthaler @0xashu We moved our localization workflow to [Crowdin](https://crowdin.com/project/uniswap-interface). It will auto-sync to this repo.

Will close this PR for now. Would appreciate your help translating the interface on Crowdin! 

Also, @0xashu you'll be happy to know we landed on English keys :)

0xashu:
@justindomingue 

Looks good, can you invite me to participate in the translation? 
I've applied to Uniswap's Crowdin translation group.

justindomingue:
Thanks @0xashu . I believe you are now approved on Crowdin. Please let me know if you have other issues.

PR #1052, opened 2020-08-18
feat(swap): setting to disable multihop routing

moodysalem:
some ux considerations:
- this persists between swaps, and may be confusing when you can't make a swap on v2 because of the setting being turned on
- a more expressive setting would be configuring the max # of hops, e.g. 2 would sometimes be better than 1 or 3
- this may be better suited for expert mode only


moodysalem:
Closing in favor of #1071

PR #1060, opened 2020-08-24
CUSTOM_BASES augments BASES_TO_CHECK_TRADES_AGAINST

brandoncurtis:
This doesn't add a token, it only adds a route hint for $BASED, [a top ten pool](https://uniswap.info/token/0x68a118ef45063051eac49c7e647ce5ace48a68a5), so that ETH → $BASED trades are correctly routed through sUSD.


moodysalem:
Only reluctant to grow this list because of issues with our routing going through illiquid pairs due to small price arbitrage opportunities without considering gas cost of additional swaps or considering the risk of revert because of routing through low liquidity pairs.

We can try to improve our routing and then add more tokens to the list.

brandoncurtis:
Here's an idea: what about, instead of using CUSTOM_BASES to filter *down* the set of possible pairs, which ultimately doesn't save very much, CUSTOM_BASES is used to *add* to the set of possible pairs for cases where unique routing possibilities exist for a particular token?

I'll push up and example.

brandoncurtis:
@moodysalem there's the concept anyway!  Liquid paths through uncommon intermediates can be added to individual tokens without adding them to the routing shortlist for all tokens.

src/hooks/Trades.ts
```diff
@@ -17,35 +18,24 @@ function useAllCommonPairs(currencyA?: Currency, currencyB?: Currency): Pair[] {
     ? [wrappedCurrency(currencyA, chainId), wrappedCurrency(currencyB, chainId)]
     : [undefined, undefined]
 
+  const customBases = CUSTOM_BASES[chainId]
+  const customBasesA: Token[] = customBases && tokenA ? customBases[tokenA.address] : []
```
NoahZinsmeister, line 22: can we do these like `CUSTOM_BASES[chainId]?.[tokenA?.address] ?? []`

src/constants/index.ts
```diff
@@ -16,6 +16,14 @@ export const USDT = new Token(ChainId.MAINNET, '0xdAC17F958D2ee523a2206206994597
 export const COMP = new Token(ChainId.MAINNET, '0xc00e94Cb662C3520282E6f5717214004A7f26888', 18, 'COMP', 'Compound')
 export const MKR = new Token(ChainId.MAINNET, '0x9f8F72aA9304c8B593d555F12eF6589cC3A579A2', 18, 'MKR', 'Maker')
 export const AMPL = new Token(ChainId.MAINNET, '0xD46bA6D942050d489DBd938a2C909A5d5039A161', 9, 'AMPL', 'Ampleforth')
+export const SUSD = new Token(
+  ChainId.MAINNET,
+  '0x57Ab1ec28D129707052df4dF418D58a2D46d5f51',
```
NoahZinsmeister, line 21: is this the correct proxy address which has uniswap liquidity?

src/constants/index.ts
```diff
@@ -16,6 +16,14 @@ export const USDT = new Token(ChainId.MAINNET, '0xdAC17F958D2ee523a2206206994597
...
+  18,
+  'sUSD',
+  'Synthetix sUSD'
+)
+export const BASED = new Token(ChainId.MAINNET, '0x68A118Ef45063051Eac49c7e647CE5Ace48a68a5', 18, '$BASED', '$BASED')
```
NoahZinsmeister, line 26: can you send a link verifying this token's identity? also, mind briefly explaining the need for the special-case routing?

src/hooks/Trades.ts
```diff
@@ -17,35 +18,24 @@ function useAllCommonPairs(currencyA?: Currency, currencyB?: Currency): Pair[] {
...
+  const customBasesB: Token[] = customBases && tokenB ? customBases[tokenB.address] : []
+  const basesWithCustom: Token[] = union(bases, customBasesA, customBasesB)
```
NoahZinsmeister, line 24: let's try to get rid of the dependency if we can. something like:

`[...bases, ...customBasesA, ...customBasesB].filter((base, i, allBases) => allBases.findIndex(allBase => allBase.equals(base)) === i)`

should work

src/constants/index.ts
brandoncurtis: that's correct, `0x57ab1ec28d129707052df4df418d58a2d46d5f51` is the proxy and the implementation target is currently `0xae38b81459d74a8c16eaa968c792207603d84480`

- https://etherscan.io/address/0x57ab1ec28d129707052df4df418d58a2d46d5f51#readContract
- https://etherscan.io/address/0xae38b81459d74a8c16eaa968c792207603d84480#readContract
- https://uniswap.info/token/0x57ab1ec28d129707052df4df418d58a2d46d5f51

src/constants/index.ts
brandoncurtis: refs:
- https://stake.based.money/
- https://etherscan.io/token/0x68a118ef45063051eac49c7e647ce5ace48a68a5
- https://uniswap.info/token/0x68a118ef45063051eac49c7e647ce5ace48a68a5
- https://uniswap.info/pair/0xaAD22f5543FCDaA694B68f94Be177B561836AE57

rationale:
- special-case routing is necessary because the project has chosen to incentivize Uniswap LPs of the sUSD:$BASED pair.
- sUSD is not on the routing shortlist (`BASES_TO_TRADE_AGAINST`) and only currently has significant liquidity with ETH and $BASED, so it does not make sense to add it to the routing shortlist for all tokens.
- the special-case routing ensures that Uniswap correctly routefinds through sUSD, allowing trades between ETH (and everything liquid with ETH) through sUSD to $BASED.

src/hooks/Trades.ts
moodysalem: if tokenA is undefined, you're trying to index into an object with string keys by the key undefined

moodysalem (changes requested):
> Here's an idea: what about, instead of using CUSTOM_BASES to filter _down_ the set of possible pairs, which ultimately doesn't save very much, CUSTOM_BASES is used to _add_ to the set of possible pairs for cases where unique routing possibilities exist for a particular token?
> 
> I'll push up and example.

Sorry, I think this is problematic, we want AMPL to *only* trade through ~USDC~DAI and ETH, as those are the only supported pairs. That was why CUSTOM_BASES was introduced.

I think the ideal behavior of CUSTOM_BASES is that if either the input or output token is present, we *only* use pairs from custom bases that intersect with both input/output

Then in the CUSTOM_BASES for sUSD, and BASED, spread (`...`) all the custom bases for mainnet + BASED/SUSD respectively

Does this make sense?

brandoncurtis:
Restricting the pairings results in significantly worse rates in several instances, for instance trading USDT to AMPL or USDC to AMPL.  A user can't be hurt by routing through an "unregistered" pool, so is the intent to punish unregistered pool LPs by depriving them of volume?  If that's the case, wouldn't that be better served with a warning/restriction on adding liquidity?

This seems like an odd choice, but I can try to make changes to preserve that behavior if that's what you want.

moodysalem:
> Restricting the pairings results in significantly worse rates in several instances, for instance trading USDT to AMPL or USDC to AMPL. A user can't be hurt by routing through an "unregistered" pool, so is the intent to punish unregistered pool LPs by depriving them of volume? If that's the case, wouldn't that be better served with a warning/restriction on adding liquidity?
> 
> This seems like an odd choice, but I can try to make changes to preserve that behavior if that's what you want.

discussed here
https://discordapp.com/channels/597638925346930701/607978109089611786/747944583496990811

brandoncurtis:
Notes to self: `CUSTOM_BASES` should be used to **restrict** what pairs a token can be traded through.  This is needed in the case of rebasing tokens like Ampleforth.

Here's why: prior to pool `sync`, which is triggered in "registered" pools by calling `sync` in the `rebase` transaction or in unregistered pools by manually calling `sync` or by the first trade or LP interaction, a pool's reported reserves are out of sync with the actual contents of the pool.  This means that the pool price is out of sync with reality, causing Uniswap to make routing decisions based on incorrect information.  Because the quoted prices are wrong, the transaction will probably fail.

I'll rewrite this PR to deal with this.

brandoncurtis:
Made some changes to reintroduce the desired filtering behavior.

React is new to me, and I'm struggling to get a unit test written to cover this.

Main blockers:
- what's the preferred way to write `jest` unit tests for a function that isn't exported? using `rewire`?
- the function I want to test, `useAllCommonPairs`, makes a hook call out to `useActiveWeb3React` to get the `chainId` and I can't figure out how to mock this up

The checks we want to do are straightforward
```
import { AMPL, DAI } from '../constants'
import { useAllCommonPairs } from './Trades'
import { CUSTOM_BASES } from '../constants'

describe('Trade', () => {

    const tokenA = DAI
    const tokenB = AMPL
    const chainId = 1
    const customBasesA = CUSTOM_BASES[chainId]?.[tokenA?.address] ?? []
    const customBasesB = CUSTOM_BASES[chainId]?.[tokenB?.address] ?? []

    it('should find all common pairs for restricted token', () => {
        const result = useAllCommonPairs(tokenA, tokenB)
        
        for (var pair of result) {
            // if customBasesA is defined, every pair with tokenA
            if (customBasesA.length && pair.involvesToken(tokenA)) {
                // must also contain an element from restrictedBasesA
                expect(customBasesA.map(base => pair.involvesToken(base)).includes(true)).toBe(true)
            }
            // if customBasesB is defined, every pair with tokenB
            if (customBasesB.length && pair.involvesToken(tokenB)) {
                // must also contain an element from restrictedBasesB
                expect(customBasesB.map(base => pair.involvesToken(base)).includes(true)).toBe(true)
            }
        }
    })
})
```

moodysalem:
> Made some changes to reintroduce the desired filtering behavior.
> 
> 
> 
> React is new to me, and I'm struggling to get a unit test written to cover this.
> 
> 
> 
> Main blockers:
> 
> - what's the preferred way to write `jest` unit tests for a function that isn't exported? using `rewire`?
> 
> - the function I want to test, `useAllCommonPairs`, makes a hook call out to `useActiveWeb3React` to get the `chainId` and I can't figure out how to mock this up

You want to extract the logic out of the hook into a function that you care about testing. This function should take any dependencies that normally come from other hooks as arguments.

src/hooks/Trades.test.ts
```diff
@@ -0,0 +1,34 @@
...
+import { generateAllRoutePairs } from './Trades'
+import { ChainId } from '@uniswap/sdk'
+
+describe('Trade', () => {
+  const chainId = ChainId.MAINNET
+  const baseTokens = BASES_TO_CHECK_TRADES_AGAINST[chainId]
+  const restrictedTokens = CUSTOM_BASES[chainId]
+
+  it('should find only restricted pairs for restricted-route tokens', () => {
+    for (const restrictedToken in restrictedTokens) {
```
moodysalem, line 11: Tests should not contain this much logic.

It is hard to understand exactly what is expected of the function by reading this test.

Better to try just a few inputs under different conditions and check their exact outputs.

eg test function with two random tokens, 1 token from common bases and 1 not, both from common bases, 1 from restricted 1 not, 2 from restricted, etc.

Each of these would be separate test cases

src/hooks/__snapshots__/Trades.test.ts.snap
```diff
@@ -0,0 +1,2000 @@
...
+      "symbol": "WETH",
+    },
+  ],
+  Array [
+    Token {
+      "address": "0xD46bA6D942050d489DBd938a2C909A5d5039A161",
+      "chainId": 1,
+      "decimals": 9,
+      "name": "Ampleforth",
+      "symbol": "AMPL",
```
moodysalem, line 107: AMPL DAI is a bug.

src/hooks/__snapshots__/Trades.test.ts.snap
```diff
@@ -0,0 +1,2000 @@
...
+      "decimals": 18,
+      "name": "Wrapped Ether",
+      "symbol": "WETH",
+    },
+    Token {
+      "address": "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
+      "chainId": 1,
+      "decimals": 6,
+      "name": "USD//C",
+      "symbol": "USDC",
```
moodysalem, line 146: WETH/USDC shows up twice.

src/hooks/__snapshots__/Trades.test.ts.snap
moodysalem: My bad, I thought it was USDC/AMPL, this is right
https://www.ampltalk.org/app/forum/technology-development-17/topic/supported-dex-pools-61/

src/hooks/Trades.ts
```diff
@@ -8,56 +8,61 @@ import { wrappedCurrency } from '../utils/wrappedCurrency'
...
 import { useActiveWeb3React } from './index'
 
+export function generateAllRoutePairs(tokenA?: Token, tokenB?: Token, chainId?: ChainId): [Token, Token][] {
+  const bases: Token[] = chainId ? BASES_TO_CHECK_TRADES_AGAINST[chainId] : []
+  const customBases = chainId !== undefined ? CUSTOM_BASES[chainId] : undefined
+  const customBasesA = customBases && tokenA ? customBases[tokenA.address] ?? [] : []
+  const customBasesB = customBases && tokenB ? customBases[tokenB.address] ?? [] : []
+
+  const allBases = [...bases, ...customBasesA, ...customBasesB].filter(
+    (base, i, allBases) => allBases.findIndex(allBases => allBases.equals(base)) === i
```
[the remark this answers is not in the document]
NoahZinsmeister: my fault on this one, i suggested the code. not clear to me that native `Set` will work in this case though, as items are objects.

moodysalem:
@brandoncurtis would love to merge this. Can you address the comments?

brandoncurtis:
I probably can't. Pretty pissed about the token. Turning this over to the Yam Finance team to close it out.

moodysalem:
We now have an additional bases feature in the routing, which would be a more suitable way to consider a specific pair in the routing for certain input/output tokens

PR #1064, opened 2020-08-26
Corrects prices displayed in PoolPriceBar

gigamesh:
The PoolPriceBar currently displays the wrong price ratio. I believe this fixes it, but someone more familiar with the code should probably test it before merging. :) 

image

### DIFF
--- a/src/state/mint/hooks.ts
+++ b/src/state/mint/hooks.ts
@@ -95,7 +95,7 @@ export function useDerivedMintInfo(
     [Field.CURRENCY_B]: independentField === Field.CURRENCY_A ? dependentAmount : independentAmount
   }
 
-  const token0Price = pair?.token0Price
+  const currentPrice = currencies[Field.CURRENCY_A]?.name === pair?.token0.name ? pair?.token0Price : pair?.token1Price
   const price = useMemo(() => {
     if (noLiquidity) {
       const { [Field.CURRENCY_A]: currencyAAmount, [Field.CURRENCY_B]: currencyBAmount } = parsedAmounts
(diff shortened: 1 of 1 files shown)


gigamesh:
The breaking test seems unrelated to my PR. Tests are all passing locally so I'm not sure what's up.

moodysalem:
Integration tests are somewhat flakey

Can you rebase on master and rerun the tests?

gigamesh:
still no luck :/

src/state/mint/hooks.ts
```diff
@@ -95,7 +95,7 @@ export function useDerivedMintInfo(
     [Field.CURRENCY_B]: independentField === Field.CURRENCY_A ? dependentAmount : independentAmount
   }
 
-  const token0Price = pair?.token0Price
+  const currentPrice = currencies[Field.CURRENCY_A]?.name === pair?.token0.name ? pair?.token0Price : pair?.token1Price
```
moodysalem, line 98: comparing names is dangerous, use `currencyEquals`

moodysalem:
you can try rebasing again

src/state/mint/hooks.ts
gigamesh: `currencyEquals` only accepts type `Currency` -- is there a utility for comparing a currency with a token from a Pair?

gigamesh: ...I figure this UI is already broken and the reliability of the comparison is pretty low-stakes, but let me know if there is a better way to do the comparison. Tests are passing :)

moodysalem: `Token` extends `Currency`, you can use `currencyEquals` here.
Comparing name is *known* to be broken as well, so I would prefer to merge a proper fix because otherwise I will just have to push another commit to fix it.

moodysalem: However you need to wrap the currency first before comparing

moodysalem:
I'll fix this on master, using `pair#priceOf`

PR #1104, opened 2020-09-14
fix: Fix the style of account transaction record list.

xiaotiandada:
Fix Style

image

Because the rounded corner of the lower left corner is 25px, and the rounded corner of the lower right and husband-level border is 20px.

In fact, you can also delete
```css
  border-bottom-left-radius: 20px;
  border-bottom-right-radius: 20px;
```


ericaricaric (approved):
lgtm

engindenizhan:
13637

PR #1121, opened 2020-09-21
Fix: Lineheight & alignment Manage Lists

nickspaargaren:
preview


nickspaargaren:
I'm not sure why we end up with a fatal

moodysalem:
now outdated

Issue #1122, opened 2020-09-21
Add information about impermantent loss risk in pool feature


3esmit:
**Is your feature request related to a problem? Please describe.**
The current information about providing liquidity can lead to misleading interpretations. 
The current message is:
```
# Liquidity provider rewards
Liquidity providers earn a 0.3% fee on all trades proportional to their share of the pool. Fees are added to the pool, accrue in real time and can be claimed by withdrawing your liquidity.
[Read more about providing liquidity](https://uniswap.org/docs/v2/core-concepts/pools/)
```
This message talks only about the potential returns, and does not mention the risk of impermantent loss.

**Describe the solution you'd like**
Change the message to explain both sides, earns fees but is at risk of impermanent losses.
Suggested message:
```
# Liquidity provider rewards
Liquidity providers earn a 0.3% fee on all trades proportional to their share of the pool. Fees are added to the pool, accrue in real time and can be claimed by withdrawing your liquidity. [Read more about providing liquidity](https://uniswap.org/docs/v2/core-concepts/pools/)
Note: Liquidity providers assume the risk of impermanent losses, pooling two tokens which have a correlated price are more risky.  [Learn more about the returns providing liquidity](https://uniswap.org/docs/v2/advanced-topics/understanding-returns)
``` 

**Describe alternatives you've considered**
Interface warns only wallets that never interacted with Uniswap right before trying to join or create a pool. 
Uniswap provides an easy guide to learn about "how to earn money as a LP" and about "reducing the risks of being a LP" and use it instead of the current docs "core-concepts", and link the core-concepts inside this guide. 

**Additional context**
The information about risk is too far from users sight, clicking the link on the current message does not explicitly mention it, instead the core-concepts docs call it "arbitrage opportunity" (which is a opportunity to others, not to the LP). 
Is easier to provider liquidity than to learn about the risks. 
There are many complaining users in Uniswap's Discord complaining about their pool shares having less value than what they initially deposited. 
By exploring the Uniswap interface, many users learned about Pooling through the interface, which only "promised winning fees", without mentioning about any risk.

PR #1135, opened 2020-09-28
Add card explaining risk of impermanent loss for LP (#1122)

jochenboesmans:
Screenshot 2020-09-28 at 13 31 28

### DIFF
--- a/src/pages/Pool/index.tsx
+++ b/src/pages/Pool/index.tsx
@@ -30,6 +30,11 @@ const VoteCard = styled(DataCard)`
   overflow: hidden;
 `
 
+const WarningCard = styled(DataCard)`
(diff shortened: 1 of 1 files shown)


callil:
@jochenboesmans I personally prefer to keep a single card in the view as demonstrated here: #1138 

I do like your copy though - it's more direct. Maybe there is a way to merge the design from @AshishGapat and the copy here?

jochenboesmans:
@callil 

The main reason for putting this in a separate card was that most of the card would be black with the additional text. Happy for the other PR to be merged instead though.

AshishGapat:
@jochenboesmans 

I like your copy better too. How do we go about this merge? I just got started with OSS so correct me if I'm wrong, but maybe you can make a pull request to my fork?

jochenboesmans:
> @jochenboesmans
> 
> I like your copy better too. How do we go about this merge? I just got started with OSS so correct me if I'm wrong, but maybe you can make a pull request to my fork?

Feel free to just add a commit to your branch, might be easier.

PR #1142, opened 2020-10-01
add right margin to mobile token list update notification lists

### LINKED ISSUE #1140: Multiple token list updates look smushed on mobile
**Bug Description**
image

**Steps to Reproduce**
1. get some dummy token list data
2. dispatch a bunch of:
```
{
          key: listUrl,
          content: {
            listUpdate: {
              listUrl,
              auto: false,
              oldList,
              newList
            }
          },
          removeAfterMs: null
        }
```

**Expected Behavior**
I expect the updates to indicate visually that there are more to scroll, but not to directly abut.

JFrankfurt:
fixes #1140

### DIFF
--- a/src/components/Popups/PopupItem.tsx
+++ b/src/components/Popups/PopupItem.tsx
@@ -30,6 +30,9 @@ export const Popup = styled.div`
 
   ${({ theme }) => theme.mediaWidth.upToSmall`
     min-width: 290px;
+    &:not(:last-of-type) {
+      margin-right: 35px;
+    }
   `}
 `
 const Fader = styled.div`


JFrankfurt:
Horizontal scroll functions already, though there's no visual affordance for it. The 35px offset is the same as the internal padding on these cards and it **does** allow enough space to see the next one on the screens I tested. (Pixel 3, Samsung something or other). Shoulda included a screenshot of the fix, but I'd already nuked my test data and was heading to bed.

JFrankfurt:
**before**
image


**after**
image

JFrankfurt:
I just tested w/ a few smaller devices and I'd actually in retrospect put this at a 20px margin instead, but it's better regardless. On the Moto G4 (not wide at all, only 360px) you almost can't see that there's a second.

JFrankfurt:
@moodysalem see #1146

PR #1144, opened 2020-10-01
Check trades against WBTC

dmihal:
WBTC is the most liquid Bitcoin-type asset on Ethereum, and has some high-liquidity token pairs (notably TBTC & DAI).

Adding this token to this array will allow better prices for trades through WBTC (for example, ETH -> WBTC -> TBTC)


moodysalem:
Reluctant to add to this list of intermediaries given this concern https://github.com/Uniswap/uniswap-interface/pull/1060#issuecomment-679306433

moodysalem:
> Reluctant to add to this list of intermediaries given this concern [#1060 (comment)](https://github.com/Uniswap/uniswap-interface/pull/1060#issuecomment-679306433)

That said, if you can finish that PR, you can add WBTC to the WETH token since WBTC/WETH is very liquid

NoahZinsmeister:
this is resolved, WBTC is searched through now

PR #1157, opened 2020-10-09
fix(ethereum-logo): centers Ethereum logo

JayWelsh:
Addresses the issue brought up on https://www.reddit.com/r/ethereum/comments/j7upcm/i_know_to_a_lot_of_you_it_might_seem_picky_but_as/

Previous Version:

ethereum-logo-off-center

Updated Version:

ethereum-logo


moodysalem:
Hmmm which one is official though? Symmetric != better

https://www.adweek.com/creativity/how-the-imperfections-in-googles-logo-are-what-make-it-perfect/

JayWelsh:
@moodysalem I think that it was obviously a mistake in the original image design, this is not a case of visual centering or a justified example, there seems to be pretty solid consensus in the Reddit thread that it isn't "for the best". I'll leave this PR open but I am not really prepared to put much effort into trying to make a case for why this should be merged, since it seems fairly straightforward. 🌈

PR #1164, opened 2020-10-12
fix(governance): modal bugs, redable styles, show common names for contracts

ianlapham:


src/theme/index.tsx
```diff
@@ -192,6 +192,10 @@ body {
   padding: 0;
 }
 
+ a {
+   color: #2172E5;
```
moodysalem, line 196: should we use a theme variable?

moodysalem (approved):

src/theme/index.tsx
ianlapham: Oh true

PR #1167, opened 2020-10-14
add conditional tokens list

auryn-macmillan:
Gnosis' conditional tokens are ERC1155 compatible. We have an [ERC20 wrapper](https://github.com/gnosis/1155-to-20) for the tokens so they can be traded one AMMs that only support ERC20. One feature of the list is that it gives each of the tokens identical names and symbols. So Gnosis would like to provide a curated list of these wrapped tokens.


TomasFlow (approved):

TomasFlow (approved):

TomasFlow (approved):

moodysalem:
The only tokens on this list are for an expired market, consider adding the list to tokenlists-org repo

PR #1173, opened 2020-10-19
improve displayed proposal end time accuracy

shanefontaine:
Proposals currently last 40,320 blocks, which equates to exactly 7 days at 15s block times.

However, the [Muir Glacier fork](https://blog.ethereum.org/2019/12/23/ethereum-muir-glacier-upgrade-announcement/) in early 2020 decreased the average block time to [somewhere between 13-13.5s](https://etherscan.io/chart/blocktime). This equates to proposals lasting approximately 6.5 days or less. This may confuse people who rely on the UI to wait until the last minute to vote on a proposal.

This PR adds a more accurate (~10%) time when displaying the approximate end of the proposal. It leaves room for variance in block times by assuming 14s block times instead of the more accurate 13-13.5s so that the proposal display time has never passed before a proposal is actually over.


NoahZinsmeister (approved):

NoahZinsmeister:
thanks @shanefontaine !

PR #1174, opened 2020-10-20
fix(token lists): stop showing notifications for lists that are not selected

moodysalem:
fixes https://github.com/Uniswap/uniswap-interface/issues/1161


ianlapham (approved):
LGTM

Issue #1184, opened 2020-10-27
Absolute imports


anxolin:
**Is your feature request related to a problem? Please describe.**
The app doesn't support absolute imports. Within a componentents directories can be nice, but for importing components or utils, you need to include the `../../` in the path.

**Describe the solution you'd like**
Add `baseUrl` in the `compilerOptions` in `tsconfig.js`

```js
{
  "compilerOptions": {
    ...,
    "baseUrl": "src"
  },
...
}
```

More info in https://create-react-app.dev/docs/importing-a-component

I've made this PR #1185 with the suggestion

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

anxolin:
@moodysalem when do you think my PRs can be reviewed?

moodysalem:
I'm not opposed but also don't necessarily see the benefit if you are using automatic imports in your IDE. And it creates 2 different ways of importing which means we should be consistent.

anxolin:
Thanks for the reply. I think it makes refactors easier. Also, easier to copy paste some imports between files. 
Looks simpler too, so easier to read.

I think within a component, if it has its own directory, is good to use relative to import the subcomponent parts. 

But to import stuff the `constants`, the `theme`,  or some components, it makes it a bit inconvenient to import using `../.../` or `../.../..` depending where you are.

moodysalem:
QQ, what happens if there is a conflict between an absolute import and a node_module import?

anxolin:
I think `src` will take precedence on a conflict. 

Normally there won't be conflicts, since imports will be like `hook/aaaa`, or `data/yyyy`

If you want to do a quick test, you can create the file `src/lodash.ts` with this content:

```ts
export interface Cancelable {
  cancel(): void
  flush(): void
}
```

Then, you can see how src code gets picked first:
```js
import { Cancelable } from 'lodash'
```

PR #1186, opened 2020-10-27
Rewrite paths as absolute

anxolin:
This PR is related to #1184 Issue.

## Context
This PR simplifies the imports by using absolute paths. It's the seconds part of this PR ( #1185) where it was added the support for absolute paths.

For context https://create-react-app.dev/docs/importing-a-component

## Some notes
* Although this PR is big, it **ONLY** modifies the routes that refer to a directory hanging directly from `src`. It doesn't do any other change in the code base.
* It was modified using a REGEX, following the pattern `import '(..\/)+(components|assets|<all other dirs>)\/` 
* These are the dirs used in the REGEX:
image
* There's one exception. The constants cannot be resolved using `import { ... } from 'constants'`, so they were changed into `import { ... } from 'constants/index'`

## Not included in this PR
Although this REGEX manages to change a big part of the imports, I decided against "going manual" and modify some relative imports like: 

```ts
import CurrencySearchModal from '../SearchModal/CurrencySearchModal'
```

This is more error prone, so they can be done in a separated PR, getting the ones changed in this PR out of the way.


TomasFlow (approved):

anxolin:
Any chance this can be also merged @moodysalem?

moodysalem:
> Any chance this can be also merged @moodysalem?

Yup sorry will take a look. It's just a lot of work to manually review so it would be easier if I just copied your approach and applied it to master

anxolin:
I understand. Whatever is easier for u, yes there's a lot of changes and not fun to review, but I think the way Github highlight the change is not that bad. You'll just see a lot of `../..` disappearing and the project working as before 😃

anxolin:
Hei @moodysalem did you have a chance to review the changes. I saw there was conflicts with master, and I solved them, but it would be nice if we review it sooner, so it doesn't diverge from master (there's a lot of small changes, so it will keep happening)

moodysalem:
I hear you but I'm worried about merging this and missing something that breaks the interface or worse and don't think it's worth the due diligence to review since it has no impact on the end user xp and minimal impact on the developer xp. If you want to get this changed across the codebase, I would be more comfortable with a PR adding a bash script that rewrites all the paths, which I can more easily review and then apply to master

anxolin:
@moodysalem , yes I totally get it. I provided some instructions here, I hope with this makes things easier https://github.com/Uniswap/uniswap-interface/issues/1207

PR #1195, opened 2020-11-03
Fix(pool page): show deposited liquidity on pool page

ianlapham:
- detect if user has any liquidity deposited into a rewards pool 
- show this along with regular liquidity so users dont get confused about what they own


src/components/PositionCard/index.tsx
```diff
@@ -157,17 +158,20 @@ export function MinimalPositionCard({ pair, showUnwrapped = false, border }: Pos
...
   const currency1 = unwrappedToken(pair.token1)
 
   const [showMore, setShowMore] = useState(false)
 
-  const userPoolBalance = useTokenBalance(account ?? undefined, pair.liquidityToken)
+  const userDefaultPoolBalance = useTokenBalance(account ?? undefined, pair.liquidityToken)
   const totalPoolTokens = useTotalSupply(pair.liquidityToken)
 
+  // if staked balance balance provided, use instead of fetched balance from pool
+  const userPoolBalance = stakedBalance ?? userDefaultPoolBalance
```
moodysalem, line 173: shouldn't it be the combination?

src/pages/Pool/index.tsx
```diff
@@ -180,10 +186,19 @@ export default function Pool() {
...
                 </ButtonSecondary>
-
                 {allV2PairsWithLiquidity.map(v2Pair => (
                   <FullPositionCard key={v2Pair.liquidityToken.address} pair={v2Pair} />
                 ))}
+                {stakingPairs.map(
+                  (stakingPair, i) =>
+                    stakingPair[1] && ( // skip pairs that arent loaded
+                      <FullPositionCard
+                        key={stakingPair[1].liquidityToken.address}
```
moodysalem, line 196: conflicting key

src/components/PositionCard/index.tsx
ianlapham: Guess its a UX question - i could see the case for showing the total, and then giving 3 buttons (add, remove, manage staked)

ianlapham: and then a line for the staked amount?

ianlapham: Updated to a combination, with an extra button and row of data if theres a staked amount

PR #1205, opened 2020-11-10
fix: token warning modal stealing focus across frame boundaries

moontools-hyperion:
The token warning modal in the Uniswap interface, when embedded in an iframe on a parent page, steals focus from the parent window. This prevents users from interacting with input elements in the parent window.

Simple example of issue:
https://jsfiddle.net/26cms93e/

Steps to reproduce:
1. Go to https://app.moontools.io/pairs/0x02f14c27037bd30f18a6578590fd40fafd3376ff
2. Click on the "Swap" tab (contains an iframe pointing to Uniswap interface) and wait for token warning modal to load
3. Notice that you cannot interact with the "Filter by address" input field above the transaction log. If you close the token warning modal, the input becomes focusable again.

This issue is also present on other DEX data explorer sites. This PR fixes the issue described above.

### DIFF
--- a/src/components/Modal/index.tsx
+++ b/src/components/Modal/index.tsx
@@ -115,7 +115,13 @@ export default function Modal({
       {fadeTransition.map(
         ({ item, key, props }) =>
           item && (
-            <StyledDialogOverlay key={key} style={props} onDismiss={onDismiss} initialFocusRef={initialFocusRef}>
+            <StyledDialogOverlay
+              key={key}
+              style={props}
+              onDismiss={onDismiss}
+              initialFocusRef={initialFocusRef}
+              dangerouslyBypassFocusLock
+            >
               <StyledDialogContent
                 {...(isMobile
                   ? {
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "devDependencies": {
     "@ethersproject/experimental": "^5.0.1",
     "@popperjs/core": "^2.4.4",
-    "@reach/dialog": "^0.10.3",
+    "@reach/dialog": "^0.11.2",
     "@reach/portal": "^0.10.3",
     "@reduxjs/toolkit": "^1.3.5",
     "@types/jest": "^25.2.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2048,19 +2048,27 @@
     penpal "3.0.7"
     pocket-js-core "0.0.3"
 
-"@reach/dialog@^0.10.3":
-  version "0.10.5"
-  resolved "https://registry.yarnpkg.com/@reach/dialog/-/dialog-0.10.5.tgz#70bd832616fc2b0f31ed65aff81ae9e218c4dbc4"
-  integrity sha512-j2+VaHCen/M5zh+q6L/xMXETMeD5yVywdhjjs0kn2uQ3AolYSj32P8Go1xSpxQNAjS1/zChd02Y/4g02eL3afg==
+"@reach/dialog@^0.11.2":
+  version "0.11.2"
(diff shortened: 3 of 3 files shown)


moodysalem:
This is very interesting--is it better fixed in chrome/safari though? Should a child iframe be able to steal focus or blur inputs on the parent page? It seems like a good thing you can't tab outside of the modal while on the Uniswap interface.

src/components/Modal/index.tsx
```diff
@@ -115,7 +115,13 @@ export default function Modal({
       {fadeTransition.map(
         ({ item, key, props }) =>
           item && (
-            <StyledDialogOverlay key={key} style={props} onDismiss={onDismiss} initialFocusRef={initialFocusRef}>
+            <StyledDialogOverlay
+              key={key}
+              style={props}
+              onDismiss={onDismiss}
+              initialFocusRef={initialFocusRef}
+              dangerouslyBypassFocusLock
```
MicahZoltu, line 123: The docs on this explicitly say that you should have some replacement focus lock if you enable this, but I don't see any replacement in this PR.

MicahZoltu:
https://github.com/w3c/webappsec-permissions-policy/issues/273

MicahZoltu:
IIUC, the browser solution is to disallow the iframe from being able to programmatically set focus at all.  Essentially this means that Uniswap should not depend on its ability to control focus for disabling page content.  Instead it should disable fields in the app and/or use an overlay.

moontools-hyperion:
> Should a child iframe be able to steal focus or blur inputs on the parent page? 

IMO it should not, but evidently it is possible in some browsers. Afaik there isn't a way to prevent this.

Probably need to do what @MicahZoltu suggested i.e. disabling all inputs in uniswap-interface when the modal is open. Is this approach okay? @moodysalem

moodysalem:
> IIUC, the browser solution is to disallow the iframe from being able to programmatically set focus at all.  Essentially this means that Uniswap should not depend on its ability to control focus for disabling page content.  Instead it should disable fields in the app and/or use an overlay.

Wdym by an overlay? I do not like the idea of disabling all other inputs on the page when a modal is open. It will either be mutating the dom of many components outside of react and is likely to break, or require changing tons of components

luanluvthu (approved):
Ok

MicahZoltu:
> I do not like the idea of disabling all other inputs on the page when a modal is open.

I'm not a huge fan either, but I don't think there is another option if you want the iframe widget to work correctly and not inappropriately interfere with the hosting page.  If you do nothing, people using the iframe have two choices:
1. Their page breaks when Uniswap has a modal open.
2. Uniswap modals don't work correctly within the iframe.

> It will either be mutating the dom of many components outside of react and is likely to break, or require changing tons of components

I agree that it increases complexity quite a bit.  I would not be worried about DOM updates since most (all) of the changes won't trigger a re-layout, just some non resizing style changes.  Component updates aren't that expensive (on this order of magnitude) as long as they don't trigger a relayout.  I suspect you can mitigate the complexity quite a bit with a good framework (e.g., React).

> I am pretty sure the browser should fix it

From what I have seen, no browser considers this a bug.  I can appreciate one disagreeing on this, but I think Uniswap should work in the real world, not in a hypothetical future that doesn't align with current trajectory for reality.

moodysalem:
We need to set crossFrame to false on the underlying focus lock that reach dialog is rendering
https://github.com/theKashey/react-focus-lock/issues/104#issue-601676942

moodysalem:
@moontools-hyperion are you able to add a replacement focus lock in our Modal component that has crossFrame set to false?

sharife3:
Any updates on this ?

moodysalem:
> Any updates on this ?

The better fix is described in the above comment (disable only cross frame focus lock) and we would accept any pull request that implements it

moontools-hyperion:
Hi @moodysalem, I've created a PR for this at https://github.com/Uniswap/uniswap-interface/pull/1326

PR #1211, opened 2020-11-15
fix: bug causing wbtc pooled amount to be off by a factor of 10

ncitron:
In the pool tab, when viewing your pooled wBTC balance, the number is off by a factor of 10. This is due to the decimal being set incorrectly in `src/constants/index.ts`.


moodysalem:
confirmed decimals here https://etherscan.io/token/0x2260fac5e5542a773aa44fbcfedf7c193bc2c599

PR #1232, opened 2020-12-08
Add uEth as routing pair. Add tokenlist.ulock.eth to tokenlists.

udev-ulock:
uETH allows safely creating tokens with high liquidity on Uniswap for no cost. All circulating uETH is backed 1:1 by ETH, but uETH permanently locked in Uniswap is not redeemable. This allows the free creation of arbitrary amounts of liquidity without any protocol upgrades to Uniswap. More information is available at [ulock.eth](https://ulock.eth).

This PR adds uETH as a routing pair for the Uniswap app, allowing new users to trade uETH/xyz pairs without using the uEth wrapping contract.

Without uETH as a routing pair, malicious liquidity providers can create low liquidity WETH/xyz pairs then withdraw the liquidity after the WETH/xyz price exceeds the uETH/xyz price.


moodysalem:
The mechanism described is solved by v3 (just add liquidity in a price range above/below the current price), and doesn't look like the project is still active

PR #1237, opened 2020-12-17
Improvement(lists): Switch to multiple active lists

ianlapham:
- update state to contain multiple active lists, search across all of these
- UI updates for list management
- allow toggle for single hop trades only 
- use single hop trades if amounts vary by less than 0.5% of input/output amount


src/components/PositionCard/index.tsx
```diff
@@ -202,7 +202,8 @@ export default function FullPositionCard({ pair, border, stakedBalance }: Positi
             <Text fontWeight={500} fontSize={20}>
               {!currency0 || !currency1 ? <Dots>Loading</Dots> : `${currency0.symbol}/${currency1.symbol}`}
             </Text>
-            {!!stakedBalance && (
+            {/* Hide this UI until a new staking program is enabled */}
+            {/* {!!stakedBalance && (
```
moodysalem, line 206: just delete

PR #1239, opened 2020-12-18
feat: allow overriding proposals in the UI

gakonst:
As title. Also includes the Uniswap grants proposal, since its description was fudged. https://gov.uniswap.org/t/rfc-uniswap-grants-program-v0-1/9081.

Overriding a proposal is as simple as adding it in the `constants/proposals` directory and exporting it via the PRELOADED_PROPOSALS map with its proposalId.


crazyrabbitLTC:
Dennison from [www.withTally.com](Tally) here, we saw that last night as well on our own interface as it blew up our formatting too. :-) 

Any thoughts about creating a standard template for Proposal submission to help community members when crafting proposals?

PR #1248, opened 2020-12-24
Add wBTC to default pairs. Changed branch name so checks maybe won't fail.

sksksksksksksk:
This is the same as [1241](https://github.com/Uniswap/uniswap-interface/pull/1241) with a different branch name and one more addition, so wBTC shows up in the recommended pools to add liquidity as well.

This one might actually merge...


moodysalem:
Don't worry about linter plugin failures, we will just override it for forks and let it fix the code style on master

PR #1254, opened 2020-12-29
improvement(routing): Add toggle for single hops only, use single hop trade if below difference threshold

ianlapham:


src/hooks/Trades.ts
```diff
@@ -78,16 +78,36 @@ function useAllCommonPairs(currencyA?: Currency, currencyB?: Currency): Pair[] {
   )
 }
 
+const OUTPUT_PERCENT_DIFFERENCE_MAX_BIPS = 50 // 0.5% difference at most in output amount
```
moodysalem, line 81: no longer just output

src/hooks/Trades.ts
```diff
@@ -101,10 +121,23 @@ export function useTradeExactOut(currencyIn?: Currency, currencyAmountOut?: Curr
...
         null
-      )
+
+      const tradeFixed =
+        Trade.bestTradeExactOut(allowedPairs, currencyIn, currencyAmountOut, { maxHops: 1, maxNumResults: 1 })[0] ??
+        null
+
+      // if input amount difference is within threshold, return single hop trade instead - only check if valid trades
+      if (tradeFixed && trade) {
+        const percent = new Percent(trade.inputAmount.raw, tradeFixed.inputAmount.raw)
```
moodysalem, line 134: this is really hard to wrap my head around, is it correct?

say the 1 hop trade is 100 input, multihop is 99 input
if (99 / 100) > 99.5% return 1 hop
shouldn't we return multihop since it's different by 1%?

would be good to extract this code to a method and add a unit test

ianlapham: I just changed this logic a bunch to make it more clear, now comparing the difference to a percent max, or just returning single if the input/output amount is favorable (based on exactIn or exactOut)

ianlapham: gunna move to a function and test

PR #1278, opened 2021-01-22
Fix typo in Import Token UI

gin:
Typo in Import Token UI
Typo in Import Token UI


callil (approved):

ianlapham:
thanks for the contribution - just added this to #1301

PR #1310, opened 2021-02-09
show hidden search results by default

ianlapham:
- replace button with list separator and always show inactive results


src/components/Card/index.tsx
```diff
@@ -18,6 +18,10 @@ export const LightCard = styled(Card)`
   background-color: ${({ theme }) => theme.bg1};
 `
 
+export const LighgGreyCard = styled(Card)`
```
JFrankfurt, line 21: `LighgGreyCard` spelling?

src/components/SearchModal/filtering.ts
```diff
@@ -42,3 +43,39 @@ export function filterTokens(tokens: Token[], search: string): Token[] {
...
+          token.symbol?.toLowerCase() !== symbolMatch[0]
+      ),
+
+      // rest that dont match upove
+      ...tokens.filter(
+        token =>
+          !token.symbol?.toLowerCase().startsWith(searchQuery.toLowerCase().trim()) &&
+          token.symbol?.toLowerCase() !== symbolMatch[0]
+      )
+    ]
```
JFrankfurt, line 79: minor nit - perf could be improved here by calling filter on the subset of tokens not chosen by prior calls, instead of re-filtering the whole list, and by only doing the `toLowerCase` and `trim` on first iteration (and once at the start for the `searchQuery`.

JFrankfurt:
Should debounce the text input; it's a little laggy right now because it's running the filters on every keystroke.

src/components/Card/index.tsx
ianlapham: Yike yep

src/components/SearchModal/filtering.ts
ianlapham: Will take a look here - the filtering is kinda a mess, theres 3 levels to it right now

ianlapham:
> Should debounce the text input; it's a little laggy right now because it's running the filters on every keystroke.

@JFrankfurt  Added a 150ms, think this feels alright?

JFrankfurt:
I've never really thought about what the *right* input debounce is. Interesting question. I just did some googling and saw someone calculate 167ms is the average time between keystrokes, so maybe we do a little more? Like 200ms?

JFrankfurt (approved):

ianlapham:
> I've never really thought about what the _right_ input debounce is. Interesting question. I just did some googling and saw someone calculate 167ms is the average time between keystrokes, so maybe we do a little more? Like 200ms?

Yeah I've seen 200 a lot. Ill update to this then merge. @JFrankfurt

Issue #1316, opened 2021-02-10
Use service workers so the site is still available when the host goes down


moodysalem:
**Is your feature request related to a problem? Please describe.**
When the host goes down, sometimes Uniswap interface becomes inaccessible.

**Describe the solution you'd like**
Use service workers to cache the site and prefer the cached version, notifying the user when updates are available

**Describe alternatives you've considered**
Switching to a more centralized hosting provider (i.e. non-ipfs gateway)

PR #1319, opened 2021-02-15
feature(service worker): add offline support

JFrankfurt:
addresses #1316

I've added pretty much the default `workbox`-enabled `create-react-app` service worker. 

Apparently there's some weirdness with the build process, so the standard implementation only runs in production. I've added `serve` https://www.npmjs.com/package/serve as a dep here and `start:service-worker: "yarn build && yarn serve -s build"` to `scripts` for dev work on service-worker things.

You can now disable network connectivity in dev tools and still expect a UI to show up when you refresh the page. 😄 Ideally this will keep the UI functional for people next time the cloudflare ipfs/cdn setup has issues.


public/manifest.json
```diff
@@ -1,6 +1,7 @@
 {
-  "short_name": "Uniswap",
-  "name": "Uniswap",
+  "background_color": "#fff",
+  "display": "standalone",
+  "homepage_url": "https://app.uniswap.org",
```
moodysalem, line 4: hmm can you remind me what this effects?

because sometimes the host isn't app.uniswap.org, e.g.
https://app-uniswap-org.ipns.dweb.link

moodysalem:
One interesting case to test is when the cloudflare host goes down, it doesn't always *completely* go down, sometimes it serves an html or text/plain response with some error message in place of a HTML/JS/CSS/JSON file. Does the workbox code handle that?

public/manifest.json
JFrankfurt: https://developer.chrome.com/docs/extensions/mv2/manifest/homepage_url/

It looks like it's really only useful for installable browser extensions. 🤷

moodysalem: Maybe just drop it, no need for the extra reference to a particular URL

JFrankfurt:
@moodysalem this will cache the app when it first loads, then load the cached version first in all cases (whether cloudflare sends one or not). It should only replace the cache when the site loads successfully and a new service worker can be activated. [more info](https://developers.google.com/web/tools/workbox/modules/workbox-precaching#how_workbox-precaching_works)

Since the browser should render this before dns resolution, my guess is existing users who visit the site won't notice anything's wrong.

moodysalem:
OK, down to try it and see what happens, after resolving all the linting errors ofc

JFrankfurt:
TBH I'm a little blown away by how well this works relative to how easy it was. It's a perfect fit for us.

PR #1321, opened 2021-02-18
undoes service worker reversion

JFrankfurt:


moodysalem:
I believe the function you need to call is `unregister`

moodysalem (approved):

PR #1336, opened 2021-03-17
fix(merkle drop): make claim a post request

JFrankfurt:
fixes https://github.com/Uniswap/uniswap-interface/issues/1334

### DIFF
--- a/src/state/claim/hooks.ts
+++ b/src/state/claim/hooks.ts
@@ -29,18 +29,16 @@ function fetchClaim(account: string, chainId: ChainId): Promise<UserClaimData | 
 
   return (CLAIM_PROMISES[key] =
     CLAIM_PROMISES[key] ??
-    fetch(`https://gentle-frost-9e74.uniswap.workers.dev/${chainId}/${formatted}`)
-      .then(res => {
-        if (res.status === 200) {
-          return res.json()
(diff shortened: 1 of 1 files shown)


src/state/claim/hooks.ts
```diff
@@ -29,18 +29,16 @@ function fetchClaim(account: string, chainId: ChainId): Promise<UserClaimData |
...
-      }))
+    fetch('https://merkle-drop-1.uniswap.workers.dev/', {
+      body: JSON.stringify({ chainId, address: formatted }),
+      headers: {
+        'Content-Type': 'application/json',
+        'Referrer-Policy': 'no-referrer'
+      },
+      method: 'POST'
+    })
+      .then(({ ok, json }) => (ok ? json() : console.log(`No claim for account ${formatted} on chain ID ${chainId}`)))
```
moodysalem, line 40: prefer just console.debug, we don't want people to be confused by this

moodysalem (approved):

src/state/claim/hooks.ts
JFrankfurt: ah, saw this and assumed the more common one would be better, but I guess it's maybe only a TS alias? 
image

PR #1338, opened 2021-03-18
stop destructuring merkle drop response object

JFrankfurt:


sham789:
What's the point of getting rid of destructuring, may I ask?

JFrankfurt:
https://stackoverflow.com/a/54817427

I added it as a simpler way to get `res.ok` and `res.json`, but `Response` objects from `fetch` don't like destructured assignment.

PR #1347, opened 2021-03-29
Refactor: default token return from URL parameter

rhacker:


src/state/swap/hooks.ts
```diff
@@ -217,14 +217,15 @@ export function useDerivedSwapInfo(): {
...
   if (typeof urlParam === 'string') {
     const valid = isAddress(urlParam)
     if (valid) return valid
-    if (urlParam.toUpperCase() === 'ETH') return 'ETH'
-    if (valid === false) return 'ETH'
+    if (urlParam.toUpperCase() === DEFAULT_TOKEN) return DEFAULT_TOKEN
+    if (valid === false) return DEFAULT_TOKEN
   }
-  return 'ETH' ?? ''
+  return DEFAULT_TOKEN ?? ''
```
moodysalem, line 228: this line is clearly some kind of typo

PR #1356, opened 2021-04-04
fix(fei): route fei through tribe only

moodysalem:
the problem with this pr is if the user selects fei/eth it can't find any route

the best solution is probably to customize the fei/eth pair we use for routing to actually return correct quotes

### DIFF
--- a/src/constants/index.ts
+++ b/src/constants/index.ts
@@ -20,6 +20,10 @@ export const USDT = new Token(ChainId.MAINNET, '0xdAC17F958D2ee523a2206206994597
 export const COMP = new Token(ChainId.MAINNET, '0xc00e94Cb662C3520282E6f5717214004A7f26888', 18, 'COMP', 'Compound')
 export const MKR = new Token(ChainId.MAINNET, '0x9f8F72aA9304c8B593d555F12eF6589cC3A579A2', 18, 'MKR', 'Maker')
 export const AMPL = new Token(ChainId.MAINNET, '0xD46bA6D942050d489DBd938a2C909A5d5039A161', 9, 'AMPL', 'Ampleforth')
+
+export const FEI = new Token(ChainId.MAINNET, '0x956F47F50A910163D8BF957Cf5846D573E7f87CA', 18, 'FEI', 'Fei USD')
+export const TRIBE = new Token(ChainId.MAINNET, '0xc7283b66Eb1EB5FB86327f08e1B5816b0720212B', 18, 'TRIBE', 'Tribe')
+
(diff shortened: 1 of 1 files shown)


sebastiandelgado:
@moodysalem thank you for putting this together so quickly. The token addresses look all good on this PR.

If the net effect of this PR is equivalent to setting "Disable Multihops" to "on" on the uniswap exchange client when trading FEI<>TRIBE, then this is exactly what we are looking for. We need TRIBE/FEI to be traded directly instead of through ETH. 

The reason routing anything through the FEI/ETH will not work on the uniswap exchange client while FEI is under peg is that the burn amount is not taken into account by the client. Because of this, the more permanent solution would be to never trade anything through the FEI-ETH pair on uniswaps' interface. That seems a bit heavy-handed for now though. This PR LGTM. Thanks again.

moodysalem:
> Because of this, the more permanent solution would be to never trade anything through the FEI-ETH pair on uniswaps' interface. That seems a bit heavy-handed for now though. This PR LGTM. Thanks again.

If that is the desired behavior, this is _more_ than setting disable multihops to on. It removes the ability to trade ETH/FEI in either direction from the interface. This was added for AMPL for which trades through any pair other than DAI/ETH would fail.

sebastiandelgado:
> If that is the desired behavior, this is more than setting disable multihops to on. It removes the ability to trade ETH/FEI in either direction from the interface. This was added for AMPL for which trades through any pair other than DAI/ETH would fail.

Ok, I wasn't super clear, so let me rephrase this a bit. 

For now, the desired solution is to route users trying to trade FEI-TRIBE through the FEI-TRIBE pool and not have them go through the FEI-ETH pool, which will cause the tx to fail if FEI is under its peg.

Thanks again and let me know if I can clarify anything else for you!

moodysalem:
> > If that is the desired behavior, this is more than setting disable multihops to on. It removes the ability to trade ETH/FEI in either direction from the interface. This was added for AMPL for which trades through any pair other than DAI/ETH would fail.
> 
> 
> 
> Ok, I wasn't super clear, so let me rephrase this a bit. 
> 
> 
> 
> For now, the desired solution is to route users trying to trade FEI-TRIBE through the FEI-TRIBE pool and not have them go through the FEI-ETH pool, which will cause the tx to fail if FEI is under its peg.
> 
> 
> 
> Thanks again and let me know if I can clarify anything else for you!
> 
> 
> 
> 

Yes, this PR does that.

However, if the user wants to sell FEI for ETH or buy FEI with ETH, the interface with this PR will not try the FEI/ETH pair nor the FEI/TRIBE pair, and it will not find a route.

If that's the desired behavior, we could merge this. More likely the user wants to go FEI->TRIBE->ETH to sell FEI, which requires a bigger code change. Our routing in the interface is lacking, which is why we added the disable multihop toggle.

PR #1362, opened 2021-04-12
Use BigNumber for userDeadline

jackaldridge:
This PR switches `userDeadline` to `BigNumber` representation to fix Uniswap/uniswap-interface#1313.

Users can now enter an arbitrarily high number for transaction deadline without the interface crashing. All other UI behaviors are unaltered.


moodysalem:
I don't think we should store complex objects in the redux store, because it is serialized to localStorage and won't be a BigNumber when it's deserialized from localStorage

jackaldridge:
Yeah that makes sense. I’ll make those changes tomorrow.

moodysalem:
sorry, i fixed this without changing the store here https://github.com/Uniswap/uniswap-interface/commit/8bea95fab2ce056ab9dd5a00601ae018b38ae1f0

PR #1363, opened 2021-04-13
add additional bases

NoahZinsmeister:


MicahZoltu:
Code seems fine, but philosophically I think this is a bad change.  UNI is *not* a reasonable denomination token for trading and adding this is just going to further bifurcate liquidity across Uniswap.  I have always been against routing through anything other than ETH and while I *understood* (but disagreed with) the arguments around stable coin bases I mostly ignored them.  WBTC was a bit silly because it didn't have the support of the stablecoin argument so it was just "how can we best encourage liquidity splitting".  This latest change very much seems like "how can we short-term pump our bags" instead of focusing on "how can we make Uniswap a better product overall" (which would mean encouraging liquidity consolidation rather than liquidity splitting).

NoahZinsmeister:
@MicahZoltu this is just a test PR for a particular class of assets which might need to be able to search through UNI pairs, it will not be merged! appreciate the feedback though :)

MicahZoltu:
> it will not be merged

😒

NoahZinsmeister:
@MicahZoltu it doesn't do what it did when you looked at it. we simply adding additional base assets for a select few input/output assets, to improve routing

MicahZoltu:
Maybe I am misunderstand something but the updates you made to the PR made the situation *much worse*.  Not only did you add UNI to the set of routable base pairs (the thing I originally argued against), but you also added 6 additional tokens to the set of routable base pairs.

The result is that liquidity will start to get split *even more* than if you only added UNI as a base pair.

NoahZinsmeister:
you've misunderstood, the tokens in the values of `ADDITIONAL_BASES` are treated as bases only for trades involving the tokens in the keys. the behavior for trades not involving any of those tokens is unchanged

PR #1375, opened 2021-04-26
feat(routing): support mirror protocol routing as additional bases

Jorropo:
This allows to trade mirror's mAssets correctly.
Thoses assets are indexed on the `UST` stablecoin and thus require a custom routing.

Related token list issue (technically not needed but would be good) : https://github.com/Uniswap/tokenlists-org/issues/722

Why not `UST` or `MIR` to global routing ? I don't think it's currently usefull, `UST` and `MIR` are kinda like a trade pocket, it would be pretty rare for trades to go through there.
The only probable routes (excluding mAssets) would be `x <-> USDT <-> UST <-> ETH` or `USDT <-> UST <-> ETH <-> x` or `USDT <-> UST <-> MIR <-> ETH` which is clearly not enough, plus on all of thoses UST is competing with USDT which have a way bigger ETH / UDST pool.
Also even tho `UST` have lots of liquidity, most of it is in mirror's dead ends `MIR` and mAssets, `USDT / UST` is only about 20m$ big, which make it way less interresting as stable for new pairs (`USDT`, `USDC`, `DAI`, `FEI` (when it will work :D ), ... would be way more interresting).
That is why I think as addtional bases is the better place currently (again this could change if more non mirror token have a pool with `UST` which would make `UST` a stronger generic stable coin, and or, if `USDT / UST` would be bigger).


Jorropo:
@moodysalem @haydenadams this has been rebased for V3.
Any status about a merge ?

PR #1379, opened 2021-04-30
Update polling block number breathing behavior

AmagiDDmxh:
I see the hovering behavior wasn't look like the codes trying to do, so I made some styling updates.

The screent shot below shows what it looks like

breath light

Polling Breath Dot

### DIFF
--- a/src/components/Header/Polling.tsx
+++ b/src/components/Header/Polling.tsx
@@ -9,19 +9,22 @@ import { useActiveWeb3React } from '../../hooks'
 const StyledPolling = styled.div`
   position: fixed;
   display: flex;
+  align-items: center;
   right: 0;
   bottom: 0;
   padding: 1rem;
-  color: white;
-  transition: opacity 0.25s ease;
   color: ${({ theme }) => theme.green1};
-  :hover {
-    opacity: 1;
-  }
 
   ${({ theme }) => theme.mediaWidth.upToMedium`
     display: none;
   `}
+`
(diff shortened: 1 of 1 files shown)


src/components/Header/Polling.tsx
```diff
@@ -9,27 +9,29 @@ import { useActiveWeb3React } from '../../hooks'
...
 
   ${({ theme }) => theme.mediaWidth.upToMedium`
     display: none;
   `}
 `
+const StyledPollingNumber = styled(TYPE.small)<{ breathe: boolean; hovering: boolean }>`
+  transition: opacity 0.25s ease;
+  opacity: ${({ breathe, hovering }) => (hovering ? 0.7 : breathe ? 1 : 0.2)};
+  :hover {
+    opacity: 1;
```
moodysalem, line 26: why is the hover css needed in addition to the hover js?

AmagiDDmxh: There were 3 states on the UI

1. js `hovering` with **0.7** opacity, that is activated when its parent is being hover
2. js `breathe` with **1** opacity for mounting, which activated when new block is coming
3. the `css hovering`, when hovering the text, it would be more lighting than hovering on the parent

I guess to anwser your question, the First state above should be clear.

The third one with transition makes it more vigour.

Thats why there were having css and js hover at the same time, because `mounting`, and `hovering` in the parent, should be js controlled.

But hovering in the text do not.

Let me know if that help :>

Issue #1166, opened 2020-10-12
use enum for `enumerateProposalState`


NoahZinsmeister:
in `src/state/governance/hooks`, `enumerateProposalState` should return a typescript enum, and everything consuming governance state should be checking against these values

PR #1381, opened 2021-05-02
use enum for enumerateProposalState #1166

jochenboesmans:
Integer enum used for tracking proposal state in proposal data. Text values are derived from inverse key, value pairs in enum (e.g. ProposalState[0] will evaluate to 'Pending').

This means text values passed to components aren't lowercased anymore, but since styles are now derived from the enum values themselves and the display text is uppercased later, this doesn't cause any changes to the UI.

### DIFF
--- a/src/pages/Vote/styled.tsx
+++ b/src/pages/Vote/styled.tsx
@@ -1,29 +1,25 @@
 import styled from 'styled-components'
+import { ProposalState } from '../../state/governance/hooks'
 
-const handleColorType = (status?: any, theme?: any) => {
+const handleColorType = (status?: ProposalState, theme?: any) => {
   switch (status) {
-    case 'pending':
+    case ProposalState.Pending:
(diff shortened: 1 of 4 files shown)


src/pages/Vote/styled.tsx
```diff
@@ -1,29 +1,25 @@
...
       return theme.red1
-    case 'queued':
-      return theme.text3
-    case 'executed':
-      return theme.green1
-    case 'canceled':
-      return theme.text3
-    case 'expired':
-      return theme.text3
+    case ProposalState.Queued:
```
moodysalem, line 14: can't you just remove these cases and fall into the default?

jochenboesmans: yeah, the explicit handling was merely meant for clarity

moodysalem:
fixes #1166

PR #1397, opened 2021-05-05
try manual approve if erc712 doesn't work

### LINKED ISSUE #1396: v2 to v3 migration signing failing (Metamask + ledger)
This may be a Metamask issue, but I am getting the below signing error as I try to sign to migrate:

_MetaMask Message Signature: Error: Not supported on this device_

Things I have tried:
- I do have contract data set to "Allowed" on the Ledger
- I have updated to the latest version of the app on the Ledger
- I have tried running from different browsers

NoahZinsmeister:
closes #1396


NoahZinsmeister:
might be better to pass in the manual approve as an arg to the permit hook so that we only have to write the fallback logic in one place

PR #1460, opened 2021-05-06
Feature Request: ETH2x-FLI to ETH2X pair routing

Muhd-Amrullah:
This allows direct routing between leveraged tokenset to tokenset pair particularly between ETH2x-FLI and ETH2X pair.

### DIFF
--- a/src/constants/index.ts
+++ b/src/constants/index.ts
@@ -39,6 +39,7 @@ export const TRIBE = new Token(ChainId.MAINNET, '0xc7283b66Eb1EB5FB86327f08e1B58
 export const FRAX = new Token(ChainId.MAINNET, '0x853d955aCEf822Db058eb8505911ED77F175b99e', 18, 'FRAX', 'Frax')
 export const FXS = new Token(ChainId.MAINNET, '0x3432B6A60D23Ca0dFCa7761B7ab56459D9C964D0', 18, 'FXS', 'Frax Share')
 export const renBTC = new Token(ChainId.MAINNET, '0xEB4C2781e4ebA804CE9a9803C67d0893436bB27D', 8, 'renBTC', 'renBTC')
+export const ETH2X-FLI = new Token(ChainId.MAINNET,'0xAa6E8127831c9DE45ae56bB1b0d4D4Da6e5665BD', 18, 'ETH2X-FLI','ETH 2x Flexible Leverage Index')
 
 // Block time here is slightly higher (~1s) than average in order to avoid ongoing proposals past the displayed time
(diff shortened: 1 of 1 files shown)


src/constants/index.ts
```diff
@@ -39,6 +39,7 @@ export const TRIBE = new Token(ChainId.MAINNET, '0xc7283b66Eb1EB5FB86327f08e1B58
 export const FRAX = new Token(ChainId.MAINNET, '0x853d955aCEf822Db058eb8505911ED77F175b99e', 18, 'FRAX', 'Frax')
 export const FXS = new Token(ChainId.MAINNET, '0x3432B6A60D23Ca0dFCa7761B7ab56459D9C964D0', 18, 'FXS', 'Frax Share')
 export const renBTC = new Token(ChainId.MAINNET, '0xEB4C2781e4ebA804CE9a9803C67d0893436bB27D', 8, 'renBTC', 'renBTC')
+export const ETH2X-FLI = new Token(ChainId.MAINNET,'0xAa6E8127831c9DE45ae56bB1b0d4D4Da6e5665BD', 18, 'ETH2X-FLI','ETH 2x Flexible Leverage Index')
```
moodysalem, line 42: doesnt build because dash

moodysalem (changes requested):

src/constants/index.ts
Muhd-Amrullah: I've replaced the dash.

Muhd-Amrullah:
> what's the relationship between ETH2x-FLI and ETH2X? is it your intention in this PR to only add 1-way routing logic for ETH2x but not ETH2x-FLI?

Both are leveraged tokensets that should operate 1:1 at first, however ETH2X will diverge little by little as it takes on a custom tokenset strategy. 

The intention of adding ETH2x-FLI as the connector token is to use it as a base against a range of leveraged tokens such as ETH2X,(next week released BTC2X-FLI) and potentially a combination of others. ETH2x-FLI also has the liquidity it takes to connect and support other 2x leveraged assets. Since their delta (i.e. for every $1 change, it moves $2) is similar, this can potentially reduce some IL. 

The idea is for a person to be able to route "ETH (or other assets) -> ETH2x-FLI -> ETH2X" and back "ETH2X ->ETH2x-FLI -> ETH" (or other assets). Similarly this can be expanded "ETH (or other assets) -> ETH2x-FLI -> BTC2x-FLI" in the future.

Muhd-Amrullah:
I'm not sure how to handle the vercel's errors. I need some assistance on what to do next to push this forward.

moodysalem:
closing in favor of https://github.com/Uniswap/uniswap-interface/pull/1644

PR #1463, opened 2021-05-06
feat: automatic slippage tolerance

moodysalem:


haydenadams:
no longer showing quotes when wallet is not connected - is this a deploy preview thing?

PR #1472, opened 2021-05-07
refactor: use new sdk version

moodysalem:


KeKo6988 (approved):

src/state/lists/wrappedTokenInfo.ts
```diff
@@ -0,0 +1,58 @@
+import { ChainId, Token } from '@uniswap/sdk-core'
+import { TokenInfo } from '@uniswap/token-lists'
+import { isAddress } from '../../utils'
+import { TagInfo } from './hooks'
+
+/**
+ * Token instances created from token info on a token list.
+ */
+export class WrappedTokenInfo implements Token {
```
piavgh, line 9: @moodysalem : I have a question. How does this lazily load the token?

piavgh: Thanks. But for example, if I enable the Coingecko list, the app still needs to run the Token constructor more than 4000 times on the first page load, right?

piavgh: why don't we store the list of token objects in Redux?

moodysalem: the wrapped token info constructor is very cheap, but yes, 4k tokens is going to be slow.

PR #1502, opened 2021-05-10
optimize sandtexture.png with .webp

### LINKED ISSUE #1499: Reduce size of sand texture
Sand texture is currently 1.38MB. Use .webp?

justindomingue:
png image size: 1.38MB, (1.04s)
optimized image size: 241KB, (2ms)

browser support: https://caniuse.com/webp
  * Only concern is Safari - only supported on macOS 11 Big Sur +

fixes #1499


euthenicsDev:
Could also do a fallback to png if you wanted to retain the browser support?
https://usefulangle.com/post/114/webp-image-in-html-with-fallback

PR #1509, opened 2021-05-11
Fix NFT SVG performance issue in browser

### LINKED ISSUE #1466: NFT SVG animation hogs CPU
**Bug Description**
The SVG NFT (liq. position) is very nice but it's heavy on CPU usage !
Hogs like 60% of CPU on latest Firefox/OS X MacBook Air M1 (!!!!)
and most of CPU under an older Ubuntu laptop on a 4th gen i5....

**Steps to Reproduce**

   ...

**Expected Behavior**
Make it optional

**Additional Context**
I really had to block it manually using a plugin. It would make my laptop overheat.

justindomingue:
Disabled NFT SVG animation, only enabled on hover

The animated SVG was causing serious performance issues because of the many layout changes. We should consider changes the SVG definition, but for now, this PR disables animations until hover.

The solution is to snapshot the SVG state onto a static canvas and show that canvas. On hover, show the actual image.

Other approaches considered:
 - SVGSVGElement.pauseAnimations(): unfortunately, the SVG is received as a data:img payload, and inserting it on the page would not be safe
 - Stopping CSS animations: CSS animations are not used in the SVG

Testing:
 - Performance was monitored in Activity Monitor
 - Low CPU usage, spike on hover, returns to low usage on leave
 
✅ macOS: Chrome, Firefox, Safari
✅ Android:  Chrome (animates on tap, stops on tap outside)
 
- only animate NFT SVG on hover by using a canvas
- handle high dpis
- animation transition between canvas and img
- set start state to not animated
- removed animations that were causing issues on Firefox

fixes #1466

### DIFF
--- a/src/pages/Pool/PositionPage.tsx
+++ b/src/pages/Pool/PositionPage.tsx
@@ -194,8 +194,7 @@ function getRatio(lower: Price, current: Price, upper: Price) {
 }
 
 function NFT({ image }: { image: string }) {
-  // Initialize to true to mount the <img> to set the width/height.
-  const [animate, setAnimate] = useState(true)
+  const [animate, setAnimate] = useState(false)
   const animatedStyles = useSpring({ opacity: animate ? 1 : 0 })
   const staticStyles = useSpring({ opacity: animate ? 0 : 1 })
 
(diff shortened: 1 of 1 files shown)


justindomingue:
@moodysalem can you take a look before I merge?

moodysalem (approved):
LGTM

PR #1517, opened 2021-05-12
Ask user to fill in reproduction steps in GitHub issue on page crash.

jochenboesmans:
Hopefully users will actually fill it in.

It also might be useful to look into giving some major components a displayName in order to make the stacktraces a bit more useful in production's minified js as it's currently quite difficult to understand what's going on from the stacktrace alone.

### DIFF
--- a/src/components/ErrorBoundary/index.tsx
+++ b/src/components/ErrorBoundary/index.tsx
@@ -108,6 +108,7 @@ function issueBody(error: Error): string {
 App crashed
 
 **Steps to Reproduce**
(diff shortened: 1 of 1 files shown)


moodysalem:
we could also include the redux store and current URL but i'm hesitant to include the store since it will have transactions they sent (even if they r submitting it). can include a slice of the store maybe

moodysalem:
the error name in the issue title would be a big improvement for marking duplicates

moodysalem:
also should link to support!

moodysalem:
cleaned it up a bit

web3devsa:
apple.com

Issue #1498, opened 2021-05-10
Header overlap makes header tabs hard to read


callil:
On scroll, header should get a solid background when content reaches the bottom of the header (peeking header).

<img width="869" alt="Screen Shot 2021-05-10 at 10 51 44 AM" src="https://user-images.githubusercontent.com/1355319/117714502-0a921c80-b1a5-11eb-8bb0-893eb1a2e9fa.png">

<img width="1016" alt="Screen Shot 2021-05-10 at 10 51 47 AM" src="https://user-images.githubusercontent.com/1355319/117714509-0bc34980-b1a5-11eb-9210-49a96cf6e506.png">

PR #1531, opened 2021-05-12
add header background on scroll body overlap

justindomingue:
using @react-hook/window-scroll to avoid performance hits and readability


justindomingue:
Screen Shot 2021-05-12 at 11 16 53 AM

moodysalem:
hmmmmm what about sliding it out of the way?

justindomingue:
> hmmmmm what about sliding it out of the way?

@moodysalem, @callil mentioned it, much better!

justindomingue:
fixes #1498

PR #1541, opened 2021-05-12
use styled-components/macro to display className in dev builds

justindomingue:
to help new hires (and probably everyone?) find components through the Chrome dev tools

will not show up in prod builds thanks for babel-plugin-macros.config (tested with yarn build && serve -s build)

Resources:
* https://styled-components.com/docs/tooling#babel-macro
* https://github.com/kentcdodds/babel-plugin-macros/blob/main/other/docs/user.md#using-with-create-react-app

image


moodysalem:
is dev mode any slower with this?

moodysalem:
i think i might have preferred changing the resolution of `styled-components`, since now we have to remember to import `styled-components/macro` everywhere

PR #1553, opened 2021-05-12
allow fee collection/liquidity removal in weth

NoahZinsmeister:


src/pages/Pool/PositionPage.tsx
```diff
@@ -269,8 +270,11 @@ export function PositionPage({
   const currency0 = token0 ? unwrappedToken(token0) : undefined
   const currency1 = token1 ? unwrappedToken(token1) : undefined
 
+  // flag for receiving WETH
+  const [receiveWETH, setReceiveWETH] = useState(false)
```
justindomingue, line 274: shouldn't this be stored in global state?

NoahZinsmeister: i'm just a liiittle bit worried that someone might toggle it for 1-time use or just to see what it does, and then mistakenly collect in WETH instead of ETH. i think the use-case of collecting WETH is rare enough that we can impose the burden of toggling every time? also you arguably might want to collect fees vs burn proceeds differently

justindomingue: LGTM

justindomingue (approved):

PR #1555, opened 2021-05-12
handle ape= query parameter

justindomingue:


src/hooks/useExpertModeQueryParamReader.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  useEffect(() => {
+    if (!search) return
+
+    const parsed = parse(search, {
+      parseArrays: false,
+      ignoreQueryPrefix: true,
+    })
+
+    const expertMode = parsed['expert_mode'] || parsed.ape
```
moodysalem, line 23: literally don't even make expert_mode an option

src/hooks/useExpertModeQueryParamReader.ts
```diff
@@ -0,0 +1,31 @@
...
+  return useExpertModeQueryParamReader(search)
+}
+
+function useExpertModeQueryParamReader(search: string): null {
+  const dispatch = useDispatch<AppDispatch>()
+
+  useEffect(() => {
+    if (!search) return
+
+    const parsed = parse(search, {
```
moodysalem, line 18: we have something called useParsedQuery... for this

src/hooks/useExpertModeQueryParamReader.ts
```diff
@@ -0,0 +1,31 @@
...
+      parseArrays: false,
+      ignoreQueryPrefix: true,
+    })
+
+    const expertMode = parsed['expert_mode'] || parsed.ape
+
+    if (typeof expertMode !== 'string') return
```
moodysalem, line 25: if ape is present but not set won't expertMode be undefined?

justindomingue: removed expertmode

src/hooks/useExpertModeQueryParamReader.ts
justindomingue: nice, thanks.. didn't know that existed

src/hooks/useApeModeQueryParamReader.ts
```diff
@@ -0,0 +1,21 @@
...
+  useApeModeQueryParamReader()
+  return null
+}
+
+function useApeModeQueryParamReader() {
+  const dispatch = useDispatch<AppDispatch>()
+  const { ape } = useParsedQueryString()
+
+  useEffect(() => {
+    if (typeof ape !== 'string') return
```
moodysalem, line 17: if they just do `?ape` is the key in the parsed qs object but set to undefined?

src/hooks/useApeModeQueryParamReader.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    dispatch(updateUserExpertMode({ userExpertMode: ape.toLowerCase() === 'true' }))
```
moodysalem, line 19: i think we should only dispatch this if `ape` is undefined or equal to `true`, otherwise do not dispatch an action that disables expert mode

src/hooks/useApeModeQueryParamReader.ts
justindomingue: set to empty string

src/hooks/useApeModeQueryParamReader.ts
justindomingue: should ?ape and ?ape= enable expert mode?

moodysalem (approved):

PR #1565, opened 2021-05-13
clean up svg animations

NoahZinsmeister:
this PR removes the image jump on mouseExit

i'm pretty sure there is also a way to save the src on mouseExit and re-hydrate it on subsequent mouseEnter, but I haven't figured it out yet...

### DIFF
--- a/src/pages/Pool/PositionPage.tsx
+++ b/src/pages/Pool/PositionPage.tsx
@@ -1,4 +1,4 @@
-import React, { SyntheticEvent, useCallback, useMemo, useRef, useState } from 'react'
+import React, { useCallback, useMemo, useRef, useState } from 'react'
 import { NonfungiblePositionManager, Pool, Position } from '@uniswap/v3-sdk'
 
 import { PoolState, usePool } from 'hooks/usePools'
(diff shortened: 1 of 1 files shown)


src/pages/Pool/PositionPage.tsx
```diff
@@ -200,20 +200,11 @@ function getRatio(
...
-
-    const { current: canvas } = canvasRef
-    const context = canvas.getContext('2d')
-
-    if (!context) return
+// snapshots a src img into a canvas
+function getSnapshot(src: HTMLImageElement, canvas: HTMLCanvasElement, targetHeight: number) {
+  const context = canvas.getContext('2d')
 
+  if (context) {
```
justindomingue, line 207: do we have a style guide for these sort of things? I tend to prefer early exits to prevent nested indentation but I can see this being more readable too.

justindomingue (approved):

src/pages/Pool/PositionPage.tsx
NoahZinsmeister: honestly if prettier doesn't fix it i don't really have a strong preference, happy to go either way really

PR #1574, opened 2021-05-14
fix header elements overlapping at around before upToMedium breakpoint

justindomingue:
Keeps the HeaderLinks centered unless the right section needs to grow

image

     ⬇️

image


callil:
If possible the 2nd state (overlap fix) should appear as so: Screen Shot 2021-05-14 at 2 02 40 PM

justindomingue:
> If possible the 2nd state (overlap fix) should appear as so: Screen Shot 2021-05-14 at 2 02 40 PM

That's tricky without new media queries (would rather not have custom media queries outside of theme.js) or js

RonSherfey (changes requested):
request changes

PR #1593, opened 2021-05-16
Add etherscan links for the two tokens in the pool on the NFT page

jab416171:


NoahZinsmeister:
thanks @jab416171 ! would you mind styling these links like the "Charts" tab button, including hover effects and the link icon? also, we shouldn't render a link for ETH (WETH), so you can probably just use a check against `currencyQuote?.address` or something similar, as ETH won't have that field but tokens will

Screen Shot 2021-05-17 at 12 02 30 PM

jab416171:
ETH does have an address (well, it's the address for WETH) so now the code is just checking for the WETH address.

PR #1610, opened 2021-05-17
Leverage Trade api for USDC price with v2 & v3 support

### LINKED ISSUE #1552: USD value way off for token with only eth pair
**Bug Description**
For this token which only has an eth pair, the quoted USD value is way off and the app says the price impact is too high.

**Steps to Reproduce**

1. Go to uniswap
2. Enter 0xc6a788f47199ce99fb08a8c8bd16aa489209d9a5 for the address for pixl
3. Try to swap, notice the price impact
   ...

**Expected Behavior**
The price for tokens which only has an eth pairing should be the same as the eth price

**Additional Context**
image

justindomingue:
- Leverage Trade API tradeExactOut(USDC) rather than custom impl. for eth-usdc-currency pairs
- Also adds support for v3 prices as fallback when v2 prices not available (relevant for pairs that do not have (sufficient) liquidity in v2)

fixes #1552

### DIFF
--- a/src/hooks/useUSDCPrice.ts
+++ b/src/hooks/useUSDCPrice.ts
@@ -1,81 +1,34 @@
-import { ChainId, Currency, CurrencyAmount, currencyEquals, Price, Token, WETH9 } from '@uniswap/sdk-core'
-import JSBI from 'jsbi'
+import { Currency, CurrencyAmount, Price, Token } from '@uniswap/sdk-core'
 import { useMemo } from 'react'
 import { USDC } from '../constants'
-import { PairState, useV2Pairs } from './useV2Pairs'
-import { useActiveWeb3React } from '../hooks'
-import { wrappedCurrency } from '../utils/wrappedCurrency'
+import { useV2TradeExactOut } from './useV2Trade'
+import { tryParseAmount } from 'state/swap/hooks'
+import { useBestV3TradeExactOut, V3TradeState } from './useBestV3Trade'
+
+// USDC amount used when calculating spot price for a given currency.
+// The amount is large enouhg to filter low liquidity pairs.
+const usdcCurrencyAmount = tryParseAmount('10000', USDC)
 
 /**
  * Returns the price in USDC of the input currency
  * @param currency currency to compute the USDC price of
  */
 export default function useUSDCPrice(currency?: Currency): Price<Currency, Token> | undefined {
-  const { chainId } = useActiveWeb3React()
-  const wrapped = wrappedCurrency(currency, chainId)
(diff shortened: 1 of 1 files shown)


src/hooks/useUSDCPrice.ts
```diff
@@ -1,34 +1,30 @@
...
-import { useActiveWeb3React } from '../hooks'
-import { wrappedCurrency } from '../utils/wrappedCurrency'
+import { useV2TradeExactOut } from './useV2Trade'
+import { tryParseAmount } from 'state/swap/hooks'
+import { useBestV3TradeExactOut, V3TradeState } from './useBestV3Trade'
+import { useActiveWeb3React } from 'hooks'
+
+// USDC amount used when calculating spot price for a given currency.
+// The amount is large enough to filter low liquidity pairs.
+const usdcCurrencyAmount = tryParseAmount('10000', USDC)
```
moodysalem, line 11: you can do something like `CurrencyAmount.fromRawAmount(USDC, 10_000e6)`

src/hooks/useUSDCPrice.ts
```diff
@@ -40,42 +36,17 @@ export default function useUSDCPrice(currency?: Currency): Price<Currency, Token
...
-    // handle usdc
-    if (wrapped.equals(USDC)) {
-      return new Price(USDC, USDC, '1', '1')
+    // use v2 price if available, v3 as fallback
+    if (v2USDCTrade) {
+      const { numerator, denominator } = v2USDCTrade.route.midPrice
+      return new Price(currency, USDC, denominator, numerator)
+    } else if (v3USDCTrade.state === V3TradeState.VALID && v3USDCTrade.trade) {
+      const { numerator, denominator } = v3USDCTrade.trade.route.midPrice
+      return new Price(currency, USDC, denominator, numerator)
```
moodysalem, line 45: can you not just return midPrice? currencies should be correct

moodysalem (approved):

src/hooks/useUSDCPrice.ts
justindomingue: midPrice is Price<Currency, Currency> rather than Price<Currency, Token>. Probably makes sense to build a Price<Currency, Token> rather than force the type of midPrice?

moodysalem: oh. we don't particularly care about the currency though (we don't use it anywhere)

justindomingue: hm changing to price<currency,currency> breaks a lot of downstream types (e.g. fiatvalue in currencyinputpanel). might just leave as-is to avoid scope creep

PR #1644, opened 2021-05-19
feat: ETH2x-FLI to ETH2X pair routing

moodysalem:


Muhd-Amrullah:
Thank you Moody for putting it neatly. ETH -> ETH2x-FLI -> ETH2x works nicely. 

**1** key observation is the routing mechanism works nicely when the initial is at ETH, but I observe it outputs blank when it tries to do a third 'hop'. For example if I use USDC or WBTC as the starting asset, it won't be able to go USDC -> ETH -> ETH2x-FLI -> ETH2x. Very okay with this limitation but keen to point out this observation/behaviour.

**Screenshot A: ETH -> ETH2x-FLI -> ETH2X shows perfectly.**
Screenshot 2021-05-21 at 9 41 06 AM

**Screenshot B: WBTC -> ETH -> ETH2x-FLI -> ETH2X does not display values**
Screenshot 2021-05-21 at 9 40 12 AM

**Screenshot C: USDC -> ETH -> ETH2x-FLI ETH2X does not display values**
Screenshot 2021-05-21 at 9 39 49 AM

 **Screenshot D: USDC -> ETH -> ETH2x-FLI shows perfectly.**
Screenshot 2021-05-21 at 9 47 08 AM

moodysalem:
> **1** key observation is the routing mechanism works nicely when the initial is at ETH, but I observe it outputs blank when it tries to do a third 'hop'. For example if I use USDC or WBTC as the starting asset, it won't be able to go USDC -> ETH -> ETH2x-FLI -> ETH2x. Very okay with this limitation but keen to point out this observation/behaviour.

That's because we limit the V3 routes to 2 hops, on this line
https://github.com/Uniswap/uniswap-interface/blob/5e30a4b4ad9c2c1027ea513e6cf654fbfd8589e7/src/hooks/useAllV3Routes.ts#L62

Muhd-Amrullah:
Alright I think we are all set. I will do a check for demand from users and may come back to it in the future. Likely a separate feature request in the future.

PR #1651, opened 2021-05-19
fix: stabilize fiat value by fixing max hop and increasing USDC out to 100k

justindomingue:
will update v3 in a follow-up PR

### DIFF
--- a/src/hooks/useV2Trade.ts
+++ b/src/hooks/useV2Trade.ts
@@ -36,15 +36,14 @@ const MAX_HOPS = 3
  */
 export function useV2TradeExactIn(
   currencyAmountIn?: CurrencyAmount<Currency>,
-  currencyOut?: Currency
+  currencyOut?: Currency,
+  { maxHops = MAX_HOPS } = {}
 ): Trade<Currency, Currency, TradeType.EXACT_INPUT> | null {
   const allowedPairs = useAllCommonPairs(currencyAmountIn?.currency, currencyOut)
 
-  const [singleHopOnly] = useUserSingleHopOnly()
-
   return useMemo(() => {
     if (currencyAmountIn && currencyOut && allowedPairs.length > 0) {
-      if (singleHopOnly) {
+      if (maxHops === 1) {
         return (
           Trade.bestTradeExactIn(allowedPairs, currencyAmountIn, currencyOut, { maxHops: 1, maxNumResults: 1 })[0] ??
           null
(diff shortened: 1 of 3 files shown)


src/hooks/useV2Trade.ts
```diff
@@ -70,26 +70,28 @@ export function useV2TradeExactIn(
 
 /**
  * Returns the best trade for the token in to the exact amount of token out
+ * @param forcedMaxHops Sets max hops to `forcedMaxHops` regardless of useUSerSingleHopOnly
```
moodysalem, line 73: strange to pull from state and have an override, when you can just take the max hops number as an argument and pass in 2 in one location and the result of state in another

src/hooks/useV2Trade.ts
```diff
@@ -73,12 +73,11 @@ export function useV2TradeExactIn(
  */
 export function useV2TradeExactOut(
   currencyIn?: Currency,
-  currencyAmountOut?: CurrencyAmount<Currency>
+  currencyAmountOut?: CurrencyAmount<Currency>,
+  { maxHops = MAX_HOPS, singleHopOnly }: { singleHopOnly?: boolean; maxHops?: number } = {}
```
moodysalem, line 77: what do you need singleHopOnly for? can't you just pass 1 for maxHops?

src/hooks/useV2Trade.ts
```diff
@@ -89,7 +88,7 @@ export function useV2TradeExactOut(
       }
       // search through trades with varying hops, find best trade out of them
       let bestTradeSoFar: Trade<Currency, Currency, TradeType.EXACT_OUTPUT> | null = null
-      for (let i = 1; i <= MAX_HOPS; i++) {
+      for (let i = 1; i <= maxHops; i++) {
```
moodysalem, line 91: a) delete the constant and b) make the same change to the other method

src/hooks/useV2Trade.ts
```diff
@@ -65,31 +64,30 @@ export function useV2TradeExactIn(
...
+  { maxHops = MAX_HOPS } = {}
 ): Trade<Currency, Currency, TradeType.EXACT_OUTPUT> | null {
   const allowedPairs = useAllCommonPairs(currencyIn, currencyAmountOut?.currency)
 
-  const [singleHopOnly] = useUserSingleHopOnly()
-
   return useMemo(() => {
     if (currencyIn && currencyAmountOut && allowedPairs.length > 0) {
-      if (singleHopOnly) {
+      if (maxHops === 1) {
```
moodysalem, line 82: this early exit seems kinda unnecessary but otherwise lgtm

moodysalem (approved):

PR #1676, opened 2021-05-21
feat(pool): add toggle to hide inactive positions

justindomingue:


src/state/user/hooks.tsx
```diff
@@ -137,6 +138,23 @@ export function useUserSlippageTolerance(): Percent | 'auto' {
   )
 }
 
+export function useUserHideClosedPositions(): [boolean, (newHideClosedPositions: boolean) => void] {
+  const dispatch = useDispatch<AppDispatch>()
+
+  const hideClosedPositions = useSelector<AppState, AppState['user']['userHideClosedPositions']>(
+    (state) => state.user.userHideClosedPositions
```
moodysalem, line 145: fyi this will be undefined if state is loaded from `localStorage`, if you want to add something to the 'updateVersion' handler to set it to false or just treat undefined like false in this hook

src/pages/Pool/index.tsx
```diff
@@ -92,14 +94,30 @@ const MainContentWrapper = styled.main`
...
 export default function Pool() {
   const { account } = useActiveWeb3React()
   const toggleWalletModal = useWalletModalToggle()
   const { t } = useTranslation()
   const theme = useContext(ThemeContext)
+  const [userHideClosedPositions, setUserHideClosedPositions] = useUserHideClosedPositions()
 
   const { positions, loading: positionsLoading } = useV3Positions(account)
 
+  const sortedPositions = positions?.sort((a, b) => (a.liquidity?.sub(b.liquidity ?? 0).isNegative() ? 1 : -1)) ?? []
```
moodysalem, line 116: sorting by liquidity is pretty much arbitrary (depends on position range, and token decimals, has no relation to dollar value), token id is going to be more meaningful

src/pages/Pool/index.tsx
```diff
@@ -174,6 +192,16 @@ export default function Pool() {
 
             <CTACards />
 
+            {positions?.some((p) => p.liquidity?.isZero()) ? (
+              <ShowInactiveToggle>
+                <TYPE.darkGray>{t('Hide closed positions')}</TYPE.darkGray>
```
moodysalem, line 197: this is separate but we shouldn't use the english translation for translation keys (can be addressed with the larger task of improving i18n throughout the app), see https://github.com/Uniswap/uniswap-interface/pull/1026#issuecomment-675541852 for discussion

src/pages/Pool/index.tsx
justindomingue: I was mostly just trying to put the closed positions at the end of the list. I'll update with sorting by tokenId for active and liquidity zero at the end

src/pages/Pool/index.tsx
justindomingue: Ack.

src/state/user/hooks.tsx
justindomingue: I think that should be okay. Tempted to leave as-is for consistency with the other boolean flags in state.

PR #1678, opened 2021-05-21
improvement(swap): hide swap placeholders to reduce visual complexity

justindomingue:


moodysalem:
maybe prefer `improvement(swap):` so it wouldn't create a minor version release

PR #1692, opened 2021-05-24
feat(i18n): improve i18n

justindomingue:
replace i18next with linguijs
- extract and compile tool built-in (reconciles translations at build time)
- use po format for development and js for prod
- replace most `t` with `Trans` component

follow-up pr to wrap hard coded english text around `Trans`

To test: https://uniswap-interface-git-lingui-uniswap.vercel.app/?lang=pseudo-en


src/index.tsx
```diff
@@ -63,17 +63,19 @@ ReactDOM.render(
...
-            <Updaters />
-            <ThemeProvider>
-              <ThemedGlobalStyle />
-              <HashRouter>
-                <App />
-              </HashRouter>
-            </ThemeProvider>
-          </Provider>
-        </Blocklist>
+        <LanguageProvider>
```
moodysalem, line 66: should it be at the top just below strict mode? doesn't depend on anything else, and theoretically the theme might depend on it

justindomingue: will move below strict mode for now and consider integrating with redux in a follw-up cl

src/components/Badge/RangeBadge.tsx
```diff
@@ -40,35 +40,48 @@ export default function RangeBadge({
...
-            <BadgeText>{t('In range')}</BadgeText>
+            <BadgeText>
+              <Trans id="pools.inRange">In range</Trans>
+            </BadgeText>
           </Badge>
         </MouseoverTooltip>
       ) : (
         <MouseoverTooltip
-          text={`The price of this pool is outside of your selected range. Your position is not currently earning fees.`}
+          text={t({
```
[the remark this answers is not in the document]
justindomingue: MouseoverTooltip enforces text:string right now. Will leave as-is to avoid scope creep of this PR.

src/components/Web3Status/index.tsx
```diff
@@ -124,7 +124,11 @@ function newTransactionsFirst(a: TransactionDetails, b: TransactionDetails) {
 }
 
 const SOCK = (
-  <span role="img" aria-label="has socks emoji" style={{ marginTop: -4, marginBottom: -4 }}>
+  <span
+    role="img"
+    aria-label={t({ id: 'wallet.hasSocks', message: 'has socks emoji' })}
```
moodysalem, line 129: does this work? i think not because it's a constant, and not a function component

src/i18n.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
+import { i18n } from '@lingui/core'
+import { I18nProvider } from '@lingui/react'
+import { detect, fromUrl } from '@lingui/detect-locale'
+import { ReactNode, useEffect } from 'react'
+
+export const locales = ['en', 'it']
+export const defaultLocale = 'en'
+
+locales.map((locale) => i18n.loadLocaleData(locale, { plurals: () => null }))
```
moodysalem, line 10: please explain this line in a comment, and since you aren't using the return value it should be `forEach`. e.g. why can't this be done with the locale is detected?

also i guess we don't care about using plurals?

src/i18n.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const detected =
+    detect(
+      fromUrl('lang'), // helps local development
+      defaultLocale
+    ) ?? defaultLocale
+  return locales.includes(detected) ? detected : defaultLocale
+}
+
+export async function dynamicActivate(locale: string) {
+  const { messages } = await import(`@lingui/loader!./locales/${locale}/messages.po`)
```
moodysalem, line 22: isn't it going to be included in the build like this? also interpolating the locale makes me think webpack would not know to include the file. how would this be excluded from the production build?

src/locales/en/messages.js
```diff
@@ -0,0 +1 @@
...:"Learn","pools.menu.migrate":"Migrate","pools.menu.more":"More","pools.menu.v2Liquidity":"V2 liquidity","pools.minPriceHint":["Your position will be 100% composed of ",["0"]," at this price"],"pools.minPriceLabel":"Min Price","pools.outOfRange":"Out of range","pools.outOfRangeTooltip":"The price of this pool is outside of your selected range. Your position is not currently earning fees.","pools.removeLiquidity":"Remove Liquidity","selectLiquidityRange":"selectLiquidityRange","swap.buttons.priceImpactTooHigh":"Price Impact Too High","swap.buttons.swap":"Swap","swap.buttons.swapAnyway":"Swap Anyway","tokenLists.manage":"Manage Token Lists","transactions.addLiquidity":["Add ",["0"],"/",["1"]," V3 liquidity"],"transactions.confirmation.button":"Add","transactions.confirmation.pendingText":["Supplying ",["0"]," ",["1"]," ",["2"]," ",["3"]," ",["4"]],"transactions.createPoolAndAddLiquidity":["Create pool and add ",["0"],"/",["1"]," V3 liquidity"],"transactions.pending":[["0"]," Pending"],"uni.claim":"Claim UNI","uni.claiming":"Claiming UNI","unsupportedAsset":"Unsupported Asset","wallet.connect":"Connect a wallet","wallet.ethBalance":[["0"]," ETH"],"wallet.hasSocks":"has socks emoji"}};
```
moodysalem, line 1: typically generated code should not be committed (e.g. the typescript types)

src/locales/en/messages.po
```diff
@@ -0,0 +1,356 @@
+msgid ""
```
moodysalem, line 1: any plan to get the rest of the existing translations in here? in a way that isn't too labor intensive? or we can just get a new pass of translators after everything has been wrapped in `Trans`

justindomingue: Current plan:
-Wrap all English text in app around <Trans>
-Review and get approval on all keys
-Generate new {locale}.po files for languages we support
-Auto-fill {locale}.po with previous translations (hoping to automate with a script)
-Notify translators to translate remaining keys

src/components/Web3Status/index.tsx
justindomingue: that's why I'm using t():string here instead of <Trans>. t returns a string.

src/pages/Pool/index.tsx
```diff
@@ -200,7 +202,9 @@ export default function Pool() {
 
             {closedPositions.length > 0 ? (
               <ShowInactiveToggle>
-                <TYPE.darkGray>{t('Hide closed positions')}</TYPE.darkGray>
+                <TYPE.darkGray>
+                  {<Trans id="pools.toggle.hideClosedPositions">Hide closed positions</Trans>}
```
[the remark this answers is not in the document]
justindomingue: good catch

src/components/Web3Status/index.tsx
moodysalem: i mean the SOCKS is a constant and is only evaluated at compile time so the label wouldnt change when the locale changes

src/i18n.tsx
justindomingue: I think we'll want to add plurals soon enough. I didn't needit yet. It increases the bundle size.

src/components/Web3Status/index.tsx
justindomingue: You're right.. it doesn't change. I'll revert this until I figure out a better solution.

src/components/Badge/RangeBadge.tsx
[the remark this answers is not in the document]
justindomingue: Will add an item to change to component so we support switching language. Leaving as t for now.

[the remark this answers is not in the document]
moodysalem: does t not support language switching? i don't think it would make a diff

src/components/Web3Status/index.tsx
moodysalem: just make it a function component instead of a const

src/i18n.tsx
justindomingue: lingui.config.js has a default locale which I believe will always be included in the bundle.

Non-default locales will be downloaded on the fly.

callil (approved):

src/components/NavigationTabs/index.tsx
```diff
@@ -80,7 +79,9 @@ export function FindPoolTabs({ origin }: { origin: string }) {
         <HistoryLink to={origin}>
           <StyledArrowLeft />
         </HistoryLink>
-        <ActiveText>Import Pool</ActiveText>
+        <ActiveText>
+          <Trans id="nav.importPools">Import Pools</Trans>
```
smsunarto, line 83: Might have added an extra 's' here, or perhaps the original text misses the 's'?

src/components/NavigationTabs/index.tsx
```diff
@@ -118,7 +119,13 @@ export function AddRemoveTabs({
           <StyledArrowLeft stroke={theme.text2} />
         </HistoryLink>
         <TYPE.mediumHeader fontWeight={500} fontSize={20}>
-          {creating ? 'Create a pair' : adding ? 'Add Liquidity' : 'Remove Liquidity'}
+          {creating ? (
+            <Trans id="pools.buttons.createPair">Create a pair</Trans>
```
smsunarto, line 123: Capitalization seems to be inconsistent

src/pages/PoolFinder/index.tsx
```diff
@@ -112,7 +119,7 @@ export default function PoolFinder() {
             </Row>
           ) : (
             <Text fontWeight={500} fontSize={20} marginLeft={'12px'}>
-              Select a Token
+              <Trans id="currencies.search.selectToken">Select a token</Trans>
```
smsunarto, line 122: Capitalization changed?

src/pages/PoolFinder/index.tsx
```diff
@@ -136,7 +143,7 @@ export default function PoolFinder() {
...
```
smsunarto, line 146: Capitalization changed?

src/components/NavigationTabs/index.tsx
justindomingue: nice catch!

src/components/NavigationTabs/index.tsx
justindomingue: I think I'll leave this one as-is fo rnow

src/pages/PoolFinder/index.tsx
justindomingue: That oe was intentional. The string 'Select a token' is used throughout the app so I just standardized it.

src/pages/PoolFinder/index.tsx
justindomingue: ditto

src/components/Web3Status/index.tsx
moodysalem: I was saying, instead of doing `const SOCK = (<jsx.../>)`

do 
```
const Sock = function Sock() {
  return <jsx.../>
}
```

This will fix the translation issue

PR #1714, opened 2021-05-26
chore(i18n): wrapped English text around <Trans> component

justindomingue:
adds more english text to translate


src/components/WalletModal/PendingView.tsx
```diff
@@ -82,20 +83,22 @@ export default function PendingView({
         <LoadingWrapper>
           {error ? (
             <ErrorGroup>
-              <div>Error connecting.</div>
+              <div>
+                <Trans id="error.connecting">Error connecting</Trans>
```
smsunarto, line 87: missing dot ".'

src/components/earn/UnstakingModal.tsx
```diff
@@ -60,58 +64,78 @@ export default function UnstakingModal({ isOpen, onDismiss, stakingInfo }: Staki
 
   let error: string | undefined
   if (!account) {
-    error = 'Connect Wallet'
+    error = t({ id: 'wallet.connect', message: 'Connect a wallet' })
```
smsunarto, line 67: Not sure if the text change here is intentional

src/components/WalletModal/PendingView.tsx
justindomingue: intentional for key consistency, thanks for flagging it!

src/components/earn/UnstakingModal.tsx
justindomingue: intentional as well for consistency across the app 👍

src/components/PositionCard/Sushi.tsx
```diff
@@ -48,7 +49,13 @@ export default function SushiPositionCard({ tokenA, tokenB, liquidityToken, bord
           <AutoRow gap="8px">
             <DoubleCurrencyLogo currency0={currency0} currency1={currency1} size={20} />
             <Text fontWeight={500} fontSize={20}>
-              {!currency0 || !currency1 ? <Dots>Loading</Dots> : `${currency0.symbol}/${currency1.symbol}`}
+              {!currency0 || !currency1 ? (
+                <Dots>
```
moodysalem, line 53: probably dots should just be the `...` part and part of the translation (might want to remove it from other languages)

justindomingue: interetsting.. DOT animates the `...`. That may be hard to localize. Maybe we could just use a spinner or something other iconography. I'll leave as-is for now since there are quite a few instance of Dots.

PR #1723, opened 2021-05-27
feat(i18n): save and load locale from redux store

justindomingue:


src/i18n.tsx
```diff
@@ -1,19 +1,30 @@
...
+  'it-IT' = 'italiano',
+  'iw' = 'Hebrew',
+  'ro' = 'română',
+  'ru' = 'русский',
+  'vi' = 'Tiếng Việt',
+  'zh-CN' = '中文 ( 中国 )',
+  'zh-TW' = '中文 ( 台灣 )',
+  'pseudo-en' = '',
+}
 export const defaultLocale = 'en'
```
moodysalem, line 23: This can be `SupportedLocale.ENGLISH` if you make the above change

src/i18n.tsx
```diff
@@ -1,19 +1,30 @@
...
+  'es-US' = 'español (Estados Unidos)',
+  'it-IT' = 'italiano',
+  'iw' = 'Hebrew',
+  'ro' = 'română',
+  'ru' = 'русский',
+  'vi' = 'Tiếng Việt',
+  'zh-CN' = '中文 ( 中国 )',
+  'zh-TW' = '中文 ( 台灣 )',
+  'pseudo-en' = '',
+}
```
moodysalem, line 22: i think they key of the locale enum should be dev readable, and the value should be the locale, e.g. `en`

a mapping from locale to label should be defined separately where it's rendered, e.g. in the switch component

e.g.
```
const LocaleLabel: {[locale in Locales]: string} = {...}
```

src/i18n.tsx
```diff
@@ -1,19 +1,30 @@
...
 
-const getDetectedLocale = () => {
-  const detected =
-    detect(
-      fromUrl('lang'), // helps local development
-      defaultLocale
-    ) ?? defaultLocale
-  return locales.includes(detected) ? detected : defaultLocale
+export const getNavigatorLocale = (): string | undefined => {
+  const detected = detect(fromNavigator())
```
moodysalem, line 28: i think we might want to parse the locale from navigator manually.. it seems like we need to split it, uppercase the second part, and fallback to the general one if the region specific version isn't available
https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language

src/i18n.tsx
```diff
@@ -24,11 +35,17 @@ export async function dynamicActivate(locale: string) {
 }
 
 export function LanguageProvider({ children }: { children: ReactNode }) {
```
moodysalem, line 39: ok how about in the redux store we store `undefined` or `selectedLocale` and if selectedLocale is undefined or not present in the locale enum values we fallback to detecting the locale

src/index.tsx
```diff
@@ -60,24 +60,24 @@ function Updaters() {
...
-          <Blocklist>
-            <Provider store={store}>
-              <Updaters />
-              <ThemeProvider>
-                <ThemedGlobalStyle />
-                <HashRouter>
+    <FixedGlobalStyle />
+    <Web3ReactProvider getLibrary={getLibrary}>
+      <Web3ProviderNetwork getLibrary={getLibrary}>
+        <Blocklist>
```
moodysalem, line 66: maybe we just need to move the router/provider up instead of the language provider down? neither of those rely on the web3 react provider and the blocklist is also a candidate for translation

src/state/user/actions.ts
```diff
@@ -16,6 +16,8 @@ export interface SerializedPair {
 export const updateMatchesDarkMode = createAction<{ matchesDarkMode: boolean }>('user/updateMatchesDarkMode')
 export const updateUserDarkMode = createAction<{ userDarkMode: boolean }>('user/updateUserDarkMode')
 export const updateUserExpertMode = createAction<{ userExpertMode: boolean }>('user/updateUserExpertMode')
+export const updateUserLocale =
+  createAction<{ userLocale: string; source: 'user' | 'qs' | 'navigator' }>('user/updateUserLocale')
```
moodysalem, line 20: do we need source?

src/state/user/reducer.ts
```diff
@@ -27,6 +28,8 @@ export interface UserState {
   userDarkMode: boolean | null // the user's choice for dark mode or light mode
   matchesDarkMode: boolean // whether the dark mode media query matches
 
+  userLocale: string
```
moodysalem, line 31: imo prefer `string | undefined` and treat undefined as falling back to detection

src/state/user/updater.tsx
```diff
@@ -30,5 +31,13 @@ export default function Updater(): null {
     }
   }, [dispatch])
 
+  // keep locale in sync with the system if a locale is detected.
+  useEffect(() => {
```
moodysalem, line 35: not sure why i did dark mode this way. we should by default always use the detected one (in dark mode's case, from url or system settings), and only set it in the store if the user sets it

moodysalem (changes requested):
can you merge master so the deleted files are removed from the diff?

src/i18n.tsx
justindomingue: Will add localelabel in a follow-up PR where i add the menu

src/i18n.tsx
justindomingue: I agree. I'm tempted to leave it as-is for now. I added a story to the github project to track the work.

src/i18n.tsx
justindomingue: I'll go back to a simple array for supported locales for now. I think the 2 letter code is dev readable enough.

src/state/user/reducer.ts
moodysalem: `SupportedLocale` is an enum? I'm in favor of using an enum instead of string.

src/i18n.tsx
moodysalem: In this case not about readability, rather it's about whether it's type safe (i.e. hard to accidentally break the code)

src/i18n.tsx
```diff
@@ -1,34 +1,43 @@
...
+  'es-US',
+  'it-IT',
+  'iw',
+  'ro',
+  'ru',
+  'vi',
+  'zh-CN',
+  'zh-TW',
+]
+export const defaultFallback = () => 'en'
```
moodysalem, line 21: why is this a function now

src/i18n.tsx
moodysalem: seems like more work than just doing it, it's hardly a story as much as a 5-10 minute coding task

here's the code from the pkg you're using
https://github.com/lingui/js-lingui/blob/main/packages/detect-locale/src/detectors/fromNavigator.ts

src/i18n.tsx
```diff
@@ -1,34 +1,43 @@
...
+    i18n.loadLocaleData(locale, { plurals: () => null })
+    i18n.load(locale, messages)
+    i18n.activate(locale)
+  } catch (error) {
+    console.error(`Failed to load locale data for ${locale}`, error)
+  }
 }
 
 export function LanguageProvider({ children }: { children: ReactNode }) {
-  useEffect(() => {
```
moodysalem, line 27: i think it makes more sense for the dynamic activate effect to live here, and you just compute the locale to use from the state in the redux store + navigator + url in the LanguageProvider component

moodysalem: my rationale is that
a) it keeps the redux state folder separate from user agent/browser stuff
b) this component can get access to the url and user agent because it is intended to be rendered in the dom

src/state/user/updater.tsx
moodysalem: see the above, this code should not be in this updater thing. we should really delete the updater and move the above code that detects dark mode into the theme provider

moodysalem (changes requested):

src/i18n.tsx
justindomingue: all the examples in lingui use a function so I decided to swtich to a function.

src/i18n.tsx
justindomingue: wasn't really about the length but staying focused for this iteration. You're right that it's quick though! Added

src/i18n.tsx
```diff
@@ -1,34 +1,77 @@
...
-      fromUrl('lang'), // helps local development
-      defaultLocale
-    ) ?? defaultLocale
-  return locales.includes(detected) ? detected : defaultLocale
+import { ReactNode } from 'react'
+import { useDispatch } from 'react-redux'
+import useParsedQueryString from 'hooks/useParsedQueryString'
+import { useEffect } from '@storybook/addons'
+import { AppDispatch } from 'state'
+import { useLocale } from 'state/user/hooks'
```
moodysalem, line 9: this is a circular import (even though not a circular dependency) and might make sense to put the locale in a constants file

src/i18n.tsx
```diff
@@ -1,34 +1,77 @@
...
+  'iw',
+  'ro',
+  'ru',
+  'vi',
+  'zh-CN',
+  'zh-TW',
+] as const
+export type SupportedLocale = typeof supportedLocales[number]
+
+const defaultFallback = () => 'en' as SupportedLocale
```
moodysalem, line 31: this is weird. i dont see how this is better than a constant string in any way

src/i18n.tsx
```diff
@@ -1,34 +1,77 @@
...
+
+function parseLocale(maybeSupportedLocale: string): SupportedLocale | undefined {
+  return supportedLocales.find((locale) => locale === maybeSupportedLocale)
+}
+
+function getDetectedLocale(
+  navigator: Partial<Navigator & IE11NavigatorLanguage> = window.navigator
+): SupportedLocale | undefined {
+  const language = navigator.language ?? navigator.language
```
moodysalem, line 40: should be ?? navigator.userLanguage

src/i18n.tsx
```diff
@@ -1,34 +1,77 @@
...
-    ) ?? defaultLocale
-  return locales.includes(detected) ? detected : defaultLocale
+import { ReactNode } from 'react'
+import { useDispatch } from 'react-redux'
+import useParsedQueryString from 'hooks/useParsedQueryString'
+import { useEffect } from '@storybook/addons'
+import { AppDispatch } from 'state'
+import { useLocale } from 'state/user/hooks'
+
+type IE11NavigatorLanguage = {
```
moodysalem, line 11: big doubt we have users on ie11 but you can optionally also extend the navigator type definition in a declaration file

src/i18n.tsx
```diff
@@ -1,34 +1,77 @@
...
+}
+
+function getDetectedLocale(
+  navigator: Partial<Navigator & IE11NavigatorLanguage> = window.navigator
+): SupportedLocale | undefined {
+  const language = navigator.language ?? navigator.language
+
+  if (!language) return undefined
+
+  return parseLocale(language) ?? parseLocale(language.split('-')[0])
```
moodysalem, line 44: what about upper casing the second piece? i recommend doing a split, checking length of pieces and branching instead

src/i18n.tsx
```diff
@@ -1,34 +1,77 @@
...
+    i18n.loadLocaleData(locale, { plurals: () => null })
+    i18n.load(locale, messages)
+    i18n.activate(locale)
+  } catch (error) {
+    console.error(`Failed to load locale data for ${locale}`, error)
+  }
 }
 
 export function LanguageProvider({ children }: { children: ReactNode }) {
+  const dispatch = useDispatch<AppDispatch>()
```
moodysalem, line 63: not used

src/i18n.tsx
```diff
@@ -1,34 +1,77 @@
...
+  navigator: Partial<Navigator & IE11NavigatorLanguage> = window.navigator
+): SupportedLocale | undefined {
+  const language = navigator.language ?? navigator.language
+
+  if (!language) return undefined
+
+  return parseLocale(language) ?? parseLocale(language.split('-')[0])
+}
+
+function getActiveLocale(cached: SupportedLocale | undefined, qs: SupportedLocale | undefined): SupportedLocale {
```
moodysalem, line 47: this function feels really unnecessary

src/i18n.tsx
```diff
@@ -1,34 +1,77 @@
...
+
   useEffect(() => {
-    dynamicActivate(getDetectedLocale()).catch((error) => {
-      console.error('Failed to load locale data', error)
-    })
-  }, [])
+    const parsedLocale = (typeof parsed.lng === 'string' && parseLocale(parsed.lng)) || undefined
+
+    const activeLocale = getActiveLocale(userLocale ?? undefined, parsedLocale)
+    if (activeLocale) {
```
moodysalem, line 71: should always be true

src/i18n.tsx
```diff
@@ -1,34 +1,77 @@
...
+  return parseLocale(language) ?? parseLocale(language.split('-')[0])
+}
+
+function getActiveLocale(cached: SupportedLocale | undefined, qs: SupportedLocale | undefined): SupportedLocale {
+  return qs ?? cached ?? getDetectedLocale() ?? defaultFallback()
+}
+
+export async function dynamicActivate(locale: SupportedLocale) {
+  try {
+    const { messages } = await import(`@lingui/loader!./locales/${locale}.po`)
```
moodysalem, line 53: if the locale loading fails what happens? i tried blocking the request and the app doesnt load so maybe this allows it to load?

moodysalem (changes requested):

src/i18n.tsx
moodysalem: i checked yesterday we had 13

src/i18n.tsx
justindomingue: The I18nProvider isn't rendered until the first call to i18n.activate. I was thinking we could catch activation errors on non-english locale and try to load the english locale then. I don't really see another way around it. 

I would surprised if a particular chunk would conssitently fail to download if the other ones are successful though.

src/i18n.tsx
justindomingue: used to do a little more work. Removing.

src/i18n.tsx
justindomingue: Seems like IE11 actually does support navigator.language:

"navigator.language is the same as navigator.userLanguage for IE."

https://zzz.buzz/2016/01/13/detect-browser-language-in-javascript/#navigatoruserlanguage

src/i18n.tsx
moodysalem: in that case, you probably want to try catch at a diff level? but also not that important to handle b/c there is no good reason for it to fail to load

moodysalem (changes requested):
just one more important change (order of preference)

moodysalem (approved):

PR #1732, opened 2021-05-28
fix: correct etherscan URL when using an ENS name

### LINKED ISSUE #1731: Etherscan links using ENS names are broken
**Bug Description**

When "View on Etherscan" (see screenshot) is clicked while using an ENS address, it leads to a broken page: https://etherscan.io/address/erik.bjareholt.eth

image

**Expected Behavior**

Either point the link to the address-resolved etherscan URL, or to the ENS lookup: https://etherscan.io/enslookup-search?search=erik.bjareholt.eth

**Additional Context**

I'll look into making a PR with a fix. The relevant pieces of code are:

 - https://github.com/Uniswap/uniswap-interface/blob/5e30a4b4ad9c2c1027ea513e6cf654fbfd8589e7/src/components/AccountDetails/index.tsx#L357
 - https://github.com/Uniswap/uniswap-interface/blob/7f1def300d3d036555bcc2e2b2368b9aeb373b8f/src/utils/getExplorerLink.ts#L22

ErikBjare:
Fixes #1731

As proposed in #1731, an option would be to point the URL to the ENS lookup instead (example: https://etherscan.io/enslookup-search?search=erik.bjareholt.eth). But that would require an addition to [`getExplorerLink`](https://github.com/Uniswap/uniswap-interface/blob/7f1def300d3d036555bcc2e2b2368b9aeb373b8f/src/utils/getExplorerLink.ts#L22).

### DIFF
--- a/src/components/AccountDetails/index.tsx
+++ b/src/components/AccountDetails/index.tsx
@@ -354,7 +354,7 @@ export default function AccountDetails({
                           <AddressLink
                             hasENS={!!ENSName}
                             isENS={true}
-                            href={getExplorerLink(chainId, ENSName, ExplorerDataType.ADDRESS)}
+                            href={getExplorerLink(chainId, account, ExplorerDataType.ADDRESS)}
                           >
                             <LinkIcon size={16} />
                             <span style={{ marginLeft: '4px' }}>View on Etherscan</span>


ErikBjare:
Closing since bug was caused by incorrectly configured ENS record.

PR #1734, opened 2021-05-28
feat(i18n): add language selector in interface

justindomingue:
note. not ready for review--polish needed


src/components/Menu/index.tsx
```diff
@@ -139,32 +213,60 @@ export default function Menu() {
...
+                {account && (
+                  <UNIbutton onClick={openClaimModal} padding="8px 16px" width="100%" borderRadius="12px" mt="0.5rem">
+                    Claim UNI
+                  </UNIbutton>
+                )}
+              </MainMenu>
+            )}
+          </Transition>
+
+          <Transition in={activeMenu === 'lang'} timeout={500} unmountOnExit onEnter={calcHeight}>
```
moodysalem, line 253: eh, the animation doesn't feel that impt, might be better without at first?

moodysalem (commented):
maybe we wait to introduce this UI until we have decent translations, and in meantime commit all the other refactoring in this UI to main? e.g. useActiveLocale

PR #1758, opened 2021-05-30
chore: simplify redux types with ts inference

justindomingue:
create new hooks useAppDispatch and useAppSelector that have types embedded

keeping this PR small, will make the change to other files in a follow-up

### DIFF
--- a/src/state/hooks.ts
+++ b/src/state/hooks.ts
@@ -1,0 +1,5 @@
+import { TypedUseSelectorHook, useDispatch, useSelector } from 'react-redux'
+import { AppDispatch, AppState } from 'state'
+
+export const useAppDispatch = () => useDispatch<AppDispatch>()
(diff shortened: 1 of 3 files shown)


src/state/hooks.ts
```diff
@@ -0,0 +1,5 @@
+import { TypedUseSelectorHook, useDispatch, useSelector } from 'react-redux'
+import { AppDispatch, AppState } from 'state'
+
+export const useAppDispatch = () => useDispatch<AppDispatch>()
```
moodysalem, line 4: you can't do the same thing on this line as the next line?

justindomingue: following this, https://react-redux.js.org/using-react-redux/usage-with-typescript

i'd rather keep it in line with the docs

moodysalem (approved):

PR #1764, opened 2021-05-31
Fix @emotion/core deprecation error on npm start

micahriggan:
On a clean `npm install && npm start` I receive an error message about `@emotion/core` being deprecated and replaced with `@emotion/react`

From investigating it looks like storybook has it's own version of `@emotion` and you don't need to install it seperately.

`@storybook/theming` uses `@emotion/core@10.1.1` and the current `@emotion/core` in package.json is deprecated

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,6 @@
   "homepage": ".",
   "private": true,
   "devDependencies": {
-    "@emotion/core": "^11.0.0",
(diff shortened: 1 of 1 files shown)


moodysalem:
i removed it from main

micahriggan:
Thanks!

PR #1765, opened 2021-05-31
move fonts to public/ to support preloading

### LINKED ISSUE #1730: Preload font (woff2) files to avoid LayoutShift on load
**Bug Description**
Font files are loaded by the css chunk, but are known at compile time. The Inter font files (woff2) should be preloaded from the initial HTML so that they are fetched immediately (at low priority this should not negatively impact other loads). This will either make the layout shift occur sooner or (luckily, if they are downloaded first) eliminate the layout shift completely by having the font ready before the first render.

**Steps to Reproduce**

1. Go to http://uniswap.exchange.
2. Use the Chrome DevTools performance pane to record the load. You may need to throttle network and disable cache.
2. Note a LayoutShift when the fonts first load, after the first render.

**Expected Behavior**
Fonts should be preloaded using a <link rel="preload"> tag on the served HTML.

zzmp:
- Adds sass dependency to compile fonts sourced from a custom folder.
- Moves fonts to public/ to support preloading.
- Preloads variable fonts (which covers all modern browsers) to prevent flash of unstyled text (FOUT).

Fixes #1730.


package.json
```diff
@@ -104,6 +104,7 @@
     "react-window": "^1.8.5",
     "rebass": "^4.0.7",
     "redux-localstorage-simple": "^2.3.1",
+    "sass": "^1.34.0",
```
zzmp, line 107: node-sass is recommended by [create-react-app](https://create-react-app.dev/docs/adding-a-sass-stylesheet/) but does not yet support the M1 chip, so I've used upstream sass instead.

public/fonts/inter.scss
```diff
@@ -0,0 +1,23 @@
+@use "inter-ui/default" with (
```
zzmp, line 1: From https://github.com/philipbelesky/inter-ui#quick-start.

PR #1776, opened 2021-06-01
fix: add a link to switch the locale conditionally

moodysalem:
Justin's idea to copy the Google UX

Adds a link below the swap page and a few other locations to switch the locale if browser locale is not english or active locale is not english

Note the text is not yet translated in the below screenshot, but the "Uniswap available in" would be translated in the active locale whereas the language would be in the target locale

image


src/constants/locales.ts
```diff
@@ -35,7 +35,36 @@ export type SupportedLocale = typeof SUPPORTED_LOCALES[number]
...
+  'pt-BR': 'português',
+  'pt-PT': 'português',
+  'ro-RO': 'Română',
+  'ru-RU': 'русский',
+  'sr-SP': 'Српски',
+  'sv-SE': 'svenska',
+  'tr-TR': 'Türkçe',
+  'uk-UA': 'Український',
+  'vi-VN': 'Tiếng Việt',
+  'zh-CN': '中文',
```
justindomingue, line 68: zh-CN: 中文 ( 中国 )
zh-TW: 中文 ( 台灣 )

src/pages/Vote/VotePage.tsx
```diff
@@ -199,161 +200,164 @@ export default function VotePage({
...
+            <RowFixed style={{ width: '100%', gap: '12px' }}>
+              <ButtonPrimary
+                padding="8px"
+                borderRadius="8px"
+                onClick={() => {
+                  setSupport(true)
+                  toggleVoteModal()
+                }}
+              >
+                Vote For
```
justindomingue, line 260: missing <Trans>

PR #1798, opened 2021-06-02
chore(i18n): support localized plural strings

justindomingue:


src/components/SearchModal/ImportToken.tsx
```diff
@@ -68,9 +71,10 @@ export function ImportToken({ tokens, list, onBack, onDismiss, handleCurrencySel
         <AutoColumn justify="center" style={{ textAlign: 'center', gap: '16px', padding: '1rem' }}>
           <AlertCircle size={48} stroke={theme.text2} strokeWidth={1} />
           <TYPE.body fontWeight={400} fontSize={16}>
-            {
-              "This token doesn't appear on the active token list(s). Make sure this is the token that you want to trade."
-            }
+            <Trans>
+              This token doesn&apos;t appear on the active token list(s). Make sure this is the token that you want to
```
moodysalem, line 75: technically this should be pluralized

moodysalem (approved):

src/components/SearchModal/ImportToken.tsx
jab416171: This could change to "any active token list" then it wouldn't need to be pluralized

PR #1799, opened 2021-06-02
feat: Add proposal creation interface

smsunarto:


src/pages/CreateProposal/ProposalEditor.tsx
```diff
@@ -0,0 +1,61 @@
...
+  onTitleInput,
+  onBodyInput,
+}: {
+  className?: string
+  title: string
+  body: string
+  onTitleInput: (title: string) => void
+  onBodyInput: (body: string) => void
+}) => {
+  const bodyPlaceholder = `## Summary
```
justindomingue, line 30: not to scope creep but did you consider adding a markdown preview?

src/pages/CreateProposal/ProposalEditor.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const ProposalEditorHeader = styled.div`
+  font-size: 14px;
+  font-weight: 500;
+  color: ${({ theme }) => theme.text2};
+`
+
+const ProposalTitle = memo(styled(TextInput)`
+  margin-top: 7.5px;
+  margin-bottom: 7.5px;
```
justindomingue, line 14: can be simplified to `margin: 7.5px 0` which I think increases readability.

src/pages/CreateProposal/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+  isFormInvalid,
+  handleCreateProposal,
+}: {
+  hasActiveOrPendingProposal: boolean
+  hasEnoughVote: boolean
+  isFormInvalid: boolean
+  handleCreateProposal: () => void
+}) => {
+  return hasActiveOrPendingProposal ? (
+    <ButtonError marginTop="18px">
```
justindomingue, line 50: does `ButtonError` also disables the button?

smsunarto: Nope, I just realized that i need to add error={true} prop to actually have it show the error style, so currently it's just a no op button. I'm checking with Cal because I'm not sure when to use button disable vs. ButtonError.

src/pages/CreateProposal/ProposalEditor.tsx
smsunarto: Yeap, I think it got p2-ed for now.

src/pages/Vote/index.tsx
```diff
@@ -135,6 +139,13 @@ export default function Vote() {
     uniBalance && JSBI.notEqual(uniBalance.quotient, JSBI.BigInt(0)) && userDelegatee === ZERO_ADDRESS
   )
 
+  // show create proposal button if they pass the proposal threshold
```
justindomingue, line 142: I wonder if we should always show the button but disabled. As a way to upsell governance.

smsunarto: Yep, Hayden mentioned this in Slack. Fixed in my local branch, haven't push the commits.

src/components/AddressInputPanel/index.tsx
```diff
@@ -23,7 +23,7 @@ const ContainerRow = styled.div<{ error: boolean }>`
   border-radius: 1.25rem;
   border: 1px solid ${({ error, theme }) => (error ? theme.red1 : theme.bg2)};
   transition: border-color 300ms ${({ error }) => (error ? 'step-end' : 'step-start')},
-    color 500ms ${({ error }) => (error ? 'step-end' : 'step-start')};
+  color: 500ms ${({ error }) => (error ? 'step-end' : 'step-start')};
```
justindomingue, line 26: I thin this was intentional. color is part of `transition`, not its own property.

smsunarto: fixed, nice catch!

src/components/AddressInputPanel/index.tsx
```diff
@@ -99,7 +105,7 @@ export default function AddressInputPanel({
           <AutoColumn gap="md">
             <RowBetween>
               <TYPE.black color={theme.text2} fontWeight={500} fontSize={14}>
-                Recipient
+                {label ?? 'Recipient'}
```
justindomingue, line 108: might be simpler to just default the value to recipient in the params

`AddressInputPanel({id, className= '...', label = 'Recipient', ...})`

src/components/NavigationTabs/index.tsx
```diff
@@ -132,3 +132,17 @@ export function AddRemoveTabs({
...
+
+export function CreateProposalTabs() {
+  return (
+    <Tabs>
+      <RowBetween style={{ padding: '1rem 1rem 0 1rem' }}>
+        <HistoryLink to="/vote">
+          <StyledArrowLeft />
+        </HistoryLink>
+        <ActiveText>Create Proposal</ActiveText>
+        <div></div>
```
justindomingue, line 144: is this needed?

src/components/TextInput/index.tsx
```diff
@@ -0,0 +1,152 @@
+import React, { memo, useCallback, useRef } from 'react'
+import styled from 'styled-components'
+
+const Input = styled.input<{ error?: boolean; fontSize?: string }>`
```
justindomingue, line 4: Couldn't you use rebass text input here? I think it could simplify the css. Not really needed if this works

src/components/TextInput/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+  ::-webkit-search-decoration {
+    -webkit-appearance: none;
+  }
+
+  ::-webkit-outer-spin-button,
+  ::-webkit-inner-spin-button {
+    -webkit-appearance: none;
+  }
+
+  ::placeholder {
```
justindomingue, line 32: isn't this already added at L17?

src/pages/App.tsx
```diff
@@ -29,6 +29,7 @@ import { RedirectDuplicateTokenIdsV2 } from './AddLiquidityV2/redirects'
 import { PositionPage } from './Pool/PositionPage'
 import AddLiquidity from './AddLiquidity'
 import ApeModeQueryParamReader from 'hooks/useApeModeQueryParamReader'
+import CreateProposal from './CreateProposal'
```
justindomingue, line 32: Consider using `const CreateProposal = React.Lazy(() => import './CreateProposal')` instead to avoid increasing the size of the bundle for all users :D

src/pages/CreateProposal/ProposalActionDetail.tsx
```diff
@@ -0,0 +1,85 @@
...
+import styled from 'styled-components'
+import { ProposalAction } from './ProposalActionSelector'
+import { Currency } from '@uniswap/sdk-core'
+
+enum ProposalActionDetailField {
+  ADDRESS,
+  CURRENCY,
+}
+
+const _ProposalActionDetail = ({
```
justindomingue, line 13: curious why you use an underscore here

src/pages/CreateProposal/ProposalActionDetail.tsx
```diff
@@ -0,0 +1,85 @@
...
+            onCurrencySelect={(currency: Currency) => onCurrencySelect(currency)}
+            showMaxButton={false}
+            showCommonBases={false}
+            showCurrencyAmount={false}
+            disableNonToken={true}
+            hideBalance={true}
+            id="currency-input"
+          />
+        ) : (
+          <></>
```
justindomingue, line 73: probably more idiomatic to return `null`

src/pages/CreateProposal/ProposalActionDetail.tsx
```diff
@@ -0,0 +1,85 @@
...
+        )
+      )}
+    </div>
+  )
+}
+
+export const ProposalActionDetail = styled(_ProposalActionDetail)`
```
justindomingue, line 80: nit. I think we usually put styled components before the function component

src/pages/CreateProposal/ProposalActionSelector.tsx
```diff
@@ -0,0 +1,119 @@
...
+            <Text fontWeight={500} fontSize={16}>
+              <Trans>Select an action</Trans>
+            </Text>
+            <CloseIcon onClick={onDismiss} />
+          </RowBetween>
+        </PaddedColumn>
+        <Separator />
+        <MenuItem onClick={() => handleProposalActionSelect(ProposalAction.TRANSFER_TOKEN)}>
+          <Column>
+            <Text title="Test" fontWeight={500}>
```
justindomingue, line 104: nit. remove title `Test`?

src/pages/CreateProposal/ProposalActionSelector.tsx
```diff
@@ -0,0 +1,119 @@
...
+              <Trans>Transfer Token</Trans>
+            </Text>
+          </Column>
+        </MenuItem>
+        <MenuItem onClick={() => handleProposalActionSelect(ProposalAction.APPROVE_TOKEN)}>
+          <Column>
+            <Text title="Test" fontWeight={500}>
```
justindomingue, line 111: ditto `Test`

src/pages/CreateProposal/ProposalEditor.tsx
```diff
@@ -0,0 +1,61 @@
...
+Lorem ipsum dolor sit amet
+  
+  `
+
+  return (
+    <div className={className}>
+      <ProposalEditorHeader>
+        <Trans>Proposal</Trans>
+      </ProposalEditorHeader>
+      <ProposalTitle value={title} onUserInput={onTitleInput} placeholder="Proposal Title" fontSize="1.5rem" />
```
justindomingue, line 49: Translate? :D 

You can use the t macro if you need a string. t<backtick>Proposal Title<backtick>

src/state/governance/hooks.ts
```diff
@@ -241,3 +251,58 @@ export function useVoteCallback(): {
...
+        createProposalData.calldatas,
+        createProposalData.description,
+      ]
+
+      return govContract.estimateGas.propose(...args, {}).then((estimatedGasLimit) => {
+        return govContract
+          .propose(...args, { value: null, gasLimit: calculateGasMargin(estimatedGasLimit) })
+          .then((response: TransactionResponse) => {
+            addTransaction(response, {
+              summary: `Submitted new proposal`,
```
justindomingue, line 280: t<backtick>Submitted new proposal<backtick>

:D

src/components/TextInput/index.tsx
smsunarto: Most of this is extracted from NumericalInput if i remember correctly.

src/pages/CreateProposal/index.tsx
```diff
@@ -0,0 +1,225 @@
...
+        break
+      case ProposalAction.APPROVE_TOKEN:
+        createProposalData.signatures = ['approve(address,uint256)']
+        calldataValues = [toAddressValue, amountValue]
+        break
+      default:
+        createProposalData.signatures = []
+    }
+
+    const typesArray = createProposalData.signatures[0].split('(').pop()?.split(')')[0]?.split(',')
```
NoahZinsmeister, line 149: a bit too spooky, we should define these per-proposal action in the switch, also this is hard-coded to only work with a length-1 array, should at least have a nod toward it being generic.

the way to do this is probably something like:

```
types = [['address', 'uint256']]
createProposalData.signatures = [`transfer(${types[0].join(',')})`]
```

src/pages/CreateProposal/index.tsx
```diff
@@ -0,0 +1,225 @@
...
+    const bytes = typesArray?.map((type, i) => generateBytesByType(type, calldataValues[i]))
```
NoahZinsmeister, line 150: should use https://docs.ethers.io/v5/api/utils/abi/coder/#AbiCoder-encode

src/state/governance/hooks.ts
```diff
@@ -241,3 +252,58 @@ export function useVoteCallback(): {
...
+export function useLatestProposalId(address: string): string | undefined {
+  const gov = useGovernanceContract()
+  const res = useSingleCallResult(gov, 'latestProposalIds', [address])
+  if (res.result && !res.loading) {
+    return res.result[0]
+  }
+  return undefined
+}
+
+export function useProposalThreshold(): number | undefined {
```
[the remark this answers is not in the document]
NoahZinsmeister: on second thought, should probably return a TokenAmount of UNI

src/state/governance/hooks.ts
```diff
@@ -241,3 +252,58 @@ export function useVoteCallback(): {
...
+        createProposalData.targets,
+        createProposalData.values,
+        createProposalData.signatures,
+        createProposalData.calldatas,
+        createProposalData.description,
+      ]
+
+      return govContract.estimateGas.propose(...args, {}).then((estimatedGasLimit) => {
+        return govContract
+          .propose(...args, { value: null, gasLimit: calculateGasMargin(estimatedGasLimit) })
```
NoahZinsmeister, line 278: shouldn't need to specify `value` here

src/state/governance/hooks.ts
```diff
@@ -241,3 +252,58 @@ export function useVoteCallback(): {
...
+
+      const args = [
+        createProposalData.targets,
+        createProposalData.values,
+        createProposalData.signatures,
+        createProposalData.calldatas,
+        createProposalData.description,
+      ]
+
+      return govContract.estimateGas.propose(...args, {}).then((estimatedGasLimit) => {
```
NoahZinsmeister, line 276: don't need empty overrides arg

NoahZinsmeister (changes requested):

maggievalentine:
Awesome work Scott! From a visual perspective, I think vertically stacking everything (lefthand side on top) rather than having a split horizontal view would improve the flow of actions. Also would suggest making the "to" address input text black or gray rather than pink, since the pink is quite close to the red text that implies an error.

moodysalem:
i think a more specific url like `/create-proposal` might make sense

also i don't see how to add more than one action? is it possible?

exposing the create proposal button to people who are not delegated 10m uni may be confusing, i would favor removing it from the /vote page

NoahZinsmeister (commented):
i made a small cleanup commit, but otherwise changes LGTM, anything else we should think about before merging?

src/state/governance/hooks.ts
```diff
@@ -11,6 +11,7 @@ import { TransactionResponse } from '@ethersproject/providers'
 import { useTransactionAdder } from '../transactions/hooks'
 import { useState, useEffect, useCallback, useMemo } from 'react'
 import { abi as GOV_ABI } from '@uniswap/governance/build/GovernorAlpha.json'
+import { t } from '@lingui/macro'
```
NoahZinsmeister, line 17: is the use of `t` discouraged @justindomingue ? remember seeing something about this

justindomingue: `t` doesn't reload on hooks (locale) change. `<Trans>` is preferred for that reason

https://github.com/lingui/js-lingui/issues/1090

PR #1802, opened 2021-06-03
feat: ETH2x-FLI to BTC2X pair routing

Muhd-Amrullah:
This is the same as #1644 with an addition of BTC2X. This allows direct routing between leveraged tokenset to tokenset pair particularly between ETH2x-FLI and the BTC2X pair, a leveraged BTC token.

The BTC2X token can already be found here: [https://www.tokensets.com/v2/set/0x882B454E9152D1c35C77F9d79399d7DBf07E1b1D](url)

The liquidity between BTC2X and ETH2X-FLI can be found here: [https://info.uniswap.org/#/pools/0x4004d94a4e6a6ca2b745b9b2e9fd262f3c3bdd98](url)

### DIFF
--- a/src/constants/routing.ts
+++ b/src/constants/routing.ts
@@ -67,6 +67,7 @@ export const ADDITIONAL_BASES: { [chainId: number]: { [tokenAddress: string]: To
   [1]: {
     ...mAssetsAdditionalBases,
     '0xF16E4d813f4DcfDe4c5b44f305c908742De84eF0': [ETH2X_FLI],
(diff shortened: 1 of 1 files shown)


Muhd-Amrullah:
I'm not sure if I miss out any steps to get the workflow above to approve. Please advise if there are steps missing.

moodysalem:
This PR cannot be merged as is, and the routing API/smart order router package should handle it, so dropping it for now

PR #1806, opened 2021-06-04
Update fr-FR.po

Lafourkad:
Reviwed the french translation quickly, might still need improvement.


justindomingue:
Thank you so much for the PR @Lafourkad ! Looking great!

We now use (crowdin)[https://crowdin.com/project/uniswap-interface] to manage translations of the app. The .po files in `src/` are uploaded automatically from crowdin. This seems like a lot to ask given the amount of edits you've made to this file but would you be open to creating an account on crowdin and applying the translations there?

PR #1816, opened 2021-06-06
feat: Display spot price in swap

MStreet3:


Burstvn:
hello @MStreet3

src/pages/Swap/index.tsx
```diff
@@ -140,7 +141,7 @@ export default function Swap({ history }: RouteComponentProps) {
     [independentField, parsedAmount, showWrap, trade]
   )
 
-  const fiatValueInput = useUSDCValue(parsedAmounts[Field.INPUT])
+  const fiatValueInput = useUSDCValue(parsedAmounts[Field.INPUT] ?? pricedAmount)
   const fiatValueOutput = useUSDCValue(parsedAmounts[Field.OUTPUT])
```
justindomingue, line 145: any reason you aren't applying this treatment to output?

I think if output token is selected, we'd want to show spot price as well

src/pages/Swap/index.tsx
```diff
@@ -140,7 +141,7 @@ export default function Swap({ history }: RouteComponentProps) {
     [independentField, parsedAmount, showWrap, trade]
   )
 
-  const fiatValueInput = useUSDCValue(parsedAmounts[Field.INPUT])
+  const fiatValueInput = useUSDCValue(parsedAmounts[Field.INPUT] ?? pricedAmount)
```
justindomingue, line 144: did you consider using useUSDCPrice instead?

useUSDCValue calls useUSDCPrice with the currency. I feel like providing pricedAmount in the hook complicates it a bit.

src/pages/Swap/index.tsx
MStreet3: ok thanks, that was a question I had.  Can do as well.

src/pages/Swap/index.tsx
MStreet3: I went this way because all the child components rely on a `CurrencyAmount<Token>` type and `useUSDCPrice` returns `Price<Currency, Token>`.  `pricedAmount` is a unitary amount of the currency, considered it to be the simplest way to use existing interfaces to achieve the display of a fiat value in the `FiatRow` of a `CurrencyInputPanel`.  Thoughts?

MStreet3: essentially, I'm passing `useUSDCValue` either the `parsedAmount` from the input field, or a unitary currency amount (i.e., `pricedAmount`) for the input currency amount (e.g., 1 ETH if there is no parsed amount and ETH is the input token).

geniusj204 (approved):

PR #1819, opened 2021-06-07
fix: behavior of useEffect

YutaSugimura:
**Bug Description**
Fixed a bug that prevented Proposals from being displayed properly on the Vote page.

**Steps to Reproduce**
[https://app.uniswap.org/#/swap](https://app.uniswap.org/#/swap)
1. Switch from the Swap tab to the Vote tab.
2. It does not finish loading and the display does not switch.
3. If you click on a tab other than Vote to switch to it, and then switch back to Vote, it will be displayed correctly. Reloading will also work.

**Expected Behavior**
switching to the Vote tab displays the Proposals correctly.

### DIFF
--- a/src/state/governance/hooks.ts
+++ b/src/state/governance/hooks.ts
@@ -74,14 +74,13 @@ export function useDataFromEventLogs() {
   const previousChainId = usePrevious(chainId)
 
   useEffect(() => {
-    if (!filter || !library || chainId === previousChainId) return
     let stale = false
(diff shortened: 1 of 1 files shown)


moodysalem:
i think the issue is the previous chain id thing which we should just remove

moodysalem:
fixed in https://github.com/Uniswap/uniswap-interface/commit/84c4219c020e4cafe74052f5ce1bc894c5bbcc66

Issue #1810, opened 2021-06-04
Add Liquidity UX - buttons and liquidity distribution chart


willhennessy:
Add a liquidity distribution chart with range sliders to the Add Liquidity UI

- [x] Maggie to deliver UX by 6/13
- [x] Justin to implement

willhennessy:
Teo and I will lock the (a) metric for auto-selecting fee tier and (b) % buttons per fee tier today or tomorrow

willhennessy:
Polish doc: https://docs.google.com/document/d/1bjy7EjFM_w4XmqzzLGoqO-DTE6GiXUbiC6ZOG8aTqhw/edit

justindomingue:
Closing feature. Fast-follows:
#1836 
#1831

PR #1829, opened 2021-06-08
feat(pool): add liquidity distribution range slider and update add liquidity layout

justindomingue:
- display active liquidity in chart
- allows users to select range from chart
- adds buttons to select %
- updates layout for wide screens


justindomingue:
#1810

src/hooks/usePoolTickData.ts
```diff
@@ -0,0 +1,109 @@
...
+  valid: boolean
+  activeTick: number | undefined
+  tickData: TickProcessed[]
+} {
+  const pool = usePool(currencyA, currencyB, feeAmount)
+
+  const tickSpacing = feeAmount && TICK_SPACINGS[feeAmount]
+
+  // Find nearest valid tick for pool in case tick is not initialized.
+  const activeTick =
```
NoahZinsmeister, line 29: hmm, should this be rounded down? if `tickCurrent` is 9 (with a tick spacing of 10), i'd define 0 as the nearest "active" tick, because liquidity is active at ticks >= lower but < upper

src/components/LiquidityDepth/hooks.ts
```diff
@@ -0,0 +1,91 @@
...
+      const newData: ChartEntry[] = []
+      let maxLiquidity = JSBI.BigInt(0)
+
+      for (let i = 0; i < tickData.length; i++) {
+        const t: TickProcessed = tickData[i]
+        const active = t.tickIdx === activeTick
+
+        maxLiquidity = JSBI.greaterThan(tickData[i].liquidityActive, maxLiquidity)
+          ? tickData[i].liquidityActive
+          : maxLiquidity
```
NoahZinsmeister, line 56:
```suggestion
        if (JSBI.greaterThan(tickData[i].liquidityActive, maxLiquidity))
            maxLiquidity = tickData[i].liquidityActive
        }
```

src/components/LiquidityDepth/DensityChart.tsx
```diff
@@ -0,0 +1,191 @@
...
+                      x: [parseFloat(leftPrice?.toSignificant(5)), parseFloat(rightPrice?.toSignificant(5))],
+                    }
+                  : undefined
+              }
+              brushComponent={
+                <Brush
+                  leftHandleColor={currencyA ? tokenAColor : theme.primary1}
+                  leftLabel={
+                    price && leftPrice
+                      ? `${((parseFloat(leftPrice.toSignificant(8)) / parseFloat(price) - 1) * 100).toFixed(2)}%`
```
NoahZinsmeister, line 122: why 8 vs 5 above?

src/components/LiquidityDepth/DensityChart.tsx
```diff
@@ -0,0 +1,191 @@
...
+                      : undefined
+                  }
+                  rightLabel={
+                    price && rightPrice
+                      ? `${((parseFloat(rightPrice.toSignificant(8)) / parseFloat(price) - 1) * 100).toFixed(2)}%`
```
NoahZinsmeister, line 127: ditto

src/components/LiquidityDepth/DensityChart.tsx
```diff
@@ -0,0 +1,191 @@
...
+            y={'activeLiquidity'}
+            interpolation="step"
+          />
+
+          {price && (
+            <VictoryLine
+              data={
+                maxLiquidity && price
+                  ? [
+                      { x: parseFloat(price), y: 0 },
```
NoahZinsmeister, line 159: i think this only works because it tries to use `toString` on price? ideally should always specify e.g. `toSignificant`

src/components/LiquidityDepth/DensityChart.tsx
```diff
@@ -0,0 +1,191 @@
...
+                      { x: parseFloat(price), y: maxLiquidity },
```
NoahZinsmeister, line 160: isn't maxLiquidity a JSBI?

src/components/LiquidityDepth/hooks.ts
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        if (active) {
+          setActiveChartEntry(chartEntry)
+        }
+
+        newData.push(chartEntry)
+      }
+
+      setMaxLiquidity(parseFloat(maxLiquidity.toString()))
```
NoahZinsmeister, line 72: hmm, i see, ideally we keep it a JSBI for as long as possible

src/components/LiquidityDepth/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { DarkBlueCard } from 'components/Card'
+import DensityChart from './DensityChart'
+import { RowBetween } from 'components/Row'
+import { Trans } from '@lingui/macro'
+import { AutoColumn } from 'components/Column'
+import { ChartContext } from './hooks'
+import ZoomButtons from './ZoomButtons'
+
+const MIN_ZOOM = 10
+const MAX_ZOOM = 200
```
NoahZinsmeister, line 13: mm these are a little confusing in the context of full range, also if you can't exceed these, the # of ticks we need data for is dramatically smaller

src/hooks/useAllV3Ticks.ts
```diff
@@ -0,0 +1,108 @@
...
+    () => (tickSpacing ? bitmapIndex(nearestUsableTick(TickMath.MAX_TICK, tickSpacing), tickSpacing) : undefined),
+    [tickSpacing]
+  )
+
+  const [tickDataLatestSynced, setTickDataLatestSynced] = useState<TickData[]>([])
+
+  // fetch the pool address
+  const factoryContract = useV3Factory()
+  const poolAddress = useSingleCallResult(factoryContract, 'getPool', [token0?.address, token1?.address, feeAmount])
+    .result?.[0]
```
NoahZinsmeister, line 44: could look this up w/o network call

src/hooks/useAllV3Ticks.ts
```diff
@@ -0,0 +1,108 @@
...
+    REFRESH_FREQUENCY,
+    3_000_000
+  )
+
+  const error = useMemo(() => callStates.some(({ error }) => error), [callStates])
+  const loading = useMemo(() => callStates.some(({ loading }) => loading), [callStates])
+  const syncing = useMemo(() => callStates.some(({ syncing }) => syncing), [callStates])
+  const valid = useMemo(() => callStates.some(({ valid }) => valid), [callStates])
+
+  const tickData = useMemo(
```
NoahZinsmeister, line 71: let's move this inside the `useEffect` conditional, it should save some cycles, and as long as `error`/`loading`/`syncing`/`valid` are referentially constant (which they should be), the behavior should be the same?

src/hooks/usePoolTickData.ts
```diff
@@ -0,0 +1,109 @@
...
+        error: error || pool[0] === PoolState.INVALID,
+        valid: false,
+        activeTick,
+        tickData: [],
+      }
+    }
+
+    const tickToInitializedTick = keyBy(tickData, 'tick')
+
+    // ensure active tick is within supported range
```
NoahZinsmeister, line 59: hmm this should never happen, right?

src/constants/ticks.ts
```diff
@@ -0,0 +1,16 @@
+import JSBI from 'jsbi'
+
+export interface TickData {
+  tick: number
+  liquidityNet: string
+  liquidityGross: string
```
NoahZinsmeister, line 6: i think we can delete all the `liquidityGross` logic as it doesn't seem like we use it anywhere?

NoahZinsmeister (commented):
first pass

src/hooks/usePoolTickData.ts
justindomingue: Theoretically  not.. That was taken from @uniswap/uniswap-v3-info. Should I remove it?

src/hooks/useAllV3Ticks.ts
justindomingue: how? I could use usePool() which will maybe have a cached response. Is there a more direct way?

src/hooks/usePoolTickData.ts
justindomingue: is there a util to round down based on tick spacing?

NoahZinsmeister: not as far as i know, but subtracting tick mod tickSpacing should probably do the trick

src/hooks/useAllV3Ticks.ts
NoahZinsmeister: ah yeah, `Pool.getAddress` https://github.com/Uniswap/uniswap-v3-sdk/blob/main/src/entities/pool.ts

src/hooks/usePoolTickData.ts
NoahZinsmeister: yeah, let's try :p

src/components/LiquidityDepth/DensityChart.tsx
justindomingue: price is actually a string here.. I think I agree that I shouldn't pass as a string but pass the Price<> and let the component format.

src/components/LiquidityDepth/DensityChart.tsx
```diff
@@ -0,0 +1,204 @@
...
+          containerComponent={
+            <VictoryBrushContainer
+              allowDraw={false}
+              allowDrag={interactive}
+              allowResize={interactive}
+              brushDimension="x"
+              brushDomain={
+                leftPrice && rightPrice
+                  ? {
+                      x: [parseFloat(leftPrice?.toSignificant(5)), parseFloat(rightPrice?.toSignificant(5))],
```
ianlapham, line 116: Will this cause issues with really weird pairs? like a BTC/X pair with >5 decimals?

src/components/LiquidityDepth/hooks.ts
```diff
@@ -0,0 +1,91 @@
...
+  const { loading, error, activeTick, tickData } = usePoolTickData(currencyA, currencyB, feeAmount)
+
+  // clear data when inputs are cleared
+  useEffect(() => {
+    if (!currencyA || !currencyB || !feeAmount) {
+      setFormattedData(undefined)
+    }
+  }, [currencyA, currencyB, feeAmount])
+
+  useEffect(() => {
```
ianlapham, line 41: Could we just memoize all this instead of setting it in state and clearing it on changes

src/components/LiquidityDepth/DensityChart.tsx
justindomingue: What's your suggestion here? Should I go to 8?

src/state/mint/v3/utils.ts
```diff
@@ -20,8 +20,13 @@ export function tryParseTick(
     return undefined
   }
 
+  // handle very small amounts manually
+  const isMinPrice = parseFloat(parseFloat(value).toFixed(quoteToken.decimals)) === 0
```
justindomingue, line 24: remove?

PR #1839, opened 2021-06-10
fix: props for ResponsiveButtonPrimary

YutaSugimura:
Fixed an error that was occurring in the development environment.

スクリーンショット 2021-06-10 20 56 14

### DIFF
--- a/src/pages/Pool/v2.tsx
+++ b/src/pages/Pool/v2.tsx
@@ -54,6 +54,7 @@ const ButtonRow = styled(RowFixed)`
 
 const ResponsiveButtonPrimary = styled(ButtonPrimary)`
   width: fit-content;
+  border-radius: 12px;
   ${({ theme }) => theme.mediaWidth.upToSmall`
     width: 48%;
   `};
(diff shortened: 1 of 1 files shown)


src/pages/Pool/v2.tsx
```diff
@@ -178,9 +178,11 @@ export default function Pool() {
                 <ResponsiveButtonPrimary
                   id="find-pool-button"
                   as={Link}
-                  padding="6px 8px"
-                  borderRadius="12px"
                   to="/pool/v2/find"
+                  style={{
```
justindomingue, line 182: the idea is to pass those props to `ResponsiveButtonPrimary` which is a `styled(RebassButton)`. I think we should be able to not use `style` here.

https://github.com/Uniswap/uniswap-interface/blob/1556ff26b28b8c2bff2229d9e9b1e901fda13d59/src/components/Button/index.tsx#L16

src/pages/Pool/v2.tsx
```diff
@@ -54,6 +54,7 @@ const ButtonRow = styled(RowFixed)`
 
 const ResponsiveButtonPrimary = styled(ButtonPrimary)`
   width: fit-content;
+  border-radius: 12px;
```
justindomingue, line 57: What about

```
const ResponsiveButtonPrimary = styled(ButtonPrimary).attrs(() => ({ borderRadius: '12px' }))BACKQUOTE
  ${({ theme }) => ...
BACKQUOTE
```

YutaSugimura: In that case, the error was not resolved.

justindomingue: Okay, will close.

PR #1845, opened 2021-06-11
ensures english .po exists on startup

JFrankfurt:


package.json
```diff
@@ -120,7 +120,7 @@
     "i18n:compile": "lingui compile",
     "integration-test": "start-server-and-test 'serve build -l 3000' http://localhost:3000 'cypress run'",
     "postinstall": "yarn compile-contract-types",
-    "start": "yarn compile-contract-types && react-scripts start",
+    "start": "yarn compile-contract-types && yarn i18n:extract && react-scripts start",
```
justindomingue, line 123: summoning @moodysalem 

I also prefer generating on start. It does slow down starting a server so could also be a postinstall or something similar.

JFrankfurt:
Was just surprised when I pulled a fresh version of the repo yesterday and
it didn't work w/ `yarn && yarn start`. Figured you may have done this on
purpose though, so tagged you.

Why is English currently omitted by default?

justindomingue:
> Was just surprised when I pulled a fresh version of the repo yesterday and it didn't work w/ `yarn && yarn start`. Figured you may have done this on purpose though, so tagged you. Why is English currently omitted by default?

It's currently set up so that en-US is generated at release time. It's also automatically uploaded to crowdin by a GitHub workflow. There's no need to have it checked in because crowdin is the source of truth. We added `yarn i18n:extract` to the `README` but I agree it might not be super straightforward.

package.json
moodysalem: if you just do a `touch` does it also work?

Yodag7354 (approved):

Yodag7354 (approved):

PR #1869, opened 2021-06-16
fix: memoize extendedEther to unnecessary rerenders

justindomingue:
`ExtendedEther.onChain` creates a new object from 'ETH' which causes dependencies to re-render

### DIFF
--- a/src/hooks/Tokens.ts
+++ b/src/hooks/Tokens.ts
@@ -179,7 +179,8 @@ export function useCurrency(currencyId: string | undefined): Currency | null | u
   const { chainId } = useActiveWeb3React()
   const isETH = currencyId?.toUpperCase() === 'ETH'
   const token = useToken(isETH ? undefined : currencyId)
+  const extendedEther = useMemo(() => (chainId ? ExtendedEther.onChain(chainId) : undefined), [chainId])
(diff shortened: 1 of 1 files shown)


JFrankfurt (commented):
🚢

src/hooks/Tokens.ts
```diff
@@ -179,7 +179,8 @@ export function useCurrency(currencyId: string | undefined): Currency | null | u
   const { chainId } = useActiveWeb3React()
   const isETH = currencyId?.toUpperCase() === 'ETH'
   const token = useToken(isETH ? undefined : currencyId)
+  const extendedEther = useMemo(() => (chainId ? ExtendedEther.onChain(chainId) : undefined), [chainId])
   const weth = chainId ? WETH9_EXTENDED[chainId] : undefined
   if (weth?.address?.toLowerCase() === currencyId?.toLowerCase()) return weth
-  return isETH ? (chainId ? ExtendedEther.onChain(chainId) : undefined) : token
```
moodysalem, line 184: might be better to memoize in the extended ether class. i actually did that in the sdk-core but didn't do it in the extension, just an oversight https://github.com/Uniswap/uniswap-sdk-core/blob/bf75b5ee01ad7ce14e5ce1fe9551620754ca6182/src/entities/ether.ts#L21-L25

PR #1899, opened 2021-06-22
feat(pools): download tick data from TickLens

justindomingue:
download tick data from TickLens and compute active liquidity


src/hooks/useAllV3Ticks.ts
```diff
@@ -0,0 +1,100 @@
...
+import { useTickLens } from './useContract'
+import { TickData } from 'constants/ticks'
+
+function bitmapIndex(tick: number, tickSpacing: number) {
+  return Math.floor(tick / tickSpacing / 256)
+}
+
+const REFRESH_FREQUENCY = { blocksPerFetch: 24 }
+
+// for now, reconsider using this function, it consumes a lot of data and cpu to fetch all the ticks.
```
moodysalem, line 16: this is why this was removed

justindomingue: Yeah, I have found perf to be acceptable (increased blocks per fetch 24, hogs the cpu for about .5-1s when processing the response) and data usage is around 300kb at 0.05% which seems reasonable given the usefulness of the data provided.

I left the comment just because it is fairly relevant.

We could consider getting the active liquidity data preprocessed by the graph to reduce both data usage and cpu.

moodysalem: we should just use the graph, we already brought it into the app. the app freezing for 1s on a high end laptop when you get to a page is a pretty bad perf degradation

justindomingue: It's closer to 400-500ms on a high-end laptop. There's a spinner while this is happening so user is aware that something is being loaded. 

We only fetch the blocks every 5 minutes or so. I think it should be reasonable to start.

Grabbing data from the graph has some other implications. Optimizing for data availability makes sense to start with I think.

NoahZinsmeister: personally i'm in favor of fetching the data on-chain:

1) the trust assumptions are more minimal
2) the total bandwidth transferred is probably not _that_ different. we're obviously using more "cloud compute" with this approach, but acceptable given the setting
3) there are almost certainly perf gains still to be had in this PR/throughout the data fetching lifecycle, up to and including moving compute into a web worker, which imo is an important thing for us to start thinking about in general, in preparation for e.g. client-side routing

src/hooks/usePoolTickData.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    const activeTickProcessed: TickProcessed = {
+      liquidityActive: JSBI.BigInt(pool[1]?.liquidity ?? 0),
+      tickIdx: activeTick,
+      liquidityNet: JSBI.BigInt(0),
+      price0: tickToPrice(token0, token1, activeTick).toFixed(PRICE_FIXED_DIGITS),
+    }
+
+    // if active tick is initialized, fill liquidity
+    if (tickToInitializedTick[activeTick]) {
```
NoahZinsmeister, line 67: maybe put as a ternary in `liquidityNet`?

src/hooks/usePoolTickData.ts
```diff
@@ -0,0 +1,99 @@
...
+      activeTickProcessed.liquidityNet = JSBI.BigInt(tickToInitializedTick[activeTick].liquidityNet)
+    }
+
+    const subsequentTicks = computeSurroundingTicks(
```
NoahZinsmeister, line 71: do these need to be memoized?

src/utils/computeSurroundingTicks.ts
```diff
@@ -0,0 +1,74 @@
...
+  ascending: boolean
+): TickProcessed[] {
+  let previousTickProcessed: TickProcessed = {
+    ...activeTickProcessed,
+  }
+
+  // Iterate outwards (either up or down depending on direction) from the active tick,
+  // building active liquidity for every tick.
+  let processedTicks: TickProcessed[] = []
+  for (let i = 0; i < Object.keys(tickIdxToInitializedTick).length; i++) {
```
NoahZinsmeister, line 25: hmm am i misunderstanding this? we're iterating through as many ticks are defined in our map, which should be only ticks that have a non-zero liquidityGross (subset of all ticks), but then we're bumping currentTickIdx by tickSpacing every time, which isn't guaranteed to get us all the way to the end unless every possible tick is "initialized"

afaict the right (naive) solution would be to just use a while loop, and wait until we hit the break, but in practice i think we simply want the return value to only contain processed ticks for which there's actually a change in liquidity, and then rely on back-filling logic for gaps, right?

justindomingue: yes, you are right. 

I've implemented your second approach. Instead of relying on a `tick->{tick, liquidityNet}` map and iterating over the tick space by `TICK_SPACING` increments, I do the following

That algorithm should have a similar time complexity (?) but span the entire initialized tick space

- sort the `{tick, liquidityNet}[]` data returned by `TickLens`
- find `pivot` in sorted array where active tick is
- iterate left/right from the pivot to build active liquidity

Anything ticks beyond those should not be initialized and liquidity is 0

src/hooks/useAllV3Ticks.ts
justindomingue: Discussed offline. While the preference is to rely on chain data in general, there are some pain points that are difficult to work around with (the current implementation of) TickLens (e.g. active liquidity from old blocks).

Will move to The Graph.

PR #1912, opened 2021-06-24
feat(L2): add Arbitrum UI

JFrankfurt:
image


src/components/Header/NetworkCard.tsx
```diff
@@ -0,0 +1,245 @@
...
+`
+export default function NetworkCard() {
+  const { chainId } = useActiveWeb3React()
+  const node = React.useRef<HTMLDivElement>(null)
+  const open = useModalOpen(ApplicationModal.ARBITRUM_OPTIONS)
+  const toggle = useToggleModal(ApplicationModal.ARBITRUM_OPTIONS)
+  useOnClickOutside(node, open ? toggle : undefined)
+  const { ethereum } = window
+  const [implements3085, setImplements3085] = React.useState(false)
+  React.useEffect(() => {
```
moodysalem, line 157: why are we doing this on mount?

src/components/Header/NetworkCard.tsx
```diff
@@ -0,0 +1,245 @@
...
+  }
+`
+export default function NetworkCard() {
+  const { chainId } = useActiveWeb3React()
+  const node = React.useRef<HTMLDivElement>(null)
+  const open = useModalOpen(ApplicationModal.ARBITRUM_OPTIONS)
+  const toggle = useToggleModal(ApplicationModal.ARBITRUM_OPTIONS)
+  useOnClickOutside(node, open ? toggle : undefined)
+  const { ethereum } = window
+  const [implements3085, setImplements3085] = React.useState(false)
```
moodysalem, line 156: can we use the imported useState, useCallback, useEffect, etc. instead of calling React...

src/components/Header/NetworkCard.tsx
```diff
@@ -0,0 +1,245 @@
...
+    background-color: ${({ theme }) => theme.bg3};
+  }
+`
+export default function NetworkCard() {
+  const { chainId } = useActiveWeb3React()
+  const node = React.useRef<HTMLDivElement>(null)
+  const open = useModalOpen(ApplicationModal.ARBITRUM_OPTIONS)
+  const toggle = useToggleModal(ApplicationModal.ARBITRUM_OPTIONS)
+  useOnClickOutside(node, open ? toggle : undefined)
+  const { ethereum } = window
```
moodysalem, line 155: can we just use the library from active web3 react? it's very possible that this will be supported in walletconnect, and referencing window in a component is ick

src/components/swap/SwapNetworkAlert.tsx
```diff
@@ -0,0 +1,100 @@
...
+        </Trans>
+      </Header>
+      <Body>
+        <Trans>
+          This is an alpha release of Uniswap on the Arbitrum network. You must bridge L1 assets to the network to swap
+          them.
+        </Trans>
+      </Body>
+      <LinkOutToBridge href="https://bridge.arbitrum.io/" target="_blank" rel="noopener noreferrer">
+        Deposit to Arbitrum
```
moodysalem, line 95: trans

src/hooks/useChainIdBackground.ts
```diff
@@ -0,0 +1,23 @@
+import { SupportedChainId } from 'constants/chains'
+import React from 'react'
+import { useActiveWeb3React } from './web3'
+
+export const useChainIdBackground = () => {
+  const { chainId } = useActiveWeb3React()
+  React.useEffect(() => {
+    const background = document.querySelector('#background-radial-gradient')
```
moodysalem, line 8: you can do this just once outside the hook. also this hurts me

src/hooks/useChainIdBackground.ts
```diff
@@ -0,0 +1,23 @@
+import { SupportedChainId } from 'constants/chains'
+import React from 'react'
+import { useActiveWeb3React } from './web3'
+
+export const useChainIdBackground = () => {
```
moodysalem, line 5: can you just make this a component? this logic does not need to be shared or ever called more than once. like call it BackgroundColorNetworkUpdater

src/hooks/useLogs.ts
```diff
@@ -0,0 +1,23 @@
...
+): LogsResponse<Data, Error> {
+  const { library } = useActiveWeb3React()
+  const response: LogsResponse<Data, Error> = {}
+  if (!library) {
+    return response
+  }
+  if (Array.isArray(filters)) {
+    return response
+  } else {
+    library.getLogs(filters)
```
moodysalem, line 20: i don't understand, is this incomplete/can we just remove?

src/pages/Swap/index.tsx
```diff
@@ -350,6 +353,31 @@ export default function Swap({ history }: RouteComponentProps) {
 
   const priceImpactTooHigh = priceImpactSeverity > 3 && !isExpertMode
 
+  // arbitrum network alert
+  const [arbitrumAlertVisible, setArbitrumAlertVisible] = React.useState(false)
+  React.useEffect(() => {
+    const onArbitrum = chainId === SupportedChainId.ARBITRUM_ONE
+    if (onArbitrum && account && library) {
+      library.getBalance(account).then((balance) => {
```
moodysalem, line 361: please use the balance fetching hook

src/pages/Swap/index.tsx
```diff
@@ -68,7 +69,10 @@ const StyledInfo = styled(Info)`
   }
 `
 
+const ARBITRUM_LOCAL_STORAGE_DISMISSED_KEY = 'HIDE_ARBITRUM_SWAP_ALERT'
```
moodysalem, line 72: please use redux for storing this state, for which the user slice is already stored to localstorage

src/react-app-env.d.ts
```diff
@@ -12,6 +12,11 @@ interface Window {
     on?: (...args: any[]) => void
     removeListener?: (...args: any[]) => void
     autoRefreshOnNetworkChange?: boolean
+    request: (payload: Record<string, any>) => Promise<any>
+    sendAsync: (
+      payload: Record<string, any>,
+      callback?: (error: Error | null, result: JsonRpcResponse | undefined) => void
+    ) => void
```
moodysalem, line 19: these should go away if we use library instead of window.ethereum

src/pages/Swap/index.tsx
```diff
@@ -350,6 +353,31 @@ export default function Swap({ history }: RouteComponentProps) {
 
   const priceImpactTooHigh = priceImpactSeverity > 3 && !isExpertMode
 
+  // arbitrum network alert
+  const [arbitrumAlertVisible, setArbitrumAlertVisible] = React.useState(false)
```
moodysalem, line 357: not sure why you need state for this, just store whether it's dismissed in redux, and you should show it if they're on arbitrum and haven't dismissed it yet

moodysalem (changes requested):

src/components/Header/NetworkCard.tsx
JFrankfurt: We want to determine if the menu option for switching back to L1 is supported and should be visible. Switching to the current network is a noop and the promise returns null if it's supported, and it throws if the api is unavailable. Do you think we should run the check somewhere else? I could see doing a global check somewhere and retaining the support info for a check here.

src/hooks/useLogs.ts
JFrankfurt: Yeah inclusion of this was an oversight. WIP for something else.

src/hooks/useChainIdBackground.ts
JFrankfurt: Yeah, I can see that and I'm fine w/ it in this case since all the other updaters use that API. I just prefer not using components that never return dom since the api surface when used tends to imply document structure; a null returning component never has that. 🤷 Hooks communicate the side effect more clearly, I think.

moodysalem: hooks do not implicitly have side effects as react components do not implicitly return dom, but making it a hook implies it is to be shared

src/components/Header/NetworkCard.tsx
moodysalem: could just check provider.isMetaMask

JFrankfurt: The thought occurred to me, but I figured this gave us better flexibility for supporting other clients that might support it now (w/o me knowing of it) or support it in the future w/o code change. Admittedly, it is super gross and I hate it.

src/hooks/useChainIdBackground.ts
JFrankfurt: What would you think of adding this to `src/state/application/updater.ts`?

src/pages/Swap/index.tsx
JFrankfurt: Cool, didn't know this was a thing

src/hooks/useChainIdBackground.ts
JFrankfurt: Seems like the existing updaters are all state-focused, and this one's kind of an oddball. Not sure where it belongs.

src/hooks/useChainIdBackground.ts
```diff
@@ -1,14 +1,15 @@
...
-import React from 'react'
+import { useEffect } from 'react'
 import { useActiveWeb3React } from './web3'
 
 export const useChainIdBackground = () => {
   const { chainId } = useActiveWeb3React()
-  React.useEffect(() => {
-    const background = document.querySelector('#background-radial-gradient')
-    if (background === null) {
+  const background = document.querySelector('#background-radial-gradient') as HTMLDivElement | null
```
moodysalem, line 7: sorry, nit, `document.getElementById('background-radial-gradient')` is faster (not that it matters at all)

src/hooks/useChainIdBackground.ts
moodysalem: don't feel strongly about where it goes

src/components/Header/NetworkCard.tsx
moodysalem: sure, but sending this request while no op on metamask may produce a strange result in another wallet

moodysalem (approved):

src/hooks/useChainIdBackground.ts
subtleGradient: `getElementById` is not faster than `querySelector` unless maybe you're executing it thousands of times in a loop in IE7 in 2007. Then… maybe 😜

PR #1926, opened 2021-06-28
feat(thegraph): auto-generate graphql types

justindomingue:
Generates GraphQL types from the schema
- also generates types for the RTK-Query API

I also simplified the base graphql query


src/state/data/slice.ts
```diff
@@ -1,21 +1,22 @@
...
+import { gql, GraphQLClient } from 'graphql-request'
 import { FeeAmount } from '@uniswap/v3-sdk'
 import { reduce } from 'lodash'
+import { graphqlRequestBaseQuery } from '@rtk-query/graphql-request-base-query'
 
-import { graphqlBaseQuery, UNISWAP_V3_GRAPH_URL } from './common'
 import { FeeTierDistribution, PoolTVL } from './types'
 
-export const dataApi = createApi({
+export const UNISWAP_V3_GRAPH_URL = 'https://api.thegraph.com/subgraphs/name/ianlapham/uniswap-v3-alt'
```
ianlapham, line 9: one note is we should be using this endpoint 

https://thegraph.com/explorer/subgraph/uniswap/uniswap-v3

ianlapham (approved):
LGTM

src/state/data/slice.ts
justindomingue: ah, thanks!

PR #1941, opened 2021-06-30
refactor: logs hook

moodysalem:
fixes https://github.com/Uniswap/uniswap-interface/issues/1832


src/state/logs/hooks.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  useEffect(() => {
+    // we do not have all the information necessary to fetch, skip
+    if (!filter || !library || typeof blockNumber !== 'number' || typeof chainId !== 'number') return
+
+    // we are already synced, skip
+    if (state && state.chainId === chainId && state.blockNumber >= blockNumber) return
+
+    // TODO: we could try to fetch only as of the last block we've seen, however there aren't good ways to force a fetch as of a block
+    const { promise, cancel } = retry(() => library.getLogs({ ...filter, fromBlock: 0, toBlock: blockNumber }), {
```
moodysalem, line 49: todo: retry isn't actually used here, since we never throw retryable error

PR #1943, opened 2021-07-01
fix: error messages from walletconnect provider

moodysalem:
https://github.com/Uniswap/uniswap-interface/issues/1560


moodysalem:
Thank you @pedrouid it looks like this helps!

PR #1949, opened 2021-07-02
feat(iframe): iframe ui updates

ianlapham:
- update for hiding components when app rendered witin iframe
- example here: https://jsfiddle.net/ianlapham/hme9ucr7/12/


src/hooks/useIsIframe.ts
```diff
@@ -0,0 +1,3 @@
+export function useIsIframe() {
+  return window !== window.parent
```
moodysalem, line 2: you don't really need a hook, you can just `export const IS_WITHIN_IFRAME = window && window !== window.parent`

src/components/Header/index.tsx
```diff
@@ -302,44 +303,51 @@ export default function Header() {
...
+      {!isIframe && (
+        <HeaderRow>
+          <Title href=".">
+            <UniIcon>
+              <img width={'24px'} src={darkMode ? LogoDark : Logo} alt="logo" />
+            </UniIcon>
+          </Title>
+        </HeaderRow>
+      )}
+      {!isIframe && (
```
justindomingue, line 324: do you need this?

ianlapham: I was intending to hide everything in the top nav - so logo and nav buttons

src/components/Header/index.tsx
```diff
@@ -308,38 +309,42 @@ export default function Header() {
...
-        </StyledNavLink>
-        <StyledNavLink id={`stake-nav-link`} to={'/vote'}>
-          <Trans>Vote</Trans>
-        </StyledNavLink>
-        <StyledExternalLink id={`stake-nav-link`} href={'https://info.uniswap.org'}>
-          <Trans>Charts</Trans>
-          <sup>↗</sup>
-        </StyledExternalLink>
-      </HeaderLinks>
+      {!IS_WITHIN_IFRAME && (
```
moodysalem, line 312: two adjacent identical checks could just be

`IS_WITHIN_IFRAME ? null : (
<>
{ stuff here }
</>
)
`

src/components/SearchModal/ImportToken.tsx
```diff
@@ -97,7 +99,9 @@ export function ImportToken({ tokens, list, onBack, onDismiss, handleCurrencySel
                 </AutoColumn>
                 {chainId && (
                   <ExternalLink href={getExplorerLink(chainId, token.address, ExplorerDataType.ADDRESS)}>
-                    <AddressText fontSize={12}>{token.address}</AddressText>
+                    <AddressText fontSize={12}>
+                      {IS_WITHIN_IFRAME ? shortenAddress(token.address) : token.address}
```
moodysalem, line 103: can you add word-break styling to the `AddressText` just in case? could also happen on small devices

src/pages/styled.tsx
```diff
@@ -4,3 +4,12 @@ export const StandardPageWrapper = styled.div`
   padding-top: 160px;
   width: 100%;
 `
+
+export const IframeBodyWrapper = styled.div`
```
moodysalem, line 8: think weneed some padding

moodysalem:
note that these two changes (domain detection and iframe detection) will make our integration tests less reliable for catching issues. we should either a) increase the scope of our integration tests to include both app.uniswap.org and localhost:x + iframe/non-iframe, or b) only test one domain and try to make the iframe stuff work responsively for all devices

jajones10s:
Yo we have a direct use case for this fix, I was wondering if there is any update on when this could be pushed.

it seems that this could change the game for us! @moodysalem

PR #1950, opened 2021-07-02
fix: add optional chaining for stakingInfo

lukedonato:
Fixes https://github.com/Uniswap/uniswap-interface/issues/1927


moodysalem:
we should probably show a message indicating the url is invalid rather than just a blank page

PR #1957, opened 2021-07-02
feat(L2): network alert cards (optimism and arbitrum)

JFrankfurt:


src/pages/App.tsx
```diff
@@ -69,6 +71,46 @@ function TopLevelModals() {
...
+const Routes = () => {
+  const { chainId } = useActiveWeb3React()
+
+  const HIDE_ON_L2 = Boolean(chainId && L2_CHAIN_IDS.includes(chainId))
+  return (
+    <Switch>
+      <Route exact strict path="/vote" component={Vote} />
+      <Route exact strict path="/vote/:governorIndex/:id" component={VotePage} />
+      <Route exact strict path="/claim" component={OpenClaimAddressModalAndRedirectToSwap} />
+      {HIDE_ON_L2 && <Route exact strict path="/uni" component={Earn} />}
```
moodysalem, line 83: this should be addressed in the page, rather than redirect to swap (and our catch-all route shouldn't direct either). i.e. it should tell you to connect to mainnet if you are at the route not supported on the current chain

PR #1968, opened 2021-07-06
fix(styled-components): no more `borderRadius` errors!

NoahZinsmeister:
bumps to styled-components v5 ([fully non-breaking](https://styled-components.com/docs/faqs#what-do-i-need-to-do-to-migrate-to-v5)), and uses new [transient props](https://styled-components.com/docs/api#transient-props) to not pass down invalid dom attributes


src/components/Button/index.tsx
```diff
@@ -12,16 +12,15 @@ const Base = styled(RebassButton)<
   {
     padding?: string
     width?: string
-    borderRadius?: string
+    $borderRadius?: string
```
justindomingue, line 15: does `$` just change the property name so there's no conflict?

justindomingue (approved):

src/components/Button/index.tsx
NoahZinsmeister: no specifically it tells styled-components to not set the prop as a DOM property

PR #1988, opened 2021-07-07
fix(L2): polish

JFrankfurt:


src/utils/getInfoLinkByChainId.ts
```diff
@@ -0,0 +1,20 @@
+import { SupportedChainId } from 'constants/chains'
+
+export default function getInfoLinkByChainId(chainId?: number) {
```
justindomingue, line 3: have you thought about making this a const object instead? I'd love to extend this actually with the subgraph url too. Would be easier to extend if it was an object.

src/components/OptimismDowntimeWarning/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+export default function OptimismDowntimeWarning() {
+  const { chainId } = useActiveWeb3React()
+  if (!chainId || ![SupportedChainId.OPTIMISM, SupportedChainId.OPTIMISTIC_KOVAN].includes(chainId)) {
+    return null
+  }
+  return (
+    <Root>
+      <TitleRow>
+        <WarningIcon />
+        Optimism <Trans>Scheduled Downtimes</Trans>
```
justindomingue, line 48: Might be better to put 'Optimism' inside the `<Trans>` (thanks for remembering to wrap the new strings inside `<Trans>` btw). Some languages have word different word ordering

src/components/OptimismDowntimeWarning/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+  align-items: center;
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-start;
+  margin: 0;
+  font-size: 20px;
+  font-weight: 600;
+  line-height: 25px;
+`
+const Body = styled.div`
```
justindomingue, line 29: have you considered using `TYPE.body`? which may give you some default styling?

src/components/Header/NetworkCard.tsx
```diff
@@ -183,24 +188,37 @@ export default function NetworkCard() {
 
   if (L2_CHAIN_IDS.includes(chainId)) {
     const info = L2_INFO[chainId]
+    let bridgeText, explorerText
+    switch (chainId) {
+      case SupportedChainId.ARBITRUM_ONE:
+      case SupportedChainId.ARBITRUM_RINKEBY:
+        bridgeText = `${NETWORK_LABELS[chainId]} Bridge`
```
justindomingue, line 195: can you wrap those with `<Trans>` unless it's the product name? Can also use `t` macro if you need a string (although be aware that there's a known bug where `t` doesn't update on locale change)

src/utils/getInfoLinkByChainId.ts
JFrankfurt: Yeah I could definitely see putting this in the chain constants file as an object

src/components/OptimismDowntimeWarning/index.tsx
JFrankfurt: Oh interesting point... Do you know how our translation setup would work with a word like Optimism that's being used as a proper noun here?

justindomingue: I think the easiest way to handle this would be to put optimism in a variable

`const name = 'Optimism'`

`<Trans>{name} Scheduled Downtimes`

such that the translation key is something like `{0} Scheduled Downtimes` and translator don't attempt to translate the word `Optimism`.

moodysalem:
bit weird to be triggering minor releases daily via `feat:` tag with no new features, we probably want to label these `fix:` and consider them visual bugs

src/components/OptimismDowntimeWarning/index.tsx
JFrankfurt: that makes sense
also, yuck

JFrankfurt:
Oh I didn't realize the release behavior distinction, thanks for the heads up @moodysalem

src/components/OptimismDowntimeWarning/index.tsx
JFrankfurt: I like this in principle, but it looks like the class specificity works such that I'd have to fill the style override w/ `!important`s to get it to work... :/

i.e. 
```
const Body = styled(TYPE.body)`
  font-size: 12px !important;
  line-height: 15px;
  margin: 8px 0 0 0 !important;
`
```

Is that the intended behavior of the `TYPE[text]` components? I didn't expect it to require that.

src/state/lists/updater.ts
```diff
@@ -28,6 +29,12 @@ export default function Updater(): null {
     )
   }, [fetchList, isWindowVisible, lists])
 
+  useEffect(() => {
+    if (chainId && [SupportedChainId.OPTIMISM, SupportedChainId.OPTIMISTIC_KOVAN].includes(chainId)) {
+      dispatch(enableList(OPTIMISM_LIST))
```
moodysalem, line 34: this isn't exactly what was discussed, which was to add to the default list of lists and enhance the UI to indicate which lists have no tokens on the current chain. did those requirements change?

src/state/user/updater.tsx
```diff
@@ -1,10 +1,24 @@
...
   const dispatch = useAppDispatch()
+  const { chainId } = useActiveWeb3React()
+  const [, setTTL] = useUserTransactionTTL()
+
+  // keep deadline appropriate to chainId
+  useEffect(() => {
+    if (!chainId) {
+      return
+    }
+    setTTL(L2_CHAIN_IDS.includes(chainId) ? L2_DEADLINE_FROM_NOW : DEFAULT_DEADLINE_FROM_NOW)
```
moodysalem, line 20: should not use useEffect like this. the result is if the user switches back to the main chain they lost their settings. instead we should just update useTransactionDeadline to return a diff deadline and the deadline input on optimism

src/components/OptimismDowntimeWarning/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const { chainId } = useActiveWeb3React()
+  if (!chainId || ![SupportedChainId.OPTIMISM, SupportedChainId.OPTIMISTIC_KOVAN].includes(chainId)) {
+    return null
+  }
+  const name = 'Optimism'
+  return (
+    <Root>
+      <TitleRow>
+        <WarningIcon />
+        <Trans>{name} Scheduled Downtimes</Trans>
```
moodysalem, line 48: you can just inline `{'Optimism'}` so it's clear that it's hard coded

src/constants/lists.ts
```diff
@@ -1,33 +1,34 @@
...
   AAVE_LIST,
-  UMA_LIST,
-  WRAPPED_LIST,
-  SET_LIST,
-  ROLL_LIST,
-  COINGECKO_LIST,
   CMC_ALL_LIST,
   CMC_STABLECOIN,
-  KLEROS_LIST,
+  COINGECKO_LIST,
```
moodysalem, line 23: this reordering is a reprioritization, is it intentional?

src/constants/misc.ts
```diff
@@ -7,6 +7,7 @@ export const NetworkContextName = 'NETWORK'
 
 // 30 minutes, denominated in seconds
 export const DEFAULT_DEADLINE_FROM_NOW = 60 * 30
```
moodysalem, line 9: we can change these to use the ms macro, eg. `ms\`30m\``

src/pages/Pool/CTACards.tsx
```diff
@@ -137,7 +144,7 @@ export default function CTACards() {
             </RowBetween>
           </ResponsiveColumn>
         </CTA1>
-        <CTA2 href={'https://info.uniswap.org/#/pools'}>
+        <CTA2 href={infoLink + '/pools'}>
```
moodysalem, line 147: i don't love this method of constructing info urls because it implies you know what infoLink ends in when it's in another file, plus this is copy pasted 4 times that i count now (and the copy pasted code isn't particularly readable) so i would prefer to move this into a function

src/components/OptimismDowntimeWarning/index.tsx
JFrankfurt: @justindomingue that works w/ your suggestion, right?

src/pages/Pool/CTACards.tsx
JFrankfurt: Take a look at the new implementation and lmk what you think. It still uses the `infoLink + '/pools'` syntax in the templates, but it removes the tedious L2 checks from all the components.

src/components/OptimismDowntimeWarning/index.tsx
justindomingue: It's neat. You should double check that it does extract the string properly with `yarn i18n:extract`. Love it if it works

src/state/lists/hooks.ts
```diff
@@ -47,13 +48,16 @@ export function useAllLists(): AppState['lists']['byUrl'] {
 }
 
 function combineMaps(map1: TokenAddressMap, map2: TokenAddressMap): TokenAddressMap {
```
moodysalem, line 50: this code is a little funny, we don't really need a key for each chain, just the chains present in either map1 or map2. would be a good candidate for adding unit tests

src/state/lists/updater.ts
moodysalem: if you remove this part you can address the token lists ux changes in a separate pr so this very broad pr isn't outstanding for too long (and just add optimism list to default list of lists)

src/constants/chains.ts
```diff
@@ -1,43 +1,118 @@
...
   ROPSTEN = 3,
   RINKEBY = 4,
   GOERLI = 5,
   KOVAN = 42,
-  ARBITRUM_ONE = 42161,
-  ARBITRUM_RINKEBY = 421611,
-  OPTIMISM = 10,
-  OPTIMISTIC_KOVAN = 69,
+}
+export const SupportedChainId = {
```
moodysalem, line 18: i don't love turning this into a const instead of an enum, since i don't think it works exactly the same as the old enum

can we achieve the same thing we want by having 

```
enum SupportedChainId { // all chains }

export type SupportedL2ChainId = SupportedChainId.ARBITRUM_ONE | SupportedChainId.OPTIMISM | ...

export SUPPORTED_CHAIN_IDS = [
 SupportedChainId.MAINNET,
 // all chains
] as const
```

moodysalem: or just help me understand why this is better

moodysalem (approved):
just the one nit about splitting chain id

src/constants/chains.ts
JFrankfurt: My approach isn't better, it just felt weird repeating the enum definition and creating two sources of truth for chainId constants. This way we lost the ability to use `SupportedChainId` as a type, but retained the ability to use a single reference point for all chainId values (`SupportedChainId.MAINNET`).

src/constants/chains.ts
```diff
@@ -1,42 +1,49 @@
...
+  [SupportedChainId.ROPSTEN]: 'Ropsten',
+  [SupportedChainId.GOERLI]: 'Görli',
+  [SupportedChainId.KOVAN]: 'Kovan',
+  [SupportedChainId.ARBITRUM_ONE]: 'Arbitrum',
+  [SupportedChainId.ARBITRUM_RINKEBY]: 'Arbitrum Rinkeby',
+  [SupportedChainId.OPTIMISM]: 'Optimistic Ethereum',
+  [SupportedChainId.OPTIMISTIC_KOVAN]: 'Optimistic Kovan',
 }
 
 export const L2_CHAIN_IDS = [
```
moodysalem, line 42: if you add `as const` maybe it'll help with type safety? not sure

src/utils/constructSameAddressMap.ts
```diff
@@ -1,16 +1,16 @@
-import { SupportedL1ChainId, SupportedL2ChainId } from '../constants/chains'
+import { SupportedChainId } from '../constants/chains'
 
-const MAINNET_AND_TESTNETS: (SupportedL1ChainId | SupportedL2ChainId)[] = [
-  SupportedL1ChainId.MAINNET,
-  SupportedL1ChainId.ROPSTEN,
-  SupportedL1ChainId.RINKEBY,
-  SupportedL1ChainId.GOERLI,
-  SupportedL1ChainId.KOVAN,
+const MAINNET_AND_TESTNETS: SupportedChainId[] = [
```
moodysalem, line 3: could move this to chains.ts

moodysalem (approved):

src/components/OptimismDowntimeWarning/index.tsx
justindomingue: hmm I think you are supposed to pass in font size and margin as component props

`<TYPE.body fontSize={12} margin='8px 0 0 0'>` or something similar. `line-height` can probably remain in the styled component.

src/utils/switchToNetwork.ts
```diff
@@ -1,10 +1,10 @@
-import { SupportedChainId } from 'constants/chains'
+import { SupportedL1ChainId, SupportedL2ChainId } from 'constants/chains'
 import { BigNumber, utils } from 'ethers'
 import { Web3Provider } from '@ethersproject/providers'
 
 interface SwitchNetworkArguments {
   library: Web3Provider
-  chainId: SupportedChainId
+  chainId: SupportedL1ChainId | SupportedL2ChainId
```
moodysalem, line 7: this can go back to SupportedChainId

PR #2014, opened 2021-07-11
fix(L2): L2 token lists

JFrankfurt:
- [x] sort lists with tokens on the current chain above those without


src/components/SearchModal/ManageLists.tsx
```diff
@@ -160,8 +169,19 @@ const ListRow = memo(function ListRow({ listUrl }: { listUrl: string }) {
 
   if (!list) return null
 
+  const { label } = CHAIN_INFO[chainId ? chainId : SupportedL1ChainId.MAINNET]
+
+  const hoverText = t`This token list currently has ${activeTokensOnThisChain} ${label} tokens in it`
```
justindomingue, line 174: You could use a plural form here. From the [docs](https://lingui.js.org/#quick-overview):

```
<Plural  value={activeTokensOnThisChain}
        one={
            This token list currently has one ${label} tokens in it
        }
        other={`This token list currently has ${activeTokensOnThisChain} ${label} tokens in it`}
```

src/constants/chains.ts
```diff
@@ -1,43 +1,118 @@
...
   ROPSTEN = 3,
   RINKEBY = 4,
   GOERLI = 5,
   KOVAN = 42,
-  ARBITRUM_ONE = 42161,
-  ARBITRUM_RINKEBY = 421611,
-  OPTIMISM = 10,
-  OPTIMISTIC_KOVAN = 69,
+}
+export const SupportedChainId = {
```
justindomingue, line 18: couldn't this just be a type? `export type SupportChainId = SupportedL2ChainId | SupportedL1ChainId` or the ts equivalent

src/components/SearchModal/ManageLists.tsx
```diff
@@ -226,12 +246,24 @@ export function ManageLists({
...
 }) {
+  const { chainId } = useActiveWeb3React()
   const theme = useTheme()
 
   const [listUrlInput, setListUrlInput] = useState<string>('')
 
   const lists = useAllLists()
 
+  const tokenCountByListName = useMemo(() => {
+    const result: any = {}
```
moodysalem, line 257: don't use any

src/components/SearchModal/ManageLists.tsx
```diff
@@ -226,12 +246,24 @@ export function ManageLists({
...
+    Object.values(lists).forEach(({ current: list }) => {
+      if (!list) {
+        return
+      }
+      result[list.name] = list.tokens.filter((token) => token.chainId === chainId).length
```
moodysalem, line 262: this counting is expensive, add a method that uses a `WeakMap` so we avoid ever repeating if the `WeakMap` is available

src/state/lists/hooks.ts
```diff
@@ -47,13 +48,16 @@ export function useAllLists(): AppState['lists']['byUrl'] {
...
 
 function combineMaps(map1: TokenAddressMap, map2: TokenAddressMap): TokenAddressMap {
-  return {
-    [1]: { ...map1[1], ...map2[1] },
-    [4]: { ...map1[4], ...map2[4] },
-    [3]: { ...map1[3], ...map2[3] },
-    [42]: { ...map1[42], ...map2[42] },
-    [5]: { ...map1[5], ...map2[5] },
-  }
+  return Object.values(SupportedChainId).reduce(
```
moodysalem, line 51: can we make the improvement of returning a map that only contains chains from map1 or map2 and add unit tests? also there's potential perf savings here of not using the spread operator when an object only exists on map1 or map2

src/state/lists/updater.ts
```diff
@@ -28,6 +28,11 @@ export default function Updater(): null {
     )
   }, [fetchList, isWindowVisible, lists])
 
+  useEffect(() => {
+    if (chainId && [SupportedChainId.OPTIMISM, SupportedChainId.OPTIMISTIC_KOVAN].includes(chainId)) {
+      dispatch(enableList(OPTIMISM_LIST))
+    }
+  }, [chainId, dispatch])
```
moodysalem, line 35: just add it to the default list of lists

src/utils/switchToNetwork.ts
```diff
@@ -1,10 +1,10 @@
-import { SupportedChainId } from 'constants/chains'
+import { SupportedL1ChainId, SupportedL2ChainId } from 'constants/chains'
 import { BigNumber, utils } from 'ethers'
 import { Web3Provider } from '@ethersproject/providers'
 
 interface SwitchNetworkArguments {
   library: Web3Provider
-  chainId: SupportedChainId
+  chainId: SupportedL1ChainId | SupportedL2ChainId
```
moodysalem, line 7: this can just be `SupportedChainId` still

src/state/lists/hooks.ts
justindomingue: you might be able to simplify the logic with [_merge](https://lodash.com/docs/4.17.15#merge)

src/state/lists/updater.ts
moodysalem: hm i see this is for enabling the list rather than adding it to the list of lists

src/components/SearchModal/ManageLists.tsx
JFrankfurt: v cool

JFrankfurt: Handled plurals, but didn't end up using the Plural component because I need a string for the title property.

src/components/SearchModal/ManageLists.tsx
```diff
@@ -171,7 +171,7 @@ const ListRow = memo(function ListRow({ listUrl }: { listUrl: string }) {
 
   const { label } = CHAIN_INFO[chainId ? chainId : SupportedChainId.MAINNET]
   const plural = activeTokensOnThisChain === 1 ? '' : 's'
-  const hoverText = t`This token list currently has ${activeTokensOnThisChain} ${label} token${plural}`
+  const hoverText = t`This token list has ${activeTokensOnThisChain} ${label} token${plural}`
```
justindomingue, line 174: this is not really going to work for non english unfortunately :(

That's because the string that will be extracted by the i18n library will look like `This token list has {0} {1} token{2}`. At run time we will try to add an 's' at position `{2}`, but languages have different ways of localizing plurals.

I think I'd rather use what you had before for this PR, not a big deal.

moodysalem:
closing in favor of pr from branch in this repo

PR #2022, opened 2021-07-12
fix: dismiss txn confirmation modal if on L2

ianlapham:
- auto dismiss txn confirmation modal if on L2 and txn is confirmed on chain and modal still open 
- update styling on network selector (padding, popup padding)

### DIFF
--- a/src/components/Popups/index.tsx
+++ b/src/components/Popups/index.tsx
@@ -4,6 +4,9 @@ import { AutoColumn } from '../Column'
 import PopupItem from './PopupItem'
 import ClaimPopup from './ClaimPopup'
 import { useURLWarningVisible } from '../../state/user/hooks'
+import { useActiveWeb3React } from 'hooks/web3'
+import { SupportedChainId } from 'constants/chains'
+import { MEDIA_WIDTHS } from 'theme'
 
 const MobilePopupWrapper = styled.div<{ height: string | number }>`
   position: relative;
@@ -30,7 +33,11 @@ const MobilePopupInner = styled.div`
   }
 `
 
-const FixedPopupColumn = styled(AutoColumn)<{ extraPadding: boolean }>`
+const StopOverflowQuery = `@media screen and (min-width: ${MEDIA_WIDTHS.upToMedium + 1}px) and (max-width: ${
+  MEDIA_WIDTHS.upToMedium + 500
+}px)`
+
+const FixedPopupColumn = styled(AutoColumn)<{ extraPadding: boolean; xlPadding: boolean }>`
   position: fixed;
   top: ${({ extraPadding }) => (extraPadding ? '80px' : '88px')};
   right: 1rem;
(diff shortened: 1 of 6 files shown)


src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -227,8 +229,15 @@ export default function TransactionConfirmationModal({
 }: ConfirmationModalProps) {
   const { chainId } = useActiveWeb3React()
 
+  const confirmed = useIsTransactionConfirmed(hash)
+
   if (!chainId) return null
 
+  // if on L2 and txn is confirmed, close automatically (if open)
+  if (isOpen && confirmed && L2_CHAIN_IDS.includes(chainId)) {
+    onDismiss()
```
moodysalem, line 238: this should be wrapped in useEffect

src/components/Popups/index.tsx
```diff
@@ -30,7 +33,11 @@ const MobilePopupInner = styled.div`
   }
 `
 
-const FixedPopupColumn = styled(AutoColumn)<{ extraPadding: boolean }>`
+const StopOverflowQuery = `@media screen and (min-width: ${MEDIA_WIDTHS.upToMedium + 1}px) and (max-width: ${
+  MEDIA_WIDTHS.upToMedium + 500
```
moodysalem, line 37: wait... why are we doing this

ianlapham: The network selector overflows into the next line - so the popups get cut off. Another solution is just bumping up z-index on popups to cover network selector -maybe thats better?

ianlapham: (they only get cut off within that screen size)

moodysalem: this smells bad. we're trying to get all the UI elements in just the right spot at each screen size. overlapping feels bad in general, but popups should definitely overlap the network selector. however it might not be z-index but rather the layering, which could make this really time consuming to fix. 

i'm ok with this change for now, but we need to address all our media queries and try to make things more responsive by using the browser's rendering engine instead of media queries and very specific margins

src/components/Popups/index.tsx
```diff
@@ -49,9 +60,13 @@ export default function Popups() {
 
   const urlWarningActive = useURLWarningVisible()
 
+  // need extra padding if network is not L1 Ethereum
+  const { chainId } = useActiveWeb3React()
+  const networkNameShown = Boolean(chainId && chainId !== SupportedChainId.MAINNET)
```
moodysalem, line 65: can you give this a better name, e.g. `isNotOnMainnet`

src/components/Popups/index.tsx
ianlapham: Got it - okay lets track that, ill try with z-index for this first in case that works

ianlapham: yeah it seems to be layering

PR #2031, opened 2021-07-13
fix(L2): add Optimism announcement on mainnet

JFrankfurt:
also includes a fix for dismissal of the network alert (dismissing on Arbitrum no longer dismisses on Optimism as well)

image


moodysalem:
blocked on announcement

PR #2055, opened 2021-07-14
feat: use modal for txn confirmation on l2

ianlapham:
- keep modal open but add confirmation state if txn confirmed on L2


moodysalem:
atm shows error on kovan, also the popup still showing up in the background, not sure what we want to do about that. if this is the ideal flow, maybe no popup after close?

image

Dunlap28 (approved):

PR #2062, opened 2021-07-15
feat: Add language flags

callil:


justindomingue:
i like the way flags look and it makes the list easier to browse. However, we should be aware that [flags do not represent languages](http://www.flagsarenotlanguages.com/blog/why-flags-do-not-represent-language/)

justindomingue (approved):

moodysalem:
> i like the way flags look and it makes the list easier to browse. However, we should be aware that [flags do not represent languages](http://www.flagsarenotlanguages.com/blog/why-flags-do-not-represent-language/)

i would agree it is eye catching but not really worth

also looks less nice here
image

zzmp:
Given that we're global, and not targeted to specific countries, I also don't think it's worth merging this.

PR #2063, opened 2021-07-15
fix: set language using url param in language dropdown

justindomingue:


src/components/Menu/index.tsx
```diff
@@ -163,6 +181,39 @@ const ToggleMenuItem = styled.button`
...
+  const location = useLocation()
+  const qs = useParsedQueryString()
+
+  return (
+    <MenuFlyout>
+      <ToggleMenuItem onClick={close}>
+        <ChevronLeft size={16} />
+      </ToggleMenuItem>
+      {SUPPORTED_LOCALES.map((locale) => (
+        <InternalLinkMenuItem
```
moodysalem, line 195: you could put this into a shared component like `SwitchLocaleLink<{ locale: SupportedLocale }>`

moodysalem: or write a hook `useLocationLinkProps(locale: SupportedLocale): { to: Location, onClick: () => void }`

justindomingue: refactored. Did I overcomplicate the hooks?

src/hooks/useLocationLinkProps.ts
```diff
@@ -0,0 +1,32 @@
...
+  to?: LocationDescriptor
+  onClick?: () => void
+} {
+  const location = useLocation()
+  const qs = useParsedQueryString()
+  const activeLocale = useActiveLocale()
+
+  if (!locale) return {}
+
+  return {
```
moodysalem, line 19: memo this whole thing

src/components/SwitchLocaleLink/index.tsx
```diff
@@ -17,46 +14,37 @@ const Container = styled(TYPE.small)`
...
+
+  const { to, onClick } = useLocationLinkProps(targetLocale)
+
+  if (!targetLocale || !to) return null
+
+  return (
+    <Container>
+      <Trans>
+        Uniswap available in:{' '}
+        {
```
moodysalem, line 42: unnecessary bracket

src/components/SwitchLocaleLink/index.tsx
```diff
@@ -17,46 +14,37 @@ const Container = styled(TYPE.small)`
   margin-top: 1rem !important;
 `
 
-export function SwitchLocaleLink() {
-  const activeLocale = useActiveLocale()
+const useTargetLocale = (activeLocale: SupportedLocale) => {
```
moodysalem, line 17: if this werent a hook then target locale would never be null i think and then your hook can always return a value w/ to and onClick

justindomingue: `targetLocale` is actually conditionally set based on active locale and default locale so I think `null` is a valid state here.

moodysalem (approved):

PR #2078, opened 2021-07-19
fix: invert selected range on RateToggle instead of clearing inputs

justindomingue:


src/pages/AddLiquidity/index.tsx
```diff
@@ -609,8 +609,8 @@ export default function AddLiquidity({
                     currencyA={baseCurrency}
                     currencyB={quoteCurrency}
                     handleRateToggle={() => {
-                      onLeftRangeInput('')
-                      onRightRangeInput('')
+                      onLeftRangeInput((invertPrice ? priceLower : priceUpper?.invert())?.toSignificant(6) ?? '')
+                      onRightRangeInput((invertPrice ? priceUpper : priceLower?.invert())?.toSignificant(6) ?? '')
```
ianlapham, line 613: How come we do a significance on the inverted version only? or does the inversion output a string

ianlapham (approved):

Issue #2086, opened 2021-07-20
POC for using trained ONNX model in the interface


NoahZinsmeister:
I would like to be able to consume [ONNX](https://onnx.ai/) models (via [onnxruntime-web](https://www.onnxruntime.ai/) or similar) in order to make predictions client-side. Specifically, models trained with [scikit-learn](https://scikit-learn.org/stable/) and exported via [sklearn-onnx](https://onnx.ai/sklearn-onnx/). Work has been started on this in https://github.com/Uniswap/uniswap-interface/tree/feat/onnx, which works, but I have so far been unable to extend this code to handle models trained and exported from python.

A successful resolution of this issue would include replicable python3 code that exports a trained model from sklearn which is then consumed by the interface and produces predictions.

0xMartian:
Using the basic example from [here](https://onnx.ai/sklearn-onnx/) I trained a model: https://deepnote.com/project/ONNX-p3P2bx05T_GzH_MMjDTtmg/%2Fnotebook.ipynb

Tried implementing it similarly using `onnxruntime-web` but while running the model it throws an error and a random number.

<img width="666" alt="Screen Shot 2021-07-21 at 01 38 34" src="https://user-images.githubusercontent.com/87619808/126408432-bb3f51b7-bf08-4ff9-ad0d-f0c07b248e94.png">

It might be that I'm calling the run function with the wrong inputs.

NoahZinsmeister:
yeah @0xMartian that's exactly the error i was getting with similar toy models, just a random integer, not at all clear what's going wrong. figuring out the source of/reason for that error is the root of this i think.

0xMartian:
Ah, looks like we need to define the output name as the second arg (#2092).

Does this solve the problem in your self-trained model as well @NoahZinsmeister?

<details>
<summary>
Docs for the second arg, "fetches"
</summary>

     * A fetches (model outputs) could be one of the following:
     *
     * - Omitted. Use model's output names definition.
     * - An array of string indicating the output names.
     * - An object that use output names as keys and OnnxValue or null as corresponding values.


</details>

NoahZinsmeister:
confirming this fixes my issue locally! thanks so much @uniyj and @0xMartian :) really appreciate the help, will keep updating the `feat/onnx` branch

PR #2092, opened 2021-07-22
fix: Feat/onnx

uniyj:
This is based on https://github.com/Uniswap/uniswap-interface/pull/2088, and should override that PR.

Addressing issue #2086.

Turns out @0xMartian's code (see PR above) just needed to pass in the output var name ('output_label') as a second arg to the `session.run` function call. See: https://github.com/microsoft/onnxruntime-inference-examples/blob/f9243417dd2561377de783ce4f6c12bcbe6597ad/js/api-usage_inference-session/inference-session-run.js

The model onnx file was generated by @0xMartian at: https://deepnote.com/project/ONNX-p3P2bx05T_GzH_MMjDTtmg/%2Fnotebook.ipynb#00005-a6ae32ef-9ff8-4397-b384-f2ddd3a97d9b


NoahZinsmeister:
thanks @uniyj , will make further changes directly to the branch :)

PR #2096, opened 2021-07-23
feat: support connecting to a gnosis safe

moodysalem:


cewlio:
this is some great source code, thanks for the freebies!

PR #2102, opened 2021-07-24
test: set up component snapshot testing

justindomingue:
Better support component testing in the interface

Adds snapshot tests for `TextInput` and `ResizableTextArea` as initial examples.


src/state/index.ts
```diff
@@ -18,6 +18,11 @@ import { routingApi } from './routing/slice'
 
 const PERSISTED_KEYS: string[] = ['user', 'transactions', 'lists']
 
+let preloadedState = undefined
+if (process.env.NODE_ENV === 'production' || process.env.NODE_ENV === 'development') {
```
moodysalem, line 22: i think we want !== 'test' or something

src/test-utils.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { createWeb3ReactRoot, Web3ReactProvider } from '@web3-react/core'
+import { NetworkContextName } from 'constants/misc'
+import { ThemedGlobalStyle } from 'theme'
+import { HashRouter } from 'react-router-dom'
+
+const Web3ProviderNetwork = createWeb3ReactRoot(NetworkContextName)
+
+const WithProviders: FC = ({ children }: { children?: ReactNode }) => {
+  return (
+    <Provider store={store}>
```
moodysalem, line 18: you don't need all this stuff for most component tests, including the ones included in this pr afaict. is this a good idea? seems like you're going to end up bundling and rendering half the app, plus making actual network requests in every unit test rather than just what's needed

src/state/index.ts
justindomingue: I removed this block and instead use `load.disableWarning` to disable the first time use warning for test runs.

src/test-utils.tsx
justindomingue: Those were needed for some other components i'm testing locally but I agree it's not ideal. I'll remove them from here for now and consider switching to `jest.mock`

moodysalem (approved):

PR #2112, opened 2021-07-26
fix: prevent error if user selects ETH and WETH in add liquidity

moodysalem:
fixes https://github.com/Uniswap/uniswap-interface/issues/1763

### DIFF
--- a/src/hooks/Tokens.ts
+++ b/src/hooks/Tokens.ts
@@ -179,10 +179,8 @@ export function useCurrency(currencyId: string | undefined): Currency | null | u
   const { chainId } = useActiveWeb3React()
   const isETH = currencyId?.toUpperCase() === 'ETH'
   const token = useToken(isETH ? undefined : currencyId)
-  if (!chainId) return undefined
-
-  const ether = ExtendedEther.onChain(chainId)
-  const weth = WETH9_EXTENDED[chainId]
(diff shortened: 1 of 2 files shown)


src/hooks/Tokens.ts
```diff
@@ -179,8 +179,10 @@ export function useCurrency(currencyId: string | undefined): Currency | null | u
   const { chainId } = useActiveWeb3React()
   const isETH = currencyId?.toUpperCase() === 'ETH'
   const token = useToken(isETH ? undefined : currencyId)
-  const extendedEther = useMemo(() => (chainId ? ExtendedEther.onChain(chainId) : undefined), [chainId])
-  const weth = chainId ? WETH9_EXTENDED[chainId] : undefined
+  if (!chainId) return undefined
+
+  const ether = ExtendedEther.onChain(chainId)
```
justindomingue, line 184: `ExtendedEther.onChain` creates a new object which causes dependencies of `useCurrency` to re-render unnecessarily. That's why we have a memoize extendedEther above.

JFrankfurt:
Seems solid to me. I defer to @justindomingue on the `extendedEther` point. I did notice while flipping the base/quote currency values around that we've got a major perf problem there. Not sure if if it's related or not.

src/hooks/Tokens.ts
moodysalem: ty sir

justindomingue (approved):

PR #2116, opened 2021-07-26
feat: routing api integration

justindomingue:
Feature branch for routing integration.

Loading animation + invalidate stale data visually


src/state/routing/useRouterTrade.ts
```diff
@@ -0,0 +1,147 @@
...
+          tokenInAddress: currencyIn.wrapped.address,
+          tokenInChainId: currencyIn.chainId,
+          tokenOutAddress: amountOut.currency.wrapped.address,
+          tokenOutChainId: amountOut.currency.chainId,
+          amount: amountOut.quotient.toString(),
+          type: 'exactOut',
+          recipient: account ?? undefined,
+          slippageTolerance:
+            typeof userSlippageTolerance === 'string' ? userSlippageTolerance : userSlippageTolerance.toSignificant(),
+          deadline: deadline.toString(),
```
willpote, line 98: Since we need all 3 or none at all, will `deadline.toString()` work? Should we check all 3 are defined first?

src/state/routing/useRoutes.ts
```diff
@@ -0,0 +1,149 @@
...
+        graph,
+        src: currencyIn.wrapped.address,
+        dest: currencyOut.wrapped.address,
+        discovered: {},
+        path: [],
+        acc: allRoutes,
+      })
+
+      return allRoutes.map((route) => {
+        const [pool, rawAmountIn, rawAmountOut] = parseRouteToPool(route, tokensInRoute, quoteResult.routeEdges)
```
willpote, line 54: pools

willpote (commented):
Looking good so far!

src/hooks/useV3Trade.ts
```diff
@@ -0,0 +1,62 @@
...
+): { state: V3TradeState; trade: Trade<Currency, Currency, TradeType.EXACT_INPUT> | null } {
+  // attempt to use multi-route trade
+  const multiRouteTradeExactIn = useRouterTradeExactIn(amountIn, currencyOut)
+
+  const useFallback = shouldUseFallback(multiRouteTradeExactIn.state)
+
+  // only local router if multi-route trade failed
+  const bestV3TradeExactIn = useLocalV3TradeExactIn(
+    useFallback ? amountIn : undefined,
+    useFallback ? currencyOut : undefined
```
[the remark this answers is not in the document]
justindomingue: can't call hooks conditionally :/

src/state/routing/useRouterTrade.ts
justindomingue: I actually check in routing/slice.ts that all 3 are defined. Here I'm just passing as much as possible to abstract away the API constraints.

justindomingue: https://github.com/Uniswap/uniswap-interface/pull/2116/files#diff-c9fd69a3658e7e36f7ab6b62fd18589fed617dad15c3184bc7335b9e6ad7ee02R57

src/state/routing/useRoutes.ts
```diff
@@ -0,0 +1,59 @@
...
+      quoteResult['route'].map((route) => {
+        const rawAmountIn = route[0].amountIn
+        const rawAmountOut = route[route.length - 1].amountOut
+
+        if (!rawAmountIn || !rawAmountOut) {
+          throw new Error('Expected both amountIn and amountOut to be present')
+        }
+
+        // extract currency[In|Out] from route[first|last]
+        const [currencyIn, currencyOut] = [parseToken(route[0].tokenIn), parseToken(route[route.length - 1].tokenOut)]
```
willpote, line 26: Could just do this once outside the map

src/utils/prices.ts
```diff
@@ -29,7 +29,8 @@ export function computeRealizedLPFeePercent(
     )
   } else {
     percent = ONE_HUNDRED_PERCENT.subtract(
-      trade.route.pools.reduce<Percent>(
+      // TODO(judo): compute across all swaps
+      trade.swaps[0].route.pools.reduce<Percent>(
```
willpote, line 33: So i _think_ we just do this per route then sum.. something like:

```
const totalPercent = new Percent(0, 0)
for (const swap of swaps) {
    const overallPercent = <compute using amount and total amount>
    overallPercent.subtract(... // same logic as before)
    totalPercent.add(overallPercent)
}
```

Looks like Moody wrote this though so should check with him

willpote (commented):
Not qualified to review the React code, but otherwise looks good

src/components/swap/SwapRoute.tsx
```diff
@@ -1,47 +1,75 @@
...
+    const path = []
+    for (let i = 1; i < tokenPath.length; i++) {
+      path.push([tokenPath[i - 1], tokenPath[i], undefined] as [Currency, Currency, FeeAmount | undefined])
+    }
+    return [{ percent: new Percent(100, 100), path }]
+  }
+
+  return trade.swaps.map(({ route: { tokenPath, pools }, inputAmount }) => {
+    const portion = inputAmount.divide(trade.inputAmount)
+    const percent = new Percent(portion.numerator, portion.denominator)
```
NoahZinsmeister, line 59: why not just `new Percent(inputAmount, trade.inputAmount)`?

src/hooks/useApproveCallback.ts
```diff
@@ -75,13 +107,9 @@ export function useApproveCallback(
...
-      // general fallback for tokens who restrict approval amounts
-      useExact = true
-      return tokenContract.estimateGas.approve(spender, amountToApprove.quotient.toString())
-    })
+    if (!estimatedGas) {
+      return
+    }
 
     return tokenContract
       .approve(spender, useExact ? amountToApprove.quotient.toString() : MaxUint256, {
```
NoahZinsmeister, line 115: see i thought we were using max approve by default, what was that other code you were showing us @moodysalem ?

src/hooks/useBetterTrade.ts
```diff
@@ -0,0 +1,212 @@
...
+  const deadline = useTransactionDeadline()
+  const routerContract = useV2RouterContract()
+
+  const [v2SwapGasEstimateWei, setV2SwapGasEstimate] = useState<BigNumber | undefined>()
+
+  useEffect(() => {
+    if (!library || !routerContract || !trade) return
+
+    const sampleSlippage = new Percent(5, 100)
+    const sampleRecipient = '0xab5801a7d398351b8be11c439e05c5b3259aec9b'
```
NoahZinsmeister, line 31: out of curiosity what is this? also should probably be checksummed

src/hooks/useBetterTrade.ts
```diff
@@ -0,0 +1,212 @@
...
+      !value || isZero(value)
+        ? { /* from: account,  */ to: address, data: calldata }
+        : {
+            /* from: account, */
+            to: address,
+            data: calldata,
+            value,
+          }
+
+    library.estimateGas(tx).then((gasEstimate) => {
```
NoahZinsmeister, line 54: does this not always fail because the `from` isn't specified, and there's no way to mock the approval (or that the address has enough tokens) in the generic case?

src/utils/prices.ts
```diff
@@ -28,13 +29,24 @@ export function computeRealizedLPFeePercent(
       )
     )
   } else {
-    percent = ONE_HUNDRED_PERCENT.subtract(
-      trade.route.pools.reduce<Percent>(
-        (currentFee: Percent, pool): Percent =>
-          currentFee.multiply(ONE_HUNDRED_PERCENT.subtract(new Fraction(pool.fee, 1_000_000))),
-        ONE_HUNDRED_PERCENT
+    //TODO(judo): validate this
```
NoahZinsmeister, line 32: seems correct after a 1st pass

NoahZinsmeister (commented):
i want to spend some more time with `useBetterTrade`, but the PR looks close as far as i can tell. left a few comments for now!

src/hooks/useBetterTrade.ts
justindomingue: was Vitalik's address 😬 will most likely be removed!

src/hooks/useApproveCallback.ts
moodysalem: you're right, this code shows we do max approve

imo, this code has really surprising behavior though, the code computing the `amountToApprove` argument uses `Trade#maximumAmountIn` (just below this function). i would prefer if `maxApprove` was a boolean argument to the function that defaulted to false

src/state/swap/hooks.ts
```diff
@@ -160,15 +161,25 @@ export function useDerivedSwapInfo(toggledVersion: Version): {
   const bestV3TradeExactIn = useV3TradeExactIn(isExactIn ? parsedAmount : undefined, outputCurrency ?? undefined)
   const bestV3TradeExactOut = useV3TradeExactOut(inputCurrency ?? undefined, !isExactIn ? parsedAmount : undefined)
 
-  const v2Trade = isExactIn ? bestV2TradeExactIn : bestV2TradeExactOut
-  const v3Trade = isExactIn ? bestV3TradeExactIn : bestV3TradeExactOut
+  const v2TradeRaw = isExactIn ? bestV2TradeExactIn : bestV2TradeExactOut
+  const v3TradeState = isExactIn ? bestV3TradeExactIn : bestV3TradeExactOut
+  const v2Trade = v2TradeRaw ?? undefined
+  const v3Trade = v3TradeState.trade && v3TradeState.state == V3TradeState.VALID ? v3TradeState.trade : undefined
```
justindomingue, line 167: @willpote this isn't ideal since the interfaces uses this trade object to render the ui

src/components/swap/SwapRoute.tsx
justindomingue: hm i'd need to format `inputAmount` into a `BigIntIsh`. Creating the percent from the fraction seemed more straightfoward.

src/components/RoutingDiagram/RoutingDiagram.tsx
```diff
@@ -0,0 +1,114 @@
...
+  currency0: Currency
+  currency1: Currency
+  feeAmount: FeeAmount | undefined
+}) {
+  return (
+    <PoolBadge>
+      <Box margin="0 5px 0 10px">
+        <DoubleCurrencyLogo currency0={currency1} currency1={currency0} size={20} />
+      </Box>
+      {feeAmount && <TYPE.small fontSize={12}>{feeAmount / 10000}%</TYPE.small>}
```
moodysalem, line 111: is `Number#toString` what we want here?

moodysalem, line 111: technically `feeAmount` can be 0 which would fail the ternary check

cmcewen (commented):
not all the way through it yet so will be back for more

src/components/Tooltip/index.tsx
```diff
@@ -15,14 +19,22 @@ interface TooltipProps extends Omit<PopoverProps, 'content'> {
...
   return <Popover content={<TooltipContainer>{text}</TooltipContainer>} {...rest} />
 }
 
-function TooltipContent({ content, ...rest }: TooltipContentProps) {
-  return <Popover content={<TooltipContainer>{content}</TooltipContainer>} {...rest} />
+function TooltipContent({ content, container, ...rest }: TooltipContentProps) {
+  return (
+    <Popover
+      content={
+        container ? cloneElement(container, { children: content }) : <TooltipContainer>{content}</TooltipContainer>
```
moodysalem, line 33: clone element smells, almost never a good idea since you're overriding one of the props (in this case the children of the component). just render the Popover in the one case where you don't want a tooltip container

src/components/swap/AdvancedSwapDetails.tsx
```diff
@@ -1,22 +1,29 @@
...
 
 interface AdvancedSwapDetailsProps {
   trade?: V2Trade<Currency, Currency, TradeType> | V3Trade<Currency, Currency, TradeType>
   allowedSlippage: Percent
+  loading?: boolean
 }
 
-export function AdvancedSwapDetails({ trade, allowedSlippage }: AdvancedSwapDetailsProps) {
+/** Renders a loading placeholder if loading and the children otherwise. */
+function DimmableText({ dim, width, children }: { dim: boolean; width: number; children: React.ReactNode }) {
```
moodysalem, line 22: dimmable not a great name

src/components/swap/AdvancedSwapDetails.tsx
```diff
@@ -30,61 +37,61 @@ export function AdvancedSwapDetails({ trade, allowedSlippage }: AdvancedSwapDeta
 
   return !trade ? null : (
     <AutoColumn gap="8px">
+      <TransactionDetailsLabel fontWeight={500} fontSize={14}>
+        <Trans>Transaction Details</Trans>
```
moodysalem, line 41: check this copy?

src/components/swap/AdvancedSwapDetails.tsx
```diff
@@ -30,61 +37,61 @@ export function AdvancedSwapDetails({ trade, allowedSlippage }: AdvancedSwapDeta
...
+        </DimmableText>
       </RowBetween>
 
       <RowBetween>
         <RowFixed>
-          <TYPE.black fontSize={12} fontWeight={400} color={theme.text2}>
-            {trade.tradeType === TradeType.EXACT_INPUT ? <Trans>Minimum received</Trans> : <Trans>Maximum sent</Trans>}
-          </TYPE.black>
+          <TYPE.subHeader color={theme.text1}>
+            <Trans>Slippage tolerance</Trans>
```
moodysalem, line 72: inconsistent capitalization

moodysalem (commented):
(per slack) we should not make requests while the window is not focused, the same way we do not make multicall/block number requests

first batch, still reviewing

src/components/CurrencyInputPanel/index.tsx
```diff
@@ -247,15 +251,14 @@ export default function CurrencyInputPanel({
...
-                }}
-              />
-            </>
+            <NumericalInput
+              className="token-amount-input"
+              value={value}
+              onUserInput={(val) => {
+                onUserInput(val)
+              }}
+              style={loading ? { opacity: '.3', transition: 'opacity 0.2s ease-in-out' } : {}}
```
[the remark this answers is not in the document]
justindomingue: i thought showing old values was awkward at first as well but I kinda like it now. It removes the jerkiness that's in the prod ui right now where the entire form clears. 

Definitely need to change this css.

src/components/swap/AdvancedSwapDetails.tsx
justindomingue: checked with design/pm

src/components/Loader/LoadingBar.tsx
```diff
@@ -0,0 +1,21 @@
...
+  50% {
+    opacity: 1;
+  }
+  100% {
+    opacity: 0.3;
+  }
+`
+
+export const LoadingBar = styled.div<{ width: number; height: number }>`
+  animation: ${pulse} 1s ease-in-out infinite alternate;
```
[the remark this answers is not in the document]
justindomingue: ah good call, didn't realize we had one. I'll use it for the swap page tooltips.

src/state/routing/useRoutes.ts
```diff
@@ -0,0 +1,74 @@
+import { Currency, CurrencyAmount, Ether, Token } from '@uniswap/sdk-core'
+import { FeeAmount, Pool, Route } from '@uniswap/v3-sdk'
+import { useMemo } from 'react'
+import { GetQuoteResult } from 'state/routing/slice'
+
+export function useRoutes(
```
moodysalem, line 6: can you explain this function in a comment? also seems like it doesn't need to be a hook unless you want to use the token info from token lists to get the logos

src/state/routing/useRoutes.ts
```diff
@@ -0,0 +1,74 @@
...
+      // `Route` constructor may throw if inputs/outputs are temporarily out of sync
+      // (RTK-Query always returns the latest data which may not be the right inputs/outputs)
+      // This is not fatal and will fix itself in future render cycles
+      console.debug(e)
+      return undefined
+    }
+  }, [currencyIn, currencyOut, quoteResult])
+}
+
+const parseToken = ({ address, chainId, decimals, symbol }: GetQuoteResult['route'][0][0]['tokenIn']): Currency => {
```
moodysalem, line 55: should probably be `[number]` instead of `[0]`?

moodysalem: why doesn't this just return Token instead of Currency and you don't need wrapped below

src/state/user/hooks.tsx
```diff
@@ -105,19 +105,21 @@ export function useExpertModeManager(): [boolean, () => void] {
...
-      dispatch(updateUserSingleHopOnly({ userSingleHopOnly: newSingleHopOnly }))
+  const setLegacyRouter = useCallback(
+    (newLegacyRouter: boolean) => {
+      dispatch(updateUserLegacyRouter({ userLegacyRouter: newLegacyRouter }))
     },
     [dispatch]
   )
 
-  return [singleHopOnly, setSingleHopOnly]
+  return [legacyRouter || chainId !== SupportedChainId.MAINNET, setLegacyRouter]
```
moodysalem, line 122: this is a bit messy because `setLegacyRouter(!legacyRouter)` won't do anything. instead you want to show the toggle as disabled when not on mainnet

src/utils/isTradeBetter.ts
```diff
@@ -18,7 +18,7 @@ export function isTradeBetter(
     !tradeA.inputAmount.currency.equals(tradeB.inputAmount.currency) ||
     !tradeA.outputAmount.currency.equals(tradeB.outputAmount.currency)
   ) {
-    throw new Error('Comparing incomparable trades')
+    return undefined
```
moodysalem, line 21: why did we change the validation?

src/state/routing/slice.ts
```diff
@@ -41,12 +53,9 @@ export const routingApi = createApi({
         tokenOutChainId: SupportedChainId
         amount: string
         type: 'exactIn' | 'exactOut'
-        recipient?: string
-        slippageTolerance?: string
-        deadline?: string
       }
     >({
-      query: (args) => `quote?${qs.stringify(args)}`,
+      query: (args) => `quote?${qs.stringify({ ...args })}`,
```
moodysalem, line 58: why copy it?

src/pages/Swap/index.tsx
```diff
@@ -118,6 +127,29 @@ export default function Swap({ history }: RouteComponentProps) {
...
   } = useDerivedSwapInfo(toggledVersion)
 
+  const toggleVersionCallback = useToggleVersionCallback()
+
+  // automatically toggle best version
+  useEffect(() => {
+    if (!bestTrade || !trade) return
+
+    if (bestTrade !== trade) {
+      toggleVersionCallback()
```
moodysalem, line 137: i would be more comfortable with this explicitly changing to the bestTrade version, but also i think we don't need to switch via the url, just always use the best trade right?

src/components/RoutingDiagram/RoutingDiagram.test.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+import RoutingDiagram, { RoutingDiagramEntry } from './RoutingDiagram'
+import { Token, Percent, Currency } from '@uniswap/sdk-core'
+import { FeeAmount } from '@uniswap/v3-sdk'
+import { render } from 'test-utils'
+
+const percent = (strings: TemplateStringsArray) => new Percent(parseInt(strings[0]), 1000)
+
+const USDC = new Token(1, '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48', 6, 'USDC')
```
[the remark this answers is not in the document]
justindomingue: Yes! Thanks, didn't realize we had those defined.

src/components/RoutingDiagram/RoutingDiagram.tsx
justindomingue: `feeAmount` is `3000` for 0.03%.

src/components/RoutingDiagram/RoutingDiagram.tsx
```diff
@@ -0,0 +1,114 @@
...
+import CurrencyLogo from 'components/CurrencyLogo'
+import DoubleCurrencyLogo from 'components/DoubleLogo'
+import Row, { AutoRow } from 'components/Row'
+import { Box } from 'rebass'
+import styled from 'styled-components/macro'
+import { TYPE } from 'theme'
+
+export interface RoutingDiagramEntry {
+  percent: Percent
+  path: [Currency, Currency, FeeAmount | undefined][]
```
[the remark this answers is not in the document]
justindomingue: Yeah, this allows us to treat v2's `Trade.route` and v3's `Trade.swaps` the same way in the UI. I'd imagine that we'd get rid of this and use `Trade.swaps` only with v2+v3 routing.

src/components/swap/RouterLabel.tsx
```diff
@@ -0,0 +1,48 @@
...
+    /* will-change: background-position; */
+  }
+`
+
+export function RouterLabel({ syncing = false }: { syncing?: boolean }) {
+  return (
+    <AutoRow gap="4px" width="auto">
+      <StyledAutoRouterIcon />
+      <GradientText fontSize={14} pulsing={syncing}>
+        Auto Router
```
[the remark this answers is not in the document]
justindomingue: This is branded so we'll keep it in English. @willhennessy @callil can you confirm?

src/components/RoutingDiagram/RoutingDiagram.tsx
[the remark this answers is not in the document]
justindomingue: The type `[Currency, Currency, FeeAmount | undefined]` doesn't really have meaning outside of `RoutingDiagramEntry`. I changed some places where I manually `[Currency, Currency, FeeAmount | undefined]` to `RoutingDiagramEntry['path'][0]` to add some readability.

src/components/Popover/index.tsx
```diff
@@ -90,6 +88,7 @@ export default function Popover({ content, show, children, placement = 'auto' }:
     modifiers: [
       { name: 'offset', options: { offset: [8, 8] } },
       { name: 'arrow', options: { element: arrowElement } },
```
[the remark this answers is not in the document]
justindomingue: is there harm in keeping the `arrow` modifier when `hideArrow=true`? Seems more readable to just leave it. `arrowElement` will be null

src/components/Tooltip/index.tsx
justindomingue: > just render the Popover in the one case where you don't want a tooltip container
not quite sure what you mean by that

I changed container to a functional component.

src/components/swap/SwapRoute.tsx
```diff
@@ -1,47 +1,82 @@
...
+
+function getTokenPath(
+  trade: V2Trade<Currency, Currency, TradeType> | V3Trade<Currency, Currency, TradeType>
+): RoutingDiagramEntry[] {
+  // convert V2 path to a list of routes
+  if (trade instanceof V2Trade) {
+    const { path: tokenPath } = (trade as V2Trade<Currency, Currency, TradeType>).route
+    const path = []
+    for (let i = 1; i < tokenPath.length; i++) {
+      path.push([tokenPath[i - 1], tokenPath[i], FeeAmount.MEDIUM] as [Currency, Currency, FeeAmount | undefined])
```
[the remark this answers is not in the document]
justindomingue: I had the fee amount set to `undefined` for v2 initially. It's always set to MEDIUM now so I can remove the `undefined` type

src/hooks/useToggledVersion.ts
```diff
@@ -21,3 +24,28 @@ export default function useToggledVersion(): Version {
       return Version.v3
   }
 }
+
+export function useToggleVersionCallback() {
```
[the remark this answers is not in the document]
justindomingue: hm in this case there is no `Link` to click, the callback is used to trigger a switch programmatically

src/hooks/useToggledVersion.ts
```diff
@@ -21,3 +24,28 @@ export default function useToggledVersion(): Version {
...
+      ...location,
+      search: `?${stringify({
+        ...rest,
+        ...newUse,
+      })}`,
+    }
+  }, [location, search, version])
+
+  return useCallback(() => {
+    history.replace(linkDestination)
```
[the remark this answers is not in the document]
justindomingue: I like `replace` because if the user hits the back arrow in the browser to go back, we'll actually redirect them again to the better version. So pushing in this case has poor behavior imo.

src/state/routing/useRoutes.ts
justindomingue: `Currency` is helpful because the input token may be native.

I don't think that `GetQuoteResult['quote'][number]` works? Is that what you mean?

src/state/routing/useRoutes.ts
justindomingue:
```
/**
 * Transforms a Routing API quote into an array of routes that
 * can be used to create a V3 `Trade`.
 */
```

I made it a hook so it would memoize itself rather than having the caller memoize it. Do you prefer if I keep this as a function and have the caller memoize? I guess a regular function is easier to test but all the tests are already written with `renderHook`

src/components/RoutingDiagram/RoutingDiagram.tsx
moodysalem: o that should be 0.3%, but i think this code is correct anyway.

src/state/routing/useRoutes.ts
moodysalem: i would rather it be a function and the caller does the memoize

src/state/routing/useRoutes.ts
moodysalem: but doesn't the parseToken method always return a Token? i don't see how it can return a currency with this implementation

justindomingue: Ah right, changed to `Token`.

JFrankfurt:
image
Just ran into this. Not sure what's causing it, but the same input is not resulting in the incorrect insufficient balance state on prod.

src/pages/Swap/index.tsx
justindomingue: agreed. changing it so that url param forces a version, no url param just uses the best trade without redirecting.

src/components/RoutingDiagram/RoutingDiagram.tsx
justindomingue: oops, yeah meant to say `0.3%`.

src/state/user/hooks.tsx
justindomingue: good call, design preferred hiding the toggle entirely on non mainnet.

src/hooks/useV3Trade.ts
```diff
@@ -0,0 +1,102 @@
...
+  gasPriceWei?: BigNumber
+  gasUseEstimate?: BigNumber
+} {
+  const debouncedAmountIn = useDebounce(amountIn, 250)
+
+  // attempt to use multi-route trade
+  const multiRouteTradeExactIn = useRouterTradeExactIn(debouncedAmountIn, currencyOut)
+
+  // consider trade debouncing when inputs/outputs do not match
+  const debouncing =
```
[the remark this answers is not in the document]
justindomingue: Hm that's a bit of a quirkiness of RTK-Query (https://github.com/reduxjs/redux-toolkit/issues/1339#issue-950885472). I need to check that the data returned by RTK-Query is what the input args are expecting, otherwise treat the query as debouncing. RTK-Query is releasing a workaround I can experiment with (https://github.com/reduxjs/redux-toolkit/pull/1500)

src/components/Loader/styled.tsx
```diff
@@ -0,0 +1,33 @@
...
+    background-size: 400%;
+    border-radius: 12px;
+    height: 2.4em;
+    will-change: background-position;
+  }
+`
+
+export function opacityLoadingStyles(loading: boolean) {
+  return { transition: 'opacity 0.2s ease-in-out', ...(loading ? { opacity: '0.3' } : { opacity: '1' }) }
+}
```
JFrankfurt, line 33: pref doing this with styled-components instead of inline styles https://styled-components.com/docs/api#css

src/components/Popover/index.tsx
```diff
@@ -78,20 +75,29 @@ export interface PopoverProps {
   show: boolean
   children: React.ReactNode
   placement?: Placement
+  hideArrow?: boolean
```
JFrankfurt, line 78: I don't think the `?` is correct here, since there's a default argument

src/components/Popover/index.tsx
```diff
@@ -103,12 +109,14 @@ export default function Popover({ content, show, children, placement = 'auto' }:
       <Portal>
         <PopoverContainer show={show} ref={setPopperElement as any} style={styles.popper} {...attributes.popper}>
           {content}
-          <Arrow
-            className={`arrow-${attributes.popper?.['data-popper-placement'] ?? ''}`}
-            ref={setArrowElement as any}
-            style={styles.arrow}
-            {...attributes.arrow}
-          />
+          {!hideArrow && (
```
JFrankfurt, line 112: nit: `!hide...` → `show`

src/hooks/useV3Trade.ts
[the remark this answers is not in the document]
mgowon7: Habit

src/components/Settings/index.tsx
```diff
@@ -123,7 +127,7 @@ export default function SettingsTab({ placeholderSlippage }: { placeholderSlippa
 
   const [expertMode, toggleExpertMode] = useExpertModeManager()
 
-  const [singleHopOnly, setSingleHopOnly] = useUserSingleHopOnly()
+  const [legacyRouter, setLegacyRouter] = useIsLegacyRouter()
```
JFrankfurt, line 130: pref more specificity than `legacy`

src/components/swap/AdvancedSwapDetails.tsx
```diff
@@ -1,22 +1,42 @@
...
+  loading: boolean
+  width: number
+  children: React.ReactNode
+}) {
+  return loading ? (
+    <LoadingRows>
+      <div style={{ height: '15px', width: `${width}px` }} />
+    </LoadingRows>
+  ) : (
+    <>{children}</>
```
JFrankfurt, line 35: do you need the fragment here?

src/components/swap/RouterLabel.tsx
```diff
@@ -0,0 +1,54 @@
...
+  stroke: #2172e5;
+`
+
+const GradientText = styled(TYPE.black)<{ pulsing: boolean }>`
+  line-height: 1rem;
+
+  /* fallback color */
+  color: ${({ theme }) => theme.green1};
+
+  @supports (-webkit-background-clip: text) and (-webkit-text-fill-color: transparent) {
```
JFrankfurt, line 22: nice

src/components/swap/RouterLabel.tsx
```diff
@@ -0,0 +1,54 @@
...
+      margin="-.25rem"
+      disabled
+      style={{
+        opacity: 0.8,
+      }}
+    >
+      <TYPE.black fontSize={12}>{version === Version.v2 ? <Trans>V2</Trans> : <Trans>V3</Trans>}</TYPE.black>
+    </ButtonGray>
+  ) : (
+    <div></div>
```
JFrankfurt, line 52: Pref avoiding nested ternaries and returning an empty div

src/components/swap/SwapRoute.tsx
```diff
@@ -1,47 +1,88 @@
...
+
+function getTokenPath(
+  trade: V2Trade<Currency, Currency, TradeType> | V3Trade<Currency, Currency, TradeType>
+): RoutingDiagramEntry[] {
+  // convert V2 path to a list of routes
+  if (trade instanceof V2Trade) {
+    const { path: tokenPath } = (trade as V2Trade<Currency, Currency, TradeType>).route
+    const path = []
+    for (let i = 1; i < tokenPath.length; i++) {
+      path.push([tokenPath[i - 1], tokenPath[i], V2_DEFAULT_FEE_TIER] as RoutingDiagramEntry['path'][0])
```
JFrankfurt, line 65: Not familiar w/ the `Type['key'][n]` syntax here. Is this for sure the best way to type the things it's being applied to?

src/components/Loader/styled.tsx
justindomingue: neat! so much better

src/components/Popover/index.tsx
justindomingue: TS complains if I remove the `?` since the props interface doesn't know about the default value.

src/components/Popover/index.tsx
justindomingue: using `hide` so it defaults to `false` rather than `true`

src/components/swap/AdvancedSwapDetails.tsx
justindomingue: got it to work by changing the type of `children` to `JSX.Element`

JFrankfurt: Might consider `ReactChild | ReactChildren` as another option. Up to you!

src/components/swap/SwapRoute.tsx
justindomingue: hm that just makes it so that i don't have to extract the inner types to an external variable. That type doesn't really have meaning outside of the super type.

src/components/swap/AdvancedSwapDetails.tsx
justindomingue: Hm can't get the typechecked to pass with `ReactChild | ReactChildren`. Might keep `jsx.element`

src/state/routing/types.ts
```diff
@@ -0,0 +1,44 @@
+export enum V3TradeState {
+  LOADING,
+  INVALID,
+  NO_ROUTE_FOUND,
+  VALID,
+  SYNCING,
+}
+
+export interface GetQuoteResult {
```
moodysalem, line 9: @willpote is there a package containing this type?

src/state/routing/computeRoutes.ts
```diff
@@ -0,0 +1,75 @@
...
+  sqrtRatioX96,
+  liquidity,
+  tickCurrent,
+  tokenIn,
+  tokenOut,
+}: GetQuoteResult['route'][0][0]): Pool =>
+  new Pool(
+    parseToken(tokenIn),
+    parseToken(tokenOut),
+    parseInt(fee) as FeeAmount,
```
moodysalem, line 71: this type should probably be number in the sdk

src/state/routing/computeRoutes.ts
```diff
@@ -0,0 +1,75 @@
...
+        route: new Route(route.map(parsePool), parsedCurrencyIn, parsedCurrencyOut),
+        inputAmount: CurrencyAmount.fromRawAmount(parsedCurrencyIn, rawAmountIn),
+        outputAmount: CurrencyAmount.fromRawAmount(parsedCurrencyOut, rawAmountOut),
+      }
+    })
+  } catch (e) {
+    // `Route` constructor may throw if inputs/outputs are temporarily out of sync
+    // (RTK-Query always returns the latest data which may not be the right inputs/outputs)
+    // This is not fatal and will fix itself in future render cycles
+    console.debug(e)
```
moodysalem, line 51: another try catch that indicates something's up with our knowledge of states/loading

moodysalem: do we need to compute routes while we're loading? don't we show a loading indicator?

src/state/index.ts
```diff
@@ -44,6 +45,8 @@ const store = configureStore({
 
 store.dispatch(updateVersion())
 
+setupListeners(store.dispatch)
```
moodysalem, line 48: they thought of everything didn't they

src/pages/Swap/index.tsx
```diff
@@ -529,15 +503,21 @@ export default function Swap({ history }: RouteComponentProps) {
...
                     ) : null)}
                 </ButtonPrimary>
+              ) : routeIsSyncing ? (
+                <GreyCard style={{ textAlign: 'center' }}>
+                  <TYPE.main mb="4px">
+                    <Dots>
+                      <Trans>Loading</Trans>
+                    </Dots>
+                  </TYPE.main>
+                </GreyCard>
```
moodysalem, line 513: looks the same as loading below

src/pages/Swap/index.tsx
```diff
@@ -438,74 +442,44 @@ export default function Swap({ history }: RouteComponentProps) {
...
-                    </MouseoverTooltipContent>
-                  </RowFixed>
-                ) : null}
+                  </LoadingOpacityContainer>
+                  <MouseoverTooltipContent
+                    Container={({ children }) => (
+                      <ResponsiveTooltipContainer origin="bottom right" width={'295px'}>
+                        {children}
+                      </ResponsiveTooltipContainer>
+                    )}
```
moodysalem, line 474: this inline container component does not need to be defined in every render cycle

src/pages/Swap/index.tsx
```diff
@@ -438,74 +442,44 @@ export default function Swap({ history }: RouteComponentProps) {
...
-                      <TYPE.black fontSize={12}>
-                        <Trans>V3</Trans>
-                      </TYPE.black>
-                    </ButtonGray>
-                  )}
+            {!showWrap && trade && (
+              <Row justify={!trade ? 'center' : 'space-between'}>
+                <RowFixed style={{ position: 'relative' }}>
+                  <MouseoverTooltipContent
+                    Container={({ children }) => <ResponsiveTooltipContainer>{children}</ResponsiveTooltipContainer>}
```
moodysalem, line 449: this inline container does not need to be defined in every render cycle

src/pages/Swap/index.tsx
```diff
@@ -438,74 +442,44 @@ export default function Swap({ history }: RouteComponentProps) {
...
+                    content={<SwapRoute trade={trade} syncing={routeIsSyncing} />}
```
moodysalem, line 450: i don't like this pattern very much. we have a container, and then we have content. why can't you just specify the container in the content? a `boolean` prop for whether to wrap the content might be more clear

src/pages/Swap/index.tsx
```diff
@@ -107,16 +114,16 @@ export default function Swap({ history }: RouteComponentProps) {
 
   // swap state
   const { independentField, typedValue, recipient } = useSwapState()
+  const newLocal = useDerivedSwapInfo(toggledVersion)
   const {
-    v2Trade,
-    v3TradeState: { trade: v3Trade, state: v3TradeState },
-    toggledTrade: trade,
+    v3Trade: { state: v3TradeState },
```
moodysalem, line 119: i think ideally we push this into the hook, and expose a single loading state for whether v2/v3 trades are loaded/current/syncing. maybe this will help clean up/prevent the inconsistent loading state issues in the component

src/components/Tooltip/index.tsx
```diff
@@ -15,14 +21,20 @@ interface TooltipProps extends Omit<PopoverProps, 'content'> {
...
 export default function Tooltip({ text, ...rest }: TooltipProps) {
   return <Popover content={<TooltipContainer>{text}</TooltipContainer>} {...rest} />
 }
 
-function TooltipContent({ content, ...rest }: TooltipContentProps) {
-  return <Popover content={<TooltipContainer>{content}</TooltipContainer>} {...rest} />
+function TooltipContent({ content, Container, ...rest }: TooltipContentProps) {
+  return (
+    <Popover
+      content={Container ? Container({ children: content }) : <TooltipContainer>{content}</TooltipContainer>}
```
moodysalem, line 34: you _can_ use jsx here and default the argument to `TooltipContainer` to avoid the ternary

src/hooks/useTokenInfoFromActiveList.ts
```diff
@@ -0,0 +1,22 @@
...
+ * or the passed token otherwise. */
+export function useTokenInfoFromActiveList(currency: Currency) {
+  const { chainId } = useActiveWeb3React()
+  const activeList = useCombinedActiveList()
+
+  return useMemo(() => {
+    if (!chainId) return
+
+    try {
+      return activeList[chainId][currency.wrapped.address].token
```
moodysalem, line 17: if it's native, do you really want the token? just return `currency`

src/state/routing/computeRoutes.test.ts
```diff
@@ -0,0 +1,215 @@
...
+            sqrtRatioX96: '2437312313659959819381354528',
+            liquidity: '10272714736694327408',
+            tickCurrent: '-69633',
+          },
+          {
+            address: '0x3f8F72aA9304c8B593d555F12eF6589cC3A579A2',
+            amountIn: amount`1`,
+            amountOut: amount`200`,
+            fee: '10000',
+            tokenIn: MKR,
```
moodysalem, line 89: it surprised me that the shape of the getquoteresult happened to work with `Token`, but that's ok

src/state/routing/useRoutingAPITrade.ts
```diff
@@ -0,0 +1,175 @@
...
+    amount: amountOut,
+    type: 'exactOut',
+  })
+
+  const { isLoading, isError, data } = useGetQuoteQuery(queryArgs ?? skipToken, {
+    pollingInterval: ms`10s`,
+    refetchOnFocus: true,
+  })
+
+  const quoteResult = useFreshData(data, Number(data?.blockNumber) ?? 0)
```
moodysalem, line 129: `Number(undefined) ?? 0` == `NaN`, not what we want. it's inconsequential in this case but could cause a bug

src/state/routing/useRoutingAPITrade.ts
```diff
@@ -0,0 +1,175 @@
...
+  type: 'exactIn' | 'exactOut'
+}) {
+  if (!tokenIn || !tokenOut || !amount || tokenIn.equals(tokenOut)) {
+    return undefined
+  }
+
+  return {
+    tokenInAddress: tokenIn.wrapped.address,
+    tokenInChainId: tokenIn.chainId,
+    tokenOutAddress: tokenOut.wrapped.address,
```
moodysalem, line 45: hmm does the api not accept ETH as some kind of argument?

src/state/swap/hooks.ts
```diff
@@ -147,20 +150,58 @@ export function useDerivedSwapInfo(toggledVersion: Version): {
...
+      // v3 trade may be debouncing or fetching and have different
+      // inputs/ouputs than v2
+      return undefined
+    }
+  }, [v2Trade, v3Trade.trade])
+
+  const bestTrade = isV2TradeBetter == undefined ? undefined : isV2TradeBetter ? v2Trade : v3Trade.trade
+
+  if (isV2TradeBetter !== undefined) {
+    console.debug(
```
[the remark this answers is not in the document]
justindomingue: this was added by @willpote for testing. I don't think we need to check this to main. Will remove unless @willpote thinks I should keep?

src/state/routing/useRoutingAPITrade.ts
justindomingue: it does accept `ETH` as `tokenAddress` but @willpote mentioned passing WETH will not change the outcome

src/state/routing/useRoutingAPITrade.ts
justindomingue: good catch! `|| 0` is a good alternative yeah? Given `NaN` cannot be evaluated to `true`

src/state/routing/computeRoutes.test.ts
justindomingue: the return type was adjusted for `Token`. What's the issue you see here? API not generic enough?

src/components/Tooltip/index.tsx
justindomingue: i like using jsx, the default arg definitely cleansi tup

justindomingue: I did update to a `wrap` boolean

src/state/index.ts
justindomingue: 😆

src/components/Loader/styled.tsx
```diff
@@ -0,0 +1,33 @@
...
+    );
+    background-size: 400%;
+    border-radius: 12px;
+    height: 2.4em;
+    will-change: background-position;
+  }
+`
+
+export function opacityLoadingStyles(loading: boolean) {
+  return { transition: 'opacity 0.2s ease-in-out', ...(loading ? { opacity: '0.3' } : { opacity: '1' }) }
```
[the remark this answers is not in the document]
justindomingue: code is outdated, should be using `css`

src/state/routing/useRoutingAPITrade.ts
moodysalem: true, we do have `currencyId` function though and may want to be accurate if it doesn't cost us anything

src/state/routing/computeRoutes.ts
justindomingue: @willpote

Issue #1727, opened 2021-05-27
Import pool title should be center aligned


willhennessy:
1. Visit https://app.uniswap.org/#/pool
2. click More > Migrate
3. click 'Import it' at the bottom of the box

Expected:  the title "Import Pool" should be center aligned in the box
Actual:  it is right aligned - and not even aligned with the right edge of the form boxes

<img width="956" alt="Screen Shot 2021-05-27 at 7 26 47 PM" src="https://user-images.githubusercontent.com/5455577/119908961-c67f7580-bf21-11eb-8bb6-130bc8846153.png">

msquinn:
I think I have a working fix for this locally. If you want to assign this to me I can open up a PR for this. Thanks

PR #2122, opened 2021-07-29
fix: center import pool text V2 

msquinn:
This is a fix for #1727


JFrankfurt:
Looks to like it's still a bit off-center: 
image


I think you'll need something like this instead:
image
with a `position: relative` on the parent.

msquinn:
@JFrankfurt ah, Thank you! That looks better. I appreciate the help!

PR #2123, opened 2021-07-30
feat(chain-connectivity): warn when the user is not receiving blocks

JFrankfurt:
### Arbitrum
image

### Optimism
image

### Mainnet 🙀
image

### DIFF
--- a/src/components/Header/ChainConnectivityWarning.tsx
+++ b/src/components/Header/ChainConnectivityWarning.tsx
@@ -57,7 +57,11 @@ export function ChainConnectivityWarning() {
         </TitleText>
       </TitleRow>
       <BodyRow>
-        <Trans>{label} may be down right now, or you may have lost your network connection.</Trans>{' '}
+        {chainId === SupportedChainId.MAINNET ? (
+          <Trans>You may have lost your network connection.</Trans>
+        ) : (
+          <Trans>{label} may be down right now, or you may have lost your network connection.</Trans>
+        )}{' '}
         {(info as L2ChainInfo).statusPage !== undefined && (
           <span>
             <Trans>Check network status</Trans>{' '}
--- a/src/components/Header/ChainConnectivityWarning.tsx
+++ b/src/components/Header/ChainConnectivityWarning.tsx
@@ -1,19 +1,19 @@
 import { Trans } from '@lingui/macro'
-import { CHAIN_INFO, L2ChainInfo } from 'constants/chains'
+import { CHAIN_INFO, L2ChainInfo, SupportedChainId } from 'constants/chains'
 import { useActiveWeb3React } from 'hooks/web3'
 import { AlertOctagon } from 'react-feather'
 import styled from 'styled-components/macro'
 import { ExternalLink, MEDIA_WIDTHS } from 'theme'
 
 const BodyRow = styled.div`
-  color: black;
+  color: ${({ theme }) => theme.black};
   font-size: 12px;
 `
 const CautionIcon = styled(AlertOctagon)`
-  color: black;
+  color: ${({ theme }) => theme.black};
 `
 const Link = styled(ExternalLink)`
-  color: black;
+  color: ${({ theme }) => theme.black};
   text-decoration: underline;
 `
 const TitleRow = styled.div`
@@ -45,7 +45,7 @@ const Wrapper = styled.div`
 
 export function ChainConnectivityWarning() {
   const { chainId } = useActiveWeb3React()
-  const info = CHAIN_INFO[chainId || '1']
+  const info = CHAIN_INFO[chainId ?? SupportedChainId.MAINNET]
   const label = info?.label
 
   return (
--- a/src/components/Header/ChainConnectivityWarning.tsx
+++ b/src/components/Header/ChainConnectivityWarning.tsx
@@ -53,7 +53,7 @@ export function ChainConnectivityWarning() {
       <TitleRow>
         <CautionIcon />
         <TitleText>
-          <Trans>Network Downtime Warning</Trans>
+          <Trans>Network Warning</Trans>
         </TitleText>
       </TitleRow>
       <BodyRow>
--- a/src/components/Header/ChainConnectivityWarning.tsx
+++ b/src/components/Header/ChainConnectivityWarning.tsx
@@ -1,0 +1,64 @@
+import { Trans } from '@lingui/macro'
+import { AlertOctagon } from 'react-feather'
+import styled from 'styled-components/macro'
+import { ExternalLink, MEDIA_WIDTHS } from 'theme'
+
+const BodyRow = styled.div`
+  color: black;
+  font-size: 12px;
+`
+const CautionIcon = styled(AlertOctagon)`
+  color: black;
+`
+const Link = styled(ExternalLink)`
+  color: black;
+  text-decoration: underline;
+`
+const TitleRow = styled.div`
+  align-items: center;
+  display: flex;
+  justify-content: flex-start;
+  margin-bottom: 8px;
+`
+const TitleText = styled.div`
+  color: black;
+  font-weight: 600;
+  font-size: 16px;
+  line-height: 20px;
+  margin: 0px 12px;
+`
+const Wrapper = styled.div`
+  background-color: ${({ theme }) => theme.yellow3};
+  border-radius: 12px;
+  bottom: 60px;
+  display: none;
+  max-width: 348px;
+  padding: 16px 20px;
+  position: absolute;
+  right: 16px;
+  @media screen and (min-width: ${MEDIA_WIDTHS.upToMedium}px) {
+    display: block;
+  }
+`
+
+export function ChainConnectivityWarning() {
+  return (
+    <Wrapper>
+      <TitleRow>
+        <CautionIcon />
+        <TitleText>
+          <Trans>Network Downtime Warning</Trans>
+        </TitleText>
+      </TitleRow>
+      <BodyRow>
+        <Trans>
+          The Optimism network is down right now, or you may have lost connection to the sequencer. Check Optimism
+          network status
+        </Trans>{' '}
+        <Link href="">
+          <Trans>here.</Trans>
+        </Link>
+      </BodyRow>
+    </Wrapper>
+  )
+}
--- a/src/components/Header/ChainConnectivityWarning.tsx
+++ b/src/components/Header/ChainConnectivityWarning.tsx
@@ -1,4 +1,6 @@
 import { Trans } from '@lingui/macro'
+import { CHAIN_INFO, L2ChainInfo } from 'constants/chains'
+import { useActiveWeb3React } from 'hooks/web3'
 import { AlertOctagon } from 'react-feather'
 import styled from 'styled-components/macro'
 import { ExternalLink, MEDIA_WIDTHS } from 'theme'
@@ -42,6 +44,10 @@ const Wrapper = styled.div`
 `
 
 export function ChainConnectivityWarning() {
+  const { chainId } = useActiveWeb3React()
+  const info = CHAIN_INFO[chainId || '1']
+  const label = info?.label
+
   return (
     <Wrapper>
       <TitleRow>
@@ -51,13 +57,15 @@ export function ChainConnectivityWarning() {
         </TitleText>
       </TitleRow>
       <BodyRow>
-        <Trans>
-          The Optimism network is down right now, or you may have lost connection to the sequencer. Check Optimism
-          network status
-        </Trans>{' '}
-        <Link href="">
-          <Trans>here.</Trans>
-        </Link>
+        <Trans>{label} may be down right now, or you may have lost your network connection.</Trans>{' '}
+        {(info as L2ChainInfo).statusPage !== undefined && (
+          <span>
+            <Trans>Check network status</Trans>{' '}
+            <Link href={(info as L2ChainInfo).statusPage || ''}>
+              <Trans>here.</Trans>
+            </Link>
+          </span>
+        )}
       </BodyRow>
     </Wrapper>
   )
--- a/src/constants/chains.ts
+++ b/src/constants/chains.ts
@@ -48,7 +48,7 @@ export const L2_CHAIN_IDS = [
 export type SupportedL2ChainId = typeof L2_CHAIN_IDS[number]
 
 interface L1ChainInfo {
-  readonly blockWaitMSBeforeWarning?: number
+  readonly blockWaitMsBeforeWarning?: number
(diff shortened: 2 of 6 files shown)


src/constants/chains.ts
```diff
@@ -110,19 +112,23 @@ export const CHAIN_INFO: ChainInfo = {
     label: 'Görli',
   },
   [SupportedChainId.OPTIMISM]: {
+    blockWaitSecondsBeforeWarning: 10 * 60, // 10 minutes
```
justindomingue, line 115: Can use 

```
ms`10m`
```

justindomingue: Actually the units may not match

src/state/application/updater.ts
```diff
@@ -42,6 +48,28 @@ export default function Updater(): null {
     [chainId, setState]
   )
 
+  const connectivityWarningActive = useAppSelector((state) => state.application.connectivityWarning)
+  const timeout = useRef<NodeJS.Timeout>()
+  useEffect(() => {
+    console.log('running healthcheck')
```
justindomingue, line 54: Can probably remove the console log

src/state/application/actions.ts
```diff
@@ -27,3 +27,4 @@ export const setOpenModal = createAction<ApplicationModal | null>('application/s
 export const addPopup =
   createAction<{ key?: string; removeAfterMs?: number | null; content: PopupContent }>('application/addPopup')
 export const removePopup = createAction<{ key: string }>('application/removePopup')
+export const setConnectivityWarning = createAction<{ warn: boolean }>('application/setConnectivityWarning')
```
justindomingue, line 30: I wonder if you should specify `BlockchainConnectivity` to make it clearer it's not Network connectivity it something else.

src/state/application/updater.ts
```diff
@@ -42,6 +48,27 @@ export default function Updater(): null {
     [chainId, setState]
   )
 
+  const chainConnectivityWarningActive = useAppSelector((state) => state.application.chainconnectivityWarning)
+  const timeout = useRef<NodeJS.Timeout>()
+  useEffect(() => {
+    const waitMSBeforeWarning =
+      (chainId ? CHAIN_INFO[chainId]?.blockWaitMSBeforeWarning : DEFAULT_SECONDS_BEFORE_WARNING_WAIT) ||
```
justindomingue, line 55: may be worth converting everything to use ms or s? I think you'd need to multiple `DEFAULT_SECONDS_BEFORE_WARNING_WAIT` by `1000` here?

src/state/application/updater.ts
```diff
@@ -42,6 +48,27 @@ export default function Updater(): null {
     [chainId, setState]
   )
 
+  const chainConnectivityWarningActive = useAppSelector((state) => state.application.chainconnectivityWarning)
```
justindomingue, line 51: consider moving this to its own custom hooks similar to `useQueryCacheInvalidator` for readability.

src/state/application/updater.ts
```diff
@@ -20,6 +22,40 @@ function useQueryCacheInvalidator() {
   }, [chainId, dispatch])
 }
 
+const NETWORK_HEALTH_CHECK_MS = ms`15s`
+const DEFAULT_MS_BEFORE_WARNING_WAIT = ms`10m`
+interface UseBlockWarningTimerArgs {
```
justindomingue, line 27: nit. new line?

src/state/application/updater.ts
```diff
@@ -20,6 +22,40 @@ function useQueryCacheInvalidator() {
...
+  chainId: number | undefined
+  dispatch: (action: any) => void
```
justindomingue, line 29: does `dispatch: typeof dispatch` works?

src/state/application/updater.ts
```diff
@@ -20,6 +22,40 @@ function useQueryCacheInvalidator() {
...
+  msSinceLastBlock: number
+  setMsSinceLastBlock: (n: number) => void
+}
+
+function useBlockWarningTimer({ chainId, dispatch, msSinceLastBlock, setMsSinceLastBlock }: UseBlockWarningTimerArgs) {
+  const chainConnectivityWarningActive = useAppSelector((state) => state.application.chainConnectivityWarning)
+  const timeout = useRef<NodeJS.Timeout>()
+  useEffect(() => {
+    const waitMsBeforeWarning =
+      (chainId ? CHAIN_INFO[chainId]?.blockWaitMsBeforeWarning : DEFAULT_MS_BEFORE_WARNING_WAIT) ||
```
justindomingue, line 39: `??` is preferred over `||` here I think in case `CHAIN_INFO[chainId]?.blockWaitMsBeforeWarning` is 0.

src/state/application/updater.ts
```diff
@@ -20,6 +22,40 @@ function useQueryCacheInvalidator() {
...
+      DEFAULT_MS_BEFORE_WARNING_WAIT
+
+    timeout.current = setTimeout(() => {
+      setMsSinceLastBlock(NETWORK_HEALTH_CHECK_MS + msSinceLastBlock)
+      if (msSinceLastBlock > waitMsBeforeWarning) {
+        dispatch(setChainConnectivityWarning({ warn: true }))
```
justindomingue, line 45: why are you passing the dispatch as an arg? Isn't it always the same `appDispatch`?

src/state/application/updater.ts
justindomingue: I think I would just `useAppDispatch` here

src/state/application/updater.ts
```diff
@@ -33,11 +69,13 @@ export default function Updater(): null {
 
   useQueryCacheInvalidator()
 
+  const [msSinceLastBlock, setMsSinceLastBlock] = useState(0)
```
justindomingue, line 72: nit. readability. I usually prefer putting `useState`s together

src/state/application/updater.ts
```diff
@@ -33,11 +69,13 @@ export default function Updater(): null {
...
   const blockNumberCallback = useCallback(
     (blockNumber: number) => {
       setState((state) => {
         if (chainId === state.chainId) {
           if (typeof state.blockNumber !== 'number') return { chainId, blockNumber }
+          setMsSinceLastBlock(0)
```
justindomingue, line 78: i wonder if we're overloading the meaning of `blockNumberCallback` here. Can you not subscribe to the redux store (`useBlockNumber`) and update `msSinceLastBlock` in there? I think that's a more 'redux-y' way of doing things.

src/state/application/updater.ts
```diff
@@ -20,6 +22,40 @@ function useQueryCacheInvalidator() {
   }, [chainId, dispatch])
 }
 
+const NETWORK_HEALTH_CHECK_MS = ms`15s`
+const DEFAULT_MS_BEFORE_WARNING_WAIT = ms`10m`
```
justindomingue, line 26: nit. not clear that `NETWORK_HEALTH_CHECK_MS` and `DEFAULT_MS_BEFORE_WARNING_WAIT` are both referring to network connectivity.

I think it would be cleaner to create a custom hook `useNetworkConnectivityWarning` that has all this wok and just call it from updater. Similar to `useQueryCacheInvalidator above.

src/components/Header/ChainConnectivityWarning.tsx
```diff
@@ -0,0 +1,72 @@
...
+  position: absolute;
+  right: 16px;
+  @media screen and (min-width: ${MEDIA_WIDTHS.upToMedium}px) {
+    display: block;
+  }
+`
+
+export function ChainConnectivityWarning() {
+  const { chainId } = useActiveWeb3React()
+  const info = CHAIN_INFO[chainId || '1']
```
justindomingue, line 48: use `??` in case `chaindId` is `0`

src/components/Header/ChainConnectivityWarning.tsx
```diff
@@ -0,0 +1,72 @@
...
+const Wrapper = styled.div`
+  background-color: ${({ theme }) => theme.yellow3};
+  border-radius: 12px;
+  bottom: 60px;
+  display: none;
+  max-width: 348px;
+  padding: 16px 20px;
+  position: absolute;
+  right: 16px;
+  @media screen and (min-width: ${MEDIA_WIDTHS.upToMedium}px) {
```
justindomingue, line 41:
```
    display: block;
  `};
```

is probably cleaner

src/components/Header/ChainConnectivityWarning.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { CHAIN_INFO, L2ChainInfo } from 'constants/chains'
+import { useActiveWeb3React } from 'hooks/web3'
+import { AlertOctagon } from 'react-feather'
+import styled from 'styled-components/macro'
+import { ExternalLink, MEDIA_WIDTHS } from 'theme'
+
+const BodyRow = styled.div`
+  color: black;
+  font-size: 12px;
+`
```
justindomingue, line 11: nit. styling -- idk that we actually follow this but I've always seen a newline separating styled components in our repo

src/components/Header/ChainConnectivityWarning.tsx
```diff
@@ -0,0 +1,72 @@
...
+const CautionIcon = styled(AlertOctagon)`
+  color: black;
+`
+const Link = styled(ExternalLink)`
+  color: black;
```
justindomingue, line 16: prefer `theme.black` so this color is tracked with our theme. It's still defined as `black` for both light and dark but if we ever change it later we'll know

src/components/Header/ChainConnectivityWarning.tsx
justindomingue: ditto below

src/components/Header/ChainConnectivityWarning.tsx
JFrankfurt: I'm intending to use the `min-width` behavior here instead of `max-width` in the theme. In general I think using min-width for query selectors is a better pattern.  Might be worth bringing up in eng review how to approach this. I don't like the pattern divergence, but in my ideal world we'd use `theme.mediaWidth.greaterThanMedium` or something instead.

src/components/Header/ChainConnectivityWarning.tsx
JFrankfurt: Always get tripped up by the name of `??` and assume it only works for `null` values, not undefined.

src/components/Header/ChainConnectivityWarning.tsx
ianlapham: Yeah we dont have a formal guideline on that but thats the standard ive been following

src/components/Header/ChainConnectivityWarning.tsx
ianlapham: I actually do like the upToMedium better 

@cmcewen mentioned in a chat we had that maybve we should formalize / agree on how we do certain thinsg like this (media queries, layering, even ordering of css properties)

justindomingue: I think we should use the theme-defined lambdas for consistency. Can you not use max-width and flip the css inside the block?

src/state/application/updater.ts
```diff
@@ -20,6 +22,40 @@ function useQueryCacheInvalidator() {
...
+interface UseBlockWarningTimerArgs {
+  chainId: number | undefined
+  dispatch: (action: any) => void
+  msSinceLastBlock: number
+  setMsSinceLastBlock: (n: number) => void
+}
+
+function useBlockWarningTimer({ chainId, dispatch, msSinceLastBlock, setMsSinceLastBlock }: UseBlockWarningTimerArgs) {
+  const chainConnectivityWarningActive = useAppSelector((state) => state.application.chainConnectivityWarning)
+  const timeout = useRef<NodeJS.Timeout>()
```
justindomingue, line 38:
```suggestion
  const timeout = useRef<NodeJS.Timeout>()
  
  const [msSinceLastBlock, setMsSinceLastBlock] = useState(0)
  const blockNumber = useBlockNumber()
    
  useEffect(() => setMsSinceLastBlock(0), [blockNumber])
```

JFrankfurt: woah, this is a cool feature.

src/components/Header/ChainConnectivityWarning.tsx
JFrankfurt: Yeah we def *can*, but going smaller→larger allows for progressive overwrites without requiring "erasing" old values that aren't relevant in the next larger screen context. The need to set css back to default values leads to more bugs in my experience (and more code).

JFrankfurt: There are obv exceptions, but I think the mobile-first (min-width) pattern often ends up being cleaner.

src/components/Header/ChainConnectivityWarning.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <BodyRow>
+        <Trans>{label} may be down right now, or you may have lost your network connection.</Trans>{' '}
+        {(info as L2ChainInfo).statusPage !== undefined && (
+          <span>
+            <Trans>Check network status</Trans>{' '}
+            <Link href={(info as L2ChainInfo).statusPage || ''}>
+              <Trans>here.</Trans>
+            </Link>
+          </span>
+        )}
```
JFrankfurt, line 68: @willhennessy current copy

src/state/application/updater.ts
```diff
@@ -20,17 +23,60 @@ function useQueryCacheInvalidator() {
   }, [chainId, dispatch])
 }
 
+function useBlockWarningTimer() {
+  const NETWORK_HEALTH_CHECK_MS = ms`15s`
```
justindomingue, line 27: should these be outside of the hook so they don't get redefined each render loop?

justindomingue (approved):

src/components/Header/ChainConnectivityWarning.tsx
willhennessy: lgtm

willhennessy (approved):

src/components/Header/ChainConnectivityWarning.tsx
callil: Sorry I didn't make it to review, once concern I have here is the primacy of the "network down" language, when we may see this often for just regular wifi disconnects for instance. @willhennessy

willhennessy:
yeah fair point. Let's flip it :
https://www.notion.so/uniswaplabs/update-network-warning-copy-54c4e9f4b0dc464eb8e557c667ac9553

PR #2138, opened 2021-08-04
fix: buying fee on transfer token when entering exactEth amounts

JamJomJim:
Issue: https://github.com/Uniswap/uniswap-interface/issues/2137

### DIFF
--- a/src/hooks/useSwapCallback.ts
+++ b/src/hooks/useSwapCallback.ts
@@ -302,8 +302,7 @@ export function useSwapCallback(
 
         // a successful estimation is a bignumber gas estimate and the next call is also a bignumber gas estimate
         let bestCallOption: SuccessfulCall | SwapCallEstimate | undefined = estimatedCalls.find(
-          (el, ix, list): el is SuccessfulCall =>
-            'gasEstimate' in el && (ix === list.length - 1 || 'gasEstimate' in list[ix + 1])
+          (el): el is SuccessfulCall => 'gasEstimate' in el
         )
 
         // check if any calls errored with a recognizable error


src/hooks/useSwapCallback.ts
```diff
@@ -302,8 +302,7 @@ export function useSwapCallback(
 
         // a successful estimation is a bignumber gas estimate and the next call is also a bignumber gas estimate
         let bestCallOption: SuccessfulCall | SwapCallEstimate | undefined = estimatedCalls.find(
-          (el, ix, list): el is SuccessfulCall =>
-            'gasEstimate' in el && (ix === list.length - 1 || 'gasEstimate' in list[ix + 1])
+          (el): el is SuccessfulCall => 'gasEstimate' in el
```
moodysalem, line 305: i think there's a reason for this, but unfortunately the comment on line 303 is not sufficient. my guess is we don't want to allow you to buy a fot that we think will give you sufficient output but it won't with given slippage tolerance

JamJomJim: I just looked back through the history, and this section was added in commit `0f91af1df26ef9456e5745557276fe8ad9118b50`. Sorry, I'm not sure if I can link to a previous PR's diff, so here's the previous code
```       
// if only 1 method exists, either:
// a) the token is doing something weird not related to FoT (e.g. enforcing a whitelist)
// b) the token is FoT and the user specified an exact output, which is not allowed
        if (swapMethods.length === 1) {
          throw Error(
            `An error occurred. If either of the tokens you're swapping take a fee on transfer, you must specify an exact input amount.`
          )
        }```
Clearly a lot has changed since then, but I'm guessing the reasoning for the error is the same? That the swap would somehow need to know the fee % to guarantee an exact output (which it can't know). If this can't be changed, would it be possible to add this more meaningful error back in? That way people will know they need to specify an exact input, instead of just having the swap fail for no clear reason.

tinaszheng:
sorry going to close this since is a change in our deprecated swap flow!

PR #2140, opened 2021-08-05
fix(config): update tsconfig, drop ie11/dead browsers support

sambacha:
### Changes

- [x] add `tslib` (or alt. remove `downlevelIteration` from tsconfig
- [x] update browserslist to explicitly not support IE11

#### 1. enable `importHelpers` 

> importHelpers flag is on, these helper functions are instead imported from the tslib module
> (...)
>Turning on `downlevelIteration` and `importHelpers` is still false - [typescript documentation, tsconfig](https://www.typescriptlang.org/tsconfig#importHelpers)

##### 2. update `browserslist` configuration

```bash
$ npx browserslist  ">0.2%, not dead, not op_mini all"
and_chr 90
and_ff 87
and_uc 12.12
android 4.4.3-4.4.4
chrome 90
chrome 89
chrome 88
chrome 87
chrome 86
chrome 85
chrome 49
edge 90
edge 89
firefox 88
firefox 87
ie 11
ios_saf 14.0-14.4
ios_saf 13.4-13.7
ios_saf 12.2-12.4
opera 75
opera 73
safari 14
safari 13.1
```

##### Updated browserslist configuration with explicit no IE11

```json
  "browserslist": {
    "production": [
      ">0.3%",
      "not dead",
      "not op_mini all",
      "not IE > 0",
      "not samsung 4",
      "not and_uc 12.12"
    ]
```

###### updated result:

```bash
$ npx browserslist
and_chr 90
chrome 90
chrome 89
chrome 88
chrome 87
edge 90
edge 89
firefox 88
firefox 87
ios_saf 14.0-14.4
ios_saf 13.4-13.7
opera 75
opera 73
safari 14
safari 13.1
samsung 14.0
samsung 13.0
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,6 +112,7 @@
     "start-server-and-test": "^1.11.0",
     "styled-components": "^5.3.0",
     "styled-system": "^5.1.5",
+    "tslib": "^2.3.0",
     "typechain": "^5.0.0",
     "typescript": "^4.2.3",
     "ua-parser-js": "^0.7.28",
@@ -148,9 +149,12 @@
   },
   "browserslist": {
     "production": [
-      ">0.2%",
+      ">0.3%",
       "not dead",
-      "not op_mini all"
+      "not op_mini all",
+      "not IE > 0",
+      "not samsung 4",
+      "not and_uc 12.12"
     ],
     "development": [
       "last 1 chrome version",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -18,6 +18,7 @@
     "noImplicitReturns": true,
     "moduleResolution": "node",
     "resolveJsonModule": true,
+    "importHelpers": true,
     "isolatedModules": true,
     "jsx": "react-jsx",
     "downlevelIteration": true,
--- a/yarn.lock
+++ b/yarn.lock
@@ -18381,7 +18381,7 @@ tslib@^2, tslib@^2.0.0, tslib@~2.2.0:
   resolved "https://registry.npmjs.org/tslib/-/tslib-2.2.0.tgz"
   integrity sha512-gS9GVHRU+RGn5KQM2rllAlR3dU6m7AcpJKdtH8gFvQiC4Otgk98XnmMU+nZenHt/+VhnBPWwgrJsyrdcw6i23w==
 
-tslib@^2.0.3, tslib@^2.1.0, tslib@~2.3.0:
+tslib@^2.0.3, tslib@^2.1.0, tslib@^2.3.0, tslib@~2.3.0:
   version "2.3.0"
   resolved "https://registry.yarnpkg.com/tslib/-/tslib-2.3.0.tgz#803b8cdab3e12ba581a4ca41c8839bbb0dacb09e"
   integrity sha512-N82ooyxVNm6h1riLCoyS9e3fuJ3AMG2zIZs2Gd1ATcSFjSA23Q0fzjjZeh0jbJvWVDZ0cJT8yaNNaaXHzueNjg==


moodysalem:
can you explain the motivation or link an issue?

JFrankfurt:
[downlevelIteration](https://www.typescriptlang.org/tsconfig#downlevelIteration)
[importHelpers](https://www.typescriptlang.org/tsconfig#importHelpers)

First I've heard of these. Curious how much of an effect they'll have.

sambacha:
> can you explain the motivation or link an issue?

Thanks @moodysalem for asking the obvious, my reasoning is a lot clearer now. Apologies for the delayed response!

## Motivation

#### Performance and Benchmarking

As of right now, Uniswap's frontend does not have an explicit browser support policy/targeted devices for support list (that I know of). Reducing loading time and increasing response time directly affects end user behavior/attention span, so reducing waiting time (to me) is important.

With an explicit policy it becomes much easier to know what can and can not be adjusted/replaced/improved etc to get increased performance (e.g. reducing time to first contentful paint, reducing bundle size on deployment, etc). 

So, we can establish that there is:

1: No need to support old browsers, seeing as how Metamask is dependent on modern environments
2: An obvious security risk w/ supporting such browsers
3: Unambiguously degrades application performance

This is the motivation at least for my usage. 


> [downlevelIteration](https://www.typescriptlang.org/tsconfig#downlevelIteration)
> [importHelpers](https://www.typescriptlang.org/tsconfig#importHelpers)
> 
> First I've heard of these. Curious how much of an effect they'll have.

@JFrankfurt to your question,  downlevelIteration is just the TypeScript term for transpiling to an older version of JavaScript (es5).


## TL:DR

Basically trying to figure out what browsers Uniswap is wanting to be compatible with, which would inform the presented decisions above 

## Action Items:
es2017+, remove downIteration in tsconfig, remove tslib, keep browserslist
*or*
es5/6: keep tslib and downIteration enabled. *note: modify browserslist for iOS13/Safari*

sambacha:
let me know if you have any questions @moodysalem / @JFrankfurt

package.json
```diff
@@ -110,6 +110,8 @@
     "serve": "^11.3.2",
     "start-server-and-test": "^1.11.0",
     "styled-components": "^5.3.0",
+    "styled-system": "^5.1.5",
```
zzmp, line 113: Why did you add styled-system?

package.json
```diff
@@ -110,6 +110,8 @@
...
+    "tslib": "^2.3.0",
```
zzmp, line 114: Is tslib required to be installed explicitly? Is it not enough to just set `importHelpers: true` in the tsconfig?

package.json
```diff
@@ -144,9 +146,12 @@
   },
   "browserslist": {
     "production": [
-      ">0.2%",
+      ">0.3%",
```
zzmp, line 149: See https://github.com/Uniswap/interface/issues/2367. This should be resolved before committing a change to browserslist. Would you mind commenting with your rationale in that issue?

tsconfig.json
```diff
@@ -19,6 +19,7 @@
     "useUnknownInCatchVariables": false,
     "moduleResolution": "node",
     "resolveJsonModule": true,
+    "importHelpers": true,
```
zzmp, line 22: Just to clarify: this reduces code duplication, and nothing else? How is it related to browser support?

If this can be turned on without modifying browserslist, can you break it out into a separate very small PR, so we can get it in before resolving the final browserslist?

yarn.lock
```diff
@@ -18492,7 +18492,7 @@ tslib@^2, tslib@^2.0.0, tslib@~2.2.0:
   resolved "https://registry.npmjs.org/tslib/-/tslib-2.2.0.tgz"
   integrity sha512-gS9GVHRU+RGn5KQM2rllAlR3dU6m7AcpJKdtH8gFvQiC4Otgk98XnmMU+nZenHt/+VhnBPWwgrJsyrdcw6i23w==
 
-tslib@^2.0.3, tslib@^2.1.0, tslib@~2.3.0:
+tslib@^2.0.3, tslib@^2.1.0, tslib@^2.3.0, tslib@~2.3.0:
```
zzmp, line 18495: That we already have tslib installed makes me think that we don't need it declared explicitly in the package.json.

package.json
zzmp: You should be able to use tslib through typescript, so you can remove this from package.json, and revert the change to the yarn.lock.

BigsteppA1 (approved):

PR #2144, opened 2021-08-06
fix: add chainid to GA calls

JFrankfurt:


justindomingue:
This works but i wonder if instead we should avoid the duplication.. something like extract `ReactGA.event` to a hook

`const logEvent = useReactGA({ ... })`

willhennessy (commented):
Looks like we simply insert {value: chainID} into every log.  I'm surprised there is not a dedicated mapping like {chain: ID} or a key-value pair like {label: "chain-id", value: chaindID}

Google Analytics seems to treat the value as an integer that gets summed up by Category, Action, or Label. https://analytics.google.com/analytics/web/#/report/content-event-events/a128182339w232099914p218018572/explorer-segmentExplorer.segmentId=analytics.eventCategory&explorer-table.plotKeys=%5B%5D&explorer-table.rowCount=500

I'm not sure this is the right field?

JFrankfurt:
Yeah, it def doesn't feel right to me, but I timeboxed finding the "right" way to do this and ended up here. 

Totally open to new approaches. This one leaves a lot to be desired.

willhennessy:
nice, definitely respect the time box. This is definitely a lower priority than your active projects. Let's pause this until we have time to do it the right way

I think a Custom Dimension might do the trick? https://support.google.com/analytics/answer/2709828

JFrankfurt:
Thanks, this looks right. If you have a free second, would you mind taking
a look at the prerequisites section at that link and making sure our
analytics account is set up to support this? It reads like gibberish to me
rn.

willhennessy:
Alright I followed these [instructions](https://support.google.com/analytics/answer/2709829#zippy=%2Cin-this-article) to set up a Custom Dimension for you called "Chain ID".  I chose [Hit Level Scope](https://support.google.com/analytics/answer/2709828#example&zippy=%2Cin-this-article)

I think you should pick it up with the "Modify your tracking code" [section](https://support.google.com/analytics/answer/2709829#zippy=%2Cin-this-article)

This code example suggests what you should do next
Screen Shot 2021-09-16 at 5 57 47 PM

And here is example configuration code

```Created Custom Dimension
Example Codes for This Dimension
Copy the following code snippet for your platform. DO NOT FORGET to replace dimensionValue with your own.

JavaScript (gtag.js)
For instructions on how to setup custom dimensions using gtag.js, please refer to the gtag.js developer documentation.

**JavaScript (Only works for Universal Analytics properties)**
var dimensionValue = 'SOME_DIMENSION_VALUE';
ga('set', 'dimension1', dimensionValue);

** ^ I think this is what we need **

**Android SDK**
String dimensionValue = "SOME_DIMENSION_VALUE";
tracker.set(Fields.customDimension(1), dimensionValue);

**iOS SDK**
NSString *dimensionValue = @"SOME_DIMENSION_VALUE";
[tracker set:[GAIFields customDimensionForIndex:1] value:dimensionValue];
Done ```

JFrankfurt:
superseded by #2440

PR #2147, opened 2021-08-07
Update chains.ts

babydingo:


JFrankfurt:
BSC is not a supported chain

legend620:
what will be the best approach to support BSC in uniswap interface?

PR #2150, opened 2021-08-09
Only one China, Please change the Language

dsky1990:


dsky1990:
@moodysalem 
It’s not China and Taiwan. Taiwan belongs to China. It’s Chinese Taiwan. You should change it to`中文简体` and `中文翻译`

PR #2204, opened 2021-08-12
feat(i18n): eagerly fetch active locale

zzmp:
Eagerly loads the active locale asset as soon as it is known, instead of waiting for React bootstrapping.

Removing the asset load from the critical path appx 100ms on pageload, as loading the asset blocks text rendering, and TTFB is appx 100msn (for unknown reasons).

### DIFF
--- a/src/hooks/useActiveLocale.ts
+++ b/src/hooks/useActiveLocale.ts
@@ -2,7 +2,7 @@ import { DEFAULT_LOCALE, SupportedLocale, SUPPORTED_LOCALES } from 'constants/lo
 import { parse } from 'qs'
 import { useEffect, useMemo } from 'react'
 import { useUserLocale, useUserLocaleManager } from 'state/user/hooks'
-import { initialState } from 'state'
+import store from 'state'
 import useParsedQueryString from './useParsedQueryString'
 
 /**
@@ -32,7 +32,7 @@ function urlLocale(): SupportedLocale | undefined {
  * Returns the supported locale read from the user state (redux)
  */
 function userLocale(): SupportedLocale | undefined {
-  return initialState.user?.userLocale ?? undefined
+  return store.getState().user.userLocale ?? undefined
 }
 
 /**
--- a/src/state/index.ts
+++ b/src/state/index.ts
@@ -18,8 +18,6 @@ import { routingApi } from './routing/slice'
 
 const PERSISTED_KEYS: string[] = ['user', 'transactions', 'lists']
 
-const preloadedState = load({ states: PERSISTED_KEYS, disableWarnings: process.env.NODE_ENV === 'test' })
-
 const store = configureStore({
   reducer: {
     application,
@@ -41,7 +39,7 @@ const store = configureStore({
       .concat(dataApi.middleware)
       .concat(routingApi.middleware)
       .concat(save({ states: PERSISTED_KEYS, debounce: 1000 })),
-  preloadedState,
+  preloadedState: load({ states: PERSISTED_KEYS, disableWarnings: process.env.NODE_ENV === 'test' }),
 })
 
 store.dispatch(updateVersion())
@@ -50,5 +48,3 @@ export default store
 
 export type AppState = ReturnType<typeof store.getState>
 export type AppDispatch = typeof store.dispatch
-
-export const initialState: Partial<AppState> = preloadedState


src/hooks/useActiveLocale.ts
```diff
@@ -14,6 +16,22 @@ function parseLocale(maybeSupportedLocale: string): SupportedLocale | undefined
...
+  const search = location.search || location.hash.match(/\?.*/)?.[0]
+  const parsed = search && search.length > 1 ? parse(search, { parseArrays: false, ignoreQueryPrefix: true }) : {}
+  return typeof parsed.lng === 'string' ? parseLocale(parsed.lng) : undefined
+}
+
+/**
+ * Returns the supported locale read from the user state (redux)
+ */
+function userLocale(): SupportedLocale | undefined {
+  return initialState.user?.userLocale ?? undefined
```
justindomingue, line 32: won't that always be `null`?

zzmp: It may be set, if you've explicitly set a locale in the past.

For example, if you navigate to `example.com?lng=fr-FR`, then `fr-FR` will be set in your `localStorage` (through redux), and this will be in your `initialState` (which is loaded from localStorage as part of initializing redux).

src/hooks/useActiveLocale.ts
```diff
@@ -14,6 +16,22 @@ function parseLocale(maybeSupportedLocale: string): SupportedLocale | undefined
   )
 }
 
+/**
+ * Returns the supported locale read from the url query parameter (lng=)
+ */
+function urlLocale(): SupportedLocale | undefined {
+  const search = location.search || location.hash.match(/\?.*/)?.[0]
```
zzmp, line 23: `location.hash.match(/\?.*/)?.[0]` is necessary in case the query parameter is placed after a hash. IDK if this is spec-compliant, but it is definitely done by our current codebase, so it must be supported.

src/hooks/useActiveLocale.ts
```diff
@@ -14,6 +16,25 @@ function parseLocale(maybeSupportedLocale: string): SupportedLocale | undefined
   )
 }
 
+/**
+ * Returns the supported locale read from the url query parameter (lng=)
+ */
+function urlLocale(): SupportedLocale | undefined {
+  const search =
+    location.search ||
```
justindomingue, line 24: this feel hacky. Can you not use react-router-dom?

src/i18n.tsx
```diff
@@ -106,3 +106,9 @@ export function LanguageProvider({ children }: { children: ReactNode }) {
     </I18nProvider>
   )
 }
+
+// eagerly import the locale to unblock pageload
+;(() => {
+  const locale = initialLocale()
+  import(`@lingui/loader!./locales/${locale}.po`)
```
justindomingue, line 113: couldn't this be done inside `LanguageProvider` so we can re-use hooks?

zzmp: This IIFE executes 65ms sooner than the LanguageProvider function component on a production build, so it must be here to be effective. It's definitely a niche optimization.

src/hooks/useActiveLocale.ts
zzmp: This is called outside of React, so I don't think any react-router-dom's are yet available.

zzmp: nvm, just found https://reactrouter.com/web/api/matchPath, will update.

zzmp: I misread - matchPath will not suffice here.

I don't see how to use react-router-dom here, as it is outside of the React render cycle. Is there an underlying parser that react-router-dom uses that we can also rely on, to avoid this?

zzmp: We only use the HashRouter, so I've updated this to only check the hash, and parse the query from there. That's fairly straightforward, and similar to what we do in other query-parsing scenarios.

justindomingue: I just don't really like the precedent this sets. I don't think we should be hacking the tools we use to save 10ms. e.g. what if we move away from react-router? This code would start failing and we have no test coverage either.

Could we consider wrapping this code with `withRouter` to be able to grab the location from the react-router `location` object?

PR #2214, opened 2021-08-18
feat: allow users to turn on frontrunning protection which will send signed transactions to a different rpc

moodysalem:
this is the work starting for mvp 'flashbots' integration, but proper integration with flashbots requires us to create and sign bundles on a server with a secret private key used to track reputation, as well as resubmitting every block

fixes https://github.com/Uniswap/uniswap-interface/issues/2202


moodysalem:
Preferred approach is to allow users to switch the RPC endpoint that their wallet uses via the interface

PR #2225, opened 2021-08-19
feat: display a leading 0 on fractional numerical input

zzmp:
Prepends a 0 when a user enters a fractional number


moodysalem:
pretty strange behavior imo

zzmp:
This is in response to the interface auto-populated the converted amount using this format. I think it looks off when the auto-populated format does not match the user-input format.

image

callil:
I like this from a UX perspective tbh, feels like it makes the input more immediately legible and match how the input gets displayed in the review screen for instance.

PR #2228, opened 2021-08-20
fix: older browser suppport

circle-hotaru:
polyfill Object.fromEntries() to support older browser.

Object.fromEntries is not a function

issue: https://github.com/Uniswap/uniswap-interface/issues/2227


moodysalem:
this is supported by pretty much every modern browser https://caniuse.com/mdn-javascript_builtins_object_fromentries

samlin425:
Hi, @moodysalem ， this issue has happened on android 8 or 9. Unfortunately, 20% of imToken users are still using those old browsers, so we need to support it with this bug fix.

src/constants/routing.ts
```diff
@@ -1,6 +1,7 @@
 // a list of tokens by chain
 import { Currency, Token } from '@uniswap/sdk-core'
 import { SupportedChainId } from './chains'
+import 'polyfill-object.fromentries'
```
zzmp, line 4: I'd rather see this in a polyfill file, as it doesn't import anything obviously used in this file. That way, it will be easier to track polyfills in the future.

circle-hotaru: I have create a polyfill file in `/src` and import it in `index.tsx` now. Is this appropriate ?

src/polyfills.js
```diff
@@ -0,0 +1 @@
+import 'polyfill-object.fromentries'
```
justindomingue, line 1: can you make it typescript?

circle-hotaru: Do you mean changing polyfills.js to polyfills.ts?

justindomingue: yes 😄

circle-hotaru: ok, It has changed😃

justindomingue (approved):

justindomingue:
LGTM. Will let @zzmp and @moodysalem merge when they approve.

zzmp (approved):

PR #2230, opened 2021-08-21
chore: rm redux-devtools-extension

zzmp:
redux-devtools-extension integration is already provided through @reduxjs/toolkit.


justindomingue (approved):

moodysalem:
devtools still work on the deployment preview so it looks good to me, but strange that it gets removed from yarn.lock

PR #2231, opened 2021-08-21
chore: use native flatMap

zzmp:
one less dependency


justindomingue (approved):
there's no existing test coverage for this, right?

NoahZinsmeister:
`flatMap` has somewhat less-than-ideal support across browsers: https://caniuse.com/?search=flatMap

that being said, at the moment we'd be introducing `flat` in this PR: https://github.com/Uniswap/uniswap-interface/pull/2271

PR #2245, opened 2021-08-27
fix: do not paint currency selector until initialized

zzmp:
Avoids a flash of "Select a token" on the input currency selector as the App initializes (and/or the output currency selector, if an inputCurrency/outputCurrency parameter is passed).

### DIFF
--- a/src/pages/Swap/index.tsx
+++ b/src/pages/Swap/index.tsx
@@ -345,7 +345,7 @@ export default function Swap({ history }: RouteComponentProps) {
     [onCurrencySelection]
   )
 
-  const swapIsUnsupported = useIsSwapUnsupported(currencies?.INPUT, currencies?.OUTPUT)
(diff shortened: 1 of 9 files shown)


src/pages/Swap/index.tsx
```diff
@@ -345,7 +345,7 @@ export default function Swap({ history }: RouteComponentProps) {
     [onCurrencySelection]
   )
 
-  const swapIsUnsupported = useIsSwapUnsupported(currencies?.INPUT, currencies?.OUTPUT)
+  const swapIsUnsupported = useIsSwapUnsupported(currencies[Field.INPUT], currencies[Field.OUTPUT])
```
ianlapham, line 348: any idea why we werent doing this before?

zzmp: no clue, but it's done this way earlier in the file. probably just multiple editors over time.

PR #2247, opened 2021-08-27
fix(governance): special case bravo parsing

zzmp:
Replace invalid UTF-8 characters in proposal descriptions with the replacement character to avoid crashes on future proposals with invalid UTF-8.

This is motivated by the Bravo proposal, which (1) has invalid characters, as U-2018 and U-2026 codepoints are reversed; and (2) omits newlines, preventing correct markdown parsing. This PR fixes the proposal programmatically instead of using a static constant (see e1f37b0ec61cb43552aeea75af249031cd3fc759) so that the data still originates on-chain, using guidance from ethers-io/ethers.js#1894.


NoahZinsmeister (approved):
this looks great, thanks @zzmp :)

PR #2264, opened 2021-08-30
Fix z-index bug which prevented modal click

### LINKED ISSUE #2131: “Read More about Unsupported Assets
**Bug Description**
Selecting iXMR shows a warning message about unsupported asset

Link at bottom says "read more about unsupported assets" but tapping it from touch screen doesn't produce anything more to read.

**Steps to Reproduce**

1. Select wBTC
2. Select iXMR
   ...

**Expected Behavior**
A clear and consice screen explaining how to pool liquidity assets of Monero and Bitcoin wrapped with your individual wallets should appear.

**Additional Context**
As of XMR v0.17.21 liquidity assets are supported in bisq p2p trades. No need to display "unsupported asset" message in the first place.

cmcewen:
- fixes #2131 

Before:
Screen Shot 2021-08-30 at 4 28 21 PM

After:
Screen Shot 2021-08-30 at 4 28 26 PM

Test link: https://app.uniswap.org/#/add/ETH/0x21cA39943E91d704678F5D00b6616650F066fD63/10000

### DIFF
--- a/src/components/swap/UnsupportedCurrencyFooter.tsx
+++ b/src/components/swap/UnsupportedCurrencyFooter.tsx
@@ -1,21 +1,23 @@
-import { useState } from 'react'
-import styled from 'styled-components/macro'
-import { TYPE, CloseIcon, ExternalLink } from 'theme'
+import { Trans } from '@lingui/macro'
+import { Currency, Token } from '@uniswap/sdk-core'
 import { ButtonEmpty } from 'components/Button'
-import Modal from 'components/Modal'
 import Card, { OutlineCard } from 'components/Card'
-import { RowBetween, AutoRow } from 'components/Row'
 import { AutoColumn } from 'components/Column'
 import CurrencyLogo from 'components/CurrencyLogo'
+import Modal from 'components/Modal'
+import { AutoRow, RowBetween } from 'components/Row'
 import { useActiveWeb3React } from 'hooks/web3'
-import { Currency, Token } from '@uniswap/sdk-core'
+import { useState } from 'react'
+import styled from 'styled-components/macro'
+import { CloseIcon, ExternalLink, TYPE, Z_INDEX } from 'theme'
 import { useUnsupportedTokens } from '../../hooks/Tokens'
 import { ExplorerDataType, getExplorerLink } from '../../utils/getExplorerLink'
-import { Trans } from '@lingui/macro'
 
 const DetailsFooter = styled.div<{ show: boolean }>`
   padding-top: calc(16px + 2rem);
   padding-bottom: 20px;
+  margin-left: auto;
+  margin-right: auto;
   margin-top: -2rem;
   width: 100%;
   max-width: 400px;
@@ -23,7 +25,7 @@ const DetailsFooter = styled.div<{ show: boolean }>`
   border-bottom-right-radius: 20px;
   color: ${({ theme }) => theme.text2};
   background-color: ${({ theme }) => theme.advancedBG};
-  z-index: -1;
+  z-index: ${Z_INDEX.zero};
 
   transform: ${({ show }) => (show ? 'translateY(0%)' : 'translateY(-100%)')};
   transition: transform 300ms ease-in-out;
@@ -100,7 +102,7 @@ export default function UnsupportedCurrencyFooter({
           </AutoColumn>
         </Card>
       </Modal>
-      <ButtonEmpty padding={'0'} onClick={() => setShowDetails(true)}>
+      <ButtonEmpty padding={'0'} style={{ textDecoration: 'none' }} onClick={() => setShowDetails(true)}>
         <TYPE.blue>
           <Trans>Read more about unsupported assets</Trans>
         </TYPE.blue>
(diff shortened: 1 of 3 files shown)


src/theme/index.tsx
```diff
@@ -20,6 +20,14 @@ export const MEDIA_WIDTHS = {
   upToLarge: 1280,
 }
 
+export enum Z_INDEX {
```
justindomingue, line 23: might be worth looking at existing uses of `z-index` (e.g. modal would be closer to 9999)

https://github.com/Uniswap/uniswap-interface/search?q=z-index

justindomingue: i like the idea of putting it there though

src/pages/AppBody.tsx
```diff
@@ -11,6 +12,7 @@ export const BodyWrapper = styled.main<{ margin?: string; maxWidth?: string }>`
     0px 24px 32px rgba(0, 0, 0, 0.01);
   border-radius: 24px;
   margin-top: 1rem;
+  z-index: ${Z_INDEX.content};
```
justindomingue, line 15: i think it would be cleaner to put `z-index` inside the `theme` object similar to color or media-queries so we don't have to import `Z_INDEX` from `theme` and instead be able to do `z-index: ${({theme}) => theme.zIndexContent`

justindomingue: see https://github.com/Uniswap/uniswap-interface/blob/56f8a7443cfa19c329f4d1e557dd8d4713dd76e5/src/theme/index.tsx#L107

src/theme/index.tsx
cmcewen: yeah something i'd like to discuss around CSS style guide but imo using too many z-index values just gets confusing. i feel like it's a bit like using `!important` - keeping things simple and constrained reduces complexity everywhere and then you don't need huge values

src/pages/AppBody.tsx
cmcewen: so i'm with you on color for sure, but i wonder when you would want media queries or z-index to be dynamic with the theme. using enums means they can be compiled into static values vs a function can just add more JS even if the value never changes

justindomingue: Really just for developer experience and consistency with how we do media queries. I hadn't really thought about compiled vs. runtime.

Probably makes sense to keep it static.

src/components/swap/UnsupportedCurrencyFooter.tsx
```diff
@@ -100,7 +102,7 @@ export default function UnsupportedCurrencyFooter({
           </AutoColumn>
         </Card>
       </Modal>
-      <ButtonEmpty padding={'0'} onClick={() => setShowDetails(true)}>
+      <ButtonEmpty padding={'0'} style={{ textDecoration: 'none' }} onClick={() => setShowDetails(true)}>
```
zzmp, line 105: Why is this bit necessary?

src/theme/index.tsx
```diff
@@ -20,6 +20,14 @@ export const MEDIA_WIDTHS = {
...
+  background = -1,
+  zero = 0,
```
zzmp, line 25: Can this be given a more semantic value?

src/components/swap/UnsupportedCurrencyFooter.tsx
JFrankfurt: I generally pref avoiding inline styles in favor of [modifying the base styled-component](https://styled-components.com/docs/basics#extending-styles).

cmcewen: this was a request from @willhennessy to remove the underline. 

@JFrankfurt agree with you there but the base component is used a few other places (https://github.com/Uniswap/uniswap-interface/blob/main/src/components/PositionCard/V2.tsx#L99) and also has some inline styles there, so I'm hesitant to modify it without knowing the cascading effects

one thing that helps me feel a little more comfortable modifying base component styles is jest snapshot tests - have y'all considered those at all?

src/theme/index.tsx
cmcewen: yeah i was hoping to have a bit of a CSS style conversation on how to approach this. i'm definitely in favor of putting some standard convention around z-index (https://getbootstrap.com/docs/5.0/layout/z-index/, https://tailwindcss.com/docs/z-index), but i think that is outside of the scope of this bug fix

i would hope that convention leads to never using zero - but it seemed like a reasonable choice to get this bug fixed and deployed

src/components/swap/UnsupportedCurrencyFooter.tsx
justindomingue: I think Jordan meant styling the base component in this file

```
const StyledButtonEmpty = styled(ButtonEmpty)`
  text-decoration: none;
`
```

justindomingue: i recently added support for jest snapshots but our coverage is almost 0 right now.

https://github.com/Uniswap/uniswap-interface/tree/main/src/components/TextInput

src/theme/index.tsx
justindomingue: I would be in favor of bootstrap's approach. I'd probably make the change in this PR too so people don't start relying on temporary `Z_INDEX` values.

PR #2276, opened 2021-08-31
chore: upgrade type system

zzmp:
Upgrade types, typescript, and formatters.

### DIFF
--- a/src/hooks/Tokens.ts
+++ b/src/hooks/Tokens.ts
@@ -76,7 +76,7 @@ export function useSearchInactiveTokenLists(search: string | undefined, minResul
       if (!list) continue
       for (const tokenInfo of list.tokens) {
         if (tokenInfo.chainId === chainId && tokenFilter(tokenInfo)) {
-          const wrapped = new WrappedTokenInfo(tokenInfo, list)
+          const wrapped: WrappedTokenInfo = new WrappedTokenInfo(tokenInfo, list)
           if (!(wrapped.address in activeTokens) && !addressSet[wrapped.address]) {
             addressSet[wrapped.address] = true
             result.push(wrapped)
--- a/src/state/lists/reducer.ts
+++ b/src/state/lists/reducer.ts
@@ -48,8 +48,6 @@ export default createReducer(initialState, (builder) =>
   builder
     .addCase(fetchTokenList.pending, (state, { payload: { requestId, url } }) => {
       state.byUrl[url] = {
-        current: null,
-        pendingUpdate: null,
         ...state.byUrl[url],
         loadingRequestId: requestId,
         error: null,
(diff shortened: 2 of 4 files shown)


src/state/lists/reducer.ts
```diff
@@ -47,10 +47,12 @@ const initialState: ListsState = {
...
   builder
     .addCase(fetchTokenList.pending, (state, { payload: { requestId, url } }) => {
+      const current = state.byUrl[url]?.current ?? null
+      const pendingUpdate = state.byUrl[url]?.pendingUpdate ?? null
+
       state.byUrl[url] = {
-        current: null,
-        pendingUpdate: null,
-        ...state.byUrl[url],
+        current: current,
```
justindomingue, line 54: can you shorten to this?

```
{
current,
pendingUpdate,
loadingRequestId: requestId,
error: null,
}
```

src/state/lists/reducer.ts
```diff
@@ -66,11 +68,10 @@ export default createReducer(initialState, (builder) =>
         if (upgradeType === VersionUpgrade.NONE) return
         if (loadingRequestId === null || loadingRequestId === requestId) {
           state.byUrl[url] = {
-            ...state.byUrl[url],
```
justindomingue, line 69: doesn't removing `...state.byUrl[url]` change the behavior? Can't remember if we're using imer in the background

zzmp: imer? According to the typing, this should not change the behavior because all of the properties from state.byUrl[url] are overwritten on the next 4 lines.

src/state/lists/reducer.ts
zzmp: I expect this to be covered by eslint. Can I just make a separate PR where I set the object-shorthand rule to "always", and fix the errors (there's a few files).

zzmp: I could do it as a new commit on this PR, but it affects 7 files, and I'd rather avoid the scope creep.

src/state/lists/reducer.ts
justindomingue: ah okay makes sense

was referring to https://github.com/immerjs/immer

src/hooks/Tokens.ts
```diff
@@ -76,7 +76,7 @@ export function useSearchInactiveTokenLists(search: string | undefined, minResul
       if (!list) continue
       for (const tokenInfo of list.tokens) {
         if (tokenInfo.chainId === chainId && tokenFilter(tokenInfo)) {
-          const wrapped = new WrappedTokenInfo(tokenInfo, list)
+          const wrapped: WrappedTokenInfo = new WrappedTokenInfo(tokenInfo, list)
```
justindomingue, line 79: surprised this is required given we're clearly instantiating a `WrappedtokenInfo`?

justindomingue (approved):

PR #2277, opened 2021-08-31
feat: adding oneUNI as a common base token

37aces:
Added oneUNI (Stable UNI) as a token.
Added oneUNI as a BASES_TO_CHECK_TRADES_AGAINST
Added oneUNI as a COMMON_BASES

### DIFF
--- a/src/constants/routing.ts
+++ b/src/constants/routing.ts
@@ -13,6 +13,7 @@ import {
   USDC,
   USDT,
   WBTC,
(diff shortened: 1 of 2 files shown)


dthibeau:
@Yorkemartin - where can we submit PRs for routing on the Uniswap protocol? No UI support requested. We simply would like to route certain pairs through the Stable UNI token we created.

PR #2281, opened 2021-09-01
ci: upgrade crust pinner adapt to crust mainnet 

zikunfan:
Crust Pinner `v1.0.8`(based on Crust Preview Network Maxwell) has already been help Uni continuously increasing IPFS replica count in the past 161 days and stored 100 times. Each version of Uni-Interface is stored by over 60+ IPFS nodes all over the world, it makes Uni-interface more distributed and decentralized.

Since Crust already launched MainNet(now already has 5000+ IPFS nodes and more than 1.5 EB storage space, you can check more detail [here](https://crust.subscan.io/storage)), I propose this PR(upgrade Crust Pinner to `v2.0.3`) to help Uni pin on Crust MainNet, which will further increase IPFS replicas. 😀


moodysalem:
Note this is considered safe because the action happens after the build and IPFS hash is generated, and stored in a way that cannot be manipulated by any other GitHub action

PR #2305, opened 2021-09-07
feat: use SDK to format calldata passed to quoter

NoahZinsmeister:
uses the new features introduced [here](https://github.com/Uniswap/uniswap-v3-sdk/pull/76) to format calldata passed to the quoter, as opposed to doing it manually.

as side effects:

- introduce new method `useSingleContractWithCallData` to call functions via multicall directly from encoded calldata. note: unlike `useSingleContractMultipleData`, this new method can call any number of different functions.
- as a cleanup, make `useSingleCallResult` a wrapper over `useSingleContractMultipleData`

### DIFF
--- a/src/state/multicall/hooks.ts
+++ b/src/state/multicall/hooks.ts
@@ -193,7 +193,7 @@ export function useMultipleContractSingleData(
   contractInterface: Interface,
   methodName: string,
   callInputs?: OptionalMethodInputs,
-  options?: Partial<ListenerOptions> & { gasRequired?: number }
+  options: Partial<ListenerOptions> & { gasRequired?: number } = {}
 ): CallState[] {
   const fragment = useMemo(() => contractInterface.getFunction(methodName), [contractInterface, methodName])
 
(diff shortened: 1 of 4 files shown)


src/state/multicall/hooks.ts
```diff
@@ -162,21 +163,31 @@ export function useSingleContractMultipleData(
 ): CallState[] {
   const fragment = useMemo(() => contract?.interface?.getFunction(methodName), [contract, methodName])
 
-  const blocksPerFetch = options?.blocksPerFetch
+  // encode callDatas
+  const callDatas = useMemo(
+    () =>
+      callInputs.every((callInput) => isValidMethodArgs(callInput)) && fragment
```
moodysalem, line 169: idr why we check that every input is a valid list of args instead of just encoding the ones that are valid and ignoring the ones that aren't

src/state/multicall/hooks.ts
```diff
@@ -162,21 +163,31 @@ export function useSingleContractMultipleData(
...
+        ? callInputs.map<string>((callInput) =>
+            // this type assertion is ok, because if fragment is defined, contract must be as well
```
moodysalem, line 171: i don't love these because while true today they may not be true if someone edits the above fragment computation, so i would just add `&& contract` after `&& fragment`

src/state/multicall/hooks.ts
```diff
@@ -162,21 +163,31 @@ export function useSingleContractMultipleData(
...
-              address: contract.address,
-              callData: contract.interface.encodeFunctionData(fragment, inputs),
-              ...(gasRequired ? { gasRequired } : {}),
-            }
-          })
-        : [],
-    [contract, fragment, callInputs, gasRequired]
+      callDatas.map<Call>((callData) => ({
+        // this type assertion is ok, because for callDatas.length to be > 0, contract must be defined
+        address: (contract as Contract).address,
```
moodysalem, line 186: same here

src/state/multicall/hooks.ts
```diff
@@ -162,21 +163,31 @@ export function useSingleContractMultipleData(
...
+        callData,
+        ...(gasRequired ? { gasRequired } : {}),
```
moodysalem, line 188: weird, why didn't we just put in `gasRequired`? it can be undefined

moodysalem (approved):
seems fine

src/state/multicall/hooks.ts
NoahZinsmeister: yeah in theory i think we could do that and use the `Call | undefined` return sig like in `useMultipleContractSingleData`, not sure it's worth it though

src/state/multicall/hooks.ts
NoahZinsmeister: yeah not sure, will change

src/state/multicall/hooks.ts
NoahZinsmeister: ok addressed actually

PR #2309, opened 2021-09-07
feat: Use frontrunning protection by Flashbots & mistX

Dire-0x:
### Option to send transactions to Flashbots using the mistX SDK
Transactions sent to the Flashbots network avoid the mempool entirely thus protecting them from frontrunning attacks carried out by bots & miners. This implementation only works using MetaMask on Mainnet

### UX Considerations

1. Gas fees are not chosen by the user during the MetaMask signing process using `eth_sign`. Gas fees are calculated by the API and shown on the swap confirm modal. Additionally, the user should not be able to sign a transaction if they do not have enough ETH to cover the gas fees. See todos 1 & 2.

Screen Shot 2021-09-07 at 11 56 59 AM

2. Failed transactions sent through Flashbots cannot link to the corresponding transaction in Etherscan because they never exist in the mempool. See todo 3.

Screen Shot 2021-09-07 at 12 09 34 PM

3. The API is limited to one transaction per EOA (transaction from address). Due to this limitation, the user should be blocked from submitting a frontrunning protected transaction while they have a pending transaction (frontrunning protected or not). See todo 4.

### TODOS
1. Review the UI & language for displaying gas fees in the swap confirm modal. 
2. Define the UX for blocking the user from signing a transaction when they do not have enough ETH to cover the gas fees.
3. Review how non-linkable transactions should be displayed in the transaction history.
4. Define the UX for blocking a user from submitting a frontrunning protected transaction while they have a pending transaction


Dire-0x:
@moodysalem I can't see the Vercel logs. Can you see why the build is failing?

moodysalem:
```
./src/pages/Swap/index.tsx
Cannot find module: 'components/swap/FrontrunningProtectionFooter'. Make sure this package is installed.
```

willhennessy:
Hi team, @ppphhhiiilll and I are working on UX design for this! We have a few questions:

1. Does flashbots require any new approves?
2. #2 in the PR says we can’t link to etherscan for failed or pending tx.
    1. is that also true of pending tx?
    2. will we be able to get the data of a failed tx? (e.g. amount)
    3. is gas still consumed on a failed flashbots tx?
3. Does flashbots support cancelling a tx by sending another tx?
4. @moodysalem  how does flashbots gas interplay with the Routing API?
5. Plans for Optimism or Arbitrum support?

moodysalem:
Note we need to add `<Trans>` around all copy

bertmiller:
- Does flashbots require any new approves?
Nope!
- In the original post point 2 says we can’t link to etherscan for failed or pending tx.
- is that also true of pending tx?
Yes, unfortunately. You could link to the pending tx, but Etherscan won't have any data on it.
- will we be able to get the data of a failed tx? (e.g. amount)
Shouldn't you already have the data since it was generated in the Uniswap UI?
- is gas still consumed on a failed flashbots tx?
So, there is a difference between failing at the Flashbots level vs failing on-chain.

It's possible for a transaction to "fail" on the Flashbots level simply by not paying a competitive enough gas price to be included and eventually expiring. In this case the transaction never lands on-chain, and to answer your question it consumes no gas. Also, transactions that would fail on-chain will be rejected by MEV-Geth and not included in blocks, which has obvious benefits for traders.

- is the user able to submit other non-flashbots tx while a flashbot tx is pending?
They can although the normal caveats around nonce management apply here.
- Does flashbots support cancelling a tx by sending another tx?
Yep! 
- @moodysalem how does flashbots gas interplay with the Routing API?
At least from my perspective this should be the same as it is normally, but you can expect the gas price needed to use Flashbots to be a bit higher than normal.
- Plans for Optimism or Arbitrum support?
We don't have anything concrete at the moment.

Dire-0x:
> Does flashbots support cancelling a tx by sending another tx?
The SDK provides a way to attempt cancelling a tx without sending another tx.

Transactions are sent to the Flashbots relay every block until success, failure, or a deadline is reached. A transaction cancellation can be submitted but does not have a 100% success rate because the transaction may be in-flight to the Flashbots relay and be included by a miner before the cancellation can occur. There are no fees for cancellations

bertmiller:
hey @moodysalem mind having a look at this again and telling us what you think? We made some initial decisions on a few of the UX questions highlighted above and feedback would be really helpful.

src/hooks/useSwapCallback.ts
```diff
@@ -337,51 +377,81 @@ export function useSwapCallback(
...
+              if (web3Provider) {
+                web3Provider.provider.isMetaMask = true
+              }
+
+              /**
+               * Hack for detecting hardware wallets connected through metamask
+               * Hardware wallets cannot sign with eth_sign. A direct hardware wallet connection should be used instead
+               */
+              const parsed = parseTransaction(rawSignedTransaction)
+              if (parsed.from !== account) {
```
joehawary, line 424: @Dire-0x 
Two things I wanted to mention
1. The 'from' address could differ on non supported wallets too - not just hardware, so maybe a copy that covers both scenarios?
2. A secondary check could also be done to check the nonce number against the `getTransactionCount` function to further identify other incompatible wallets.

Dire-0x:
Moving to a new PR that implements the Flashbots Protect RPC https://github.com/Uniswap/interface/pull/2713

PR #2321, opened 2021-09-10
feat: add @davatar/react for robust avatar support

carlosdp:
## Description

This PR adds robust avatar image lookup support for ENS `avatar` text record lookup for the logged in user's avatar, using the [@davatar/react](https://github.com/TBDAO/davatar-helpers/tree/master/packages/react) package, which seeks to fulfill the spec outlined in [avatars.md](https://gist.github.com/Arachnid/9db60bd75277969ee1689c8742b75182).

Presently, we support all URI types except the NFT-metadata reference type, although that one is planned in a future package release. Our goal is to push forward usage of decentralized avatars using ENS, and this PR works toward making sure as people do so, they see their chosen avatar in the Uniswap app!

The component defaults to a Jazzicon, exactly as Uniswap currently behaves today, so if no avatar has been set by the user on their ENS reverse record, they see exactly the same thing as they always have in the app.

## Security / Supply Chain risk consideration
The `@davatar/react` library uses minimal dependencies (just the `color` package and the `merssene-twister` package, which the `@metamask/jazzicon` package that this one replaces already used, and `ethers`). This adds no extra third-party dependencies to the project, other than the `@davatar/react` package itself, which aims to be very small.

## List of features added/changed

- Add `@davatar/react` dependency (small, 7kb)
- Remove `@metamask/jazzicon` (no longer needed)
- Replace the jazzicon SVG component for rendering avatars with the `<Davatar />` component from `@davatar/react`, which supports several decentralized URI protocols (in addition to Jazzicon default)

## How Has This Been Tested?

Connected my wallet in the local dev environment and checked that it rendered my avatar correctly.

## Screenshots (if appropriate):
Screen Shot 2021-09-10 at 2 24 08 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "homepage": ".",
   "private": true,
   "devDependencies": {
+    "@davatar/react": "^1.0.2",
     "@ethersproject/experimental": "^5.4.0",
     "@gnosis.pm/safe-apps-web3-react": "^0.6.0",
     "@graphql-codegen/cli": "1.21.5",
@@ -154,7 +155,5 @@
     ]
   },
   "license": "GPL-3.0-or-later",
-  "dependencies": {
-    "@davatar/react": "^1.0.2"
-  }
+  "dependencies": {}
 }
--- a/package.json
+++ b/package.json
@@ -155,5 +155,7 @@
     ]
   },
   "license": "GPL-3.0-or-later",
-  "dependencies": {}
+  "dependencies": {
+    "@davatar/react": "^1.0.2"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "homepage": ".",
   "private": true,
   "devDependencies": {
-    "@davatar/react": "^1.0.2",
+    "@davatar/react": "1.1.0",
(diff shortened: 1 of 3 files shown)


NoahZinsmeister:
cool project @carlosdp thanks for the PR! will review in more depth shortly but for now could you add the dependency as a devDependency instead?

justindomingue (approved):

carlosdp:
Thanks 😊 Yep, will make that change when I get back to my laptop!

carlosdp:
Moved to `devDependencies`!

NoahZinsmeister:
ok @carlosdp this LGTM with one more change, could you please pin a specific semver? in the name of reducing dependency risk :)

carlosdp:
Do you mean an exact version? It's currently pinned to ^1.0.1, same as all other current deps except graphql-cli

carlosdp:
@NoahZinsmeister (forgot to tag you in previous comment)

NoahZinsmeister:
sorry yes an exact version @carlosdp :)

carlosdp:
@NoahZinsmeister gotcha, pushed commit with that change! 😄

PR #2287, opened 2021-09-01
style: Add prettier organize imports plugin and run on all files

cmcewen:
Add a prettier plugin to organize imports - h/t @willpote

PR #2339, opened 2021-09-13
chore: update lint-action version

cmcewen:
I'm having some trouble with #2287, which I believe stems from the fact that the lint-action commit uses an older version of eslint and prettier. It runs the master workflow on PRs, so I need to merge this first before it will use the proper newer versions

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -39,7 +39,7 @@ jobs:
         run: yarn install --frozen-lockfile
 
       - name: Run linters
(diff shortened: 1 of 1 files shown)


zzmp:
lint-actions _should_ be using our own installed version of `prettier` and `eslint`. It does have the eslint plugin for prettier in its dependencies, so I'm not sure if it's using ours or its own.

zzmp (approved):

moodysalem:
i updated the workflow file on main

PR #2340, opened 2021-09-13
fix: only try SafeApp connection in an iframe

### LINKED ISSUE #2338: gnosis SafeApp introduces a 300ms pageload timeout
To use the gnosis SafeApp, you must check for it before using other injectors, so it sits in the pageload critical path.

To check for SafeApp, the SDK races an iframe postMessage with a 300ms timeout (see https://github.com/gnosis/safe-apps-sdk/blob/master/packages/safe-apps-web3-react/src/connector.ts#L52). This should be skipped if we are not in an iframe, and there should be an immediate, deterministic way to tell if we are in a SafeApp.

zzmp:
Improves non-iframe pageload by 300ms. Fixes #2338.

The Gnosis check for a SafeApp races a postMessage and a 300ms timeout ([src]). The SafeApp embeds the interface in an iframe, so this avoids the check when not in iframes.

[src]: https://github.com/gnosis/safe-apps-sdk/blob/f224869dd5ae810db8cecad08fbbcfaa6c046d9d/packages/safe-apps-web3-react/src/connector.ts#L52

### DIFF
--- a/src/hooks/web3.ts
+++ b/src/hooks/web3.ts
@@ -1,6 +1,5 @@
 import { Web3Provider } from '@ethersproject/providers'
-import { useSafeAppConnection } from '@gnosis.pm/safe-apps-web3-react'
-import { useWeb3React as useWeb3ReactCore } from '@web3-react/core'
+import { useWeb3React } from '@web3-react/core'
 import { Web3ReactContextInterface } from '@web3-react/core/dist/types'
 import { useEffect, useState } from 'react'
 import { isMobile } from 'react-device-detect'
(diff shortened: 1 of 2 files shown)


src/hooks/web3.ts
```diff
@@ -1,26 +1,44 @@
...
+  const { activate, active } = useWeb3React()
+  const [triedSafe, setTriedSafe] = useState(false)
   const [tried, setTried] = useState(false)
 
-  const triedToConnectToSafe = useSafeAppConnection(gnosisSafe)
+  // first, try connecting to a gnosis safe
+  useEffect(() => {
+    // gnosisSafe.isSafeApp() races a timeout against postMessage, so it delays pageload if we are not in a safe app;
+    // if we are not in an iframe, it is not worth checking
+    if (window && window.parent !== window) {
```
moodysalem, line 24: we have a constant with this check

src/hooks/web3.ts
```diff
@@ -1,26 +1,44 @@
...
+      gnosisSafe.isSafeApp().then((loadedInSafe) => {
+        if (loadedInSafe) {
+          activate(gnosisSafe, undefined, true).catch(() => {
+            setTriedSafe(true)
+          })
+        } else {
+          setTriedSafe(true)
```
moodysalem, line 31: you can just default triedSafe to true if you're not in an iframe and avoid a setState

moodysalem (approved):
nice find

src/hooks/web3.ts
zzmp: added to src/constants/misc.ts

PR #2343, opened 2021-09-14
feat(i18n): initialize locale before React

zzmp:
Eagerly loads the active locale asset as soon as it is known, instead of waiting for React to bootstrap.
A continuation of #2204.

### DIFF
--- a/src/hooks/useParsedQueryString.ts
+++ b/src/hooks/useParsedQueryString.ts
@@ -2,10 +2,15 @@ import { parse, ParsedQs } from 'qs'
 import { useMemo } from 'react'
 import { useLocation } from 'react-router-dom'
 
+export function parsedQueryString(search?: string): ParsedQs {
+  if (!search) {
+    const hash = window.location.hash
+    search = hash.substr(hash.indexOf('?'))
+  }
+  return search && search.length > 1 ? parse(search, { parseArrays: false, ignoreQueryPrefix: true }) : {}
+}
+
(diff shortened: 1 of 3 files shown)


src/hooks/useParsedQueryString.ts
```diff
@@ -2,10 +2,15 @@ import { parse, ParsedQs } from 'qs'
 import { useMemo } from 'react'
 import { useLocation } from 'react-router-dom'
 
+export function parsedQueryString(search?: string): ParsedQs {
+  if (!search) {
+    const hash = window.location.hash
```
justindomingue, line 7: probably worth a comment explaining why we are doing this for future readers

src/i18n.tsx
```diff
@@ -82,21 +83,24 @@ async function dynamicActivate(locale: SupportedLocale) {
...
+        setUserLocale(locale) // stores the selected locale to persist across sessions
         setLoaded(true)
       })
       .catch((error) => {
         console.error('Failed to activate locale', locale, error)
       })
-  }, [locale])
+  }, [locale, setUserLocale])
 
   // prevent the app from rendering with placeholder text before the locale is loaded
```
justindomingue, line 105: do we still need this bit?

justindomingue (approved):

src/i18n.tsx
zzmp: yes - in the case that the locale is not the default/fallback locale, this prevents the extra network request for the fallback if the locale is not yet loaded.

PR #2349, opened 2021-09-14
fix: only connect to safe on mount

zzmp:


zzmp:
Tested on gnosis, works as expected.

PR #2362, opened 2021-09-15
style: ban esnext syntax

zzmp:


zzmp:
This could also be accomplished via browserslist compliance, with something like [eslint-plugin-compat](https://www.npmjs.com/package/eslint-plugin-compat).

justindomingue:
why are we banning esnext?

zzmp:
This was an attempt at browser compatibility, but we couldn't get the CI linter to accept it.

sambacha:
> This could also be accomplished via browserslist compliance, with something like [eslint-plugin-compat](https://www.npmjs.com/package/eslint-plugin-compat).

would be interested to hear your thoughts on the subject of browserlist usage , see https://github.com/Uniswap/interface/pull/2140

PR #2371, opened 2021-09-16
refactor: simplify trade hooks

justindomingue:
previously trade hooks were split in `*ExactIn|ExactOut`. this pr merges the two


src/hooks/useClientSideV3Trade.ts
```diff
@@ -59,31 +69,49 @@ export function useClientV3TradeExactIn(
...
         if (!result) return currentBest
 
-        if (currentBest.amountOut === null) {
-          return {
-            bestRoute: routes[i],
-            amountOut: result.amountOut,
+        // overwrite the current best if it's not defined or if this route is better
+        if (tradeType === TradeType.EXACT_INPUT) {
+          const amountOut = CurrencyAmount.fromRawAmount(currencyOut, result.amountOut.toString())
+          if (currentBest.amountOut === null || JSBI.lessThan(currentBest.amountOut.quotient, amountOut.quotient)) {
```
justindomingue, line 87: @NoahZinsmeister curious why did you change from `CurrencyAmount.lt` to `CurrencyAmount.fromRawAmount` & `JSBI.lessThan`?

ianlapham (commented):
Looks good to me aside from naming on v3Trade hook

src/hooks/useClientSideV3Trade.ts
ianlapham: I actually think we can just do `currentBest.amountOut.lessThan(amountOut.quotient)`

Not sure why we do this - this hook was written way back even before we used SDK frequently

PR #2391, opened 2021-09-19
perf: lazy load routes to reduce initial bundle size

chenxsan:
I noticed there was [a pull request](https://github.com/Uniswap/interface/pull/1597) doing what this pull request want to do, but it was reverted later in https://github.com/Uniswap/interface/pull/1611.

This pull request does almost the same thing like the initial pull request, however with a suspense.

Currently the loading indicator is simply a `loading...`, I think it can be improved to conform to the uniswap's theme.

### DIFF
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -25,7 +25,6 @@ const MigrateV2Pair = lazy(() => import('./MigrateV2/MigrateV2Pair'))
 const OpenClaimAddressModalAndRedirectToSwap = lazy(() =>
   import('./Swap/redirects').then((module) => ({ default: module.OpenClaimAddressModalAndRedirectToSwap }))
 )
-const RedirectToSwap = lazy(() => import('./Swap/redirects').then((module) => ({ default: module.RedirectToSwap })))
 const Pool = lazy(() => import('./Pool'))
 const PoolFinder = lazy(() => import('./PoolFinder'))
 const PoolV2 = lazy(() => import('./Pool/v2'))
@@ -39,6 +38,7 @@ const RedirectDuplicateTokenIdsV2 = lazy(() =>
 const RedirectPathToSwapOnly = lazy(() =>
   import('./Swap/redirects').then((module) => ({ default: module.RedirectPathToSwapOnly }))
 )
+const RedirectToSwap = lazy(() => import('./Swap/redirects').then((module) => ({ default: module.RedirectToSwap })))
 const RemoveLiquidity = lazy(() => import('./RemoveLiquidity'))
 const RemoveLiquidityV3 = lazy(() => import('./RemoveLiquidity/V3'))
 const Vote = lazy(() => import('./Vote'))
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -1,3 +1,4 @@
+import * as React from 'react'
 import ApeModeQueryParamReader from 'hooks/useApeModeQueryParamReader'
 import { Route, Switch } from 'react-router-dom'
 import styled from 'styled-components/macro'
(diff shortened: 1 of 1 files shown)


src/pages/App.tsx
```diff
@@ -11,24 +12,37 @@ import Web3ReactManager from '../components/Web3ReactManager'
...
-import PoolV2 from './Pool/v2'
-import PoolFinder from './PoolFinder'
-import RemoveLiquidity from './RemoveLiquidity'
-import RemoveLiquidityV3 from './RemoveLiquidity/V3'
 import Swap from './Swap'
-import { OpenClaimAddressModalAndRedirectToSwap, RedirectPathToSwapOnly, RedirectToSwap } from './Swap/redirects'
-import Vote from './Vote'
-import VotePage from './Vote/VotePage'
+
+const { lazy, Suspense } = React
```
justindomingue, line 17: styling nit. we would typically import those from `'react'` rather than destructuring

src/pages/App.tsx
```diff
@@ -84,46 +98,51 @@ export default function App() {
...
-
-              <Route exact strict path="/send" component={RedirectPathToSwapOnly} />
-              <Route exact strict path="/swap/:outputCurrency" component={RedirectToSwap} />
-              <Route exact strict path="/swap" component={Swap} />
-
-              <Route exact strict path="/pool/v2/find" component={PoolFinder} />
-              <Route exact strict path="/pool/v2" component={PoolV2} />
-              <Route exact strict path="/pool" component={Pool} />
-              <Route exact strict path="/pool/:tokenId" component={PositionPage} />
+            <Suspense fallback={<div>loading...</div>}>
```
justindomingue, line 101: English text should be wrapped in `<Trans>` to get translations.

Use `<Dots>` in `components/swap/styleds` instead of `...`

justindomingue: @callil did you ever look at what the suspense animation would look like? I like the running horse from v1..

justindomingue (commented):
looking good, would definitely like to refine the loading animation

fyi the previous pr was reverted because there was no suspense which meant the header would get removed on fetch.

src/pages/App.tsx
chenxsan: @justindomingue I've updated the pull request per your review, thanks for the guides.

zzmp:
This looks great, thanks! Have you instrumented or tested the performance implications of this change? Lazy loading _every_ route means that _none_ of them can be rendered without an additional fetch. I think this should be implemented more gradually so we can see what the actual effect will be. Specifically, I don't think that `/swap` should ever be lazy loaded (as it is the default route), and I think that each route that is lazy loaded should be instrumented first to see what the actual savings/costs are.

Could you focus on a few routes, so that you can collect that info and make a compelling case for code splitting here? You can use Chrome DevTools, along with [source map explorer](https://create-react-app.dev/docs/analyzing-the-bundle-size), to start to drill down on these. I know it makes this a much bigger task, so reach out with any questions/concerns about it.

chenxsan:
@zzmp 

Regarding your concern about that `/swap`, it's not lazy loaded in this pull request, you can check the code here https://github.com/Uniswap/interface/pull/2391/files#diff-11b6c68fcd0989e7541fed72fed10b836738171fe1b5f1bb01721f51b3e5e622R16. Generally I think it's possible to find out which routes are used most often through Google Analytics and avoid lazy load them.

I'll present some data I can collect for this pull request.

## development environment

We'll target the size of `main.chunk.js` file here.

### before the change

Size: 4.1MB

before

### after the change

Size: 2.5MB

after

## production environment

I'll run the `yarn build` command locally to build assets for production environment, then analysis the built asset with `source-map-explorer`.

### before the change

Size: 715.71KB

image

### after the change

Size: 473.84KB

image

zzmp:
> Regarding your concern about that /swap, it's not lazy loaded in this pull request

My mistake, good job there.

> development environment

Perf improvements only need to target production bundles. Sometimes what looks like a win (or loss) in development drastically changes when using more aggressive webpack/optimization options.

> production environment

This looks like a 34% reduction in size, also great, but the largest chunk that we deliver is the vendors chunk. Does this have any effect there?

Another thing to look at is how this affects the pageload. Does loading the other pages in the initial chunk holding up the main thread? You can take a performance trace (using a production bundle) to determine if there is any improvement here, but I suspect it will be minimal.

Because the vendors chunk is so much larger, I think this should be judged based on the effect there, because lazy loading means that pages _other than Swap_ will load slightly slower, so we need to see some substantial improvement to Swap to make it worthwhile. Do you agree / disagree?

chenxsan:
1. Changes in this pull request are application-specific code, none of them are from `node_modules`, so they won't affect the vendors chunk. From what I can tell, create-react-app takes advantage of the default [`optimization.splitChunks`](https://v4.webpack.js.org/plugins/split-chunks-plugin/#optimizationsplitchunks) when creating the vendors chunk, since the configuration is not ejected, so there're not much things we can do but lazy-loading some vendors, however that would be another topic.
2. I'll see if I can take a performance trace later. But even the improvement is minimal, it might still worth it. For example, when 80% uniswap users use `/swap` while only 30% unswap users use other routes, in this case, we could still save a lot of time for most users.
3. yeah, it's about trade-off here. It's not possible to please everyone. The reason I filed this pull request is I use `/swap` most, and I want to make it fast for myself :)

zzmp:
You can affect the vendors chunk by lazy-loading application code, if it is the only code using that vendor package. For example, react-markdown is only used in /vote, so just lazy-loading the /vote routes will also reduce the initial vendors chunk size, because vendor packages specific to /vote will be lazy-loaded along with the application code.

For now, I'd like to see /swap and /pool remain eagerly loaded (as they are the most heavily used), and the rest loaded in appropriate sections. Using /vote as an example again, I think any page under /vote should be lazy-loaded together, which means a slight restructuring of the files/imports/folders. Because this is section-specific lazy-loading, I think each top-level route should be modified as a separate Pull Request. Lastly, I'd like the loading element (`fallback`) to either be our existing spinner, or a page-specific skeleton layout, although this should go to our designers for their opinion.

Would you be willing to make a new PR that just separates out the /vote page?

zzmp:
Another place where code-splitting would probably see some easy gains is by splitting the V2 pages, which can probably be worked on in parallel with #2468.

vm:
Closing for now as we likely want to add lazy loading incrementally (like in the linked PR above)

PR #2400, opened 2021-09-20
fix(web3): render children while awaiting connection

zzmp:


NoahZinsmeister (approved):
LGTM, don't think this is a problem!

PR #2401, opened 2021-09-20
fix(redux): setup store listeners immediately

zzmp:
Allows Redux DevTools to record all dispatched actions.

### DIFF
--- a/src/state/index.ts
+++ b/src/state/index.ts
@@ -43,9 +43,9 @@ const store = configureStore({
   preloadedState: load({ states: PERSISTED_KEYS, disableWarnings: process.env.NODE_ENV === 'test' }),
 })
 
-store.dispatch(updateVersion())
-
 setupListeners(store.dispatch)
+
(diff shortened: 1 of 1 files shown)


moodysalem:
i'm pretty sure the listeners are for the redux toolkit caching stuff and the devtools are from the middleware

src/state/index.ts
```diff
@@ -43,10 +43,10 @@ const store = configureStore({
   preloadedState: load({ states: PERSISTED_KEYS, disableWarnings: process.env.NODE_ENV === 'test' }),
 })
 
-store.dispatch(updateVersion())
-
 setupListeners(store.dispatch)
 
+store.dispatch(updateVersion())
```
justindomingue, line 48: What is the goal here?

`setupListeners(store.dispatch)` sets up the RTK-Query listeners for on window focus/blur and such.

zzmp:
You're right. I misread the DevTools UI and assumed this did more than it did.

Issue #972, opened 2020-07-18
Keep URL synced with swap interface


robertuniqid:
**Is your feature request related to a problem? Please describe.**

It seems that uniswap does not detect the hashtag change in the URL ( tough it would change when swapping between currencies, and would require to have an "ignore" functionality in that part if it's already set on the interface or still processing.

**Describe the solution you'd like**

I'd like for uniswap to detect the hashchange, and update the currencies accordingly, this seems very practical if using bookmarks, and in general it's a best practice on the web.

**Additional context**

Simple video showcasing the issue : https://d.pr/v/nlqFyh

moodysalem:
In addition, the app should change the query parameters as you select parameters in the swapping interface

NoahZinsmeister:
linking my solution for this on hypertext for consideration: https://github.com/NoahZinsmeister/hypertext/blob/79676225a680ce9bc5dc1010e434f5b398e3ed07/components/Settings.tsx#L149

i chose to populate initial UI from query params, then clear them. if at any point a user has modified any part of the ui that could be encoded as a query param, i show a share button. if possible, it triggers the native share ui via window.navigator.share, else falls back to copying

PR #2402, opened 2021-09-20
perf(swap): initialize swap inputs from url

zzmp:
Allows swap inputs to be rendered in the first frame.


src/state/swap/reducer.ts
```diff
@@ -33,10 +25,10 @@ export default createReducer<SwapState>(initialState, (builder) =>
       (state, { payload: { typedValue, recipient, field, inputCurrencyId, outputCurrencyId } }) => {
         return {
           [Field.INPUT]: {
-            currencyId: inputCurrencyId,
+            currencyId: inputCurrencyId || null,
```
justindomingue, line 28: would `??` be better here in case `currencyId` is falsy? I'm assuming an id of `0` wouldn't be valid anyway but we probably shouldn't bake those assumptions into the code.

justindomingue: dito in `swap/hooks` if so

src/hooks/Tokens.ts
```diff
@@ -181,8 +182,16 @@ export function useCurrency(currencyId: string | null | undefined): Currency | n
   const { chainId } = useActiveWeb3React()
   const isETH = currencyId?.toUpperCase() === 'ETH'
   const token = useToken(isETH ? undefined : currencyId)
-  const extendedEther = useMemo(() => (chainId ? ExtendedEther.onChain(chainId) : undefined), [chainId])
+  const extendedEther = useMemo(
```
justindomingue, line 185: idk if it translates but this block reminds me of `useCurrency`

https://github.com/Uniswap/interface/blob/f59c5f687742d9d50adf18f0756dff3eb6d659b6/src/hooks/Tokens.ts#L180

src/state/swap/hooks.test.ts
```diff
@@ -26,7 +26,7 @@ describe('hooks', () => {
         queryParametersToSwapState(parse('?outputCurrency=invalid', { parseArrays: false, ignoreQueryPrefix: true }))
```
justindomingue, line 26: above

`'?inputCurrency=ETH&outputCurrency=0x6b175474e89094c44da98b954eedeac495271d0f&exactAmount=20.5&exactField=outPUT',`

is `outPUT` supposed to have that casing?

justindomingue (approved):

moodysalem:
maybe the better refactor is to have the swap state be in the url

https://github.com/Uniswap/interface/issues/972
not sure why that was closed

zzmp:
This initializes the swap inputs from the URL in time for first render, so it is related but separate work from #972 (which I agree should be considered).

src/state/swap/reducer.ts
zzmp: the initial check needs to set it to null if the currencyId is '', as it is parsed from the url query param. otherwise, sgtm.

justindomingue (approved):

src/hooks/Tokens.ts
```diff
@@ -181,8 +182,16 @@ export function useCurrency(currencyId: string | null | undefined): Currency | n
...
+    () =>
+      chainId
+        ? ExtendedEther.onChain(chainId)
+        : // display mainnet when not connected
+          ExtendedEther.onChain(SupportedChainId.MAINNET),
```
moodysalem, line 190: i really don't like this default chain id here (i think this is the only place we do this), what needs to change is the app needs to control the selected network instead of the connector

zzmp: That should be a larger change in the future. I think it will also help with first paint and linking semantics (ie links can rely on the chain being correct).

PR #2403, opened 2021-09-20
perf(font): serve inter var from public

zzmp:
Preloads `Inter var` from public, using a custom name, so that it is not clobbered by the `inter-ui` package and will be ready for rendering ASAP.


public/index.html
```diff
@@ -26,15 +26,31 @@
...
         * {
             font-family: 'Inter', sans-serif;
             box-sizing: border-box;
         }
 
+        /**
+          Explicitly load Inter var from public/ so it does not block LCP's critical path.
+        */
+        @font-face {
+            font-family: 'Inter custom';
```
moodysalem, line 41: does this override the inter-ui import?

moodysalem (approved):

PR #2404, opened 2021-09-20
perf: render components on first frame

zzmp:
- Inlines the logo SVG so it renders with no additional fetches
- Renders "Vote" link in navbar while awaiting connection to avoid layout shift
- Renders "Connect to a wallet" while awaiting connection to improve LCP


mgowon7 (approved):

ianlapham (approved):
LGTM

PR #2405, opened 2021-09-20
perf(i18n): bundle default locale with main chunk

zzmp:
Effectively renders the page with one less network request in the default case.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,8 +4,6 @@
 /src/types/v3
 /src/abis/types
 /src/locales/**/*.js
-/src/locales/**/*.ts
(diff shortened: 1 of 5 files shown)


justindomingue:
is the downside here that english always has to be bundled in the main chunk regardless of the locale?

zzmp:
Yes, exactly that. It will omit one fetch from the default case, at the expense of increasing the main bundle size.

justindomingue:
> Yes, exactly that. It will omit one fetch from the default case, at the expense of increasing the main bundle size.

LGTM

PR #2412, opened 2021-09-21
fix: improve the meta tags

moodysalem:
i know the user scalable doesn't work on iOS anymore, but maybe it's nice for android


justindomingue (approved):
setting `user-scale=no` is nice on android 👍

PR #2419, opened 2021-09-22
perf: split web3 libs into a separate chunk

zzmp:
Splits the vendors library chunk from one file (2.3MB) to two (948kB and 1.3MB) by splitting out web3 libraries into their own chunk. This should improve loading times for browsers where the initial vendor chunks' downloads can be parallelized, and prevents us from using one very large (2.3MB) chunk.

Installs craco to extend the webpack configuration to configure chunking behavior.


moodysalem:
I would prefer not to do this one, I don't think splitting the chunk into two pieces will make it any more cacheable or make it load faster

PR #2424, opened 2021-09-22
perf: import @ethersproject directly to enforce modularization

zzmp:


zzmp:
The entire `ethers` package is also brought in as a dependency to @davatar/react, which should be resolved by https://github.com/TBDAO/davatar-helpers/pull/4. This commit will be ineffective without an update to @davatar/react.

NoahZinsmeister (approved):
LGTM! will be nice once we get the @davatar update in

zzmp:
Waiting to merge this until @davatar/react updates, so that the version bump can be included in this PR.

PR #2428, opened 2021-09-23
fix(L2): add common bases to L2s

JFrankfurt:
These common bases are now available on L2 networks. The tokens don't have assets in the trustwallet repo though, so images won't show up until that data is populated.

image

### DIFF
--- a/src/components/CurrencyLogo/index.tsx
+++ b/src/components/CurrencyLogo/index.tsx
@@ -10,8 +10,29 @@ import Logo from '../Logo'
 
 type Network = 'ethereum' | 'arbitrum' | 'optimism'
 
-export const getTokenLogoURL = (address: string, network: Network = 'ethereum') =>
-  `https://raw.githubusercontent.com/Uniswap/assets/master/blockchains/${network}/assets/${address}/logo.png`
+function chainIdToNetworkName(networkId: SupportedChainId): Network {
+  switch (networkId) {
+    case SupportedChainId.MAINNET:
+      return 'ethereum'
+    case SupportedChainId.ARBITRUM_ONE:
+      return 'arbitrum'
+    case SupportedChainId.OPTIMISM:
+      return 'optimism'
+    default:
+      return 'ethereum'
+  }
+}
+
+export const getTokenLogoURL = (
+  address: string,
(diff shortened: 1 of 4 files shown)


src/components/CurrencyLogo/index.tsx
```diff
@@ -7,8 +8,10 @@ import useHttpLocations from '../../hooks/useHttpLocations'
 import { WrappedTokenInfo } from '../../state/lists/wrappedTokenInfo'
 import Logo from '../Logo'
 
-export const getTokenLogoURL = (address: string) =>
-  `https://raw.githubusercontent.com/uniswap/assets/master/blockchains/ethereum/assets/${address}/logo.png`
+type Network = 'ethereum' | 'arbitrum' | 'optimism'
```
justindomingue, line 11: love this typing 👏

src/components/CurrencyLogo/index.tsx
```diff
@@ -25,6 +28,19 @@ const StyledLogo = styled(Logo)<{ size: string }>`
...
+function chainIdToNetwork(networkId: SupportedChainId): Network {
+  switch (networkId) {
+    case SupportedChainId.MAINNET:
+      return 'ethereum'
+    case SupportedChainId.ARBITRUM_ONE:
+      return 'arbitrum'
+    case SupportedChainId.OPTIMISM:
+      return 'optimism'
+    default:
+      return 'ethereum'
```
justindomingue, line 40: tempted to leave this as `undefined` given we default to `ethereum` in `getTokenLogoURL`.

src/components/CurrencyLogo/index.tsx
```diff
@@ -41,7 +57,12 @@ export default function CurrencyLogo({
     if (!currency || currency.isNative) return []
 
     if (currency.isToken) {
-      const defaultUrls = currency.chainId === 1 ? [getTokenLogoURL(currency.address)] : []
+      const defaultUrls = [SupportedChainId.ARBITRUM_ONE, SupportedChainId.MAINNET, SupportedChainId.OPTIMISM].includes(
```
justindomingue, line 60: would it be useful to move this check inside `getTokenLogoURL` so the caller doesn't need to be aware of the impl details?

justindomingue (approved):

PR #2440, opened 2021-09-24
fix: set custom dimension on chainId change

JFrankfurt:
adds a custom dimension to ga tracking w/ current chainId

### DIFF
--- a/src/components/analytics/GoogleAnalyticsReporter.tsx
+++ b/src/components/analytics/GoogleAnalyticsReporter.tsx
@@ -1,3 +1,4 @@
+import { useActiveWeb3React } from 'hooks/web3'
 import { useEffect } from 'react'
 import ReactGA from 'react-ga'
 import { RouteComponentProps } from 'react-router-dom'
@@ -20,6 +21,11 @@ export default function GoogleAnalyticsReporter({ location: { pathname, search }
     getLCP(reportWebVitals)
     getCLS(reportWebVitals)
   }, [])
+
+  const { chainId } = useActiveWeb3React()
+  useEffect(() => {
+    ReactGA.set({ ['Chain ID']: chainId ?? 1 })
+  }, [chainId])
 
   useEffect(() => {
     ReactGA.pageview(`${pathname}${search}`)
--- a/src/components/analytics/GoogleAnalyticsReporter.tsx
+++ b/src/components/analytics/GoogleAnalyticsReporter.tsx
@@ -25,7 +25,7 @@ export default function GoogleAnalyticsReporter({ location: { pathname, search }
 
   const { chainId } = useActiveWeb3React()
   useEffect(() => {
-    ReactGA.set({ ['Chain ID']: chainId ?? SupportedChainId.MAINNET })
+    ReactGA.set({ ['Chain ID']: chainId ?? 0 })
   }, [chainId])
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


src/components/analytics/GoogleAnalyticsReporter.tsx
```diff
@@ -21,6 +22,11 @@ export default function GoogleAnalyticsReporter({ location: { pathname, search }
     getCLS(reportWebVitals)
   }, [])
 
+  const { chainId } = useActiveWeb3React()
+  useEffect(() => {
+    ReactGA.set({ ['Chain ID']: chainId ?? 1 })
```
zzmp, line 27: Should we use a chainId of `0` for default/unconnected?

zzmp (changes requested):

src/components/analytics/GoogleAnalyticsReporter.tsx
JFrankfurt: 1 (mainnet) is the default chainId and rpc connection.

JFrankfurt: w/ `useActiveWeb3React` "active" means the user's wallet connection. If they don't have a connection to the application, we use a fallback network connector that only uses data from mainnet.

zzmp: If a user is connected to mainnet, it will be `1`, but if a user doesn't connect a wallet, it will still be `1` from the fallback connection. AFAIK we don't allow switching from mainnet when using the fallback. Could we use `0` to denote "mainnet with fallback", as opposed to "mainnet with connected user wallet"?

JFrankfurt: @willhennessy Would you prefer 0?

JFrankfurt: Will says 0 works, will add.

src/components/analytics/GoogleAnalyticsReporter.tsx
```diff
@@ -25,7 +25,7 @@ export default function GoogleAnalyticsReporter({ location: { pathname, search }
 
   const { chainId } = useActiveWeb3React()
   useEffect(() => {
-    ReactGA.set({ ['Chain ID']: chainId ?? SupportedChainId.MAINNET })
+    ReactGA.set({ ['Chain ID']: chainId ?? 0 })
```
zzmp, line 28: Should this be in the enum, something like `SupportedChainId.DEFAULT` or `SupportedChainId.DISCONNECTED`?
I'm not sure, just something to consider.

zzmp (approved):

JFrankfurt:
@willhennessy I just checked the analytics and didn't see any custom dimension stuff. Where should this be showing up?

willhennessy:
we should see custom dimensions on this page:  https://analytics.google.com/analytics/web/#/report/content-event-events/a128182339w232099914p226273589/explorer-table.secSegmentId=analytics.customDimension1&explorer-table.plotKeys=%5B%5D/

in column #2.  For some reason the value of the chainId dimension is being logged as "Soundcloud Music Downloader" or "YouTube Music Downloader". Why would that be happening? I would expect 1, 10, 42161, etc

 And the number of events if much lower than I would expect for L1

PR #2441, opened 2021-09-24
add usdc price in TradePrice

### LINKED ISSUE #2389: Add USD price in swap box
example:
1 ETH = 3402 DAI ($3401.63)

justindomingue:
closes #2389

### DIFF
--- a/src/components/swap/TradePrice.tsx
+++ b/src/components/swap/TradePrice.tsx
@@ -1,4 +1,6 @@
+import { Trans } from '@lingui/macro'
 import { Currency, Price } from '@uniswap/sdk-core'
+import useUSDCPrice from 'hooks/useUSDCPrice'
 import { useCallback, useContext } from 'react'
 import { Text } from 'rebass'
 import styled, { ThemeContext } from 'styled-components/macro'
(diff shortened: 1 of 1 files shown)


src/components/swap/TradePrice.tsx
```diff
@@ -10,21 +13,23 @@ interface TradePriceProps {
...
   border: none;
-  height: 24px;
   cursor: pointer;
+  display: grid;
+  height: 24px;
+  justify-content: center;
+  padding: 0;
+  grid-template-columns: 1fr auto;
+  grid-gap: 0.25rem;
 `
```
JFrankfurt, line 26: 🙏

JFrankfurt (approved):

willhennessy:
I think when the USD price is <$1 we are rounding up to an even "$1".  bug?
Screen Shot 2021-09-27 at 10 39 08 AM

willhennessy:
nvm, I see that's the unit price of the input asset so $1 makes sense. 

other nit:  spotted 3 decimal places here. Can we always round to 2 decimals for USD?
Screen Shot 2021-09-27 at 10 43 28 AM

callil:
I think we have the price reversed, it should show the $Value for 1 Unit.

From the examples above it would show:
1 USDC = 0.0003 ETH (~$1.01)
1 ETH = 10.76 AAVE (~$3,149)

PR #2451, opened 2021-09-27
chore: rm lodash

zzmp:
Updates node-vibrant to use a webpack-compatible version.
lodash was implicitly installed through node-vibrant (and its lack of tree-shaking); with this fixed, lodash is no longer provided.


zzmp:
@justindomingue another option is to just install lodash/inRange (or lodash-es) for this use case.

justindomingue (approved):
seems good, i don't think there was a reason to use `inRange` except for readability but doing it manually may be more readable tbh

PR #2452, opened 2021-09-27
chore: use esm for luxon

zzmp:
Using ESM (over CJS) allows for tree-shaking and improves source-mapping.


zzmp:
This is necessary because, although luxon exports a webpack-compatible bundle in the `module` field, it overrides it with the CJS bundle in the `browser` field.

NoahZinsmeister (approved):

PR #2454, opened 2021-09-27
fix: quick fix for broken token select on add

### LINKED ISSUE #2453: Add liquidity page broken - cannot select a second token "Invalid pair"
**Bug Description**
When you load app.uniswap.org and navigate to the add liquidity page, the second token selector is not available and prevents a users from exploring pools and liquidity. 

**Expected behavior** 
I can select a second token and explore pools.

Screen Shot 2021-09-27 at 3 38 39 PM

ianlapham:
- fixes #2453


zzmp (approved):
lgtm, but please open an issue to track this and back it out when the underlying issue (update to null if no token found) is fixed

justindomingue (approved):

PR #2461, opened 2021-09-28
fix: upgrade @davatar/react to 1.6.2, minor fix for ERC token URIs

carlosdp:
There was a small bug identified in the library when ERC721/1155 token URIs were provided in uppercase, while The Graph expects lowercase contract IDs. This patch version fixes that bug and ensures IDs are all in lowercase.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "homepage": ".",
   "private": true,
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


zzmp:
Good catch. Thanks for keeping us up-to-date!

PR #2463, opened 2021-09-28
feat: routing api v2+v3 support & smart order router locally

justindomingue:
- replaces instances of `v2-sdk/trade` and `v3-sdk/trade` with `router-sdk/trade`
- uses `smart-order-router` as client-side routing api
- runs in main thread as perf was worse in a webworker


src/worker/smartOrderRouter.ts
```diff
@@ -0,0 +1,235 @@
...
+const provider = new ethers.providers.JsonRpcProvider(NETWORK_URLS[chainId])
+
+const subgraphProvider = new URISubgraphProvider(chainId, 'https://ipfs.io/ipfs/QmfArMYESGVJpPALh4eQXnjF8HProSF1ky3v8RmuYLJZT4')
+
+const multicall2Provider =
+  new UniswapMulticallProvider(chainId, provider, 375_000)
+
+const poolProvider = new PoolProvider(ID_TO_CHAIN_ID(chainId), multicall2Provider)
+
+const tokenListProvider = new CachingTokenListProvider(chainId, DEFAULT_TOKEN_LIST, new Cache())
```
willpote, line 92: We should probably use the token list selected by the user, and maybe re-initialize the web worker if they change it

src/worker/smartOrderRouter.ts
```diff
@@ -0,0 +1,235 @@
...
+    },
+    {
+      gasLimitOverride: 2_000_000,
+      multicallChunk: 70,
+    }
+  ),
+  gasPriceProvider: new CachingGasStationProvider(chainId, new EIP1559GasPriceProvider(provider), new Cache()),
+  gasModelFactory: new HeuristicGasModelFactory(),
+  blockedTokenListProvider: undefined,
+  tokenProvider: tokenListProvider,
```
willpote, line 122: Using just a tokenListProvider for this means that if the user specifies a token not in the token list, the API will return a 404. Should use a CachingTokenProviderWithFallback (it will check the token list first, then search on-chain if it doesnt find the token, and do caching)

Same for pool provider actually, we should use the caching version

Best to try and match as closely to the config here: https://github.com/Uniswap/routing-api/blob/main/lib/handlers/quote/injector.ts#L204-L286

src/worker/smartOrderRouter.ts
```diff
@@ -0,0 +1,235 @@
...
+  topNWithEachBaseToken: 3,
+  topNWithBaseToken: 6,
+  topNWithBaseTokenInSet: false,
+  maxSwapsPerPath: 3,
+  minSplits: 1,
+  maxSplits: 7,
+  distributionPercent: 5,
+}
+
+class Cache<T> implements ICache<T> {
```
willpote, line 68: Curious about plans for this. Implementing it should increase performance quite a lot

src/worker/smartOrderRouter.ts
```diff
@@ -0,0 +1,235 @@
...
+      quoteMinSuccessRate: 0.15,
+    },
+    {
+      gasLimitOverride: 2_000_000,
+      multicallChunk: 70,
+    }
+  ),
+  gasPriceProvider: new CachingGasStationProvider(chainId, new EIP1559GasPriceProvider(provider), new Cache()),
+  gasModelFactory: new HeuristicGasModelFactory(),
+  blockedTokenListProvider: undefined,
```
willpote, line 121: Reminder to fill in

src/worker/smartOrderRouter.ts
```diff
@@ -0,0 +1,235 @@
...
+}
+
+class MetricLogger extends IMetric {
+  putDimensions() {
+    return
+  }
+
+  putMetric(key: string, value: number, unit?: MetricLoggerUnit) {
+    console.info({ key, value, unit }, `[Metric]: ${key}: ${value} | ${unit ? unit : ''}`)
+  }
```
willpote, line 50: Reminder for react-ga

willpote (commented):
Looks good so far

JFrankfurt (dismissed):

src/utils/memoryCache.test.ts
```diff
@@ -0,0 +1,78 @@
+import { MemoryCache } from './memoryCache'
```
justindomingue, line 1: think can delete

src/utils/memoryCache.ts
```diff
@@ -0,0 +1,45 @@
+import { ICache } from '@uniswap/smart-order-router'
```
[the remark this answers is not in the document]
justindomingue: delete? using `nodecache`

src/utils/memoryCache.ts
```diff
@@ -0,0 +1,45 @@
...
+
```
justindomingue, line 2: delete? nodecache now

src/utils/prices.ts
```diff
@@ -1,61 +1,47 @@
...
-            (currentFee: Percent, pool): Percent =>
-              currentFee.multiply(ONE_HUNDRED_PERCENT.subtract(new Fraction(pool.fee, 1_000_000))),
-            ONE_HUNDRED_PERCENT
-          )
+    const routeRealizedLPFeePercent = overallPercent.multiply(
+      ONE_HUNDRED_PERCENT.subtract(
+        swap.route.pools.reduce<Percent>(
+          (currentFee: Percent, pool): Percent =>
+            currentFee.multiply(
+              ONE_HUNDRED_PERCENT.subtract(new Fraction(pool instanceof Pair ? FeeAmount.MEDIUM : pool.fee, 1_000_000))
```
justindomingue, line 29: is this correct to calculate routes for either v2 or v3?

src/hooks/useUSDCPrice.ts
```diff
@@ -45,7 +46,8 @@ export default function useUSDCPrice(currency?: Currency): Price<Currency, Token
       const { numerator, denominator } = v2USDCTrade.route.midPrice
       return new Price(currency, stablecoin, denominator, numerator)
     } else if (v3USDCTrade.trade) {
-      const { numerator, denominator } = v3USDCTrade.trade.route.midPrice
+      // TODO(judo): is looking at a single midprice enough?
+      const { numerator, denominator } = v3USDCTrade.trade.routes[0].midPrice
```
willpote, line 50: This is using the old client side routing logic right? If so I don't understand the TODO since the old logic doesn't do split routes

If we change this to use the API / client side router I could see us having this issue (though we could use maxSplits = 1 when getting this price to get around it)

src/state/routing/types.ts
```diff
@@ -21,6 +20,27 @@ export type PoolInRoute = {
   fee: string
   amountIn?: string
   amountOut?: string
+
+  // no used in the interface
+  address?: string
```
willpote, line 25: Address is always returned by the api

src/state/routing/types.ts
```diff
@@ -21,6 +20,27 @@ export type PoolInRoute = {
...
+  type: 'v2-pool'
+  tokenIn: TokenInRoute
+  tokenOut: TokenInRoute
+  reserve0: V2Reserve
+  reserve1: V2Reserve
+  amountIn?: string
+  amountOut?: string
+
+  // no used in the interface
+  address?: string
```
willpote, line 43: .

src/state/routing/useRoutingAPITrade.ts
```diff
@@ -1,13 +1,22 @@
...
+import { TradeState } from './types'
+import { computeRoutes, transformRoutesToTrade } from './utils'
+
+/** Picks required properties of a `Token` to make it serializable */
+const tokenToSerializable = ({ address, chainId, symbol, decimals }: Token) => ({
+  address,
+  chainId,
+  symbol,
+  decimals,
+})
```
willpote, line 19: Do we still need this?

src/components/swap/SwapRoute.tsx
```diff
@@ -22,16 +23,18 @@ const Separator = styled.div`
   width: 100%;
 `
 
-const V2_DEFAULT_FEE_TIER = 3000
+const V2_DEFAULT_FEE_TIER = FeeAmount.MEDIUM
```
ianlapham, line 26: small nit i still am not a fan of low medium high - but for this release seems okay

src/constants/addresses.ts
```diff
@@ -76,6 +76,8 @@ export const SOCKS_CONTROLLER_ADDRESSES: AddressMap = {
   [SupportedChainId.MAINNET]: '0x65770b5283117639760beA3F867b69b3697a91dd',
 }
 export const SWAP_ROUTER_ADDRESSES: AddressMap = constructSameAddressMap('0xE592427A0AEce92De3Edee1F18E0157C05861564', [
+  //TODO: set new swap router addresses
```
ianlapham, line 79: can we delete this?

ianlapham (commented):
some small nits

src/state/routing/types.ts
justindomingue: It is but the client side smart-order-router needs a provider to generate it and it's not actually used so I made it optional here.

src/state/routing/useRoutingAPITrade.ts
justindomingue: just a utility to make tokens serializable in the redux store

src/components/swap/SwapRoute.tsx
justindomingue: lol yeah i dislike it too especially that we have a lowest tier now.. :/ but definitely a big change

PR #2468, opened 2021-09-29
perf: lazy load vote related routes

chenxsan:
See the related comment here https://github.com/Uniswap/interface/pull/2391#issuecomment-928004001. cc @zzmp Let me know if I get your comment right.


src/pages/App.tsx
```diff
@@ -86,8 +87,10 @@ export default function App() {
             <Polling />
             <TopLevelModals />
             <Switch>
-              <Route exact strict path="/vote" component={Vote} />
-              <Route exact strict path="/vote/:governorIndex/:id" component={VotePage} />
+              <Suspense fallback={<Loader />}>
```
chenxsan, line 90: I'm reusing the `<Loader />` component here.

zzmp: @callil is this an appropriate loading element?

src/pages/App.tsx
```diff
@@ -86,8 +87,10 @@ export default function App() {
...
+                <Route exact strict path="/vote" component={Vote} />
+              </Suspense>
```
zzmp, line 92: Considering that other routes may be lazy-loaded in the future, I think it's appropriate to wrap the Switch in Suspense, instead of just the route. wdyt?

src/pages/App.tsx
```diff
@@ -86,8 +87,10 @@ export default function App() {
             <Polling />
             <TopLevelModals />
             <Switch>
-              <Route exact strict path="/vote" component={Vote} />
-              <Route exact strict path="/vote/:governorIndex/:id" component={VotePage} />
+              <Suspense fallback={<Loader />}>
+                <Route exact strict path="/vote" component={Vote} />
```
zzmp, line 91: I think `exact` prevents this from matching for sub-paths of `/vote`, and needs to be omitted.

src/pages/App.tsx
```diff
@@ -122,7 +125,6 @@ export default function App() {
               <Route exact strict path="/migrate/v2" component={MigrateV2} />
               <Route exact strict path="/migrate/v2/:address" component={MigrateV2Pair} />
 
-              <Route exact strict path="/create-proposal" component={CreateProposal} />
```
zzmp, line 125: I like that you've moved this under `/vote`, but it will no longer match. You'll need to add a redirect in App.tsx from /create-proposal -> /vote/create-proposal (or /vote/propose, or /vote/create), and change the route under `/vote` to the redirected-to path.

src/pages/Vote/index.tsx
```diff
@@ -11,8 +11,9 @@ import DelegateModal from 'components/vote/DelegateModal'
 import ProposalEmptyState from 'components/vote/ProposalEmptyState'
 import { useActiveWeb3React } from 'hooks/web3'
 import JSBI from 'jsbi'
+import CreateProposal from 'pages/CreateProposal'
```
zzmp, line 14: @moodysalem what would be the appropriate file structure for nested pages? As-is, they are all top-level files. Would it be ok to nest them under vote, ie renaming them to `pages/vote/index.ts`, `pages/vote/Governance.tsx`, `pages/vote/Vote.tsx`, `pages/vote/Propose.tsx`?

zzmp (changes requested):
The deployed vercel app is not loading the swap page for me, do you know why?

This looks great, just needs some additional file cleanup and a redirect put in.

src/pages/App.tsx
chenxsan: Sure, I just assumed you might want to specify different loading indicators for different sections.

src/pages/App.tsx
```diff
@@ -85,46 +86,54 @@ export default function App() {
...
-
-              <Route exact strict path="/create-proposal" component={CreateProposal} />
-              <Route component={RedirectPathToSwapOnly} />
-            </Switch>
+            <Suspense fallback={<Loader />}>
+              <Switch>
+                <Route strict path="/vote" component={Vote} />
+                <Route exact strict path="/create-proposal">
+                  <Redirect to="/vote/create-proposal" />
+                </Route>
```
chenxsan, line 94: I add a `Redirect` here.

src/pages/Vote/index.tsx
```diff
@@ -122,136 +124,140 @@ export default function Vote() {
...
-                <RowBetween>
-                  <TYPE.white fontWeight={600}>
-                    <Trans>Uniswap Governance</Trans>
-                  </TYPE.white>
-                </RowBetween>
-                <RowBetween>
-                  <TYPE.white fontSize={14}>
+      <Route exact strict path="/vote/:governorIndex/:id" component={VotePage} />
+      <Route exact strict path="/vote/create-proposal" component={CreateProposal} />
+      <Route exact strict path="/vote">
```
chenxsan, line 129: I'm adding a `<Route />` here so those elements won't get rendered for other routes like `/vote/1/2` or `/vote/create-proposal`.

chenxsan:
> The deployed vercel app is not loading the swap page for me, do you know why?
> 
> This looks great, just needs some additional file cleanup and a redirect put in.

I can't reproduce it on my computer.

zzmp:
Hi @chenxsan, sorry for my delay. Can you move the Vote/index component into its own file, so that the routes are the  only thing in the index file? Maybe something like Vote/Landing.tsx (although you're welcome to use another name).

Once that's done, this should be ready to merge! Just make sure to merge the current main branch back in so that it's up-to-date.

zzmp (approved):

PR #2485, opened 2021-10-03
upgrade @davatar/react to 1.8.0 for performance improvements

carlosdp:
## Description

This PR updates the [@davatar/react](https://github.com/TBDAO/davatar-helpers/tree/master/packages/react) package to 1.8.0, which includes performance improvements for avatar loading. The library now uses more direct RPC calls for token metadata loading, and caches avatars for 24 hours in local storage to avoid refetching avatars all the time.

## Security / Supply Chain risk consideration
The `@davatar/react` library uses minimal dependencies and is exact version matched. This new version adds `@ethersproject/contracts` and `bn.js`, which are already included in Uniswap, so no net increase in dependencies.

## List of features added/changed

- Upgrade `@davatar/react` to 1.8.0

## How Has This Been Tested?

Connected my wallet in the local dev environment and checked that it rendered avatars correctly.


zzmp:
It looks like the github repo is only at 1.7.0. Where is the source code for 1.8.0?

carlosdp:
Woops, the push had failed during release. Code is in the repo now: https://github.com/TBDAO/davatar-helpers/tree/v1.8.0

zzmp:
The repo doesn't describe the package's use of localStorage at all, and I want to make sure it's not going to interact with other things.

Are you planning on updating your README? Adding a roadmap? Is this still in active development, or should we expect this version to stick for a bit?

carlosdp:
@zzmp I've added a note about the avatar caching behavior, which uses localStorage, into the [README](https://github.com/TBDAO/davatar-helpers/tree/master/packages/react#caching). I also confirmed the library doesn't conflict with any other local storage usage in the Uniswap app, but also made a small patch change to fully namespace keys to ensure no future conflict!

Not sure what else would be missing or out of date. There's no roadmap at the moment because the library is meant to be tightly scoped: just show ENS avatars based on the spec developed and now being pushed by the ENS team.

Recent changes were from unforeseen complications with how to properly query NFT data and such, as the spec is just now being implemented everywhere. At this point, the library has been simplified down enough and tested that I wouldn't expect anymore upgrades for a while. There are also no plans to increase the scope of the library at all at present.

If there's anything else I can do to alleviate any concerns, please let me know! 😃  Uniswap is critical infrastructure for Ethereum, I'm trying to do my best to make sure we respect that and ensure we can't interrupt normal operation with this library, with fail-safes and just keeping it super small in scope.

PR #2516, opened 2021-10-08
fix: replace `t` with `<Trans>` where possible

justindomingue:


moodysalem:
is there a linter error we can add for importing `t` from the macro?

Issue #2519, opened 2021-10-08
show usdc price in add liquidity below upper and lower price tickers


justindomingue:
example:
1 ETH = 3402 DAI ($3401.63)

willhennessy:
Thanks prakash!  Here is a design spec from our UX designer. We recommend putting the USD value in parentheses on the same line as the token price, just above the bar chart.

<img width="759" alt="Screen Shot 2021-10-15 at 3 08 19 PM" src="https://user-images.githubusercontent.com/5455577/137781169-697d7488-b4ce-41f2-9564-99157af48767.png">

justindomingue:
decision is to show dollar price for ETH and when either token is a stable coin (we do not have a strong list rn, so can probably just do for DAI,USDC,USDT?)

tinaszheng:
@willhennessy @justindomingue it looks like the USD price we're displaying is for the base currency - so for a ETH-DAI pairing we are displaying the USD value of DAI (is this understanding correct?) 

It seems like from the copy, e.g. "3820.76 DAI per ETH ($3,900)" it's suggesting that is the ETH price? Thoughts on changing it to something like "3820.76 DAI ($3,900) per ETH"?

callil:
I had the same idea @tinaszheng! maybe we can see it in practice and review it.

tinaszheng:
@callil yea! here are the two side by side for comparison :D i guess either looks fine imo, as the savvy user can probably figure out what's going on (famous last words tho)

![Screen Shot 2021-12-15 at 2 28 51 PM](https://user-images.githubusercontent.com/59578595/146274827-2a123b6a-2d84-42ad-9b13-091db649508a.png)
![Screen Shot 2021-12-15 at 2 29 36 PM](https://user-images.githubusercontent.com/59578595/146274840-a0d31f50-8b2d-42ee-bb99-e4338808263c.png)

PR #2619, opened 2021-10-15
fix: #2519 - adds USDC price for currencies in liquidity pool creator

prakashsanker:
1. Not sure if I should be using `currencyA` or `currencyB`
2. Now sure what the use case is - could someone elaborate?
3. Screenshot
Screen Shot 2021-10-13 at 4 04 46 PM


prakashsanker:
Fixes #2519

prakashsanker:
@justindomingue  bumping this!

src/components/InputStepCounter/InputStepCounter.tsx
```diff
@@ -140,6 +144,11 @@ const StepCounter = ({
     }
   }, [localValue, useLocalValue, value])
 
+  const baseCurrencyPriceInUSDC = useUSDCPrice(currencyB ?? undefined)
```
justindomingue, line 147: sorry to be pedantic but this is a bit hard to follow. Could you move it to a custom hook  at the top of this file? Something like

```
function useBaseCurrencyAmount(baseCurrency: Currency | undefined, quoteAmount: number) {
  const currencyUSDCPrice = useUSDCPrice(baseCurrency ?? undefined)
  if (!currencyUSDCPrice) return undefined
  return Number(currencyUSDCPrice.toSignificant(6)) * quoteAmount
}
```

src/pages/AddLiquidity/index.tsx
```diff
@@ -137,7 +137,10 @@ export default function AddLiquidity({
     baseCurrency ?? undefined,
     existingPosition
   )
-
+  const usdcBaseCurrencyPriceInNumber = Number(baseCurrencyPriceInUSDC?.toSignificant(4))
```
justindomingue, line 140: this could also be in its own hook for readability

justindomingue (commented):
@willhennessy have we settled on an approach?

src/constants/tokens.ts
```diff
@@ -165,6 +165,8 @@ export const UNI: { [chainId: number]: Token } = {
   [SupportedChainId.KOVAN]: new Token(SupportedChainId.KOVAN, UNI_ADDRESS[42], 18, 'UNI', 'Uniswap'),
 }
 
+export const STABLECOINS = [DAI.symbol, USDC.symbol, USDT.symbol]
```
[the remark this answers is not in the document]
tinaszheng: oooh TIL about symbols not being unique. ty!!

src/utils/index.ts
```diff
@@ -55,3 +56,10 @@ export function isTokenOnList(tokenAddressMap: TokenAddressMap, token?: Token):
 export function formattedFeeAmount(feeAmount: FeeAmount): number {
   return feeAmount / 10000
 }
+
+export function isEthStablecoinPair(currencyA: Currency | undefined, currencyB: Currency | undefined) {
+  const isEthAndStable = currencyA?.isNative && STABLECOINS.includes(currencyB?.symbol)
```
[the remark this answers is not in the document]
tinaszheng: ooh if we should return for stable-stable pairs it might be easier to just to something like `isEthOrStable(currencyA) && isEthOrStable(currencyB)`

src/components/PositionPreview/index.tsx
```diff
@@ -1,20 +1,19 @@
...
-import { Break } from 'components/earn/styled'
+import { Trans } from '@lingui/macro'
 import { Currency } from '@uniswap/sdk-core'
-import CurrencyLogo from 'components/CurrencyLogo'
-import DoubleCurrencyLogo from 'components/DoubleLogo'
-import JSBI from 'jsbi'
-import { LightCard } from 'components/Card'
 import { Position } from '@uniswap/v3-sdk'
 import RangeBadge from 'components/Badge/RangeBadge'
+import { LightCard } from 'components/Card'
```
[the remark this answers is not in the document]
tinaszheng: yeah i think the sort order has since changed 🤔 looks diff on main

src/components/InputStepCounter/InputStepCounter.tsx
```diff
@@ -144,10 +144,8 @@ const StepCounter = ({
     }
   }, [localValue, useLocalValue, value])
 
-  const baseCurrencyPriceInUSDC = useUSDCPrice(currencyB ?? undefined)
-  const baseCurrencyPriceInUSDCNumber = Number(baseCurrencyPriceInUSDC?.toFixed(4))
-  const baseCurrencyPriceIsNotNaN = !isNaN(baseCurrencyPriceInUSDCNumber)
   const quoteCurrencyAmount = Number(value)
+  const baseCurrencyUSDPrice = useCurrencyUSDPrice(currencyB ?? undefined, quoteCurrencyAmount)
```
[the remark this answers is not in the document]
tinaszheng: ooh hmm a few typing issues are making this difficult to use the `CurrencyAmount` format: `currencyB` might be undefined, and `quoteCurrencyAmount` is a decimal instead of a `BigInt` as the constructor expects. Maybe cleaner to do it without `CurrencyAmount`?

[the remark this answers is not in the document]
tinaszheng: oooh nvm, i found a helper function that gives me what i need: `tryParsedAmount` which converts the input string into `CurrencyAmount`!

src/pages/AddLiquidity/index.tsx
```diff
@@ -137,11 +138,12 @@ export default function AddLiquidity({
     baseCurrency ?? undefined,
     existingPosition
   )
-  const usdcBaseCurrencyPriceInNumber = Number(baseCurrencyPriceInUSDC?.toSignificant(4))
-  const quoteCurrencyPrice =
-    usdcBaseCurrencyPriceInNumber * Number(invertPrice ? price?.invert().toSignificant(4) : price?.toSignificant(4))
-  const roundedQuoteCurrencyPrice = isNaN(usdcBaseCurrencyPriceInNumber) ? undefined : quoteCurrencyPrice.toFixed(2)
-  const showBaseCurrencyPrice = isEthStablecoinPair(baseCurrency ?? undefined, quoteCurrency ?? undefined)
+
+  const quoteCurrencyAmount = invertPrice ? price?.invert().toSignificant(6) : price?.toSignificant(6)
```
tinaszheng, line 142: would love another set of eyes on this math here 👀  i believe this is accomplishing the same thing as before, but simpler?

src/components/InputStepCounter/InputStepCounter.tsx
```diff
@@ -140,6 +146,10 @@ const StepCounter = ({
     }
   }, [localValue, useLocalValue, value])
 
+  const baseCurrencyUSDPrice = useCurrencyUSDPrice(tryParseAmount(value, currencyB))
+  const showBaseCurrencyUSDPrice =
+    isEthOrStablecoin(currencyA ?? undefined) && isEthOrStablecoin(currencyB ?? undefined)
```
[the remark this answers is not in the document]
tinaszheng: ah yeah i think `currencyA` could be `null` so it's to cast that into `undefined` in those cases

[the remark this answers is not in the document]
tinaszheng: ah good catch!

src/components/InputStepCounter/InputStepCounter.tsx
```diff
@@ -140,6 +146,9 @@ const StepCounter = ({
     }
   }, [localValue, useLocalValue, value])
 
+  const baseCurrencyUSDPrice = useCurrencyUSDPrice(tryParseAmount(value, currencyB))
+  const showBaseCurrencyUSDPrice = isEthOrStablecoin(currencyA) && isEthOrStablecoin(currencyB)
```
justindomingue, line 150: i would probably skip running `useCurrencyUSDPrice` if `showBaseCurrencyUSDPrice` is false by passing `undefined` to the hook in that case?

justindomingue: then L190 only needs to check `baseCurrencyUSDPrice` is defined

src/hooks/useCurrencyUSDPrice.ts
```diff
@@ -0,0 +1,9 @@
+import { Currency, CurrencyAmount } from '@uniswap/sdk-core'
```
justindomingue, line 1: isn't this `useUSDValue`? 

https://github.com/Uniswap/interface/blob/main/src/hooks/useUSDCPrice.ts#L58

src/pages/AddLiquidity/index.tsx
```diff
@@ -137,6 +139,12 @@ export default function AddLiquidity({
     existingPosition
   )
 
+  const quoteCurrencyAmount = invertPrice ? price?.invert().toSignificant(6) : price?.toSignificant(6)
+  const quoteCurrencyPrice = useCurrencyUSDPrice(tryParseAmount(quoteCurrencyAmount, quoteCurrency ?? undefined))
+
+  const showQuoteCurrencyPrice =
```
justindomingue, line 145: ditto. could probably skip `useCurrencyUSDPrice` if this is false

src/hooks/useCurrencyUSDPrice.ts
tinaszheng: my god yes it is

tinaszheng: thank u 🤯

BigsteppA1:
Ok

tinaszheng:
after playing around with the change, it feels like showing the USDC price isn't as helpful since only showing for stablecoin pairs means that it will roughly always match the stablecoin amount anyway. going to close this issue for now - sorry @prakashsanker and thank you for your contribution!

PR #2634, opened 2021-10-18
fix: display common bases logos on l2

### LINKED ISSUE #2185: Add Common Bases on L2 Networks
On Optimism and Arbitrum, the interface does not present any "Common Bases" aside from ETH

It should recommend the same as L1.
* If L1 uses chain data to pick popular pairs, then go with that on L2 as well
* If L1 hardcodes popular pairs, go with that.  (ETH, DAI, USDC, USDT, WBTC, WETH)

justindomingue:
common bases are generated statically. this fetches the right token from the active lists

closes #2185


justindomingue:
@willhennessy created #2640 and #2641

PR #2649, opened 2021-10-20
fix: restrict @davatar usage to avoid 3p fetches

zzmp:
See https://github.com/metaphor-xyz/davatar-helpers/issues/18


zzmp:
This should be sufficient to stop 3p requests, but it is based on inspection of the code.
If you find that it is not sufficient, we should revert @davatar/react to use @metamask/jazzicon until this can be fixed by @davatar/react.

PR #2652, opened 2021-10-20
fix(L2): ensure chainIds match before fetching pool data

JFrankfurt:


moodysalem:
i think the fact that we're passing currencies from different chains to this function might have to do with where we're doing the debounces, so i'd rather see that fixed. working around invariants like this is usually code smell, and you can also see the debounce issues in the ux when you change input values and the impacts immediately change

moodysalem:
> > with where we're doing the debounces
> 
> @moodysalem can you expand on this? The data flow here (as far as I've worked out) is:
> 
> * initial swap reducer state parsed from query string
> * user input dispatched to store from via `useSwapActionHandlers` (line 163-178 in `Swap/index.ts`
> * user changes `chainId`
> * swap store state can now fail invariant requirement due to sortsBefore chainId requirement making, for example, the selected Arbitrum DAI mismatch the newly selected ETH.
> 
> I'm not sure where/how a debouncing system works into this.

how is the arbitrum DAI currency/currency amount being produced when the state only contains an address and the hook constructs the token? seems like the swap state of `currencyId: arbitrum_dai_address` should result in an undefined token when you switch chains

PR #2655, opened 2021-10-20
fix: Update walletlink-connector to 6.2.8

brendanww:
Updates walletlink-connector to 6.2.8 which bumps walletlink to a version that handles addEthereumChain+switchEthereumChain requests

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "@web3-react/injected-connector": "^6.0.7",
     "@web3-react/portis-connector": "^6.0.9",
     "@web3-react/walletconnect-connector": "^7.0.2-alpha.0",
-    "@web3-react/walletlink-connector": "^6.2.6",
+    "@web3-react/walletlink-connector": "^6.2.7",
     "ajv": "^6.12.3",
     "array.prototype.flat": "^1.2.4",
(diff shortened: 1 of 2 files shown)


brendanww:
@moodysalem would it be possible to get some 👀 on this?

moodysalem:
> @moodysalem would it be possible to get some 👀 on this?

it looks fine but what's with all the extra deps
https://bundlephobia.com/package/walletlink@2.2.6
vs 
https://bundlephobia.com/package/walletlink@2.1.6

brendanww:
bundlephobia can be a bit misleading. walletlink used to depend on nodejs polyfills (eg crypto-browserify) that wouldn't show up on bundlephobia but would still increase the size of the application bundle. we still have a few polyfills left to replace, and if there is not a web native api we can use, then the bundle size will have appeared to increase on bundlephobia.

`js-sha256` is a lib we added to help us move off `crypto-browserify`. the equivalent web api is async and would have forced a riskier refactor.

the only other lib that stands out in the diff of those two links is `readable-stream` which comes in through `stream-browserify` which was formerly an implicit dependency added by build tools like webpack4.

PR #2656, opened 2021-10-20
added support for fetching tokenlins from the ceramic network 

lucasespinosa28:
the ceramic network is open-source platform for creating, hosting, and sharing streams of data.

I added support to fetch token list storage in ceramic
Screenshot from 2021-10-20 19-48-51
c

### DIFF
--- a/package.json
+++ b/package.json
@@ -170,5 +170,6 @@
       "last 1 safari version"
     ]
   },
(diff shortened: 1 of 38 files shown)


moodysalem:
Are there any token list maintainers that want to host on ceramic? Or, is there a reason to recommend ceramic over ENS + IPFS?

Issue #2692, opened 2021-10-26
WalletConnect modal doesn't open on retry after cancelling 1st attempt


M0kY:
**Bug Description**
When trying to connect a wallet through WalletConnect, in case of user rejection (closing the modal with QR code) it is impossible to reopen the modal with the QR without a page refresh. After closing the WalletConnect modal and clicking the "Try again" button it just gets stuck at "Initializing...". The same happens even after completely closing the wallet modal and trying again to connect through WalletConnect.

**Steps to Reproduce**

1. Click on "Connect Wallet"
2. Click on "WalletConnect" option
3. Click on "X" to close QR code modal
4. Click on "Try again" to reopen QR code modal
5. Modal doesn't reopen, app is stuck at "Initializing..."
6. Close wallet modal
7. Click on "Connect Wallet"
8. Click on "WalletConnect" option
9. App is still stuck at "Initializing..."

**Expected Behavior**
I expect the QR code modal to show every time I retry connecting with WalletConnect.

**Additional Context**

![initializing](https://user-images.githubusercontent.com/46133205/138870001-76af9f06-38b0-4ea8-a8f9-a938608dc7f8.PNG)

The issue is in `src\components\WalletModal\index.tsx` line 198.

```js
if (connector instanceof WalletConnectConnector && connector.walletConnectProvider?.wc?.uri) {
      connector.walletConnectProvider = undefined
}
```
`connector.walletConnectProvider?.wc` and `walletConnectProvider?.wc?.uri` are always `undefined` and the resetting is never executed.

PR #2693, opened 2021-10-26
fix: walletconnect modal reopen after user rejection

M0kY:
This change fixes Uniswap/interface#2692


willhennessy:
thanks for the PR! Justin can you please review because this is near the wallet connect code you shipped last week? https://github.com/Uniswap/interface/issues/2692

PR #2696, opened 2021-10-26
fix: Add trailing slash to L2 info links

willhennessy:
Some links were broken. For example on /pools/ page click the 'Top Pools' CTA. It would mistakenly direct you to info.uniswap.org/optimismpools instead of optimism/pools


willhennessy:
@JFrankfurt ptal

Issue #2130, opened 2021-08-02
Link to the Help Center


willhennessy:
Now that help.uniswap.org is live, we should link to it in a few places

TODO: confirm the official name with Cal. 'Help Center'?

App
- add 'Help Center' in the three dot dropdown menu right after 'About'

Org
- in 'Community' dropdown, add 'Help Center' as the first item
- update the big FAQ button [TODO: Will to write new FAQ in help site for this]
- further down the page in the "A global community." section, add a box for the Help Center

willhennessy:
@callil how is this for a quick Help Center integration spec?

PR #2709, opened 2021-10-28
feat: Menu update. Add help center & feature requests. Remove analytics & github.

willhennessy:
@callil ptal I added Help Center to the menu. It got so long that a scroll bar appeared (see screenshot), so I deleted Analytics because we have that in the Chart tab in the header. Wdyt?

Other options
- we could also extend the height of the menu with no scroll bar
- we could selectively hide the Claim UNI button, but that's a bit more complex

for a quick fix this seems optimal

https://github.com/Uniswap/interface/issues/2130

### DIFF
--- a/src/components/Menu/index.tsx
+++ b/src/components/Menu/index.tsx
@@ -13,6 +13,7 @@ import {
   Code,
   FileText,
   Globe,
+  HelpCircle,
   Info,
   MessageCircle,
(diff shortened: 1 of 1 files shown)


willhennessy:
Do not merge yet. Looking into Canny price tiers

callil (approved):

PR #2713, opened 2021-10-29
feat: Flashbots Protect RPC mistX

Dire-0x:
- On block change, check if the user is using the protection RPC.
- Display protection status in header
- Display text in the `TransactionSubmittedContent` telling the user not to change the RPC while the tx is pending.

TODO:

We need to handle the FAILED state for a tx sent through the protection RPC. There is no `SerializableTransactionReceipt` for a tx that fails off-chain. I don't want to make any changes to the way you are handling state, can you suggest how you would like this to be handled?


src/hooks/useIsFlashbotsProtectRPC.ts
```diff
@@ -0,0 +1,42 @@
+import { Contract } from '@ethersproject/contracts'
+import { useEffect, useState } from 'react'
+
+import { useBlockNumber } from '../state/application/hooks'
+import { useActiveWeb3React } from './web3'
+
+export default function useIsFlashbotsProtectRPC(): [boolean, () => Promise<boolean>] {
```
moodysalem, line 7: you should use our `useSingleCallResult` hook

Dire-0x: Is infura used to make the calls through the multicall hooks? I had tried using `useSingleCallResult` previously but the result from the call was always false. In order for the call to return true the call needs to be made through MetaMask while the Flashbots Protect RPC is in use.

Dire-0x: The RPC intercepts `FlashbotsRPCContract.isFlashRPC()` calls and returns `true` when calls through another RPC would return `false`. Since the multicall pattern batches these calls and routes them through a multicall contract the RPC cannot detect this call and intercept it.

src/state/transactions/updater.tsx
```diff
@@ -15,6 +16,23 @@ interface TxInterface {
   lastCheckedBlockNumber?: number
 }
 
+export function checkPrivateTxStatusFailed(txHash: string): Promise<boolean> {
+  return new Promise((resolve) => {
+    axios.get(`https://protection-api.flashbots.net/tx/${txHash}`).then((result) => {
+      if (
+        result &&
+        result.data &&
+        result.data.status &&
```
JFrankfurt, line 25: can you do `result?.data?.status === x` here?

src/components/Header/FlashbotsProtectStatus.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+export default function FlashbotsProtect() {
+  const { chainId, library } = useActiveWeb3React()
+  const node = useRef<HTMLDivElement>()
+  const [isFlashRPC, detectFlashRPC] = useIsFlashbotsProtectRPC()
+  const open = useModalOpen(ApplicationModal.FLASHBOTS_PROTECT)
+  const toggle = useToggleModal(ApplicationModal.FLASHBOTS_PROTECT)
+  useOnClickOutside(node, open ? toggle : undefined)
+
+  const showSelector = Boolean(chainId === 1)
```
justindomingue, line 143:
```suggestion
  const showSelector = Boolean(chainId === SupportedChainId.MAINNET)
```

src/components/Header/FlashbotsProtectStatus.tsx
```diff
@@ -0,0 +1,187 @@
...
+          <ShieldLogo src={shieldLogo} />
+        </ShieldWrapper>
+        <SelectorLabel>MEV Shield</SelectorLabel>
+        {showSelector && <StyledChevronDown />}
+      </SelectorControls>
+      {open && (
+        <FlyoutMenu>
+          <FlyoutHeader>
+            <Trans>
+              Sandwich Protection: <StatusText active={isFlashRPC}>{isFlashRPC ? 'ON' : 'OFF'}</StatusText>
```
justindomingue, line 170: probably need to split up into two `<Trans>` blocks to properly extract 'on' and 'off'

src/hooks/useIsFlashbotsProtectRPC.ts
```diff
@@ -0,0 +1,42 @@
...
+  }, [blockNumber])
+
+  async function detectFlashRPC(): Promise<boolean> {
+    console.log('detect start')
+    if (!library || !chainId || chainId !== 1) {
+      setIsFlashRPC(false)
+      return false
+    }
+    const flasbotsRPCContract = new Contract(
+      '0xf1a54b0759b58661cea17cff19dd37940a9b5f1a',
```
justindomingue, line 24: contract addresses are typically put in https://github.com/Uniswap/interface/blob/main/src/constants/addresses.ts

src/hooks/useIsFlashbotsProtectRPC.ts
```diff
@@ -0,0 +1,42 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [blockNumber])
+
+  async function detectFlashRPC(): Promise<boolean> {
+    console.log('detect start')
+    if (!library || !chainId || chainId !== 1) {
+      setIsFlashRPC(false)
+      return false
+    }
+    const flasbotsRPCContract = new Contract(
```
justindomingue, line 23: `useContract` as a convenience hook https://github.com/Uniswap/interface/blob/main/src/hooks/useContract.ts

src/hooks/useIsFlashbotsProtectRPC.ts
```diff
@@ -0,0 +1,42 @@
...
+import { useBlockNumber } from '../state/application/hooks'
+import { useActiveWeb3React } from './web3'
+
+export default function useIsFlashbotsProtectRPC(): [boolean, () => Promise<boolean>] {
+  const [isFlashRPC, setIsFlashRPC] = useState<boolean>(false)
+  const { library, chainId } = useActiveWeb3React()
+  const blockNumber = useBlockNumber()
+
+  useEffect(() => {
+    detectFlashRPC()
```
justindomingue, line 13: is there no other way to detect an rpc change?

src/state/transactions/updater.tsx
```diff
@@ -15,6 +16,23 @@ interface TxInterface {
   lastCheckedBlockNumber?: number
 }
 
+export function checkPrivateTxStatusFailed(txHash: string): Promise<boolean> {
+  return new Promise((resolve) => {
```
justindomingue, line 20: we typically use `rtk-query` over fetching manually

https://redux-toolkit.js.org/rtk-query/overview

would involve creating a new api slice to interact with `api.flashbots.net`, similar to https://github.com/Uniswap/interface/blob/main/src/state/data/slice.ts

src/state/transactions/updater.tsx
```diff
@@ -118,13 +136,21 @@ export default function Updater(): null {
                 dispatch(updateBlockNumber({ chainId, blockNumber: receipt.blockNumber }))
               }
             } else {
+              checkPrivateTxStatusFailed(hash).then((isFailed) => {
+                // Set transaction status as failed or dropped
```
justindomingue, line 140: hm maybe we can add a field to `TransactionDetails` to track this status?

https://github.com/Uniswap/interface/blob/main/src/state/transactions/reducer.ts#L15

interface TransactionDetails {
  receipt?: Serializable..
  privateTransactionStatus: 'idle' | 'failed'
}

or something like that, then we can add a new action in the slice to set private tx status?

cc @moodysalem

src/hooks/useIsFlashbotsProtectRPC.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (!library || !chainId || chainId !== 1) {
+      setIsFlashRPC(false)
+      return false
+    }
+    const flasbotsRPCContract = new Contract(
+      '0xf1a54b0759b58661cea17cff19dd37940a9b5f1a',
+      [
+        {
+          inputs: [],
+          name: 'isFlashRPC',
```
justindomingue, line 28: if you do use a hook then this property bag can probably be move outside of the function body so it doesn't get redefined every render cycle

justindomingue (changes requested):

vm:
closing as this was superseded by https://github.com/Uniswap/interface/pull/3606

PR #2721, opened 2021-11-01
fix(L2): block L2 tokens explicitly linked to L1 tokens that are blocked

JFrankfurt:
uses this pattern: https://github.com/Uniswap/token-lists/issues/51

see also: https://github.com/ethereum-optimism/ethereum-optimism.github.io/pull/35/files

### DIFF
--- a/src/hooks/Tokens.ts
+++ b/src/hooks/Tokens.ts
@@ -1,10 +1,9 @@
 import { arrayify } from '@ethersproject/bytes'
 import { parseBytes32String } from '@ethersproject/strings'
 import { Currency, Token } from '@uniswap/sdk-core'
-import { SupportedChainId } from 'constants/chains'
+import { CHAIN_INFO, L2_CHAIN_IDS, SupportedChainId, SupportedL2ChainId } from 'constants/chains'
 import { ARBITRUM_LIST, OPTIMISM_LIST } from 'constants/lists'
 import { useMemo } from 'react'
-import { useAppSelector } from 'state/hooks'
 
 import { createTokenFilterFunction } from '../components/SearchModal/filtering'
 import { ExtendedEther, WETH9_EXTENDED } from '../constants/tokens'
@@ -59,22 +58,24 @@ export function useAllTokens(): { [address: string]: Token } {
   return useTokensFromMap(allTokens, true)
 }
 
-interface BridgeInfo {
-  [chainId: SupportedChainId | number | string]: {
+type BridgeInfo = Record<
+  SupportedChainId,
+  {
     tokenAddress: string
     originBridgeAddress: string
     destBridgeAddress: string
   }
-}
+>
+
 export function useUnsupportedTokens(): { [address: string]: Token } {
   const { chainId } = useActiveWeb3React()
-  const listsByUrl = useAppSelector((state) => state.lists.byUrl)
+  const listsByUrl = useAllLists()
   const unsupportedTokensMap = useUnsupportedTokenList()
   const unsupportedTokens = useTokensFromMap(unsupportedTokensMap, false)
 
   // checks the default L2 lists to see if `bridgeInfo` has an L1 address value that is unsupported
-  const L2InferredBlockedTokens: { [address: string]: Token } = useMemo(() => {
-    if (!chainId || ![SupportedChainId.ARBITRUM_ONE, SupportedChainId.OPTIMISM].includes(chainId)) {
+  const l2InferredBlockedTokens: typeof unsupportedTokens = useMemo(() => {
+    if (!chainId || !L2_CHAIN_IDS.includes(chainId)) {
       return {}
     }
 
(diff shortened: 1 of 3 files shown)


src/hooks/Tokens.ts
```diff
@@ -57,9 +59,53 @@ export function useAllTokens(): { [address: string]: Token } {
   return useTokensFromMap(allTokens, true)
 }
 
+interface BridgeInfo {
```
zzmp, line 62: I feel like this interface should be available through the token list package.

src/hooks/Tokens.ts
```diff
@@ -57,9 +59,53 @@ export function useAllTokens(): { [address: string]: Token } {
...
+}
 export function useUnsupportedTokens(): { [address: string]: Token } {
+  const { chainId } = useActiveWeb3React()
+  const listsByUrl = useAppSelector((state) => state.lists.byUrl)
   const unsupportedTokensMap = useUnsupportedTokenList()
-  return useTokensFromMap(unsupportedTokensMap, false)
+  const unsupportedTokens = useTokensFromMap(unsupportedTokensMap, false)
+
+  // checks the default L2 lists to see if `bridgeInfo` has an L1 address value that is unsupported
+  const L2InferredBlockedTokens: { [address: string]: Token } = useMemo(() => {
```
zzmp, line 76: nit: lowercase, ie l2InferredBlockedTokens

zzmp: I think a better typing (because it is used for expansion) would be `typeof unsupportedTokens`.

src/hooks/Tokens.ts
```diff
@@ -57,9 +59,53 @@ export function useAllTokens(): { [address: string]: Token } {
...
+    if (!chainId || ![SupportedChainId.ARBITRUM_ONE, SupportedChainId.OPTIMISM].includes(chainId)) {
```
zzmp, line 77: Can this be written more generically? Maybe have the `constants/lists` file export an L2_CHAIN_IDS array? I'm afraid it might be missed in the future if more L2 chains are added.

zzmp (changes requested):

src/hooks/Tokens.ts
JFrankfurt: We do that currently, but that array of chainIds includes L2 testnets where the token lists needed here aren't active by default. I'll take a look at options for exporting constants with and without the testnet ids.

src/hooks/Tokens.ts
JFrankfurt: @moodysalem thoughts on this? It crossed my mind, but I'm not sure how prescriptive you intend extensions/BridgeInfo to be.

src/hooks/Tokens.ts
```diff
@@ -57,9 +59,53 @@ export function useAllTokens(): { [address: string]: Token } {
...
+
+    const { current: list } = listsByUrl[chainId === SupportedChainId.OPTIMISM ? OPTIMISM_LIST : ARBITRUM_LIST]
+    if (!list) {
+      return {}
+    }
+
+    const unsupportedSet = new Set(Object.keys(unsupportedTokens))
+
+    return list.tokens.reduce((acc, tokenInfo) => {
+      const bridgeInfo = tokenInfo.extensions?.bridgeInfo as unknown as BridgeInfo
```
justindomingue, line 93: hmmmm not leaving `unknown` here..

src/hooks/Tokens.ts
```diff
@@ -57,9 +59,53 @@ export function useAllTokens(): { [address: string]: Token } {
...
+  const L2InferredBlockedTokens: { [address: string]: Token } = useMemo(() => {
+    if (!chainId || ![SupportedChainId.ARBITRUM_ONE, SupportedChainId.OPTIMISM].includes(chainId)) {
+      return {}
+    }
+
+    if (!listsByUrl) {
+      return {}
+    }
+
+    const { current: list } = listsByUrl[chainId === SupportedChainId.OPTIMISM ? OPTIMISM_LIST : ARBITRUM_LIST]
```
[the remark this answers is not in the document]
JFrankfurt: @zzmp

src/hooks/Tokens.ts
JFrankfurt: yeah, I think `extensions` is mistyped to not allow `Record<string, any>` 😭

src/hooks/Tokens.ts
moodysalem: it's currently only an extension

justindomingue (approved):

zzmp (approved):

willhennessy:
@JFrankfurt friendly bump to test this after clearing the explicitly unsupported L2 tokens to make sure the dynamic logic is working properly. https://github.com/Uniswap/interface/blob/main/src/constants/tokenLists/unsupported.tokenlist.json

JFrankfurt:
@willhennessy confirmed

Issue #2686, opened 2021-10-25
Select a token dropdown UI flickering on iOS 


jforaker:
**Bug Description**
Firstly, awesome work folks!  

Long time user of Uniswap via Coinbase Wallet on iOS here. Suddenly maybe around 2 weeks ago, the "Select a token" dropdown has become nearly unusable. Each list element seems unresponsive to taps and it looks as if the item flickers when you try and tap it - almost as if the current state is being toggled many many times per second.  

Video: https://share.getcloudapp.com/nOu58RN7 - you can see the list items become highlighted when I tap them and then they struggle and flicker for some time. Its hard to say what actually kicks off a successful tap - whether its timing or number of taps. 

**Steps to Reproduce**

iPhone Xr, iOS 15 > running Coinbase Wallet latest (24.15.534)

1. Go to Swap
2. Click on Select a token
3. Observe the dropdown list and try to select a token. 
4. Observe the list item "flashing" between selected state and non selected state when attempting to tap it
5. Only after some time (sometimes over 15 seconds or so) or some number of taps does the item actually get selected.
   ...

**Expected Behavior**
List item should respond to a single tap and load correctly 

**Additional Context**
Bug only occurs only in mobile browser environments (only tested on iOS). I'm wondering if there was some kind of iOS 15 regression that wasn't caught, because the timeframe matches for when I first started seeing the issue and when iOS 15 was released (not verified but a hunch) 

Thanks

mirshko:
I notice the same on desktop Chrome, seems clicking on a token in the token list takes a few clicks to go through as well.

PR #2724, opened 2021-11-01
fix: memoize the list stuff so the tokens are consistently clickable

moodysalem:
before this fix, clicking a token in the list would just sometimes not work because the list re-renders constantly

fixes https://github.com/Uniswap/interface/issues/2686


justindomingue (approved):

justindomingue:
related to #2686 ?

moodysalem:
> related to #2686 ?

yep, didn't see that

PR #2753, opened 2021-11-08
fix: Enable 3085 requests for coinbase wallet

brendanww:
Enable 3085 request menu for Coinbase Wallet now that 3085 requests are supported across CB Wallet extension, walletlink, and in-app dapp browser.


src/state/application/updater.ts
```diff
@@ -119,7 +119,13 @@ export default function Updater(): null {
   }, [dispatch, debouncedState.chainId])
 
   useEffect(() => {
-    if (!account || !library?.provider?.request || !library?.provider?.isMetaMask) {
+    // @ts-ignore
+    const isCbWalletDappBrowser = window?.ethereum?.isCoinbaseWallet
```
moodysalem, line 123: please add this to the type declarations instead of using ts-ignore

src/state/application/updater.ts
```diff
@@ -119,7 +119,13 @@ export default function Updater(): null {
...
+    // @ts-ignore
+    const isWalletlink = !!window?.WalletLinkProvider || !!window?.walletLinkExtension
+    const isCbWallet = isCbWalletDappBrowser || isWalletlink
+    const isMetamaskOrCbWallet = library?.provider?.isMetaMask || isCbWallet
+    if (!account || !library?.provider?.request || !isMetamaskOrCbWallet) {
       return
     }
     switchToNetwork({ library })
```
moodysalem, line 131: i'm not sure we need to do this test anymore, @JFrankfurt this was just for older versions of metamask right?

brendanww: fwiw, I am seeing some dapps just assume that wallets support 3085 requests. might be better to just remove this check so every other wallet doesn't have to add special flags here?

src/react-app-env.d.ts
```diff
@@ -7,7 +7,10 @@ declare module '@metamask/jazzicon' {
 declare module 'fortmatic'
 
 interface Window {
+  WalletLinkProvider?: any
+  walletLinkExtension?: any
   ethereum?: {
+    isCoinbaseWallet?: false
```
moodysalem, line 13: why is this false or undefined? i'd think it would be true or undefined

src/state/application/updater.ts
```diff
@@ -119,7 +119,11 @@ export default function Updater(): null {
   }, [dispatch, debouncedState.chainId])
 
   useEffect(() => {
-    if (!account || !library?.provider?.request || !library?.provider?.isMetaMask) {
+    const isCbWalletDappBrowser = window?.ethereum?.isCoinbaseWallet
+    const isWalletlink = !!window?.WalletLinkProvider || !!window?.walletLinkExtension
```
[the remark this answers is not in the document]
brendanww: Ah I see, yes true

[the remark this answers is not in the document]
brendanww: not pretty, but addressed here for the extension

src/react-app-env.d.ts
```diff
@@ -7,7 +7,10 @@ declare module '@metamask/jazzicon' {
 declare module 'fortmatic'
 
 interface Window {
+  WalletLinkProvider?: any
```
moodysalem, line 10: sorry can you just add some comments? when is WalletLinkProvider on the window, when is walletLinkExtension on the window, when is window.ethereum on the window with isCoinbaseWallet=true

src/react-app-env.d.ts
```diff
@@ -7,7 +7,10 @@ declare module '@metamask/jazzicon' {
...
+  walletLinkExtension?: any
   ethereum?: {
+    isCoinbaseWallet?: boolean
```
moodysalem, line 13: should never be false

src/react-app-env.d.ts
moodysalem: the code below is very hard to understand, especially not knowing why there are 3 different ways for coinbase to be injected to the window

brendanww: that's fair. added docs.

src/state/application/updater.ts
```diff
@@ -119,7 +121,14 @@ export default function Updater(): null {
   }, [dispatch, debouncedState.chainId])
 
   useEffect(() => {
-    if (!account || !library?.provider?.request || !library?.provider?.isMetaMask) {
+    const isCbWalletDappBrowser = window?.ethereum?.isCoinbaseWallet
+    const isWalletlinkOrInjectedConnector =
+      connector instanceof WalletLinkConnector || connector instanceof InjectedConnector
+    const isWalletlink =
+      !!window?.WalletLinkProvider || (!!window?.walletLinkExtension && isWalletlinkOrInjectedConnector)
```
moodysalem, line 128: is `window?.WalletLinkProvider` not redundant with `connector instanceof WalletLinkConnector`?

src/state/application/updater.ts
```diff
@@ -119,7 +121,14 @@ export default function Updater(): null {
...
+    const isCbWallet = isCbWalletDappBrowser || isWalletlink
+    const isMetamaskOrCbWallet = library?.provider?.isMetaMask || isCbWallet
+    if (!account || !library?.provider?.request || !isMetamaskOrCbWallet) {
```
moodysalem, line 131: it looks like this code actually has a bug... if you have a wallet that supports EIP 3085 and then switch to one that doesn't, it looks like `implements3085` will have last been set to true

src/state/application/updater.ts
brendanww: good call. updated to simplify.

src/state/application/updater.ts
JFrankfurt: This check was for older versions of mm, yeah. My gut says it's still necessary (since Brave still ships w/ an old version.)

JFrankfurt (changes requested):
@brendanww I just tested this using walletconnect via rainbow and it's giving me the supported UX when it shouldn't. Can you take a look at that?

image

JFrankfurt:
Actually, I think this is an instance of the preexisting bug moody mentioned.

JFrankfurt (approved):
lgtm!

brendanww:
@JFrankfurt Do you know when we'll see this change on `main` branch deployed to production?

PR #2767, opened 2021-11-10
feat(widgets): new web3 react for widgets

JFrankfurt:


src/lib/cosmos.decorator.tsx
```diff
@@ -0,0 +1,215 @@
...
+import { UnsupportedChainIdError, useWeb3React, Web3ReactProvider } from '@web3-react/core'
+import {
+  InjectedConnector,
+  NoEthereumProviderError,
+  UserRejectedRequestError as UserRejectedRequestErrorInjected,
+} from '@web3-react/injected-connector'
+import { NetworkConnector } from '@web3-react/network-connector'
+import { useEffect, useState } from 'react'
+
+const RPC_URLS: { [chainId: number]: string } = {
```
zzmp, line 11: Use the constants already defined in the repo.
We're going to need to move them to `lib` eventually, but for now I think we should use them as-is.

src/lib/cosmos.decorator.tsx
```diff
@@ -0,0 +1,215 @@
...
+}
+
+export const injected = new InjectedConnector({ supportedChainIds: [1, 3, 4, 5, 42] })
+
+export const network = new NetworkConnector({
+  urls: { 1: RPC_URLS[1], 4: RPC_URLS[4] },
+  defaultChainId: 1,
+})
+
+export function useEagerConnect() {
```
zzmp, line 23: Use the existing logic in the repo.
Instead of copying this logic over to `lib`, I think it'd be better to just expand the interface logic so that it is usable by `lib`, and worry about completely porting in only when widgets are ready to replace the interface functionality.

src/lib/cosmos.decorator.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+function getErrorMessage(error: Error) {
+  if (error instanceof NoEthereumProviderError) {
+    return 'No Ethereum browser extension detected, install MetaMask on desktop or visit from a dApp browser on mobile.'
+  } else if (error instanceof UnsupportedChainIdError) {
+    return "You're connected to an unsupported network."
+  } else if (error instanceof UserRejectedRequestErrorInjected) {
+    return 'Please authorize this website to access your Ethereum account.'
+  } else {
+    console.error(error)
```
zzmp, line 110: Cover error code -32002 as well (request already pending), and 4001 (rejection).

zzmp (commented):
Will it be possible to use existing code with a new web3-react implementation, or does using a new web3-react necessitate a reimplementation? Have you thought about how to facilitate code re-use?

PR #2768, opened 2021-11-10
refactor: Replace multicall implementation with library

jmrossy:
This replaces the interface's implementation of multicall with the [new redux-multicall library](https://github.com/Uniswap/redux-multicall)

The library works the same way but it's interfaces are slightly more generalized to enable usage in other projects. I.e. it doesn't rely on hooks to retrieve chainId / latestBlockNumber, those are now parameters.

Tested by running interface locally and clicking around. Open to more specific testing if anyone has suggestions 🙂


package.json
```diff
@@ -96,6 +96,7 @@
     "luxon": "^1.25.0",
     "microbundle": "^0.13.3",
     "ms.macro": "^2.0.0",
+    "multicall-query": "^0.1.5",
```
jmrossy, line 99: TODO using my personal version just to test but we need to publish @uniswap/redux-multicall and use that instead.

src/state/multicall/instance.tsx
```diff
@@ -0,0 +1,4 @@
+import { createMulticall } from 'multicall-query' // TODO switch out for '@uniswap/redux-multicall' when published
```
justindomingue, line 1: change extension to `ts`?

src/state/multicall/updater.tsx
```diff
@@ -1,269 +1,13 @@
...
-}
-
-export default function Updater(): null {
-  const dispatch = useAppDispatch()
-  const state = useAppSelector((state) => state.multicall)
-  // wait for listeners to settle before triggering updates
-  const debouncedListeners = useDebounce(state.callListeners, 100)
+// Create Updater wrappers that pull needed info from store
+export default function Updater() {
+  const Updater = multicall.Updater
```
justindomingue, line 8: don't think we need to extract the updater here? It's also a big confusing because the component is called `Updater`

justindomingue (approved):

yarn.lock
```diff
@@ -4654,6 +4654,18 @@
     base64-sol "1.0.1"
     hardhat-watcher "^2.1.1"
 
+"@uniswap/v3-periphery@^1.2.1":
```
moodysalem, line 4657: hm can we make sure there's only one of these?

jmrossy: The multicall lib is using a newer version of v3-periphery than the interface is. I can downgrade its version to match or we can upgrade the interface. @moodysalem what do you recommend?

moodysalem: upgrade the interface, don't think there's any breaking changes (at least that affect the interface)

src/state/multicall/updater.tsx
jmrossy: Will fix, thanks!

yarn.lock
jmrossy: I've made multicall + interface consistent on 1.3.0 but the @uniswap/v3-sdk uses an old periphery version (1.1.1). 
I've added a yarn resolution for now. Alternatively we can upgrade the v3-sdk and ship an update.

NoahZinsmeister (approved):

moodysalem:
My biggest concern with this PR atm is, do we have enough testing infrastructure in the multicall repo to release and version it separately from all its clients? If not, the workflow for updating the redux multicall code will involve `yarn link` (cumbersome) or creating new releases (bad)

Is it possible to restructure the multicall library to get more testing on the `Updater` component of the code, where we make most the changes? Perhaps by using a middleware instead of a react component

jmrossy:
Regarding @moodysalem's comment above, we've created integration tests for the redux-multicall lib here:
https://github.com/Uniswap/redux-multicall/pull/5

moodysalem (approved):

PR #2769, opened 2021-11-10
feat: add support for 0.01% tier

justindomingue:
* in routing diagram for routes returned by the routing api
* in add liquidity

(assuming governance proposal passes)

todo: make router change per chain


src/components/FeeSelector/shared.tsx
```diff
@@ -0,0 +1,30 @@
+import { Trans } from '@lingui/macro'
+import { FeeAmount } from '@uniswap/v3-sdk'
+import { SupportedChainId } from 'constants/chains'
+import { ReactNode } from 'react'
+
+export const FEE_AMOUNT_DETAIL: Record<
+  FeeAmount,
+  { label: string; description: ReactNode; supportedChains: SupportedChainId[] | 'all' }
+> = {
+  [FeeAmount.LOWEST]: {
```
ianlapham, line 10: Would it be better to label them after bits?

ianlapham (commented):
just small nit - rest looks good

src/components/FeeSelector/shared.tsx
justindomingue: probably 😓 that would have to be done in the v3-sdk though :/

ianlapham: Ah in that case probably not worth th refactor

src/components/FeeSelector/shared.tsx
```diff
@@ -0,0 +1,30 @@
...
+    label: '0.01',
+    description: <Trans>Best for very stable pairs.</Trans>,
+    supportedChains: [SupportedChainId.MAINNET],
+  },
+  [FeeAmount.LOW]: {
+    label: '0.05',
+    description: <Trans>Best for stable pairs.</Trans>,
+    supportedChains: 'all',
```
moodysalem, line 18: ALL_SUPPORTED_CHAIN_IDS?

yarn.lock
```diff
@@ -4654,18 +4666,28 @@
...
     tiny-warning "^1.0.3"
 
+"@uniswap/v3-staker@1.0.0":
+  version "1.0.0"
+  resolved "https://registry.yarnpkg.com/@uniswap/v3-staker/-/v3-staker-1.0.0.tgz#9a6915ec980852479dfc903f50baf822ff8fa66e"
+  integrity sha512-JV0Qc46Px5alvg6YWd+UIaGH9lDuYG/Js7ngxPit1SPaIP30AlVer1UYB7BRYeUVVxE+byUyIeN5jeQ7LLDjIw==
+  dependencies:
+    "@openzeppelin/contracts" "3.4.1-solc-0.7-2"
+    "@uniswap/v3-core" "1.0.0"
+    "@uniswap/v3-periphery" "^1.0.1"
```
moodysalem, line 4689: lol periphery getting duplicated by this

src/components/FeeSelector/shared.tsx
moodysalem: or just `allowedChains: undefined | SupportedChainId[]`

moodysalem: one thing to consider is tick spacing may differ between the fee amounts on different chains. not sure where tick spacing is coming from here, is it being queried from the factory?

justindomingue: TICK_SPACINGS is updated here: https://github.com/Uniswap/v3-sdk/blob/main/src/constants.ts#L25

moodysalem: that constant already exists in the chains constants file

i'm also fine with `'all'`

justindomingue: oh nice, i prefer `ALL_SUPPORTED_CHAIN_IDS` if it's already defined.

Magkoooh (approved):

Magkoooh (approved):

PR #2770, opened 2021-11-10
fix: support 0.01% fee tier in client side router

justindomingue:
- chore: add support for 0.01% tier
- only show 1bps on mainnet
- rename VERY_LOW to LOWEST
- upgrade to v3-sdk 3.7.0
- consider 0.01% tier in pool


moodysalem (approved):
ideally on a per chain basis 😅

justindomingue:
will make per-chain in the base pr

PR #2777, opened 2021-11-11
fix(L2): update block warning updater to check most recent block timestamp

JFrankfurt:
prior code was assuming that the most recent block was received when the component mounted :|

### DIFF
--- a/src/state/application/updater.ts
+++ b/src/state/application/updater.ts
@@ -1,4 +1,5 @@
 import { CHAIN_INFO } from 'constants/chains'
+import useCurrentBlockTimestamp from 'hooks/useCurrentBlockTimestamp'
 import useDebounce from 'hooks/useDebounce'
 import useIsWindowVisible from 'hooks/useIsWindowVisible'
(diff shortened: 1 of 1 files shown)


src/state/application/updater.ts
```diff
@@ -35,11 +35,27 @@ function useBlockWarningTimer() {
...
+        if (!chainConnectivityWarningActive) {
+          dispatch(setChainConnectivityWarning({ warn: true }))
+        }
+      } else {
+        if (chainConnectivityWarningActive) {
+          dispatch(setChainConnectivityWarning({ warn: false }))
+        }
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
moodysalem, line 57: smell. also this doesn't need state anymore, it's just whether the ms since last block is greater than wait ms before warning

JFrankfurt: 🤦‍♂️

Viperboi283 (approved):

PR #2781, opened 2021-11-12
fix(L2): remove redux from chain connectivity

JFrankfurt:


src/hooks/useMachineTime.ts
```diff
@@ -0,0 +1,16 @@
+import { useEffect, useState } from 'react'
+
+const useMachineTimeMs = (updateInterval: number): number => {
+  const [now, setNow] = useState(Date.now())
+  useEffect(() => {
```
moodysalem, line 5: `useInterval`

PR #2799, opened 2021-11-15
feat(widgets): add new @web3-react cosmos decorator

JFrankfurt:


package.json
```diff
@@ -93,6 +87,13 @@
     "graphql-request": "^3.4.0",
     "inter-ui": "^3.13.1",
     "jest-styled-components": "^7.0.5",
+    "legacy-web3-react/core": "npm:@web3-react/core",
```
zzmp, line 90: These should retain the same name so that the current interface can continue to use them unchanged. The new version should be renamed.

src/lib/cosmos/connectors/network.ts
```diff
@@ -0,0 +1,19 @@
+import { initializeConnector } from '@web3-react/core'
+import { Network } from '@web3-react/network'
+import { SupportedChainId } from 'constants/chains'
+
+export const URLS = {
```
zzmp, line 5: Can we re-use our existing URLs list?

src/lib/cosmos/components/index.tsx
```diff
@@ -0,0 +1,222 @@
+import { BigNumber } from '@ethersproject/bignumber'
```
zzmp, line 1: src/lib/cosmos/components/Connectors.tsx?

src/lib/cosmos/components/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+  hooks: { useChainId, useAccounts, useError },
+}: {
+  connector: Connector
+  hooks: Web3ReactHooks
+}) {
+  const chainId = useChainId()
+  const accounts = useAccounts()
+  const error = useError()
+
+  const connected = Boolean(chainId && accounts)
```
zzmp, line 37: After disconnecting from MetaMask, this is still marked true in the UI, with "Chain Id: 1", "Accounts: None".

src/lib/cosmos/components/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+    )
+  }
+}
+
+const Wrapper = styled.div`
+  display: flex;
+  flex-direction: row;
+`
+const ConnectorWrapper = styled.div`
+  align-items: center;
```
zzmp, line 185: flex-start

src/lib/cosmos/components/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+      </button>
+    )
+  }
+}
+
+const Wrapper = styled.div`
+  display: flex;
+  flex-direction: row;
+`
+const ConnectorWrapper = styled.div`
```
zzmp, line 184: I'd like this to be styled more like the cosmos interface, so it's more obviously part of the configurator and not a part of the fixture being viewed. WDYT?

JFrankfurt: sure, seems reasonable

src/lib/components/Widget.tsx
```diff
@@ -27,9 +29,14 @@ export interface WidgetProps {
...
   locale?: SupportedLocale
+  provider?: EthProvider
 }
 
-export default function Widget({ children, theme, locale = DEFAULT_LOCALE }: WidgetProps) {
+export default function Widget({ children, theme, locale = DEFAULT_LOCALE, provider }: WidgetProps) {
+  const [, setProvider] = useAtom(providerAtom)
+  useEffect(() => {
+    setProvider(provider)
+  }, [setProvider, provider])
```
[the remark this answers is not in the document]
JFrankfurt: Yes, but we typically include even referentially stable hook dependencies:
`React Hook useEffect has a missing dependency: 'setProvider'. Either include it or remove the dependency array.eslintreact-hooks/exhaustive-deps`

For example, we usually include `dispatch` when it is used, even though that's not needed. I'd rather include unneeded dependencies than override w/ `disable-next-line` when I can, since doing so might lead to unintentional omissions.

src/lib/cosmos/connectors/network.ts
JFrankfurt: No, since that relies on getting the infura key from `process.env.REACT_APP_INFURA_KEY` (and CRA values aren't available in cosmos)

src/lib/components/Widget.tsx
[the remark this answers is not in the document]
zzmp: I'm definitely in favor of just included unneeded deps over adding eslint overrides.

package.json
```diff
@@ -182,6 +187,7 @@
     "rebass": "^4.0.7",
     "styled-components": "^5.3.0",
     "wcag-contrast": "^3.0.0",
-    "wicg-inert": "^3.1.1"
+    "wicg-inert": "^3.1.1",
+    "zustand": "^3.6.5"
```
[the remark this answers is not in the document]
JFrankfurt: Can't until the new web3-react moves zustand out of peerDependencies (will pr this today)

src/lib/state/index.ts
```diff
@@ -0,0 +1,4 @@
+import { Provider as EthProvider } from '@ethersproject/abstract-provider'
+import { atom } from 'jotai'
+
+export const providerAtom = atom<EthProvider | undefined>(undefined)
```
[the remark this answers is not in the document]
JFrankfurt: What could we accomplish with this pattern?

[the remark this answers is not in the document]
zzmp: Multiple widgets sharing a single provider, which can be specified in a Context; this would allow us to support both a widgetized App (using Context), and standalone widgets (using passed provider props).

This can probably just be done by defaulting a Context-specified atom's value to that of the Context, too. That actually seems cleaner. I just wanted to mention the design need now to make sure we didn't accidentally preclude it.

src/lib/cosmos/components/Connectors.tsx
```diff
@@ -0,0 +1,230 @@
...
+  hooks: { useChainId, useAccounts, useError },
+}: {
+  connector: Connector
+  hooks: Web3ReactHooks
+}) {
+  const chainId = useChainId()
+  const accounts = useAccounts()
+  const error = useError()
+
+  const connected = Boolean(chainId && accounts && accounts.length > 0)
```
zzmp, line 38: MetaMask continues to show as active after disconnection.

I think there is a bug in web3-react where the chainId persists despite disconnection. I also think there should be a hook that abstracts this "connected" logic - is that useIsActive()?

src/lib/cosmos/components/Connectors.tsx
```diff
@@ -0,0 +1,230 @@
...
+
+  return (
+    <div>
+      <b>{getName(connector)}</b>
+      <br />
+      {error ? (
+        <>
+          🛑 {error.name ?? 'Error'}: {error.message}
```
zzmp, line 46: nit: Use react-feather instead of emojis for a cleaner UI.

src/lib/cosmos/components/Connectors.tsx
zzmp: Network does not require accounts to be connected.

src/lib/cosmos/components/Connectors.tsx
```diff
@@ -0,0 +1,230 @@
...
+
+function useBalances(
+  provider?: ReturnType<Web3ReactHooks['useProvider']>,
+  accounts?: string[]
+): BigNumber[] | undefined {
+  const [balances, setBalances] = useState<BigNumber[] | undefined>()
+
+  useEffect(() => {
+    if (provider && accounts?.length) {
+      let stale = false
```
zzmp, line 71: Should this be outside of useEffect? I think within useEffect, L74 will always be true.

src/lib/cosmos/components/Connectors.tsx
```diff
@@ -0,0 +1,230 @@
...
+    <div>
+      Accounts:
+      {accounts === undefined
+        ? ' -'
+        : accounts.length === 0
+        ? ' None'
+        : accounts?.map((account, i) => (
+            <ul key={account} style={{ margin: 0, overflow: 'hidden', textOverflow: 'ellipsis' }}>
+              <b>{ENSNames?.[i] ?? account}</b>
+              {balances?.[i] ? ` (Ξ${formatEther(balances[i])})` : null}
```
zzmp, line 115: Display this on a separate line

src/lib/cosmos/components/Connectors.tsx
```diff
@@ -0,0 +1,230 @@
...
+  const provider = useProvider(useAnyNetwork ? 'any' : undefined)
+  const accounts = useAccounts()
+  const ENSNames = useENSNames(provider)
+
+  const balances = useBalances(provider, accounts)
+
+  return (
+    <div>
+      Accounts:
+      {accounts === undefined
```
zzmp, line 108: I think we should display '-' for both undefined and length === 0.

src/lib/cosmos/components/Connectors.tsx
```diff
@@ -0,0 +1,230 @@
...
+  margin-bottom: 14px;
+`
+const ConnectorWrapper = styled.div`
+  align-items: flex-start;
+  border: 1px solid rgba(255, 255, 255, 0.1);
+  border-radius: 4px;
+  color: rgb(248, 248, 248);
+  display: flex;
+  flex-direction: column;
+  font-family: -apple-system, BlinkMacSystemFont, Ubuntu, 'Helvetica Neue', Helvetica, sans-serif;
```
zzmp, line 197: font-size should be 14 for consistency

src/lib/cosmos/components/Connectors.tsx
```diff
@@ -0,0 +1,230 @@
...
+}) {
+  const chainId = useChainId()
+  const accounts = useAccounts()
+  const error = useError()
+
+  const connected = Boolean(chainId && accounts && accounts.length > 0)
+
+  return (
+    <div>
+      <b>{getName(connector)}</b>
```
zzmp, line 42: bold/em should match the styling of cosmos: labels should look plain, and values should be in #808080 color. This can be pushed to a future PR.

zzmp (commented):
I think this needs more styling to make the styling more consistent with cosmos (eg vertical stacking, label/value colors, expando button), but it's low prioritiy / can be done in a future PR.

src/lib/cosmos/components/Connectors.tsx
JFrankfurt: I am an emoji-in-js maximalist

PR #2806, opened 2021-11-16
feat: display an ENS avatar

zzmp:
- Replaces @davatar/react so that interface arch can be leveraged
- [x] Renders https uris
- [x] Renders ipfs uris
- [x] Renders data uris
- [x] Renders ERC721 NFTs
- [x] Renders ERC1155 NFTs

### DIFF
--- a/src/hooks/useENSAvatar.ts
+++ b/src/hooks/useENSAvatar.ts
@@ -1,46 +1,50 @@
 import { namehash } from '@ethersproject/hash'
-import { useMemo } from 'react'
+import { useEffect, useMemo, useState } from 'react'
 import uriToHttp from 'utils/uriToHttp'
 
 import { useSingleCallResult } from '../state/multicall/hooks'
 import { isAddress } from '../utils'
 import isZero from '../utils/isZero'
-import { useENSRegistrarContract, useENSResolverContract } from './useContract'
+import { useENSRegistrarContract, useENSResolverContract, useERC721Contract } from './useContract'
 import useDebounce from './useDebounce'
 import useENSName from './useENSName'
+import { useActiveWeb3React } from './web3'
 
 /**
- * Does an ENS avatar lookup.
+ * Returns the ENS avatar URI, if available.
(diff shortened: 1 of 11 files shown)


src/hooks/useENSAvatar.ts
```diff
@@ -0,0 +1,57 @@
...
+
+function useENSNodeAvatar(node = ''): { avatar: string | null; loading: boolean } {
+  const registrarContract = useENSRegistrarContract(false)
+  const resolverAddress = useSingleCallResult(registrarContract, 'resolver', [node])
+  const resolverAddressResult = resolverAddress.result?.[0]
+  const resolverContract = useENSResolverContract(
+    resolverAddressResult && !isZero(resolverAddressResult) ? resolverAddressResult : undefined,
+    false
+  )
+  const avatar = useSingleCallResult(resolverContract, 'text', [node, 'avatar'])
```
zzmp, line 51: @moodysalem is `useSingleCallResult` the correct API to use here? Is there a way to batch these requests, or will that be done already?

src/hooks/useENSAvatar.ts
```diff
@@ -0,0 +1,57 @@
...
+  const addrAvatar = useENSNodeAvatar(ensNodeArgument[0])
+
+  const name = useENSName(address)
+  const nameAvatar = useENSNodeAvatar(namehash(name.ENSName ?? ''))
+
+  const changed = debouncedAddress !== address
+  const avatar = addrAvatar.avatar || nameAvatar.avatar
+
+  const uri = avatar ? uriToHttp(avatar)[0] : null
+  // TODO: Support ERC721/ERC1155.
```
zzmp, line 35: NB: This is not yet supported by @davatar/react, so there is no loss in functionality.

src/hooks/useENSAvatar.ts
moodysalem: it is already batched, but make sure you memoize array and object arguments because otherwise they are recreated every time

moodysalem (changes requested):
memoize arguments passed to the multicall hooks

PR #2815, opened 2021-11-17
feat: set the auto slippage tolerance by the dollar value of gas

moodysalem:
only works on mainnet because of the available onchain gas price oracle


markemcintosh:
That was fast.

src/hooks/useSwapSlippageTolerance.ts
```diff
@@ -1,25 +1,61 @@
...
 
 const V2_SWAP_DEFAULT_SLIPPAGE = new Percent(50, 10_000) // .50%
 const V3_SWAP_DEFAULT_SLIPPAGE = new Percent(50, 10_000) // .50%
 const ONE_TENTHS_PERCENT = new Percent(10, 10_000) // .10%
 
+/**
+ * Return a guess of the gas cost used in computing slippage tolerance for a given trade
+ * @param trade the trade for which to _guess_ the amount of gas it would cost to execute
+ */
+function guestimateGas(
```
zzmp, line 22: nit: s/guestimate/guesstimate

src/hooks/useSwapSlippageTolerance.ts
```diff
@@ -1,25 +1,61 @@
...
+  trade: V2Trade<Currency, Currency, TradeType> | V3Trade<Currency, Currency, TradeType> | undefined
+): number | undefined {
+  if (!trade) return undefined
```
zzmp, line 25: I think this would be more clear as

    if (trade instanceof V2Trade) {
      // ...
    } else if (trade instanceof V3Trade) {
      // ...
    }
    return undefined

src/hooks/useSwapSlippageTolerance.ts
```diff
@@ -1,25 +1,61 @@
...
+  if (trade instanceof V2Trade) return 90_000 + trade.route.pairs.length * 20_000
```
zzmp, line 26: Include the rationale (or a link to it) of these constants.

src/hooks/useSwapSlippageTolerance.ts
```diff
@@ -1,25 +1,61 @@
...
   const defaultSlippageTolerance = useMemo(() => {
     if (!trade || onL2) return ONE_TENTHS_PERCENT
+
+    const ethGasCost =
+      ethGasPrice && typeof gasEstimate === 'number' ? JSBI.multiply(ethGasPrice, JSBI.BigInt(gasEstimate)) : undefined
+    const dollarGasCost =
+      ether && ethGasCost && etherPrice ? etherPrice.quote(CurrencyAmount.fromRawAmount(ether, ethGasCost)) : undefined
+
+    if (outputDollarValue && dollarGasCost) {
+      const fraction = dollarGasCost.asFraction.divide(outputDollarValue.asFraction)
```
zzmp, line 55: Comment this logic. What are you trying to set? Slippage can be up to or equal to gas cost?

src/hooks/useSwapSlippageTolerance.ts
zzmp: This can just be named `estimateGas`, as it is assigned to `gasEstimate`.

moodysalem: maybe worth renaming when it's a better estimate imo

src/hooks/useSwapSlippageTolerance.ts
moodysalem: this is pretty much a yolo, see: name of method

zzmp (approved):

PR #2821, opened 2021-11-18
fix: display Uniswap token list in UI

### LINKED ISSUE #2756: Display Uniswap token list in UI
the Uniswap Labs default token list should be a visible option in the token list menu

* logo: brand
* name: Uniswap Labs Token List
* state: on by default
* Gemini token list should continue to be on by default as well

bkrochta:
Fixes #2756. 
Removes Uniswap default-token-list build dependency in favor of treating Uniswap token-list like every other list.


src/constants/lists.ts
```diff
@@ -1,3 +1,4 @@
+const UNI_LIST = 'https://gateway.ipfs.io/ipns/tokens.uniswap.org'
```
moodysalem, line 1: we should use the ENS name `tokens.uniswap.eth`, which points to `ipns://tokens.uniswap.org`, which points at the latest version of the Uniswap token list.

src/state/lists/hooks.ts
```diff
@@ -116,7 +113,7 @@ export function useInactiveListUrls(): string[] {
 export function useCombinedActiveList(): TokenAddressMap {
   const activeListUrls = useActiveListUrls()
   const activeTokens = useCombinedTokenMapFromUrls(activeListUrls)
-  return useMemo(() => combineMaps(activeTokens, TRANSFORMED_DEFAULT_TOKEN_LIST), [activeTokens])
+  return activeTokens
```
moodysalem, line 116: this means that users can disable all lists and see no tokens now @willhennessy

moodysalem (changes requested):

moodysalem:
Thank you for the PR @bkrochta, can you also merge main into this branch? To get the linter checks to run

src/constants/lists.ts
bkrochta: Fixed

willhennessy:
Thank bkrochta! Two comments
1. The 'view list' URL is not working. https://tokenlists.org/token-list?url=tokens.uniswap.eth is failing to find a result. Should we try https://tokenlists.org/token-list?url=https://gateway.ipfs.io/ipns/tokens.uniswap.org instead?
2. We should rename the list to "Uniswap Labs Default List". We'll need to make this change in the uniswap token list repo and deploy a new version of the list

bkrochta:
It seems `tokens.uniswap.eth` is not getting resolved correctly by some services. 

`uniswap.tokens.eth` points to `ipns://tokens.uniswap.org`, but this is a DNSLink name which is being deprecated by ens (see [here](https://github.com/ensdomains/ens-app/issues/849#issuecomment-776819864) and [here](https://github.com/ensdomains/content-hash/pull/5)). 

However, the Uniswap interface can still resolve `tokens.uniswap.eth` because it uses its own contenthash decoder [here](https://github.com/Uniswap/interface/blob/a5b152da065a992911d8d8ed7bacb9fd61d1120b/src/utils/contenthashToUri.ts#L21). Other sites such as tokenlists.org use other services to resolve which may not be able to decode a DNSLink contenthash. 

Other tokenlists that use an ens name such as `tokenlist.aave.eth` resolve to `ipfs://{cid}`, so they have no issue being resolved by the Uniswap interface or tokenslist.org or any other service.

This can be fixed by changing `tokens.uniswap.eth` back to `https://gateway.ipfs.io/ipns/tokens.uniswap.org` or changing the contenthash of `tokens.uniswap.eth`

willhennessy:
Thanks for the thorough answer! @moodysalem do you have an opinion on the two options presented in the last sentence? I'm in favor of any minimal effort change to make the 'View list' link in the app work. 

Also, we've updated the name of the token list to `Uniswap Labs List` so that is ready to go.

moodysalem:
does `ipns://tokens.uniswap.org` work?

moodysalem:
> The list will load for the Uniswap interface, but the 'view list' URL https://tokenlists.org/token-list?url=ipns://tokens.uniswap.org will fail since the tokenlist website only expects URLs that start with "https://" or end with ".eth"

Unfortunate, let's just use https://tokens.uniswap.org then.

moodysalem (approved):

PR #2824, opened 2021-11-19
fix(L2): better eip 3085 support

JFrankfurt:
check for support on all connectors, and disable when the connector (or lack thereof) no longer supports 3085

### DIFF
--- a/src/state/application/updater.ts
+++ b/src/state/application/updater.ts
@@ -1,5 +1,3 @@
-import { InjectedConnector } from '@web3-react/injected-connector'
-import { WalletLinkConnector } from '@web3-react/walletlink-connector'
 import useDebounce from 'hooks/useDebounce'
 import useIsWindowVisible from 'hooks/useIsWindowVisible'
 import { useActiveWeb3React } from 'hooks/web3'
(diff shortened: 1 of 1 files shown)


src/state/application/updater.ts
```diff
@@ -79,19 +77,20 @@ export default function Updater(): null {
...
+
   useEffect(() => {
-    const isCbWalletDappBrowser = window?.ethereum?.isCoinbaseWallet
-    const isWalletlink =
-      connector instanceof WalletLinkConnector || (connector instanceof InjectedConnector && window.walletLinkExtension)
-    const isCbWallet = isCbWalletDappBrowser || isWalletlink
-    const isMetamaskOrCbWallet = library?.provider?.isMetaMask || isCbWallet
-    if (!account || !library?.provider?.request || !isMetamaskOrCbWallet) {
-      return
+    if (account && library?.provider?.request) {
```
JFrankfurt, line 83: doesn't make as much sense to check for specific connectors now that we know more than metamask supports the eip

justindomingue (approved):
LGTM, though

PR #2832, opened 2021-11-20
fix: Add routes for stakewise tokens

tsudmi:
# Intoduction

[StakeWise](https://stakewise.io) is a liquid Ethereum 2.0 staking protocol that tokenizes users' deposits as `sETH2` (principal) tokens and staking rewards as `rETH2` (reward) tokens. Both tokens are mapped 1:1 to ETH. StakeWise has the following main pools on Uniswap:

- [ETH - sETH2 pool](https://info.uniswap.org/#/pools/0x7379e81228514a1d2a6cf7559203998e20598346) has 99.36m $ of liquidity and allows users to withdraw staked ETH. The withdrawals from the ETH2 chain will be enabled in the future phases.
- [rETH2 - sETH2 pool](https://info.uniswap.org/#/pools/0xa9ffb27d36901f87f1d0f20773f7072e38c5bfba) has 3.75m $ of liquidity and allows users to compound staking rewards by swapping the reward token (rETH2) into the principal token (sETH2).

# Problem

Unfortunately, our users can't swap the staking rewards into ETH or any other tokens through the Uniswap interface as the path `rETH2 -> sETH2 -> ETH` is not detected.

Also, we are about to create the `SWISE - sETH2` pool for our governance token that is expected to become our main pool and attract a good amount of liquidity. What worries us the most is the fact that our users won't be able to buy/sell SWISE as the path `SWISE -> sETH2 -> ETH` won't be detected.

We would very appreciate the team to have a look into those small additions as they're crucial for the StakeWise project to keep all our three main pools on Uniswap.


brianchilders:
As a StakeWise user, I would love to see these routes defined.  Thank you Uniswap for your consideration.

Teoyovani:
Envíe su clave privada

justindomingue (approved):
LGTM

Note that the the interface uses quotes from https://github.com/Uniswap/smart-order-router which uses pool TVL to determine routes. That router doesn't find the route even with your PR. @willpote is looking into this

The interface then falls back to the client-side v3 router which you modified in this PR.

justindomingue:
@tsudmi @willpote updated the routing api. Change should be live at app.uniswap.org

image


> The API was finding the rETH/stETH pool, but wasn’t finding the stETH/WETH pool. Changed the pool filter logic to better handle this case

tsudmi:
Thanks for the update @justindomingue @willpote . Was it only updated for the `rETH2` token or also for `SWISE` (`0x48c3399719b582dd63eb5aadf12a40b4c3f52fa2`) as we will create the similar pool between `sETH2 - SWISE` in the upcoming days?

justindomingue:
> Thanks for the update @justindomingue @willpote . Was it only updated for the `rETH2` token or also for `SWISE` (`0x48c3399719b582dd63eb5aadf12a40b4c3f52fa2`) as we will create the similar pool between `sETH2 - SWISE` in the upcoming days?

SWISE will also be covered as long as it has liquidity. The logic routes via the rETH2 pools that have the most liquidity

JFrankfurt:
@tsudmi I think you'll want to include your pairs in both directions (see FEI:TRIBE and TRIBE:FEI)
```
export const ADDITIONAL_BASES: { [chainId: number]: { [tokenAddress: string]: Token[] } } = {
  [SupportedChainId.MAINNET]: {
...
    [FEI.address]: [TRIBE],
    [TRIBE.address]: [FEI],
...
  },
}
```

tsudmi:
@JFrankfurt only `sETH2` will be used as a basis for us, so no need to add both directions.

JFrankfurt:
@tsudmi could you reference the addresses via sdk token entities (import from `src/constants/tokens`, like most of the entries here).

tsudmi:
@JFrankfurt replaced the addresses with the SDK tokens. Looks like there is an issue with Vercel billing according to the error message above.

tsudmi:
@JFrankfurt @justindomingue I've fixed the comment, but Vercel fails to deploy it for some reason. Any ideas?

moodysalem:
Think it's ok to merge this for the client side routing which will soon be obsoleted by the smart-order-router package

PR #2837, opened 2021-11-22
fix: enforce contract memoization

### LINKED ISSUE #2822: Brand memoized args from React.useMemo
Certain functions, like multicall, will only be performant when memoized. This can currently only be caught in code reviews (eg #2806).

`useMemo` should be retyped to return a branded type, and `useSingleCallResult` should enforce (via typing) that passed arguments are memoized.

**Describe alternatives you've considered**
N/A

**Additional context**
Example tutorial of branding (like branding a cow, not like a company's brand): https://medium.com/@KevinBGreene/surviving-the-typescript-ecosystem-branding-and-type-tagging-6cf6e516523d

zzmp:
Enforce contract memoization by branding the return type of `useMemo`, and ensuring that that brand is present in the argument to `useSingleCallResult`. This is enforce at compile-time only - there should be zero runtime cost, and not changes to the compiled code.

This strategy could likely be used with other hooks to enforce performance optimizations through typings.

Fixes #2822.


cmcewen (commented):
copying from slack. basically i think we are fixing something in the consumer that should be fixed in the library

```
so i think the hook that is doing the expensive computation should be the useMemo-er. so in this case https://github.com/Uniswap/redux-multicall/blob/fb9890729b4251a9d6118d298afa8cd7da6c6051/src/hooks.ts#L26 useCallsData ensures that it only adds new listeners when any of the relevant variables change https://github.com/Uniswap/redux-multicall/blob/fb9890729b4251a9d6118d298afa8cd7da6c6051/src/hooks.ts#L69.

everyone else up the chain should useMemo on doing expensive work/transforming data (https://github.com/Uniswap/redux-multicall/blob/fb9890729b4251a9d6118d298afa8cd7da6c6051/src/hooks.ts#L155), or useCallback for function/object comparison (https://github.com/Uniswap/redux-multicall/blob/fb9890729b4251a9d6118d298afa8cd7da6c6051/src/hooks.ts#L188) so that we don't get variable thrash down the chain.

basically, you'll mess up the useCallsData optimization if you feed in different values that aren't truly different```

jmrossy (commented):
I agree with @cmcewen, feels like these useMemo calls can happen in redux-multicall if we need them.
Note there's a PR open to move the interface to the new lib: https://github.com/Uniswap/interface/pull/2768

vm:
Seems close-able for now?

PR #2846, opened 2021-11-26
feat: integrate with ledger dappsBrowser

flocks:
This makes uniswap interface compatible with ledger DappsBrowser by injecting `@ledger/iframe-provider` into `window.ethereum` when the  interface is run inside an iframe, **and** the url contains `?embed=ledgerdappbrowser`

Please note this PR is done in the context of [Ledger Vault](https://enterprise.ledger.com/products/vault/), the B2B solution and not directly ledger-live. The vault code reuse the [DappsBrowser](https://github.com/LedgerHQ/eth-dapp-browser) but in order to be visible in Ledger-Live, other steps are needed, described in the [doc](https://developers.ledger.com/docs/dapp/introduction/)

I'm fairly new to ethereum world, so feel free to give me any feedbacks on this.

I have tested this by running the code locally, running an iframe with the url `https://localhost:3000?embed=ledgerdappbrowser`.

Also let met know, if you feel like the code injecting the iframe-provider should be done elsewhere in the code base.

### DIFF
--- a/src/connectors/index.ts
+++ b/src/connectors/index.ts
@@ -7,6 +7,8 @@ import { WalletLinkConnector } from '@web3-react/walletlink-connector'
 
 import UNISWAP_LOGO_URL from '../assets/svg/logo.svg'
 import { ALL_SUPPORTED_CHAIN_IDS, SupportedChainId } from '../constants/chains'
+import { IS_IN_IFRAME } from '../constants/misc'
+import { parsedQueryString } from '../hooks/useParsedQueryString'
 import getLibrary from '../utils/getLibrary'
 import { FortmaticConnector } from './Fortmatic'
 import { NetworkConnector } from './NetworkConnector'
@@ -17,6 +19,16 @@ const PORTIS_ID = process.env.REACT_APP_PORTIS_ID
 
 if (typeof INFURA_KEY === 'undefined') {
   throw new Error(`REACT_APP_INFURA_KEY must be a defined environment variable`)
+}
+
+const isInsideLedgerDappBrowser = () => {
+  const queryParams = parsedQueryString(window.location.search)
+  return IS_IN_IFRAME && queryParams.embed === 'ledgerdappbrowser'
+}
+if (isInsideLedgerDappBrowser()) {
+  import('@ledgerhq/iframe-provider').then(({ IFrameEthereumProvider }) => {
+    window.ethereum = new IFrameEthereumProvider()
+  })
 }
 
 const NETWORK_URLS: { [key in SupportedChainId]: string } = {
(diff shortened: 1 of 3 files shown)


src/connectors/index.ts
```diff
@@ -19,6 +22,14 @@ if (typeof INFURA_KEY === 'undefined') {
   throw new Error(`REACT_APP_INFURA_KEY must be a defined environment variable`)
 }
 
+const isInsideLedgerDappBrowser = () => {
+  const queryParams = parsedQueryString(window.location.search)
+  return IS_IN_IFRAME && queryParams.embed === 'ledgerdappbrowser'
+}
+if (isInsideLedgerDappBrowser()) {
+  window.ethereum = new IFrameEthereumProvider()
```
justindomingue, line 30: would probably lazy load `IFrameEthereumProvider` here. Something like

```
if (isInsideLedgerDappBrowser()) {
  import('@ledgerhq/iframe-provider').then(({ IFrameEthereumProvider }) => {
    window.ethereum = new IFrameEthereumProvider()
  }
}
```

flocks: Good point, updated the PR.

flocks:
> Is there any way we can standardize between this and the gnosis safe provider, which serves the same purpose? What functionality does the gnosis safe require that this provider doesn't have?

Hello @moodysalem thank you for your comment.
To be honest, I'm not familiar at all with `gnosis safe`, and even less with their code base. Did you have an idea in mind on how to standardize between ledger and gnosis? 

I'm not either a maintainer of ledger dapp browser. I just know that a dapp, in order to be integrated in it, needs to add the specific ledger iframe-provider to its list of provider. Ledger's `iframe-provider` itself is a fork of https://github.com/ethvault/iframe-provider with the following [change](https://github.com/LedgerHQ/iframe-provider/commit/3bb05b8803162f1edf4df126755987489461be97) on top of it.

Is gnosis using the original iframe-provider package? 

This PR is rather small, with minimal code changes, and it works, but I understand the will to standardize. However, I feel like I don't know which direction to go to.

moodysalem:
> Is gnosis using the original iframe-provider package?

The gnosis safe web3 provider works similarly to the ethvault iframe provider, except it doesn't use the same JSON RPC interface as an ETH node for the cross-frame communication

I think the best path forward would be for the ledger dapp browser to implement support for the gnosis safe provider if possible. The result would be that the ledger dapp browser immediately supports all gnosis safe apps.

Context:
https://github.com/gnosis/safe-apps-sdk/issues/191
https://github.com/gnosis/safe-apps-sdk/tree/master/packages/safe-apps-web3-react

Issue #2572, opened 2021-10-13
do not link to the create issue UI if the URL is not the production URL or a known IPFS gateway


moodysalem:
this will prevent issues from being created for forks of the interface

PR #2855, opened 2021-11-29
fix: do not show urls if issue is not occurring on app.uniswap.org

moodysalem:
fixes https://github.com/Uniswap/interface/issues/2572

### DIFF
--- a/src/components/ErrorBoundary/index.tsx
+++ b/src/components/ErrorBoundary/index.tsx
@@ -47,7 +47,7 @@ type ErrorBoundaryState = {
   error: Error | null
 }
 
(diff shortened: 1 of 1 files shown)


src/components/ErrorBoundary/index.tsx
```diff
@@ -47,6 +47,8 @@ type ErrorBoundaryState = {
   error: Error | null
 }
 
+const IS_UNISWAP = window.location.host === 'app.uniswap.org'
```
zzmp, line 50: nit: use `hostname` to match other check in service-worker.ts.

zzmp (approved):

PR #2863, opened 2021-11-29
fix: remove orphaned node

zzmp:


ianlapham (commented):
Will this ```icon``` always be defined?

ianlapham (approved):
looka good now

PR #2886, opened 2021-11-30
fix(lint): clean up the eslint config

moodysalem:
we had eslint config in two different places, which seemed to cause the `react-app` config to be ignored 

also the lint action still fails on prs, so this fixes the action to just run `yarn eslint .` on forks and the setup-node step sets up matchers to read the output of the linter to create annotations


.github/workflows/lint.yml
```diff
@@ -38,10 +38,15 @@ jobs:
...
         with:
           github_token: ${{ secrets.github_token }}
           eslint: true
           eslint_extensions: js,jsx,ts,tsx,json
-          auto_fix: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.owner.login == github.repository_owner }}
+          auto_fix: true
+
+      - name: Run eslint
+        if: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.owner.login == github.repository_owner }}
+        run: yarn eslint .
```
moodysalem, line 52: the setup-node action sets up problem matchers that should report all the errors output by this job

moodysalem:
to be aware of for prs from forks https://github.com/actions/setup-node/issues/96

.github/workflows/lint.yml
```diff
@@ -38,10 +38,15 @@ jobs:
...
         uses: wearerequired/lint-action@36c7e6689e80d785d27a22f71d970f3a3b4fcb70
         with:
           github_token: ${{ secrets.github_token }}
           eslint: true
-          eslint_extensions: js,jsx,ts,tsx,json
-          auto_fix: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.owner.login == github.repository_owner }}
+          eslint_args: "-c .eslintrc.json"
+          auto_fix: true
+
+      - name: Run eslint
```
zzmp, line 50: nit: Use the same name for the two lint tasks (eg "Run linters")

src/components/ErrorBoundary/index.tsx
```diff
@@ -126,7 +126,7 @@ function getRelevantState(): null | keyof AppState {
   if (!path.startsWith('#/')) {
     return null
   }
-  const pieces = path.substring(2).split(/[\/\\?]/)
+  const pieces = path.substring(2).split(/[/\\?]/)
```
zzmp, line 129: Is this change intentional?

src/components/LiquidityChartRangeInput/Area.tsx
```diff
@@ -37,7 +37,7 @@ export const Area = ({
             .y0(yScale(0))(
             series.filter((d) => {
               const value = xScale(xValue(d))
-              return value > 0 && value <= innerWidth
+              return value > 0 && value <= window.innerWidth
```
zzmp, line 40: Is this done by the linter? If not, I think it's better to scope this to the window for mocking.

src/hooks/useMonitoringEventCallback.ts
```diff
@@ -42,7 +42,7 @@ function useMonitoringEventCallback() {
       try {
         push(ref(db, 'trm'), {
           chainId,
-          origin: location.origin,
+          origin: window.location.origin,
```
zzmp, line 45: I like that this is now scoped to `window`.

src/state/governance/hooks.ts
```diff
@@ -136,7 +136,7 @@ function useFormattedProposalCreatedLogs(
 
         // Bravo and one bip proposals omit newlines
         if (startBlock === BRAVO_START_BLOCK || startBlock === ONE_BIP_START_BLOCK) {
-          description = description.replace(/  /g, '\n').replace(/\d\. /g, '\n$&')
+          description = description.replace(/ {2}/g, '\n').replace(/\d\. /g, '\n$&')
```
zzmp, line 139: Is this intentional? Is it meant to make it more readable?

zzmp (approved):
nit: `TextPreset` is a weird name, considering it's just themed text. Would it be difficult to change it to `ThemedText` instead?

src/components/ErrorBoundary/index.tsx
moodysalem: yeah, linter caught an unnecessary escape

src/components/LiquidityChartRangeInput/Area.tsx
moodysalem: the rule was something like no access to globals without the prefix, this is equivalent to what it was doing before

src/hooks/useMonitoringEventCallback.ts
moodysalem: yeah, linter asked for it

src/state/governance/hooks.ts
moodysalem: not intetional, linter autofixed it

PR #2911, opened 2021-12-02
fix: undo commits directly to main

justindomingue:


zzmp:
I don't think this is up-to-date, as running `git diff HEAD 82843ff16aa685015ad2ab3468afb22c04770768` still shows differences. I think it omits some more recent commits, and am worried that they'll be clobbered by the revert. Also, being able to check with `git diff` makes the review a lot safer.

Can you pull main and re-revert it?

PR #2920, opened 2021-12-03
feat: optimize trade for approvals and execution

justindomingue:
- uses smart-order-router when users disable auto router
- chooses SwapRouter based on approval statuses and trade routes
- output of use*Route is now a `@router-sdk/Trade`. Before swapping, we generate a v2/trade, v3/trade or router-sdk/trade
    - that's why there a lot of `V2Trade | V3Trade` simplification to a single `Trade`. It makes the ui layer easier to work with, and the logic layer is eventually consistent to a specific protocol `Trade` object


src/hooks/useApproveCallback.ts
```diff
@@ -102,15 +134,76 @@ export function useApproveCallback(
...
+ */
+export function useApprovalOptimizedTrade(
+  trade: Trade<Currency, Currency, TradeType> | undefined,
+  allowedSlippage: Percent
+):
+  | V2Trade<Currency, Currency, TradeType>
+  | V3Trade<Currency, Currency, TradeType>
+  | Trade<Currency, Currency, TradeType>
+  | undefined {
+  const hasV2Routes = trade?.routes.some((route) => route.protocol === Protocol.V2)
```
justindomingue, line 176: @justindomingue should be `every`

src/hooks/useSwapSlippageTolerance.ts
```diff
@@ -11,31 +11,27 @@ import useGasPrice from './useGasPrice'
...
 
 /**
  * Return a guess of the gas cost used in computing slippage tolerance for a given trade
  * @param trade the trade for which to _guess_ the amount of gas it would cost to execute
  */
-function guesstimateGas(
-  trade: V2Trade<Currency, Currency, TradeType> | V3Trade<Currency, Currency, TradeType> | undefined
-): number | undefined {
+function guesstimateGas(trade: Trade<Currency, Currency, TradeType> | undefined): number | undefined {
   if (trade instanceof V2Trade) {
```
justindomingue, line 22: @NoahZinsmeister how would you reconcile this here? We only have a router-sdk/Trade object to wor kwith. Should we stop estimating gas for v2?

src/hooks/useApproveCallback.ts
```diff
@@ -103,23 +135,85 @@ export function useApproveCallback(
...
+
+  const approveCallback = useApproveCallback(
     amountToApprove,
     chainId
       ? trade instanceof V2Trade
         ? V2_ROUTER_ADDRESS[chainId]
         : trade instanceof V3Trade
-        ? v3SwapRouterAddress
-        : undefined
+        ? V3_ROUTER_ADDRESS[chainId]
```
ianlapham, line 157: trade will only be a `V3Trade` if its from legeacy router logic?

src/hooks/useApproveCallback.ts
```diff
@@ -103,23 +135,85 @@ export function useApproveCallback(
...
+
+  const optimizedSwapRouter = useMemo(
+    () => getTxOptimizedSwapRouter({ onlyV2Routes, onlyV3Routes, routeHasSplits, approvalStates }),
+    [approvalStates, routeHasSplits, onlyV2Routes, onlyV3Routes]
+  )
+
+  return useMemo(() => {
+    if (!trade) return undefined
+
+    switch (optimizedSwapRouter) {
```
ianlapham, line 195: Can we make this `optimizedSwapRouter` enum or soemthing?

src/hooks/useAutoRouterSupported.tsx
```diff
@@ -0,0 +1,8 @@
+import { SUPPORTED_CHAINS } from 'state/routing/clientSideSmartOrderRouter/constants'
+
+import { useActiveWeb3React } from './web3'
+
+export default function useAutoRouterSupported(): boolean {
+  const { chainId } = useActiveWeb3React()
+  return Boolean(chainId && SUPPORTED_CHAINS.includes(chainId))
```
ianlapham, line 7: Can we change this to AUTO_ROUTER_SUPPORTED_CHAINS

src/hooks/useSwapSlippageTolerance.ts
ianlapham: Related - teh auto router only returns gas cost estimate for V3 currently right?

src/utils/getTxOptimizedSwapRouter.ts
```diff
@@ -0,0 +1,26 @@
+import { ApprovalState } from 'hooks/useApproveCallback'
+
+/**
+ * Returns the swap router that will result in the least amount of txs (less gas) for a given swap.
+ * Heuristic:
+ * - if trade contains a single v2-only trade & V2 SwapRouter is approved: use V2 SwapRouter
+ * - if trade contains only v3 & V3 SwapRouter is approved: use V3 SwapRouter
+ * - else: approve and use V2+V3 SwapRouter
+ */
+export function getTxOptimizedSwapRouter({
```
ianlapham, line 10: Can we return some enum or something here

NoahZinsmeister (approved):
looking great, left a few small comments

src/hooks/useSwapSlippageTolerance.ts
justindomingue: @NoahZinsmeister so basically always run the existing logic we had for v3

@ianlapham possibly yeah in your PR where you carry the gas estimate around we could use it. We do however still fall back to the legacy v3 router some times so we need a solution for that case.

src/state/routing/utils.ts
```diff
@@ -49,22 +52,37 @@ export function computeRoutes(
...
 
+export function transformRoutesToTrade<TTradeType extends TradeType>(
+  route: ReturnType<typeof computeRoutes>,
+  tradeType: TTradeType
+): Trade<Currency, Currency, TTradeType> {
+  return new Trade({
+    v2Routes:
+      route
+        ?.filter((r) => r.routev2 !== null)
+        // typescript does not perform type narrowing from filter
```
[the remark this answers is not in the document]
justindomingue: neat the type is kind of verbose but i like type safety 🙏 

```
r is typeof route[0] & { routev2: NonNullable<typeof route[0]['routev2']> }
```

src/components/swap/SwapRoute.tsx
```diff
@@ -28,10 +25,16 @@ export default memo(function SwapRoute({
...
-  trade: V2Trade<Currency, Currency, TradeType> | V3Trade<Currency, Currency, TradeType>
+  trade: Trade<Currency, Currency, TradeType>
   syncing: boolean
 }) {
-  const routingAPIEnabled = useRoutingAPIEnabled()
+  const autoRouterSupported = useAutoRouterSupported()
+
+  const routes = getTokenPath(trade)
+
+  // currently, routes cannot be mixed so while `every` is more correct, some is more efficient
```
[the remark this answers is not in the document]
justindomingue: yeah you're right, i'll just remove the comment it doesn't add much

src/hooks/useApproveCallback.ts
justindomingue: hm in this case actually we pass a v2 | v3 | trade which is derived from the routing api `Trade` by looking at approvals. So if a user has already approved the v3 router, we convert the `Trade` into a `V3Trade`

PR #2923, opened 2021-12-03
chore: merge main into widgets

JFrankfurt:


package.json
```diff
@@ -18,7 +18,10 @@
     "@graphql-codegen/typescript-operations": "^1.18.2",
     "@graphql-codegen/typescript-rtk-query": "^1.1.1",
     "@lingui/cli": "^3.9.0",
+    "@lingui/macro": "^3.9.0",
```
zzmp, line 21: lingui and popperjs should stay in deps, not devDeps

package.json
```diff
@@ -57,6 +60,7 @@
     "@uniswap/governance": "^1.0.2",
     "@uniswap/liquidity-staker": "^1.0.2",
     "@uniswap/merkle-distributor": "1.0.1",
+    "@uniswap/redux-multicall": "^1.0.0",
```
zzmp, line 63: This should now go in devDeps, not deps, if it's to be used for widgets (ok to move in a later PR)

zzmp (changes requested):

package.json
JFrankfurt: reversed?

PR #2950, opened 2021-12-09
fix: memoize more swap

zzmp:
NB: This is easiest to review if you go by commit, rather than looking at all files changed at once.


src/state/swap/hooks.tsx
```diff
@@ -185,46 +185,57 @@ export function useDerivedSwapInfo(toggledVersion: Version | undefined): {
...
 
-  if (balanceIn && amountIn && balanceIn.lessThan(amountIn)) {
-    inputError = <Trans>Insufficient {amountIn.currency.symbol} balance</Trans>
-  }
+    if (balanceIn && amountIn && balanceIn.lessThan(amountIn)) {
+      inputError = <Trans>Insufficient {amountIn.currency.symbol} balance</Trans>
+    }
+
+    return inputError
+  }, [account, allowedSlippage, bestTrade, currencies, currencyBalances, parsedAmount, to, v2Trade])
```
justindomingue, line 238: note that this dependency array is very large so the memoization won't be as effective. Probably okay

src/hooks/useCurrentBlockTimestamp.ts
```diff
@@ -1,10 +1,19 @@
...
 
 import { useSingleCallResult } from '../state/multicall/hooks'
 import { useMulticall2Contract } from './useContract'
 
 // gets the current timestamp from the blockchain
 export default function useCurrentBlockTimestamp(): BigNumber | undefined {
+  const [lastBlockTimestamp, setLastBlockTimestamp] = useState<BigNumber | undefined>(undefined)
   const multicall = useMulticall2Contract()
-  return useSingleCallResult(multicall, 'getCurrentBlockTimestamp')?.result?.[0]
+  // TODO:zzmp
```
justindomingue, line 11: what's the todo here? (going by commits, may be missing where you remove this)

src/hooks/useCurrentBlockTimestamp.ts
```diff
@@ -1,10 +1,19 @@
...
+  const currentBlockTimestamp = useSingleCallResult(multicall, 'getCurrentBlockTimestamp')?.result?.[0]
```
justindomingue, line 12: i'm surprised this is causing issues. Shouldn't `useSingleCassResult` return the same object every time?

also i wonder if there's merit to using `number` over `BigNumber` for primitive equality.

src/hooks/useBestV3Trade.ts
```diff
@@ -26,7 +27,10 @@ export function useBestV3Trade(
   const routingAPIEnabled = useRoutingAPIEnabled()
   const isWindowVisible = useIsWindowVisible()
 
-  const [debouncedAmount, debouncedOtherCurrency] = useDebounce([amountSpecified, otherCurrency], 200)
+  const [debouncedAmount, debouncedOtherCurrency] = useDebounce(
+    useMemo(() => [amountSpecified, otherCurrency], [amountSpecified, otherCurrency]),
```
justindomingue, line 31: interesting catch. This was a pretty bad bug i think.. i wonder if splitting up the debounce in 2 would be more efficient

src/hooks/useDebounce.ts
```diff
@@ -1,5 +1,9 @@
 import { useEffect, useState } from 'react'
 
+/**
+ * Debounces updates to a value.
+ * Non-primitivevs *must* wrap the value in useMemo, or the value will be updated due to referential inequality.
```
justindomingue, line 5:
```suggestion
 * Non-primitives *must* wrap the value in useMemo, or the value will be updated due to referential inequality.
```

justindomingue (approved):

src/state/swap/hooks.tsx
```diff
@@ -185,46 +185,57 @@ export function useDerivedSwapInfo(toggledVersion: Version | undefined): {
...
-  }
+  const currencyBalances = useMemo(
+    () => ({
+      [Field.INPUT]: relevantTokenBalances[0],
+      [Field.OUTPUT]: relevantTokenBalances[1],
+    }),
+    [relevantTokenBalances]
+  )
 
-  const currencies: { [field in Field]?: Currency | null } = {
```
ianlapham, line 193: @zzmp Just so i understand here - the issue here is that `currencies` will change its value every render? (if we reference currencies anywhere else) 

But if we only destructure currencies then its fine? (aka no use of currencies itself)

src/hooks/useCurrentBlockTimestamp.ts
zzmp: I was just marking the file with a diff so I could find it again.

src/state/swap/hooks.tsx
zzmp: exactly this, yes

src/hooks/useCurrentBlockTimestamp.ts
zzmp: I don't know this library's internals. I'll ask someone else to review this to make sure multicall is working correctly, but in my local testing this was returning a new `BigNumber` (with the same value) every time.

src/hooks/useBestV3Trade.ts
zzmp: I think this is more efficient, because it doesn't create an extra setTimeout.

justindomingue (approved):

src/hooks/useCurrentBlockTimestamp.ts
zzmp: This appears to be working correctly now. I believe the library was being mis-used.

Issue #2744, opened 2021-11-05
Deadline limited to 180 minutes


wjmelements:
In "Expert Mode" I should be able to specify a longer deadline. For low-volume tokens, it is desirable to save on gas.
<img width="526" alt="Screen Shot 2021-11-04 at 6 02 00 PM" src="https://user-images.githubusercontent.com/799573/140440952-88f5e858-e0dd-4c8a-8960-ab5c1b1e21f7.png">
I suggest 3 days as a better maximum than 3 hours.

bgits:
The same should apply to removing liquidity which is generally less sensitive to price changes than swaps.

bgits:
PR opened: https://github.com/Uniswap/interface/pull/2982

PR #2982, opened 2021-12-15
fix: extend transaction deadline to 3 days

bgits:
As per https://github.com/Uniswap/interface/issues/2744

### DIFF
--- a/src/components/TransactionSettings/index.tsx
+++ b/src/components/TransactionSettings/index.tsx
@@ -3,6 +3,7 @@ import { Percent } from '@uniswap/sdk-core'
 import { L2_CHAIN_IDS } from 'constants/chains'
 import { DEFAULT_DEADLINE_FROM_NOW } from 'constants/misc'
 import { useActiveWeb3React } from 'hooks/web3'
+import ms from 'ms.macro'
 import { darken } from 'polished'
 import { useContext, useState } from 'react'
 import { useSetUserSlippageTolerance, useUserSlippageTolerance, useUserTransactionTTL } from 'state/user/hooks'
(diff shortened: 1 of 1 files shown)


src/components/TransactionSettings/index.tsx
```diff
@@ -142,7 +142,8 @@ export default function TransactionSettings({ placeholderSlippage }: Transaction
     } else {
       try {
         const parsed: number = Math.floor(Number.parseFloat(value) * 60)
-        if (!Number.isInteger(parsed) || parsed < 60 || parsed > 180 * 60) {
+        const threeDaysInSeconds = 60 * 60 * 24 * 3
```
justindomingue, line 145: better to keep const in the global space to avoid recomputing.

also can use [ms](https://www.npmjs.com/package/ms)

```
ms`3 days` * 1000
```

bgits: Global space of `index.tsx` or in a constants file?

justindomingue: you can leave it in `index.tsx` yeah

zzmp: Prefer ```ms`3 days` * 1000``` for readability. Once this is done and moved into the file's global space, it should be good to go. Thank you!

bgits: Updated

src/components/TransactionSettings/index.tsx
```diff
@@ -142,7 +145,7 @@ export default function TransactionSettings({ placeholderSlippage }: Transaction
     } else {
       try {
         const parsed: number = Math.floor(Number.parseFloat(value) * 60)
-        if (!Number.isInteger(parsed) || parsed < 60 || parsed > 180 * 60) {
+        if (!Number.isInteger(parsed) || parsed < 60 || parsed > THREE_DAYS_IN_SECONDS) {
```
wjmelements, line 148: should this be dependent on "expert mode"

bgits: Probably not IMO. The user can already change the deadline outside of expert mode and the default deadline is set to 30 minutes. Pending transactions that are "stale" can also usually be nullified by submitting a new TX with high gas.

zzmp: AFAIK expert mode mostly refers to slippage, not deadline.

src/components/TransactionSettings/index.tsx
zzmp: we both got it wrong with `* 1000`, glad you used `/ 1000` instead :)

zzmp (approved):

PR #3015, opened 2021-12-20
feat(polygon): polygon mainnet and testnet support

moodysalem:


src/components/CurrencyLogo/index.tsx
```diff
@@ -87,7 +88,17 @@ export default function CurrencyLogo({
   }, [currency, uriLocations])
 
   if (currency?.isNative) {
-    return <StyledEthereumLogo src={EthereumLogo} alt="ethereum logo" size={size} style={style} {...rest} />
+    let nativeLogoUrl: string
+    switch (currency.chainId) {
```
justindomingue, line 92: have you considered putting this in `CHAIN_INFO` instead? Would probably need to lazy load the logos but might be cleaner

src/components/DowntimeWarning/index.tsx
```diff
@@ -28,50 +30,54 @@ const ReadMoreLink = styled(ExternalLink)`
...
+function Wrapper({ children }: { children: React.ReactNode }) {
+  return (
+    <Root>
+      <WarningIcon />
+      {children}
+    </Root>
+  )
+}
+
+/**
```
justindomingue, line 42: nit. usually prefer collapsing into a single line when it fits

src/components/DowntimeWarning/index.tsx
```diff
@@ -28,50 +30,54 @@ const ReadMoreLink = styled(ExternalLink)`
   text-decoration: underline;
 `
 
+function Wrapper({ children }: { children: React.ReactNode }) {
```
justindomingue, line 33:
```suggestion
const Wrapper: React.FC = ({ children }) {
```

if you want to avoid typing children manually

src/components/NetworkAlert/NetworkAlert.tsx
```diff
@@ -209,59 +197,47 @@ interface NetworkAlertProps {
   thin?: boolean
 }
 
+const BETA_TAG_COLORS: { [chainId in SupportedL2ChainId]: string } = {
```
justindomingue, line 200: I understand that `CHAIN_INFO` may get bloated but have you considered putting that there to colocate chain info and make the code easier to navigate for newer engs?

src/components/NetworkAlert/NetworkAlert.tsx
```diff
@@ -209,59 +197,47 @@ interface NetworkAlertProps {
...
 export function NetworkAlert(props: NetworkAlertProps) {
   const { account, chainId } = useActiveWeb3React()
   const [darkMode] = useDarkModeManager()
-  const [arbitrumAlphaAcknowledged, setArbitrumAlphaAcknowledged] = useArbitrumAlphaAlert()
-  const [optimismAlphaAcknowledged, setOptimismAlphaAcknowledged] = useOptimismAlphaAlert()
-  const [locallyDismissed, setLocallyDimissed] = useState(false)
-  const userEthBalance = useETHBalances(account ? [account] : [])?.[account ?? '']
+  const [alertAcknowledged, acknowledgeAlert] = useNetworkAlertStatus(chainId)
+  const [locallyDismissed, setLocallyDimissed] = useState(alertAcknowledged)
+  const userNativeCurrencyBalance = useETHBalances(account ? [account] : [])?.[account ?? '']
```
justindomingue, line 212: will `useETHBalances` memoize `[account]`?

src/components/CurrencyLogo/index.tsx
moodysalem: not great to import a file from the constants directory, e.g. for your refactoring. i think the logo is a rendering concern

moodysalem: just realized we're already doing it for the network logo

Issue #1945, opened 2021-07-01
manage token list modal scrolls with many lists


moodysalem:
the manage token list modal scrolls with a large number of lists
![image](https://user-images.githubusercontent.com/7897876/124144838-eee31e00-da51-11eb-9958-71d862dc2f63.png)

PR #3020, opened 2021-12-21
fix: double scroll in manage token list

tinaszheng:
fixes https://github.com/Uniswap/interface/issues/1945

### DIFF
--- a/src/components/SearchModal/Manage.tsx
+++ b/src/components/SearchModal/Manage.tsx
@@ -16,8 +16,8 @@ import { PaddedColumn, Separator } from './styleds'
 const Wrapper = styled.div`
   width: 100%;
   position: relative;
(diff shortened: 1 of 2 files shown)


justindomingue:
@tinaszheng  added `fixes` before the work item in the description to auto-close it.

justindomingue (approved):

moodysalem:
It looks like the scrollbar is being clipped, I wonder if there's a more 'correct' fix
image

tinaszheng:
@moodysalem i completely missed that 80px of padding at the bottom 🤦‍♀️  that shouldnt be necessary if we're already defining the height in the parent container so i fixed it to let the scroll-list expand to the rest of the height

moodysalem (approved):

Issue #2477, opened 2021-10-01
Unsupported network causes app to crash.


JFrankfurt:
**Bug Description**
App crashed when connected to polygon.

TypeError: Cannot read property 'infoLink' of undefined
    at vf (https://app.uniswap.org/static/js/main.f97146cb.chunk.js:1:338179)
    at ia (https://app.uniswap.org/static/js/35.ec3e0318.chunk.js:2:1189283)
    at za (https://app.uniswap.org/static/js/35.ec3e0318.chunk.js:2:1198850)
    at Wu (https://app.uniswap.org/static/js/35.ec3e0318.chunk.js:2:1242732)
    at Ss (https://app.uniswap.org/static/js/35.ec3e0318.chunk.js:2:1228915)
    at Es (https://app.uniswap.org/static/js/35.ec3e0318.chunk.js:2:1228843)
    at xs (https://app.uniswap.org/static/js/35.ec3e0318.chunk.js:2:1228706)
    at ys (https://app.uniswap.org/static/js/35.ec3e0318.chunk.js:2:1225672)
    at https://app.uniswap.org/static/js/35.ec3e0318.chunk.js:2:1175051
    at https://app.uniswap.org/static/js/35.ec3e0318.chunk.js:2:1252470

**Steps to Reproduce**

1. connect to polygon
2. Use app.uniswap.org

justindomingue:
idk what useActiveWeb3React returns when chainid isn't supported by the app. I'm assuming it doesn't care and returns the id and then we try to destructure here https://github.com/Uniswap/interface/search?q=infolink

to test you can add say polygon to metasmask and then switch network.

tinaszheng:
hmm not able to repro the issue locally or on prod - just wanted to confirm if this is still an issue @JFrankfurt? 😮

JFrankfurt:
Yeah I'm p sure this has been resolved.

JFrankfurt:
@tinaszheng I was wrong, this is still an issue. (see https://github.com/Uniswap/interface/issues/2749)

Just got a report of this happening on polygon via discord. Some of the issues referenced at #2749 include trade info in the url, so I'm guessing this crash is triggered by some trade processing logic.

PR #3024, opened 2021-12-21
fix: app crash on coinbase wallet + polygon network

tinaszheng:
fixes https://github.com/Uniswap/interface/issues/2477

walletlink/coinbase wallet used to only support mainnet but added other networks along the way
so we need to let `react-web3` know to only support certain chains

to reproduce the bug on main:
- disconnect your metamask/existing wallet connection
- connect to coinbase wallet using the QR code scanner
- on the coinbase wallet app, change your default network to any unsupported network (like polygon)

(thank you @NoahZinsmeister for your help!!)


NoahZinsmeister (approved):

justindomingue (approved):

JFrankfurt:
🎉
thanks @tinaszheng

PR #3025, opened 2021-12-21
chore: adding weth to common bases

snreynolds:


src/constants/routing.ts
```diff
@@ -133,9 +135,15 @@ export const COMMON_BASES: ChainCurrencyList = {
     ExtendedEther.onChain(SupportedChainId.POLYGON),
     WRAPPED_NATIVE_CURRENCY[SupportedChainId.POLYGON],
   ],
+  [SupportedChainId.POLYGON]: [
+    ExtendedEther.onChain(SupportedChainId.POLYGON),
```
justindomingue, line 139: nit. also defined at 135, small perf improvement by defining once.. not a big dal

src/constants/routing.ts
```diff
@@ -133,9 +135,15 @@ export const COMMON_BASES: ChainCurrencyList = {
...
+    WRAPPED_NATIVE_CURRENCY[SupportedChainId.POLYGON],
+    WETH_POLYGON,
```
justindomingue, line 141: should we add usdc here as well?

justindomingue: and dai too maybe

justindomingue (approved):

src/constants/routing.ts
willhennessy: yes please, all the usual:  ethereum logo
ETH (n/a on polygon)
DAI
USDC
USDT
WBTC
WETH

PR #3036, opened 2021-12-23
fix(style): improve the bridge assets style

callil:
Re-styles the bridge card to be more minimal and persistent. Also removes downtime warnings as they are no longer necessary.

### DIFF
--- a/src/components/NetworkAlert/NetworkAlert.tsx
+++ b/src/components/NetworkAlert/NetworkAlert.tsx
@@ -105,7 +105,7 @@ const LinkOutToBridge = styled(ExternalLink)`
   padding: 6px 8px;
   margin-right: 12px;
   text-decoration: none;
-  width: auto;
+  width: 100%;
   :hover {
     text-decoration: none;
   }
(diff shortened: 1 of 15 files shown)


src/pages/Swap/index.tsx
```diff
@@ -56,7 +56,7 @@ import { warningSeverity } from '../../utils/prices'
 import AppBody from '../AppBody'
 
 const AlertWrapper = styled.div`
-  max-width: 480px;
+  max-width: 460px;
```
moodysalem, line 59: can we add `width: 100%`

src/pages/Pool/index.tsx
```diff
@@ -168,6 +147,8 @@ export default function Pool() {
   const showConnectAWallet = Boolean(!account)
   const showV2Features = Boolean(chainId && V2_FACTORY_ADDRESSES[chainId])
 
+  console.log(chainId)
```
moodysalem, line 150: delete

src/components/NetworkAlert/NetworkAlert.tsx
```diff
@@ -189,126 +141,36 @@ function shouldShowAlert(chainId: number | undefined): chainId is SupportedChain
 }
 
 export function NetworkAlert() {
-  const { account, chainId } = useActiveWeb3React()
-  const [darkMode] = useDarkModeManager()
-  const [alertAcknowledged, acknowledgeAlert] = useNetworkAlertStatus(chainId)
```
moodysalem, line 194: looks like we no longer need these hooks/state, if you want to remove

Issue #2838, opened 2021-11-23
should check forward resolution matches of ENS name set in the reverse ENS record


moodysalem:
**Bug Description**
if you set a reverse ENS record to a name, such as abc.eth, and do not control abc.eth, the app will show abc.eth as if that is your ENS name

**Expected Behavior**
we should also check the forward resolution matches the current wallet address before showing the ens name or other ens attributes in the header

moodysalem:
@zzmp you may be interested in this context

uniyj:
Hi @moodysalem, @zzmp. I think there might be a few other problems related to ENS, as reported by a number of tickets on Uniswap Discord.

A total of three tickets mentioned their addresses showing up as `0x000...` on Uniswap UI:
- Two of them seem to have multiple ENS names (including 'subdomain' names) resolving to their addresses.
- The third one does not seem to have any ENS reverse name set, but had a tx to ENS Reverse Registrar calling `setName` function with input variable string `0x000...`

I don't know if the above observations were indeed the reason(s) that they get addresses showing up as `0x000...` on the UI, more a guess from my part. I can provide more details (including links to the Discord tickets channels and their addresses) in DMs, just let me know what you need. Thanks.

uniyj:
This is now resolved by PR 3040. Please close. Thanks.

PR #3040, opened 2021-12-25
fix: Add ENSName and address check to Web3Status

uniyj:
Issue #2838

Note that I don't know TypeScript and frontend. In my tests (just printing `console.log`s), what I did here seems to work, but I have no idea if this is acceptable code or not... Someone should please review this. Thanks.

### DIFF
--- a/src/components/Web3Status/index.tsx
+++ b/src/components/Web3Status/index.tsx
@@ -8,7 +8,6 @@ import { Activity } from 'react-feather'
 import styled, { css } from 'styled-components/macro'
 
 import { NetworkContextName } from '../../constants/misc'
-import useENSAddress from '../../hooks/useENSAddress'
 import useENSName from '../../hooks/useENSName'
 import { useHasSocks } from '../../hooks/useSocksBalance'
 import { useWalletModalToggle } from '../../state/application/hooks'
@@ -142,10 +141,7 @@ function WrappedStatusIcon({ connector }: { connector: AbstractConnector }) {
 function Web3StatusInner() {
   const { account, connector, error } = useWeb3React()
 
-  const revName = useENSName(account ?? undefined)
-  // check that reverse-lookup ENS name resolves to account address
-  const fwdAddr = useENSAddress(revName.ENSName)
-  const ENSName = account === fwdAddr.address ? revName.ENSName : null
+  const { ENSName } = useENSName(account ?? undefined)
 
   const allTransactions = useAllTransactions()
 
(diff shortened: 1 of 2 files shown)


uniyj:
@moodysalem, for Christmas I wrote a few lines of JS/TS =P
(Please check haha)

tinaszheng:
Hey @uniyj, thanks for your contribution! We should probably be doing the equality check everywhere we use ENS name. Could you move that `useENSAddress` + equality check logic inside of `useENSName`?

uniyj:
Hi @tinaszheng. That was what I wanted to do initially, but I am really not familiar with TS, React hooks, etc., and so did not want to mess with the hooks – in comparison, the Web3Status file seemed more straightforward to me...

In any case, I've now tried to move the ENSAddress check into ENSName hook, and reverted the previous changes to Web3Status. Again, no idea if it's acceptable code or not.

In my quick tests, it seems like a "stray" reverse-name will still show up momentarily before being replaced by the 0xXXX address again when the ENSAddress check kicks in. I have no idea why/how it does that, because I cannot actually decipher much of what's going on in TS and React hooks flows..!

tinaszheng:
thanks @uniyj! your PR looks fine to me. can you explain more about what you mean by the "stray" reverse-name display? 

Are you seeing something like:
if 0xXX points to a.eth:

1) 0xXX address shows for a few seconds and then
2) a.eth shows? 

If so, that's fine. Are you seeing a different behavior?

uniyj:
I haven't actually tried to set a reverse-ENS for my address to an ENS name that does not forward back to that address, so cannot fully test what happens there (see what's mentioned in comments in issue #2838).

I tried to somewhat replicate that behaviour by flipping the logic check to `const checkedName = address !== fwdAddr?.address ? name : null` and seeing what happens. What I observed in the status box was then:

1. Status shows `0xXXX` address
2. Eth balance loads in status
3. `0xXXX` changes to show the reverse-ENS set for that address (not sure why this happens first)
4. The flipped logic check seems to then trigger (mimicking that the ENS name does not forward to the address)
5. The ENS name changes back to show `0xXXX`

tinaszheng:
That logic doesn't really replicate the behavior of having a different reverse proxy but that's ok. Given the logic that behavior seems expected and fine to me. The order of sequence for the states will be:

1. `address` is defined, `fwdAddr`, `name` are both undefined
2. `address` is defined, `name` is defined, `fwdAddr` is undefined
3. `address`, `name`, and `fwdAddr` are all defined

i added an inline comment just as a note to future developers, but otherwise this looks great to me~ thanks so much!

tinaszheng (approved):

Issue #1893, opened 2021-06-22
Add network parameter to uniswap URLs


rahul-kothari:
**Is your feature request related to a problem? Please describe.**
I am working for the EF's Devcon ticketing team. We use uniswap as a way for users to pay for the tickets. And this devcon we are also integrating some L2 networks.

We basically create a uniswap URL and let user's pay through that. So for example:
https://uniswap.exchange/send?exactField=output&exactAmount={exact_amount}&outputCurrency={output_currency}&recipient={recipient_address}

Except this always defaults to Ethereum mainnet. So say, I wanted to do this on Rinkeby testnet or optimism, because there is no network parameter, the website defaults to ETH Mainnet and reports that the tokens are unrecognised. The user would have to then manually connect their wallet to the right network to see the tokens. This isn't just poor UX for the user but also potentially dangerous. The non-sophisticated user wouldn't even realise what went wrong and may end up paying on the wrong network.

**Describe the solution you'd like**
The URL to have a `network`/`chainId` parameter (default = 1).
So if a chainId is provided, then the UI should default to the appropriate network (instead of currently assuming it is always eth mainnet).

https://uniswap.exchange/send?exactField=output&exactAmount={exact_amount}&outputCurrency={output_currency}&recipient={recipient_address}&chainId=3

This would be super helpful for when uniswap scales to other L2s. 

**Additional context**
![image](https://user-images.githubusercontent.com/35169918/122901925-eef26800-d345-11eb-9640-861c34949895.png)
https://app.uniswap.org/#/swap?exactAmount=3500&outputCurrency=ETH&inputCurrency=0xc7AD46e0b8a400Bb3C915120d284AafbA8fc4735recipient=0xc7AD46e0b8a400Bb3C915120d284AafbA8fc4735&exactField=output

^ Convert DAI to ETH on Rinkeby testnet. However to a user, it would seem like there is nothing wrong here and end up paying on the wrong network.

willhennessy:
Hi Rahul, thanks for the suggestion! @JFrankfurt  are we able to change the network on command like this? I thought metamask imposes some limits

rahul-kothari:
Any update on changing the URL? More useful now that uniswap is on mainnet

willhennessy:
Agreed, but we have a bunch of other tasks for L2 as well so we haven't had
a chance to start yet sorry!

moodysalem:
[...]

Issue #3032, opened 2021-12-23
Chain specific (agnostic) URL


leandercapuozzo:
Would improve user experience if LP positions were viewable at URLs which were agnostic to current chain (network) which my wallet is set to. If I am set Polygon https://app.uniswap.org/#/pool/129 views one pool, while when set to Ethereum it views another. 

A potential work around would be to add a further divide to the URL

/pool **/polygon** /129

willhennessy:
in progress! thanks leander. going to close this since we're tracking it in https://github.com/Uniswap/interface/issues/1893

PR #3057, opened 2021-12-29
feat: add chainId (network) as url parameter

tinaszheng:
fixes https://github.com/Uniswap/interface/issues/1893 https://github.com/Uniswap/interface/issues/3032

This allows users to add `chainId` to the URL to prompt the user to switch their wallet network if necessary.

Some considerations/assumptions here:
- By default (no `chainId` parameter), the network should be the wallet's network
- When no wallet is connected, the behavior should match that of switching the network option with no wallet that is currently in the app
- if `chainId` parameter exists, and is a supported chain id and is not the wallet current network, then fire a request to prompt network change
- We consider the wallet's network as the source of truth -> if the user rejects the network change request, remain on the current network
- When we switch networks on the app, should we update the URL too? ~I'm thinking no, since not changing it would be consistent with the behavior of `inputCurrency` and `outputCurrency` parameters, where the url param acts as one-directional data flow~

~to try it out: append `swap?inputCurrency=0x8e915a77e301ad12ad1f62012734cf7557eee81a&&chainId=137` to the URL~

edit: 
- when we switch networks on the app, we *should* update the URL and vice versa
- we're using `chain` name instead of `chainId`. alternate URL: `swap?inputCurrency=0x8e915a77e301ad12ad1f62012734cf7557eee81a&&chain=polygon`

### DIFF
--- a/src/components/Header/NetworkSelector.tsx
+++ b/src/components/Header/NetworkSelector.tsx
@@ -286,7 +286,7 @@ export default function NetworkSelector() {
     if (chainId && chainId !== prevChainId) {
       history.replace({ search: replaceURLParam(history.location.search, 'chain', getChainNameFromId(chainId)) })
       // otherwise assume network change originates from URL
-    } else if (urlChainId) {
+    } else if (urlChainId && urlChainId !== chainId) {
       handleChainSwitch(urlChainId, true)
     }
   }, [chainId, urlChainId, prevChainId, handleChainSwitch, history])
--- a/src/components/Header/NetworkSelector.tsx
+++ b/src/components/Header/NetworkSelector.tsx
@@ -225,7 +225,7 @@ const getParsedChainId = (parsedQs?: ParsedQs) => {
 export default function NetworkSelector() {
   const { chainId, library } = useActiveWeb3React()
   const parsedQs = useParsedQueryString()
-  const [promptedNetworkChange, setPromptedNetworkChange] = useState(false)
+  const [lastNetworkChange, setLastNetworkChange] = useState<number | null>(null)
   const node = useRef<HTMLDivElement>()
   const open = useModalOpen(ApplicationModal.NETWORK_SELECTOR)
   const toggle = useToggleModal(ApplicationModal.NETWORK_SELECTOR)
@@ -250,14 +250,14 @@ export default function NetworkSelector() {
   )
 
   useEffect(() => {
-    if (!chainId || promptedNetworkChange) return
+    if (!chainId || lastNetworkChange === chainId) return
     const newChainId = getParsedChainId(parsedQs)
 
     if (newChainId && chainId !== newChainId) {
       handleChainSwitch(newChainId, true)
-      setPromptedNetworkChange(true)
+      setLastNetworkChange(chainId)
     }
-  }, [handleChainSwitch, promptedNetworkChange, chainId, parsedQs])
+  }, [handleChainSwitch, lastNetworkChange, chainId, parsedQs])
 
   if (!chainId || !info || !library) {
     return null
--- a/src/components/Header/NetworkSelector.tsx
+++ b/src/components/Header/NetworkSelector.tsx
@@ -7,10 +7,12 @@ import useParsedQueryString from 'hooks/useParsedQueryString'
 import { ParsedQs } from 'qs'
 import { useCallback, useEffect, useRef, useState } from 'react'
 import { ArrowDownCircle, ChevronDown } from 'react-feather'
+import { useHistory } from 'react-router-dom'
 import { useModalOpen, useToggleModal } from 'state/application/hooks'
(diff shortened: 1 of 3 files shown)


hanniabu:
Not a reviewer, but I just tested this on multiple chains to link our community directly to our trade pairs on Arbitrum, Optimism, and Polygon and it works like a charm. Thank you for this!

src/components/Header/NetworkSelector.tsx
```diff
@@ -235,6 +248,16 @@ export default function NetworkSelector() {
     [dispatch, library, toggle]
   )
 
+  useEffect(() => {
+    if (!chainId || promptedNetworkChange) return
+    const newChainId = getParsedChainId(parsedQs)
+
+    if (newChainId && chainId !== newChainId) {
+      handleChainSwitch(newChainId)
```
moodysalem, line 256: ideally do not open the network popup when this is triggered

hanniabu:
Everything works as expected again. However, the Metamask popup requesting to switch networks is showing, which @moodysalem requested to not show above.

> ideally do not open the network popup when this is triggered

moodysalem:
> Everything works as expected again. However, the Metamask popup requesting to switch networks is showing, which @moodysalem requested to not show above.
> 
> 
> 
> > ideally do not open the network popup when this is triggered

Sorry I wasn't very clear, I meant the in-app switch network menu. The current behavior of having the metamask switch network popup is intended (let us know if you have any feedback!)

JFrankfurt:
I'm getting some behavior I didn't expect w/ `History`. (back/forward buttons in browser)

- am on mainnet
- use the dropdown to switch to polygon
- use my wallet to switch to optimism
- hit back button
- am on mainnet (!?)

willhennessy:
Also noticed some odd behavior with the add liquidity page URL. This page uses URL params to store token0, token1, and fee tier from the user input fields. 

Steps:
1. on mainnet, select ETH and DAI, and 0.3% tier
2. observe the url is add/ETH/0x6B175474E89094C44Da98b954EedeAC495271d0F/3000?chain=mainnet
3. switch to polygon
4. observe the input form values have been cleared, but the URL params remain. the chainID param switched to polygon though. add/ETH/0x6B175474E89094C44Da98b954EedeAC495271d0F/3000?chain=polygon

we should clear the URL params. (this might have been introduced by the chain switching feature and be unrelated to your PR, if so please ignore)

tinaszheng:
@JFrankfurt replacing the URL in place makes more sense in my head, where clicking on a dropdown value or changing networks in your wallet doesn't feel like visiting a new destination so im not sure i expect the same behavior 🤔

i was also thinking it should be easy for users to use the back button to escape the Uniswap site if necessary, and pushing new states on network change means we'll be crowding their browser history

but im also not super firm on this opinion, wdyt?

tinaszheng:
> Also noticed some odd behavior with the add liquidity page URL. This page uses URL params to store token0, token1, and fee tier from the user input fields.
> 
> Steps:
> 
> 1. on mainnet, select ETH and DAI, and 0.3% tier
> 2. observe the url is add/ETH/0x6B175474E89094C44Da98b954EedeAC495271d0F/3000?chain=mainnet
> 3. switch to polygon
> 4. observe the input form values have been cleared, but the URL params remain. the chainID param switched to polygon though. add/ETH/0x6B175474E89094C44Da98b954EedeAC495271d0F/3000?chain=polygon
> 
> we should clear the URL params. (this might have been introduced by the chain switching feature and be unrelated to your PR, if so please ignore)

@willhennessy looks like this is an existing issue with switching networks on that page - happy to address it in a separate PR!

JFrankfurt:
@tinaszheng I think you're right. I was confused by the behavior at first, but after clicking around a little more I get it and agree with you.

Separately--I notice that when I click between swap and pool, the query param flickers in/out like it's being added after the fact. Is that the case? If so, I think the links to those pages should include any existing network parameters.

tinaszheng:
@JFrankfurt ah yeah the network param is indeed added on top of the URL after mount if it doesn't already exist. this is so that we can keep the wallet network <-> URL in sync while maintaining the wallet network as source of truth. i think the flickering isnt too noticeable and adding in the current network to in-app links sounds like it could be messy 🤔

src/constants/chains.ts
```diff
@@ -18,6 +18,20 @@ export enum SupportedChainId {
   POLYGON_MUMBAI = 80001,
 }
 
+export const CHAIN_IDS_TO_NAMES = {
```
justindomingue, line 21: have you considered just putting that in `chain_ids` itself?

justindomingue (approved):

src/constants/chains.ts
tinaszheng: do you mean in `ALL_SUPPORTED_CHAIN_IDS`? i think its fine to separate it now

PR #3065, opened 2022-01-04
chore: merge widgets work into main

JFrankfurt:


package.json
```diff
@@ -99,32 +99,23 @@
...
     "react-redux": "^7.2.2",
     "react-router-dom": "^5.0.0",
     "react-scripts": "^4.0.3",
     "react-spring": "^8.0.27",
     "react-use-gesture": "^6.0.14",
-    "react-virtualized-auto-sizer": "^1.0.2",
-    "react-window": "^1.8.5",
-    "rebass": "^4.0.7",
     "redux-localstorage-simple": "^2.3.1",
+    "sass": "^1.45.1",
```
JFrankfurt, line 116: @zzmp what's this doing here?

JFrankfurt:
@justindomingue per our conversation--nuking the translation files here and will let the next generation step pick them up.

package.json
zzmp: `microbundle` is not as featureful as I thought, so there is an intermediate asset generation step which (1) generates React components out of svgs and (2) runs sass over our font files: `yarn assets:generate`.

The `sass` step in particular is done so that the font files can be set to use `font-display: block` (I think that `swap` is the default), which prevents the flash of unstyled content (FOUC for google-ability) that would otherwise occur. Note that it's in `devDependencies`, and not `dependencies`.
*Aside: this is done slightly differently in the interface (vs. the widget), which serves a static copy of the font for cache-ability, and is something we can change in the future if need be.*

.github/workflows/depcheck.yaml
```diff
@@ -0,0 +1,45 @@
+name: Bundle Dependency Check
+on:
+  push:
+    branches:
+      - main
+      - widgets
```
zzmp, line 6: Remove `widgets` from both `branches` lists

.github/workflows/integration-tests.yaml
```diff
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - main
+      - widgets
```
zzmp, line 7: Remove `widgets` from both `branches` lists

.github/workflows/lint.yml
```diff
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - main
+      - widgets
```
zzmp, line 7: Remove `widgets` from both `branches` lists

.github/workflows/unit-tests.yaml
```diff
@@ -3,9 +3,11 @@ on:
   push:
     branches:
       - main
+      - widgets
```
zzmp, line 6: Remove `widgets` from both `branches` lists

cosmos.config.json
```diff
@@ -1,7 +1,7 @@
 {
-  "staticPath": "public",
-  "watchDirs": ["src"],
+  "watchDirs": ["src/lib"],
```
zzmp, line 2: Change this to `src`, or add `src` to it, as we'll now be mixing the codebases while we implement hooks

src/constants/routing.ts
```diff
@@ -149,6 +149,7 @@ export const COMMON_BASES: ChainCurrencyList = {
     WRAPPED_NATIVE_CURRENCY[SupportedChainId.POLYGON_MUMBAI],
     WETH_POLYGON_MUMBAI,
   ],
+  [SupportedChainId.OPTIMISTIC_KOVAN]: [nativeOnChain(SupportedChainId.OPTIMISTIC_KOVAN)],
```
zzmp, line 152: Where did this come from? Could it have been introduced to `widgets` through an earlier merge of `main`, then deleted from `main`? I think it should be removed.

src/hooks/useActiveWeb3React.ts
```diff
@@ -0,0 +1,22 @@
...
+import { useWeb3React } from '@web3-react/core'
+import { default as useWidgetsWeb3React } from 'lib/hooks/useActiveWeb3React'
+
+import { NetworkContextName } from '../constants/misc'
+
+export default function useActiveWeb3React() {
+  const widgetsContext = useWidgetsWeb3React()
+  const interfaceContext = useWeb3React<Web3Provider>()
+  const interfaceNetworkContext = useWeb3React<Web3Provider>(
+    process.env.REACT_APP_IS_WIDGET ? undefined : NetworkContextName
```
zzmp, line 11: Is this branching strictly necessary? Or could `NetworkContextName` always be passed here?

src/lib/assets/fonts/index.scss
```diff
@@ -0,0 +1,22 @@
+// Use Inter mixin to set font-display: block.
```
zzmp, line 1: This is where `sass` is used - to generate only the fonts needed, and with `font-display: block`.

zzmp (changes requested):
Could you run `npx yarn-deduplicate` just to be sure you're not duplicating deps?

cosmos.config.json
JFrankfurt: yeeah, was wondering about this one

src/constants/routing.ts
JFrankfurt: Oh hmm, I thought it was already in. I think you're right about the source.

src/hooks/useActiveWeb3React.ts
JFrankfurt: It would work to do the branching at build time, right? just export the ternary?

JFrankfurt: nvm that won't work.

I'm hesitant to fiddle more w/ the functionality here. Can follow up on this later.

zzmp (approved):

PR #3066, opened 2022-01-04
fix: web3-react-fixes

NoahZinsmeister:
Uses exported EMPTY singleton instead of homegrown class.

Uses dedicated Url connector instead of Network.


src/lib/components/Web3Provider.tsx
```diff
@@ -30,9 +30,7 @@ function useConnector<T extends { new (actions: Actions, initializer: I): Connec
 
 export default function Web3Provider({ jsonRpcEndpoint, provider, children }: Web3ProviderProps) {
   const setNetwork = useUpdateAtom(networkAtom)
```
zzmp, line 32: Update this to `setUrl` and change the atom name to `urlAtom` for consistency

zzmp (approved):

PR #3067, opened 2022-01-05
chore: refactor state organization

ianlapham:
- Refactors state organization to move all state related files into the `lib/state` folder
- Specifically, moves `lib/components/swap/state.ts` to `lib/state/swap.ts`

Motivation is that as more state is added to the widget over time (add liquidity, remove, user settings, etc), all state logic can be contained in one folder. Result is that our `components` folder and files within it only include UI logic, and dont need to define state structure


zzmp (changes requested):
Could you also move `lib/utils/atoms.ts`, and namespace the connector state by moving it to something like `lib/state/web3.ts`?

ianlapham:
> Could you also move `lib/utils/atoms.ts`, and namespace the connector state by moving it to something like `lib/state/web3.ts`?

Yeah i think this is a good idea - updating with this

PR #3070, opened 2022-01-05
chore: add focus-visible for menu links

bigint:
This PR will increase a11y for users

Will pass this w3c criterion https://www.w3.org/WAI/WCAG21/Understanding/focus-visible.html

This only work for keyboard users ⌨️ 

https://user-images.githubusercontent.com/69431456/148234913-db5df2b9-a5a4-49bd-8f88-67d5d353758b.mov

image

image

### DIFF
--- a/src/components/Menu/index.tsx
+++ b/src/components/Menu/index.tsx
@@ -127,6 +127,11 @@ const MenuItem = styled(ExternalLink)`
     cursor: pointer;
     text-decoration: none;
   }
+  :focus-visible {
+    border-radius: 0.5rem;
+    text-decoration: none;
+    outline: 1px solid ${({ theme }) => theme.bg3};
+  }
 `
 
 const InternalMenuItem = styled(Link)`
@@ -175,6 +180,10 @@ const ToggleMenuItem = styled.button`
     color: ${({ theme }) => theme.text1};
     cursor: pointer;
     text-decoration: none;
+  }
+  :focus-visible {
+    border-radius: 0.5rem;
+    outline: 1px solid ${({ theme }) => theme.bg3};
   }
 `
 


bigint:
cc @tinaszheng

src/components/Menu/index.tsx
```diff
@@ -127,6 +127,11 @@ const MenuItem = styled(ExternalLink)`
...
     text-decoration: none;
   }
+  :focus-visible {
+    border-radius: 0.5rem;
+    text-decoration: none;
+    outline: 1px solid ${({ theme }) => theme.bg3};
+  }
 `
 
 const InternalMenuItem = styled(Link)`
```
zzmp, line 137: Should this use StyledInternalLink? If so, the :focus-visible be on that component, so it correctly displays for all internal links.

src/components/Menu/index.tsx
```diff
@@ -127,6 +127,11 @@ const MenuItem = styled(ExternalLink)`
     cursor: pointer;
     text-decoration: none;
   }
+  :focus-visible {
```
zzmp, line 130: Should this be on ExternalLink, so it correctly displays for all external links?

zzmp (changes requested):

zzmp:
Is there a checker available that checks for the w3c focus-visible criterion? How did you find these issues?

zzmp:
The w3c criterion is already achieved through changing the color, but I agree this is more obvious. Is there any reason this can't go in the already existing :focus selector? Why does there need to be a separate :focus-visible selector?

bigint:
@zzmp If we add it in `focus` it also gets visible for mouse users too

some platforms like Twitter, GitHub shows focus ring only for keyboard users

Screenshot 2022-01-06 at 8 11 13 AM

Screenshot 2022-01-06 at 8 12 56 AM

src/components/Menu/index.tsx
bigint: Yes I tried there, but the outline style is not matching for all links, its good to customize the focus-visible on demand!

src/components/Menu/index.tsx
bigint: https://github.com/Uniswap/interface/pull/3070#discussion_r780028757

ianlapham:
> 

I think the changes look good - but could we update to have color be same as secondary text - `text2` probably

bigint:
@ianlapham Updated 👍🏻

PR #3080, opened 2022-01-07
feat: populate the widget token selector

zzmp:
- adds swap defaults
- moves some utils to lib/utils for use in token list fetching
- expands token list fetching to allow TokenInfo[]
- loads the default token list in the token selector
  - does not sort, search, or load balances - this is deferred to future work to keep this PR more focused

For testing (via deployed Widgets), this will only be visible if accessing the token selector via Swap. It is not visible if going directly to the token selector (which now appears empty). This is working as intended.


src/lib/hooks/useTokenList.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export type TokenMap = Readonly<{ [tokenAddress: string]: WrappedTokenInfo }>
+type ChainTokenMap = Readonly<{ [chainId: number]: TokenMap }>
+type Mutable<T> = {
+  -readonly [P in keyof T]: Mutable<T[P]>
+}
+
+const chainTokenMapAtom = atom<ChainTokenMap>({})
+
+function tokensToTokenMap(tokens: TokenList | TokenInfo[]) {
```
JFrankfurt, line 17: For cases like this where we're duping code (sorry if there's a diff I'm missing) I think maybe we should create a new file in src/hooks or wherever it is currently, and import it in both its prior and new use locations, instead of duping.

ianlapham: yeah agreed we should just re-use in a shared file

src/lib/components/Swap/index.tsx
```diff
@@ -12,7 +14,38 @@ import Settings from './Settings'
...
+  tokenList: string | TokenList['tokens']
+  input: DefaultTokenAmount
+  output: DefaultTokenAmount
+}
+
+const DEFAULT_INPUT: DefaultTokenAmount = { address: 'ETH' }
+const DEFAULT_OUTPUT: DefaultTokenAmount = {}
+
+function useSwapDefaults(defaults: Partial<SwapDefaults> = {}): SwapDefaults {
+  const tokenList = defaults.tokenList || 'https://gateway.ipfs.io/ipns/tokens.uniswap.org'
```
JFrankfurt, line 32: noice

src/lib/components/Swap/index.tsx
```diff
@@ -12,7 +14,38 @@ import Settings from './Settings'
...
 import Toolbar from './Toolbar'
 
-export default function Swap() {
+interface DefaultTokenAmount {
+  address?: string | { [chainId: number]: string }
+  amount?: number
+}
+
+interface SwapDefaults {
+  tokenList: string | TokenList['tokens']
```
JFrankfurt, line 23: Are you sure `TokenList['tokens']` is what we want here?

src/lib/components/Swap/index.tsx
JFrankfurt: Thoughts on adding a schema validation check?

src/lib/components/Swap/index.tsx
zzmp: `TokenList['tokens'] === TokenInfo[]`, which is what we want. I didn't realize the equivalence until after writing this line before, so you see it as `TokenInfo[]` elsewhere. Will update.

src/lib/components/Widget.tsx
```diff
@@ -86,7 +85,7 @@ export default function Widget({
   dialog,
   className,
   onError,
-}: WidgetProps) {
+}: WidgetProps<{ children: ReactNode }>) {
```
JFrankfurt, line 88: `PropsWithChildren<WidgetProps>`

src/lib/theme/index.tsx
```diff
@@ -1,4 +1,4 @@
-import 'lib/assets/fonts/index.css'
+import '../assets/fonts/index.css' // microbundle requires relative css paths
```
JFrankfurt, line 1: ew :(

src/lib/utils/fetchTokenList.ts
```diff
@@ -0,0 +1,128 @@
...
+enum ValidationSchema {
+  LIST = 'list',
+  INFO = 'info',
+}
+
+/**
+ * Returns a ValidateFunction for token lists.
+ * Lazily loads ajv and the schema for performance.
+ */
+const getValidateFunction = (() => {
```
JFrankfurt, line 19: see https://github.com/Uniswap/interface/pull/3080/files#r780404165

yarn.lock
```diff
@@ -4937,7 +4937,7 @@
     "@typescript-eslint/types" "4.30.0"
     eslint-visitor-keys "^2.0.0"
 
-"@uniswap/default-token-list@^2.0.0", "@uniswap/default-token-list@^2.1.0":
+"@uniswap/default-token-list@^2.0.0":
```
JFrankfurt, line 4940: mistake?

src/lib/theme/index.tsx
zzmp: microbundle has been a let down, but I think transitioning would be a huge time sink, and there's better things we could be spending time on. See `yarn assets:generate`.

src/lib/types.d.ts
```diff
@@ -4,5 +4,5 @@ export interface Token {
   chainId: number
   decimals: number
   address: string
-  logoURI: string
+  logoURI?: string
```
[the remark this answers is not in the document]
zzmp: I want this file to be eventually deleted. Anywhere this type is used should eventually be replaced with `WrappedTokenInfo` or `Currency`. It was just added to make UI development faster.

src/lib/utils/fetchTokenList.ts
zzmp: src/utils/fetchTokenList.ts has been renamed to this file, so no duplication here :)

yarn.lock
zzmp: intentional: the deleted version is from an explicit dependency in the package.json which was not actual used, and so deleted. (this was found with `npx yarn-deduplicate`, but manually verified.)

src/lib/components/Swap/index.tsx
zzmp: the passed token list (whether it’s an object literal or a url string) will go through ajv to be validated, that validation just doesn’t happen until `useTokenList`.

BigsteppA1:
Ok

BigsteppA1:
Ok

BigsteppA1:
Ok

src/lib/components/Swap/index.tsx
```diff
@@ -12,7 +14,38 @@ import Settings from './Settings'
 import SwapButton from './SwapButton'
 import Toolbar from './Toolbar'
 
-export default function Swap() {
+interface DefaultTokenAmount {
```
ianlapham, line 17: How come we dont just use TokenAmount from SDK? In my refactor all the input and output is changed to represent this - so maybe best to address in #3081

src/lib/hooks/useColor.ts
```diff
@@ -16,26 +16,23 @@ function UriForEthToken(address: string) {
  */
 async function getColorFromToken(token: Token, cb: (color: string | undefined) => void = () => void 0) {
```
ianlapham, line 17: Changed this to accept `Currency` in #3081  but again - can probably address in that PR 

reasoning was basically mimic the existing useColor hook

ianlapham (commented):
Looks solid overall - like the renaming in the token map construction a lot 

Had a few ideas about some things with `Currency` but like i said it might be better to review in a different pr

src/lib/hooks/useColor.ts
JFrankfurt: @ianlapham Did you run into #3085 in your build process when you did that?

src/lib/components/Swap/index.tsx
zzmp: These defaults are simpler interfaces passed as a prop when setting up the widget, so they don't need to adhere to the full TokenAmount interface.

src/lib/hooks/useColor.ts
ianlapham: Yeah i think #3081 is the right time to do this - @JFrankfurt i actually didnt run into this somehow - but my build wasnt very stable can address in #3081

PR #3083, opened 2022-01-10
fix(widgets): correctly position the arrows

JFrankfurt:
The reverse button was too high, which caused collision with large balances.


src/lib/components/Swap/ReverseButton.tsx
```diff
@@ -8,9 +8,10 @@ import Button from '../Button'
 import Row from '../Row'
 
 const ReverseRow = styled(Row)`
-  bottom: -1.5em;
+  left: 50%;
   position: absolute;
-  width: 100%;
+  top: 118px;
```
zzmp, line 13: widgets uses `em` instead of `px` throughout

PR #3087, opened 2022-01-11
refactor: x-platform multicall

zzmp:
- Moves multicall into `src/lib`
- Moves block number from the application state into its own jotai atom. This allows it (and multicall, which relies on it) to be used outside of the redux application (ie in widgets)

`useTokenBalances` (in `useQueriedTokenList.ts`) was used as a motivation for this change, so some changes to it and related files (ie `TokenSelect/`) are included as well.


src/lib/components/TokenSelect/useQueriedTokenList.ts
```diff
@@ -0,0 +1,33 @@
...
+  const debouncedQuery = useDebounce(query, 200)
+  const filteredTokens = useMemo(
+    () => sortedTokens.filter(createTokenFilterFunction(debouncedQuery)),
+    [debouncedQuery, sortedTokens]
+  )
+
+  // Re-sorts tokens (ie to bump up exact matches)
+  const sortedFilteredTokens = useSortedTokensByQuery(filteredTokens, debouncedQuery)
+
+  // TODO(zzmp): Include native Currency
```
ianlapham, line 31: This is a requirement for merge right? If app will be using this

src/state/transactions/updater.tsx
```diff
@@ -112,11 +112,6 @@ export default function Updater(): null {
                 hash,
                 isL2 ? L2_TXN_DISMISS_MS : DEFAULT_TXN_DISMISS_MS
               )
-
-              // the receipt was fetched before the block, fast forward to that block to trigger balance updates
-              if (receipt.blockNumber > lastBlockNumber) {
```
ianlapham, line 117: Can we keep this logic by just adding a jotai hook to fast forward block number

src/lib/hooks/useBlockNumber.ts
```diff
@@ -0,0 +1,62 @@
...
+  const block = useBlock()
+  useEffect(() => {
+    setBlock(block)
+  }, [block, setBlock])
+  return null
+}
+
+/** Requires that BlockUpdater be installed in the DOM tree. */
+export default function useBlockNumber() {
+  return useAtomValue(blockAtom)
```
ianlapham, line 61: the old code returns undefined if no `chainId` , should we also keep that pattern here?

src/state/application/hooks.ts
```diff
@@ -2,16 +2,9 @@ import { DEFAULT_TXN_DISMISS_MS } from 'constants/misc'
...
 import { useAppDispatch, useAppSelector } from 'state/hooks'
 
-import { useActiveWeb3React } from '../../hooks/web3'
 import { AppState } from '../index'
 import { addPopup, ApplicationModal, PopupContent, removePopup, setOpenModal } from './reducer'
 
-export function useBlockNumber(): number | undefined {
-  const { chainId } = useActiveWeb3React()
-
-  return useAppSelector((state: AppState) => state.application.blockNumber[chainId ?? -1])
```
ianlapham, line 12: This is that old check for `chainId`

ianlapham (commented):
Just a few nits about checking for chainId

src/index.tsx
```diff
@@ -39,6 +40,7 @@ function Updaters() {
       <UserUpdater />
       <ApplicationUpdater />
       <TransactionUpdater />
+      <BlockUpdater />
       <MulticallUpdater />
```
zzmp, line 44: Do I need to include a MulticallUpdater in the widget?

src/lib/components/TokenSelect/useQueriedTokenList.ts
zzmp: This is a requirement for launch, but may be deferred to a later PR

src/lib/hooks/useBlockNumber.ts
zzmp: Good call. I'll add this back in.

src/state/transactions/updater.tsx
zzmp: Done.

src/index.tsx
zzmp: > you have to render it or you won't get any data

https://uniswapteam.slack.com/archives/C023ZR8Q1DE/p1641946622034500?thread_ts=1641946408.032600&cid=C023ZR8Q1DE

PR #3093, opened 2022-01-11
fix: set wallectconnect chainId explicitly to ethereum

PeterTheOne:
Sets wallectconnect chainId explicitly to Ethereum, to make a point that it is only configured for Ethereum mainnet, which is the current state.

To support walletconnect on networks other than Ethereum this value needs to be set to the correct chainId!

When the network is changed before connecting via walletconnect (see https://github.com/Uniswap/interface/blob/main/src/components/Header/NetworkSelector.tsx#L227 ) the `chainId` value needs to be set to the correct `chainId`.

### DIFF
--- a/src/connectors/index.ts
+++ b/src/connectors/index.ts
@@ -34,6 +34,7 @@ export const walletconnect = new WalletConnectConnector({
   supportedChainIds: ALL_SUPPORTED_CHAIN_IDS,
   rpc: INFURA_NETWORK_URLS,
   qrcode: true,
+  chaiId: 1,
 })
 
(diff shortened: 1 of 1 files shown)


dougransom:
This is affecting me i think, it seems when i connect to polygon in the pool ap it is pulling values from the ethereum chain.  i am using walletconnect & safepal., since i have a hw wallet i cannot switch to metamask.

dougransom:
is this going to be merged?

vm:
We upgraded the version of web3-react recently. This issue should be resolved, but let us know if not.

PR #3094, opened 2022-01-11
chore: import from ethers

zzmp:
- Imports @ethersproject from the umbrella package, ethers

Importing from ethers may prevent some tree-shaking, but it leads to a better devX. In particular, microbundle did not recognize nested packages, so it required any used @ethersproject packages to be listed explicitly in addition to the umbrella package.


moodysalem:
I don't understand, is the better devx that you don't have to install `@ethersproject` dependencies you import?

zzmp:
> I don't understand, is the better devx that you don't have to install `@ethersproject` dependencies you import?

Yes. Right now, they are being installed twice: once through `ethers`, and again explicitly so the `microbundle` recognizes it as a dependency. They should only be installed once.

`ethers` also installs them using pegged versions, so it may lead to version mismatch and double-installs in the future.

.eslintrc.json
```diff
@@ -59,28 +59,24 @@
       {
         "paths": [
           {
-            "name": "lodash",
-            "message": "Please import from 'lodash/module' directly to support tree-shaking."
-          },
```
JFrankfurt, line 64: intentional removal?

JFrankfurt (approved):

.eslintrc.json
zzmp: We no longer use lodash, so I took it out as tech debt.

zzmp:
> why not just remove the ethers dependency?

that would have been fine, too. it's the same maintainer, so this just seemed easier for our own ongoing maintenance.

PR #3104, opened 2022-01-12
chore: import @ethersprojects directly

zzmp:
- Removes dependency on ethers
- Restores eslint preventing imports from ethers
- Directly depends on @ethersprojects, using compatible versions (eg ^5.4.0, instead of pinning 5.4.0; this will improve lib compatibility)


JFrankfurt (approved):

package.json
```diff
@@ -134,7 +134,7 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test --env=./custom-test-env.js",
-    "test:e2e": "start-server-and-test 'serve build -l 3000' http://localhost:3000 'cypress run'",
+    "test:e2e": "start-server-and-test 'serve build -l 3000' http://localhost:3000 'cypress run --record'",
```
moodysalem, line 137: accident? you can add record argument to the ci job if you use this script locally a lot

package.json
```diff
@@ -156,6 +156,19 @@
   },
   "license": "GPL-3.0-or-later",
   "dependencies": {
+    "@ethersproject/abi": "^5.4.1",
```
moodysalem, line 159: slightly concerned w the diff patch versions. maybe just set them all to the same one and do a yarn install?

moodysalem (approved):
couple nits

package.json
zzmp: I must have accidentally removed the `--record` from ci in #3094. I'm fine removing it (as we don't pay for recording), but this is just reverting it to what it was.

package.json
zzmp: I didn't want to change any versions in this PR, so I copied the pinned versions that the `ethers` package was using, prepended them with `^` for compatibility, and manually edited the `yarn.lock` to make sure the versions weren't bumped. Changing patch versions to all match would effectively change the versions we use.

package.json
moodysalem: feel free to remove it, but should make the corresponding change to the ci (add `--record` [here](https://github.com/Uniswap/interface/blob/aa97ec01d311b8e2adb8dc736896f12ccf47b044/.github/workflows/integration-tests.yaml#L47))

PR #3105, opened 2022-01-12
refactor: split chain info from chains

zzmp:
Splits chain info from chains.ts, so that chains.ts will be usable by `src/lib` without bringing in unnecessary dependencies.


JFrankfurt:
is the goal just to avoid bundling these?
```

import ethereumLogoUrl from 'assets/images/ethereum-logo.png'
import arbitrumLogoUrl from 'assets/svg/arbitrum_logo.svg'
import optimismLogoUrl from 'assets/svg/optimistic_ethereum.svg'
import polygonMaticLogo from 'assets/svg/polygon-matic-logo.svg'
import ms from 'ms.macro'

```

nit: diff `chain` and `chainInfo` files is a little awkward

zzmp:
It's to avoid bundling the assets _and_ infura urls.

If you think we'll eventually need the assets, I can instead move them to `src/lib/assets` so that they're compatible with the widget lib. wdyt?

zzmp:
This imports the svg/png files as string urls, which is not supported by `microbundle`, so it can't be done without modifying the build pipeline for the widget. I think this is a simpler, unblocking solution than changing the build pipeline.

ianlapham (approved):
LGTM

PR #3108, opened 2022-01-12
feat: mv block number to atom

zzmp:
- Moves block number from redux application state to an independent atom.

Allows redux-multicall to function without a dependency on application state. This will be necessary for multicall to be shared between interface and widget.


src/lib/hooks/useBlockNumber.ts
```diff
@@ -0,0 +1,69 @@
...
+    setBlock(block)
+  }, [block, setBlock])
+  return null
+}
+
+/** Requires that BlockUpdater be installed in the DOM tree. */
+export default function useBlockNumber(): number | undefined {
+  const { chainId } = useActiveWeb3React()
+  const block = useAtomValue(blockAtom)
+  return chainId ? block : undefined
```
ianlapham, line 64: Will this behave strangely on other chains - shouldnt we index the block by chainId?

src/state/application/hooks.ts
```diff
@@ -1,17 +1,10 @@
...
 import { useCallback, useMemo } from 'react'
 import { useAppDispatch, useAppSelector } from 'state/hooks'
 
 import { AppState } from '../index'
 import { addPopup, ApplicationModal, PopupContent, removePopup, setOpenModal } from './reducer'
 
-export function useBlockNumber(): number | undefined {
-  const { chainId } = useActiveWeb3React()
-
-  return useAppSelector((state: AppState) => state.application.blockNumber[chainId ?? -1])
```
ianlapham, line 12: See here how block is indexed by chain id

ianlapham (changes requested):
Wondering how atom state works across chain

src/lib/hooks/useBlockNumber.ts
zzmp: If there's no chainId, or if the chainId is chainged, the atom should be cleared, so I think this will behave fine across chains. wdyt?

ianlapham: Yep looks like in original state gets reset  in the useEffect when chainId changes

ianlapham (approved):

PR #3110, opened 2022-01-12
add error reporting component, INTEGRATION ERROR type, and Missing provider error

JFrankfurt:


src/lib/components/Error/ErrorReporter.tsx
```diff
@@ -0,0 +1,23 @@
+import { useAtomValue } from 'jotai/utils'
+import { providerAtom } from 'lib/state/web3'
+import { useEffect } from 'react'
+import { EMPTY } from 'widgets-web3-react/empty'
+
+class IntegrationError extends Error {
```
zzmp, line 6: Can you move error definitions to a separate file? Maybe `src/lib/errors.ts`?

src/lib/components/Error/ErrorReporter.tsx
```diff
@@ -0,0 +1,23 @@
...
+  constructor(message: string) {
+    super(message)
+    this.name = 'INTEGRATION ERROR'
+  }
+}
+
+const missingProviderError = new IntegrationError('Missing provider')
```
zzmp: better to define this inline for readability - if we're throwing an error, whether it's a singleton doesn't matter as much.

src/lib/components/Error/ErrorReporter.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export default function ErrorReporter() {
+  const [connector] = useAtomValue(providerAtom)
+  useEffect(() => {
+    if (connector === EMPTY) {
```
zzmp, line 10: Doesn't this start as empty? Will this always throw?

src/lib/components/Error/ErrorReporter.tsx
```diff
@@ -0,0 +1,23 @@
...
+      throw missingProviderError
```
zzmp, line 11: Do we need to throw on this error? We already have defined UI for when there is no connector.

src/lib/state/web3.ts
```diff
@@ -12,6 +13,14 @@ export type Web3ReactState = [Connector, Web3ReactHooks]
 
 export const urlAtom = atomWithDefault<Web3ReactState>(() => EMPTY_CONNECTOR)
 export const injectedAtom = atomWithDefault<Web3ReactState>(() => EMPTY_CONNECTOR)
+export const providerAtom = atom<Web3ReactState>((get) => {
+  const url = get(urlAtom)
+  const injected = get(injectedAtom)
+  if (injected[0] !== EMPTY) {
+    return get(injectedAtom)
```
zzmp, line 20: `return injected`.

nit: this could also be condensed using a ternary in the return statement

src/lib/components/Error/ErrorReporter.tsx
```diff
@@ -0,0 +1,23 @@
...
+class IntegrationError extends Error {
+  constructor(message: string) {
+    super(message)
+    this.name = 'INTEGRATION ERROR'
+  }
+}
+
+const missingProviderError = new IntegrationError('Missing provider')
+
+export default function ErrorReporter() {
```
zzmp, line 7: What's the purpose of this component?

JFrankfurt: rendering it alongside children at the lowest level of the `Widget` wrapper to trigger the error boundary. Will extend w/ more error tests as necessary.

src/lib/errors.ts
```diff
@@ -0,0 +1,8 @@
+class IntegrationError extends Error {
+  constructor(message: string) {
+    super(message)
+    this.name = 'INTEGRATION ERROR'
+  }
+}
+
+export const missingProviderError = new IntegrationError('Missing provider')
```
zzmp, line 8: Error definitions that I'm used to (including native errors) just export classes, and have you check equality with `instanceof`. Can this error instead extend an integration error, and just be exported as a class?

```
export class MissingProviderError extends IntegrationError {
  constructor() {
    super('Missing provider')
  }
}
```

This would change its usage in the generator to `throw new MissingProviderError()`.

src/lib/errors.ts
```diff
@@ -0,0 +1,8 @@
+class IntegrationError extends Error {
```
zzmp, line 1: export this class

src/lib/components/Error/ErrorReporter.tsx
JFrankfurt: Looks like I was out of the loop on how this actually works. I was thinking EMPTY is used primarily as a filler value and the component would always render w/ the integrator's provider value first.

src/lib/components/Widget.tsx
```diff
@@ -90,19 +91,19 @@ export type WidgetProps<T extends JSXElementConstructor<any> | undefined = undef
...
-  locale = DEFAULT_LOCALE,
-  provider,
-  jsonRpcEndpoint,
-  width = 360,
-  dialog,
-  className,
-  onError,
-}: PropsWithChildren<WidgetProps>) {
+export default function Widget(props: PropsWithChildren<WidgetProps>) {
+  const {
```
zzmp, line 95: Why do you prefer destructuring outside of the function signature?

zzmp: Is it to make it easier to pass all of the props to ErrorGenerator?

src/lib/errors.ts
```diff
@@ -0,0 +1,12 @@
+export class IntegrationError extends Error {
+  constructor(message: string) {
+    super(message)
+    this.name = 'INTEGRATION ERROR'
```
zzmp, line 4: Can this be changed to Sentence or Title Case?

src/lib/errors.ts
```diff
@@ -0,0 +1,12 @@
...
+  }
+}
+
+export class MissingProviderError extends IntegrationError {
```
zzmp, line 8: Unused - can this be removed?

zzmp (approved):
Some nits, but lgtm

src/lib/components/Swap/index.tsx
```diff
@@ -39,6 +39,8 @@ function useSwapDefaults(defaults: Partial<SwapDefaults> = {}): SwapDefaults {
 }
 
 export interface SwapProps {
+  convenienceFee?: number
+  convenienceFeeRecipient?: string // todo: improve typing to require recipient when fee is set
```
zzmp, line 43: nit: all-caps the TODO so it is more searchable

src/lib/errors.ts
```diff
@@ -1,12 +1,12 @@
...
     super(message)
-    this.name = 'INTEGRATION ERROR'
+    this.name = 'Integration Error'
   }
 }
-
-export class MissingProviderError extends IntegrationError {
-  constructor() {
-    super(`Missing provider`)
+export class ChainIdError extends Error {
```
zzmp, line 7: nit: newline between classes

src/lib/components/Error/ErrorGenerator.tsx
```diff
@@ -1,13 +1,43 @@
...
+import { SwapWidgetProps } from 'lib/index'
 import { useEffect } from 'react'
 
 export default function ErrorGenerator(swapWidgetProps: SwapWidgetProps) {
+  const { chainId } = useActiveWeb3React()
   const { jsonRpcEndpoint, provider } = swapWidgetProps
   useEffect(() => {
     if (!provider && !jsonRpcEndpoint) {
-      throw new IntegrationError('No provider or jsonRpcEndpoint provided')
+      throw new IntegrationError('This widget requires provider or jsonRpcEndpoint prop.')
```
zzmp, line 12: nit: "...requires a provider or jsonRpcEndpoint."

src/lib/components/Error/ErrorGenerator.tsx
```diff
@@ -1,13 +1,43 @@
...
     }
   }, [provider, jsonRpcEndpoint])
+
+  useEffect(() => {
+    if (chainId && !ALL_SUPPORTED_CHAIN_IDS.includes(chainId)) {
+      throw new ChainIdError('Please switch to a network supported by the Uniswap Protocol.')
```
zzmp, line 18: nit: I think we should omit "Please" in all of these errors

src/lib/components/Error/ErrorGenerator.tsx
```diff
@@ -0,0 +1,57 @@
...
+      throw new IntegrationError('Set widget width to at least 300px.')
+    }
+  }, [width])
+
+  // convenience fee constraints
+  const { convenienceFee, convenienceFeeRecipient } = swapWidgetProps
+  useEffect(() => {
+    if (convenienceFee) {
+      if (convenienceFee > 100 || convenienceFee < 0) {
+        throw new IntegrationError('Set widget convenienceFee to at least 400px.')
```
willhennessy, line 37: @JFrankfurt I think 400px is a typo here.  did you mean "ConvenienceFee must be between 0 and 1.0 inclusive."

PR #3112, opened 2022-01-12
feat: add multicall to widget

zzmp:


src/lib/components/Widget.tsx
```diff
@@ -64,6 +67,15 @@ const WidgetWrapper = styled.div<{ width?: number | string }>`
   }
 `
 
+function Updater() {
```
ianlapham, line 70: Nit - can we changer to `Updaters`

ianlapham (commented):
Just small nit rest looks good

src/lib/components/Widget.tsx
zzmp: lol my personal nit was to change from `Updaters` to singular. I'll use the plural 😆

src/hooks/useContract.ts
```diff
@@ -38,7 +38,8 @@ import useActiveWeb3React from 'hooks/useActiveWeb3React'
 import { useMemo } from 'react'
 import { NonfungiblePositionManager, Quoter, UniswapInterfaceMulticall } from 'types/v3'
 import { V3Migrator } from 'types/v3/V3Migrator'
-import { getContract } from 'utils'
+
+import { getContract } from '../utils'
```
zzmp, line 42: This is changed as an alternative to adding `utils` to paths for `tsconfig.lib.ts`.

PR #3114, opened 2022-01-13
feat: sort the widget token select

zzmp:
- Fetches balances and sorts the widget token select
- Does not include native currency in the token select (deferred to a future PR)


ianlapham (approved):
Looks good - one non-block, why does the content in querying.ts need its own file

zzmp:
> Looks good - one non-blocking comment, why does the content in querying.ts need its own file

Personal preference. useTokenList uses distinct logic from useQueryTokenList, but I wanted them to be exported from the same file (because the base logic and use cases are the same). I didn't want to make the useTokenList.ts file too complex to read through, so I split it.

PR #3116, opened 2022-01-14
feat: add survey popup for survey monkey

ianlapham:
- adds popup link to survey monkey 
- uses random number to only show to 1% of audience 
- allow user to dismiss - save users view as boolean to show popup or not 
- this may be hard to test - as only shown 1% of time 

Screen Shot 2022-01-13 at 7 19 05 PM


src/components/Popups/SurveyPopup.tsx
```diff
@@ -0,0 +1,107 @@
...
+import ReactGA from 'react-ga'
+import { useShowSurveyPopup } from 'state/user/hooks'
+import styled from 'styled-components/macro'
+import { ExternalLink, ThemedText } from 'theme'
+
+import BGImage from '../../assets/images/survey-orb.svg'
+import useTheme from '../../hooks/useTheme'
+
+const Wrapper = styled(AutoColumn)`
+  background: #edeef2;
```
justindomingue, line 15: you can't get an approximate color from the theme? This is bg2 in darkmode

src/components/Popups/SurveyPopup.tsx
```diff
@@ -0,0 +1,107 @@
...
+    max-width: 100%;
+  `}
+`
+
+const BGOrb = styled.img`
+  position: absolute;
+  right: -64px;
+  top: -64px;
+  width: 180px;
+  z-index: 1;
```
justindomingue, line 34: we have a `Z_INDEX` object at `src/theme`

src/components/Popups/SurveyPopup.tsx
```diff
@@ -0,0 +1,107 @@
...
+  height: 20px;
+  stroke: #7c7c80;
+  z-index: 2;
+  :hover {
+    cursor: pointer;
+    opacity: 0.8;
+  }
+`
+
+const END_TIMESTAMP = 1642215971 // Jan 15th
```
justindomingue, line 51: when is this launching? Jan 15 is p close

src/components/Popups/SurveyPopup.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+export default function SurveyPopup() {
+  const theme = useTheme()
+  const [showPopup, setShowSurveyPopup] = useShowSurveyPopup()
+
+  // show popup to 1% of users
+  useEffect(() => {
+    // has not visited page during A/B testing if undefined
+    if (showPopup === undefined) {
+      const id = Math.floor(Math.random() * 101) // random between 0 -> 100
```
justindomingue, line 61: can't you avoid some math and just do `Math.random() < 0.01`

src/components/Popups/SurveyPopup.tsx
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }, [setShowSurveyPopup, showPopup])
+
+  // limit survey to 24 hours based on timestamp
+  const timestamp = useCurrentBlockTimestamp()
+  const durationOver = timestamp ? timestamp.toNumber() > END_TIMESTAMP : false
+
+  return (
+    <>
+      {!showPopup || durationOver ? null : (
```
justindomingue, line 80: there's something to be said here about making `SurveyPop` accept `children` element instead so it's more reusable

src/components/Popups/SurveyPopup.tsx
```diff
@@ -0,0 +1,107 @@
...
+              })
+              setShowSurveyPopup(false)
+            }}
+          />
+          <BGOrb src={BGImage} />
+          <ExternalLink href="https://www.surveymonkey.com/r/YGWV9VD">
+            <RowFixed>
+              <MessageCircle stroke={theme.black} size="20px" strokeWidth="1px" />
+              <ThemedText.White fontWeight={600} color={theme.black} ml="6px">
+                Tell us what you think ↗
```
justindomingue, line 96: i18n

src/components/Popups/SurveyPopup.tsx
```diff
@@ -0,0 +1,107 @@
...
+              </ThemedText.White>
+            </RowFixed>
+          </ExternalLink>
+          <ThemedText.Black style={{ zIndex: 3 }} fontWeight={400} fontSize="12px" color={theme.black}>
+            Take a 10 minute survey to help us improve your experience in the Uniswap app.
```
justindomingue, line 101: ditto i18n

src/components/Popups/SurveyPopup.tsx
```diff
@@ -0,0 +1,107 @@
...
+          <BGOrb src={BGImage} />
+          <ExternalLink href="https://www.surveymonkey.com/r/YGWV9VD">
+            <RowFixed>
+              <MessageCircle stroke={theme.black} size="20px" strokeWidth="1px" />
+              <ThemedText.White fontWeight={600} color={theme.black} ml="6px">
+                Tell us what you think ↗
+              </ThemedText.White>
+            </RowFixed>
+          </ExternalLink>
+          <ThemedText.Black style={{ zIndex: 3 }} fontWeight={400} fontSize="12px" color={theme.black}>
```
justindomingue, line 100: zindex can also be from the zindex object

src/components/Popups/SurveyPopup.tsx
```diff
@@ -0,0 +1,107 @@
...
+`
+
+const END_TIMESTAMP = 1642215971 // Jan 15th
+
+export default function SurveyPopup() {
+  const theme = useTheme()
+  const [showPopup, setShowSurveyPopup] = useShowSurveyPopup()
+
+  // show popup to 1% of users
+  useEffect(() => {
```
justindomingue, line 58: also from a code organization standup i almost want this to be in a `SurveyManager` instead of inside the popup itself. Not a big deal though since it's still pretty small

src/components/Popups/index.tsx
```diff
@@ -59,6 +60,9 @@ export default function Popups() {
   // get all popups
   const activePopups = useActivePopups()
 
+  // show survey popup if user has not closed
+  const [showSurveyPopup] = useShowSurveyPopup()
```
justindomingue, line 64: seems a bit like a smell to have to grab the popup state from outside the popup 🤔

justindomingue (commented):
some comments that could make it more future-proof but otherwise LGTM

src/components/Popups/SurveyPopup.tsx
ianlapham: goal is by EOD today so updatying timestamp right before

src/components/Popups/SurveyPopup.tsx
ianlapham: yeah i think if we had more time this would be the way - if we ever run another one i think we should

src/components/Popups/index.tsx
ianlapham: Yeah i think this could be best fixed in a survey manager component - the dynamic width that this is being used for does feel kinda weird

PR #3119, opened 2022-01-14
feat: polish copy on import token flow, remove tokenlist.svg

willhennessy:
polishing copy on the import token flow based on December meeting: https://www.notion.so/uniswaplabs/Copy-Polish-Session-d9ce25daf58543be86b71a99278ac371


src/components/SearchModal/ImportToken.tsx
```diff
@@ -47,22 +50,25 @@ export function ImportToken(props: ImportProps) {
       <PaddedColumn gap="14px" style={{ width: '100%', flex: '1 1' }}>
         <RowBetween>
           {onBack ? <ArrowLeft style={{ cursor: 'pointer' }} onClick={onBack} /> : <div />}
-          <ThemedText.MediumHeader>
-            <Plural value={tokens.length} one="Import token" other="Import tokens" />
-          </ThemedText.MediumHeader>
+          <ThemedText.MediumHeader>Import token</ThemedText.MediumHeader>
```
willhennessy, line 53: I'm cheating here - let me know if this is bad. But I've never seen this modal received more than one token. When would that happen

src/components/SearchModal/ImportToken.tsx
```diff
@@ -47,22 +50,25 @@ export function ImportToken(props: ImportProps) {
...
-            <Trans>
-              This token doesn&apos;t appear on the active token list(s). Make sure this is the token that you want to
-              trade.
-            </Trans>
-          </ThemedText.Body>
+          {chainId && (
+            <ThemedText.Body fontWeight={400} fontSize={16}>
+              <Trans>
+                This token is not on your active token lists. Please check the{' '}
+                <ExternalLink href={getExplorerLink(chainId, tokens[0].address, ExplorerDataType.ADDRESS)}>
```
willhennessy, line 65: I'm cheating again here. Open to a less hacky solution to get the token's address. This works for the common (only afaik?) use case of importing a single token.

src/components/SearchModal/ImportToken.tsx
zzmp: Not sure when/if that would happen, but this will still need to be wrapped in `Trans` or `Plural` for translations.

src/components/SearchModal/ImportToken.tsx
```diff
@@ -47,22 +50,25 @@ export function ImportToken(props: ImportProps) {
...
+                  Etherscan page
+                </ExternalLink>{' '}
+                to verify this is the right token.
```
zzmp, line 68: Could "right token" be clearer? Maybe "the token you intended to trade"?

src/components/SearchModal/ImportToken.tsx
```diff
@@ -33,6 +35,7 @@ interface ImportProps {
 export function ImportToken(props: ImportProps) {
   const { tokens, list, onBack, onDismiss, handleCurrencySelect } = props
```
zzmp, line 36: If tokens is always a list of one, replace it with `token: Token` in ImportProps and then fix it where those props are used. Otherwise, if it's possible for tokens to be plural, it should be supported correctly.

zzmp (changes requested):
Either update `ImportProps` to restrict imports to a singular token, or support plural cases.

src/components/SearchModal/ImportToken.tsx
zzmp: @willhennessy I think, if it's never going to be plural, this is the best way forward.

willhennessy:
Maybe you can help me solve the root cause:  the Plural tag was always returning the string "tokens" even when the `tokens` array had only 1 element in console.log.  Any idea why?

PR #3122, opened 2022-01-14
refactor: mv token hooks to lib

zzmp:
Separates the hook logic from application state so that the logic may be re-used by widgets.


src/lib/hooks/useTokenList/index.ts
```diff
@@ -44,6 +46,19 @@ export default function useTokenList(list?: string | TokenInfo[]): WrappedTokenI
   }, [chainId, chainTokenMap])
 }
 
+export type TokenMap = { [address: string]: Token }
+
+export function useTokenMap(): TokenMap {
```
ianlapham, line 51: Non-blocking - but it kinda confuses me that this is also called TokenMap but its different that the type TokenMap in utils below

ianlapham (approved):

PR #3123, opened 2022-01-14
feat: adding support for Kintsugi testnet

asood123:


aballerr:
I talked with Aseem,

He thinks this logic was merged in at one point and then potentially removed since it's no longer relevant.  With the merge occurring in a week, it's likely any new test needed will be resolved outside of this pr

PR #3124, opened 2022-01-15
feat: include native currency in widget select

zzmp:
- Updates widgets swap state to use Currency (and deals with downstream updates)
- Refactors logoURI code to a new lib/hooks/useCurrencyLogoURIs
- Adds native currency to useQueryTokenList


cosmos.override.js
```diff
@@ -1,14 +1,25 @@
...
+      return new HtmlWebpackPlugin({
+        templateContent: '<body></body>',
+      })
+    }
+    if (plugin instanceof DefinePlugin) {
+      return new DefinePlugin({
+        ...plugin.definitions,
+        'process.env': {
+          ...plugin.definitions['process.env'],
+          REACT_APP_IS_WIDGET: true,
```
JFrankfurt, line 19: 👏

src/lib/components/Swap/Summary/Details.tsx
```diff
@@ -24,8 +24,8 @@ function Detail({ label, value }: DetailProps) {
 
 interface DetailsProps {
   swap: Required<State>['swap']
-  input: Token
-  output: Token
+  input: Currency
+  output: Currency
```
JFrankfurt, line 28: 👏

src/lib/hooks/useCurrencyColor.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  colors.set(key, color)
+  return cb(color)
+}
+
+async function getColorFromUriPath(uri: string): Promise<string | undefined> {
+  try {
+    const palette = await Vibrant.from(uri).getPalette()
+    return palette.Vibrant?.hex
+  } catch {}
```
JFrankfurt, line 40: should we use this?

```
let v = Vibrant.from(src)
  .useQuantizer(Vibrant.Quantizer.WebWorker)
```

src/lib/hooks/useCurrencyColor.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }, [token, logoURIs, theme.tokenColorExtraction])
+}
+
+export default function useCurrencyColor(token?: Currency) {
+  const [color, setColor] = useState<string | undefined>(undefined)
+  const theme = useTheme()
+  const logoURIs = useCurrencyLogoURIs(token)
+
+  useLayoutEffect(() => {
```
JFrankfurt, line 60: I'm not super familiar w/ this hook so I looked up its usage. It looks like it's supposed to be used for reading rendered dom stuff. At first I thought it was appropriate here because of `getColorFromLogoURIs`, but after looking at the implementation, I'm not so sure.

I thiiink this should be a normal `useEffect`

src/lib/hooks/useCurrencyLogoURIs.ts
```diff
@@ -0,0 +1,59 @@
+import { Currency } from '@uniswap/sdk-core'
+import { SupportedChainId } from 'constants/chains'
+import useHttpLocations from 'hooks/useHttpLocations'
+import { useMemo } from 'react'
+import { WrappedTokenInfo } from 'state/lists/wrappedTokenInfo'
+
+import EthereumLogo from '../../assets/images/ethereum-logo.png'
+import MaticLogo from '../../assets/svg/matic-token-icon.svg'
```
JFrankfurt, line 8: so glad we can use this again 😅

src/lib/hooks/useTokenList/querying.ts
```diff
@@ -10,13 +11,25 @@ import { tokenComparator, useSortTokensByQuery } from './sorting'
 export function useQueryTokens(query: string, tokens: WrappedTokenInfo[]) {
   const { account } = useActiveWeb3React()
   const balances = useTokenBalances(account, tokens)
-  const sortedTokens = useMemo(() => [...tokens.sort(tokenComparator.bind(null, balances))], [balances, tokens])
+  const sortedTokens = useMemo(
+    () => [
+      // spread to create a new memo, because sort is in-place and returns referentially equivalent
+      ...tokens.sort(tokenComparator.bind(null, balances)),
+    ],
```
JFrankfurt, line 18: nit: prefer `Array.from()`

JFrankfurt (approved):

src/lib/hooks/useCurrencyColor.ts
zzmp: `useLayoutEffect` is necessary to update the color in the same frame as the token is updated. Without it, the effect will only run in the frame _after_ the token is updated, and there will be one frame rendered with incorrect colors, which is noticeable.

src/lib/hooks/useCurrencyColor.ts
zzmp: I don't think the logos are complex/large enough images to need this - I think it already gets the color within a frame or two, and a webworker would not be able to beat that (because it already requires an additional frame or two to communicate with the webworker).

src/lib/hooks/useTokenList/querying.ts
zzmp: also, this only creates a new array _after_ sorting in place, which might muck up the data structure where it is generated. good call!

PR #3127, opened 2022-01-15
Update LICENSE

serafimcloud:


NoahZinsmeister:
the 2007 copyright is part of the [canonical GPLv3 license](https://www.gnu.org/licenses/gpl-3.0.en.html)

PR #3136, opened 2022-01-17
feat: swap input fields on token select

zzmp:
- Filters already selected token out of token selection
- On selecting the other field's token, swaps fields (token and value)

### DIFF
--- a/src/lib/components/TokenSelect/index.tsx
+++ b/src/lib/components/TokenSelect/index.tsx
@@ -2,7 +2,7 @@ import { t, Trans } from '@lingui/macro'
 import { Currency } from '@uniswap/sdk-core'
 import { useQueryTokenList } from 'lib/hooks/useTokenList'
 import styled, { ThemedText } from 'lib/theme'
-import { ElementRef, useCallback, useEffect, useRef, useState } from 'react'
+import { ElementRef, useCallback, useEffect, useMemo, useRef, useState } from 'react'
 
 import Column from '../Column'
 import Dialog, { Header } from '../Dialog'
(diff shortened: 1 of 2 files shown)


src/lib/components/TokenSelect/index.tsx
```diff
@@ -17,9 +17,15 @@ const SearchInput = styled(StringInput)`
...
+  onSelect: (token: Currency) => void
+}
+
+export function TokenSelectDialog({ value, onSelect }: TokenSelectDialogProps) {
   const [query, setQuery] = useState('')
-  const tokens = useQueryTokenList(query)
+  const queriedTokens = useQueryTokenList(query)
+  const tokens = useMemo(() => queriedTokens.filter((token) => token !== value), [queriedTokens, value])
 
   const baseTokens: Currency[] = [] // TODO(zzmp): Add base tokens to token list functionality
```
zzmp, line 31: Rm comment on L32: this is done in this PR.

src/lib/components/TokenSelect/index.tsx
```diff
@@ -48,9 +52,11 @@ export function TokenSelectDialog({ onSelect }: { onSelect: (token: Currency) =>
         </Row>
         {Boolean(baseTokens.length) && (
           <Row pad={0.75} gap={0.25} justify="flex-start" flex>
-            {baseTokens.map((token) => (
-              <TokenBase value={token} onClick={onSelect} key={token.wrapped.address} />
-            ))}
+            {baseTokens
+              .filter((token) => token !== value)
+              .map((token) => (
+                <TokenBase value={token} onClick={onSelect} key={token.wrapped.address} />
```
justindomingue, line 58: isn't there a possibility of conflict between native and wrapped native?

can't you use `currencyId(token)` instead as the key?

zzmp: yes and yes

Issue #2917, opened 2021-12-02
Liquidity distribution data on Arbitrum


willhennessy:
**Bug Description**
On the Add Liquidity page, the liquidity distribution does not appear on the Arbitrum network. 

graph returns empty array for tick data `{"data":{"ticks":[]}}`

`arbitrum-minimal` doesn’t have tick data because the graph cant index fast enough when we make `eth_calls` on arbitrum. As a result, we had to remove all `eth_calls` in the graph.

the unexplored solution would be to reimplement all the tick math in the graph itself - but unclear if info emitted in events is enough to back populate the contract tick state

willhennessy:
write a fallback function that grabs tick data on chain for that display using the the `TickLens` contract

Also, if we ever fail to load liquidity data let's change the copy from 
* current:  "There is no liquidity data"
* new:  "Failed to load liquidity data"

PR #3149, opened 2022-01-18
feat: use TickLens on chains where subgraph is not functional

justindomingue:
- uses `TickLens` on Arbitrum to get tick data (plugs into the existing hook as an alternative on arbitrum, reuses all existing active liquidity calculation utilities / graph as-is)

see #2917

arbitrum subgraph does not contain tick data. This PR relies on TickLens to display a liquidity distribution


src/state/data/slice.ts
```diff
@@ -26,7 +26,7 @@ export const api = createApi({
         document: gql`
           query allV3Ticks($poolAddress: String!, $skip: Int!) {
             ticks(first: 1000, skip: $skip, where: { poolAddress: $poolAddress }, orderBy: tickIdx) {
-              tickIdx
+              tick: tickIdx
```
justindomingue, line 29: this was just renamed from `tickIdx` to `tick` so it's more consisten

willhennessy:
looks great justin! There are still a few cases with no liquidity data - just curious if you know why?

Also either here or in a separate PR can we change the error copy? Is the following new version be accurate:
current: "There is no liquidity data"
new: "Failed to load liquidity data"

src/hooks/usePoolTickData.ts
```diff
@@ -1,50 +1,171 @@
...
+  const poolAddress =
+    currencyA && currencyB && feeAmount ? Pool.getAddress(currencyA?.wrapped, currencyB?.wrapped, feeAmount) : undefined
+
+  return useAllV3TicksQuery(poolAddress ? { poolAddress: poolAddress?.toLowerCase(), skip: 0 } : skipToken, {
+    pollingInterval: ms`30s`,
+  })
+}
+
+// Fetches all ticks for a given pool
+function useAllV3Ticks(
```
ianlapham, line 147: Why dont we just pass in the pool? or is it easier based on where we call this function

justindomingue: i think it's slightly simpler because the subgraph doesn't need a pool, just the address which is a static method on `Pool`.

justindomingue:
@willhennessy just merged this, will tal at the issue you raised and the copy.

PR #3160, opened 2022-01-20
feat: polish select

zzmp:
- Replaces test JSON-RPC urls with our dev infura urls
- Filters selected token from the token select
- Loads native currency with the other chain tokens (smoother loading)
- Switches currencies when selecting the other currency


src/lib/components/TokenSelect/index.tsx
```diff
@@ -17,14 +18,18 @@ const SearchInput = styled(StringInput)`
   ${inputCss}
 `
 
-export function TokenSelectDialog({ onSelect }: { onSelect: (token: Currency) => void }) {
+interface TokenSelectDialogProps {
+  value?: Currency
```
ianlapham, line 22: can this be called `currency`?

src/lib/components/TokenSelect/index.tsx
```diff
@@ -81,7 +86,7 @@ export default function TokenSelect({ value, collapsed, disabled, onSelect }: To
       <TokenButton value={value} collapsed={collapsed} disabled={disabled} onClick={() => setOpen(true)} />
```
ianlapham, line 86: same as above - or `selectedCurrency` or something

src/lib/hooks/swap/index.ts
```diff
@@ -45,3 +35,38 @@ export function useSwitchSwapCurrencies() {
...
+    },
+    [otherCurrency, setCurrency, switchCurrencies]
+  )
+  return [currency, setOrSwitchCurrency]
+}
+
+export function useSwapAmount(field: Field): [string | undefined, (amount: string) => void] {
+  const amount = useAtomValue(amountAtom)
+  const independentField = useAtomValue(independentFieldAtom)
+  const value = useMemo(() => (independentField === field ? amount : undefined), [amount, independentField, field])
```
ianlapham, line 61: Ahh i like having this reset to undefined (not sure if it actually did that before)

ianlapham (approved):
Just small naming nits

src/lib/components/TokenSelect/index.tsx
zzmp: I want to keep it as value to mirror native select elements, and make it obvious that this is the currently selected value.
I think all the tokens should be replaced with currency, but am waiting to do that until we've got some down time.

src/lib/components/TokenSelect/index.tsx
zzmp: see above - `value` is intentional

PR #3161, opened 2022-01-20
feat(widgets): use default input/output

JFrankfurt:
adds defaults and related prop validation

todo:
- [x] require that input and output addresses are included in the token list the integrator provided
- [x] support 'NATIVE' input address value

### DIFF
--- a/src/lib/components/Error/PropValidator.tsx
+++ b/src/lib/components/Error/PropValidator.tsx
@@ -45,7 +45,20 @@ export default function PropValidator(swapWidgetProps: WidgetProps) {
     }
   }, [convenienceFee, convenienceFeeRecipient])
 
-  const { defa } = swapWidgetProps
-  useEffect(() => {}, [])
+  const { defaultInputAddress, defaultInputAmount, defaultOutputAddress, defaultOutputAmount } = swapWidgetProps
+  useEffect(() => {
+    if (defaultInputAddress && !isAddress(defaultInputAddress)) {
+      throw new IntegrationError('defaultInputAddress must be a valid address.')
+    }
+    if (defaultInputAmount && defaultInputAmount < 0) {
+      throw new IntegrationError('defaultInputAmount must be a positive number.')
+    }
+    if (defaultOutputAddress && !isAddress(defaultOutputAddress)) {
+      throw new IntegrationError('defaultOutputAddress must be a valid address.')
+    }
+    if (defaultOutputAmount && defaultOutputAmount < 0) {
+      throw new IntegrationError('defaultOutputAmount must be a positive number.')
+    }
+  }, [defaultInputAddress, defaultInputAmount, defaultOutputAddress, defaultOutputAmount])
   return null
 }
--- a/src/lib/components/Error/PropValidator.tsx
+++ b/src/lib/components/Error/PropValidator.tsx
@@ -1,8 +1,20 @@
+import { BigNumber } from '@ethersproject/bignumber'
+import { DefaultAddress } from 'lib/components/Swap'
 import { WidgetProps } from 'lib/components/Widget'
 import { IntegrationError } from 'lib/errors'
 import { useEffect } from 'react'
 
 import { isAddress } from '../../../utils'
+
+function isAddressOrAddressMap(addressOrMap: DefaultAddress): boolean {
+  if (typeof addressOrMap === 'object') {
+    return Object.values(addressOrMap).every((address) => isAddress(address))
+  }
+  if (typeof addressOrMap === 'string') {
+    return typeof isAddress(addressOrMap) === 'string'
+  }
+  return false
+}
 
 export default function PropValidator(swapWidgetProps: WidgetProps) {
   const { jsonRpcEndpoint, provider } = swapWidgetProps
@@ -47,18 +59,19 @@ export default function PropValidator(swapWidgetProps: WidgetProps) {
 
   const { defaultInputAddress, defaultInputAmount, defaultOutputAddress, defaultOutputAmount } = swapWidgetProps
   useEffect(() => {
-    if (defaultInputAddress && !isAddress(defaultInputAddress)) {
-      throw new IntegrationError('defaultInputAddress must be a valid address.')
-    }
-    if (defaultInputAmount && defaultInputAmount < 0) {
+    if (defaultInputAmount && BigNumber.from(defaultInputAmount).lt(0)) {
       throw new IntegrationError('defaultInputAmount must be a positive number.')
     }
-    if (defaultOutputAddress && !isAddress(defaultOutputAddress)) {
-      throw new IntegrationError('defaultOutputAddress must be a valid address.')
-    }
-    if (defaultOutputAmount && defaultOutputAmount < 0) {
+    if (defaultOutputAmount && BigNumber.from(defaultOutputAmount).lt(0)) {
       throw new IntegrationError('defaultOutputAmount must be a positive number.')
     }
+    if (defaultInputAddress && !isAddressOrAddressMap(defaultInputAddress) && defaultInputAddress !== 'NATIVE') {
+      throw new IntegrationError('defaultInputAddress must be a valid address.')
+    }
+    if (defaultOutputAddress && !isAddressOrAddressMap(defaultOutputAddress)) {
+      throw new IntegrationError('defaultOutputAddress must be a valid address.')
+    }
   }, [defaultInputAddress, defaultInputAmount, defaultOutputAddress, defaultOutputAmount])
+
   return null
 }
--- a/src/lib/components/Error/PropValidator.tsx
+++ b/src/lib/components/Error/PropValidator.tsx
@@ -1,10 +1,10 @@
+import { WidgetProps } from 'lib/components/Widget'
 import { IntegrationError } from 'lib/errors'
-import { SwapWidgetProps } from 'lib/index'
 import { useEffect } from 'react'
 
 import { isAddress } from '../../../utils'
 
-export default function ErrorGenerator(swapWidgetProps: SwapWidgetProps) {
+export default function PropValidator(swapWidgetProps: WidgetProps) {
   const { jsonRpcEndpoint, provider } = swapWidgetProps
   useEffect(() => {
     if (!provider && !jsonRpcEndpoint) {
@@ -44,5 +44,8 @@ export default function ErrorGenerator(swapWidgetProps: SwapWidgetProps) {
       }
     }
   }, [convenienceFee, convenienceFeeRecipient])
+
+  const { defa } = swapWidgetProps
+  useEffect(() => {}, [])
   return null
 }
--- a/src/lib/components/Error/WidgetPropValidator.tsx
+++ b/src/lib/components/Error/WidgetPropValidator.tsx
@@ -1,0 +1,49 @@
+import { DefaultAddress } from 'lib/components/Swap'
+import { WidgetProps } from 'lib/components/Widget'
+import { IntegrationError } from 'lib/errors'
+import { PropsWithChildren, useEffect, useMemo, useState } from 'react'
+
+import { isAddress } from '../../../utils'
+
+function isAddressOrAddressMap(addressOrMap: DefaultAddress): boolean {
+  if (typeof addressOrMap === 'object') {
+    return Object.values(addressOrMap).every((address) => isAddress(address))
+  }
+  if (typeof addressOrMap === 'string') {
+    return typeof isAddress(addressOrMap) === 'string'
+  }
+  return false
+}
+
+export default function WidgetPropValidator(props: PropsWithChildren<WidgetProps>) {
+  const { jsonRpcEndpoint, provider } = props
+
+  const [providerChecked, setProviderChecked] = useState(false)
+  useEffect(() => {
+    setProviderChecked(false)
+    if (!provider && !jsonRpcEndpoint) {
+      throw new IntegrationError('This widget requires a provider or jsonRpcEndpoint.')
+    }
+    setProviderChecked(true)
+  }, [provider, jsonRpcEndpoint])
+
+  // size constraints
+  const [sizeConstraintsChecked, setSizeConstraintsChecked] = useState(false)
+  const { width } = props
+  useEffect(() => {
+    setSizeConstraintsChecked(false)
+    if (width && width < 300) {
+      throw new IntegrationError('Set widget width to at least 300px.')
+    }
+    setSizeConstraintsChecked(true)
+  }, [width])
+
+  const propsChecked = useMemo(
+    () => providerChecked && sizeConstraintsChecked,
+    [providerChecked, sizeConstraintsChecked]
+  )
+  if (propsChecked) {
+    return <>{props.children}</>
+  }
+  return null
+}
--- a/src/lib/components/Error/WidgetPropValidator.tsx
+++ b/src/lib/components/Error/WidgetPropValidator.tsx
@@ -1,19 +1,6 @@
-import { DefaultAddress } from 'lib/components/Swap'
 import { WidgetProps } from 'lib/components/Widget'
 import { IntegrationError } from 'lib/errors'
 import { PropsWithChildren, useEffect, useMemo, useState } from 'react'
-
-import { isAddress } from '../../../utils'
-
-function isAddressOrAddressMap(addressOrMap: DefaultAddress): boolean {
-  if (typeof addressOrMap === 'object') {
-    return Object.values(addressOrMap).every((address) => isAddress(address))
-  }
-  if (typeof addressOrMap === 'string') {
-    return typeof isAddress(addressOrMap) === 'string'
-  }
-  return false
-}
 
 export default function WidgetPropValidator(props: PropsWithChildren<WidgetProps>) {
   const { jsonRpcEndpoint, provider } = props
--- a/src/lib/components/Error/WidgetsPropsValidator.tsx
+++ b/src/lib/components/Error/WidgetsPropsValidator.tsx
@@ -1,36 +1,23 @@
 import { WidgetProps } from 'lib/components/Widget'
 import { IntegrationError } from 'lib/errors'
-import { PropsWithChildren, useEffect, useMemo, useState } from 'react'
+import { PropsWithChildren, useEffect } from 'react'
 
 export default function WidgetsPropsValidator(props: PropsWithChildren<WidgetProps>) {
   const { jsonRpcEndpoint, provider } = props
 
-  const [providerChecked, setProviderChecked] = useState(false)
   useEffect(() => {
-    setProviderChecked(false)
     if (!provider && !jsonRpcEndpoint) {
       throw new IntegrationError('This widget requires a provider or jsonRpcEndpoint.')
     }
-    setProviderChecked(true)
   }, [provider, jsonRpcEndpoint])
 
   // size constraints
-  const [sizeConstraintsChecked, setSizeConstraintsChecked] = useState(false)
   const { width } = props
   useEffect(() => {
-    setSizeConstraintsChecked(false)
     if (width && width < 300) {
       throw new IntegrationError('Set widget width to at least 300px.')
     }
-    setSizeConstraintsChecked(true)
   }, [width])
 
-  const propsChecked = useMemo(
-    () => providerChecked && sizeConstraintsChecked,
-    [providerChecked, sizeConstraintsChecked]
-  )
-  if (propsChecked) {
-    return <>{props.children}</>
-  }
-  return null
+  return <>{props.children}</>
 }
--- a/src/lib/components/Error/WidgetsPropsValidator.tsx
+++ b/src/lib/components/Error/WidgetsPropsValidator.tsx
@@ -11,11 +11,10 @@ export default function WidgetsPropsValidator(props: PropsWithChildren<WidgetPro
     }
   }, [provider, jsonRpcEndpoint])
 
-  // size constraints
   const { width } = props
   useEffect(() => {
     if (width && width < 300) {
-      throw new IntegrationError('Set widget width to at least 300px.')
+      throw new IntegrationError(`Set widget width to at least 300px. (You set it to ${width}.)`)
     }
   }, [width])
 
--- a/src/lib/components/Error/WidgetsPropsValidator.tsx
+++ b/src/lib/components/Error/WidgetsPropsValidator.tsx
@@ -2,7 +2,7 @@ import { WidgetProps } from 'lib/components/Widget'
 import { IntegrationError } from 'lib/errors'
 import { PropsWithChildren, useEffect, useMemo, useState } from 'react'
 
-export default function WidgetPropValidator(props: PropsWithChildren<WidgetProps>) {
+export default function WidgetsPropsValidator(props: PropsWithChildren<WidgetProps>) {
   const { jsonRpcEndpoint, provider } = props
 
   const [providerChecked, setProviderChecked] = useState(false)
--- a/src/lib/components/Swap/Swap.fixture.tsx
+++ b/src/lib/components/Swap/Swap.fixture.tsx
@@ -24,7 +24,7 @@ function Fixture() {
     }
   }, [color, setColor])
 
-  return <Swap defaults={{ tokenList: tokens }} />
+  return <Swap tokenList={tokens} />
 }
 
 export default <Fixture />
--- a/src/lib/components/Swap/Swap.fixture.tsx
+++ b/src/lib/components/Swap/Swap.fixture.tsx
@@ -1,4 +1,5 @@
 import { tokens } from '@uniswap/default-token-list'
+import { Token } from '@uniswap/sdk-core'
 import { DAI, USDC } from 'constants/tokens'
 import { useUpdateAtom } from 'jotai/utils'
 import { useEffect } from 'react'
@@ -16,6 +17,8 @@ const validateColor = (() => {
   }
 })()
 
+const UNSUPPORTED_TOKEN = new Token(1, '0xeb57bf569ad976974c1f861a5923a59f40222451', 18, 'LOOMI', 'Loomi')
+
 function Fixture() {
   const setColor = useUpdateAtom(colorAtom)
   const [color] = useValue('token color', { defaultValue: '' })
@@ -25,21 +28,21 @@ function Fixture() {
     }
   }, [color, setColor])
 
-  const addressOptions = ['', 'NATIVE', DAI.address, USDC.address]
+  const addressOptions = ['', 'NATIVE', DAI.address, USDC.address, UNSUPPORTED_TOKEN.address]
   const inputOptions = ['', '0', '-1', '100']
-  const [defaultInputAddress] = useSelect('default InputAddress', {
+  const [defaultInputAddress] = useSelect('defaultInputAddress', {
     options: addressOptions,
     defaultValue: addressOptions[2],
   })
-  const [defaultInputAmount] = useSelect('default InputAmount', {
+  const [defaultInputAmount] = useSelect('defaultInputAmount', {
     options: inputOptions,
     defaultValue: inputOptions[3],
   })
-  const [defaultOutputAddress] = useSelect('default OutputAddress', {
+  const [defaultOutputAddress] = useSelect('defaultOutputAddress', {
     options: addressOptions,
-    defaultValue: addressOptions[3],
+    defaultValue: addressOptions[1],
   })
-  const [defaultOutputAmount] = useSelect('default OutputAmount', {
+  const [defaultOutputAmount] = useSelect('defaultOutputAmount', {
     options: inputOptions,
     defaultValue: inputOptions[0],
   })
--- a/src/lib/components/Swap/Swap.fixture.tsx
+++ b/src/lib/components/Swap/Swap.fixture.tsx
@@ -36,7 +36,6 @@ function Fixture() {
     options: addressOptions,
     defaultValue: addressOptions[2],
   })
-
   const inputOptions = ['', '0', '100', '-1']
   const [defaultInputAmount] = useSelect('defaultInputAmount', {
     options: inputOptions,
--- a/src/lib/components/Swap/Swap.fixture.tsx
+++ b/src/lib/components/Swap/Swap.fixture.tsx
@@ -1,5 +1,4 @@
 import { tokens } from '@uniswap/default-token-list'
-import { Token } from '@uniswap/sdk-core'
 import { DAI, USDC } from 'constants/tokens'
 import { useUpdateAtom } from 'jotai/utils'
 import { useEffect } from 'react'
@@ -16,8 +15,6 @@ const validateColor = (() => {
     return validator.color !== ''
   }
 })()
-
-const UNSUPPORTED_TOKEN = new Token(1, '0xeb57bf569ad976974c1f861a5923a59f40222451', 18, 'LOOMI', 'Loomi')
 
 function Fixture() {
   const setColor = useUpdateAtom(colorAtom)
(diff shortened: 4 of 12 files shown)


src/lib/components/Error/PropValidator.tsx
```diff
@@ -44,5 +56,22 @@ export default function ErrorGenerator(swapWidgetProps: SwapWidgetProps) {
       }
     }
   }, [convenienceFee, convenienceFeeRecipient])
+
+  const { defaultInputAddress, defaultInputAmount, defaultOutputAddress, defaultOutputAmount } = swapWidgetProps
+  useEffect(() => {
+    if (defaultInputAmount && BigNumber.from(defaultInputAmount).lt(0)) {
```
zzmp, line 62: Is `lt` the same as `lessThan`? Use the more verbose method for readability.

zzmp, line 62: What happens if a string is passed instead of a number? Will `BigNumber.from` throw?

src/lib/components/Swap/index.tsx
```diff
@@ -18,54 +23,50 @@ import Settings from './Settings'
...
-  const output: DefaultTokenAmount = defaults.output || DEFAULT_OUTPUT
-  input.amount = input.amount || 0
-  output.amount = output.amount || 0
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  return useMemo(() => ({ tokenList, input, output }), [])
-}
-
+export type DefaultAddress = string | { [chainId: number]: string } | 'NATIVE'
 export interface SwapProps {
+  theme?: Theme
```
zzmp, line 28: A lot of these props are already specified as part of `WidgetProps`, not in `SwapProps`. `SwapProps` should only be those things that will be specific to swapping, otherwise you break the abstraction layer between widget/swap. I think the only props should be:
- default props
- convenience fee

src/lib/components/Swap/index.tsx
```diff
@@ -1,13 +1,18 @@
 import { Trans } from '@lingui/macro'
 import { TokenInfo } from '@uniswap/token-lists'
+import { SupportedLocale } from 'constants/locales'
 import { nativeOnChain } from 'constants/tokens'
 import { useSwapAmount, useSwapCurrency } from 'lib/hooks/swap'
 import { SwapInfoUpdater } from 'lib/hooks/swap/useSwapInfo'
 import useActiveWeb3React from 'lib/hooks/useActiveWeb3React'
-import useTokenList, { DEFAULT_TOKEN_LIST } from 'lib/hooks/useTokenList'
+import useSwapDefaults from 'lib/hooks/useSwapDefaults'
```
zzmp, line 8: This is swap specific, so it should come from `lib/hooks/swap/useSwapDefaults`

src/lib/components/Swap/index.tsx
```diff
@@ -18,54 +23,50 @@ import Settings from './Settings'
...
   const [, updateSwapOutputCurrency] = useSwapCurrency(Field.OUTPUT)
   const [, updateSwapInputAmount] = useSwapAmount(Field.INPUT)
+
   useLayoutEffect(() => {
-    if (chainId) {
-      updateSwapInputCurrency(nativeOnChain(chainId))
-      updateSwapOutputCurrency()
-      updateSwapInputAmount('')
+    if (chainId !== lastChainId) {
+      setLastChainId(chainId)
```
zzmp, line 62: This will reload defaults if the chain experiences a temporary disconnection, which is jarring to a user. Consider having the hook have only `chainId` as a dep, and not storing `lastChainId` in state (see `main` branch for an example).

src/lib/components/Swap/index.tsx
```diff
@@ -18,54 +23,50 @@ import Settings from './Settings'
...
+      if (chainId) {
+        updateSwapInputCurrency(nativeOnChain(chainId))
```
zzmp, line 64: These should be set to the defaults when chainId changes.

src/lib/components/Widget.tsx
```diff
@@ -77,19 +77,7 @@ function Updaters() {
...
-  jsonRpcEndpoint?: string
-  width?: string | number
-  dialog?: HTMLElement | null
-  className?: string
-  onError?: ErrorHandler
-} & (T extends JSXElementConstructor<any>
-  ? ComponentProps<T>
-  : // eslint-disable-next-line @typescript-eslint/ban-types
-    {})
+export type WidgetProps = SwapProps
```
zzmp, line 80: These props should stay in `Widget` to make it easier when we build add'l widget components.

src/lib/components/Widget.tsx
```diff
@@ -115,7 +104,7 @@ export default function Widget(props: PropsWithChildren<WidgetProps>) {
                   <AtomProvider>
                     <Web3Provider provider={provider} jsonRpcEndpoint={jsonRpcEndpoint}>
                       <Updaters />
-                      <ErrorGenerator {...props} />
+                      <PropValidator {...props} />
```
zzmp, line 107: Can propValidator be a hook instead of a component? It's not rendered. Is there an advantage to having a component instead of a hook?

src/lib/components/Swap/index.tsx
zzmp: Alternatively, should this logic be a part of `useSwapDefaults`?

src/lib/state/swap.ts
```diff
@@ -16,13 +18,19 @@ export interface Swap {
...
   amount: '',
   [Field.INPUT]: nativeOnChain(SupportedChainId.MAINNET),
 })
 
+type SwapDefaults = Pick<
+  SwapProps,
+  'defaultInputAddress' | 'defaultInputAmount' | 'defaultOutputAddress' | 'defaultOutputAmount'
+>
+
+export const swapDefaultsAtom = atom<SwapDefaults>({})
```
zzmp, line 32: Store these as local state on the `Swap` component, instead of in an atom.

src/lib/hooks/useTokenList/index.ts
```diff
@@ -40,25 +40,23 @@ export default function useTokenList(list?: string | TokenInfo[]): WrappedTokenI
...
-  return useMemo(() => {
-    return Object.values((chainId && chainTokenMap[chainId]) || {}).map(({ token }) => token)
-  }, [chainId, chainTokenMap])
 }
 
-export type TokenMap = { [address: string]: Token }
+export type TokenMap<T extends Token> = { [address: string]: T }
 
-export function useTokenMap(): TokenMap {
+export function useTokenMap(): TokenMap<WrappedTokenInfo> {
```
zzmp, line 47: Are you sure this doesn't break other typings? This also makes it less generic (because we're using our own interface instead of the one provided by the sdk), so I'd rather continue using `Token`.

src/lib/hooks/useSwapDefaults.ts
```diff
@@ -0,0 +1,57 @@
...
+  const outputAddress =
+    typeof defaultOutputAddress === 'string'
+      ? defaultOutputAddress
+      : defaultOutputAddress && chainId
+      ? defaultOutputAddress[chainId]
+      : undefined
+  const defaultOutputToken = useToken(outputAddress)
+
+  useEffect(() => {
+    // alter default values if they have changed
```
zzmp, line 39: I don't think defaults should be able to change - they should just be set on first render. wdyt?

src/lib/hooks/useSwapDefaults.ts
```diff
@@ -0,0 +1,57 @@
...
+    setDefaults({ defaultInputAddress, defaultInputAmount, defaultOutputAddress, defaultOutputAmount })
+    // if no swap fields have been set, set them to defaults
+    if (chainId && !inputCurrency && !outputCurrency && !inputAmount && !outputAmount) {
```
zzmp, line 42: These still need to be changed if the chainId is changed, even if they are set.
They also must be set with a `useLayoutEffect` to avoid a lagging frame rendered with old token values.

src/lib/components/Widget.tsx
JFrankfurt: Yeah, rendering it in the component tree here causes the errors it throws to be caught by the error boundary above.

src/lib/components/Widget.tsx
JFrankfurt: I was running into some weird/major ts drama trying to make it work w/ the prior setup. My plan here is to eventually make this something like `type WidgetProps = SwapProps | PriceProps | OtherWidgetProps`

src/lib/hooks/useSwapDefaults.ts
JFrankfurt: I figured someone's going to run into issues expecting it to work this way one day, and it's not hard to enable it, so why not. Not married to it tho.

src/lib/hooks/useTokenList/index.ts
JFrankfurt: This... does not look familiar to me. Not sure where this change came from. Might have been an editor thing? idk. reverting.

src/lib/components/Error/PropValidator.tsx
JFrankfurt: no, that's supported usage 👌

JFrankfurt: didn't know lessThan was a thing. cool.

src/lib/hooks/useTokenList/index.ts
JFrankfurt: ah this was in the cherry-picked commit, will take a closer look at these changes

src/lib/components/Error/WidgetPropValidator.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { PropsWithChildren, useEffect, useMemo, useState } from 'react'
+
+import { isAddress } from '../../../utils'
+
+function isAddressOrAddressMap(addressOrMap: DefaultAddress): boolean {
+  if (typeof addressOrMap === 'object') {
+    return Object.values(addressOrMap).every((address) => isAddress(address))
+  }
+  if (typeof addressOrMap === 'string') {
+    return typeof isAddress(addressOrMap) === 'string'
```
JFrankfurt, line 13: weird syntax, but `isAddress` returns an address or `false`

src/lib/index.tsx
```diff
@@ -1,9 +1,9 @@
-import Swap from './components/Swap'
+import Swap, { SwapProps } from './components/Swap'
 import Widget, { WidgetProps } from './components/Widget'
 
-export type SwapWidgetProps = WidgetProps<typeof Swap>
+type SwapWidgetProps = SwapProps & WidgetProps
```
zzmp, line 4: :+1: to changing `WidgetProps` to a simpler concrete type. Good call.

src/lib/components/Widget.tsx
```diff
@@ -1,17 +1,19 @@
-import { DEFAULT_LOCALE, SupportedLocale } from 'constants/locales'
+import { DEFAULT_LOCALE } from 'constants/locales'
```
zzmp, line 1: revert: all imports from `constants/locales` in one statement

src/lib/components/Widget.tsx
```diff
@@ -1,17 +1,19 @@
...
+import { SupportedLocale } from 'constants/locales'
 import { Provider as AtomProvider } from 'jotai'
 import { BlockUpdater } from 'lib/hooks/useBlockNumber'
 import { UNMOUNTING } from 'lib/hooks/useUnmount'
 import { Provider as I18nProvider } from 'lib/i18n'
 import { MulticallUpdater, store as multicallStore } from 'lib/state/multicall'
-import styled, { keyframes, Theme, ThemeProvider } from 'lib/theme'
-import { ComponentProps, JSXElementConstructor, PropsWithChildren, StrictMode, useRef } from 'react'
+import styled, { keyframes, ThemeProvider } from 'lib/theme'
```
zzmp, line 8: revert: all imports from `lib/theme` in one statement

src/lib/components/Error/WidgetPropValidator.tsx
```diff
@@ -0,0 +1,36 @@
+import { WidgetProps } from 'lib/components/Widget'
+import { IntegrationError } from 'lib/errors'
+import { PropsWithChildren, useEffect, useMemo, useState } from 'react'
+
+export default function WidgetPropValidator(props: PropsWithChildren<WidgetProps>) {
```
zzmp, line 5: nit: Rename to `WidgetsPropsValidator` (rename the file too)

src/lib/components/Error/WidgetPropValidator.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const { jsonRpcEndpoint, provider } = props
+
+  const [providerChecked, setProviderChecked] = useState(false)
```
zzmp, line 8: - Why can't this be checked every time?
- Instead of introducing state, you can set the effect to run only on mount by not passing a deps array (or passing an empty array). If the linter stops you, disable the check and leave a comment explaining the behavior.

This applies to all checks. I don't think we need ot introduce local state to only check them once. What if the props change, and future values are invalid?

src/lib/components/Error/WidgetPropValidator.tsx
```diff
@@ -0,0 +1,36 @@
...
+      throw new IntegrationError('Set widget width to at least 300px.')
+    }
+    setSizeConstraintsChecked(true)
+  }, [width])
+
+  const propsChecked = useMemo(
+    () => providerChecked && sizeConstraintsChecked,
+    [providerChecked, sizeConstraintsChecked]
+  )
+  if (propsChecked) {
```
zzmp, line 32: To reduce nesting, have this return `null` instead, and make it an adjacent DOM node

src/lib/components/Swap/Swap.fixture.tsx
```diff
@@ -24,7 +25,34 @@ function Fixture() {
     }
   }, [color, setColor])
 
-  return <Swap defaults={{ tokenList: tokens }} />
+  const addressOptions = ['', 'NATIVE', DAI.address, USDC.address]
```
zzmp, line 28: I don't think these defaults need to be a part of cosmos. Can they be removed, and instead included in the demo site that we produce for integrators? wdyt?

src/lib/components/Widget.tsx
```diff
@@ -111,15 +111,16 @@ export default function Widget(props: PropsWithChildren<WidgetProps>) {
...
-                <ReduxProvider store={multicallStore}>
-                  <AtomProvider>
-                    <Web3Provider provider={provider} jsonRpcEndpoint={jsonRpcEndpoint}>
-                      <Updaters />
-                      <ErrorGenerator {...props} />
-                      {children}
-                    </Web3Provider>
-                  </AtomProvider>
-                </ReduxProvider>
+                <WidgetPropValidator {...props}>
```
zzmp, line 114: Can this be made a sibling DOM node, like the `<Updaters />`, to reduce nesting?

src/lib/components/Swap/SwapPropValidator.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export default function SwapPropValidator(props: PropsWithChildren<SwapProps>) {
+  // convenience fee constraints
+  const [convenienceFeeChecked, setConvenienceFeeChecked] = useState(false)
+  const { convenienceFee, convenienceFeeRecipient } = props
+  useEffect(() => {
+    setConvenienceFeeChecked(false)
+    if (convenienceFee) {
+      if (convenienceFee > 100 || convenienceFee < 0) {
+        throw new IntegrationError('Set widget convenienceFee to at least 400px.')
```
zzmp, line 26: copypasta

src/lib/components/Swap/SwapPropValidator.tsx
```diff
@@ -0,0 +1,78 @@
...
+      }
+      if (!convenienceFeeRecipient) {
+        throw new IntegrationError('convenienceFeeRecipient is required when convenienceFee is set.')
+      }
+      const MustBeValidAddressError = new IntegrationError('convenienceFeeRecipient must be a valid address.')
```
zzmp, line 31: Instead of reusing the error, can you include more context in the two error cases' messages?

src/lib/components/Swap/SwapPropValidator.tsx
```diff
@@ -0,0 +1,78 @@
...
+      throw new IntegrationError('defaultOutputAddress must be a valid address.')
+    }
+    setDefaultsChecked(true)
+  }, [defaultInputAddress, defaultInputAmount, defaultOutputAddress, defaultOutputAmount])
+
+  const propsChecked = useMemo(
+    () => convenienceFeeChecked && defaultsChecked && defaultsChecked,
+    [convenienceFeeChecked, defaultsChecked]
+  )
+  if (propsChecked) {
```
zzmp, line 74: Return null and include this as an empty sibling DOM node.

src/lib/components/Swap/index.tsx
```diff
@@ -16,60 +14,31 @@ import Output from './Output'
...
-  const tokenList = defaults.tokenList || DEFAULT_TOKEN_LIST
-  const input: DefaultTokenAmount = defaults.input || DEFAULT_INPUT
-  const output: DefaultTokenAmount = defaults.output || DEFAULT_OUTPUT
-  input.amount = input.amount || 0
-  output.amount = output.amount || 0
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  return useMemo(() => ({ tokenList, input, output }), [])
-}
-
+export type DefaultAddress = string | { [chainId: number]: string } | 'NATIVE'
```
zzmp, line 20: string | { [chainId: number]: string | 'NATIVE' } | 'NATIVE'
Write Preview

src/lib/hooks/swap/useSwapDefaults.ts
```diff
@@ -0,0 +1,68 @@
...
+  defaultOutputAmount: string | undefined
+) {
+  const [, updateSwapInputCurrency] = useSwapCurrency(Field.INPUT)
+  const [, updateSwapOutputCurrency] = useSwapCurrency(Field.OUTPUT)
+  const [, updateSwapInputAmount] = useSwapAmount(Field.INPUT)
+  const [, updateSwapOutputAmount] = useSwapAmount(Field.OUTPUT)
+  const { chainId } = useActiveWeb3React()
+
+  const inputAddress =
+    typeof defaultInputAddress === 'string'
```
zzmp, line 21: All of this logic should be moved to a helper function in the same file, something like `useDefaultToken(address: DefaultAddress): ReturnType<typeof useToken>`.

src/lib/hooks/swap/useSwapDefaults.ts
```diff
@@ -0,0 +1,68 @@
...
+    defaultInputAmount,
+    defaultOutputAmount,
+    updateSwapInputCurrency,
+    updateSwapOutputCurrency,
+    updateSwapInputAmount,
+    updateSwapOutputAmount,
+  ])
+
+  const [previousChainId, setPreviousChainId] = useState(chainId)
+  useEffect(() => {
```
zzmp, line 60: This *must* be `useLayoutEffect` in order to update the tokens before rendering with a new chain.

zzmp (changes requested):
Note that checking for input/output addresses inclusion in the token list can only be done async.

src/lib/components/Error/PropValidator.tsx
zzmp: but if `defaultInputAmount` is invalid (eg `abc`) it will throw. I think this is caught by the ErrorBoundary, though.

src/lib/components/Widget.tsx
zzmp: This makes sense for WidgetPropsValidator, but couldn't SwapPropsValidator still be a hook?

src/lib/components/Error/WidgetPropValidator.tsx
zzmp: instead of using a typeof check, return `Boolean(isAddress(addressOrMap)`. More readable imo

src/lib/components/Error/WidgetPropValidator.tsx
JFrankfurt: The nesting is intentional. We don't want to render dom that's going to use these props if they are invalid.

src/lib/components/Error/WidgetPropValidator.tsx
JFrankfurt: These checks are run every time. The state does not prevent re-running the hook. It exists to keep the child dom tree from re-rendering with a potentially invalid config.

src/lib/components/Swap/Swap.fixture.tsx
JFrankfurt: They're useless in cosmos at runtime, actually, but included for completeness. Is there a way to do object mapping for select options in cosmos? I also wanted something cleaner than addresses, but didn't see anything when I skimmed the docs.

src/lib/components/Widget.tsx
JFrankfurt: I don't think so, since the child tree is intentionally not rendered until props have been validated.

src/lib/components/Error/WidgetPropValidator.tsx
zzmp: They will still not render, because the error should be thrown first, causing the `ErrorBoundary` to render instead.

src/lib/components/Swap/index.tsx
```diff
@@ -16,60 +14,31 @@ import Output from './Output'
...
   convenienceFee?: number
-  convenienceFeeRecipient?: string // TODO: improve typing to require recipient when fee is set
-  defaults?: Partial<SwapDefaults>
+  convenienceFeeRecipient?: string | { [chainId: number]: string }
 }
 
-export default function Swap({ defaults }: SwapProps) {
-  const { tokenList } = useSwapDefaults(defaults)
+export default function Swap(props: SwapProps) {
+  const { defaultInputAddress, defaultInputAmount, defaultOutputAddress, defaultOutputAmount, tokenList } = props
```
zzmp, line 32: If `useSwapDefaults` uses an object, this destructure can be omitted. `useTokenList` can be passed `props.tokenList`.

src/lib/hooks/swap/useSwapDefaults.ts
zzmp: ***Both*** of these `useEffects` ***must*** be `useLayoutEffect`

src/lib/hooks/swap/useSwapDefaults.ts
```diff
@@ -0,0 +1,63 @@
...
+import { DefaultAddress } from 'lib/components/Swap'
+import useActiveWeb3React from 'lib/hooks/useActiveWeb3React'
+import { useToken } from 'lib/hooks/useCurrency'
+import { Field, Swap, swapAtom } from 'lib/state/swap'
+import { useCallback, useEffect, useState } from 'react'
+
+function useDefaultToken(
+  defaultAddress: DefaultAddress | undefined,
+  chainId: number | undefined
+): ReturnType<typeof useToken> | NativeCurrency {
```
zzmp, line 13: I think this return type is already aliased to `Currency`, imported from `@uniswap/sdk-core`. Use that instead.

src/lib/hooks/swap/useSwapDefaults.ts
```diff
@@ -0,0 +1,63 @@
...
+    newSwap[Field.OUTPUT] = defaultOutputToken || undefined
+    // independentField is Field.INPUT unless defaultOutputAmount is set
+    newSwap.independentField = Field.INPUT
+    if (defaultInputAmount && defaultInputToken) {
+      newSwap.amount = defaultInputAmount
+    } else if (defaultOutputAmount && defaultOutputToken) {
+      newSwap.independentField = Field.OUTPUT
+      newSwap.amount = defaultOutputAmount
+    }
+    updateSwap((swap) => ({ ...swap, ...newSwap }))
```
zzmp, line 50: You should be able to set `updateSwap(newSwap)`. `integratorFee` belongs in options.
I think you should use `updateSwap(newSwap)` and make a note next to `integratorFee`'s definition to move it to options.

src/lib/hooks/swap/useSwapDefaults.ts
```diff
@@ -0,0 +1,63 @@
...
+  defaultOutputAddress: DefaultAddress | undefined,
+  defaultOutputAmount: string | undefined
+) {
+  const updateSwap = useUpdateAtom(swapAtom)
+  const { chainId } = useActiveWeb3React()
+  const defaultInputToken = useDefaultToken(defaultInputAddress, chainId)
+  const defaultOutputToken = useDefaultToken(defaultOutputAddress, chainId)
+
+  const setToDefaults = useCallback(() => {
+    const newSwap: Partial<Swap> = {}
```
zzmp, line 39: nit: name this `defaultSwap`.

src/lib/hooks/swap/useSwapDefaults.ts
```diff
@@ -0,0 +1,63 @@
...
+    newSwap[Field.INPUT] = defaultInputToken || undefined
+    newSwap[Field.OUTPUT] = defaultOutputToken || undefined
+    // independentField is Field.INPUT unless defaultOutputAmount is set
+    newSwap.independentField = Field.INPUT
```
zzmp, line 43: similarly, `newSwap.amount = ''` unless otherwise set by defaults.

zzmp (changes requested):
I still think the validators can return null and be adjacent nodes, instead of parent nodes, to reduce nesting.
You pointed out that having them be parent nodes prevents the children from being rendered in the case of an error, but I think throwing a synchronous error should have that same effect, effectively rendering the ErrorBoundary instead of the children.

Separately, error messages should include the offending value if they can. eg instead of "convenienceFee must be between 0 and 100", it should be "convenienceFee  (${convenienceFee}) must be between 0 and 100" or something like that.

src/lib/state/swap.ts
```diff
@@ -11,9 +11,9 @@ export enum Field {
 
 export interface Swap {
   independentField: Field
-  readonly amount: string
-  readonly [Field.INPUT]?: Currency
-  readonly [Field.OUTPUT]?: Currency
+  amount: string
```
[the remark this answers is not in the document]
JFrankfurt: because if you're constructing a new `Swap` entity you can't have `readonly` fields

src/lib/components/Swap/Swap.fixture.tsx
```diff
@@ -24,7 +28,34 @@ function Fixture() {
     }
   }, [color, setColor])
 
-  return <Swap defaults={{ tokenList: tokens }} />
+  const addressOptions = ['', 'NATIVE', DAI.address, USDC.address, UNSUPPORTED_TOKEN.address]
```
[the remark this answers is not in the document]
JFrankfurt: This doesn't break when changing chains because the widget clears unsupported inputs if the user is on a chain the integrator didn't intend.

src/lib/index.tsx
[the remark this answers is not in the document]
JFrankfurt: I prefer not exporting things that aren't consumed outside the file if they're not used anywhere else.

src/lib/components/Swap/index.tsx
```diff
@@ -16,60 +14,31 @@ import Output from './Output'
...
-  const input: DefaultTokenAmount = defaults.input || DEFAULT_INPUT
-  const output: DefaultTokenAmount = defaults.output || DEFAULT_OUTPUT
-  input.amount = input.amount || 0
-  output.amount = output.amount || 0
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  return useMemo(() => ({ tokenList, input, output }), [])
-}
-
+export type DefaultAddress = string | { [chainId: number]: string | 'NATIVE' } | 'NATIVE'
 export interface SwapProps {
```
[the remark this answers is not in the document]
JFrankfurt: I tried to pull this off for quite a while but was unable to. The type above won't work, unfortunately.

src/lib/components/Error/WidgetsPropsValidator.tsx
```diff
@@ -0,0 +1,23 @@
...
+export default function WidgetsPropsValidator(props: PropsWithChildren<WidgetProps>) {
+  const { jsonRpcEndpoint, provider } = props
+
+  useEffect(() => {
+    if (!provider && !jsonRpcEndpoint) {
+      throw new IntegrationError('This widget requires a provider or jsonRpcEndpoint.')
+    }
+  }, [provider, jsonRpcEndpoint])
+
+  // size constraints
```
zzmp, line 14: nit: I don't think this comment is necessary because it's labeled by the error message. Same goes for the comments in the Swap validator.

src/lib/components/Error/WidgetsPropsValidator.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const { width } = props
+  useEffect(() => {
+    if (width && width < 300) {
+      throw new IntegrationError('Set widget width to at least 300px.')
```
zzmp, line 18: nit: `Set widget width to at least 300px (was ${width}).`

src/lib/components/Swap/Swap.fixture.tsx
```diff
@@ -24,7 +25,41 @@ function Fixture() {
...
+    Native: 'NATIVE',
+    DAI: DAI.address,
+    USDC: USDC.address,
+  }
+  const addressOptions = Object.keys(optionsToAddressMap)
+  const [defaultInput] = useSelect('defaultInputAddress', {
+    options: addressOptions,
+    defaultValue: addressOptions[2],
+  })
+
```
zzmp, line 39: nit: rm newline

src/lib/components/Swap/SwapPropValidator.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+type ValidatorProps = PropsWithChildren<SwapProps>
+
+export default function SwapPropValidator(props: ValidatorProps) {
+  // convenience fee constraints
+  const { convenienceFee, convenienceFeeRecipient } = props
+  useEffect(() => {
+    if (convenienceFee) {
+      if (convenienceFee > 100 || convenienceFee < 0) {
+        throw new IntegrationError('convenienceFee must be between 0 and 100')
```
zzmp, line 26: nit: append `(was ${convenienceFee}).` (don't forget the period for consistency)
do this in all IntegrationErrors for this validator.

src/lib/components/Swap/index.tsx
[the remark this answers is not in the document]
zzmp: Any idea why it won't work? Could you add a TODO to investigate this when we have more time?

src/lib/components/Swap/index.tsx
```diff
@@ -16,60 +14,34 @@ import Output from './Output'
...
-  defaults?: Partial<SwapDefaults>
+  convenienceFeeRecipient?: string | { [chainId: number]: string }
 }
 
-export default function Swap({ defaults }: SwapProps) {
-  const { tokenList } = useSwapDefaults(defaults)
+export default function Swap(props: SwapProps) {
+  const { defaultInputAddress, defaultInputAmount, defaultOutputAddress, defaultOutputAmount, tokenList } = props
   useTokenList(tokenList)
+  const useSwapDefaultsArgs = useMemo(
```
zzmp, line 34: I was wrong to tell you to memoize here. You should not destructure props, and just pass `props` directly to `useSwapDefaults`. `useSwapDefaults` destructures before adding anything to a deps array, so you shouldn't have any perf issues.

src/lib/hooks/swap/useSwapDefaults.ts
```diff
@@ -0,0 +1,71 @@
...
+      defaultSwapState.amount = defaultOutputAmount
+    }
+    updateSwap((swap) => ({ ...swap, ...defaultSwapState }))
+  }, [defaultInputToken, defaultOutputToken, defaultInputAmount, defaultOutputAmount, updateSwap])
+
+  const [previousChainId, setPreviousChainId] = useState(chainId)
+  useLayoutEffect(() => {
+    setPreviousChainId(chainId)
+  }, [chainId])
+
```
zzmp, line 65: nit: rm newline, as the logic between these layouteffects work in tandem.

zzmp (approved):
Pass `props` directly through to `useSwapDefaults`.
There are some other nits, but once that is fixed lgtm.

PR #3185, opened 2022-01-25
refactor: track txs

zzmp:
- Guts the Status fixture to avoid overcomplicating this refactor
- Refactors the transactions state updater to be generic; wraps it for app and lib
- Update approvals to be tracked from SwapButton


src/lib/state/transactions.ts
```diff
@@ -0,0 +1,53 @@
+import { TransactionReceipt, TransactionResponse } from '@ethersproject/abstract-provider'
```
zzmp, line 1: NB: I would consider it valid feedback to tell me to re-use the existing data structures from app. This feels simplified to me, but is definitely a duplication of work.

src/lib/components/Swap/SwapButton.tsx
```diff
@@ -26,17 +28,24 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
     setActiveTrade((activeTrade) => activeTrade && trade.trade)
   }, [trade])
 
-  // TODO(zzmp): Track pending approval
-  const useIsPendingApproval = () => false
-
   // TODO(zzmp): Return an optimized trade directly from useSwapInfo.
   const optimizedTrade = useSwapApprovalOptimizedTrade(trade.trade, allowedSlippage, useIsPendingApproval)
-  const [approval, getApproval] = useSwapApproval(optimizedTrade, allowedSlippage, useIsPendingApproval)
+  const [approval, getApproval] = useSwapApproval(optimizedTrade || trade.trade, allowedSlippage, useIsPendingApproval)
```
ianlapham, line 33: why do we pass in both trades here?

src/lib/hooks/transactions/updater.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { SupportedChainId } from 'constants/chains'
+import useActiveWeb3React from 'hooks/useActiveWeb3React'
+import useBlockNumber, { useFastForwardBlockNumber } from 'lib/hooks/useBlockNumber'
+import { useCallback, useEffect } from 'react'
+import { retry, RetryableError, RetryOptions } from 'utils/retry'
+
+interface Transaction {
+  addedTime: number
+  receipt?: unknown
+  lastCheckedBlockNumber?: number
```
ianlapham, line 11: do we want to add additional fields to match what we have in app? It gives us access to things like amounts and currencies involved (we'd only need this if we want to show advanced details about the txn anywhere else in app)

ianlapham (commented):
Some minor questions

src/lib/hooks/transactions/updater.tsx
zzmp: yes, absolutely. it just isn't necessary for approval txn's. I expected you to augment the `SwapTransactionInfo` interface in future work to include those, as you worked through swap submission.

src/lib/components/Swap/SwapButton.tsx
zzmp: Added a comment to clarify usage.

src/lib/components/Swap/SwapButton.tsx
```diff
@@ -26,17 +28,26 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
...
   }, [trade])
 
-  // TODO(zzmp): Track pending approval
-  const useIsPendingApproval = () => false
-
   // TODO(zzmp): Return an optimized trade directly from useSwapInfo.
-  const optimizedTrade = useSwapApprovalOptimizedTrade(trade.trade, allowedSlippage, useIsPendingApproval)
+  const optimizedTrade =
+    // Use trade.trade if there is no swap optimized trade. This occurs if approvals are still pending.
+    useSwapApprovalOptimizedTrade(trade.trade, allowedSlippage, useIsPendingApproval) || trade.trade
```
ianlapham, line 34: Couldnt this cause a case where the user submits the non optimized trade? aka shouldnt we always wait for approvals? Or is this out of scope

src/lib/hooks/transactions/updater.tsx
ianlapham: Sure makes sense

ianlapham (commented):
Just one question on trade usage

src/lib/components/Swap/SwapButton.tsx
zzmp: trade.trade will still return the correct approvals from `useSwapApproval`, so this will cause the `approval` to be `ApprovalState.PENDING`, which is what we want. The follow-up PR will amend the UI to "wait" for approval so that a user does not submit a non-optimized trade.

ianlapham: ahh i see, makes sense

ianlapham (approved):

PR #3186, opened 2022-01-25
feat: pending approval ui

zzmp:
Note that transaction tracking and swap hooks are kept distinct. This is intentional.


ianlapham (commented):
Looks good

PR #3189, opened 2022-01-26
feat: pending tx state

zzmp:


src/lib/state/swap.ts
```diff
@@ -24,5 +24,6 @@ export const swapAtom = atomWithImmer<Swap>({
 })
 
 export const independentFieldAtom = pickAtom(swapAtom, 'independentField')
-export const integratorFeeAtom = pickAtom(swapAtom, 'integratorFee')
-export const amountAtom = pickAtom(swapAtom, 'amount')
+
+// If set to a transaction hash, displays that transaction's status.
+export const pendingTxHashAtom = atom<string | undefined>(undefined)
```
zzmp, line 29: Should this be something more clear? Maybe `txHashToDisplay`?

ianlapham: yeah its a little confusing to me because multiple can be pending - do we want to always show the oldest txn or the most recent - could be newestTxHash

ianlapham (approved):
Looks good - one comment on naming but okay with whatever you decide

PR #3193, opened 2022-01-27
feat: Make pending txn status functional 

ianlapham:


src/lib/components/Swap/SwapButton.tsx
```diff
@@ -129,18 +131,46 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
...
+
   const onConfirm = useCallback(() => {
     swapCallback?.()
-      .then((transactionResponse) => {
-        // TODO(ianlapham): Add the swap tx to transactionsAtom
-        console.log(transactionResponse)
-        setDisplayTxHash(transactionResponse.hash)
+      .then((response) => {
+        setDisplayTxHash(response.hash)
+        if (inputCurrency && outputCurrency && inputCurrencyAmount && outputCurrencyAmount) {
```
zzmp, line 139: Will this always be true? If so, use 'invariant' from 'tiny-invariant' instead of an if statement.

zzmp, line 139: You never actually use 'inputCurrency' and 'outputCurrency' here, you should omit them and not destructure 'outputCurrency'.

src/lib/state/transactions.ts
```diff
@@ -21,20 +21,20 @@ export interface ApprovalTransactionInfo extends BaseTransactionInfo {
...
-  inputCurrencyAddress: string
-  outputCurrencyAddress: string
+  inputCurrencyAmount: CurrencyAmount<Currency>
+  outputCurrencyAmount: CurrencyAmount<Currency>
 }
 
 export interface InputSwapTransactionInfo extends SwapTransactionInfo {
   tradeType: TradeType.EXACT_INPUT
-  inputCurrencyAmount: string
+  inputCurrencyAmount: CurrencyAmount<Currency>
```
zzmp, line 30: Duplicate from base interface - omit

src/lib/state/transactions.ts
```diff
@@ -21,20 +21,20 @@ export interface ApprovalTransactionInfo extends BaseTransactionInfo {
...
   expectedOutputCurrencyAmount: string
   minimumOutputCurrencyAmount: string
 }
 
 export interface OutputSwapTransactionInfo extends SwapTransactionInfo {
   tradeType: TradeType.EXACT_OUTPUT
-  outputCurrencyAmount: string
+  outputCurrencyAmount: CurrencyAmount<Currency>
```
zzmp, line 37: Duplicate from bad interface - omit

zzmp (approved):

src/lib/components/Swap/SwapButton.tsx
ianlapham: ah was wondering why that was used - yeah it always will be

src/lib/components/Swap/SwapButton.tsx
```diff
@@ -129,18 +132,36 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
...
+          })
+        }
+        if (independentField === Field.OUTPUT) {
+          addTransaction({
+            response,
+            type: TransactionType.SWAP,
+            tradeType: TradeType.EXACT_OUTPUT,
+            inputCurrencyAmount,
+            outputCurrencyAmount,
+          })
```
JFrankfurt, line 157: can you just do a conditional on the single field that's different here to cut down on repetition?

src/lib/state/transactions.ts
```diff
@@ -21,20 +21,18 @@ export interface ApprovalTransactionInfo extends BaseTransactionInfo {
 export interface SwapTransactionInfo extends BaseTransactionInfo {
   type: TransactionType.SWAP
   tradeType: TradeType
-  inputCurrencyAddress: string
-  outputCurrencyAddress: string
+  inputCurrencyAmount: CurrencyAmount<Currency>
+  outputCurrencyAmount: CurrencyAmount<Currency>
 }
```
JFrankfurt, line 26: confirming we don't need addresses here?

JFrankfurt (approved):

src/lib/components/Swap/SwapButton.tsx
ianlapham: i thought of this - i think this pattern works better if we expect more fields to be added in future, aka a ternary may not resolve correctly if there are 3 fields

but, i dont see us addign a third field to swap unless some weir changes to protoocl or router are added, so ill go with your suggestion

src/lib/state/transactions.ts
ianlapham: yep the currency amount can give us token which can give us address

PR #3194, opened 2022-01-27
feat(widgets): move uniswap protocol branding from header to bottom of output

JFrankfurt:
image

### DIFF
--- a/src/lib/components/BrandingFooter.tsx
+++ b/src/lib/components/BrandingFooter.tsx
@@ -1,0 +1,40 @@
+import { Trans } from '@lingui/macro'
+import Row from 'lib/components/Row'
+import { Logo } from 'lib/icons'
+import styled from 'lib/theme'
+
+import ExternalLink from './ExternalLink'
+
+const UniswapA = styled(ExternalLink)`
+  color: ${({ theme }) => theme.secondary};
+  cursor: pointer;
+  display: flex;
+  font-size: 0.75em;
+  justify-content: center;
+  line-height: 1.35em;
+  text-decoration: none;
+
+  ${Logo} {
+    fill: ${({ theme }) => theme.secondary};
+    height: 1em;
+    transition: transform 0.25s ease;
+    width: 1em;
+    will-change: transform;
+
+    :hover {
+      fill: ${({ theme }) => theme.onHover(theme.secondary)};
+      transform: rotate(-5deg);
+    }
+  }
+`
+
+export default function BrandingFooter() {
+  return (
+    <UniswapA href={`https://app.uniswap.org/`}>
+      <Row gap={0.4}>
+        <Logo />
+        <Trans>Powered by the Uniswap protocol</Trans>
+      </Row>
+    </UniswapA>
+  )
+}
--- a/src/lib/components/ExternalLink.tsx
+++ b/src/lib/components/ExternalLink.tsx
@@ -1,0 +1,17 @@
+import { HTMLProps } from 'react'
+
+/**
+ * Outbound link
+ */
+export default function ExternalLink({
+  target = '_blank',
+  href,
+  rel = 'noopener noreferrer',
+  ...rest
+}: Omit<HTMLProps<HTMLAnchorElement>, 'as' | 'ref' | 'onClick'> & { href: string }) {
+  return (
+    <a target={target} rel={rel} href={href} {...rest}>
+      {rest.children}
+    </a>
+  )
+}
--- a/src/lib/components/Header.tsx
+++ b/src/lib/components/Header.tsx
@@ -1,25 +1,8 @@
-import { largeIconCss, Logo } from 'lib/icons'
+import { largeIconCss } from 'lib/icons'
 import styled, { ThemedText } from 'lib/theme'
 import { ReactElement, ReactNode } from 'react'
 
 import Row from './Row'
-
-const UniswapA = styled.a`
-  cursor: pointer;
-
-  ${Logo} {
-    fill: ${({ theme }) => theme.secondary};
-    height: 1.5em;
-    transition: transform 0.25s ease;
-    width: 1.5em;
-    will-change: transform;
-
-    :hover {
-      fill: ${({ theme }) => theme.onHover(theme.secondary)};
-      transform: rotate(-5deg);
-    }
-  }
-`
 
 const HeaderRow = styled(Row)`
   height: 1.75em;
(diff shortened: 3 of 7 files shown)


src/lib/components/Widget.tsx
```diff
@@ -39,7 +39,6 @@ const WidgetWrapper = styled.div<{ width?: number | string }>`
   font-size: 16px;
   font-smooth: always;
   font-variant: none;
-  height: 348px;
```
zzmp, line 42: This should not be removed - it constrains the height in cases where it would otherwise overflow, like with token lists

src/lib/components/Swap/index.tsx
```diff
@@ -59,7 +59,7 @@ export default function Swap(props: SwapProps) {
   return (
     <SwapPropValidator {...props}>
       <SwapInfoUpdater />
-      <Header logo title={<Trans>Swap</Trans>}>
+      <Header title={<Trans>Swap</Trans>}>
```
zzmp, line 62: Instead of removing the `logo` and including it in the `Output`, make a new `Footer` component that you can put _after_ `Output`. The footer has nothing to do with the output, so it should be a separate component.

src/lib/components/Swap/Output.tsx
```diff
@@ -101,6 +126,11 @@ export default function Output({ disabled, children }: OutputProps) {
           </ThemedText.Body2>
         </TokenInput>
         {children}
+        <UniswapA href={`https://app.uniswap.org/`}>
+          <Logo />
+          &nbsp;
```
zzmp, line 131: Instead of a non-blocking space, use a `Row` with a `gap` so you have obvious control over the spacing between the logo and the text:

```
`<Row gap={0.5}><Logo /><Trans>Powered by...</Trans></Row>
```

src/lib/components/Swap/Output.tsx
```diff
@@ -31,6 +32,30 @@ const OutputColumn = styled(Column)<{ hasColor: boolean | null }>`
   }
 `
 
+const UniswapA = styled.a`
+  align-items: center;
```
zzmp, line 36: I don't think these addition properties (like `align-items`) are necessary if you use a `Row`.

zzmp (changes requested):
Move the footer to its own `Footer` component, and continue to use a `Row` to format it.

src/lib/components/BrandingFooter.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { Logo } from 'lib/icons'
+import styled from 'lib/theme'
+
+import ExternalLink from './ExternalLink'
+
+const UniswapA = styled(ExternalLink)`
+  color: ${({ theme }) => theme.secondary};
+  cursor: pointer;
+  display: flex;
+  font-size: 0.75em;
```
zzmp, line 12: Font size and line height should come from a ThemedText element and not specified explicitly in order to keep styles consistent and semantically named.

src/lib/components/BrandingFooter.tsx
```diff
@@ -0,0 +1,40 @@
...
+import Row from 'lib/components/Row'
+import { Logo } from 'lib/icons'
+import styled from 'lib/theme'
+
+import ExternalLink from './ExternalLink'
+
+const UniswapA = styled(ExternalLink)`
+  color: ${({ theme }) => theme.secondary};
+  cursor: pointer;
+  display: flex;
```
zzmp, line 11: Display and justify-content should be enabled through props in the Row, not specified explicitly.

src/lib/components/ExternalLink.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Outbound link
+ */
+export default function ExternalLink({
+  target = '_blank',
+  href,
+  rel = 'noopener noreferrer',
+  ...rest
+}: Omit<HTMLProps<HTMLAnchorElement>, 'as' | 'ref' | 'onClick'> & { href: string }) {
```
ianlapham, line 11: Is there a reason to rewrite the External link we already have? guessing styling stuff

src/lib/components/Header.tsx
```diff
@@ -30,21 +13,13 @@ const HeaderRow = styled(Row)`
...
     <HeaderRow iconSize={1.2}>
-      <Row gap={0.5}>
-        {logo && (
-          <UniswapA href={`https://app.uniswap.org/`}>
-            <Logo />
-          </UniswapA>
-        )}
-        {title && <ThemedText.Subhead1>{title}</ThemedText.Subhead1>}
-      </Row>
+      <Row gap={0.5}>{title && <ThemedText.Subhead1>{title}</ThemedText.Subhead1>}</Row>
```
ianlapham, line 22: do we need to wrap this in translation?

src/lib/components/BrandingFooter.tsx
```diff
@@ -0,0 +1,40 @@
...
+    will-change: transform;
+
+    :hover {
+      fill: ${({ theme }) => theme.onHover(theme.secondary)};
+      transform: rotate(-5deg);
+    }
+  }
+`
+
+export default function BrandingFooter() {
```
zzmp, line 31: Nit: BrandedFooter

ianlapham (approved):
Just small comments - go to go if theyre addresses

src/lib/components/ExternalLink.tsx
```diff
@@ -0,0 +1,17 @@
+import { HTMLProps } from 'react'
+
+/**
+ * Outbound link
+ */
+export default function ExternalLink({
```
zzmp, line 6: Would love a follow up pr to use this component for etherscan links

zzmp (changes requested):

src/lib/components/ExternalLink.tsx
zzmp: This doesn't use GA. We haven't discussed if we want analytics in the widget, but I'm assuming it will be a different arch then just plugging in GA.

src/lib/components/Header.tsx
zzmp: Should already be wrapped

src/lib/components/BrandingFooter.tsx
zzmp: (If there isn't an existing ThemedText component to match Phil's spec, you should make/name a new one for branding.)

src/lib/components/BrandingFooter.tsx
JFrankfurt: fwiw I really dislike this pattern of componentizing css rules.

JFrankfurt: Much prefer interpolating css into the `styled` definition.

PR #3196, opened 2022-01-27
feat: format usd prices, add loading states

ianlapham:
- fixes https://www.notion.so/uniswaplabs/dcde9145238445b4bfd5436a3c5d813d?v=be065901c104466c9dfb0f948accb2fd&p=4f3710105c14469ca9490fb481be04c5

- fixes https://www.notion.so/uniswaplabs/dcde9145238445b4bfd5436a3c5d813d?v=be065901c104466c9dfb0f948accb2fd&p=b78599c919a34d71918acd546674456e

### DIFF
--- a/src/lib/components/Swap/Toolbar.tsx
+++ b/src/lib/components/Swap/Toolbar.tsx
@@ -2,14 +2,13 @@ import { Trans } from '@lingui/macro'
 import { Currency, CurrencyAmount, TradeType } from '@uniswap/sdk-core'
 import { ALL_SUPPORTED_CHAIN_IDS } from 'constants/chains'
 import useUSDCPrice from 'hooks/useUSDCPrice'
-import { useAtomValue } from 'jotai/utils'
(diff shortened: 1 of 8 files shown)


src/lib/components/Swap/Toolbar.tsx
```diff
@@ -118,6 +121,14 @@ export default function Toolbar({ disabled }: { disabled?: boolean }) {
           </>
         )
       }
+      if (routeNotFound && inputCurrency && outputCurency && !routeIsLoading && !routeIsSyncing) {
```
JFrankfurt, line 124: `routeNotFound` is a liiittle confusing here, since we're doing boolean checks for other negative statements w/ !

JFrankfurt (approved):
one syntax suggestion, otherwise lgtm

src/lib/components/Swap/Toolbar.tsx
ianlapham: ill change to isRouteFound and reverse

PR #3207, opened 2022-01-31
fix: input/output value/balance styles

zzmp:


ianlapham (approved):
lgtm

src/lib/theme/type.tsx
```diff
@@ -5,8 +5,10 @@ import { Color } from './theme'
 
 type TextProps = Omit<TextPropsWithCss, 'css' | 'color'> & { color?: Color }
 
-const TextWrapper = styled(Text)<{ color?: Color }>`
+const TextWrapper = styled(Text)<{ color?: Color; lineHeight: string }>`
   color: ${({ color = 'currentColor', theme }) => theme[color as Color]};
+  // Avoid the need for placeholders by setting min-height to line-height.
+  min-height: ${({ lineHeight }) => lineHeight};
```
zzmp, line 11: I like this because line-height is unavailable when _using_ `ThemedText`, so this solves the issue of setting correct height without making you think about it downstream. It can probably replace some existing CSS that was added to get element height to match line-height in other instances.

PR #3208, opened 2022-01-31
feat: take tick range from URL

dmihal:
Lets an outside application pre-set the price range, for example:

`/add/ETH/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48/3000?minPrice=100&maxPrice=500`

### DIFF
--- a/src/pages/AddLiquidity/index.tsx
+++ b/src/pages/AddLiquidity/index.tsx
@@ -150,6 +150,31 @@ export default function AddLiquidity({
 
   useEffect(() => setShowCapitalEfficiencyWarning(false), [baseCurrency, quoteCurrency, feeAmount])
 
+  useEffect(() => {
+    if (history.location.search) {
+      const search = new URLSearchParams(history.location.search)
+      let updated = false
+
+      const minPrice = search.get('minPrice')
+      if (minPrice) {
+        updated = true
+        onLeftRangeInput(minPrice)
+        search.delete('minPrice')
+      }
+
+      const maxPrice = search.get('maxPrice')
+      if (maxPrice) {
+        updated = true
+        onRightRangeInput(maxPrice)
+        search.delete('maxPrice')
+      }
+
+      if (updated) {
+        history.replace({ search: search.toString() })
+      }
+    }
+  }, [history, onRightRangeInput, onLeftRangeInput])
+
   // txn values
   const deadline = useTransactionDeadline() // custom from users settings
 
--- a/src/pages/AddLiquidity/index.tsx
+++ b/src/pages/AddLiquidity/index.tsx
@@ -5,6 +5,7 @@ import { Currency, CurrencyAmount, Percent } from '@uniswap/sdk-core'
 import { FeeAmount, NonfungiblePositionManager } from '@uniswap/v3-sdk'
 import UnsupportedCurrencyFooter from 'components/swap/UnsupportedCurrencyFooter'
 import useActiveWeb3React from 'hooks/useActiveWeb3React'
+import useParsedQueryString from 'hooks/useParsedQueryString'
 import { useCallback, useContext, useEffect, useState } from 'react'
 import { AlertTriangle } from 'react-feather'
 import ReactGA from 'react-ga'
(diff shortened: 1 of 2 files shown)


dmihal:
Working test link:
https://interface-git-fork-dmihal-tick-range-uniswap.vercel.app/#/add/ETH/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48/3000?chain=mainnet&minPrice=2000&maxPrice=3000

JFrankfurt:
I love this. @tinaszheng what do you think of the behavior here? @dmihal has the effect removing the param instead of maintaining state sync w/ the ui like we do for fee level/asset/network. My gut says we should sync as the user changes so the url is shareable into a specific range.

src/pages/AddLiquidity/index.tsx
```diff
@@ -150,6 +150,31 @@ export default function AddLiquidity({
 
   useEffect(() => setShowCapitalEfficiencyWarning(false), [baseCurrency, quoteCurrency, feeAmount])
 
+  useEffect(() => {
+    if (history.location.search) {
+      const search = new URLSearchParams(history.location.search)
```
moodysalem, line 155: we have a `useParsedQueryString()` hook

src/pages/AddLiquidity/index.tsx
```diff
@@ -150,6 +150,31 @@ export default function AddLiquidity({
...
+
+      const maxPrice = search.get('maxPrice')
+      if (maxPrice) {
+        updated = true
+        onRightRangeInput(maxPrice)
+        search.delete('maxPrice')
+      }
+
+      if (updated) {
+        history.replace({ search: search.toString() })
```
moodysalem, line 173: it seems like this would cause an infinite loop?

tinaszheng:
ooo thanks so much for your contribution @dmihal !! looks great but i do agree that we should probably persist the query params and also sync the URL with the app ranges. you can take a look at https://github.com/Uniswap/interface/blob/main/src/components/Header/NetworkSelector.tsx#L284 for some inspiration for how we sync network params with the app (although that one is probably more complicated)

src/pages/AddLiquidity/index.tsx
dmihal: Cool, I've switched to using it

src/pages/AddLiquidity/index.tsx
dmihal: Updated, query params now stay in the URL

dmihal:
Hey everyone, thanks for the review!

I've updated the PR to persist the URL param, so users can easily have "sharable" URLs, similar to the network param shared by @tinaszheng

I've also added  useParsedQueryString at @moodysalem's suggestion.

src/pages/AddLiquidity/index.tsx
```diff
@@ -150,6 +153,16 @@ export default function AddLiquidity({
 
   useEffect(() => setShowCapitalEfficiencyWarning(false), [baseCurrency, quoteCurrency, feeAmount])
 
+  useEffect(() => {
+    if (typeof parsedQs.minPrice === 'string' && parsedQs.minPrice !== leftRangeTypedValue) {
+      onLeftRangeInput(parsedQs.minPrice)
+    }
+
+    if (typeof parsedQs.maxPrice === 'string' && parsedQs.minPrice !== rightRangeTypedValue) {
```
tinaszheng, line 161: i believe you want `parsedQs.maxPrice` here?

tinaszheng (changes requested):
just 2 requested changes - otherwise looks good so far!

src/pages/AddLiquidity/index.tsx
dmihal: Good catch!

JFrankfurt (approved):
lgtm -- I noticed though that some ticks end up with a bunch of trailing 0's, while others do not. (i.e., `40.100000` and `60.2`)

tinaszheng (approved):

JFrankfurt:
thanks @dmihal !

dmihal:
Thanks for merging this @JFrankfurt & @tinaszheng!

Great timing too, we just had a team call today, and hopefully launching the staker soon!

PR #3210, opened 2022-01-31
feat: update widget with client side SOR

ianlapham:
- replace sub-optimal client side routing with SOR on widget 
- create new `useBestTrade` hook that returns fallback route if SOR package fails


justindomingue (approved):

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -0,0 +1,114 @@
...
+        data: GetQuoteResult
+        error?: unknown
+      }
+    | undefined
+  >()
+
+  useEffect(() => {
+    async function fetchQuote() {
+      if (queryArgs) {
+        const result = await getClientSideQuote(queryArgs)
```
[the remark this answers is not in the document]
ianlapham: Yeah good idea, will need error state to use fallback

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -43,12 +43,20 @@ export default function useClientSideSmartOrderRouterTrade<TTradeType extends Tr
   useEffect(() => {
     async function fetchQuote() {
       if (queryArgs) {
```
justindomingue, line 45: Hm if query args isn’t defined you may never reset to loading(false)?might be better to set loading right before the await get client side quote ?

package.json
```diff
@@ -162,7 +161,7 @@
     "@ethersproject/constants": "^5.4.0",
     "@ethersproject/contracts": "^5.4.1",
     "@ethersproject/hash": "^5.4.0",
-    "@ethersproject/providers": "^5.4.5",
+    "@ethersproject/providers": "5.4.0",
```
JFrankfurt, line 164: version drop intended?

src/lib/hooks/routing/clientSideSmartOrderRouter/index.ts
```diff
@@ -50,3 +51,49 @@ export async function getQuote(
...
+  tokenInChainId: ChainId
+  tokenInDecimals: number
+  tokenInSymbol?: string
+  tokenOutAddress: string
+  tokenOutChainId: ChainId
+  tokenOutDecimals: number
+  tokenOutSymbol?: string
+  amount: string
+  type: 'exactIn' | 'exactOut'
+}) {
```
JFrankfurt, line 78: can you move this up to a type definition?

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -0,0 +1,122 @@
...
+      if (queryArgs) {
+        try {
+          const result = await getClientSideQuote(queryArgs)
+          setLoading(false)
+          setFetchedResult({
+            data: result.data,
+            error: result.error,
+          })
+        } catch (e) {
+          setLoading(false)
```
JFrankfurt, line 54: nit: use `finally` for `setLoading(false)`

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    const otherAmount =
+      tradeType === TradeType.EXACT_INPUT
+        ? currencyOut && quoteResult
+          ? CurrencyAmount.fromRawAmount(currencyOut, quoteResult.quote)
+          : undefined
+        : currencyIn && quoteResult
+        ? CurrencyAmount.fromRawAmount(currencyIn, quoteResult.quote)
+        : undefined
```
JFrankfurt, line 101: My preference is to avoid nested ternaries in favor of `if (variableDescribingTheSituation) { .. }`

src/lib/hooks/routing/clientSideSmartOrderRouter/index.ts
JFrankfurt: there are a few function signatures here that I think also merit their own type definitions

JFrankfurt (approved):
A few nitpicky things

package.json
zzmp: did you talk to SOR authors about this version drop? might be better for them to bump than for you to drop.
Alternatively, they should be using `@ethersproject` and not `ethers`, and using `^` (at-least) versioning, so that the projects can coalesce on one version.

src/lib/hooks/swap/useSwapInfo.tsx
```diff
@@ -55,10 +55,7 @@ function useComputeSwapInfo(): SwapInfo {
     [inputCurrency, isExactIn, outputCurrency, amount]
   )
 
-  /**
-   * @TODO (ianlapham): eventually need a strategy for routing API here
```
zzmp, line 59: Is this comment resolved?

yarn.lock
```diff
@@ -1828,6 +1828,31 @@
   dependencies:
     "@ethersproject/logger" "^5.5.0"
 
+"@ethersproject/providers@5.4.0":
```
zzmp, line 1831: Run `npx yarn-deduplicate` to remove the extra copy. (This should be it's own commit.)

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -0,0 +1,122 @@
...
+        } catch (e) {
+          setLoading(false)
+          setFetchedResult({
+            data: undefined,
+            error: true,
+          })
+        }
+      }
+    }
+    setLoading(true)
```
zzmp, line 62: nit: move these calls above the definition and separate with a newline. the definiton will be hoisted, and IMO it's more readable that way.

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -0,0 +1,122 @@
...
+    fetchQuote()
+  }, [queryArgs])
+
+  const [quoteResult, error] = useMemo(() => {
```
zzmp, line 66: This is unnecessary - `fetchedResult` should take this typing initially.

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
zzmp: or wrap the whole thing in a `try {} finally {}`

src/lib/hooks/routing/useRoutingAPIArguments.ts
```diff
@@ -0,0 +1,40 @@
+import { Currency, CurrencyAmount, TradeType } from '@uniswap/sdk-core'
+import { useMemo } from 'react'
+
+/**
+ * Returns query arguments for the Routing API query or undefined if the
+ * query should be skipped.
```
zzmp, line 6: Include in the doc comment that the input does not need to be memoized, as it will be destructured.

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -0,0 +1,122 @@
...
+  const [fetchedResult, setFetchedResult] = useState<
+    | {
+        data: GetQuoteResult | undefined
+        error?: unknown
+      }
+    | undefined
+  >()
+
+  useEffect(() => {
+    async function fetchQuote() {
```
zzmp, line 44: If the only thing this depends on is `queryArgs`, move the definition outside of the component, and pass in `queryArgs` explicitly, to avoid the inline definition and long component definition.

package.json
ianlapham: yeah we need this version to match the version from SOR package for types to be compatiable

src/lib/hooks/swap/useSwapInfo.tsx
ianlapham: Kinda depends, id say its more of a product decision now if we want to use the API, but now that we have SOR we aren't missing anything significant - ill leave a note that this is where routing API logic would go

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
ianlapham: Would need to accept the state setters too, is that weird?

JFrankfurt (approved):

PR #3213, opened 2022-02-01
fix: update walletlink-connector to 6.2.11

brendanww:
walletlink-connector 6.2.11 updates walletlink to 2.4.6.

Coinbase Wallet clients will update on Monday to support `wallet_addEthereumChain` requests for any network. We made a change on the clients that is not compatible with older versions of walletlink (anything before 2.4.5).

2.4.6 also includes a fix for a common connectivity issue https://github.com/walletlink/coinbase-wallet-sdk/pull/277

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "@web3-react/injected-connector": "^6.0.7",
     "@web3-react/portis-connector": "^6.0.9",
     "@web3-react/walletconnect-connector": "^7.0.2-alpha.0",
-    "@web3-react/walletlink-connector": "^6.2.8",
+    "@web3-react/walletlink-connector": "^6.2.11",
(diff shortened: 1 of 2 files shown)


moodysalem:
not sure why all the integrity hashes were removed in this pr, i reverted that part of the pr [here](https://github.com/Uniswap/interface/commit/e5be3ebf8f7955f8d8ddaeb86d90746166566ee9) and did a yarn install to get the new lockfile [here](https://github.com/Uniswap/interface/commit/1b78ceec1080e9e9c185b2ceaaa2c1abc2c238f5)

brendanww:
I think I probably switched to use an older yarn version with `yvm` for another repo. thanks for the heads up and fixing for me; I updated my `yvm` default so I am on a recent version by default.

PR #3216, opened 2022-02-01
fix: update quote amount for price summary

ianlapham:


aballerr:
asked ian, said I could close

PR #3219, opened 2022-02-02
feat(widgets): support convenience fee in trades

JFrankfurt:


src/hooks/useSwapCallArguments.tsx
```diff
@@ -167,15 +169,16 @@ export function useSwapCallArguments(
       ]
     }
   }, [
-    trade,
-    recipient,
-    library,
     account,
```
zzmp, line 172: You should add sorting to eslint in a separate PR so that your PRs don't add additional churn.

src/hooks/useSwapCallback.tsx
```diff
@@ -33,7 +33,7 @@ export function useSwapCallback(
     state,
     callback: libCallback,
     error,
-  } = useLibSwapCallBack(trade, allowedSlippage, recipient, signatureData, deadline)
+  } = useLibSwapCallBack({ trade, allowedSlippage, recipientAddressOrName: recipient, signatureData, deadline })
```
zzmp, line 36: Was this broken before?

src/lib/hooks/swap/useSwapCallback.tsx
```diff
@@ -17,18 +18,33 @@ export enum SwapCallbackState {
...
+  callback: null | (() => Promise<TransactionResponse>)
+  error: ReactNode | null
+}
+interface UseSwapCallbackArgs {
+  trade: AnyTrade | undefined // trade to execute, required
+  allowedSlippage: Percent // in bips
+  recipientAddressOrName: string | null | undefined // the ENS name or address of the recipient of the trade, or null if swap should be returned to sender
+  signatureData: SignatureData | null | undefined
+  deadline: BigNumber | undefined
+  fee?: FeeOptions
```
zzmp, line 32: Use consistent typing. fee should be `FeeOptions | undefined`, or else the other optional inputs should be marked as optional.

zzmp: I would prefer they all be `X | undefined` so that they are passed, even if undefined, and future updates don't miss them because they are optional.

src/lib/hooks/swap/useSwapDefaults.ts
```diff
@@ -37,7 +37,7 @@ export default function useSwapDefaults({
   defaultOutputAddress,
   defaultOutputAmount,
 }: UseSwapDefaultsArgs) {
-  const updateSwap = useUpdateAtom(swapAtom)
+  const [swap, updateSwap] = useAtom(swapAtom)
```
zzmp, line 40: This adds a duplicate definition for `useAtomValue(swapAtom)`, so I don't think `swap` should be returned (because multiple codepaths doing the same thing leaves more room for error).

If you're worried about clarity, instead rename `useSwapDefaults` to `useSetSwapDefaults` so it is clear it only returns a setter.

src/lib/hooks/swap/useSwapInfo.tsx
```diff
@@ -42,6 +44,8 @@ function useComputeSwapInfo(): SwapInfo {
     [Field.OUTPUT]: outputCurrency,
   } = useAtomValue(swapAtom)
```
zzmp, line 46: nit: newline not necessary.

src/lib/hooks/swap/useSwapInfo.tsx
```diff
@@ -42,6 +44,8 @@ function useComputeSwapInfo(): SwapInfo {
...
 
+  const feeOptions = useAtomValue(feeOptionsAtom)
```
zzmp, line 47: nit: rename to `fee` instead of `feeOptions` (and `feeAtom`), and use the shorthand when you instantiate the object below.

src/lib/hooks/swap/useSwapInfo.tsx
```diff
@@ -152,6 +157,10 @@ const swapInfoAtom = atom<SwapInfo>({
   currencyAmounts: {},
   trade: { state: TradeState.INVALID },
   allowedSlippage: new Percent(0),
+  fee: {
+    fee: new Percent(0),
```
zzmp, line 161: This should be the default value for the atom, so import it from the file defining the atom.

src/lib/state/swap.ts
```diff
@@ -27,3 +28,8 @@ export const independentFieldAtom = pickAtom(swapAtom, 'independentField')
 
 // If set to a transaction hash, that transaction will display in a status dialog.
 export const displayTxHashAtom = atom<string | undefined>(undefined)
+
+export const feeOptionsAtom = atom<FeeOptions>({
+  fee: new Percent(0),
```
zzmp, line 33: Define this as a default so it can be re-used in `useSwapInfo` (see comment in that file).

src/lib/components/Swap/SwapButton.tsx
```diff
@@ -119,13 +120,14 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
   const deadline = useTransactionDeadline()
 
   // the callback to execute the swap
-  const { callback: swapCallback } = useSwapCallback(
-    optimizedTrade,
+  const { callback: swapCallback } = useSwapCallback({
+    trade: optimizedTrade,
```
zzmp, line 124: Did you rename this? Was it broken before?

src/lib/hooks/swap/useConvenienceFee.ts
```diff
@@ -0,0 +1,43 @@
...
+import { useAtom } from 'jotai'
+import { feeOptionsAtom } from 'lib/state/swap'
+import { useCallback, useEffect } from 'react'
+
+interface FeeOptionsArgs {
+  convenienceFee?: number
+  convenienceFeeRecipient?: string | string | { [chainId: number]: string }
+}
+
+export default function useConvenienceFee({ convenienceFee, convenienceFeeRecipient }: FeeOptionsArgs) {
```
zzmp, line 12: This is only ever used as a setter, so rename it (same with `useSwapDefaults`): `useSetConvenienceFee`; and don't return any value.

src/lib/hooks/swap/useConvenienceFee.ts
```diff
@@ -0,0 +1,43 @@
...
+  const { chainId } = useActiveWeb3React()
+  const [feeOptions, updateFeeOptions] = useAtom(feeOptionsAtom)
+
+  const setToDefaults = useCallback(() => {
+    updateFeeOptions({
+      fee: new Percent(0),
```
zzmp, line 18: import a default value from the state definition

src/lib/hooks/swap/useConvenienceFee.ts
```diff
@@ -0,0 +1,43 @@
+import { Percent } from '@uniswap/sdk-core'
+import useActiveWeb3React from 'hooks/useActiveWeb3React'
+import { useAtom } from 'jotai'
+import { feeOptionsAtom } from 'lib/state/swap'
+import { useCallback, useEffect } from 'react'
+
+interface FeeOptionsArgs {
+  convenienceFee?: number
+  convenienceFeeRecipient?: string | string | { [chainId: number]: string }
```
zzmp, line 9: duplicate string

src/lib/hooks/swap/useConvenienceFee.ts
```diff
@@ -0,0 +1,43 @@
...
+  }, [updateFeeOptions])
+
+  useEffect(() => {
+    if (convenienceFee && convenienceFeeRecipient) {
+      if (typeof convenienceFeeRecipient === 'string') {
+        updateFeeOptions({
+          fee: new Percent(convenienceFee, 10_000),
+          recipient: convenienceFeeRecipient,
+        })
+      } else if (chainId && typeof convenienceFeeRecipient === 'object' && convenienceFeeRecipient[chainId]) {
```
zzmp, line 30: At this point you know that convenienceFeeRecipient is an object, so make this simpler:
`} else if (chainId && convenienceFeeRecipient?.[chainId]) {`

src/lib/hooks/swap/useConvenienceFee.ts
```diff
@@ -0,0 +1,43 @@
...
+interface FeeOptionsArgs {
+  convenienceFee?: number
+  convenienceFeeRecipient?: string | string | { [chainId: number]: string }
+}
+
+export default function useConvenienceFee({ convenienceFee, convenienceFeeRecipient }: FeeOptionsArgs) {
+  const { chainId } = useActiveWeb3React()
+  const [feeOptions, updateFeeOptions] = useAtom(feeOptionsAtom)
+
+  const setToDefaults = useCallback(() => {
```
zzmp, line 16: nit: If you import a default value this becomes a one-liner, and I think you should omit the callback and just use it in the `useEffect` directly.

src/lib/hooks/swap/useConvenienceFee.ts
```diff
@@ -0,0 +1,43 @@
...
+        updateFeeOptions({
+          fee: new Percent(convenienceFee, 10_000),
+          recipient: convenienceFeeRecipient,
+        })
+      } else if (chainId && typeof convenienceFeeRecipient === 'object' && convenienceFeeRecipient[chainId]) {
+        updateFeeOptions({
+          fee: new Percent(convenienceFee, 10_000),
+          recipient: convenienceFeeRecipient[chainId],
+        })
+      } else {
```
zzmp, line 35: Instead of `else` branches, do you think it would be cleaner to `return` after calling `updateFeeOptions`, and if you haven't returned yet call `updateFeeOptions(DEFAULT_CONVENIENCE_FEE)`? Seems like less branching.

src/lib/hooks/swap/useConvenienceFee.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  const setToDefaults = useCallback(() => {
+    updateFeeOptions({
+      fee: new Percent(0),
+      recipient: '',
+    })
+  }, [updateFeeOptions])
+
+  useEffect(() => {
+    if (convenienceFee && convenienceFeeRecipient) {
```
zzmp, line 24: You should derive the convenience fee as a percent once, here, and it should follow this pattern:

https://github.com/Uniswap/interface/blob/main/src/lib/components/Swap/Settings/MaxSlippageSelect.tsx#L112-L114

I think that you are dividing the intended amount by 100 here.

zzmp (changes requested):
See comments

src/lib/hooks/swap/useSwapDefaults.ts
JFrankfurt: `useSet` is just so awful though

JFrankfurt: `useSync` maybe

src/lib/hooks/swap/useSwapCallback.tsx
JFrankfurt: > future updates don't miss them because they are optional

I'm not following here

src/lib/hooks/swap/useSwapInfo.tsx
JFrankfurt: up for doing the shorthand, but the sdk type calls this `FeeOptions`, so I want to stick with that for the atom and type

src/lib/hooks/swap/useConvenienceFee.ts
JFrankfurt: The integrator is passing an integer value for bips, so if we use that as the numerator we need to divide by 10,000 for a valid %. i.e., 100/10_000 = 1%

PR #3223, opened 2022-02-03
fix(widgets): light-mode button styling

JFrankfurt:


zzmp:
Could you update the description to include what you're trying to do? I _think_ you're trying to update the foreground color.
This applies to _every_ button - is there a reason you're not focusing on a specific button?

src/lib/components/Button.tsx
```diff
@@ -6,7 +6,7 @@ export const BaseButton = styled.button`
   background-color: transparent;
   border: none;
   border-radius: 0.5em;
-  color: currentColor;
+  color: ${({ theme: { primary } }) => primary};
```
zzmp, line 9: Use `theme.color` instead of destructuring it to match all other uses.

PR #3226, opened 2022-02-03
feat: create use best trade hook for widgets

ianlapham:
- create an intermediate hook for fetching the best trade based on current swap state 
- the new hook mimics the same behavior as app and is used for two purposes: 

 1. adds in logic to use the simple client side routing if the SOR package fails for any reason (this becomes extra helpful as we move from client side SOR to API, where API has potential downtime)
 
 2. helps debounce amount values so less calls are made to the SOR logic (as it is very data intensive). This is how we get an accurate "syncing" state which the UI components can reference for loading animations

### DIFF
--- a/src/lib/hooks/swap/useBestTrade.ts
+++ b/src/lib/hooks/swap/useBestTrade.ts
@@ -21,6 +21,7 @@ export function useBestTrade(
   trade: InterfaceTrade<Currency, Currency, TradeType> | undefined
 } {
   // debounce used to prevent excessive requests to SOR, as it is data intensive
+  // this helps provide a "syncing" state the UI can reference for loading animations
   const [debouncedAmount, debouncedOtherCurrency] = useDebounce(
     useMemo(() => [amountSpecified, otherCurrency], [amountSpecified, otherCurrency]),
     200
--- a/src/lib/hooks/swap/useBestTrade.ts
+++ b/src/lib/hooks/swap/useBestTrade.ts
@@ -1,0 +1,62 @@
+import { Currency, CurrencyAmount, TradeType } from '@uniswap/sdk-core'
+import { useClientSideV3Trade } from 'hooks/useClientSideV3Trade'
+import useDebounce from 'hooks/useDebounce'
+import { useMemo } from 'react'
+import { InterfaceTrade, TradeState } from 'state/routing/types'
+
+import useClientSideSmartOrderRouterTrade from '../routing/useClientSideSmartOrderRouterTrade'
+
+/**
+ * Returns the best v2+v3 trade for a desired swap.
+ * @param tradeType whether the swap is an exact in/out
+ * @param amountSpecified the exact amount to swap in/out
+ * @param otherCurrency the desired output/payment currency
+ */
+export function useBestTrade(
+  tradeType: TradeType,
+  amountSpecified?: CurrencyAmount<Currency>,
+  otherCurrency?: Currency
+): {
+  state: TradeState
+  trade: InterfaceTrade<Currency, Currency, TradeType> | undefined
+} {
+  // debounce used to prevent excessive requests to SOR, as it is data intensive
+  const [debouncedAmount, debouncedOtherCurrency] = useDebounce(
+    useMemo(() => [amountSpecified, otherCurrency], [amountSpecified, otherCurrency]),
+    200
+  )
+
+  const routingAPITrade = useClientSideSmartOrderRouterTrade(tradeType, debouncedAmount, debouncedOtherCurrency)
+
+  const isLoading = amountSpecified !== undefined && debouncedAmount === undefined
+
+  // consider trade debouncing when inputs/outputs do not match
+  const debouncing =
+    routingAPITrade.trade &&
+    amountSpecified &&
+    (tradeType === TradeType.EXACT_INPUT
+      ? !routingAPITrade.trade.inputAmount.equalTo(amountSpecified) ||
+        !amountSpecified.currency.equals(routingAPITrade.trade.inputAmount.currency) ||
+        !debouncedOtherCurrency?.equals(routingAPITrade.trade.outputAmount.currency)
+      : !routingAPITrade.trade.outputAmount.equalTo(amountSpecified) ||
+        !amountSpecified.currency.equals(routingAPITrade.trade.outputAmount.currency) ||
+        !debouncedOtherCurrency?.equals(routingAPITrade.trade.inputAmount.currency))
+
+  const useFallback = !debouncing && routingAPITrade.state === TradeState.NO_ROUTE_FOUND
+
+  // use simple client side logic as backup if SOR is not available
+  const bestV3Trade = useClientSideV3Trade(
+    tradeType,
+    useFallback ? debouncedAmount : undefined,
+    useFallback ? debouncedOtherCurrency : undefined
+  )
+
+  return useMemo(
+    () => ({
+      ...(useFallback ? bestV3Trade : routingAPITrade),
+      ...(debouncing ? { state: TradeState.SYNCING } : {}),
+      ...(isLoading ? { state: TradeState.LOADING } : {}),
+    }),
+    [bestV3Trade, debouncing, isLoading, routingAPITrade, useFallback]
+  )
+}
--- a/src/lib/hooks/swap/useBestTrade.ts
+++ b/src/lib/hooks/swap/useBestTrade.ts
@@ -6,38 +6,44 @@ import { InterfaceTrade, TradeState } from 'state/routing/types'
 
 import useClientSideSmartOrderRouterTrade from '../routing/useClientSideSmartOrderRouterTrade'
 
+/**
+ * Returns the currency amount from independent field, currency from independent field,
+ * and currency from dependent field.
+ */
 function getTradeInputs(
   trade: InterfaceTrade<Currency, Currency, TradeType> | undefined,
   tradeType: TradeType
 ): [CurrencyAmount<Currency> | undefined, Currency | undefined, Currency | undefined] {
-  if (!trade) return [undefined, undefined, undefined]
-  if (tradeType === TradeType.EXACT_INPUT) {
-    return [trade.inputAmount, trade.inputAmount.currency, trade.outputAmount.currency]
-  }
-  if (tradeType === TradeType.EXACT_OUTPUT) {
-    return [trade.outputAmount, trade.outputAmount.currency, trade.inputAmount.currency]
+  if (trade) {
+    if (tradeType === TradeType.EXACT_INPUT) {
+      return [trade.inputAmount, trade.inputAmount.currency, trade.outputAmount.currency]
+    }
+    if (tradeType === TradeType.EXACT_OUTPUT) {
+      return [trade.outputAmount, trade.outputAmount.currency, trade.inputAmount.currency]
+    }
   }
   return [undefined, undefined, undefined]
 }
 
-// Checks for case where debounced values are stale compared to latest values from trade.
-function isSyncing(
-  trade: InterfaceTrade<Currency, Currency, TradeType> | undefined,
-  debouncedAmount: CurrencyAmount<Currency> | undefined,
-  debouncedOtherCurrency: Currency | undefined,
-  amountFromLatestTrade: CurrencyAmount<Currency> | undefined,
-  currencyFromTrade: Currency | undefined,
-  otherCurrencyFromTrade: Currency | undefined
-): boolean {
+interface TradeDebouncingParams {
+  amounts: [CurrencyAmount<Currency> | undefined, CurrencyAmount<Currency> | undefined]
+  indepdenentCurrencies: [Currency | undefined, Currency | undefined]
+  dependentCurrencies: [Currency | undefined, Currency | undefined]
+}
+
+/**
+ * Returns wether debounced values are stale compared to latest values from trade.
+ */
+function isTradeDebouncing({ amounts, indepdenentCurrencies, dependentCurrencies }: TradeDebouncingParams): boolean {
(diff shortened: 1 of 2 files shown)


src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -0,0 +1,63 @@
...
+  const debouncing =
+    routingAPITrade.trade &&
+    amountSpecified &&
+    (tradeType === TradeType.EXACT_INPUT
+      ? !routingAPITrade.trade.inputAmount.equalTo(amountSpecified) ||
+        !amountSpecified.currency.equals(routingAPITrade.trade.inputAmount.currency) ||
+        !debouncedOtherCurrency?.equals(routingAPITrade.trade.outputAmount.currency)
+      : !routingAPITrade.trade.outputAmount.equalTo(amountSpecified) ||
+        !amountSpecified.currency.equals(routingAPITrade.trade.outputAmount.currency) ||
+        !debouncedOtherCurrency?.equals(routingAPITrade.trade.inputAmount.currency))
```
JFrankfurt, line 44: can you simplify this?

JFrankfurt (approved):
lgtm - one ternary logic simplification req

src/lib/hooks/swap/useBestTrade.ts
ianlapham: yeah this is leftover from app and agreed its brutal to read. Will attempt to clean up

JFrankfurt: yeah, figured that was prob the case so just gave it an approval anyway

if you feel like fixing I will love you forever ❤️

src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -0,0 +1,63 @@
...
+  amountSpecified?: CurrencyAmount<Currency>,
+  otherCurrency?: Currency
+): {
+  state: TradeState
+  trade: InterfaceTrade<Currency, Currency, TradeType> | undefined
+} {
+  // debounce used to prevent excessive requests to SOR, as it is data intensive
+  // this helps provide a "syncing" state the UI can reference for loading animations
+  const [debouncedAmount, debouncedOtherCurrency] = useDebounce(
+    useMemo(() => [amountSpecified, otherCurrency], [amountSpecified, otherCurrency]),
```
zzmp, line 26: good use of `useMemo` within the debounce

src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -0,0 +1,63 @@
...
+    200
+  )
+
+  const routingAPITrade = useClientSideSmartOrderRouterTrade(tradeType, debouncedAmount, debouncedOtherCurrency)
+
+  const isLoading = amountSpecified !== undefined && debouncedAmount === undefined
```
zzmp, line 32: debouncedAmount will only be undefined if it has not been debounced yet; is that the intention?

src/lib/hooks/swap/useBestTrade.ts
zzmp: this seems like it is examining effects to determine if debounce has occurred. Can you rewrite the debounce to keep state so you don't need this complex logic?

src/lib/hooks/swap/useBestTrade.ts
ianlapham: Yeah or if the amount in is undefined as well, that is the intention: to track the case of no `amountSpecified` or a new `amountSpecified` that hasnt been debounced yet

src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -0,0 +1,76 @@
...
+  trade: InterfaceTrade<Currency, Currency, TradeType> | undefined
+} {
+  // debounce used to prevent excessive requests to SOR, as it is data intensive
+  // this helps provide a "syncing" state the UI can reference for loading animations
+  const [debouncedAmount, debouncedOtherCurrency] = useDebounce(
+    useMemo(() => [amountSpecified, otherCurrency], [amountSpecified, otherCurrency]),
+    200
+  )
+
+  const routingAPITrade = useClientSideSmartOrderRouterTrade(tradeType, debouncedAmount, debouncedOtherCurrency)
```
zzmp, line 30: nit: rename to `clientSORTrade` for clarity

src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  // check if the artifical delay we impose is still occuring
+  const debouncing = amountSpecified !== undefined && debouncedAmount === undefined
```
zzmp, line 33: This doesn't seem accurate to me. If `amountSpecified && debouncedAmount` but `amountSpecified !== debouncedAmount` (and the same for `otherCurrency`) _then_ we're still debouncing, no?

src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -0,0 +1,76 @@
...
+ */
+export function useBestTrade(
+  tradeType: TradeType,
+  amountSpecified?: CurrencyAmount<Currency>,
+  otherCurrency?: Currency
+): {
+  state: TradeState
+  trade: InterfaceTrade<Currency, Currency, TradeType> | undefined
+} {
+  // debounce used to prevent excessive requests to SOR, as it is data intensive
```
zzmp, line 23: General nit: comments should still be full sentences for clarity (ie capitalized, ending with a period, etc), unless it's something like an inline label for a field.

src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -0,0 +1,76 @@
...
+    200
+  )
+
+  const routingAPITrade = useClientSideSmartOrderRouterTrade(tradeType, debouncedAmount, debouncedOtherCurrency)
+
+  // check if the artifical delay we impose is still occuring
+  const debouncing = amountSpecified !== undefined && debouncedAmount === undefined
+
+  // the amount associated with field user is typing in
+  const amountFromLatestTrade =
```
zzmp, line 36: nit: could this be expressed simpler as

```
tradeType === routingApiTrade.trade?.[TradeType.EXACT_INPUT ? 'inputAmount' : 'outputAmount']
```

zzmp: nit: this treats the enum as a binary, but that's not a feature of an enum.

src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -0,0 +1,76 @@
...
+    tradeType === TradeType.EXACT_INPUT ? routingAPITrade.trade?.inputAmount : routingAPITrade.trade?.outputAmount
+
+  // ordered currencies from API trade
+  const [currencyFromTrade, otherCurrencyFromTrade] =
+    tradeType === TradeType.EXACT_INPUT
```
zzmp, line 41: nit: this treats the enum as binary
nit: destructure this out (and above) so that these are less verbose.

src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  // check if the artifical delay we impose is still occuring
+  const debouncing = amountSpecified !== undefined && debouncedAmount === undefined
+
+  // the amount associated with field user is typing in
+  const amountFromLatestTrade =
+    tradeType === TradeType.EXACT_INPUT ? routingAPITrade.trade?.inputAmount : routingAPITrade.trade?.outputAmount
+
+  // ordered currencies from API trade
+  const [currencyFromTrade, otherCurrencyFromTrade] =
```
zzmp, line 40: there's no reason you can't destructure all 3 together (currencies and amountFromLatestTrade). I think it'd be clearer if you did, but did so as part of an if/else statement. You can do so as a separate helper function so that you are returning in each case and don't need to use `let`.

```
const [amount, currency, otherCurrency] = getTradeInputs(routingApiTrade.trade)
```

src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -0,0 +1,76 @@
...
+    tradeType === TradeType.EXACT_INPUT
+      ? [routingAPITrade.trade?.inputAmount?.currency, routingAPITrade.trade?.outputAmount?.currency]
+      : [routingAPITrade.trade?.outputAmount?.currency, routingAPITrade.trade?.inputAmount?.currency]
+
+  // check that amount from user input matches latest trade
```
zzmp, line 45: L45-55 could be moved into their own function, `isSyncing` (or similar), to make it easier to follow inline with the complex logic isolated.

src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -0,0 +1,76 @@
...
+    debouncedAmount?.currency?.equals(currencyFromTrade) &&
+    debouncedOtherCurrency?.equals(otherCurrencyFromTrade)
+
+  // either amount or currencies from latest API trade do not match swap state
+  const syncing = !amountsMatch || !currenciesMatch
+
+  const useFallback = !syncing && routingAPITrade.state === TradeState.NO_ROUTE_FOUND
+
+  // use simple client side logic as backup if SOR is not available
+  const bestV3Trade = useClientSideV3Trade(
```
zzmp, line 62: Either rename this to `fallbackTrade` or `clientV3Trade` or rename `useFallback` to `useV3Trade` or `fallbackToClientV3Trade` or similar. Right now it's unclear that they're coupled.

src/lib/hooks/swap/useBestTrade.ts
ianlapham: yeah i think youre right here, will change

src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -0,0 +1,103 @@
...
+  // Use a simple client side logic as backup if SOR is not available.
+  const fallbackTrade = useClientSideV3Trade(
+    tradeType,
+    useFallback ? debouncedAmount : undefined,
+    useFallback ? debouncedOtherCurrency : undefined
+  )
+
+  return useMemo(
+    () => ({
+      ...(useFallback ? fallbackTrade : clientSORTrade),
```
zzmp, line 97: nit: Will `fallbackTrade` be undefined if the amount/otherCurrency are not passed to it? If that's the case, I think `fallbackTrade || clientSORTrade` is clearer.

zzmp (approved):

src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -0,0 +1,103 @@
...
+import { useMemo } from 'react'
+import { InterfaceTrade, TradeState } from 'state/routing/types'
+
+import useClientSideSmartOrderRouterTrade from '../routing/useClientSideSmartOrderRouterTrade'
+
+function getTradeInputs(
+  trade: InterfaceTrade<Currency, Currency, TradeType> | undefined,
+  tradeType: TradeType
+): [CurrencyAmount<Currency> | undefined, Currency | undefined, Currency | undefined] {
+  if (!trade) return [undefined, undefined, undefined]
```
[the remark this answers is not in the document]
ianlapham: yep will update conditional, Typecheck prevents me from limiting to [] though

src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -0,0 +1,103 @@
...
+  if (tradeType === TradeType.EXACT_OUTPUT) {
+    return [trade.outputAmount, trade.outputAmount.currency, trade.inputAmount.currency]
+  }
+  return [undefined, undefined, undefined]
+}
+
+// Checks for case where debounced values are stale compared to latest values from trade.
+function isSyncing(
+  trade: InterfaceTrade<Currency, Currency, TradeType> | undefined,
+  debouncedAmount: CurrencyAmount<Currency> | undefined,
```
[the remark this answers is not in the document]
ianlapham: agreed lot of room for error right now, will update to a type

src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -0,0 +1,103 @@
...
+  debouncedOtherCurrency: Currency | undefined,
+  amountFromLatestTrade: CurrencyAmount<Currency> | undefined,
+  currencyFromTrade: Currency | undefined,
+  otherCurrencyFromTrade: Currency | undefined
+): boolean {
+  // Ensure that amount from user input matches latest trade.
+  const amountsMatch = trade && debouncedAmount && amountFromLatestTrade?.equalTo(debouncedAmount)
```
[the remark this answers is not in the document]
ianlapham: yep, was copying logic from app but after looking at this more i dont think its needed

src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -0,0 +1,103 @@
...
+    return [trade.inputAmount, trade.inputAmount.currency, trade.outputAmount.currency]
+  }
+  if (tradeType === TradeType.EXACT_OUTPUT) {
+    return [trade.outputAmount, trade.outputAmount.currency, trade.inputAmount.currency]
+  }
+  return [undefined, undefined, undefined]
+}
+
+// Checks for case where debounced values are stale compared to latest values from trade.
+function isSyncing(
```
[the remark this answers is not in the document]
ianlapham: Yeah im okay with that, this essentially is a second check that the trade from SOR is updating

src/lib/hooks/swap/useBestTrade.ts
ianlapham: the goal here is to only use the fallback when trades are syncing and the SOR package hasnt found a route, which is why we have the useFallback check

Theres a case where both trades are valid, but we dont wanna use fallbacktrade in that case

PR #3227, opened 2022-02-03
fix: reset settings incl max slippage

zzmp:


src/lib/components/Swap/Settings/MaxSlippageSelect.tsx
```diff
@@ -97,10 +107,16 @@ const Warning = memo(function Warning({ state, showTooltip }: { state: WarningSt
   )
 })
 
-export default function MaxSlippageSelect() {
+interface MaxSlippageSelectHandle {
+  reset: () => void
+}
+
+const MaxSlippageSelect = forwardRef<MaxSlippageSelectHandle>(function MaxSlippageSelect(_, ref) {
```
JFrankfurt, line 114: variable shadowing here intentional?

Rds2008ca-libellule: I really want to learn about this and also contribute, please if I can assist let me know what to do

src/lib/components/Swap/Settings/index.tsx
```diff
@@ -16,19 +16,26 @@ import TransactionTtlInput from './TransactionTtlInput'
...
+        <TextButton onClick={onReset}>
           <ThemedText.ButtonSmall>
             <Trans>Reset</Trans>
           </ThemedText.ButtonSmall>
         </TextButton>
       </Header>
       <Column gap={1} style={{ paddingTop: '1em' }} ref={setBoundary} padded css={scrollbar}>
         <BoundaryProvider value={boundary}>
-          <MaxSlippageSelect />
+          <MaxSlippageSelect ref={setMaxSlippage} />
```
JFrankfurt, line 38: Are you intending to set the ref here to the setState fn?

src/lib/components/Swap/Settings/MaxSlippageSelect.tsx
zzmp: this is intentional - lint requires me to name the function, but the function and the wrapped function are functionally the same so I don't want to introduce a new name.

src/lib/components/Swap/Settings/index.tsx
zzmp: Typings are broken when using `useRef` instead of `useState`, so I intentionally use `useState`.
If I were to use `useRef` I would need to coerce the type, and I'd rather not do that.
This pattern is also used in https://github.com/Uniswap/interface/blob/main/src/lib/components/TokenSelect/index.tsx#L63, for the same reasons.

src/lib/components/Swap/Settings/MaxSlippageSelect.tsx
zzmp: Hi @Rds2008ca-libellule, glad to see you want to contribute!

We maintain [a list of "good first issues" on GitHub](https://github.com/Uniswap/interface/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22). That is the best place to get started contributing.

Rds2008ca-libellule: Ok thanks

src/lib/components/Swap/Settings/index.tsx
JFrankfurt: Is it not an option to fix the types?

zzmp: We can coerce the type, but it functions fine to just use state instead of a ref, and does so without subverting the type system, so I prefer it.
The types are broken in React, so it's more complex to fix the types than to work around them.

PR #3232, opened 2022-02-03
feat: widget loading animations polish 

ianlapham:


src/components/Loader/styled.tsx
```diff
@@ -29,8 +29,8 @@ export const LoadingRows = styled.div`
 `
 
 export const loadingOpacityMixin = css<{ $loading: boolean }>`
-  filter: ${({ $loading }) => ($loading ? 'grayscale(1)' : 'none')};
-  opacity: ${({ $loading }) => ($loading ? '0.4' : '1')};
+  filter: ${({ $loading }) => ($loading ? 'grayscale(1)' : 'none')} !important;
```
zzmp, line 32: why is `!important` necessary?

src/lib/components/Swap/Input.tsx
```diff
@@ -39,6 +43,13 @@ export default function Input({ disabled }: InputProps) {
   // extract eagerly in case of reversal
   usePrefetchCurrencyColor(swapInputCurrency)
 
+  const routeIsLoading = useMemo(
```
zzmp, line 46: This should be `isLoading` or `isRouteLoading`. Starting a boolean with `is`/`was` makes it a clearer variable name.

src/lib/components/Swap/Input.tsx
```diff
@@ -1,10 +1,13 @@
 import { Trans } from '@lingui/macro'
+import { LoadingOpacityContainer } from 'components/Loader/styled'
```
zzmp, line 2: If this is a simple component, recreate it in `lib`, or move it to `lib`. Right now `lib` does not rely on _any_ components from the app, and adding one in will make it harder to decouple later on.

src/lib/components/Swap/TokenInput.tsx
```diff
@@ -70,17 +73,20 @@ export default function TokenInput({
...
       <TokenInputRow gap={0.5} onBlur={onBlur}>
-        <ThemedText.H2>
-          <ValueInput
-            value={amount}
-            onFocus={onFocus}
-            onChange={onChangeInput}
-            disabled={disabled || !currency}
-          ></ValueInput>
-        </ThemedText.H2>
+        <LoadingOpacityContainer $loading={!!loading}>
```
zzmp, line 80: nit: I prefer `Boolean(loading)` because it is more explicit

src/lib/components/Swap/Input.tsx
zzmp: If it is CSS, you can also just import the CSS and target it to child components based on their ReactNode. This would allow you to style the TokenInput without modifying TokenInput - it's also the strategy used for unapproved tokens.

zzmp (changes requested):

src/components/Loader/styled.tsx
ianlapham: Thought it was an issue with how the styles were being passed but looks like we dont need it.

src/lib/components/Swap/Input.tsx
ianlapham: Because its component decided to move into Loading component within lib

src/lib/components/Swap/Input.tsx
```diff
@@ -39,6 +43,13 @@ export default function Input({ disabled }: InputProps) {
   // extract eagerly in case of reversal
   usePrefetchCurrencyColor(swapInputCurrency)
 
+  const isRouteLoading = useMemo(
```
zzmp, line 46: Should this be `isTradeLoading`? Routes aren't really relevant in this file.

src/lib/components/Swap/Input.tsx
```diff
@@ -39,6 +43,13 @@ export default function Input({ disabled }: InputProps) {
...
+    () => TradeState.LOADING === tradeState || TradeState.SYNCING === tradeState,
+    [tradeState]
+  )
+  const independentField = useAtomValue(independentFieldAtom)
+  const showLoading = independentField === Field.OUTPUT && isRouteLoading
```
zzmp, line 51: For clarity, can this be rewritten as:

```
const isDependentField = independentField !== Field.INPUT
const isLoading = isDependentField && isRouteLoading
```

src/lib/components/Swap/Output.tsx
```diff
@@ -40,13 +42,22 @@ interface OutputProps {
...
   const [swapOutputAmount, updateSwapOutputAmount] = useSwapAmount(Field.OUTPUT)
   const [swapOutputCurrency, updateSwapOutputCurrency] = useSwapCurrency(Field.OUTPUT)
 
+  //loading status of the trade
+  const isRouteLoading = useMemo(
+    () => TradeState.LOADING === tradeState || TradeState.SYNCING === tradeState,
+    [tradeState]
+  )
+  const independentField = useAtomValue(independentFieldAtom)
+  const showLoading = independentField === Field.INPUT && isRouteLoading
```
zzmp, line 59: See comments on naming for `Input.tsx`. Note that you should be comparing `!== Field.OUTPUT` for the output file, to stay relevant.

src/lib/components/Swap/Toolbar/Caption.tsx
```diff
@@ -53,7 +53,7 @@ export function Trade({ trade }: { trade: InterfaceTrade<Currency, Currency, Tra
     const [a, b] = flip ? [outputAmount, inputAmount] : [inputAmount, outputAmount]
     const priceString = (!flip ? executionPrice : executionPrice?.invert())?.toSignificant(6)
 
-    const ratio = `1 ${a.currency.symbol} = ${priceString}} ${b.currency.symbol}`
+    const ratio = `1 ${a.currency.symbol} = ${priceString} ${b.currency.symbol}`
```
zzmp, line 56: Fix in a separate (tiny) PR if you can. No need for approval - just let it run the tests.

src/lib/components/Loader.tsx
```diff
@@ -0,0 +1,11 @@
+import styled, { css } from 'lib/theme'
+
+export const loadingOpacityMixin = css<{ $loading: boolean }>`
+  filter: ${({ $loading }) => ($loading ? 'grayscale(1)' : 'none')};
+  opacity: ${({ $loading }) => ($loading ? '0.4' : '1')};
+  transition: opacity 0.2s ease-in-out;
+`
+
+export const LoadingOpacityContainer = styled.div<{ $loading: boolean }>`
```
zzmp, line 9: I don't think this merits being a component. It seems more like a theme/css addition to me.

wdyt about moving it to `lib/css/loading.ts`, and only exporting the css. It would then be used by creating a new styled component out of whichever element needed it:

```
const LoadingH2 = themed<{ loading: boolean }>(ThemedText.H2)`
  ${loadingOpacityCss};
`
```

src/lib/components/Swap/Input.tsx
ianlapham: Yeah isTradeLoading seems better

src/lib/components/Loader.tsx
ianlapham: Yeah - this is probably better because it avoids us creating unneccesary container divs

ianlapham: one issue im having though is styled in our theme wont let us import css like this into a styled component

ianlapham: One argument for this though is that we end up writing styled components in every file that needs a loading state instead of referring to one component

ianlapham: Pushed CSS version for now though

src/lib/components/Swap/Input.tsx
```diff
@@ -1,17 +1,24 @@
...
 
 import Column from '../Column'
 import Row from '../Row'
 import TokenImg from '../TokenImg'
 import TokenInput from './TokenInput'
 
+const LoadingSpan = styled.span<{ $loading: boolean }>`
+  ${loadingOpacityCss};
+`
+
```
JFrankfurt, line 21: how does the $ prefix work here? and is the prop being used anywhere?

JFrankfurt: not sure how props get passed to css interpolations, but assumed they were not

ianlapham: we just cant use `loading` as its a reserved prop name, will give us TS errors

JFrankfurt: cool that this functions. TIL.

zzmp: Leave a comment that `loading` is reserved, and that's why you use `$loading`. Otherwise someone will waste half an hour trying to "fix" it in a month.

src/lib/components/Swap/Input.tsx
```diff
@@ -39,6 +47,14 @@ export default function Input({ disabled }: InputProps) {
   // extract eagerly in case of reversal
   usePrefetchCurrencyColor(swapInputCurrency)
 
+  const isTradeLoading = useMemo(
+    () => TradeState.LOADING === tradeState || TradeState.SYNCING === tradeState,
+    [tradeState]
+  )
+  const independentField = useAtomValue(independentFieldAtom)
+  const isDependentField = independentField !== Field.INPUT
```
zzmp, line 55: nit: You could inline this if you wanted: `const isDependentField = Field.INPUT !== useAtomValue(independentFieldAtom)`

src/lib/components/Swap/TokenInput.tsx
```diff
@@ -9,6 +10,10 @@ import { DecimalInput } from '../Input'
 import Row from '../Row'
 import TokenSelect from '../TokenSelect'
 
+const LoadingH2 = styled(ThemedText.H2)<{ $loading: boolean }>`
```
zzmp, line 13: Instead of a new `LoadingH2` element, this can be wrapped into the existing `ValueInput`.

Just update `ValueInput` to take `$loading`, and plop the css into there.

src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -81,9 +81,9 @@ export function useBestTrade(
     (debouncedOtherCurrency && otherCurrency && debouncedOtherCurrency !== otherCurrency)
 
   const syncing = isTradeDebouncing({
-    amounts: [amountFromLatestTrade, debouncedAmount],
-    indepdenentCurrencies: [currencyFromTrade, debouncedOtherCurrency],
-    dependentCurrencies: [otherCurrencyFromTrade, otherCurrency],
+    amounts: [amountFromLatestTrade, amountSpecified],
```
zzmp, line 84: What's going on here? Was this a mistake from a past PR? Seems out of place.

zzmp (approved):
nit: It _might_ be cleaner to do this by computing `isLoading` in `Swap/index.tsx`, and then passing a `loading` prop to `Input`, `Output`, and `Toolbar`.

src/lib/hooks/swap/useBestTrade.ts
ianlapham: Yeah this is a mistake - this logic updates the logic to match the current app which does make the loading work correctly. 

Without this we enter a perpetual loading state

PR #3238, opened 2022-02-04
fix(widgets): white accentText color on some buttons

JFrankfurt:
image

image

image

### DIFF
--- a/src/lib/components/ActionButton.tsx
+++ b/src/lib/components/ActionButton.tsx
@@ -74,7 +74,10 @@ export default function ActionButton({
   return (
     <Overlay update={Boolean(update)} flex align="stretch">
       <StyledButton color={color} disabled={disabled} onClick={update ? onUpdate : onClick}>
-        <ThemedText.TransitionButton buttonSize={update ? 'medium' : 'large'} color="currentColor">
+        <ThemedText.TransitionButton
+          buttonSize={update ? 'medium' : 'large'}
+          color={color === 'accent' && !disabled ? 'accentText' : 'currentColor'}
+        >
           {update ? update.action : children}
         </ThemedText.TransitionButton>
       </StyledButton>
--- a/src/lib/components/ActionButton.tsx
+++ b/src/lib/components/ActionButton.tsx
@@ -1,6 +1,6 @@
 import { AlertTriangle, Icon, LargeIcon } from 'lib/icons'
 import styled, { Color, css, keyframes, ThemedText } from 'lib/theme'
-import { ReactNode } from 'react'
+import { ReactNode, useMemo } from 'react'
 
 import Button from './Button'
 import Row from './Row'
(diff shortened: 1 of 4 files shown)


src/lib/theme/index.tsx
```diff
@@ -26,6 +26,7 @@ export const lightTheme: Colors = {
   dialog: '#FFFFFF',
 
   // text
+  accentText: 'hsl(0, 0%, 100%)',
```
zzmp, line 29: use #FFFFFF to be consistent with other definitions (and how it comes from figma)

src/lib/theme/theme.ts
```diff
@@ -9,6 +9,7 @@ export interface Colors {
 
   // text
   primary: string
+  accentText: string
```
zzmp, line 12: Call this `accent` - none of the other properties are suffixed with `text`.

zzmp: hmm...`accent` is taken. Maybe `contrast`?

zzmp (commented):
some nits, longer review later

src/lib/theme/theme.ts
JFrankfurt: this is specifically text to be overlayed over `accent`.

src/lib/theme/index.tsx
JFrankfurt: We *should* be using HSL everywhere. Hex is impossible to parse at a glance but HSL has human-readable info. Figma output is user-configurable, so you can read out hsl values wherever you would hex.

src/lib/theme/theme.ts
JFrankfurt: tbh I don't think we should make this change (@ppphhhiiilll )

black is a higher contrast text over the pink we are using, and I think we should make all text just be `readableColor(<backgroundColor>)`

src/lib/theme/index.tsx
zzmp: can you use #FFFFFF to be consistent here, and update _all_ values to use HSL in another PR?

src/lib/theme/theme.ts
zzmp: > this is specifically text to overlay `accent`.

In that case, can you call it `onAccent`?

src/lib/components/TokenSelect/TokenButton.tsx
```diff
@@ -36,7 +36,7 @@ interface TokenButtonProps {
 export default function TokenButton({ value, collapsed, disabled, onClick }: TokenButtonProps) {
   return (
     <StyledTokenButton onClick={onClick} empty={!value} color={value ? 'interactive' : 'accent'} disabled={disabled}>
```
zzmp, line 38: Can you pull all these colors out into a useMemo? Seems redundant and confusing with all the ternaries.

src/lib/components/ActionButton.tsx
```diff
@@ -71,10 +71,14 @@ export default function ActionButton({
   onUpdate,
   children,
 }: ActionButtonProps) {
+  const textColor = useMemo(
```
zzmp, line 74: Sorry for the churn but I don't know that `useMemo` adds any value here. Can't you just compute the textColor directly?

zzmp: Same with other `useMemo` uses.

src/lib/theme/index.tsx
```diff
@@ -40,16 +41,19 @@ export const lightTheme: Colors = {
...
-  accent: '#2172E5',
+  accent: darkThemeAccent,
   container: '#191B1F',
   module: '#2C2F36',
   interactive: '#40444F',
   outline: '#565A69',
   dialog: '#000000',
 
   // text
+  onAccentText: readableColor(darkThemeAccent),
```
zzmp, line 56: This should be onAccent, not onAccentText.
I'm more concerned with consistency, as this will be visible to external consumers.

zzmp (approved):
lgtm, just change `onAccentText` to `onAccent`.

src/lib/theme/index.tsx
JFrankfurt: are you sure `on` is clear? it sounds like a handler, and even in the context of a theme definition I find the name v confusing. including `text` at least makes it clear how the value will be applied.

PR #3241, opened 2022-02-04
fix: isolate infura

zzmp:
Prevents leaking infura keys to the lib


src/utils/switchToNetwork.ts
```diff
@@ -32,8 +33,8 @@ export async function switchToNetwork({ library, chainId }: SwitchNetworkArgumen
           {
             chainId: formattedChainId,
             chainName: info.label,
-            rpcUrls: [info.addNetworkInfo.rpcUrl],
-            nativeCurrency: info.addNetworkInfo.nativeCurrency,
+            rpcUrls: [INFURA_NETWORK_URLS[chainId]],
```
moodysalem, line 36: these are deliberately not the same

we don't want our infura urls in the user's metamask or wallet, because we can change the ids at any time. we should use one of the public endpoints provided by the network

src/constants/chainInfo.ts
```diff
@@ -83,10 +53,7 @@ export const CHAIN_INFO: ChainInfoMap = {
     infoLink: 'https://info.uniswap.org/#/',
     label: 'Ethereum',
     logoUrl: ethereumLogoUrl,
-    addNetworkInfo: {
```
ianlapham, line 86: is this `addNetworkInfo` for the interaction with metamask? Will this break if we remove it?

ianlapham, line 86: is this `addNetworkInfo` for the interaction with metamask? Will this break if we remove it?

ianlapham, line 86: is this `addNetworkInfo` for the interaction with metamask? Will this break if we remove it?

zzmp: Yes, but I've refactored the usage with metamask (see `switchToNetwork.ts`), so it should not break.

PR #3243, opened 2022-02-04
chore: rename web3-react-alpha

zzmp:
Renames widgets-web3-react/* to @widgets/web3-react/*.
npm treats nested packages as scoped, and requires scoped packages to begin with @.


JFrankfurt:
confirmed this works w/ yarn and npm?

zzmp:
> confirmed this works w/ yarn and npm?

works with yarn, haven't checked with npm

ianlapham (approved):

PR #3244, opened 2022-02-04
feat: bundle widgets

zzmp:


ianlapham (commented):
Nothing obvious to me but if anyone has more experience with rollups could be helpful

package.json
```diff
@@ -131,7 +146,8 @@
     "i18n:extract": "lingui extract --locale en-US",
     "i18n:compile": "yarn i18n:extract && lingui compile",
     "i18n:pseudo": "lingui extract --locale pseudo && lingui compile",
-    "postinstall": "yarn contracts:compile && yarn graphql:generate && yarn i18n:compile",
+    "prepare": "yarn contracts:compile && yarn graphql:generate && yarn i18n:compile",
```
JFrankfurt, line 149: will `prepare` fail w/o an existing `node_modules` dir, since `contracts:compile` references dependencies?

tsconfig.lib.json
```diff
@@ -2,20 +2,22 @@
...
       "types/*": ["../types/*"],
-      "utils/*": ["../utils/*"],
-    },
+      "utils/*": ["../utils/*"]
+    }
   },
-  "exclude": ["node_modules", "src/lib/**/*.test.*"],
-  "include": ["src/lib/**/*"]
+  "exclude": ["node_modules", "src/**/*.test.*"],
+  "include": ["src/lib/**/*.d.ts", "src/lib/index.tsx"]
```
justindomingue, line 22: is this redundant with `emitDeclarationOnly`?

rollup.config.ts
```diff
@@ -3,61 +3,125 @@
...
-    replace({ ...replacements, preventAssignment: true }),
-    url(), // imports files (including svgs) as data URIs
+    ...plugins,
+
+    // Source code transformation
+    url({ include: ASSET_EXTENSIONS.map((extname) => '**/*' + extname) }), // imports assets as data URIs
     svgr({ exportType: 'named', svgo: false }), // imports svgs as React components
-    sass(), // imports sass styles
-    typescript({ tsconfig: './tsconfig.lib.json', useTsconfigDeclarationDir: true }),
+    sass({ insert: true }), // imports inlined sass styles
```
justindomingue, line 95: is that for extensions?

tsconfig.lib.json
zzmp: No, this is a declaration file that must be included so that asset imports function (for CRA compatibility).
`src/lib/index.tsx` is not required, as it's included by the rollup toolchain, but including it allows you to run `tsc` as a standalone, so it's here for completeness.

src/lib/icons/index.tsx
```diff
@@ -1,6 +1,10 @@
-/* eslint-disable no-restricted-imports */
+import { ReactComponent as CheckIcon } from 'lib/assets/svg/check.svg'
+import { ReactComponent as ExpandoIcon } from 'lib/assets/svg/expando.svg'
+import { ReactComponent as LogoIcon } from 'lib/assets/svg/logo.svg'
+import { ReactComponent as SpinnerIcon } from 'lib/assets/svg/spinner.svg'
 import styled, { Color, css, keyframes } from 'lib/theme'
 import { FunctionComponent, SVGProps } from 'react'
+/* eslint-disable no-restricted-imports */
```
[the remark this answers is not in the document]
zzmp: `lib` has this in its eslint, so that all icon's are re-exported through this file and typed/styled the same way.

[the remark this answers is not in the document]
zzmp: All I did was target `react-feather` explicitly, instead of blanket disabling the rule.

rollup.config.ts
zzmp: `sass` is required for font imports, and `{ insert: true }` is required to bundle the font css in with the package (instead of requiring users to import both the lib and the css themselves).

package.json
```diff
@@ -210,6 +219,17 @@
     "styled-components": "^5.3.0",
     "tiny-invariant": "^1.2.0",
     "wcag-contrast": "^3.0.0",
+    "web3-react-alpha-core": "npm:@web3-react/core@8.0.16-alpha.0",
```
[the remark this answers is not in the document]
zzmp: `lib` (aka widgets) uses a newer version of `web-react`, so it needs to be renamed in order to co-exist with the version used by the app.

package.json
zzmp: yes, but it shouldn't be run until after install according to the docs: https://docs.npmjs.com/cli/v7/using-npm/scripts#life-cycle-scripts

JFrankfurt: >Runs on local `npm install` without any arguments

Am I misunderstanding this?

JFrankfurt: testing for myself

JFrankfurt: Okay, seems to work fine. I read those docs as `prepare` running before `yarn` but it definitely doesn't do that. 🤷

JFrankfurt (approved):

justindomingue (approved):
LGTM

PR #3245, opened 2022-02-05
feat(widgets): ensure passed locale is supported

JFrankfurt:
ensures the integrator passes only supported locales

exports `SUPPORTED_LOCALES`


zzmp:
Why should we crash the whole widget in this case instead of just falling back to the default en-US?

JFrankfurt:
we shouldn't

PR #3253, opened 2022-02-07
fix: update widget loading state detection

ianlapham:
- only show loading status, insufficient liquidity, or trade info if the user has typed in an amount

### DIFF
--- a/src/lib/components/Swap/Toolbar/index.tsx
+++ b/src/lib/components/Swap/Toolbar/index.tsx
@@ -1,5 +1,5 @@
 import { ALL_SUPPORTED_CHAIN_IDS } from 'constants/chains'
-import { useSwapAmount, useSwapInfo } from 'lib/hooks/swap'
+import { useIsAmountPopulated, useSwapInfo } from 'lib/hooks/swap'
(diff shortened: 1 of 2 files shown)


src/lib/components/Swap/Toolbar/index.tsx
```diff
@@ -42,11 +45,10 @@ export default function Toolbar({ disabled }: { disabled?: boolean }) {
       return <Caption.InsufficientBalance currency={trade.inputAmount.currency} />
     }
 
-    if (inputCurrency && outputCurrency) {
+    if (inputCurrency && outputCurrency && (swapInputAmount || swapOutputAmount)) {
```
zzmp, line 48: What you actually want here is whether there is a populated amount at all. Could you make a new small hook in `lib/hooks/swap` to do that? It only needs to look at `useAtomValue(amountAtom)`, so it should have less churn.

zzmp (changes requested):

src/lib/components/Swap/Toolbar/index.tsx
ianlapham: Yep makes sense

PR #3256, opened 2022-02-08
fix(widgets): fix broken unsupported network message

JFrankfurt:
Stops rendering the swap info updater on chains we don't support

image

### DIFF
--- a/src/lib/components/Swap/index.tsx
+++ b/src/lib/components/Swap/index.tsx
@@ -9,7 +9,7 @@ import useActiveWeb3React from 'lib/hooks/useActiveWeb3React'
 import useTokenList from 'lib/hooks/useTokenList'
 import { displayTxHashAtom } from 'lib/state/swap'
 import { SwapTransactionInfo, Transaction, TransactionType } from 'lib/state/transactions'
-import { useState } from 'react'
+import { useMemo, useState } from 'react'
 
 import Dialog from '../Dialog'
 import Header from '../Header'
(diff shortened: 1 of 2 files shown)


src/lib/components/Swap/index.tsx
```diff
@@ -47,20 +47,25 @@ export interface SwapProps {
 }
 
 export default function Swap(props: SwapProps) {
-  useTokenList(props.tokenList)
+  const list = useTokenList(props.tokenList)
```
zzmp, line 50: This sets the token list, it wasn't meant to _get_ it.

src/lib/components/Swap/index.tsx
```diff
@@ -47,20 +47,25 @@ export interface SwapProps {
...
-  const { active, account } = useActiveWeb3React()
+  const { active, account, chainId } = useActiveWeb3React()
   const [boundary, setBoundary] = useState<HTMLDivElement | null>(null)
 
   const [displayTxHash, setDisplayTxHash] = useAtom(displayTxHashAtom)
   const pendingTxs = usePendingTransactions()
   const displayTx = getSwapTx(pendingTxs, displayTxHash)
 
+  const onSupportedChain = useMemo(
+    () => list.reduce((acc, cur) => acc || cur.chainId === chainId, false),
```
zzmp, line 62: This is not necessarily true, because a token list loads asynchronously. It'll need three states - loading, supported, unsupported.

zzmp, line 62: This should use `some`, not `reduce`.

src/lib/components/Swap/index.tsx
```diff
@@ -47,20 +47,25 @@ export interface SwapProps {
...
+    [chainId, list]
+  )
+
   return (
     <SwapPropValidator {...props}>
-      <SwapInfoUpdater />
+      {onSupportedChain && <SwapInfoUpdater />}
```
zzmp, line 68: This doesn't need to be disabled on an unsupported chain, and it adds complexity. I'd prefer to leave it regardless.

zzmp (changes requested):

src/lib/components/Swap/index.tsx
```diff
@@ -47,20 +48,28 @@ export interface SwapProps {
...
 
   const [displayTxHash, setDisplayTxHash] = useAtom(displayTxHashAtom)
   const pendingTxs = usePendingTransactions()
   const displayTx = getSwapTx(pendingTxs, displayTxHash)
 
+  const onSupportedChain = useMemo(
+    () =>
+      chainId &&
+      ALL_SUPPORTED_CHAIN_IDS.includes(chainId) &&
+      list.reduce((acc, cur) => acc || cur.chainId === chainId, false),
```
zzmp, line 66: `list.some((token) => token.chainId === chainId)`

PR #3259, opened 2022-02-09
feat: routing tooltip for widget

ianlapham:


src/lib/components/Swap/RoutingTooltip/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+      </BaseBadge>
+    </TransparentBadgeWrapper>
+  )
+}
+
+export default function RoutingTooltip({ trade }: { trade: InterfaceTrade<Currency, Currency, TradeType> }) {
+  const routes: RoutingDiagramEntry[] = useMemo(() => getTokenPath(trade), [trade])
+
+  return (
+    <Tooltip icon={Info} placement="right">
```
zzmp, line 118: `placement="bottom"`

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+  padding: 0.25em;
+`
+
+const RouteRow = styled(Row)`
+  grid-template-columns: 1em 1.15em 1fr 1em;
+  min-width: 430px;
+`
+
+const RouteDetailsContainer = styled(Row)`
+  padding: 0.1rem 0.5rem;
```
zzmp, line 25: why rem?

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+  position: relative;
+`
+
+const ShortDottedLine = styled.div`
+  align-items: center;
+  display: flex;
+  opacity: 0.5;
+  overflow: hidden;
+  width: 71px;
```
zzmp, line 34: Why 71?

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+const RouteDetailsContainer = styled(Row)`
+  padding: 0.1rem 0.5rem;
+  position: relative;
+`
+
+const ShortDottedLine = styled.div`
+  align-items: center;
+  display: flex;
+  opacity: 0.5;
```
zzmp, line 32: Instead of setting the opacity, set the color appropriately (to match the color of the entries I think, using the theme)

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+  overflow: hidden;
+  width: 71px;
+  z-index: 1;
```
zzmp, line 35: Don't use literal z-index values. Use the constants from lib/theme/layer

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+  display: flex;
+  opacity: 0.5;
+  position: absolute;
+  width: calc(100% - 1em);
+  z-index: 1;
+`
+
+const DotColor = styled(DotLine)`
+  path {
+    stroke: ${({ theme }) => theme.secondary};
```
zzmp, line 49: how come this also needs opacity from DottedLine?

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+  }
+`
+
+const BaseBadge = styled(Row)`
+  background-color: ${({ theme }) => theme.outline};
+  border-radius: 0.5em;
```
zzmp, line 55: Should this be customizable? If so, use a multiplier with `theme.borderRadius`

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+  grid-gap: 0.375em;
+  padding: 0.25em 0.375em;
+  z-index: 2; // To cover the dotted line.
```
zzmp, line 58: use a layer const

src/lib/components/Swap/RoutingDiagram/index.tsx
zzmp: with multiple layers _in this file_ it may make sense to define local layers:

```
enum RoutingDiagramLayer {
  DOTTED_LINE = 1,
  BADGE =2,
  ...
}
```

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+const DotColor = styled(DotLine)`
+  path {
+    stroke: ${({ theme }) => theme.secondary};
+  }
+`
+
+const BaseBadge = styled(Row)`
+  background-color: ${({ theme }) => theme.outline};
+  border-radius: 0.5em;
+  grid-gap: 0.375em;
```
zzmp, line 56: Use the Row's prop, gap, instead of specifying it in css.

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+  display: grid;
+  grid-template-columns: 4.8125em 1fr;
+  width: 100%;
+`
+
+const StyledAutoRouterLabel = styled(ThemedText.Body1)`
+  color: #27ae60;
+  font-weight: 500;
+  line-height: 1rem;
+  @supports (-webkit-background-clip: text) and (-webkit-text-fill-color: transparent) {
```
zzmp, line 83: nit: newline before the @supports

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+`
+
+const DetailsRow = styled(Row)`
+  display: grid;
+  grid-template-columns: 4.8125em 1fr;
+  width: 100%;
+`
+
+const StyledAutoRouterLabel = styled(ThemedText.Body1)`
+  color: #27ae60;
```
zzmp, line 80: can this be a theme color?

zzmp, line 80: can this be a theme color?

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+  font-weight: 500;
+  line-height: 1rem;
+  @supports (-webkit-background-clip: text) and (-webkit-text-fill-color: transparent) {
+    background-image: linear-gradient(90deg, #2172e5 0%, #54e521 163.16%);
```
zzmp, line 84: these colors (or the whole gradient) should be moved to a const to make it clearer where it is defined

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+  return (
+    <Wrapper gap={0.75}>
+      <Row justify="space-between">
+        <Row gap={0.25}>
+          <StyledAutoRouterIcon />
+          <StyledAutoRouterLabel fontSize={14}>Auto Router</StyledAutoRouterLabel>
+        </Row>
+        <ThemedText.Body1 fontSize={14}>
+          Best routes via {routes.length} hop{routes.length > 1 ? 's' : ''}
```
zzmp, line 123: wrap in trans (there may be special syntax for a potential plural, check lingui docs)

zzmp: Should it be "Best route" (singular)?

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+export default function RoutingDiagram({ trade }: { trade: InterfaceTrade<Currency, Currency, TradeType> }) {
+  const routes: RoutingDiagramEntry[] = useMemo(() => getTokenPath(trade), [trade])
+
+  return (
+    <Wrapper gap={0.75}>
+      <Row justify="space-between">
+        <Row gap={0.25}>
+          <StyledAutoRouterIcon />
+          <StyledAutoRouterLabel fontSize={14}>Auto Router</StyledAutoRouterLabel>
```
zzmp, line 120: If you need to change the font size you should be using a different ThemedText. If there isn't an appropriate one ask Phil to make one or update the design to use one.

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+  z-index: 1;
+`
+
+const DotColor = styled(DotLine)`
+  path {
+    stroke: ${({ theme }) => theme.secondary};
+  }
+`
+
+const BaseBadge = styled(Row)`
```
zzmp, line 53: Can you move all of the badge logic to its own file? Do we use badges anywhere else in the interface? It feels like a complex enough component (just from styling) that I think it shouldn't just be inlined into this file.

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+import Rule from 'lib/components/Rule'
+import TokenImg from 'lib/components/TokenImg'
+import styled, { ThemedText } from 'lib/theme'
+import { useMemo } from 'react'
+import { InterfaceTrade } from 'state/routing/types'
+
+import { getTokenPath, RoutingDiagramEntry } from './utils'
+
+const Wrapper = styled(Column)`
+  padding: 0.25em;
```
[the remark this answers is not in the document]
ianlapham: Are you saying the entire header should be outside of the diagram component?

src/lib/components/Swap/RoutingDiagram/index.tsx
ianlapham: this just happened to be the specific pixels that resulted in 3 dots (to match the spec in figma)

an alternative approach is exporting an SVG that is only the 3 dots, or building th dots in html/css

src/lib/components/Swap/RoutingDiagram/index.tsx
ianlapham: do we want to start defining colors like we do in app? 

for example: green1, blue1, red1 etc

src/lib/components/Badge.tsx
```diff
@@ -0,0 +1,14 @@
+import styled from 'lib/theme'
+
+import Row from './Row'
+
+export const Badge = styled(Row)<{ borderRadius?: number; padding?: string }>`
+  background-color: ${({ theme }) => theme.outline};
+  border-radius: ${({ borderRadius }) => `${borderRadius ?? 0.5}em`};
+  padding: ${({ padding }) => padding ?? '0.25em 0.375em'};
+  z-index: 2;
```
zzmp, line 9: This should probably use OVERLAY from lib/theme/layer.ts. It should not be defined inline, it should use a Layer constant.

src/lib/components/Badge.tsx
```diff
@@ -0,0 +1,14 @@
...
+`
+
+export const BadgeDark = styled(Badge)`
```
zzmp, line 12: Instead of a separate BadgeDark, could you include `color` as a prop for `Badge`, and default it to `outline`: `background-color: ${({ theme, color = 'outline' }) => theme[color]};`? That way you don't need a separate BadgeDark.

src/lib/components/Badge.tsx
```diff
@@ -0,0 +1,14 @@
+import styled from 'lib/theme'
+
+import Row from './Row'
+
+export const Badge = styled(Row)<{ borderRadius?: number; padding?: string }>`
```
zzmp, line 5: Make this a default export.

src/lib/components/Swap/RoutingDiagram/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export interface RoutingDiagramEntry {
+  percent: Percent
+  path: [Currency, Currency, FeeAmount][]
+  protocol: Protocol
+}
+
+const V2_DEFAULT_FEE_TIER = 3000
+
+export const routerGradientCss = css`
```
zzmp, line 16: nit: I think this should be defined in `RoutingDiagram/index.tsx` where it is used. It could also just be inlined there. It's a one-off style, right?

src/lib/components/Swap/RoutingDiagram/index.tsx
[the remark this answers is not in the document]
zzmp: Will the header always be included with the diagram? I was thinking that the header/rule was specific to the use case of the tooltip, and so this should only export the diagram (in case it could be re-used elsewhere).

That said, I think this should be left as-is - we can always break them apart later.

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+`
+
+const DotColor = styled(DotLine)`
+  path {
+    stroke: ${({ theme }) => theme.secondary};
+  }
+`
+
+// Used to cutoff space between badges and dotted lines.
+const TransparentBadgeWrapper = styled(Badge)`
```
zzmp, line 53: If you include `color` as an optional prop for Badge, this can be expressed as a <Badge color="dialog" padding="0 4px" /> instead of a whole new component

src/lib/components/Swap/RoutingDiagram/index.tsx
zzmp: we don't yet have a large palette for widget. If you can't use the gradient, could this default to an existing color (like "primary")? Seems strange to have a single green header. The gradient makes more sense to me because it mirrors the colors used by the chips. Might be a phil question.

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+const DetailsRow = styled(Row)`
+  display: grid;
+  grid-template-columns: 4.8125em 1fr;
+  width: 100%;
+`
+
+const StyledAutoRouterLabel = styled(ThemedText.ButtonSmall)`
+  color: #27ae60;
+  font-weight: 500;
+  line-height: 1rem;
```
zzmp, line 67: Why `rem`? Could this be `em`?

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+  ${routerGradientCss}
+`
+
+const StyledAutoRouterIcon = styled(AutoRouterIcon)`
+  height: 0.875em;
+  width: 0.875em;
+  :hover {
+    filter: brightness(1.3);
```
zzmp, line 76: Could this be an onHover color, instead of a custom css filter? What is this meant to be?

src/lib/components/Swap/RoutingDiagram/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+  @supports (-webkit-background-clip: text) and (-webkit-text-fill-color: transparent) {
+    background-image: linear-gradient(90deg, #2172e5 0%, #54e521 163.16%);
```
zzmp, line 18: In the spec, the colors of the auto router label mirror the colors of the tokens being swapped. Is this supposed to use static colors instead? Seems strange to have this one thing be colorful, when everything else in the widget uses a very limited color palette.

zzmp (commented):
Mostly style nits, plus a refactor for Badge

src/lib/components/Swap/RoutingDiagram/index.tsx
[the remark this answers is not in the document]
ianlapham: Agreed can break apart if we want to use later ^^

src/lib/components/Swap/RoutingDiagram/utils.ts
ianlapham: yeah its one off here, could def just include in the index file

src/lib/components/Swap/RoutingDiagram/index.tsx
ianlapham: Can be em

- also i see what youre saying now for the palette, yes i think we can use primary here and if support is available use the gradient

src/lib/components/Swap/RoutingDiagram/index.tsx
ianlapham: i think we can just delete this because we're not using it as a link like we are in app?

src/lib/components/Swap/RoutingDiagram/utils.ts
ianlapham: i actually think this is a coincidence where the spec just happens to use ETH and MKR - i think based on past designs for tooltip we've created this mini branding for the auto router, which uses this specific gradient. Ill check with Phill though to confirm that

ianlapham: confirmed in slack

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+  z-index: ${Layer.UNDERLAYER};
+`
+
+const ShortDottedLine = styled(DottedLine)`
+  overflow: hidden;
+  position: relative;
+  width: 71px;
+`
+
+const DotColor = styled(DotLine)`
```
zzmp, line 46: nit: `DotColor` isn't very descriptive. Could this just be called `Dots`?

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+    -webkit-text-fill-color: transparent;
+  }
+`
+
+const StyledAutoRouterIcon = styled(AutoRouterIcon)`
+  height: 0.875em;
+  width: 0.875em;
+`
+
+function Pool({ currency0, currency1, feeAmount }: { currency0: Currency; currency1: Currency; feeAmount: FeeAmount }) {
```
zzmp, line 75: nit: currency0 is not very descriptive. Could these be called `baseCurrency`? `originCurrency`? `targetCurrency`? or even just `origin`/`target`? or `input`/`output`?

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+  return (
+    <Badge padding="0 4px" color="dialog">
+      <Badge gap={0.375}>
+        <Row>
+          <TokenImg token={currency0} />
+          <TokenImg token={currency1} />
+        </Row>
+        <ThemedText.Subhead2>{feeAmount / 10000}%</ThemedText.Subhead2>
```
zzmp, line 83: nit: type this as `10_000` to make it more readable (see https://v8.dev/features/numeric-separators).

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+const DetailsRow = styled(Row)`
+  display: grid;
+  grid-template-columns: 4.8125em 1fr;
+  width: 100%;
+`
+
+const StyledAutoRouterLabel = styled(ThemedText.ButtonSmall)`
+  color: ${({ theme }) => theme.primary};
+  font-weight: 500;
```
zzmp, line 60: Can this use an existing ThemedText to set font-weight, line-height, and color? I think only the gradient should be set here.

zzmp: I think it's similar to the ButtonSmall.

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+  line-height: 1em;
+
+  @supports (-webkit-background-clip: text) and (-webkit-text-fill-color: transparent) {
+    background-image: linear-gradient(90deg, #2172e5 0%, #54e521 163.16%);
+    -webkit-background-clip: text;
+    -webkit-text-fill-color: transparent;
+  }
+`
+
+const StyledAutoRouterIcon = styled(AutoRouterIcon)`
```
zzmp, line 70: If this were wrapped as an icon would this still be necessary? See other imported svgs in lib/icons.

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+          <RouteDetailsContainer justify="flex-start" flex>
+            <DottedLine>
+              <DotColor />
+            </DottedLine>
+            <DetailsRow>
+              <Badge padding="0 4px" color="dialog">
+                <Badge gap={0.375}>
+                  <ThemedText.ButtonSmall color="secondary">{route.percent.toSignificant(2)}%</ThemedText.ButtonSmall>
+                  <Badge padding="0.125em" borderRadius={0.25} color="module">
+                    <ThemedText.Badge color="secondary">{route.protocol.toUpperCase()}</ThemedText.Badge>
```
zzmp, line 123: This is not sized correctly in the vercel preview.

src/lib/components/Swap/RoutingDiagram/index.tsx
ianlapham: Updated to use primary as a backup

src/lib/components/Swap/RoutingDiagram/index.tsx
ianlapham: deleted

src/lib/components/Swap/RoutingDiagram/index.tsx
ianlapham: updated with Dots and better naming on container

src/lib/components/Swap/RoutingDiagram/index.tsx
ianlapham: i think origin/target is most accurate, updating with that

src/lib/components/Swap/RoutingDiagram/index.tsx
ianlapham: yep converted to icon

src/lib/components/Swap/RoutingDiagram/index.tsx
ianlapham: im wondering if the classname "badge" is causing this weird behvaior, for now i can fix with setting the size explicilty

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+      <Row justify="space-between">
+        <Row gap={0.25}>
+          <AutoRouter />
+          <StyledAutoRouterLabel color="primary" lineHeight={'16px'}>
+            <Trans>Auto Router</Trans>
+          </StyledAutoRouterLabel>
+        </Row>
+        <ThemedText.ButtonSmall>
+          <Trans>
+            Best route via {routes.length} <Plural value={routes.length} one="hop" other="hops" />
```
zzmp, line 102: I don't think nesting a Plural inside a Trans will yield valid translations. Instead, use:
```
          <Plural value={routes.length} one="Best route via 1 hop" other="Best route via # hops" />
```

zzmp: I also couldn't get this working with `one`, but had it working with `_1`, so please change to that.

zzmp (approved):
lgtm but make sure to update the `<Plural>` usage and test it locally for a singleton case before merging.

src/lib/components/Swap/RoutingDiagram/index.tsx
ianlapham: good catch, forgot to test this logic

Issue #3204, opened 2022-01-31
use localstorage to store a client identifier for analytics


moodysalem:
https://developers.google.com/analytics/devguides/collection/analyticsjs/cookies-user-id#using_localstorage_to_store_the_client_id

right now it looks like we aren't using any client identifier at all (i assumed that GA would do this automatically)
https://github.com/Uniswap/interface/blob/7aecf5d3980daa09069bb1847fa0cf57aa1f6bd3/src/components/analytics/index.ts#L6-L11

tinaszheng:
![image](https://user-images.githubusercontent.com/59578595/153129724-afd9bbc7-dc37-4a66-bd01-6463eaa6f627.png)

it looks like clientId is being sent correctly 🤔  is the problem that it is unique each time? [GA](https://developers.google.com/analytics/devguides/collection/analyticsjs/cookies-user-id) recommends that we persist clientIds through cookies instead of `localStorage`:

> Caution: Unlike cookies, localStorage is bound by the [same-origin policy](https://en.wikipedia.org/wiki/Same-origin_policy) . If parts of your site are on different subdomains, or if some pages use http and others pages use https, you cannot use localStorage to measure user activity between those pages. For this reason, cookies continues to be the officially recommended way to store the Client ID.

if so we'd just have to remove that option parameter:

`ReactGA.initialize(GOOGLE_ANALYTICS_ID)`

@moodysalem wdyt?

willhennessy:
Reading that excerpt, I think it might be okay to use local storage to track client ID.  The goal is to anonymously count clients using https://app.uniswap.org/*  so the subdomain and https will always be the same. We shouldn't have an issue based on that excerpt

PR #3264, opened 2022-02-09
feat: track google analytics clientIds

tinaszheng:
fixes https://github.com/Uniswap/interface/issues/3204

### DIFF
--- a/src/components/analytics/GoogleAnalyticsReporter.tsx
+++ b/src/components/analytics/GoogleAnalyticsReporter.tsx
@@ -1,6 +1,6 @@
 import useActiveWeb3React from 'hooks/useActiveWeb3React'
 import { useEffect } from 'react'
-import ReactGA, { ga } from 'react-ga'
+import ReactGA from 'react-ga'
 import { RouteComponentProps } from 'react-router-dom'
 import { getCLS, getFCP, getFID, getLCP, Metric } from 'web-vitals'
 
@@ -36,7 +36,7 @@ export default function GoogleAnalyticsReporter({ location: { pathname, search }
 
   useEffect(() => {
     // typed as 'any' in react-ga -.-
-    ga((tracker: any) => {
+    ReactGA.ga((tracker: any) => {
       if (!tracker) return
 
       const clientId = tracker.get('clientId')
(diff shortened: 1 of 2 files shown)


src/components/analytics/GoogleAnalyticsReporter.tsx
```diff
@@ -31,5 +33,15 @@ export default function GoogleAnalyticsReporter({ location: { pathname, search }
   useEffect(() => {
     ReactGA.pageview(`${pathname}${search}`)
   }, [pathname, search])
+
+  useEffect(() => {
+    // typed as 'any' in react-ga -.-
+    ga((tracker: any) => {
```
moodysalem, line 39: i think you can use `ReactGA.ga`

src/components/analytics/index.ts
```diff
@@ -1,12 +1,17 @@
 import ReactGA from 'react-ga'
 import { isMobile } from 'utils/userAgent'
 
+export const GOOGLE_ANALYTICS_CLIENT_ID_STORAGE_KEY = 'ga_client_id'
 const GOOGLE_ANALYTICS_ID: string | undefined = process.env.REACT_APP_GOOGLE_ANALYTICS_ID
+
+const storedClientId = window.localStorage.getItem(GOOGLE_ANALYTICS_CLIENT_ID_STORAGE_KEY)
```
moodysalem, line 7: how come we don't generate it here?

moodysalem: it's just a randomly generated id right? we could technically use something like [nanoid](https://www.npmjs.com/package/nanoid), but i understand why we might want to just use GA's built in client id generation

src/components/analytics/GoogleAnalyticsReporter.tsx
```diff
@@ -4,6 +4,8 @@ import ReactGA from 'react-ga'
 import { RouteComponentProps } from 'react-router-dom'
 import { getCLS, getFCP, getFID, getLCP, Metric } from 'web-vitals'
 
+import { GOOGLE_ANALYTICS_CLIENT_ID_STORAGE_KEY } from '.'
```
moodysalem, line 7: weird import, kinda circular

moodysalem (approved):
just nits

src/components/analytics/index.ts
tinaszheng: yeah we should use GA's format in case we ever decide to go down the cookies route :P

src/components/analytics/GoogleAnalyticsReporter.tsx
deauna: maybe

PR #3267, opened 2022-02-09
fix(widgets): complete etherscan link and stop timer on tx inclusion

JFrankfurt:


cosmos.config.json
```diff
@@ -5,5 +5,6 @@
   "webpack": {
     "configPath": "react-scripts/config/webpack.config",
     "overridePath": "cosmos.override.js"
-  }
+  },
+  "port": 5001
```
zzmp, line 9: revert

ianlapham, line 9: Should we revert this?

src/lib/components/Swap/Status/StatusDialog.tsx
```diff
@@ -68,12 +70,18 @@ interface TransactionStatusProps {
...
 function TransactionStatus({ tx, onClose }: TransactionStatusProps) {
+  const { chainId } = useActiveWeb3React()
+
   const Icon = useMemo(() => {
     return tx.receipt?.status ? CheckCircle : Spinner
   }, [tx.receipt?.status])
   const heading = useMemo(() => {
     return tx.receipt?.status ? <Trans>Transaction submitted</Trans> : <Trans>Transaction pending</Trans>
   }, [tx.receipt?.status])
+  const etherscanLink = useMemo(
```
zzmp, line 81: I thought you'd made an `Etherscan` component to abstract this away, but I was wrong. I think that'd be a good thing to have, but can be done in another PR.

ianlapham (approved):
Looks good aside from port change

src/lib/components/Swap/Status/StatusDialog.tsx
```diff
@@ -68,12 +70,18 @@ interface TransactionStatusProps {
...
+    () => getExplorerLink(chainId || 1, tx.info.response.hash, ExplorerDataType.TRANSACTION),
```
zzmp, line 82: 1 should be SupportedChainId.MAINNET. Even better would be if it were defaulted by getExplorerLink.

src/lib/components/Swap/Status/StatusDialog.tsx
```diff
@@ -33,7 +35,7 @@ function ElapsedTime({ tx }: { tx: Transaction<SwapTransactionInfo> }) {
         setElapsedMs(Date.now() - tx.addedTime)
       }
     },
-    elapsedMs === tx.info.response.timestamp ? null : 1000
+    tx.receipt ? null : 1000
```
zzmp, line 38: L33 should also be updated to reflect the actual time as reported by the chain.

cosmos.config.json
JFrankfurt: I'm adding this to help out people like @willhennessy who have issues w/ port 5000. I just got a new computer and ran into it too.

src/lib/components/Swap/Status/StatusDialog.tsx
```diff
@@ -23,17 +25,22 @@ const TransactionRow = styled(Row)`
   flex-direction: row-reverse;
 `
 
+const Link = styled(EtherscanLink)`
```
zzmp, line 28: rm: This css is already included in EtherscanLink

src/lib/components/Swap/Status/StatusDialog.tsx
```diff
@@ -23,17 +25,22 @@ const TransactionRow = styled(Row)`
...
+`
+
 function ElapsedTime({ tx }: { tx: Transaction<SwapTransactionInfo> }) {
   const [elapsedMs, setElapsedMs] = useState(0)
   useInterval(
     () => {
-      if (tx.info.response.timestamp) {
-        setElapsedMs(tx.info.response.timestamp - tx.addedTime)
+      if (tx.receipt && tx.info.response.timestamp) {
+        setElapsedMs(tx.addedTime - tx.info.response.timestamp)
```
zzmp, line 37: wouldn't this be a negative number?

src/lib/components/Swap/Status/StatusDialog.tsx
```diff
@@ -82,9 +85,9 @@ function TransactionStatus({ tx, onClose }: TransactionStatusProps) {
       </StatusHeader>
       <TransactionRow flex>
         <ThemedText.ButtonSmall>
-          <EtherscanA href="//etherscan.io" target="_blank">
+          <Link type={ExplorerDataType.TRANSACTION} data={tx.info.response.hash}>
```
zzmp, line 88: Use `EtherscanA` directly

zzmp (changes requested):

src/lib/components/Swap/Status/StatusDialog.tsx
JFrankfurt: oh, tricky bug. Wasn't showing up because the null below halts the timer at the last `Date.now() - tx.addedTime` and it *looks* correct because this block doesn't run. Good catch.

PR #3275, opened 2022-02-10
feat: connect wallet ux

zzmp:
- Updates language, spacing, and icon
- Adds an onConnectWallet prop that is triggered when clicking "Connect your wallet"


ianlapham (approved):
very cool - can trigger wallet flow from widget now

PR #3279, opened 2022-02-10
fix: adjust max button for native currencies

ianlapham:


src/lib/components/Swap/Input.tsx
```diff
@@ -59,12 +60,15 @@ export default function Input({ disabled }: InputProps) {
   //TODO(ianlapham): mimic logic from app swap page
   const mockApproved = true
 
+  // account for gas needed if using max on native token
+  const maxInputAmount = useMemo(() => maxAmountSpend(balance), [balance])
```
zzmp, line 64: nit: we're in Input.tsx, just call this `maxAmount`

zzmp (approved):

PR #3286, opened 2022-02-10
feat: focus input on token select

zzmp:


src/lib/components/Swap/TokenInput.tsx
```diff
@@ -76,6 +76,15 @@ export default function TokenInput({
     }
   }, [])
 
+  const input = useRef<HTMLInputElement>(null)
+  const onSelect = useCallback(
+    (currency: Currency) => {
+      onChangeCurrency(currency)
+      setTimeout(() => input.current?.focus(), 0)
```
zzmp, line 83: Necessary to defer this by one render loop.
If it focuses synchronously, then it will fail, as the element is not focusable while the TokenSelect dialog is still open.

- Can you think of a better way to do this?
- If not, should I leave this rationale as a comment?

JFrankfurt: smells like it might be possible to accomplish this w/ `useLayoutEffect`

JFrankfurt: maybe use a layout effect that fires focus when a new currency is selected?

zzmp: it should be in response to user interaction, and a new currency might be selected if the chain changes (eg from defaulting). I think it needs to be triggered from `onSelect`.

JFrankfurt (approved):

PR #3287, opened 2022-02-10
feat: focus and hover hooks

zzmp:
Uses native listeners for focus and hover instead of adding boilerplate for onFocus/onBlur/onMouseEnter/onMouseLeave. This implementation is slightly more complex, but the usages are significantly simpler.


src/lib/components/Swap/Settings/MaxSlippageSelect.tsx
```diff
@@ -33,30 +33,28 @@ const Custom = styled(BaseButton)<{ selected: boolean }>`
...
 `
 
-interface OptionProps extends Partial<TooltipHandlers> {
+interface OptionProps {
   wrapper: typeof Button | typeof Custom
   selected: boolean
   onSelect: () => void
   icon?: ReactNode
+  tabIndex?: number
+  children: ReactNode
```
JFrankfurt, line 42: PropsWithChildren

JFrankfurt (approved):

PR #3294, opened 2022-02-11
feat: optimize client side SOR for widgets 

ianlapham:
- goal is to reduce latency while using client side SOR package 
- bump up the min distribution % used in routes to multiples of 5
- can see logic here https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/alpha-router.ts#L295

### DIFF
--- a/src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
+++ b/src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
@@ -10,7 +10,11 @@ import { getClientSideQuote } from './clientSideSmartOrderRouter'
 import { useRoutingAPIArguments } from './useRoutingAPIArguments'
 
 const protocols: Protocol[] = [Protocol.V2, Protocol.V3]
-const config = { protocols }
+
+// The following configs are used to reduce latency when using the client side SOR
+const MIN_DISTRIBUTION_PERCENT = 5
+
+const config = { protocols, distributionPercent: MIN_DISTRIBUTION_PERCENT }
 
 export default function useClientSideSmartOrderRouterTrade<TTradeType extends TradeType>(
   tradeType: TTradeType,


src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -10,7 +11,16 @@ import { getClientSideQuote } from './clientSideSmartOrderRouter'
 import { useRoutingAPIArguments } from './useRoutingAPIArguments'
 
 const protocols: Protocol[] = [Protocol.V2, Protocol.V3]
-const config = { protocols }
+
+// The following configs are used to reduce latency when using the client side SOR
```
zzmp, line 15: Instead of making this a const, in-line it in the config object, and just add a comment to every line of the config explaining why you're setting it as you are and what it does, so it's easier to reason through in the future.

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -10,7 +11,16 @@ import { getClientSideQuote } from './clientSideSmartOrderRouter'
...
+const MIN_DISTRIBUTION_PERCENT = 5
+
+const config = {
```
zzmp, line 18: Should you also be setting `maxSwapsPerPath` and/or `maxSplits`?

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -10,7 +11,16 @@ import { getClientSideQuote } from './clientSideSmartOrderRouter'
 import { useRoutingAPIArguments } from './useRoutingAPIArguments'
 
 const protocols: Protocol[] = [Protocol.V2, Protocol.V3]
-const config = { protocols }
+
+// The following configs are used to reduce latency when using the client side SOR
+const MIN_DISTRIBUTION_PERCENT = 5
```
zzmp, line 16: This seems to be a regression from the defaults for some chains. If this defaults to 5, maybe we should just leave it?

See https://github.com/Uniswap/smart-order-router/blob/b19ebcb3f3e2b6b10a8021884f5336c8735ba8a5/src/routers/alpha-router/config.ts#L4

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -10,7 +11,16 @@ import { getClientSideQuote } from './clientSideSmartOrderRouter'
...
+
+const config = {
+  protocols,
+  distributionPercent: MIN_DISTRIBUTION_PERCENT,
+  v3PoolProvider: StaticV3SubgraphProvider,
```
zzmp, line 21: This worries me a little bit:

https://github.com/Uniswap/smart-order-router/blob/b19ebcb3f3e2b6b10a8021884f5336c8735ba8a5/src/providers/v2/static-subgraph-provider.ts#L49-L50

zzmp (changes requested):
Can you talk to someone about how the defaults were chosen (https://github.com/Uniswap/smart-order-router/blob/309e6f6603984d3b5aef0733b0cfaf129c29f602/src/routers/alpha-router/config.ts), and figure out which numbers there you can turn down?

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
ianlapham: I also learned that teh hoardcoded provider only helps on the first load, so probably not worth the change here

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -9,8 +9,13 @@ import useActiveWeb3React from '../useActiveWeb3React'
 import { getClientSideQuote } from './clientSideSmartOrderRouter'
 import { useRoutingAPIArguments } from './useRoutingAPIArguments'
 
-const protocols: Protocol[] = [Protocol.V2, Protocol.V3]
-const config = { protocols }
+const config = {
+  // Limit to only V2 and V3.
+  protocols: [Protocol.V2, Protocol.V3],
+
+  // Used to reduce latency while SOR is used client side. Default is 5
```
zzmp, line 16: Default is 5 for MAINNET, but 10 for OPTIMISM and 25 for ARBITRUM, so this may end up slowing down L2s. Can this be chain-specific? Should OPTIMISM stay at 10, or should we bump all L2s to 25?

zzmp (changes requested):

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
zzmp: Instead of putting the default here, you should link to the code where it is defined (in case it changes in the future). Make sure to use a stable link, like https://github.com/Uniswap/smart-order-router/blob/309e6f6603984d3b5aef0733b0cfaf129c29f602/src/routers/alpha-router/config.ts#L83.

zzmp: Also, say what it does to reduce latency (this can be somewhat copypasta from the docs, as it's just more documentation), eg:

`Increase the minimum percentage of the input token to use for each route in a split route.`

(Original docs: https://github.com/Uniswap/smart-order-router/blob/309e6f6603984d3b5aef0733b0cfaf129c29f602/src/routers/alpha-router/alpha-router.ts#L260)

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -9,8 +10,30 @@ import useActiveWeb3React from '../useActiveWeb3React'
 import { getClientSideQuote } from './clientSideSmartOrderRouter'
 import { useRoutingAPIArguments } from './useRoutingAPIArguments'
 
-const protocols: Protocol[] = [Protocol.V2, Protocol.V3]
-const config = { protocols }
+const DistributionPercents: { [key: number]: number } = {
+  [ChainId.MAINNET]: 10,
```
zzmp, line 14: Comment on this line with the rationale

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -9,8 +10,30 @@ import useActiveWeb3React from '../useActiveWeb3React'
...
+function getConfig(chainId: ChainId | undefined) {
+  return {
+    // Limit to only V2 and V3.
+    protocols: [Protocol.V2, Protocol.V3],
+
+    /**
+     * Reduces client-side latency by increasing the minimum percentage of the input token to use for each route in a split route while SOR is used client-side.
+     * Defaults are defined in https://github.com/Uniswap/smart-order-router/blob/309e6f6603984d3b5aef0733b0cfaf129c29f602/src/routers/alpha-router/config.ts#L83.
+     */
+    distributionPercent: chainId
```
zzmp, line 32: Simplify as chainId && DistributionPercents[chainId] ?? DEFAULT...

zzmp (approved):

PR #3172, opened 2022-01-21
refactor: isolate approval callback hooks

zzmp:
Decouples approval logic from transactions, by making "hasPendingApproval" lookups use a callback hook instead of coupled logic. This is necessary to allow the widget to use the same approval codepaths, as the widget does not yet track transactions, and will not do so using redux, as the app does.

PR #3301, opened 2022-02-11
feat(widgets): support wrapping native assets

JFrankfurt:
this adds support for native asset wrapping within the existing framework for swap transactions


src/lib/components/Swap/SwapButton.tsx
```diff
@@ -76,6 +77,12 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
     })
   }, [addTransaction, getApproval])
 
+  const {
```
zzmp, line 80: Can you move this into useSwapInfo, and return a swap type enum saying if it's a swap, wrap, or unwrap? That way, the toolbar and swap button don't need to both call useWrapCallback - it's all taken care of in one place.

src/lib/components/Swap/SwapButton.tsx
```diff
@@ -153,6 +160,20 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
       })
   }, [addTransaction, independentField, inputCurrencyAmount, outputCurrencyAmount, setDisplayTxHash, swapCallback])
 
+  if (wrapType !== WrapType.NOT_APPLICABLE) {
```
zzmp, line 163: The same Action button needs to be rendered regardless of wrap/swap in order for css transitions to function for the button. Instead of returning early, set the props to then be passed to the ActionButton.

src/lib/components/Swap/SwapButton.tsx
zzmp: I think doing so would also allow you to short-circuit a fetch if only a wrap is needed.

src/lib/components/Swap/SwapButton.tsx
```diff
@@ -1,6 +1,7 @@
 import { Trans } from '@lingui/macro'
 import { Token, TradeType } from '@uniswap/sdk-core'
 import { useERC20PermitFromTrade } from 'hooks/useERC20Permit'
+import useWrapCallback, { WrapErrorText, WrapType } from 'hooks/useWrapCallback'
```
zzmp, line 4: This will need to be refactored to return a receipt so that it can be tracked as a pending transaction. This is also not currently usable by the widget because it is coupled to app state.

zzmp: See #3172

src/lib/components/ActionButton.tsx
```diff
@@ -65,7 +65,7 @@ export interface ActionButtonProps {
   color?: Color
   disabled?: boolean
   action?: Action
-  onClick: () => void
+  onClick?: () => void
```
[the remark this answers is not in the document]
JFrankfurt: How about setting the type to match native button + custom behavior?

```
type BaseProps = {
  color?: Color
  disabled?: boolean
  action?: Action
}

export type ActionButtonProps = BaseProps & Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, keyof BaseProps>

```

src/hooks/useWrapCallback.tsx
```diff
@@ -19,33 +18,6 @@ export enum WrapType {
...
-    case WrapInputError.ENTER_WRAPPED_AMOUNT:
-      return <Trans>Enter {wrapped?.symbol} amount</Trans>
-
-    case WrapInputError.INSUFFICIENT_NATIVE_BALANCE:
-      return <Trans>Insufficient {native?.symbol} balance</Trans>
-    case WrapInputError.INSUFFICIENT_WRAPPED_BALANCE:
-      return <Trans>Insufficient {wrapped?.symbol} balance</Trans>
-  }
-}
-
```
JFrankfurt, line 48: removed to `src/lib/hooks/swap/useWrapCallback.tsx`

src/lib/components/ActionButton.tsx
```diff
@@ -45,13 +45,13 @@ const actionCss = css`
...
-export const Overlay = styled(Row)<{ action?: boolean }>`
+export const Overlay = styled(Row)<{ showAction: boolean }>`
   border-radius: ${({ theme }) => theme.borderRadius}em;
   flex-direction: row-reverse;
   min-height: 3.5em;
   transition: padding 0.25s ease-out;
 
-  ${({ action }) => action && actionCss}
+  ${({ showAction }) => (showAction ? actionCss : '')}
 `
```
JFrankfurt, line 55: react was complaining about this--smelled like a styled-components issue?

src/lib/components/Swap/Status/StatusDialog.tsx
```diff
@@ -12,7 +12,7 @@ import { ExplorerDataType } from 'utils/getExplorerLink'
 import ActionButton from '../../ActionButton'
 import Column from '../../Column'
 import Row from '../../Row'
-import Summary from '../Summary'
+import SwapSummary from '../Summary'
```
JFrankfurt, line 15: renamed to help make clear the behavior below where this is only used for `SWAP` transaction types. (not wrap or approvals)

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -48,9 +49,13 @@ export default function useClientSideSmartOrderRouterTrade<TTradeType extends Tr
     quoteResult: undefined,
     error: undefined,
   })
+  const { type: wrapType } = useWrapCallback()
 
   // When arguments update, make a new call to SOR for updated quote
   useEffect(() => {
+    if (wrapType !== WrapType.NOT_APPLICABLE) {
+      return
+    }
```
JFrankfurt, line 58: this short-circuits routing calculations when the user is wrapping

src/lib/state/transactions.ts
```diff
@@ -37,7 +38,14 @@ export interface OutputSwapTransactionInfo extends SwapTransactionInfo {
...
 
-export type TransactionInfo = ApprovalTransactionInfo | SwapTransactionInfo
+export interface WrapTransactionInfo extends BaseTransactionInfo {
+  type: TransactionType.WRAP
+  unwrapped: boolean
+  currencyAmountRaw: string
+  chainId?: number
+}
+
+export type TransactionInfo = ApprovalTransactionInfo | SwapTransactionInfo | WrapTransactionInfo
```
JFrankfurt, line 48: copied from app

src/lib/components/Swap/Status/StatusDialog.tsx
ianlapham: Fan of this - it always confused me what this was

src/lib/components/Swap/Toolbar/Caption.tsx
```diff
@@ -25,6 +26,16 @@ function Caption({ icon: Icon = AlertTriangle, caption }: CaptionProps) {
   )
 }
 
+export function WrapCurrency({ loading, wrapType }: { loading: boolean; wrapType: WrapType.UNWRAP | WrapType.WRAP }) {
+  const WrapText = useCallback(() => {
+    if (wrapType === WrapType.WRAP) {
+      return loading ? <Trans>Wraping native currency.</Trans> : <Trans>Wrap native currency.</Trans>
```
ianlapham, line 32: change to "Wrapping"

src/lib/components/Swap/Toolbar/Caption.tsx
```diff
@@ -25,6 +26,16 @@ function Caption({ icon: Icon = AlertTriangle, caption }: CaptionProps) {
...
+    }
+    return loading ? <Trans>Unwraping native currency.</Trans> : <Trans>Unwrap native currency.</Trans>
```
ianlapham, line 34: "Unwrapping"

src/lib/components/Swap/index.tsx
```diff
@@ -31,7 +31,7 @@ export type DefaultAddress = string | { [chainId: number]: string | 'NATIVE' } |
 function getSwapTx(txs: { [hash: string]: Transaction }, hash?: string): Transaction<SwapTransactionInfo> | undefined {
```
ianlapham, line 31: should we change the name of this function? If its also containing wraps now?

src/lib/hooks/swap/useWrapCallback.tsx
```diff
@@ -0,0 +1,132 @@
...
+    independentField,
+    [Field.INPUT]: inputCurrency,
+    [Field.OUTPUT]: outputCurrency,
+  } = useAtomValue(swapAtom)
+
+  const wrapType = useMemo(() => {
+    if (!inputCurrency || !outputCurrency || !chainId) {
+      return WrapType.NOT_APPLICABLE
+    }
+    const weth = WRAPPED_NATIVE_CURRENCY[chainId]
```
ianlapham, line 68: should we change this from `weth` as we support more chains without ETH as native currency? (MATIC) for example

ianlapham (commented):
Some small requests

src/lib/components/Swap/Toolbar/Caption.tsx
JFrankfurt: 🙈 thanks for catching

src/lib/hooks/swap/useWrapCallback.tsx
JFrankfurt: I think so. This is the pattern currently used in app, so I was a little hesitant to change.

src/lib/components/Swap/Status/StatusDialog.tsx
zzmp: Should you also rename the file?

src/lib/components/Swap/SwapButton.tsx
```diff
@@ -167,17 +172,52 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
       })
   }, [addTransaction, inputCurrencyAmount, outputCurrencyAmount, setDisplayTxHash, swapCallback, tradeType])
 
+  const ButtonText = useCallback(() => {
```
zzmp, line 175: Never seen a useCallback used to make a react component. Why this over `useMemo`?

src/lib/components/Swap/Toolbar/Caption.tsx
```diff
@@ -25,6 +26,16 @@ function Caption({ icon: Icon = AlertTriangle, caption }: CaptionProps) {
...
+export function WrapCurrency({ loading, wrapType }: { loading: boolean; wrapType: WrapType.UNWRAP | WrapType.WRAP }) {
+  const WrapText = useCallback(() => {
+    if (wrapType === WrapType.WRAP) {
+      return loading ? <Trans>Wrapping native currency.</Trans> : <Trans>Wrap native currency.</Trans>
+    }
+    return loading ? <Trans>Unwrapping native currency.</Trans> : <Trans>Unwrap native currency.</Trans>
+  }, [loading, wrapType])
+
+  return <Caption icon={Info} caption={<WrapText />} />
+}
```
zzmp, line 38: newline between exports

src/lib/components/Swap/index.tsx
```diff
@@ -28,10 +28,13 @@ import Toolbar from './Toolbar'
 
 export type DefaultAddress = string | { [chainId: number]: string | 'NATIVE' } | 'NATIVE'
 
-function getSwapTx(txs: { [hash: string]: Transaction }, hash?: string): Transaction<SwapTransactionInfo> | undefined {
+function getTransactionFromMap(
+  txs: { [hash: string]: Transaction },
+  hash?: string
+): Transaction<SwapTransactionInfo> | undefined {
```
zzmp, line 34: Should there be a separate `WrapTransactionInfo` type?

src/lib/state/transactions.ts
```diff
@@ -37,7 +38,14 @@ export interface OutputSwapTransactionInfo extends SwapTransactionInfo {
   maximumInputCurrencyAmount: string
 }
 
-export type TransactionInfo = ApprovalTransactionInfo | SwapTransactionInfo
+export interface WrapTransactionInfo extends BaseTransactionInfo {
+  type: TransactionType.WRAP
+  unwrapped: boolean
```
zzmp, line 43: Is `unwrapped` copied from the app? I'd rather have this be `wrapType`, with the options being `WRAP | UNWRAP`.

src/lib/hooks/swap/useWrapCallback.tsx
```diff
@@ -0,0 +1,137 @@
...
+import tryParseCurrencyAmount from 'lib/utils/tryParseCurrencyAmount'
+import { useCallback, useMemo, useState } from 'react'
+
+import { WRAPPED_NATIVE_CURRENCY } from '../../../constants/tokens'
+import useActiveWeb3React from '../useActiveWeb3React'
+import { useCurrencyBalances } from '../useCurrencyBalance'
+import useNativeCurrency from '../useNativeCurrency'
+
+export enum WrapInputError {
+  NO_ERROR, // must be equal to 0 so all other errors are truthy
```
zzmp, line 15: you should explicitly set this to `= 0` if this is the case.

src/lib/components/Swap/Toolbar/index.tsx
```diff
@@ -25,7 +26,7 @@ export default function Toolbar({ disabled }: { disabled?: boolean }) {
   } = useSwapInfo()
   const isRouteLoading = state === TradeState.SYNCING || state === TradeState.LOADING
   const isAmountPopulated = useIsAmountPopulated()
-
+  const { type: wrapType, loading: wrapLoading } = useWrapCallback()
```
zzmp, line 29: I don't believe the state from the hook will be shared by multiple invocations of the hook. You'd need an atom to accomplish that.

zzmp (changes requested):
I don't think the Toolbar and ActionButton usages of the useWrapCallback will share the loading state, as intended. I think you need an atom to coordinate the shared state.

ianlapham (commented):
Looks good assuming @zzmp comments addressed

src/lib/components/Swap/SwapButton.tsx
JFrankfurt: useMemo gives you a value, useCallback gives you a function. You need the function call for the component constructor. 

I've **also** never seen useCallback used this way, but it... feels right? I usually see this pattern used w/o the hook to define a locally scoped function component, but I think that's strictly worse, right?


image

src/lib/components/Swap/Status/StatusDialog.tsx
JFrankfurt: I chose not to because it's already in the `components/Swap/Summary` directory. Making it `components/Swap/SwapSummary` feels weirdly repetitive and none of the other components follow that pattern. I do think changing the import could help though! I'll do that.

src/lib/hooks/swap/useWrapCallback.tsx
JFrankfurt: copied from app. I really dislike this usage of the 0 index as a falsy value :|

src/lib/state/transactions.ts
JFrankfurt: yeah, this is copied over like this for consistency.

src/lib/components/Swap/Toolbar/Caption.tsx
```diff
@@ -25,21 +26,37 @@ function Caption({ icon: Icon = AlertTriangle, caption }: CaptionProps) {
   )
 }
 
+export function WrapCurrency({ loading, wrapType }: { loading: boolean; wrapType: WrapType.UNWRAP | WrapType.WRAP }) {
+  const WrapText = useCallback(() => {
+    if (wrapType === WrapType.WRAP) {
```
zzmp, line 31: I prefer to make this explicit, so that you check against both types. A good way to do this and have tsc ensure completeness is to use switch:
```
switch (wrapType) {
  case WrapType.WRAP:
    ...
  case WrapType.UNWRAP:
    ...
}
```

I prefer to do it this way because WrapType is an enum, not a boolean, so there's not a guarantee (in code) that there are only two cases.

src/lib/components/Swap/Toolbar/Caption.tsx
zzmp: all of the "simple" captions are grouped up at the top. put this one at the bottom, above or below Trade, so the files easier to grok on first glance.

src/lib/components/Swap/Toolbar/Caption.tsx
zzmp: Just realized there _is_ a third type to WrapType, but you've already excluded it in your typing on L29.

src/lib/hooks/swap/useWrapCallback.tsx
zzmp: you should still explicitly set this to 0, as the comment indicates that you are relying on that value.

src/lib/hooks/swap/useWrapCallback.tsx
```diff
@@ -0,0 +1,153 @@
...
+  ENTER_NATIVE_AMOUNT,
+  ENTER_WRAPPED_AMOUNT,
+  INSUFFICIENT_NATIVE_BALANCE,
+  INSUFFICIENT_WRAPPED_BALANCE,
+}
+
+export function WrapErrorText({ wrapInputError }: { wrapInputError: WrapInputError }) {
```
zzmp, line 23: If this is only used in the toolbar, I think it should live in the Toolbar/Caption file.

src/lib/hooks/swap/useWrapCallback.tsx
```diff
@@ -0,0 +1,153 @@
...
+    if (inputCurrency.isNative && wrappedNativeCurrency.equals(outputCurrency)) {
+      return WrapType.WRAP
+    }
+    if (wrappedNativeCurrency.equals(inputCurrency) && outputCurrency.isNative) {
+      return WrapType.UNWRAP
+    }
+    return WrapType.NOT_APPLICABLE
+  }, [chainId, inputCurrency, outputCurrency])
+
+  const isExactIn: boolean = independentField === Field.INPUT
```
zzmp, line 89: unnecessary typing

src/lib/hooks/swap/useWrapCallback.tsx
```diff
@@ -0,0 +1,153 @@
...
+  )
+  const currencyBalances = useMemo(
+    () => ({
+      [Field.INPUT]: relevantTokenBalances[0],
+      [Field.OUTPUT]: relevantTokenBalances[1],
+    }),
+    [relevantTokenBalances]
+  )
+
+  const hasInputAmount = Boolean(parsedAmount?.greaterThan('0'))
```
zzmp, line 108: This needs a Boolean wrapper?

src/lib/hooks/swap/useWrapCallback.tsx
```diff
@@ -0,0 +1,153 @@
...
+  }, [hasInputAmount, setWrapState, sufficientBalance, wrapType])
+
+  const callback = useCallback(async () => {
+    if (
+      !wrappedNativeCurrencyContract ||
+      !sufficientBalance ||
+      !parsedAmountIn ||
+      wrapType === WrapType.NOT_APPLICABLE
+    ) {
+      return Promise.reject()
```
zzmp, line 134: Include a rejection reason for troubleshooting

src/lib/hooks/swap/useWrapCallback.tsx
```diff
@@ -0,0 +1,153 @@
...
+      setWrapState((state) => ({
+        ...state,
+        error: hasInputAmount ? WrapInputError.INSUFFICIENT_WRAPPED_BALANCE : WrapInputError.ENTER_WRAPPED_AMOUNT,
+      }))
+    }
+  }, [hasInputAmount, setWrapState, sufficientBalance, wrapType])
+
+  const callback = useCallback(async () => {
+    if (
+      !wrappedNativeCurrencyContract ||
```
zzmp, line 129: nit: add a new variable to cover this: const canWrap = ...?

src/lib/hooks/swap/useWrapCallback.tsx
```diff
@@ -0,0 +1,153 @@
...
+      !parsedAmountIn ||
+      wrapType === WrapType.NOT_APPLICABLE
+    ) {
+      return Promise.reject()
+    }
+    setWrapState((state) => ({ ...state, loading: true }))
+    const result = await (wrapType === WrapType.WRAP
+      ? wrappedNativeCurrencyContract.deposit({ value: `0x${parsedAmountIn.quotient.toString(16)}` })
+      : wrappedNativeCurrencyContract.withdraw(`0x${parsedAmountIn.quotient.toString(16)}`))
+    result.wait(1).then(() => setWrapState((state) => ({ ...state, loading: false })))
```
zzmp, line 140: Add a comment documenting why you are waiting one block.

zzmp, line 140: This should be a finally (not a then) to ensure loading is set to false.

src/lib/hooks/swap/useWrapCallback.tsx
JFrankfurt: this value going from undefined -> false -> undefined would trigger a rerun of the `useEffect` below. Not a big deal, but undesirable imo.

src/lib/components/Swap/Status/StatusDialog.tsx
```diff
@@ -26,7 +26,9 @@ const TransactionRow = styled(Row)`
   flex-direction: row-reverse;
 `
 
-function ElapsedTime({ tx }: { tx: Transaction<SwapTransactionInfo> }) {
+type PendingTransaction = Transaction<SwapTransactionInfo | WrapTransactionInfo>
```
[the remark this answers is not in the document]
JFrankfurt: You could, but it would just move the `<SwapTransactionInfo | WrapTransactionInfo>` down to the usage of `PendingTransaction`. I don't think doing that gets us anything rn.

src/lib/components/Swap/Status/StatusDialog.tsx
```diff
@@ -63,14 +65,24 @@ function TransactionStatus({ tx, onClose }: TransactionStatusProps) {
     return tx.receipt?.status ? CheckCircle : Spinner
   }, [tx.receipt?.status])
   const heading = useMemo(() => {
-    return tx.receipt?.status ? <Trans>Transaction submitted</Trans> : <Trans>Transaction pending</Trans>
-  }, [tx.receipt?.status])
+    if (tx.info.type === TransactionType.SWAP) {
+      return tx.receipt?.status ? <Trans>Swap confirmed</Trans> : <Trans>Swap pending</Trans>
```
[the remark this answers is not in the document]
JFrankfurt: "submitted" is incorrect here
submission is when the transaction is broadcast to the network
confirmation is when the transaction is included in a block

The prior implementation was using "submitted" to mean "confirmed".

src/lib/components/Swap/WrapErrorText.tsx
```diff
@@ -0,0 +1,29 @@
+import { Trans } from '@lingui/macro'
+import useNativeCurrency from 'lib/hooks/useNativeCurrency'
+
+export enum WrapInputError {
```
zzmp, line 4: WrapInputError should be defined with the hook that returns it, in useWrapCallback.tsx.
Also, could you just call it `WrapError`?

WrapErrorText should be defined separately (ie should stay in this file).
nit: Would also like WrapErrorText to be renamed to `WrapError`, but up to you.

zzmp (approved):

Scheusal13 (commented):
0xEa5C920973eBfff95270dd52c8045c4e6eBFc847

Scheusal13 (commented):
0xEa5C920973eBfff95270dd52c8045c4e6eBFc847

PR #3313, opened 2022-02-14
fix: ignore stale SOR fetches

zzmp:


src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -100,9 +41,8 @@ export function useBestTrade(
   return useMemo(
     () => ({
       ...(useFallback ? fallbackTrade : clientSORTrade),
-      ...(syncing ? { state: TradeState.SYNCING } : {}),
```
ianlapham, line 103: I have a feeling removing the syncing check is what is causing the break on the UI dim state

ianlapham (changes requested):

src/lib/hooks/swap/useBestTrade.ts
zzmp: I updated useClientSideSmartOrderRouterTrade to include the trade as a consequence of its fetched state, so the trade will no longer update until a new one is ready (or none is found). I think retaining the trade with the loading state is cleaner than re-inferring it one level down.

ianlapham (approved):
Looks good - one last request if you can think of solution is that the loading text flickers on every input now (looks like it flickers between the "fetching" text and the rate?)

if its not obvious could fix later - not sure its worth spending much more time on

PR #3324, opened 2022-02-16
fix: connect wallet button

zzmp:
Updates the "Connect you wallet" text to be a button when a handler is present, and adds a handler to the cosmos to demo that.

This makes the text/button focusable (using tab-selection) and gives it a cursor on hover, making it more accessible.


src/lib/components/Swap/Swap.fixture.tsx
```diff
@@ -71,6 +71,7 @@ function Fixture() {
       defaultOutputAddress={optionsToAddressMap[defaultOutput]}
       defaultOutputAmount={defaultOutputAmount}
       tokenList={tokens}
+      onConnectWallet={() => console.log('onConnectWallet')}
```
ianlapham, line 74: is this intentional?

ianlapham (commented):
Small question

src/lib/components/Swap/Swap.fixture.tsx
zzmp: Yes - this way it will show as a button in our cosmos viewer, and _something_ will happen when you press it.

src/lib/components/Dialog.tsx
```diff
@@ -40,7 +40,7 @@ export function Provider({ value, children }: ProviderProps) {
     }
   }, [active])
   return (
-    <div ref={ref}>
+    <div ref={ref} style={{ isolation: 'isolate' }}>
```
JFrankfurt, line 43: is this necessary? Looks like it's usually used for opacity/z-index stuff

zzmp: yes - without this, the button's click-target is behind the modal (which is not rendered).

JFrankfurt:
Can you add a screenshot or description of what's happening here?

src/lib/components/Dialog.tsx
zzmp: Documented this.

src/lib/components/Swap/Swap.fixture.tsx
zzmp: Removed this.

zzmp: jk I just documented this with a comment.

ianlapham (approved):

PR #3341, opened 2022-02-21
fix: bring back url phishing warning

moodysalem:


src/pages/App.tsx
```diff
@@ -52,12 +54,12 @@ const BodyWrapper = styled.div`
   `};
 `
 
-const HeaderWrapper = styled.div`
+const HeaderWrapper = styled.div<{ urlWarningVisible?: boolean }>`
```
moodysalem, line 57: required isn't it?

src/components/Header/URLWarning.tsx
```diff
@@ -0,0 +1,54 @@
...
+          <code style={{ padding: '0 4px', display: 'inline', fontWeight: 'bold' }}>app.uniswap.org</code>
+        </Trans>
+      </div>
+      <StyledClose size={12} onClick={toggleURLWarning} />
+    </PhishAlert>
+  ) : (
+    <PhishAlert>
+      <div style={{ display: 'flex' }}>
+        <AlertTriangle style={{ marginRight: 6 }} size={12} />
+        <Trans>
```
moodysalem, line 45: only this is different, the rest is the same, move the ternary in

src/pages/App.tsx
```diff
@@ -52,12 +54,12 @@ const BodyWrapper = styled.div`
...
   ${({ theme }) => theme.flexRowNoWrap}
   width: 100%;
   justify-content: space-between;
   position: fixed;
-  top: 0;
+  top: ${({ urlWarningVisible }) => (urlWarningVisible ? '20px' : 0)};
```
moodysalem, line 62: this bit is odd, the warning may be larger if the text wraps

aballerr:
confirmed with Moody this can be closed

PR #3347, opened 2022-02-23
feat(swap): default 1 native to usdc on the swap page

moodysalem:


moodysalem:
alternately, we shouldn't show this warning if the token came from a shorthand

src/constants/tokens.ts
```diff
@@ -31,7 +40,7 @@ export const DAI_OPTIMISM = new Token(
   'DAI',
   'Dai stable coin'
 )
-export const USDC = new Token(
+export const USDC_MAINNET = new Token(
```
zzmp, line 43: Are these not exported from `@uniswap/smart-order-router`? Seems weird that all of the other networks _are_ exported, but not MAINNET, ARBITRUM, POLYGON, or OPTIMISM.

src/lib/hooks/useCurrency.ts
```diff
@@ -28,65 +29,71 @@ function parseStringOrBytes32(str: string | undefined, bytes32: string | undefin
...
+  const tokenNameBytes32 = useSingleCallResult(tokenContractBytes32, 'name', undefined, NEVER_RELOAD)
+  const symbol = useSingleCallResult(tokenContract, 'symbol', undefined, NEVER_RELOAD)
+  const symbolBytes32 = useSingleCallResult(tokenContractBytes32, 'symbol', undefined, NEVER_RELOAD)
+  const decimals = useSingleCallResult(tokenContract, 'decimals', undefined, NEVER_RELOAD)
 
   return useMemo(() => {
-    if (token) return token
-    if (tokenAddress === null) return null
-    if (!chainId || !address) return undefined
+    if (typeof tokenAddress !== 'string') return tokenAddress
```
zzmp, line 47: This should explicitly return undefined, because the fn will never return a `string`. You should elide it with the following line.

src/lib/hooks/useCurrency.ts
```diff
@@ -28,65 +29,71 @@ function parseStringOrBytes32(str: string | undefined, bytes32: string | undefin
...
+/**
+ * Returns a Token from the tokenAddress.
+ * Returns null if token is loading or null was passed.
+ * Returns undefined if tokenAddress is invalid or token does not exist.
+ */
+export function useTokenFromMapOrNetwork(tokens: TokenMap, tokenAddress?: string | null): Token | null | undefined {
+  const address = isAddress(tokenAddress)
+  const token: Token | undefined = address ? tokens[address] : undefined
+
+  const tokenFromNetwork = useTokenFromNetwork(token ? undefined : address ? address : undefined)
```
zzmp, line 84: can this be simplified as `useTokenFromNetwork(token ?? address ? address : undefined)`? Nested ternary is hard to read.

If you change address to be `isAddress(tokenAddress) || undefined`, you can just have it be `token ?? address`.

src/state/swap/hooks.tsx
```diff
@@ -216,9 +219,14 @@ function validatedRecipient(recipient: any): string | null {
 export function queryParametersToSwapState(parsedQs: ParsedQs): SwapState {
   let inputCurrency = parseCurrencyFromURLParameter(parsedQs.inputCurrency)
   let outputCurrency = parseCurrencyFromURLParameter(parsedQs.outputCurrency)
-  if (inputCurrency === '' && outputCurrency === '') {
+  let typedValue = parseTokenAmountURLParameter(parsedQs.exactAmount)
+  const independentField = parseIndependentFieldURLParameter(parsedQs.exactField)
+
+  if (inputCurrency === '' && outputCurrency === '' && typedValue === '' && independentField === Field.INPUT) {
     // default to ETH input
```
zzmp, line 226: Update this comment (eg `// Defaults to 1 ETH -> DAI`)

src/pages/Swap/index.tsx
```diff
@@ -84,10 +86,19 @@ export default function Swap({ history }: RouteComponentProps) {
...
-      urlLoadedTokens.filter((token: Token) => {
-        return !Boolean(token.address in defaultTokens)
-      }),
-    [defaultTokens, urlLoadedTokens]
+      urlLoadedTokens
+        .filter((token: Token) => {
+          return !Boolean(token.address in defaultTokens)
+        })
+        .filter((token: Token) => {
+          const supported = supportedChainId(chainId)
```
zzmp, line 94: What is this doing? Could you leave a comment to clarify it?

zzmp (approved):

src/constants/tokens.ts
moodysalem: we now use the exported tokens from smart-order-router

src/lib/hooks/useCurrency.ts
zzmp: token || (address ? ...)?

moodysalem: also not equivalent, we want it to be undefined if token is defined

PR #3354, opened 2022-02-23
feat: use permit when valid on approvals

ianlapham:


src/lib/components/Swap/SwapButton.tsx
```diff
@@ -61,72 +59,73 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
...
+      disabled ||
+      !chainId ||
+      loadingSignature ||
+      (inputCurrencyAmount && inputCurrencyBalance && inputCurrencyBalance.lessThan(inputCurrencyAmount))
+    ) {
+      return { disabled: true }
+    }
+    if (notApproved) {
+      const currency = inputCurrency || approvalCurrencyAmount?.currency
+      invariant(currency)
```
JFrankfurt, line 85: Will using invariant here throw to the worst case error screen?

src/lib/components/Swap/SwapButton.tsx
```diff
@@ -61,72 +59,73 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
...
-          },
-        }
-      } else if (inputCurrencyAmount && inputCurrencyBalance && !inputCurrencyBalance.lessThan(inputCurrencyAmount)) {
-        return {}
+    if (
+      disabled ||
+      !chainId ||
+      loadingSignature ||
+      (inputCurrencyAmount && inputCurrencyBalance && inputCurrencyBalance.lessThan(inputCurrencyAmount))
+    ) {
```
JFrankfurt, line 80: should we also be returning disabled when the `inputCurrencyAmount` and `inputCurrencyBalance` are falsy?

src/lib/components/Swap/SwapButton.tsx
```diff
@@ -61,72 +59,73 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
...
 
   const addTransaction = useAddTransaction()
-  const addApprovalTransaction = useCallback(() => {
-    getApproval().then((transaction) => {
-      if (transaction) {
-        addTransaction({ type: TransactionType.APPROVAL, ...transaction })
-      }
-    })
-  }, [addTransaction, getApproval])
+  const deadline = useTransactionDeadline()
```
JFrankfurt, line 72: nit: can you move this down beneath the action props to be closer to its usage in `useSwapCallback`?

src/lib/components/Swap/SwapButton.tsx
```diff
@@ -61,72 +59,73 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
...
-            children: <Trans>Approve</Trans>,
-          },
-        }
-      } else if (inputCurrencyAmount && inputCurrencyBalance && !inputCurrencyBalance.lessThan(inputCurrencyAmount)) {
-        return {}
+    if (
+      disabled ||
+      !chainId ||
+      loadingSignature ||
+      (inputCurrencyAmount && inputCurrencyBalance && inputCurrencyBalance.lessThan(inputCurrencyAmount))
```
zzmp, line 79: nit: pull this out into its own variable (with useMemo)

src/lib/components/Swap/SwapButton.tsx
```diff
@@ -61,72 +59,73 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
...
+    ) {
+      return { disabled: true }
+    }
```
zzmp, line 82: nit: I like to put a newline after my guard statements

src/lib/components/Swap/SwapButton.tsx
```diff
@@ -61,72 +59,73 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
...
+      const currency = inputCurrency || approvalCurrencyAmount?.currency
+      invariant(currency)
+      return {
+        action: {
+          message: supportsPermit ? (
+            <Trans>Allow {currency.symbol} first</Trans>
+          ) : (
+            <Trans>Approve {currency.symbol} first</Trans>
+          ),
+          onClick: handleApproveOrPermit,
```
zzmp, line 93: This should be wrapped in a callback to include the transaction tracking (from useAddTransaction).

src/lib/components/Swap/SwapButton.tsx
zzmp: This must also check that inputCurrencyAmount && inputCurrencyBalance are both truthy

src/lib/components/Swap/SwapButton.tsx
```diff
@@ -61,72 +59,73 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
...
+      return {
+        disabled: true,
+        action: {
+          message: (
+            <EtherscanLink type={ExplorerDataType.TRANSACTION} data={approvalHash}>
+              <Trans>Approval pending</Trans>
+            </EtherscanLink>
+          ),
+          icon: Spinner,
+          onClick: handleApproveOrPermit,
```
zzmp, line 108: omit this property. if typings do not allow it, make it a nop. (Jordan is updating typings to allow it.)

zzmp (changes requested):

src/lib/components/Swap/SwapButton.tsx
ianlapham: i think so? this logic was just taken from existing approval logic - so maybe thats a question for another PR?

src/lib/components/Swap/SwapButton.tsx
ianlapham: Ah yes we need to do this

src/lib/hooks/swap/useSwapApproval.ts
```diff
@@ -134,3 +138,69 @@ export function useSwapApprovalOptimizedTrade(
...
+  // Check status of permit and wether token supports it
+  const {
+    state: signatureState,
+    signatureData,
+    gatherPermitSignature,
+  } = useERC20PermitFromTrade(trade, allowedSlippage, deadline)
+
+  const notApproved =
+    approval === ApprovalState.NOT_APPROVED &&
+    !(signatureState === UseERC20PermitState.SIGNED && Boolean(gatherPermitSignature))
```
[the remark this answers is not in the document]
ianlapham: This was also copied from app logic, but yeah we should be able to only check signed status

src/lib/hooks/swap/useSwapApproval.ts
```diff
@@ -134,3 +138,69 @@ export function useSwapApprovalOptimizedTrade(
...
+          addApprovalTransaction()
+        }
+      }
+    } else {
+      addApprovalTransaction()
+    }
+  }, [signatureState, gatherPermitSignature, addApprovalTransaction])
+
+  return {
+    notApproved,
```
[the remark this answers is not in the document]
ianlapham: ahhh i like this

src/lib/components/Swap/SwapButton.tsx
ianlapham: dont we also need to check outputCurrencyAmount?

src/lib/components/Swap/SwapButton.tsx
zzmp: yes it will. I believe that I put that here. The other option is to check that `currency` is defined, but I think that with the logic above that's already checked for. It may be possible to rewrite these conditionals so that the typechecker understands all of this betetr.

src/lib/components/Swap/SwapButton.tsx
zzmp: I don't think we care about that for approvals, only for trades.

src/lib/components/Swap/SwapButton.tsx
```diff
@@ -60,42 +61,59 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
...
+    })
+  }, [addTransaction, handleApproveOrPermit])
 
   const actionProps = useMemo((): Partial<ActionButtonProps> | undefined => {
     if (
       disabled ||
       !chainId ||
-      loadingSignature ||
+      approvalState === ApproveOrPermitState.PENDING_SIGNATURE ||
+      !(inputCurrencyAmount && outputCurrencyAmount && inputCurrencyBalance) ||
```
zzmp, line 90: We can prompt for approval/permit without an outputCurrencyAmount.

src/lib/hooks/swap/useSwapApproval.ts
```diff
@@ -155,51 +160,52 @@ export const useApproveAndPermit = (
...
-    })
-  }, [addTransaction, getApproval])
-
-  // if permit is supported, trigger a signature, if not create approval transaction
+  // If permit is supported, trigger a signature, if not create approval transaction.
   const handleApproveOrPermit = useCallback(async () => {
     if (signatureState === UseERC20PermitState.NOT_SIGNED && gatherPermitSignature) {
       try {
-        await gatherPermitSignature()
+        return await gatherPermitSignature()
```
zzmp, line 179: nit: split this into two lines so noone removes the `await` (which is needed for the try/catch)

src/lib/hooks/swap/useSwapApproval.ts
```diff
@@ -155,51 +160,52 @@ export const useApproveAndPermit = (
...
       } catch (error) {
-        // try to approve if gatherPermitSignature failed for any reason other than the user rejecting it
+        // Try to approve if gatherPermitSignature failed for any reason other than the user rejecting it
```
zzmp, line 181: nit: period.

src/lib/hooks/swap/useSwapApproval.ts
```diff
@@ -155,51 +160,52 @@ export const useApproveAndPermit = (
...
+      return ApproveOrPermitState.PENDING_SIGNATURE
+    }
+    if (notApproved && Boolean(gatherPermitSignature)) {
+      return ApproveOrPermitState.REQUIRES_SIGNATURE
+    }
+    if (notApproved) {
+      return ApproveOrPermitState.REQUIRES_APPROVAL
     }
-  }, [signatureState, gatherPermitSignature, addApprovalTransaction])
+    return undefined
```
zzmp, line 204: Should this be another state, ApprovedOrPermitState.APPROVED?

zzmp (changes requested):
We should still prompt for approval/permit without an outputCurrencyAmount.

src/lib/components/Swap/SwapButton.tsx
ianlapham: got it, removing check on output

src/lib/hooks/swap/useSwapApproval.ts
ianlapham: yeah i like that

src/lib/components/Swap/SwapButton.tsx
```diff
@@ -88,26 +89,28 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
...
       disabled ||
       !chainId ||
+      wrapLoading ||
+      (wrapType !== WrapType.NOT_APPLICABLE && wrapError) ||
       approvalState === ApproveOrPermitState.PENDING_SIGNATURE ||
-      !(inputCurrencyAmount && outputCurrencyAmount && inputCurrencyBalance) ||
-      (inputCurrencyAmount && inputCurrencyBalance && inputCurrencyBalance.lessThan(inputCurrencyAmount)),
-    [approvalState, chainId, disabled, inputCurrencyAmount, inputCurrencyBalance, outputCurrencyAmount]
+      !(inputTradeCurrencyAmount && inputCurrencyBalance) ||
+      (inputTradeCurrencyAmount && inputCurrencyBalance && inputCurrencyBalance.lessThan(inputTradeCurrencyAmount)),
```
zzmp, line 102: this is redundant with L101 - can you replace `(inputTradeCurrencyAmount && inputCurrencyBalance && inputCurrencyBalance.lessThan(inputTradeCurrencyAmount))` with `inputCurrencyBalance.lessThan(inputTradeCurrencyAmount)`?

zzmp (approved):

PR #3355, opened 2022-02-24
fix: wrap Eip1193Bridge to fix outstanding bugs

zzmp:


src/lib/hooks/useEip1193Provider.ts
```diff
@@ -17,32 +16,42 @@ class Eip1193Bridge extends ExperimentalEip1193Bridge {
         const result = await this.provider.getNetwork()
         return '0x' + result.chainId.toString(16)
       }
+      case 'eth_sendTransaction': {
+        if (!this.signer) break
+
+        // TODO(zzmp): JsonRpcProvider filters from/gas fields from the params.
```
zzmp, line 22: This should be opened as an issue on github.com/ethers-io/ethers.js, and someone else should verify this as a bug.

ianlapham (approved):

ianlapham (changes requested):
Just noticed styling issue on bottom border

I've noticed in cosmos we get this extra padding - but my guess is when we import the widget into a react app this will go away

ianlapham (approved):

PR #3364, opened 2022-02-25
fix: Italian translation

jacko06v:
the translation "Balance" was wrong
"Bilanciamento" was used which means "balancing".

Changed to "Disponibile" which means "Available" and is much more appropriate


NoahZinsmeister:
hi @jacko06v , thanks for this! would you mind making this change via Crowdin instead? that's how we manage translation fixes, rather than changes to the `.po` files. instructions here: https://github.com/Uniswap/interface/blob/main/CONTRIBUTING.md#translations

PR #3373, opened 2022-02-26
fix: convert to commonjs

zzmp:
Includes pinning exact @web3-react versions which support cjs


JFrankfurt (approved):
What's the deal with the vibrant changes?

zzmp:
> What's the deal with the vibrant changes?

Should we export both cjs and esm in the future, esm resolution requires that cjs modules be fully specified, so the vibrant imports will need to include an extension. See https://nodejs.org/api/esm.html#mandatory-file-extensions.

This is an artifact of my first working to fix esm before abandoning it for cjs (due to too many upstream deps incorrectly supporting esm). I left it in because it is strictly more correct, and may save a headache in the future.

PR #3376, opened 2022-02-28
fix(widgets): disable interactions and don't show summary screen on unsupported networks

JFrankfurt:


src/lib/components/Swap/SwapButton.tsx
```diff
@@ -162,6 +163,8 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
       })
   }, [addTransaction, inputCurrencyAmount, outputCurrencyAmount, setDisplayTxHash, swapCallback, tradeType])
 
+  const onSupportedNetwork = useMemo(() => chainId && ALL_SUPPORTED_CHAIN_IDS.includes(chainId), [chainId])
```
ianlapham, line 166: Can we just make this a standalone hook somewhere? seems pretty useful

ianlapham (commented):
Interaction looks good, quick q about the hooks

src/lib/components/Swap/Input.tsx
```diff
@@ -81,11 +81,24 @@ export default function Input({ disabled, focused }: InputProps) {
     return insufficientBalance ? 'error' : undefined
   }, [balance, inputCurrencyAmount, swapInputCurrencyAmount])
 
+  const amount = useMemo(() => {
+    if (disabled) {
+      return ''
```
zzmp, line 86: You should ping @ppphhhiiilll and ask what this should show when disabled. Showing nothing may look broken.

src/lib/components/Swap/Input.tsx
```diff
@@ -81,11 +81,24 @@ export default function Input({ disabled, focused }: InputProps) {
     return insufficientBalance ? 'error' : undefined
   }, [balance, inputCurrencyAmount, swapInputCurrencyAmount])
 
+  const amount = useMemo(() => {
```
zzmp, line 84: This is the same between `Input` and `Output`. Can you make it an exported hook from `Input` and reuse it in `Output`? See `LoadingRow`, `Balance` as prior work.

src/lib/components/Swap/SwapButton.tsx
```diff
@@ -171,7 +174,7 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
       >
         <Trans>Review swap</Trans>
       </ActionButton>
-      {activeTrade && (
+      {activeTrade && onSupportedNetwork && (
```
zzmp, line 177: Does this mean that changing networks will close this dialog?
I think instead, changing networks should `setActiveTrade(undefined)`. Otherwise you'll still have an active trade for the wrong network (and switching between supported networks will show the old network's trade).

src/lib/components/Swap/index.tsx
```diff
@@ -74,15 +75,15 @@ export default function Swap(props: SwapProps) {
       {isSwapSupported && <SwapInfoUpdater />}
       <Header title={<Trans>Swap</Trans>}>
         {active && <Wallet disabled={!account} onClick={props.onConnectWallet} />}
-        <Settings disabled={!active} />
+        <Settings disabled={!active || !onSupportedNetwork} />
```
zzmp, line 78: Pull out `disabled` into its own variable.

zzmp (changes requested):

src/lib/components/Swap/Input.tsx
JFrankfurt: It actually looks more borked when there is a value here. The text input/output looks interactive, but the field itself is disabled.

src/lib/components/Swap/SwapButton.tsx
JFrankfurt: This is actually already happening in the `useSyncSwapDefaults` hook. Any chain changes will clear fields to default values for the new chain.

src/lib/components/Swap/Input.tsx
JFrankfurt: discussed w/ @ppphhhiiilll and confirmed this behavior

src/lib/components/Swap/index.tsx
```diff
@@ -54,34 +54,36 @@ export default function Swap(props: SwapProps) {
...
-  const onSupportedNetwork = useMemo(() => chainId && ALL_SUPPORTED_CHAIN_IDS.includes(chainId), [chainId])
+  const onSupportedNetwork = useOnSupportedNetwork()
   const isSwapSupported = useMemo(
     () => Boolean(active && onSupportedNetwork && tokenList?.length),
     [active, onSupportedNetwork, tokenList?.length]
   )
 
   const focused = useHasFocus(wrapper)
 
+  const interactionsDisabled = useMemo(() => !active || !onSupportedNetwork, [active, onSupportedNetwork])
```
zzmp, line 73: as simple booleans, doesn't need to use `useMemo`.
also, prefer a positive variable name, and I prefer to prefix booleans with "is" or "should" - `isInteractive`?

src/lib/components/Swap/Input.tsx
```diff
@@ -8,6 +8,7 @@ import {
   useSwapCurrencyAmount,
   useSwapInfo,
 } from 'lib/hooks/swap'
+import useInputOutputFieldAmount from 'lib/hooks/swap/useInputOutputFieldAmount'
```
zzmp, line 11: As a hook specific to Input/Output, I think this should be inlined to Input.tsx and not in a separate file. I don't think it's weird to export from Input and import in Output.

src/lib/components/Swap/Input.tsx
```diff
@@ -81,11 +82,22 @@ export default function Input({ disabled, focused }: InputProps) {
     return insufficientBalance ? 'error' : undefined
   }, [balance, inputCurrencyAmount, swapInputCurrencyAmount])
 
+  const amount = useInputOutputFieldAmount(
+    useMemo(
```
zzmp, line 86: This does not need to be memoized because it is destructured by the hook and used in a useMemo within the hook, so the references should be the same.

src/lib/components/Swap/WrapErrorText.tsx
```diff
@@ -0,0 +1,22 @@
+import { Trans } from '@lingui/macro'
+import { WrapInputError } from 'lib/hooks/swap/useWrapCallback'
+import useNativeCurrency from 'lib/hooks/useNativeCurrency'
+
+export function WrapErrorText({ wrapInputError }: { wrapInputError: WrapInputError }) {
```
zzmp, line 5: I think this file came in from a different PR and shouldn't be here?

zzmp (changes requested):

src/lib/components/Swap/Input.tsx
```diff
@@ -81,11 +82,22 @@ export default function Input({ disabled, focused }: InputProps) {
     return insufficientBalance ? 'error' : undefined
   }, [balance, inputCurrencyAmount, swapInputCurrencyAmount])
 
+  const amount = useInputOutputFieldAmount(
```
ianlapham, line 85: can we call this `useFormattedFieldAmount` or something? I feel like we have so many vars named Amount rn. Could help to clarify this is a formatting thing

ianlapham (commented):
Small nits

src/lib/components/Swap/WrapErrorText.tsx
JFrankfurt: gah thanks

src/lib/components/Swap/Input.tsx
```diff
@@ -46,6 +46,31 @@ export interface InputProps {
   focused: boolean
 }
 
+interface UseInputOutputFormattedFieldArguments {
```
zzmp, line 49: nit: I think you should just inline this, as it's not a component.

src/lib/components/Swap/Input.tsx
```diff
@@ -46,6 +46,31 @@ export interface InputProps {
...
+  disabled: boolean
+  currencyAmount: CurrencyAmount<Currency> | undefined
```
zzmp, line 51: nit: I prefer to specify this (and below) as `currencyAmount?: CurrencyAmount<Currency>` instead of explicitly using `undefined`, as it's less verbose and usually the same.

src/lib/components/Swap/Input.tsx
```diff
@@ -45,6 +46,31 @@ export interface InputProps {
...
+  fieldAmount: string | undefined
+}
+
+export function useInputOutputFormattedField({
```
zzmp, line 55: update name to `useFormattedFieldAmount` - it's independent of InputOutput.

src/lib/components/Swap/SwapButton.tsx
zzmp: I'm still confused here - I think that any change to the chain should invalidate any active trade, which should be done as

```
useEffect(() => setActiveTrade(undefined), [chainId])
```

I don't think that `onSupportedNetwork` belongs in this file, as you wouldn't have an active trade if you aren't on a supported network (so it's a prerequisite to getting to this point).

zzmp (changes requested):
Still confused about the use of onSupportedNetwork in the SwapButton.

src/lib/components/Swap/SwapButton.tsx
JFrankfurt: roughly, here's the logic flow:

`SwapButton.tsx`
```
const { trade, ... } = useSwapInfo()
...
useEffect(() => setActiveTrade(activeTrade => activeTrade && trade.trade), [trade])
```

`useSyncSwapDefaults`
```
const updateSwap = useUpdateAtom(swapAtom)

const setToDefaults = useCallback(() => updateSwap(DEFAULTS_FROM_INTEGRATOR))

useLayoutEffect(() => if (chainId && chainId !== previousChainId) setToDefaults(), [chainId, previousChainId, setToDefaults])
```

`useSwapInfo`
```
this is a lot of code, but it reads out swapAtom to build SwapInfo
any time swapAtom changes (like because the chain changes and triggers the useSyncSwapDefaults effect) it rebuilds SwapInfo. 

THIS is where we get a new trade entity. This is the trade.trade reference in the first code snippet. 
```

JFrankfurt: chain change:
1. triggers useLayoutEffect
1. triggers setToDefaults
1. triggers updateSwap (resets values for swapAtom)
1. triggers rerun of useSwapInfo
1. clears calculated trade from `useBestTrade`
1. triggers rerun of the hook in SwapButton
2. calls `setActiveTrade((activeTrade) => truthyActiveTrade && undefined)`
3. activeTrade is now undefined

zzmp (approved):

PR #3379, opened 2022-02-28
fix: exports

zzmp:
- ignores cosmos' build artifact, /cosmos-export (from `yarn cosmos-export`)
- builds and exports cjs and esm modules (index, tokenlist spec, and locales)
  - fixes the builds to include chunks for efficient lazy loading and locales
  - inlines json and @lingui into the bundles to avoid esm limitations:
    - json cannot be natively loaded by esm
    - @lingui uses an unsupported file extension (it is typed as commonjs, but uses js for esm), so it does not load as esm
  - pins the newest cjs-compatible @web3-react versions


JFrankfurt (approved):
It feels like I'd need a looot of background knowledge to review this effectively, but nothing jumps out at me as being amiss here. lgtm!

NoahZinsmeister (approved):

ianlapham (approved):

PR #3392, opened 2022-03-02
fix: set appropriate types to nowrap

zzmp:
Set appropriate types to nowrap. This prevents linebreaks during element resizing, as reported [1].

[1]: https://discord.com/channels/597638925346930701/947940283167936532/948229837649240114

### DIFF
--- a/src/lib/theme/type.tsx
+++ b/src/lib/theme/type.tsx
@@ -5,10 +5,11 @@ import { Color } from './theme'
 
 type TextProps = Omit<TextPropsWithCss, 'css' | 'color'> & { color?: Color }
 
-const TextWrapper = styled(Text)<{ color?: Color; lineHeight: string }>`
+const TextWrapper = styled(Text)<{ color?: Color; lineHeight: string; noWrap?: true }>`
   color: ${({ color = 'currentColor', theme }) => theme[color as Color]};
   // Avoid the need for placeholders by setting min-height to line-height.
   min-height: ${({ lineHeight }) => lineHeight};
+  white-space: ${({ noWrap }) => noWrap && 'nowrap'};
 `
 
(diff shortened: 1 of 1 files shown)


JFrankfurt:
This feels like the wrong way to be configuring component styles. I think if a consumer of the styled component wants to extend or overwrite default behavior, it should be done w/ the standard wrapping pattern:

```
const SomeSpecialText = styled(TextWrapper)`
    white-space: nowrap;
`
const SomeSpecialText2 = styled(TextWrapper)`
    white-space: pre;
`
```

ianlapham (approved):

ianlapham:
> This feels like the wrong way to be configuring component styles. I think if a consumer of the styled component wants to extend or overwrite default behavior, it should be done w/ the standard wrapping pattern:
> 
> ```
> const SomeSpecialText = styled(TextWrapper)`
>     white-space: nowrap;
> `
> const SomeSpecialText2 = styled(TextWrapper)`
>     white-space: pre;
> `
> ```

I wonder if this is okay though - because a majority of the time in our app we dont want to wrap

zzmp:
My thinking here is that these types should never be wrapped because they are headers, captions, or buttons, so it's ok to encode that directly in the type.

JFrankfurt:
In my experience these kinds of base components end up getting wrapped pretty frequently (i.e., to add a bit of padding, a special case for opacity effects, etc.). Unless we want to come up with a way of explicitly and programmatically restricting the most common pattern of styled component reuse in cases like this, I don't think we should use this pattern.

In many cases we won't wrap and will use the base component. But when we do need to add some styles, I think it should be done w/ wrapping, and not by adding a new prop to the base component so we can pass it through to the `styled template string.

PR #3403, opened 2022-03-02
feat: create donation banner

ianlapham:


src/components/Popups/index.tsx
```diff
@@ -65,16 +69,24 @@ export default function Popups() {
   const { chainId } = useActiveWeb3React()
   const isNotOnMainnet = Boolean(chainId && chainId !== SupportedChainId.MAINNET)
 
+  // donation popup, hide if after fixed duration
```
zzmp, line 72: This logic should be encapsulated in useShowDonationLink.

src/state/user/hooks.tsx
```diff
@@ -117,6 +118,19 @@ export function useShowSurveyPopup(): [boolean | undefined, (showPopup: boolean)
   return [showSurveyPopup, toggleShowSurveyPopup]
 }
 
+export function useShowDonationLink(): [boolean | undefined, (showDonationLink: boolean) => void] {
```
zzmp, line 121: This should always return a boolean. If it is undefined, just return `true`.

src/components/Popups/index.tsx
```diff
@@ -65,16 +69,24 @@ export default function Popups() {
...
+  const showDonation = donationVisible !== false && !durationOver
+
   return (
     <>
       <FixedPopupColumn gap="20px" extraPadding={urlWarningActive} xlPadding={isNotOnMainnet}>
         <ClaimPopup />
         {activePopups.map((item) => (
           <PopupItem key={item.key} content={item.content} popKey={item.key} removeAfterMs={item.removeAfterMs} />
         ))}
+        {showDonation ? <DonationLink /> : null}
```
zzmp, line 85: Why does DonationLink not get pushed to activePopups?

ianlapham: active popups are from transactions , this one is a custom element

src/components/Popups/DonationLink.tsx
```diff
@@ -0,0 +1,132 @@
...
+    overflow: hidden;
+    background-size: 300px;
+    content: '';
+    height: 1200px;
+    width: 400px;
+    opacity: 0.1;
+    position: absolute;
+    transform: rotate(25deg) translate(-140px, -60px);
+    width: 300px;
+    z-index: 1;
```
justindomingue, line 46: can you try to use `Z_INDEX` here 🙈

justindomingue: ditto L56

src/components/Popups/DonationLink.tsx
```diff
@@ -0,0 +1,132 @@
...
+  transform: rotate(90deg);
+`
+
+const StyledLink = styled(ExternalLink)`
+  text-decoration: none !important;
+`
+
+export default function DonationLink() {
+  const [darkMode] = useDarkModeManager()
+  const [, setVisible] = useShowDonationLink()
```
justindomingue, line 95: as a follow-up this should probably live in the `user` slice? will be annoying to have to dismiss constantly

src/components/Popups/DonationLink.tsx
```diff
@@ -0,0 +1,132 @@
...
+        ReactGA.event({
+          category: 'Donate',
+          action: 'Link to Ukraine site.',
+        })
+      }}
+    >
+      <RowBetween>
+        <RowFixed>
+          <StyledFlagImage />
+          <ThemedText.Body fontWeight={600} ml="6px" style={{ textDecoration: 'none' }}>
```
justindomingue, line 114: is text decoration needed on a `body`?

src/components/Popups/index.tsx
```diff
@@ -65,16 +67,20 @@ export default function Popups() {
   const { chainId } = useActiveWeb3React()
   const isNotOnMainnet = Boolean(chainId && chainId !== SupportedChainId.MAINNET)
 
+  const [showDonation] = useShowDonationLink()
+
   return (
     <>
       <FixedPopupColumn gap="20px" extraPadding={urlWarningActive} xlPadding={isNotOnMainnet}>
         <ClaimPopup />
         {activePopups.map((item) => (
```
justindomingue, line 80: not familiar with `activePopups` but could yo not leverage that here instead of doing a custom donation popup?

src/components/Popups/DonationLink.tsx
ianlapham: good catch - not needed was using to hide styles from link

src/components/Popups/index.tsx
ianlapham: i think for custom styling and link we'd need a custom component here - unless we wanted to change design to work with txn popups (which have a text string and a link)

src/state/user/hooks.tsx
```diff
@@ -117,6 +119,26 @@ export function useShowSurveyPopup(): [boolean | undefined, (showPopup: boolean)
...
+  const showDonationLink = useAppSelector((state) => state.user.showDonationLink)
+
+  const toggleShowDonationLink = useCallback(
+    (showPopup: boolean) => {
+      dispatch(updateShowDonationLink({ showDonationLink: showPopup }))
+    },
+    [dispatch]
+  )
+
+  const timestamp = useCurrentBlockTimestamp()
```
justindomingue, line 135: surprised this isn't a number to begin with 🤔

ianlapham: i think its because we use this within contract calls?

PR #3408, opened 2022-03-02
fix: token balance reloading

zzmp:
- Fixes token balance UI to avoid horizontal overflow
- Fixes useBlock to avoid clearing the block number on window visibility change
  - This prevents re-fetching still valid data

### DIFF
--- a/src/lib/hooks/useBlockNumber.ts
+++ b/src/lib/hooks/useBlockNumber.ts
@@ -25,7 +25,8 @@ function useBlock() {
 
   useEffect(() => {
     if (library && chainId && windowVisible) {
-      setState({ chainId })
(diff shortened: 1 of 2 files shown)


src/lib/hooks/useBlockNumber.ts
```diff
@@ -25,7 +25,8 @@ function useBlock() {
 
   useEffect(() => {
     if (library && chainId && windowVisible) {
-      setState({ chainId })
+      // If chainId hasn't changed, don't clear the block. This prevents re-fetching still valid data.
```
ianlapham, line 28: smra

ianlapham (approved):

PR #3410, opened 2022-03-02
feat: user select

zzmp:
- Updates the UI to explicitly specify user-selectable elements
  - This is done via styled-components, through the text components, so that styled-components can expand the user-select to cross-browser compatible prefixed properties (eg -webkit-user-select)
- Updates useHasFocus to check against all contained elements, by using tabIndex=-1


src/lib/components/Swap/Summary/Summary.tsx
```diff
@@ -38,13 +38,13 @@ function TokenValue({ input, usdc, change }: TokenValueProps) {
     <Column justify="flex-start">
       <Row gap={0.375} justify="flex-start">
         <TokenImg token={input.currency} />
-        <ThemedText.Body2>
+        <ThemedText.Body2 style={{ userSelect: 'text' }}>
```
ianlapham, line 41: shouldnt this just be `userSelect`

src/lib/components/Swap/Summary/Summary.tsx
```diff
@@ -38,13 +38,13 @@ function TokenValue({ input, usdc, change }: TokenValueProps) {
...
           {formatCurrencyAmount(input, 6, i18n.locale)} {input.currency.symbol}
         </ThemedText.Body2>
       </Row>
       {usdc && usdcAmount && (
         <Row justify="flex-start">
-          <ThemedText.Caption color="secondary">
+          <ThemedText.Caption color="secondary" style={{ userSelect: 'text' }}>
```
ianlapham, line 47: same here

ianlapham (commented):
Can we remove the inline styles and make use of the useSelect props

ianlapham (approved):

src/lib/theme/type.tsx
```diff
@@ -3,12 +3,17 @@ import { Text, TextProps as TextPropsWithCss } from 'rebass'
...
+  userSelect?: true
+}
 
-const TextWrapper = styled(Text)<{ color?: Color; lineHeight: string }>`
+const TextWrapper = styled(Text)<{ color?: Color; lineHeight: string; userSelect?: true }>`
   color: ${({ color = 'currentColor', theme }) => theme[color as Color]};
   // Avoid the need for placeholders by setting min-height to line-height.
   min-height: ${({ lineHeight }) => lineHeight};
+  // user-select must be configured through styled-components for cross-browser compat.
+  user-select: ${({ userSelect }) => userSelect && 'text'};
```
JFrankfurt, line 16: - Can you add a comment here about `src\lib\components\Widget.tsx`'s `WidgetWrapper`'s `user-select: none`, explaining why this works the way it does? 

- I think the value should be `auto` here instead of `text`. [That should be more robust against edge case behavior.](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select#syntax)

zzmp: - comment added
- `auto` will not work, because the parent has a value of `none`, so this will adopt a value of `none`

JFrankfurt: `auto !important` or `unset`, then?

I'm surprised the simple `auto` doesn't work. I'd expected the value lower in the tree to take precedence. 🤷

PR #3411, opened 2022-03-02
feat: export types

zzmp:
- Exports types used in SwapWidget
- Cleans up token amount defaulting for cleaner exports


src/lib/components/Swap/SwapPropValidator.tsx
```diff
@@ -15,7 +16,7 @@ function isAddressOrAddressMap(addressOrMap: DefaultAddress): boolean {
   return false
 }
 
-type ValidatorProps = PropsWithChildren<SwapProps>
+type ValidatorProps = PropsWithChildren<DefaultTokenAmounts & FeeOptions>
```
JFrankfurt, line 19: Is this change necessary? It looks like it does almost the same thing:

`export interface SwapProps extends DefaultTokenAmounts, FeeOptions {`

src/lib/hooks/swap/useSyncTokenAmounts.ts
```diff
@@ -24,23 +32,16 @@ function useDefaultToken(
...
-  defaultInputAddress?: DefaultAddress
-  defaultInputAmount?: string
-  defaultOutputAddress?: DefaultAddress
-  defaultOutputAmount?: string
-}
-
-export default function useSyncSwapDefaults({
-  defaultInputAddress,
+export default function useSyncTokenAmounts({
+  defaultInputTokenAddress,
```
JFrankfurt, line 36: I think this hook should retain at least the `defaults` portion

maybe `useSyncTokenDefaults`?

also, note that the `Token` entities are being synced here as well, not just the amounts.

JFrankfurt (changes requested):

src/lib/components/Swap/SwapPropValidator.tsx
zzmp: This prevents circular dependencies (SwapPropValidator imports Swap, and Swap imports SwapPropValidator). I think typescript allows it with types, but it still leads to weird coupling.

src/lib/hooks/swap/useSyncTokenAmounts.ts
zzmp: I like `useSyncTokenDefaults`, will change.

JFrankfurt (approved):

PR #3428, opened 2022-03-03
feat: update loading icon

ianlapham:
- Used loading icon from app with updated styles to match widget theme


src/lib/icons/index.tsx
```diff
@@ -133,3 +133,48 @@ export const Spinner = styled(icon(SpinnerIcon))<{ color?: Color }>`
   stroke-linecap: round;
   stroke-width: 2;
 `
+
```
zzmp, line 136: This is long enough that I think it should go into its own file, src/lib/icons/DetailedSpinner.
Also, everything exported from this file is an Icon. Do the types work in Caption, or does the definition of Icon need to be expanded?

zzmp (changes requested):

src/lib/icons/index.tsx
ianlapham: types actually work fine in Caption no changes needed, but yeah can move this into its own file.

src/lib/icons/DetailedSpinner.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+to {
+  transform: rotate(360deg);
+}
+`
+
+const DetailedSpinnerWrapper = styled.div`
+  align-items: center;
+  display: flex;
+  height: 1rem;
```
zzmp, line 15: Considering that the InnerDot and OuterRing are defined in terms of px, I think the height and width should be set to 16px and not 1rem.

src/lib/icons/DetailedSpinner.tsx
```diff
@@ -0,0 +1,57 @@
+import styled, { keyframes } from 'lib/theme'
```
zzmp, line 1: nit: this spinner is statically sized to 16px, so I think it should be renamed InlineSpinner.

zzmp (approved):
two quick changes, but lgtm, no re-review needed

src/lib/icons/DetailedSpinner.tsx
ianlapham: Good call

PR #3429, opened 2022-03-03
fix: Improve Vote page

NoahZinsmeister:
- shows ` / QUORUM` after the For vote totals for all proposals against the current governor
- fixes the For/Against percentage bar logic
- makes vote totals currency amounts

### DIFF
--- a/src/pages/Vote/VotePage.tsx
+++ b/src/pages/Vote/VotePage.tsx
@@ -1,7 +1,6 @@
 import { BigNumber } from '@ethersproject/bignumber'
-// eslint-disable-next-line no-restricted-imports
-import { t, Trans } from '@lingui/macro'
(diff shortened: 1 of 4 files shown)


src/pages/Vote/VotePage.tsx
```diff
@@ -146,10 +146,14 @@ export default function VotePage({
     params: { governorIndex, id },
   },
 }: RouteComponentProps<{ governorIndex: string; id: string }>) {
+  const parsedGovernorIndex = Number.parseInt(governorIndex)
+
   const { chainId, account } = useActiveWeb3React()
 
+  const quorum = useQuorum(parsedGovernorIndex)
```
ianlapham, line 153: nit: could we call this quorumAmount

ianlapham (approved):
lgtm, small nit if you think its valid

PR #3435, opened 2022-03-04
fix(error handling): try reloading the app when encountering a javasc…

moodysalem:
…ript error

### DIFF
--- a/src/components/ErrorBoundary/index.tsx
+++ b/src/components/ErrorBoundary/index.tsx
@@ -49,6 +49,11 @@ type ErrorBoundaryState = {
 
 const IS_UNISWAP = window.location.hostname === 'app.uniswap.org'
 
+async function updateServiceWorker(): Promise<void> {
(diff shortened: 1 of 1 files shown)


src/components/ErrorBoundary/index.tsx
```diff
@@ -49,13 +49,25 @@ type ErrorBoundaryState = {
...
 export default class ErrorBoundary extends React.Component<unknown, ErrorBoundaryState> {
   constructor(props: unknown) {
     super(props)
     this.state = { error: null }
   }
 
   static getDerivedStateFromError(error: Error): ErrorBoundaryState {
+    updateServiceWorker()
+      .then(() => {
+        window.location.reload()
```
zzmp, line 66: Do you want to reload, or prompt the user to reload?

NoahZinsmeister:
merging so that the fix in https://github.com/Uniswap/interface/pull/3634 propagates more quickly - can think through reload UX later!

Issue #3359, opened 2022-02-25
disabled lists have weird child opacity on their popovers


JFrankfurt:
![image](https://user-images.githubusercontent.com/5773490/155637777-e2c36c3b-0e5e-484d-93e0-bf0a37331746.png)

ashutosh887:
I would like to work on this  @curly210102 @JFrankfurt

mrchurches:
Hello, can I be assigned to resolve this issue?

PR #3440, opened 2022-03-05
fix(ui): popover with zero token list has weird opacity

curly210102:
Close https://github.com/Uniswap/interface/issues/3359


JFrankfurt:
Can you make the child popover lists retain their non disabled styling?

incorrect:
image

correct:
image



Also, it looks like your changes broke other dropdowns like the network selector:
image


I'm guessing the useOnClickOutside changes are the culprit here

PR #3444, opened 2022-03-06
fix(ui): Auto Router icon not appearing on safari/ios

### LINKED ISSUE #3138: Auto Router icon not appearing on mobile
**Bug Description**
When you expand the transaction details, the Auto Router box is missing the Auto Router icon

**Device**
 iOS latest version. Repro'd in both the Chrome browser and Coinbase browser..

**Steps to Reproduce**
1. app.uniswap.org
2. get a quote like DAI:ETH
3. expand the transaction details

observe there is no Auto Router icon. there should be, to the left of the text "auto router"

IMG_65208C39A18D-1

curly210102:
Closes #3138


JFrankfurt:
should we remove the old svg asset as well, or is it still in use elsewhere?

curly210102:
Yes, It should be removed.

PR #3454, opened 2022-03-07
fix: update type on wrapped native currency

ianlapham:
- currently `WRAPPED_NATIVE_CURRENCY` is typed to return `Token`, however this is not always the case if the indexing key chainId is not a valid key in the mapping

### DIFF
--- a/src/constants/routing.ts
+++ b/src/constants/routing.ts
@@ -46,7 +46,7 @@ type ChainCurrencyList = {
 const WRAPPED_NATIVE_CURRENCIES_ONLY: ChainTokenList = Object.fromEntries(
   Object.entries(WRAPPED_NATIVE_CURRENCY)
     .map(([key, value]) => [key, [value]])
-    .filter((token) => !!token)
+    .filter(Boolean)
 )
 
 // used to construct intermediary pairs for trading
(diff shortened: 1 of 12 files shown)


src/constants/tokens.ts
```diff
@@ -265,7 +265,7 @@ class MaticNativeCurrency extends NativeCurrency {
 
   get wrapped(): Token {
     if (!isMatic(this.chainId)) throw new Error('Not matic')
-    return WRAPPED_NATIVE_CURRENCY[this.chainId]
+    return WRAPPED_NATIVE_CURRENCY[this.chainId] as Token
```
zzmp, line 268: Should you check this? Or use an invariant? Don't cast as it might hide errors from the type system.

src/constants/routing.ts
```diff
@@ -44,7 +44,9 @@ type ChainCurrencyList = {
 }
 
 const WRAPPED_NATIVE_CURRENCIES_ONLY: ChainTokenList = Object.fromEntries(
-  Object.entries(WRAPPED_NATIVE_CURRENCY).map(([key, value]) => [key, [value]])
+  Object.entries(WRAPPED_NATIVE_CURRENCY)
+    .map(([key, value]) => [key, [value]])
+    .filter((token) => !!token)
```
zzmp, line 49: You're just using a boolean filter, so be more explicit if typescript allows you to: `.filter(Boolean)`

zzmp (approved):

JFrankfurt:
Feels weird to be doing type assertions on a const we control.. Would prefer to find a way that does this where `WRAPPED_NATIVE_CURRENCY` is defined

JFrankfurt (approved):

PR #3455, opened 2022-03-07
fix: widget sizing

zzmp:


src/lib/components/Widget.tsx
```diff
@@ -23,14 +23,15 @@ const WidgetWrapper = styled.div<{ width?: number | string }>`
...
+  box-sizing: border-box;
   color: ${({ theme }) => theme.primary};
   display: flex;
   flex-direction: column;
   font-feature-settings: 'ss01' on, 'ss02' on, 'cv01' on, 'cv03' on;
   font-size: 16px;
   font-smooth: always;
   font-variant: none;
-  height: 348px;
+  height: 356px;
```
ianlapham, line 34: is there any way to test this within an integration? i remember changing this didnt expose errors in cosmos but it did when used within other apps 

it affected the bottom amount of padding

ianlapham (commented):
Wondering if we can test change within an integration

JFrankfurt (approved):
🤷‍♀️

src/lib/components/Widget.tsx
zzmp: I manually tested this with widgets-demo to come up with this number.

PR #3459, opened 2022-03-07
fix: routing tooltip styles

zzmp:


src/lib/theme/type.tsx
```diff
@@ -65,7 +65,7 @@ export function Caption(props: TextProps) {
 }
 
 export function Badge(props: TextProps) {
-  return <TextWrapper className="badge" fontSize={8} fontWeight={600} lineHeight="8px" noWrap />
+  return <TextWrapper className="badge" fontSize="8px" fontWeight={600} lineHeight="8px" noWrap {...props} />
```
zzmp, line 68: idk why but passing `{8}` as a number rendered as `72px`, so I passed it in as a string to avoid that.

src/lib/components/Swap/RoutingDiagram/index.tsx
```diff
@@ -14,53 +13,72 @@ import { InterfaceTrade } from 'state/routing/types'
...
-  width: calc(100% - 1em);
   z-index: ${Layer.UNDERLAYER};
 `
 
-const DotsContainerShort = styled(DotsContainer)`
-  overflow: hidden;
-  position: relative;
-  width: 71px;
+const RouteRow = styled(Row)`
+  flex-wrap: nowrap;
```
ianlapham, line 49: does row not take a wrap property?

ianlapham: nvm it doesnt

src/lib/theme/type.tsx
ianlapham: i had this same issue and could not debug either - i had a guess it was something to do with the className

ianlapham (approved):
Looks good

PR #3470, opened 2022-03-08
fix: update default tokens on unsupported networks

ianlapham:
- currentlty the widget will crash when `ExtendedEther` is use as a token input on a network that has no native currency 
- this issue is only exposed through network switching (as we disable token select on those networks)
- when switching from unsupported to supported, a race condition causes ExtendedEther to be used on a network where there is no native currency 

-  to avoid this we should only use "NATIVE" as a token default if we have instantiated the native currency to be used within `ExtendedEther`

### DIFF
--- a/src/lib/hooks/swap/useSyncTokenDefaults.ts
+++ b/src/lib/hooks/swap/useSyncTokenDefaults.ts
@@ -47,12 +47,13 @@ export default function useSyncTokenDefaults({
 }: TokenDefaults) {
   const updateSwap = useUpdateAtom(swapAtom)
   const { chainId } = useActiveWeb3React()
+  const onSupportedNetwork = useOnSupportedNetwork()
   const nativeCurrency = useNativeCurrency()
   const defaultOutputToken = useDefaultToken(defaultOutputTokenAddress, chainId)
   const defaultInputToken =
     useDefaultToken(defaultInputTokenAddress, chainId) ??
     // Default the input token to the native currency if it is not the output token.
-    (defaultOutputToken === nativeCurrency ? nativeCurrency : undefined)
+    (defaultOutputToken === nativeCurrency && onSupportedNetwork ? nativeCurrency : undefined)
 
   const setToDefaults = useCallback(() => {
     const defaultSwapState: Swap = {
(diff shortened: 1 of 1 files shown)


src/lib/hooks/swap/useSyncTokenDefaults.ts
```diff
@@ -26,7 +27,8 @@ function useDefaultToken(
     address = defaultAddress[chainId]
   }
   const token = useToken(address)
-  if (chainId && address === 'NATIVE') {
+  // Only use native currency if chain ID is in supported chains. ExtendedEther will error otherwise.
+  if (chainId && address === 'NATIVE' && chainId in SupportedChainId) {
```
zzmp, line 31: use `useOnSupportedNetwork` instead of reimplementing this logic.

zzmp (changes requested):
Re-use `useOnSupportedNetwork`.
There are two codepaths that use native currency in `useSyncTokenDefaults` - be sure to fix the latter, too (see L46).

ianlapham:
> Re-use `useOnSupportedNetwork`. There are two codepaths that use native currency in `useSyncTokenDefaults` - be sure to fix the latter, too (see L46).

Would both be using the same hook here?

JFrankfurt (approved):

src/lib/hooks/swap/useSyncTokenDefaults.ts
```diff
@@ -41,12 +47,13 @@ export default function useSyncTokenDefaults({
...
   const updateSwap = useUpdateAtom(swapAtom)
   const { chainId } = useActiveWeb3React()
+  const onSupportedNetwork = useOnSupportedNetwork()
   const nativeCurrency = useNativeCurrency()
   const defaultOutputToken = useDefaultToken(defaultOutputTokenAddress, chainId)
   const defaultInputToken =
     useDefaultToken(defaultInputTokenAddress, chainId) ??
     // Default the input token to the native currency if it is not the output token.
-    (defaultOutputToken === nativeCurrency ? nativeCurrency : undefined)
+    (defaultOutputToken === nativeCurrency && onSupportedNetwork ? nativeCurrency : undefined)
```
zzmp, line 56: Looks like I had a bug here - it should be `defaultOutputToken !== nativeCurrency && onSupportedNetwork`.

ianlapham: Yep updating

zzmp (approved):

PR #3471, opened 2022-03-09
fix: incorrectly memoized hooks

zzmp:
Fixes hooks which returned new instances of objects every call, by memoizing the return values to prevent re-renders.

This fixes all hooks that are used in a typical use of `src/lib/components/Swap/SwapButton`. No other codepaths were tested.

### DIFF
--- a/src/hooks/useCurrentBlockTimestamp.ts
+++ b/src/hooks/useCurrentBlockTimestamp.ts
@@ -1,10 +1,20 @@
 import { BigNumber } from '@ethersproject/bignumber'
 import { useSingleCallResult } from 'lib/hooks/multicall'
+import { useEffect, useState } from 'react'
 
 import { useInterfaceMulticall } from './useContract'
 
 // gets the current timestamp from the blockchain
 export default function useCurrentBlockTimestamp(): BigNumber | undefined {
+  const [lastBlock, setLastBlock] = useState<BigNumber | undefined>()
   const multicall = useInterfaceMulticall()
-  return useSingleCallResult(multicall, 'getCurrentBlockTimestamp')?.result?.[0]
+  const block: BigNumber | undefined = useSingleCallResult(multicall, 'getCurrentBlockTimestamp')?.result?.[0]
+  useEffect(() => {
+    // If block xor lastBlock are undefined, or if block has progressed, then update lastBlock.
+    // This prevents updates when the block doesn't change, because the returned BigNumber will still be referentially unique.
+    if (Boolean(block) !== Boolean(lastBlock) || (block && lastBlock && !block.eq(lastBlock))) {
+      setLastBlock(block)
+    }
+  }, [block, lastBlock])
+  return lastBlock
 }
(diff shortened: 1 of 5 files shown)


src/hooks/useUSDCPrice.ts
```diff
@@ -78,17 +78,16 @@ export function useStablecoinAmountFromFiatValue(fiatValue: string | null | unde
   const { chainId } = useActiveWeb3React()
   const stablecoin = chainId ? STABLECOIN_AMOUNT_OUT[chainId]?.currency : undefined
 
-  if (fiatValue === null || fiatValue === undefined || !chainId || !stablecoin) {
-    return undefined
-  }
-
   // trim for decimal precision when parsing
-  const parsedForDecimals = parseFloat(fiatValue).toFixed(stablecoin.decimals).toString()
+  const parsedForDecimals =
```
ianlapham, line 82: Should this be memoized as well?

ianlapham (commented):
Few questions

src/hooks/useUSDCPrice.ts
zzmp: It may be memoized, but it doesn't need to be in order to effect a performance improvement.
I think it's a cheap operation, and the actual re-render is caused by returning distinct objects (from `tryParseCurrencyAmount`), so I'd rather just leave it be.

src/hooks/useCurrentBlockTimestamp.ts
```diff
@@ -1,10 +1,20 @@
 import { BigNumber } from '@ethersproject/bignumber'
 import { useSingleCallResult } from 'lib/hooks/multicall'
+import { useEffect, useState } from 'react'
 
 import { useInterfaceMulticall } from './useContract'
 
 // gets the current timestamp from the blockchain
 export default function useCurrentBlockTimestamp(): BigNumber | undefined {
```
justindomingue, line 8: would it be simpler to make `block` a primitive `Number` instead ?

src/hooks/useCurrentBlockTimestamp.ts
```diff
@@ -1,10 +1,20 @@
...
+  const [lastBlock, setLastBlock] = useState<BigNumber | undefined>()
```
justindomingue, line 9: can you simplify with `usePreviousValue`?

src/hooks/useUSDCPrice.ts
```diff
@@ -78,17 +78,16 @@ export function useStablecoinAmountFromFiatValue(fiatValue: string | null | unde
...
+    (fiatValue && chainId && stablecoin && parseFloat(fiatValue).toFixed(stablecoin.decimals).toString()) || undefined
```
justindomingue, line 83: style wise i think we tend to prefer ternaries in the codebase, nbd

src/hooks/useCurrentBlockTimestamp.ts
zzmp: `usePrevious` will not work because it is still based on referential equality.

src/hooks/useCurrentBlockTimestamp.ts
zzmp: It would require converting it back to `BigNumber` for some uses, so I don't think it simplifies things overall.

justindomingue: i don't think it needs to be a BigNumber at all though if i look at usages 🤷

ianlapham (approved):
Looks good

PR #3472, opened 2022-03-09
fix: restructure web3 to memoize

zzmp:
Refactors web3 usage so that it does not hit the external store with every use of useActiveWeb3React, which was causing multi-second mount times.


ianlapham (approved):
Looks good

PR #3478, opened 2022-03-09
fix: add formatting to USDC values

ianlapham:
- add formatting to input and output USDC values on widget 
- update formatting util to accept fixed decimal cutoff (used to cutoff decimals at 2 for USD prices)

PR #3479, opened 2022-03-09
fix: use group separator to add commas to USD values 

ianlapham:
- this is an alternative solution to #3478 
- uses group separator to add commas without incorporating local formatting


ianlapham:
C

PR #3480, opened 2022-03-09
fix: token list fetching performance

zzmp:
Avoids refetching an already-fetched token list

### DIFF
--- a/src/lib/hooks/useTokenList/index.ts
+++ b/src/lib/hooks/useTokenList/index.ts
@@ -1,7 +1,7 @@
 import { NativeCurrency, Token } from '@uniswap/sdk-core'
 import { TokenInfo, TokenList } from '@uniswap/token-lists'
-import { atom } from 'jotai'
-import { useAtomValue, useUpdateAtom } from 'jotai/utils'
+import { atom, useAtom } from 'jotai'
+import { useAtomValue } from 'jotai/utils'
 import useActiveWeb3React from 'lib/hooks/useActiveWeb3React'
 import resolveENSContentHash from 'lib/utils/resolveENSContentHash'
 import { useCallback, useEffect, useMemo, useState } from 'react'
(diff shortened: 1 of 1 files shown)


src/lib/hooks/useTokenList/index.ts
```diff
@@ -14,31 +14,34 @@ import { validateTokens } from './validateTokenList'
...
-        // TODO(zzmp): Use network resolver when wallet is not on chainId === 1.
         return resolveENSContentHash(ensName, library)
       }
       throw new Error('Could not construct mainnet ENS resolver')
     },
     [chainId, library]
   )
+
   useEffect(() => {
+    if (tokenList.map) return
```
JFrankfurt, line 43: maybe `Array.isArray` instead? checking for method presence feels funky

JFrankfurt: oh, `.map` is not the method, it's a property w/ a `TokenMap` value. THAT's confusing af.

src/lib/hooks/useTokenList/index.ts
```diff
@@ -14,31 +14,34 @@ import { validateTokens } from './validateTokenList'
 
 export const DEFAULT_TOKEN_LIST = 'https://gateway.ipfs.io/ipns/tokens.uniswap.org'
 
-const chainTokenMapAtom = atom<ChainTokenMap | undefined>(undefined)
+const tokenListAtom = atom<{ list?: string | TokenInfo[]; map?: ChainTokenMap }>({})
```
JFrankfurt, line 17: pref change `map` to `tokenMap` and `list` to `tokenInfoList`

JFrankfurt (approved):

src/lib/hooks/useTokenList/index.ts
ianlapham: is `list` the url?

ianlapham (commented):
Wondering about naming

Issue #2808, opened 2021-11-16
Remove useBestV2Trade from the interface


justindomingue:
Only remaining dependency should be https://github.com/Uniswap/interface/blob/main/src/hooks/useUSDCPrice.ts#L28

Few options
* Rely on client-side v3 router only (won't pick up long tail assets)
* Use client-side smart-order router (possible when L2s supported)
* Write simplified logic similar to what we had before

PR #3481, opened 2022-03-09
feat(perf): cache usdc prices per block

zzmp:
The `useUSDCPrice` hook is used throughout both the app and widgets, and nested within other hooks. It is computationally expensive, because it computes a best route to USDC.
Rather than try to move it to a top level (so that `useMemo` would appropriately cache), I implemented a cache external to React for the `useUSDCPrice` hook.

This significantly speeds up rendering when multiple prices are displayed on the screen (as they all call into this hook), and also improves re-rendering an already known pair (eg if you hit the "switch" button twice).

### DIFF
--- a/src/hooks/useUSDCPrice.ts
+++ b/src/hooks/useUSDCPrice.ts
@@ -1,7 +1,8 @@
 import { Currency, CurrencyAmount, Price, Token, TradeType } from '@uniswap/sdk-core'
 import useActiveWeb3React from 'hooks/useActiveWeb3React'
+import useBlockNumber from 'lib/hooks/useBlockNumber'
 import tryParseCurrencyAmount from 'lib/utils/tryParseCurrencyAmount'
-import { useMemo } from 'react'
+import { useEffect, useMemo } from 'react'
 
 import { SupportedChainId } from '../constants/chains'
 import { DAI_OPTIMISM, USDC_ARBITRUM, USDC_MAINNET, USDC_POLYGON } from '../constants/tokens'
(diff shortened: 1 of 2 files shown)


src/hooks/useUSDCPrice.ts
```diff
@@ -17,43 +18,64 @@ export const STABLECOIN_AMOUNT_OUT: { [chainId: number]: CurrencyAmount<Token> }
...
   const amountOut = chainId ? STABLECOIN_AMOUNT_OUT[chainId] : undefined
   const stablecoin = amountOut?.currency
 
   // TODO(#2808): remove dependency on useBestV2Trade
-  const v2USDCTrade = useBestV2Trade(TradeType.EXACT_OUTPUT, amountOut, currency, {
-    maxHops: 2,
-  })
-  const v3USDCTrade = useClientSideV3Trade(TradeType.EXACT_OUTPUT, amountOut, currency)
+  const v2Currency = currency && v2USDCPriceCache.has(currency) ? undefined : currency
+  const v2USDCTrade = useBestV2Trade(TradeType.EXACT_OUTPUT, amountOut, v2Currency, { maxHops: 2 })
```
justindomingue, line 44: a better fix is probably to not rely on the v2 router at all. The client side v3 trade should already be cached for a given polling interval.

zzmp: The v3 trade may be cached, but running the hook still takes 1-3ms, and this hook is invoked in multiple components to compute usdc prices, so this easily causes dropped frames. Also, until the trade is cached, it is computed in each of the components that calls the hook, so it's not just a dropped frame, but ends up being a few 100ms of non-interactivity.

I think a best solution is to do both:
- Just use the v3 trade
- Cache the v3 trade from the first caller, for usdc price only (as this hook is the only entry point AFAIK that is re-used across both app and widget)

src/hooks/useUSDCPrice.ts
```diff
@@ -1,11 +1,11 @@
 import { Currency, CurrencyAmount, Price, Token, TradeType } from '@uniswap/sdk-core'
 import useActiveWeb3React from 'hooks/useActiveWeb3React'
+import useBlockNumber from 'lib/hooks/useBlockNumber'
 import tryParseCurrencyAmount from 'lib/utils/tryParseCurrencyAmount'
-import { useMemo } from 'react'
+import { useEffect, useMemo, useState } from 'react'
 
 import { SupportedChainId } from '../constants/chains'
 import { DAI_OPTIMISM, USDC_ARBITRUM, USDC_MAINNET, USDC_POLYGON } from '../constants/tokens'
-import { useBestV2Trade } from './useBestV2Trade'
```
zzmp, line 8: @justindomingue If this is OK I can remove the code and close #2808 in a follow-up PR.

src/hooks/useUSDCPrice.ts
```diff
@@ -17,43 +17,84 @@ export const STABLECOIN_AMOUNT_OUT: { [chainId: number]: CurrencyAmount<Token> }
...
+  const amountSpecified = chainId ? STABLECOIN_AMOUNT_OUT[chainId] : undefined
+  const stablecoin = amountSpecified?.currency
 
-  // TODO(#2808): remove dependency on useBestV2Trade
-  const v2USDCTrade = useBestV2Trade(TradeType.EXACT_OUTPUT, amountOut, currency, {
-    maxHops: 2,
-  })
-  const v3USDCTrade = useClientSideV3Trade(TradeType.EXACT_OUTPUT, amountOut, currency)
+  const useCache = useUSDCPriceCache(currency)
+  const { trade } = useClientSideV3Trade(TradeType.EXACT_OUTPUT, amountSpecified, useCache ? undefined : currency)
```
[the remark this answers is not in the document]
zzmp: I'd like to not modify this logic in this PR. I put V2 back for now.

zzmp:
Closing for #3523

PR #3484, opened 2022-03-10
fix: summary expando height

zzmp:


JFrankfurt (approved):
magic numbas big spooki 👻

PR #3493, opened 2022-03-10
fix: memoize on-chain results

zzmp:
Memo-izes on-chain non-primitive results by converting them to a string, which is a primitive, and using that in `useMemo`/`useEffect` deps arrays. This prevents unnecessary re-renders when the same result is returned.

### DIFF
--- a/src/hooks/useGasPrice.ts
+++ b/src/hooks/useGasPrice.ts
@@ -23,5 +23,5 @@ export default function useGasPrice(): JSBI | undefined {
   const contract = useContract(address ?? undefined, CHAIN_DATA_ABI, false)
 
   const resultStr = useSingleCallResult(contract, 'latestAnswer').result?.[0]?.toString()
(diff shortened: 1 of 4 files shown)


src/hooks/useGasPrice.ts
```diff
@@ -22,5 +23,5 @@ export default function useGasPrice(): JSBI | undefined {
   const contract = useContract(address ?? undefined, CHAIN_DATA_ABI, false)
 
   const resultStr = useSingleCallResult(contract, 'latestAnswer').result?.[0]?.toString()
-  return typeof resultStr === 'string' ? JSBI.BigInt(resultStr) : undefined
+  return useMemo(() => (resultStr === 'string' ? JSBI.BigInt(resultStr) : undefined), [resultStr])
```
ianlapham, line 26: why aren't we checking `typeof` here? with `resultStr === 'string'`

ianlapham (commented):
Just wondering on this one comparison

ianlapham (approved):
lgtm

PR #3495, opened 2022-03-10
fix: update wrap fail state callback and input values

ianlapham:
This PR contains 4 changes. 

1. Add error catching to the `wrapCallback`. Before, if a user rejected the transaction in their wallet the `WrapState` would not reset the value for `loading`, which causes button to be disabled until page relod. Fix is to catch error within the callback and reset loading state if detected. 

2. Update amount parsing when wrapping and disable trade fetching. When a user is wrapping or unwrapping, we dont want to fetch an ETH <-> WETH trade, and we dont want to show the input/output amounts from that trade. Instead, force input/output amounts to be the typed value from user. 

3. Fix logic for detecting balance in wrapCallback. Currently we are only checking balance if the user has typed in the input field. This leads to an error when users type in the output field, as `sufficientBalance` in `useWrapCallback` is always false if they type in output field. 

4. Dont show wrapping errors if user is not wrapping or unwrapping (in swap button)

### DIFF
--- a/src/lib/components/Swap/SwapButton.tsx
+++ b/src/lib/components/Swap/SwapButton.tsx
@@ -184,7 +184,7 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
   }, [addTransaction, inputTradeCurrencyAmount, outputTradeCurrencyAmount, setDisplayTxHash, swapCallback, tradeType])
 
   const ButtonText = useCallback(() => {
-    if (wrapError !== WrapError.NO_ERROR) {
+    if (wrapError !== WrapError.NO_ERROR && (wrapType === WrapType.WRAP || wrapType === WrapType.UNWRAP)) {
       return <WrapErrorText wrapError={wrapError} />
     }
     switch (wrapType) {
--- a/src/lib/components/Swap/SwapButton.tsx
+++ b/src/lib/components/Swap/SwapButton.tsx
@@ -207,14 +207,16 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
       setActiveTrade(trade.trade)
     } else {
       const transaction = await wrapCallback()
-      addTransaction({
-        response: transaction,
-        type: TransactionType.WRAP,
-        unwrapped: wrapType === WrapType.UNWRAP,
-        currencyAmountRaw: transaction.value?.toString() ?? '0',
-        chainId,
-      })
-      setDisplayTxHash(transaction.hash)
+      if (transaction) {
+        addTransaction({
+          response: transaction,
+          type: TransactionType.WRAP,
+          unwrapped: wrapType === WrapType.UNWRAP,
+          currencyAmountRaw: transaction.value?.toString() ?? '0',
+          chainId,
+        })
+        setDisplayTxHash(transaction.hash)
+      }
     }
   }, [addTransaction, chainId, setDisplayTxHash, trade.trade, wrapCallback, wrapType])
 
--- a/src/lib/components/Swap/SwapButton.tsx
+++ b/src/lib/components/Swap/SwapButton.tsx
@@ -184,7 +184,7 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
   }, [addTransaction, inputTradeCurrencyAmount, outputTradeCurrencyAmount, setDisplayTxHash, swapCallback, tradeType])
 
   const ButtonText = useCallback(() => {
-    if (wrapError !== WrapError.NO_ERROR && (wrapType === WrapType.WRAP || wrapType === WrapType.UNWRAP)) {
+    if ((wrapType === WrapType.WRAP || wrapType === WrapType.UNWRAP) && wrapError !== WrapError.NO_ERROR) {
       return <WrapErrorText wrapError={wrapError} />
     }
     switch (wrapType) {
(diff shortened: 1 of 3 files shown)


src/lib/hooks/swap/useSwapInfo.tsx
```diff
@@ -52,6 +53,10 @@ function useComputeSwapInfo(): SwapInfo {
     useMemo(() => [inputCurrency ?? undefined, outputCurrency ?? undefined], [inputCurrency, outputCurrency])
   )
 
+  // Short circuit trade fetching and amount parsing if wrapping or unwrapping.
+  const { type: wrapType } = useWrapCallback()
+  const isWrapping = wrapType === WrapType.WRAP || wrapType === WrapType.UNWRAP
```
ianlapham, line 58: @zzmp do i need to memoize this?

zzmp: no, it's a primitive type (boolean), so not memoizing will not cause any extra re-renders

src/lib/components/Swap/SwapButton.tsx
```diff
@@ -184,7 +184,7 @@ export default function SwapButton({ disabled }: SwapButtonProps) {
   }, [addTransaction, inputTradeCurrencyAmount, outputTradeCurrencyAmount, setDisplayTxHash, swapCallback, tradeType])
 
   const ButtonText = useCallback(() => {
-    if (wrapError !== WrapError.NO_ERROR) {
+    if (wrapError !== WrapError.NO_ERROR && (wrapType === WrapType.WRAP || wrapType === WrapType.UNWRAP)) {
```
zzmp, line 187: nit: would prefer these to be reversed, as it makes sense to only check for an error if the wrapType is applicable.

src/lib/hooks/swap/useSwapInfo.tsx
```diff
@@ -52,6 +53,10 @@ function useComputeSwapInfo(): SwapInfo {
     useMemo(() => [inputCurrency ?? undefined, outputCurrency ?? undefined], [inputCurrency, outputCurrency])
   )
 
+  // Short circuit trade fetching and amount parsing if wrapping or unwrapping.
```
zzmp, line 56: If you're short-circuiting here, can you remove the short-circuiting in useClientSideSmartOrderRouterTrade? Or are they both necessary?

src/lib/hooks/swap/useWrapCallback.tsx
```diff
@@ -16,7 +16,7 @@ export enum WrapType {
   UNWRAP,
 }
 interface UseWrapCallbackReturns {
-  callback: () => Promise<ContractTransaction>
+  callback: () => Promise<ContractTransaction | void>
```
zzmp, line 19: When would this return Promise<void>?

src/lib/components/Swap/SwapButton.tsx
ianlapham: Yep

src/lib/hooks/swap/useSwapInfo.tsx
ianlapham: I actually didnt know we did that in `useClientSideSmartOrderRouterTrade`, ill remove here instead

src/lib/hooks/swap/useWrapCallback.tsx
ianlapham: We dont need this actually - was leftover from old error catching logic

src/lib/hooks/swap/useWrapCallback.tsx
```diff
@@ -118,7 +111,11 @@ export default function useWrapCallback(): UseWrapCallbackReturns {
     setWrapState((state) => ({ ...state, loading: true }))
     const result = await (wrapType === WrapType.WRAP
       ? wrappedNativeCurrencyContract.deposit({ value: `0x${parsedAmountIn.quotient.toString(16)}` })
-      : wrappedNativeCurrencyContract.withdraw(`0x${parsedAmountIn.quotient.toString(16)}`))
+      : wrappedNativeCurrencyContract.withdraw(`0x${parsedAmountIn.quotient.toString(16)}`)
+    ).catch(() => {
+      setWrapState((state) => ({ ...state, loading: false }))
+      return Promise.reject('User denied wallet transaction.')
```
zzmp, line 117: This may fail for other reasons. Is there a reason you can't just have the actual error returned?

What about replacing the catch with a `.finally(() => setWrapState((state) => ({ ...state, loading: false })))` to deal with resetting the loading state?

src/lib/hooks/swap/useSwapInfo.tsx
zzmp, line 56: Update this comment.

src/lib/hooks/swap/useWrapCallback.tsx
ianlapham: we do this on line 120, but if the request to wallet fails then the `finally` doesnt get triggered. I think this is probably due to the `wait(1)` where its waiting for block confirmation (which doesnt happen if error occurs when sent to wallet)

ianlapham: we could replace this catch with a finally though, which may be more correct (because it can fail for other reasons)

zzmp: I see. Could you keep it as a catch, but rethrow?

```
).catch((e: unknown) => {
  setWrapState(...)
  throw e
})
```

ianlapham: I like this, will update with that

zzmp (approved):

PR #3498, opened 2022-03-11
fix: destructure currencies for ref equality

zzmp:
Destructures the call to useCurrencyBalances. As all currency balances come in, this object will be updated with every balance - however, the destructured balances are referentially stable. Destructuring in this way prevents extra renders, which is important because useComputeSwapInfo is expensive to run, and the re-renders were multiplying that cost.

Rearranges the hook by "theme", but does not change any other logic.


ianlapham (approved):
Looks good, noticed how a lot of logic was leftover from app, but not needed in widget (to address)

PR #3499, opened 2022-03-11
fix(perf): cache pools to avoid cost of instantiation

zzmp:
Pools are all fetched together, but re-instantiated any time any of them loads. This leads to repeated instantiation of the same pools (in a manual test, there were ~60 pools, but over 1000 instantiations when first fetching data, ie a >93% hit rate). Avoiding repeated instantiations shaves off ~15-25ms per data fetch, which leads to displaying swap prices ~45-75ms faster (because pools are fetched for 3 routes: the swap, and stablecoin prices for both coins).

This adds a simple FIFO cache of 128 pools (appx. 2 swaps) to avoid pool reinstantiation.

---

NB: The current bottleneck in rendering is both Pool instantiation and redux-multicall. If this is resolved, redux-multicall will be the bottleneck for rendering.

Reviewers: If you see code that would benefit from more comments/documentation please call them out.

### DIFF
--- a/src/hooks/usePools.ts
+++ b/src/hooks/usePools.ts
@@ -1,9 +1,10 @@
 import { Interface } from '@ethersproject/abi'
-import { Currency, Token } from '@uniswap/sdk-core'
+import { BigintIsh, Currency, Token } from '@uniswap/sdk-core'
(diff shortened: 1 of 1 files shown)


justindomingue (approved):
LGTM

PR #3503, opened 2022-03-11
fix: input amount lag

zzmp:
Propagates the exact amount to the trade currency amount without waiting for the trade to be computed. This allows things like insufficient balance or approval to be reflected in the UI while typing, instead of having a lag.

This screencap shows the issue that is being fixed:
https://user-images.githubusercontent.com/5403956/157987706-dbdbaa6f-8027-4bf6-aeac-ec9330be7cb1.mov


ianlapham (approved):
lgtm

PR #3504, opened 2022-03-11
fix: token button overflow

zzmp:
Removes the overflow property from TokenButton, which prevents the vertical scrollbar from rendering.

See https://www.notion.so/uniswaplabs/y-overflow-causing-scrollbar-on-token-in-output-d5aa67a355694c86b5d9530878a121c2


ianlapham (approved):
lgtm

JFrankfurt (approved):
expected the solution would be making the text line height not be weird, but this is good too 👍

PR #3508, opened 2022-03-12
chore: simplify Price in its own component

zzmp:
- Simplifies the price logic on the toolbar (price is derived from execution price)
- Refactors `Price` to its own component so it can be used in both the toolbar and the summary screen
- Incidentally, add "click to reverse basis" functionality to the summary screen

As part of refactoring price logic, also fixes a bug in `formatLocaleNumber` where `fixedDecimals` was not respected.

### DIFF
--- a/src/lib/components/Swap/Price.tsx
+++ b/src/lib/components/Swap/Price.tsx
@@ -11,10 +11,11 @@ import { TextButton } from '../Button'
 
 interface PriceProps {
   trade: Trade<Currency, Currency, TradeType>
-  quoteUSDC?: CurrencyAmount<Token>
+  outputUSDC?: CurrencyAmount<Token>
 }
 
-export default function Price({ trade, quoteUSDC }: PriceProps) {
+/** Displays the price of a trade. If outputUSDC is included, also displays the unit price. */
+export default function Price({ trade, outputUSDC }: PriceProps) {
   const { i18n } = useLingui()
   const { inputAmount, outputAmount, executionPrice } = trade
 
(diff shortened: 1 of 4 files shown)


willhennessy:
thanks zach. can you explain "(as price may be different in the other direction, and allowing the user to toggle this is misleading)" ?

zzmp:
Considering that this would be divergent from the interface, where you can click to reverse the price quote basis, I restored that functionality, so "(as price may be different in the other direction, and allowing the user to toggle this is misleading)" is no longer relevant.

To simplify the code (and re-use the component), I've adding the "click to reverse basis" functionality to the summary page.

src/lib/components/Swap/Price.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { ThemedText } from 'lib/theme'
+import formatLocaleNumber from 'lib/utils/formatLocaleNumber'
+import { useCallback, useMemo, useState } from 'react'
+import { formatCurrencyAmount, formatPrice } from 'utils/formatCurrencyAmount'
+
+import { TextButton } from '../Button'
+
+interface PriceProps {
+  trade: Trade<Currency, Currency, TradeType>
+  quoteUSDC?: CurrencyAmount<Token>
```
ianlapham, line 14: Could we just add a comment explaining what `quoteUSDC` is? Took me a second to parse everything

zzmp: Renamed to `outputUSDC` to clarify this.

PR #3513, opened 2022-03-14
fix: add check for undefined slot0 and liquidity call results

ianlapham:
- in `usePools` an error can surface when network is changed to an unsupported network 
- in the widget, on unsupported networks `slot0s` and `liquidities` will be empty arrays
- this fix adds a check that the element in the result array exists before attempting to reference it

- may fix notion issue [here ](https://www.notion.so/uniswaplabs/Widgets-UI-Polish-a7c6d4ece6124d4cb96e9be926e9e67f?p=7f1c80a961fd4e19a528ed46fc6afc8a)

### DIFF
--- a/src/hooks/usePools.ts
+++ b/src/hooks/usePools.ts
@@ -106,9 +106,9 @@ export function usePools(
       const [token0, token1, fee] = tokens
 
       if (!slot0s[index]) return [PoolState.INVALID, null]
-      if (!liquidities[index]) return [PoolState.INVALID, null]
-
       const { result: slot0, loading: slot0Loading, valid: slot0Valid } = slot0s[index]
+
+      if (!liquidities[index]) return [PoolState.INVALID, null]
       const { result: liquidity, loading: liquidityLoading, valid: liquidityValid } = liquidities[index]
 
       if (!tokens || !slot0Valid || !liquidityValid) return [PoolState.INVALID, null]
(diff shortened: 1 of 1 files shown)


src/hooks/usePools.ts
```diff
@@ -105,8 +105,18 @@ export function usePools(
       if (!tokens) return [PoolState.INVALID, null]
       const [token0, token1, fee] = tokens
 
-      const { result: slot0, loading: slot0Loading, valid: slot0Valid } = slot0s[index]
-      const { result: liquidity, loading: liquidityLoading, valid: liquidityValid } = liquidities[index]
+      const {
```
zzmp, line 108: nit: Instead of a nullish operator, can you explicitly check for these? eg

```
if (!slot0s[index]) return [PoolState.INVALID, null]
```

This may be easier to read, but will also bring it in line with how tokens are checked directly above it.

zzmp (approved):

src/hooks/usePools.ts
```diff
@@ -105,8 +105,12 @@ export function usePools(
       if (!tokens) return [PoolState.INVALID, null]
       const [token0, token1, fee] = tokens
 
+      if (!slot0s[index]) return [PoolState.INVALID, null]
```
zzmp, line 108: nit: do the check before the destructure to make it apparent why you're doing them:

```
if (!slots0[index]) ...
const {...} = slots0[index]

if (!liquidities[index]) ...
...
```

zzmp (approved):

PR #3515, opened 2022-03-14
feat: make Expando element

zzmp:
Refactors Expando into its own component, which is used in the status dialog and in the summary. Doing so greatly simplifies the dialog and summary.

Fixes https://www.notion.so/uniswaplabs/Swap-button-shifts-down-slightly-when-expanded-details-bc610bbad9364b2fb8e3d8115018d229.

### DIFF
--- a/src/lib/components/Expando.tsx
+++ b/src/lib/components/Expando.tsx
@@ -1,0 +1,80 @@
+import { IconButton } from 'lib/components/Button'
+import Column from 'lib/components/Column'
+import Row from 'lib/components/Row'
+import Rule from 'lib/components/Rule'
+import useScrollbar from 'lib/hooks/useScrollbar'
+import { Expando as ExpandoIcon } from 'lib/icons'
+import styled from 'lib/theme'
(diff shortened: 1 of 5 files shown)


src/lib/components/Expando.tsx
```diff
@@ -0,0 +1,66 @@
...
+import Column from 'lib/components/Column'
+import Row from 'lib/components/Row'
+import Rule from 'lib/components/Rule'
+import useScrollbar from 'lib/hooks/useScrollbar'
+import { Expando as ExpandoIcon } from 'lib/icons'
+import styled from 'lib/theme'
+import { PropsWithChildren, ReactNode, useState } from 'react'
+
+const HeaderColumn = styled(Column)`
+  transition: gap 0.25s;
```
ianlapham, line 11: Do we need this line if we have the same styling applied in `Column` ?

zzmp: We don't. I put it in Column for two use cases, but really it should be special-cased.

ianlapham (approved):
Looks good now

PR #3530, opened 2022-03-16
feat: usePoll

zzmp:


src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -23,18 +25,23 @@ const DistributionPercents: { [key: number]: number } = {
...
 function getConfig(chainId: ChainId | undefined) {
   return {
     // Limit to only V2 and V3.
     protocols: [Protocol.V2, Protocol.V3],
-
     distributionPercent: (chainId && DistributionPercents[chainId]) ?? DEFAULT_DISTRIBUTION_PERCENT,
   }
 }
 
+function useFreshData<T>(data: T, dataBlockNumber: number, maxBlockAge = 10): T | undefined {
```
ianlapham, line 37: Can we reuse this in the other location its needed, for routingapi?

ianlapham (commented):
wondering about reusing hooks, otherwise looks great

src/lib/hooks/usePoll.ts
```diff
@@ -0,0 +1,57 @@
...
+export default function usePoll<T>(
+  fetch: () => Promise<T>,
+  key = '',
+  pollingInterval = DEFAULT_POLLING_INTERVAL,
+  keepUnusedDataFor = DEFAULT_KEEP_UNUSED_DATA_FOR
+): T | undefined {
+  const cache = useMemo(() => new Map<string, { ttl: number; result?: T }>(), [])
+  const [data, setData] = useState<{ key: string; result?: T }>({ key })
+
+  useEffect(() => {
```
justindomingue, line 16: can you add a comment above each useEffect to indicate what variable it tracks? The dependency array contains a few more objects that are used inside the body

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -68,45 +75,25 @@ export default function useClientSideSmartOrderRouterTrade<TTradeType extends Tr
...
-          setResult(result)
-          setLoading(false)
-        }
-      }
-    }
-  }, [queryArgs, params, config, isDebouncing, wrapType])
+  }, [config, params, queryArgs, wrapType])
+  const { data, error } = usePoll(
+    getQuoteResult,
+    useMemo(() => JSON.stringify(queryArgs), [queryArgs])
```
justindomingue, line 92: do you need `useMemo` if you stringify? won't react compare string value?

justindomingue (approved):

src/lib/hooks/usePoll.ts
zzmp: Added a comment to the other useEffect, and a `void key` to enforce it in code.
Did not add a comment to this one, as all tracked variables are used.

PR #3460, opened 2022-03-08
fix: the logic of display injected connectors on mobile app.

naiba:
In the old logic, the return on line 253 would cause wallet apps like TokenPocket and BitizenWallet (built by our team) that provide injected connectors to display an empty list of wallets.

image

So I moved the logic to check for non-mobile and non-mobile exclusive on line 288 inside `isMobile` so that the mobile injected connector can be displayed in the wallet list properly.

Now:

image

PR #3531, opened 2022-03-16
feat: replace manual query with rtk-query

ianlapham:
- This replicated the logic in app for fetching quotes using SOR 
- In this case for widgets, we always want to use the client side SOR package 

- The goal with using RTK here is to allow for polling to get fresh quotes on some interval

### DIFF
--- a/src/lib/hooks/routing/slice.ts
+++ b/src/lib/hooks/routing/slice.ts
@@ -1,0 +1,59 @@
+import { Web3Provider } from '@ethersproject/providers'
+import { createApi, fetchBaseQuery, FetchBaseQueryError } from '@reduxjs/toolkit/query/react'
+import { Protocol } from '@uniswap/router-sdk'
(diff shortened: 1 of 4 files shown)


src/lib/hooks/routing/slice.ts
```diff
@@ -0,0 +1,59 @@
...
+      }
+    >({
+      async queryFn(args, _api, _extraOptions) {
+        let result
+        try {
+          const { params, config } = args
+          result = await getClientSideQuote(args, params, config)
+          return { data: result.data as GetQuoteResult }
+        } catch (e) {
+          // TODO: fall back to client-side quoter when auto router fails.
```
justindomingue, line 46: can remove these two comments

naiba:
Is there anything wrong with this PR https://github.com/Uniswap/interface/pull/3460? Why does no one care about your experience on the mobile app? Why is there no feedback for 5 days? I even looked for developers on discord and twitter and still no response?

PR #3539, opened 2022-03-17
chore: include usdc in swap info

zzmp:
Prevents USDC from being computed multiple times each render. Greatly reduces blocking tasks on the main thread, and minimizes their time (whereas before they occured at every update for 150+ms, now they only occur for >60ms on the final update of a trade).

Removes cruft from computeSwapInfo to make it more maintainable / readable.

### DIFF
--- a/src/lib/hooks/swap/useSwapInfo.tsx
+++ b/src/lib/hooks/swap/useSwapInfo.tsx
@@ -1,36 +1,34 @@
-import { Trans } from '@lingui/macro'
 import { Currency, CurrencyAmount, Percent, TradeType } from '@uniswap/sdk-core'
-import { FeeOptions } from '@uniswap/v3-sdk'
 import { atom } from 'jotai'
 import { useAtomValue, useUpdateAtom } from 'jotai/utils'
 import { useCurrencyBalances } from 'lib/hooks/useCurrencyBalance'
-import { feeOptionsAtom, Field, swapAtom } from 'lib/state/swap'
+import { Field, swapAtom } from 'lib/state/swap'
 import tryParseCurrencyAmount from 'lib/utils/tryParseCurrencyAmount'
-import { ReactNode, useEffect, useMemo } from 'react'
+import { useEffect, useMemo } from 'react'
 import { InterfaceTrade, TradeState } from 'state/routing/types'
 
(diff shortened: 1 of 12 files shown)


src/lib/hooks/swap/useSwapInfo.tsx
```diff
@@ -1,143 +1,98 @@
...
+    () => (isWrapping || !isExactIn ? parsedAmount : trade.trade?.outputAmount),
+    [isExactIn, isWrapping, parsedAmount, trade.trade?.outputAmount]
   )
-  const currencyBalances = useMemo(
-    () => ({
-      [Field.INPUT]: inputCurrencyBalance,
-      [Field.OUTPUT]: outputCurrencyBalance,
-    }),
-    [inputCurrencyBalance, outputCurrencyBalance]
+  const [balanceIn, balanceOut] = useCurrencyBalances(
```
ianlapham, line 60: will this cause that issue we had before with the balance fetching? I could be wrong, but did we destructure the balances on purpose? aka fetch `inputCurrencyBalance`

ianlapham: ah wait nvm, this seems to be the same

src/lib/hooks/swap/useSwapInfo.tsx
```diff
@@ -1,143 +1,98 @@
...
+    account,
+    useMemo(() => [currencyIn, currencyOut], [currencyIn, currencyOut])
   )
 
-  const inputError = useMemo(() => {
-    let inputError: ReactNode | undefined
```
ianlapham, line 87: how are we doing errors here now?

ianlapham (commented):
Generally looks good just couple small questions

src/lib/hooks/swap/useSwapInfo.tsx
zzmp: These errors weren't actually used anywhere, everything is recomputed in the appropriate child components.
This is cruft from porting from interface.

ianlapham: Ahhhhhhh very nice, good catch

ianlapham (approved):
looks good

PR #3540, opened 2022-03-17
fix: only debounce input amounts

zzmp:
Fixes https://www.notion.so/uniswaplabs/Display-thrashing-with-route-data-3ea2ae65beb044e88b3bdce6a61e937c
(May need to be in conjunction with https://github.com/Uniswap/interface/pull/3539).


ianlapham (approved):
lgtm

PR #3543, opened 2022-03-18
fix: swap lag

zzmp:
Fixes layout thrashing and re-implements hysteresis for trades (ie displaying stale trades while loading).


src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -41,22 +41,20 @@ export default function useClientSideSmartOrderRouterTrade<TTradeType extends Tr
   state: TradeState
   trade: InterfaceTrade<Currency, Currency, TTradeType> | undefined
 } {
+  const amount = useMemo(() => amountSpecified?.asFraction, [amountSpecified])
+  const [currencyIn, currencyOut] =
```
ianlapham, line 45: are we not memoizing this on purpose?

zzmp: I don't think we gain anything by memoizing this, because it's just a boolean check.

ianlapham (approved):
Just wondefing about the memo, otherwise looks good.

PR #3548, opened 2022-03-18
feat: tally integration

DhruvJain1122:
Tally Wallet Integration

image


willhennessy:
Hi Dhruv, thanks for the PR. We are very hesitant to add more wallets to this list because it doesn't scale.

1. How does Tally wallet connection work? is it via WalletConnect? is it an injected provider?
2. roughly how many monthly active users or market share does Tally Wallet have?

DhruvJain1122:
@willhennessy 

It’s an Injected provider.

I don’t have exact numbers of users, I’m one of the community developers working on integrations. The wallet is currently released as a Community Edition and will be publicly launched soon, which will be a big event with the DAO launch.

We’re hoping to have our target integrations live at that stage.

willhennessy:
will it work if Tally extension is installed and the user clicks the Metamask button? or is the connection protocol inherently different?

DhruvJain1122:
@willhennessy

Naxsun:
Hi @willhennessy

Working with @DhruvJain1122 and others from our working group on wallet integrations. Would the integration suggested, so having 'Injected' change to 'Tally Ho' for users that have Tally Ho as their default wallet be an option?

Similar as to how it currently changes from 'MetaMask' to Injected now for those users.
Screen Shot 2022-04-11 at 22 14 02

PR #3553, opened 2022-03-19
fix: impact rendering

zzmp:
- Fixes price impact rendering, which was broken when attempting to render the PriceImpact object 😱 
- Fixes the fade-out gradient in WebKit by using `transparent` explicitly (instead of a hex version)

### DIFF
--- a/src/lib/components/Expando.tsx
+++ b/src/lib/components/Expando.tsx
@@ -18,7 +18,7 @@ const ExpandoColumn = styled(Column)<{ height: number; open: boolean }>`
   transition: height 0.25s, padding 0.25s;
 
   :after {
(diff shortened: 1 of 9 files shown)


JFrankfurt:
not super happy about the switch from `priceImpact` to just `impact` (though I see it wasn't all `priceImpact` initially)

JFrankfurt (approved):

zzmp:
> not super happy about the switch from `priceImpact` to just `impact` (though I see it wasn't all `priceImpact` initially)

I just wanted it to be consistent, and generally go for more concise wording. Happy to do either way. Also wouldn't be offended if you put in a PR to update all the names to `priceImpact`, as long as it stays consistent.

PR #3559, opened 2022-03-20
chore: clean dialog mounting logic

zzmp:
- Updates useScrollbar to take effect on mount, instead of waiting a frame for state to be applied
  - This can be seen when clicking into TokenSelect on Safari. Without this fix, it will first render with scrollbars, then re-render when the mounting animation ends.
- Improves unmounting animation detection to work without an animation
  - This improves a codepath, but has no effect on the currently executing codepaths. It is necessary in case `dialog` is ever exposed as a prop.
- General cleanup around dialog unmounting variable names


ianlapham (approved):
Nice fix

PR #3560, opened 2022-03-20
fix: prevent scroll on token select focus

zzmp:
Fixes [jank when mounting TokenSelect](https://www.notion.so/uniswaplabs/safari-Token-selector-animation-wonky-551d5757740d42d990eec081b4ae7cdf) caused by WebKit scrolling the focused element (the search bar) into view during the Dialog mounting animation.


ianlapham (approved):
looks good

PR #3562, opened 2022-03-21
fix: token img jank

zzmp:
Re-render the token image when it changes by setting the alt as a key. This prevents rendering the stale image for a frame while the new image loads.

Alternatively, the currencyId could be used as a key, but it likely should not be re-rendered across chains with different addresses for the same currency name, so alt seems more appropriate.

The issue can be seen in this screencap. This can be reproduced by opening DevTools and disabling the cache.

https://user-images.githubusercontent.com/5403956/159192131-79f9bba9-9cb3-4baf-ae34-ac82143c9f8b.mov


ianlapham (approved):
lg

PR #3563, opened 2022-03-21
fix: trade UI latency

zzmp:
Fixes remaining latency on trade loading UI.

- Fixes useClientSideSmartOrderRouterTrade to return _the trade_, unless it is unavailable, and _then_ return the last trade
  - This fixes a bug where the trade amount took an extra frame to render (visible when reversing currencies on already-loaded trades), because the _last trade_ was always returned first, until it was updated.
- Fixes Input to immediately render any changed value. The validation is removed, so that it now only occurs when the Input changes, not on the initial value that is set.
  - This fixes a bug where the Input would take a frame to update (visible when reversing currencies to a trade that was already loaded, but is no longer, in which case it would flash the old trade then update to 0).


ianlapham (approved):
lgtm

PR #3568, opened 2022-03-21
fix: use a min fresh block

zzmp:
Maintain a map of fresh blocks per chain, and consider trades stale if they are older than this fresh block.
This updates trades immediately after a user-intiated trade goes through, preventing issues like [a failure to submit a second identical trade with high price impact](https://www.notion.so/uniswaplabs/Cannot-submit-second-swap-with-same-settings-as-one-recently-submitted-d6241f64bea9423bbb71170929dede01).


JFrankfurt (approved):
The concept of a "minimum" or "fresh" block seems a little ambiguous, so I'd consider using terms that are a bit more standard if possible (maybe `head`? `lead`?)

src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
```diff
@@ -80,11 +81,13 @@ export default function useClientSideSmartOrderRouterTrade<TTradeType extends Tr
     }
   }, [config, params, queryArgs, wrapType])
 
-  const { data, error } = usePoll(getQuoteResult, JSON.stringify(queryArgs), isDebouncing) ?? {
+  const getIsFreshBlock = useGetIsFreshBlock()
+  const { data: quoteResult, error } = usePoll(getQuoteResult, JSON.stringify(queryArgs), {
+    debounce: isDebouncing,
+    staleCallback: useCallback(({ data }) => getIsFreshBlock(Number(data.blockNumber) || 0), [getIsFreshBlock]),
```
ianlapham, line 87: will `data` always be defined here?

zzmp: No! You're right. This should read `data?.blockNumber`. `useCallback` seems to erase the type.

PR #3569, opened 2022-03-21
fix: approval crash

zzmp:
Removes incorrect invariant from approval checking. See [notion](https://www.notion.so/uniswaplabs/Crash-when-switching-input-tokens-420e089cceb04cbd98400ced48a28e50) for context.


ianlapham (approved):
This is actually okay - we never return the legacy router as the spender for an approval so this error case is no longer valid

PR #3570, opened 2022-03-21
fix: set approval to pending on wallet request

zzmp:
Flips ApprovalState to pending as soon as a wallet interaction is requested, so that the interface does not look hung if the request is going to a mobile wallet. See [notion](https://www.notion.so/uniswaplabs/Wallet-connect-signing-states-88e49d02eb934a4597c7902f3e232a43) for more context.

### DIFF
--- a/src/lib/components/ActionButton.tsx
+++ b/src/lib/components/ActionButton.tsx
@@ -57,7 +57,7 @@ export const Overlay = styled(Row)<{ hasAction: boolean }>`
 export interface Action {
   message: ReactNode
   icon?: Icon
-  onClick: () => void
+  onClick?: () => void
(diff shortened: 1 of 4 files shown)


ianlapham (commented):
Implementation looks good. Question: do we also want this behavior when people are clicking submit on swap (aka they click submit on review page)? 

I guess this may be out of scope because we need a view for that ^^

zzmp:
> Implementation looks good. Question: do we also want this behavior when people are clicking submit on swap (aka they click submit on review page)?

Yes, absolutely, it'll just be a different codepath so I was putting it in a different PR (still working on it, honestly) for ease of review.

PR #3572, opened 2022-03-21
fix: invert stale callback

zzmp:


src/lib/hooks/usePoll.ts
```diff
@@ -35,29 +40,33 @@ export default function usePoll<T>(
...
-    if (entry && entry.ttl + keepUnusedDataFor > Date.now() && !isStale) {
-      // If there is a fresh entry, return it and queue the next poll.
-      setData({ key, result: entry.result })
-      timeout = setTimeout(poll, Math.max(0, entry.ttl - Date.now()))
+    if (entry) {
+      // If there is not a pending fetch (and there should be), queue one.
+      if (entry.ttl) {
+        if (isStale) {
+          poll() // stale results should be refetched immediately
+        } else if (entry.ttl && (entry.ttl + keepUnusedDataFor > Date.now() || isStale)) {
```
ianlapham, line 48: wont `isStale` on this line always be false

ianlapham (commented):
One small question

ianlapham (commented):
Looks like the loading state is broken again (no dimmed stale data) - need to confirm its not because of this change

PR #3580, opened 2022-03-22
fix: retain permit validity for lesser amounts

zzmp:


ianlapham (approved):
lgtm

moodysalem:
there are a couple of concerns with this
- if you permit less than the exact amount, you will not get the storage refund for spending the exact amount (and thus clearing the approval), so it might be slightly more expensive. not relevant for exact output trades though
- if you approve more than the user is spending, they will have leftover approval, which is slightly less safe (i.e., as far as we know this is safe with our existing swap router)

PR #3585, opened 2022-03-23
feat: prompt for interaction "in your wallet"

zzmp:
Prompts the user to complete an interaction (approve, permit, swap, wrap) "in your wallet". Spec'ed out in [notion](https://www.notion.so/uniswaplabs/Wallet-connect-signing-states-88e49d02eb934a4597c7902f3e232a43).

- Updates `ActionButton` to accomodate these prompting states.
- Updates logic for approval/swap/wrap to show these prompts.
  - Of note, adds `animationend` handlers in SwapButton/index.tsx for swap/wrap to avoid layout thrashing from `ActionButton` animations because of the large color difference between the prompt and the button. Specifically, delays the button from reappearing if it will then be immediately covered by a status dialog.

### DIFF
--- a/src/lib/components/Swap/Summary.fixture.tsx
+++ b/src/lib/components/Swap/Summary.fixture.tsx
@@ -41,7 +41,7 @@ function Fixture() {
   return trade ? (
     <Modal color="dialog">
       <SummaryDialog
-        onConfirm={() => void 0}
+        onConfirm={async () => void 0}
         trade={trade}
         slippage={slippage}
         inputUSDC={inputUSDC}
--- a/src/lib/components/Swap/SwapButton/useApprovalData.tsx
+++ b/src/lib/components/Swap/SwapButton/useApprovalData.tsx
@@ -42,7 +42,8 @@ export default function useApprovalData(
     }
     setIsPending(false)
   }, [addTransaction, handleApproveOrPermit])
-  useEffect(() => setIsPending(false), [onApprove])
+  // Reset the pending state if currency changes.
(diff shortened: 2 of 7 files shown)


src/lib/components/Swap/SwapButton/useApprovalData.tsx
```diff
@@ -32,61 +32,57 @@ export default function useApprovalData(
...
 
   const approvalHash = usePendingApproval(currency?.isToken ? currency : undefined, useSwapRouterAddress(trade))
-  const approvalData = useMemo((): Partial<ActionButtonProps> | undefined => {
+  const approvalAction = useMemo((): Action | undefined => {
     if (!trade || !currency) return
 
-    if (approvalState === ApproveOrPermitState.REQUIRES_APPROVAL) {
-      return {
-        action: {
+    switch (approvalState) {
```
zzmp, line 52: Note that only returning from within the `switch` statement ensures that all cases are covered, as the compiler will err if there is a codepath with no return.

src/lib/components/Swap/Summary.fixture.tsx
```diff
@@ -41,7 +41,7 @@ function Fixture() {
   return trade ? (
     <Modal color="dialog">
       <SummaryDialog
-        onConfirm={() => void 0}
+        onConfirm={async () => void 0}
```
JFrankfurt, line 44: what's happening here? (I've never used the void operator, and the docs I just read aren't helping me understand this usage)

PR #3594, opened 2022-03-24
fix: trade displays

zzmp:
- Fix a nit on the toolbar where "insufficient balance" shows while trade is still loading
- Move "lastTrade" logic to `useBestTrade` so that it applies to both SOR and V3 router results
  - This prevents jank from occurring when using the V3 router (because SOR has failed)
  - This can be reproduced by trading large amounts of MATIC to AAVE on polygon, using metamask (which restricts gas usage for these queries)


src/lib/components/Swap/Toolbar/index.tsx
```diff
@@ -38,22 +37,23 @@ export default memo(function Toolbar({ disabled }: { disabled?: boolean }) {
...
+        return <Caption.InsufficientLiquidity />
+      }
+      if (trade?.inputAmount && trade.outputAmount) {
+        return <Caption.Trade trade={trade} outputUSDC={outputUSDC} impact={impact} />
       }
-      if (trade) {
-        if (!trade.swaps) {
-          return <Caption.InsufficientLiquidity />
-        }
-        if (trade.inputAmount && trade.outputAmount) {
```
JFrankfurt, line 54: confirming the re-ordering here is intentional

JFrankfurt: ah, I see that it is. disregard

src/lib/hooks/swap/useBestTrade.ts
```diff
@@ -28,5 +30,24 @@ export function useBestTrade(
...
+  return useMemo(() => {
+    if ((trade.state !== TradeState.LOADING && trade.state !== TradeState.SYNCING) || trade.trade) return trade
+
+    // Dont return last trade if currencies dont match.
+    const isStale =
+      (currencyIn && !lastTrade?.inputAmount?.currency.equals(currencyIn)) ||
+      (currencyOut && !lastTrade?.outputAmount?.currency.equals(currencyOut))
+    if (isStale) return trade
+
+    return { state: trade.state, trade: lastTrade }
```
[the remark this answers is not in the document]
JFrankfurt: looks like it's just a naming issue--can you find a way to make the state, trade, trade.state stuff a little more clear?

JFrankfurt (approved):

PR #3595, opened 2022-03-24
fix: memoize client side v3 router call data

zzmp:
Memoizes the call data passed into the V3 router. Without this fix, it is possible that the max stack depth will be exceeded, because the router will be requeried every frame (because the call data is constructed anew each render).

### DIFF
--- a/src/hooks/useClientSideV3Trade.ts
+++ b/src/hooks/useClientSideV3Trade.ts
@@ -4,7 +4,7 @@ import { SupportedChainId } from 'constants/chains'
 import useActiveWeb3React from 'hooks/useActiveWeb3React'
 import JSBI from 'jsbi'
 import { useSingleContractWithCallData } from 'lib/hooks/multicall'
-import { useMemo } from 'react'
+import { useEffect, useMemo } from 'react'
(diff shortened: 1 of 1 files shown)


src/hooks/useClientSideV3Trade.ts
```diff
@@ -28,26 +28,25 @@ export function useClientSideV3Trade<TTradeType extends TradeType>(
...
+    () =>
+      amountSpecified
+        ? routes.map((route) => SwapQuoter.quoteCallParameters(route, amountSpecified, tradeType).calldata)
+        : [],
+    [amountSpecified, routes, tradeType]
   )
+  const quotesResults = useSingleContractWithCallData(quoter, callData, {
+    gasRequired: chainId ? QUOTE_GAS_OVERRIDES[chainId] ?? DEFAULT_GAS_QUOTE : undefined,
+  })
+  useEffect(() => console.log('quotesResults'), [quotesResults])
```
zzmp, line 49: Whoops - should have removed this. I'll leave it in for now so that you can see when this is called if you want to test the vercel, and remove it before I merge.

justindomingue (approved):
LGTM! (minus `useEffect`)

PR #3598, opened 2022-03-24
fix: tick before returning quote

zzmp:
After fetching quote results, the SOR package takes >80ms to process the data (on a production build), resulting in dropped frames. This isn't _terrible_, but we then trigger a re-render by setting the quoteResult using a `useState` hook, tacking on an additional 40-80ms. This is a performance improvement which delays the re-render by a tick, allowing the browser to render a frame sooner and making the page feel snappier for integrations with other things happening on the page (like FWB's background graphic).

### DIFF
--- a/src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
+++ b/src/lib/hooks/routing/useClientSideSmartOrderRouterTrade.ts
@@ -1,3 +1,5 @@
+import 'setimmediate'
+
 import { Protocol } from '@uniswap/router-sdk'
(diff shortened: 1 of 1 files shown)


JFrankfurt (approved):
Is there a substantial diff between this and `setTimeout(fn, 0)`?

zzmp:
> Is there a substantial diff between this and `setTimeout(fn, 0)`?

No! It may even be the same as a resolved resolved promise, but I'd have to check. I just use setImmediate because it reads clearer to me.

PR #3599, opened 2022-03-24
feat: migrate to GA4

cmcewen:


src/hooks/useFeeTierDistribution.ts
```diff
@@ -112,9 +112,7 @@ function usePoolTVL(token0: Token | undefined, token1: Token | undefined) {
     }
 
     if (latestBlock - (_meta?.block?.number ?? 0) > MAX_DATA_BLOCK_AGE) {
-      ReactGA.exception({
-        description: `Graph stale (latest block: ${latestBlock})`,
-      })
+      ReactGA.event('exception', { description: `Graph stale (latest block: ${latestBlock})` })
```
zzmp, line 115: Does this need a 'fatal' property?

zzmp (approved):
This looks like a pretty straightforward migration. Testing is hard because you only see GA events after a delay to aggregate them, so I'd just merge this, wait a few days, and re-check the GA dashboard.

src/hooks/useFeeTierDistribution.ts
cmcewen: It's not required so I read that as default to false, but happy to change if you prefer https://developers.google.com/analytics/devguides/collection/ga4/exceptions#exception_parameters

PR #3606, opened 2022-03-26
feat: flashbots protect integration

zeroXbrock:
Implements Flashbots Protect RPC check into UI.

hijacked [this PR](https://github.com/Uniswap/interface/pull/2713) from @Dire-0x. Not sure I resolved merge conflicts perfectly, but I addressed the comments raised in the original PR to the extent that made sense to me.

@justindomingue I wasn't able to get rtk-query to work (can't use callbacks in a non-component function) -- maybe you can provide some supporting info on how this should work?

The last commits (from Dire) on this were around 5 months ago, so it would be great to get a fresh review on this.

Thanks all! Tagging the reviewers from the previous PR: @justindomingue @moodysalem @JFrankfurt


src/state/transactions/updater.tsx
```diff
@@ -1,20 +1,71 @@
...
+  lastCheckedBlockNumber?: number
+}
+
+export function checkPrivateTxStatusFailed(txHash: string): Promise<boolean> {
+  return new Promise((resolve) => {
+    axios.get(`https://protect.flashbots.net/tx/${txHash}`).then((result) => {
+      return resolve(result?.data?.status === 'FAILED' || result?.data?.status === 'FAILED_BUNDLE')
+    })
+  })
+}
```
JFrankfurt, line 26: Could drop axios here and simplify this call a bit.

```
export function checkPrivateTxStatusFailed(txHash: string): Promise<boolean> {
  return fetch(`https://protect.flashbots.net/tx/${txHash}`)
    .then((result) => result?.data?.status === 'FAILED' || result?.data?.status === 'FAILED_BUNDLE'))
}
```

src/components/Header/FlashbotsProtectStatus.tsx
```diff
@@ -0,0 +1,196 @@
...
+      {open && (
+        <FlyoutMenu>
+          <FlyoutHeader>
+            <Trans>
+              Sandwich Protection: <StatusText active={isFlashRPC}>{isFlashRPC ? 'ON' : 'OFF'}</StatusText>
+            </Trans>
+          </FlyoutHeader>
+          {!isFlashRPC && (
+            <FlyoutRow>
+              <Trans>Connect to the RPC</Trans>
```
justindomingue, line 176: the values probably should not be translated. We use Google Translate as first step and then our community picks up the translations to approve and confirm.

src/state/transactions/updater.tsx
```diff
@@ -1,20 +1,71 @@
...
+const DEFAULT_RETRY_OPTIONS: RetryOptions = { n: 1, minWait: 0, maxWait: 0 }
+
 export default function Updater() {
-  const { chainId } = useActiveWeb3React()
+  const { chainId, library } = useActiveWeb3React()
+
+  const lastBlockNumber = useBlockNumber()
+
+  const dispatch = useAppDispatch()
```
JFrankfurt, line 59: can drop this diff in favor of the old line next to its usage

src/state/transactions/updater.tsx
```diff
@@ -1,20 +1,71 @@
...
+
+export function checkPrivateTxStatusFailed(txHash: string): Promise<boolean> {
+  return new Promise((resolve) => {
+    axios.get(`https://protect.flashbots.net/tx/${txHash}`).then((result) => {
+      return resolve(result?.data?.status === 'FAILED' || result?.data?.status === 'FAILED_BUNDLE')
+    })
+  })
+}
+
+export function shouldCheck(lastBlockNumber: number, tx: TxInterface): boolean {
```
justindomingue, line 28: does this need to be exported?

vm:
Think we can close this for now, since a good amount of the codebase has changed since implementation

PR #3609, opened 2022-03-27
feat: Added lido steth wrap/unwrap functionality

dylan5394:
Added wrap/unwrap functionality for Lido stETH.

Certain code could be modularized in the future to more easily support wrap/unwrap of additional tokens though given this may be an uncommon use case and in an effort to keep complexity low for my first contribution I opted not to do this. If it is predicted to be a common future use case I would be happy to abstract some of the hard coded instances of weth/steth contracts and naming.

Wrap token approval is handled in the useWrapCallback hook since the approval hooks in Swap/index.tsx are focused on SwapRouter approval. In the event of adding wrap support for other tokens, the wrap approval complexity can remain encapsulated in the useWrapCallback hook without adding any code to Swap/index.tsx.

I did not add swap functionality in the widget components as it did not seem in scope of the ticket.

### DIFF
--- a/src/abis/wsteth.json
+++ b/src/abis/wsteth.json
@@ -1,0 +1,202 @@
+[
+  {
+    "inputs": [{ "internalType": "contract IStETH", "name": "_stETH", "type": "address" }],
+    "stateMutability": "nonpayable",
+    "type": "constructor"
+  },
+  {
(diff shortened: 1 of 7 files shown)


mattbspector:

PR #3627, opened 2022-03-30
feat(widgets): empty token list on network alert

JFrankfurt:
Adds an empty state to the token selector and a configuration option for token lists to the swap fixture.

image

image


src/lib/components/TokenSelect/NoTokensAvailableOnNetwork.tsx
```diff
@@ -0,0 +1,32 @@
...
+const Text = styled(ThemedText.Body1)`
+  text-align: center;
+`
+
+const Wrapper = styled(ColumnCenter)`
+  height: 80%;
+  justify-content: center;
+`
+
+function NoTokensAvailableOnNetwork() {
```
zzmp, line 22: Should be `export default function ...` for consistency within lib.

src/lib/components/TokenSelect/index.tsx
```diff
@@ -72,45 +88,59 @@ export function TokenSelectDialog({ value, onSelect }: TokenSelectDialogProps) {
...
-      </Column>
-      {isLoaded ? (
-        tokens.length ? (
-          <TokenOptions tokens={tokens} onSelect={onSelect} ref={setOptions} />
-        ) : (
-          <Column padded>
-            <Row justify="center">
-              <ThemedText.Body1 color="secondary">
-                <Trans>No results found.</Trans>
+      {listHasMoreTokensThanJustNativeAsset || !isLoaded ? (
```
zzmp, line 101: This inline logic is getting hard to follow. Can you break it into a
```
const Body = useCallback(() => {
  if (listHasTokens) {
    ...
  }
  return <NoTokensAvailableOnNetwork />
}, [])
```

src/lib/components/TokenSelect/index.tsx
```diff
@@ -45,10 +46,25 @@ interface TokenSelectDialogProps {
 }
 
 export function TokenSelectDialog({ value, onSelect }: TokenSelectDialogProps) {
+  const { chainId } = useActiveWeb3React()
```
zzmp, line 49: I don't think we should be clearing the query on chain changes - it adds complexity that we don't need to handle, and is out of scope of this change.

src/lib/components/TokenSelect/index.tsx
```diff
@@ -45,10 +46,25 @@ interface TokenSelectDialogProps {
...
+  const [previousChainId, setPreviousChainId] = useState(chainId)
+
   const [query, setQuery] = useState('')
   const queriedTokens = useQueryCurrencies(query)
   const tokens = useMemo(() => queriedTokens?.filter((token) => token !== value), [queriedTokens, value])
```
zzmp, line 54: You can't base `listHasMoreTokensThanJustNativeAsset` on `tokens` because it's already filtered by the query, and no tokens for a query has a different UI ("No results found"). I think you'll need to update useQueryCurrencies to take a token list, and then get the token list here to check if there are tokens on the network.

Also not that native asset is not technically a token, so you could feasibly call it something like `listHasTokens`.

src/lib/components/TokenSelect/NoTokensAvailableOnNetwork.tsx
```diff
@@ -0,0 +1,32 @@
...
+  return (
+    <Wrapper>
```
zzmp, line 24: Use a `Column` instead of `Wrapper` to avoid custom styling. You can use gap to add the spacing, and you don't need to specify a height if it doesn't overflow.

src/lib/components/TokenSelect/NoTokensAvailableOnNetwork.tsx
```diff
@@ -0,0 +1,32 @@
...
+      <HelpCircleIcon />
+      <Text color="primary">
```
zzmp, line 26: Use a `Row` to center this content, instead of custom styling. Wrap the `ThemedText.Body1` in the `Row` to ensure the font is correct.

src/lib/components/TokenSelect/NoTokensAvailableOnNetwork.tsx
```diff
@@ -0,0 +1,32 @@
+import { Trans } from '@lingui/macro'
+import { ColumnCenter } from 'components/Column'
+import { HelpCircle } from 'lib/icons'
+import styled, { ThemedText } from 'lib/theme'
+
+const HelpCircleIcon = styled(HelpCircle)`
```
zzmp, line 6: Can you use `LargeIcon` here instead of custom styling?

zzmp (changes requested):

src/lib/components/TokenSelect/index.tsx
zzmp: Alternatively, you can move the Header into the parent component, and have this entire componet just be the body to simplify a little. Then you can have early returns. I think that'd make this component a lot easier to read.

src/lib/components/TokenSelect/index.tsx
```diff
@@ -65,55 +65,54 @@ export function TokenSelectDialog({ value, onSelect }: TokenSelectDialogProps) {
...
-            />
-          </ThemedText.Body1>
-        </Row>
-        {Boolean(baseTokens.length) && (
-          <Row pad={0.75} gap={0.25} justify="flex-start" flex>
-            {baseTokens.map((token) => (
-              <TokenBase value={token} onClick={onSelect} key={currencyId(token)} />
-            ))}
+  const { chainId } = useActiveWeb3React()
+  const list = useTokenList()
```
zzmp, line 74: Remove `useQueryCurrencies` from `useTokenList/index.ts`, and instead re-export `useQueryTokens` from `useTokenList/querying.ts` (`export { useQueryTokens } from './querying'`), which takes the token list as a parameter. This allows re-use of the same token list for the "no tokens" check that you use in the "query" filter, and will be more performant.

Alternatively, instead of re-exporting, you can move `useQueryTokens` to its own file, maybe `useQueryTokenList.ts` (with a rename)?

src/lib/components/TokenSelect/index.tsx
```diff
@@ -65,55 +65,54 @@ export function TokenSelectDialog({ value, onSelect }: TokenSelectDialogProps) {
...
+  const listHasTokens = useMemo(() => {
+    return Boolean(list.find((token) => token.chainId === chainId && !token.isNative))
+  }, [chainId, list])
+
+  if (listHasTokens || !isLoaded) {
```
zzmp, line 79: It may be easier to read if you express this as a guard:

```
if (isLoaded && !listHasTokens) {
  return (
    <TokenOptionsSkeleton />
  )
}
```

src/lib/components/TokenSelect/index.tsx
```diff
@@ -140,6 +139,7 @@ export default memo(function TokenSelect({ value, collapsed, disabled, onSelect
       <TokenButton value={value} collapsed={collapsed} disabled={disabled} onClick={onOpen} />
       {open && (
         <Dialog color="module" onClose={() => setOpen(false)}>
+          <DialogHeader title={<Trans>Select a token</Trans>} />
```
zzmp, line 142: Moving the header here breaks the TokenSelect.fixture.tsx. Instead, how about a new

```
function TokenSelectDialog() {
  return (
    <>
      <Header ...>
      <TokenSelectBody />
    <>
  )
}
```

with the existing logic moved into `TokenSelectBody` (or something similar).

zzmp (changes requested):

src/lib/components/Swap/Swap.fixture.tsx
```diff
@@ -57,6 +59,35 @@ function Fixture() {
   })
   const [defaultOutputAmount] = useValue('defaultOutputAmount', { defaultValue: 0 })
 
+  const tokenListNameMap: Record<string, TokenInfo[] | string> = {
+    'default list': tokens,
+    'mainnet only': tokens.filter((token) => SupportedChainId.MAINNET === token.chainId),
+    'arbitrum only': [
```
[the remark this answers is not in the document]
zzmp: I don't know that any of these are necessary considering that arbitrum already has no tokens, so this edge case can already be triggered in cosmos.

src/lib/components/TokenSelect/index.tsx
```diff
@@ -65,55 +65,54 @@ export function TokenSelectDialog({ value, onSelect }: TokenSelectDialogProps) {
...
-        </Row>
-        {Boolean(baseTokens.length) && (
-          <Row pad={0.75} gap={0.25} justify="flex-start" flex>
-            {baseTokens.map((token) => (
-              <TokenBase value={token} onClick={onSelect} key={currencyId(token)} />
-            ))}
+  const { chainId } = useActiveWeb3React()
+  const list = useTokenList()
+  const listHasTokens = useMemo(() => {
+    return Boolean(list.find((token) => token.chainId === chainId && !token.isNative))
```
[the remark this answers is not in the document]
JFrankfurt: I had a vague recollection that you'd recently mentioned some array method that was better for this case when I was writing this code, actually. `some` it is.

src/lib/components/Swap/Swap.fixture.tsx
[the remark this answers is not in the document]
JFrankfurt: That requires the viewer to have arbitrum already available to them, which isn't necessarily the case if they are using metamask.

PR #3628, opened 2022-03-30
feat: upgrade web3-react

zzmp:
- Upgrades @web3-react
- Updates the Web3Provider to be a context-based provider
  - Improves performance, but also a refactor was necessary with @web3-react updates
- Replaces the Eip1193Bridge with a JsonRpcConnector
- Allows a JsonRpcProvider for the jsonRpcEndpoint, as this is natively supported by @web3-react/url


src/lib/utils/JsonRpcConnector.ts
```diff
@@ -0,0 +1,39 @@
+import { JsonRpcProvider } from '@ethersproject/providers'
+import { Actions, Connector, ProviderConnectInfo, ProviderRpcError } from '@web3-react/types'
+
+function parseChainId(chainId: string) {
+  return Number.parseInt(chainId, 16)
+}
+
+export default class JsonRpcConnector extends Connector {
+  constructor(actions: Actions, public customProvider: JsonRpcProvider) {
```
NoahZinsmeister, line 9: does this `public` syntax automatically set the class attribute? interesting...

src/lib/hooks/useActiveWeb3React.tsx
```diff
@@ -1,71 +1,94 @@
...
+  const injectedConnector = useConnector(injectedConnectorAtom, Injected, provider)
   const urlConnector = useConnector(urlConnectorAtom, Url, jsonRpcEndpoint)
-  const priorityConnector = getPriorityConnector(injectedConnector, urlConnector)
-  const priorityProvider = priorityConnector.usePriorityProvider()
-  const priorityWeb3React = priorityConnector.usePriorityWeb3React(priorityProvider)
-  const setWeb3 = useUpdateAtom(web3Atom)
-  useEffect(() => {
-    setWeb3(priorityWeb3React)
-  }, [priorityWeb3React, setWeb3])
+  const [connector, hooks] = injectedConnector[1].useIsActive() ? injectedConnector : urlConnector ?? EMPTY_CONNECTOR
```
NoahZinsmeister, line 69: is it ok that these hooks are called conditionally (swapped out at runtime)? the benefit of the `usePriority*` approach is that all hooks remain mounted at all times

zzmp: My understanding is that the hooks must be invoked in the same order, but these hooks will always be populated and always be called in the same order (even though they may change between renders), so I _think_ this is fine, and it functions correctly when tested.

zzmp: I cannot use the `usePriority` approach because the providers may be swapped out during runtime.

NoahZinsmeister: got it, okay! yeah i know exactly what you mean, i was never able to get a firm answer as to whether this specifically was ok or not, but if it works it works

NoahZinsmeister (approved):
lgtm!

PR #3633, opened 2022-03-31
fix: bug where user cannot burn lp position if fetching fee values fails.

guil-lambert:
Sometimes the `positionManager.callStatic.collect()` call fails in hooks/useV3PositionFees.ts. 

When this happens, the `useDerivedV3BurnInfo(position, receiveWETH)` function on line `87` of pages/RemoveLiquidity/V3.tsx returns `undefined` for `feeValue0` and `feeValue1` 

If `feeValue0` and `feeValue1` is `undefined`, then the `burn = useCallback` function on line 103  fails because `!feeValue0` and `!feeValue0` return `True`. So LPs cannot burn their position if the `collect()` call fails (which happens quite frequently for me)

To solve this, I removed the `!feeValue0` and `!feeValue1` conditions lines on 112+113, and also replaced `feeValue0`+`feeValue1` with `liquidityValue0`+`liquidityValue1` for `expectedCurrencyOwed0`+`expectedCurrencyOwed1` on lines 127 and 128. 

Maybe the `expectedCurrencyOwed0` and `expectedCurrencyOwed1` could be set to 0 as in https://docs.uniswap.org/sdk/guides/liquidity/removing#removing-liquidity-from-a-position (?)

### DIFF
--- a/src/hooks/useV3PositionFees.ts
+++ b/src/hooks/useV3PositionFees.ts
@@ -23,13 +23,11 @@ export function useV3PositionFees(
   const tokenIdHexString = tokenId?.toHexString()
   const latestBlockNumber = useBlockNumber()
 
-  // TODO find a way to get this into multicall
+  // we can't use multicall for this because we need to simulate the call from a specific address
   // latestBlockNumber is included to ensure data stays up-to-date every block
-  const [amounts, setAmounts] = useState<[BigNumber, BigNumber]>()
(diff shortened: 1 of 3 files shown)


NoahZinsmeister:
thanks for this @guil-lambert ! took the liberty of implementing a slightly different approach - falling back to 0 fees is indeed the correct behavior here, which i've introduced in the case when fees fail to fetch for whatever reason

JFrankfurt (approved):

PR #3643, opened 2022-04-01
fix: walletconnect numeric chain id

zzmp:
This should be backed out once it is fixed upstream.


JFrankfurt (approved):
Heads up @NoahZinsmeister

PR #3657, opened 2022-04-04
fix: stale data edge cases

zzmp:
- Simplifies block and swap info updaters
  - Fixes block updater to not return another chain's block number
- Support chains with no token list (eg to show a balance)
  - Gate unnecessary RPCs called by swap info updater

### DIFF
--- a/src/hooks/useUSDCPrice.ts
+++ b/src/hooks/useUSDCPrice.ts
@@ -24,7 +24,7 @@ export const STABLECOIN_AMOUNT_OUT: { [chainId: number]: CurrencyAmount<Token> }
 export default function useUSDCPrice(currency?: Currency): Price<Currency, Token> | undefined {
   const chainId = currency?.chainId
 
-  const amountOut = currency && chainId ? STABLECOIN_AMOUNT_OUT[chainId] : undefined
+  const amountOut = chainId ? STABLECOIN_AMOUNT_OUT[chainId] : undefined
   const stablecoin = amountOut?.currency
 
   // TODO(#2808): remove dependency on useBestV2Trade
--- a/src/hooks/useUSDCPrice.ts
+++ b/src/hooks/useUSDCPrice.ts
@@ -24,7 +24,7 @@ export const STABLECOIN_AMOUNT_OUT: { [chainId: number]: CurrencyAmount<Token> }
 export default function useUSDCPrice(currency?: Currency): Price<Currency, Token> | undefined {
   const chainId = currency?.chainId
 
-  const amountOut = chainId ? STABLECOIN_AMOUNT_OUT[chainId] : undefined
+  const amountOut = currency && chainId ? STABLECOIN_AMOUNT_OUT[chainId] : undefined
   const stablecoin = amountOut?.currency
 
   // TODO(#2808): remove dependency on useBestV2Trade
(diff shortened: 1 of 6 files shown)


src/hooks/useUSDCPrice.ts
```diff
@@ -24,7 +24,7 @@ export const STABLECOIN_AMOUNT_OUT: { [chainId: number]: CurrencyAmount<Token> }
 export default function useUSDCPrice(currency?: Currency): Price<Currency, Token> | undefined {
   const chainId = currency?.chainId
 
-  const amountOut = chainId ? STABLECOIN_AMOUNT_OUT[chainId] : undefined
+  const amountOut = currency && chainId ? STABLECOIN_AMOUNT_OUT[chainId] : undefined
```
justindomingue, line 27: does this actually prevent any work? Won't `v2USDCTrade` and `v3USDCTrade` skip the query when `currency` is undefined, and `useMemo` has an if check on `currency` early

src/lib/hooks/swap/useSwapInfo.tsx
```diff
@@ -95,23 +104,15 @@ function useComputeSwapInfo(): SwapInfo {
...
 }
 
-const swapInfoAtom = atom<SwapInfo>({
-  [Field.INPUT]: {},
-  [Field.OUTPUT]: {},
-  trade: { state: TradeState.INVALID },
-  slippage: { auto: true, allowed: new Percent(0) },
-})
-
 export function SwapInfoUpdater() {
```
justindomingue, line 114: hm why are we using a swap info atom here instead of just using the return value? that hook shouldn't be called from multiple places?

src/lib/hooks/swap/useSwapInfo.tsx
```diff
@@ -45,8 +45,8 @@ function useComputeSwapInfo(): SwapInfo {
   )
   const trade = useBestTrade(
     isExactIn ? TradeType.EXACT_INPUT : TradeType.EXACT_OUTPUT,
-    parsedAmount,
-    (isExactIn ? currencyOut : currencyIn) ?? undefined
+    currencyIn && currencyOut && parsedAmount,
+    (currencyIn && currencyOut && (isExactIn ? currencyOut : currencyIn)) ?? undefined
   )
```
JFrankfurt, line 50: could this result in a weird behavior where i.e. `currencyIn && currencyOut && parsedAmount` resolves to the value at `currencyOut` when `parsedAmount` is falsy?

src/lib/hooks/useBlockNumber.ts
```diff
@@ -1,32 +1,38 @@
...
     (block: number) => {
-      setState((state) => {
-        if (state.chainId === chainId) {
-          if (typeof state.block !== 'number') return { chainId, block }
-          return { chainId, block: Math.max(block, state.block) }
+      setChainBlock((chainBlock) => {
+        if (chainBlock.chainId === chainId) {
+          if (chainBlock.block === block) return chainBlock
+          if (typeof chainBlock.block !== 'number') return { chainId, block }
+          return { chainId, block: Math.max(block, chainBlock.block) }
```
JFrankfurt, line 24: Do you recall why the `Math.max` is used here? It feels like it could be wrong to me. Why would we be in a situation where `.max` determines correctness?

src/lib/hooks/swap/useSwapInfo.tsx
zzmp: yes!

src/lib/hooks/useBlockNumber.ts
zzmp: my assumption is that a provider may rewind given its sources, and that this prevents that from being reflected in our ui, although this was copied over from interface code so idk for sure.

PR #3661, opened 2022-04-05
fix: missing token balance

alexdorsch:
Increases the gas requirement to 200,000 when calling the `balanceOf` function on a token.
Ensures that the token balance and "MAX" button appear in the Swap input component.

Tested on token 0xab167E816E4d76089119900e941BEfdfA37d6b32 which requires ~173k gas to read `balanceOf` with multicall.

### DIFF
--- a/src/lib/hooks/useCurrencyBalance.ts
+++ b/src/lib/hooks/useCurrencyBalance.ts
@@ -47,7 +47,7 @@ export function useNativeCurrencyBalances(uncheckedAddresses?: (string | undefin
 }
 
 const ERC20Interface = new Interface(ERC20ABI) as Erc20Interface
-const tokenBalancesGasRequirement = { gasRequired: 200_000 }
+const tokenBalancesGasRequirement = { gasRequired: 185_000 }
 
 /**
  * Returns a map of token addresses to their eventually consistent token balances for a single account.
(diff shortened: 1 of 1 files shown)


zzmp:
Thanks @alexdorsch! Do you have any documentation that you can link supporting the increased gas cost for getting this token's balance?

alexdorsch:
@zzmp Here is the test I used to read the token balance with multicall, which outputs the gas usage:
https://github.com/alexdorsch/uniswap-multicall-test

image

zzmp:
Would it be reasonable to change the gas limit to 175_000 instead of 200_000? Is there a risk (with 200_000) of multiple calls exceeding a node's gas cap?

alexdorsch:
@zzmp My only concern would be that for this specific token, the gas used by `balanceOf` can fluctuate a bit due to the way reflections are calculated in the contract. Perhaps it could be lowered to 185_000?

According to the docs, Infura limits `eth_call` to 10x the current block gas limit. If my math is correct, this would allow for about 1500 batched token balances at 200_000 (as of today)

https://docs.infura.io/infura/networks/ethereum/json-rpc-methods/eth_call

zzmp:
Great - let's make it 185_000 and then I'll merge this in. Thank you!

alexdorsch:
Updated to 185_000. Thanks!

zzmp (approved):

PR #3662, opened 2022-04-05
fix: JsonRpc url wrapper

zzmp:
- Fixes jsonRpcEndpoint wrapper so that if a JsonRpcProvider is used, it will be featureful (eg able to change networks)

### DIFF
--- a/src/lib/hooks/useActiveWeb3React.tsx
+++ b/src/lib/hooks/useActiveWeb3React.tsx
@@ -23,7 +23,7 @@ type Web3ContextType = {
 
 const EMPTY_CONNECTOR = initializeConnector(() => EMPTY)
 const EMPTY_CONTEXT: Web3ContextType = { connector: EMPTY }
-const urlConnectorAtom = atom<[Connector, Web3ReactHooks, Web3ReactStore]>(EMPTY_CONNECTOR)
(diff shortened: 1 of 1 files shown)


src/lib/hooks/useActiveWeb3React.tsx
```diff
@@ -69,8 +69,14 @@ export function ActiveWeb3Provider({
     return EIP1193
   }, [provider]) as { new (actions: Actions, initializer: typeof provider): Connector }
   const injectedConnector = useConnector(injectedConnectorAtom, Injected, provider)
-  const urlConnector = useConnector(urlConnectorAtom, Url, jsonRpcEndpoint)
-  const [connector, hooks] = injectedConnector[1].useIsActive() ? injectedConnector : urlConnector ?? EMPTY_CONNECTOR
+  const JsonRpc = useMemo(() => {
+    if (JsonRpcProvider.isProvider(jsonRpcEndpoint)) return JsonRpcConnector
+    return Url
```
JFrankfurt, line 74: such an unfortunate name 😅

JFrankfurt (approved):
lgtm

src/lib/hooks/useActiveWeb3React.tsx
zzmp: to be fair it is scoped to web3-react

PR #3665, opened 2022-04-06
fix: simplify validation

zzmp:
Simplify validation by switching to inline hooks (or, for swap, a dedicated hook) instead of non-presentational components.

- No longer throws for non-fatal errors in the Widget (previously the widget validator)
- Also updates cosmos viewer to be able to exercise edge cases:
  - Switch to fake locale to see the warning in the console
  - Make the width under 300 to see the warning in the console


cmcewen (approved):
LGTM!

PR #3673, opened 2022-04-06
fix: do not fetch wrap price

zzmp:
- Prevents the toolbar from showing "Fetching best price" when the "swap" is just a "wrap", as the price in this case is fixed.


JFrankfurt (approved):
Wondering if it might be better to investigate why this line is `true` for wrap transactions in the first place. I wonder if there's a routing process running behind the scenes for wraps? There shouldn't be, but I'd expect this to be `true` in that case. `(state === TradeState.SYNCING || state === TradeState.LOADING)`

PR #3683, opened 2022-04-07
fix: show i18n keys while messages load

zzmp:


src/lib/i18n.tsx
```diff
@@ -104,6 +102,20 @@ export function Provider({ locale, forceRenderAfterLocaleChange = true, onActiva
       })
   }, [locale, onActivate])
 
+  // Initialize the locale immediately if it is DEFAULT_LOCALE, so that keys are shown while loading the translation messages.
+  // This renders the translation _keys_, not the translation _messages_, which is only acceptable while loading the DEFAULT_LOCALE,
+  // as [there are no "default" messages](https://github.com/lingui/js-lingui/issues/388#issuecomment-497779030).
+  // See https://github.com/lingui/js-lingui/issues/1194#issuecomment-1068488619.
+  const isInitialized = useRef(false)
+  if (!isInitialized.current) {
```
justindomingue, line 110: shouldn't this run in a useeffect?

zzmp: A useEffect is too late to avoid the warning generated from lingui (see https://github.com/lingui/js-lingui/issues/1194), and would also prevent SSR from including the default locale.

justindomingue (approved):

PR #3685, opened 2022-04-07
fix: omit document ref

zzmp:
Defers the use of `document` until the `useEffect` hook to avoid breaking nextjs SSR, which has no global `document`.

### DIFF
--- a/src/hooks/useIsWindowVisible.ts
+++ b/src/hooks/useIsWindowVisible.ts
@@ -12,13 +12,14 @@ function isWindowVisible() {
  * Returns whether the window is currently visible to the user.
  */
 export default function useIsWindowVisible(): boolean {
-  const [focused, setFocused] = useState<boolean>(isWindowVisible())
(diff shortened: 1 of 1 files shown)


src/hooks/useIsWindowVisible.ts
```diff
@@ -12,13 +12,14 @@ function isWindowVisible() {
...
 export default function useIsWindowVisible(): boolean {
-  const [focused, setFocused] = useState<boolean>(isWindowVisible())
+  const [focused, setFocused] = useState<boolean>(false)
   const listener = useCallback(() => {
     setFocused(isWindowVisible())
   }, [setFocused])
 
   useEffect(() => {
     if (!isVisibilityStateSupported()) return undefined
+    setFocused((focused) => isWindowVisible())
```
tinaszheng, line 22: u can also do `setFocused(() => isWindowVisible())` since `focused` isnt used

tinaszheng (approved):

PR #3695, opened 2022-04-08
fix: skewed swap info state

zzmp:
Fixes useSwapInfo such that it returns up-to-date currency values, instead of lagging behind, because it is triggered by updates to swapAtom and so is not always consistent with swapAtom.

As an example of what this fixes, reverse the currencies in the widget. You'll notice on the [current build](https://widgets-uniswap.vercel.app/), there is a lag in the UI when updating the swap info. This manifests in two ways, both of which should be fixed by this PR:
- When reversing a wrap, the text will update twice. You can see this if you capture the frames using the Chrome DevTools Performance tab.
- When reversing swap, the amounts will update twice, as the two fields will update on different frames.

### DIFF
--- a/src/lib/hooks/swap/useSwapInfo.tsx
+++ b/src/lib/hooks/swap/useSwapInfo.tsx
@@ -1,15 +1,15 @@
 import { Currency, CurrencyAmount, Percent, TradeType } from '@uniswap/sdk-core'
 import { atom } from 'jotai'
 import { useAtomValue, useUpdateAtom } from 'jotai/utils'
+import useActiveWeb3React from 'lib/hooks/useActiveWeb3React'
 import { useCurrencyBalances } from 'lib/hooks/useCurrencyBalance'
+import useSlippage, { Slippage } from 'lib/hooks/useSlippage'
+import useUSDCPriceImpact, { PriceImpact } from 'lib/hooks/useUSDCPriceImpact'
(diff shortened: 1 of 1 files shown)


src/lib/hooks/swap/useSwapInfo.tsx
```diff
@@ -118,5 +97,37 @@ export function SwapInfoUpdater() {
...
+        (currencyIn && !trade.trade.inputAmount.currency.equals(currencyIn)) ||
+        (currencyOut && !trade.trade.outputAmount.currency.equals(currencyOut))
+      // swapInfo has not yet caught up to swapAtom.
+      if (isTradeStale) return TradeState.LOADING
+    }
+    return trade.state
+  }, [currencyIn, currencyOut, swapInfo])
+
+  const { account } = useActiveWeb3React()
+  const [balanceIn, balanceOut] = useCurrencyBalances(
```
ianlapham, line 116: why do we specifically pull balance out of the other hook?

ianlapham (commented):
Wondering about balances

src/lib/hooks/swap/useSwapInfo.tsx
zzmp: I wanted to ensure that balance was being computed from the correct currency, and not from the stale currency.

ianlapham (approved):
lgtm

Issue #2442, opened 2021-09-25
Package scripts are not compatible with Windows


pengchew:
**Bug Description**
./src/constants/locales.ts
Module not found: Can't resolve '../locales/en-US' in 'C:\Programming\interface\src\constants'

**Steps to Reproduce**

1. yarn install && yarn start

pengchew:
<img width="960" alt="1" src="https://user-images.githubusercontent.com/64325702/134804620-f026ca91-e35a-401d-bc8e-33c9e04b7b4a.png">

justindomingue:
`touch locales/en-US` should do it.

cc @zzmp

zzmp:
Thanks for including the terminal output. It looks like contract compilation failed in postinstall, so i18n compilation never occurred. I'll double check the scripts today to see what causes this.

zzmp:
@pengchew this occurred because our package scripts were not written to support windows, and `./node_modules/@uniswap/?(v3-core|v3-periphery)/artifacts/contracts/**/*.json` is not a valid shell expansion for you. It will need to be split up to separate commands to be more compatible.

pengchew:
> @pengchew this occurred because our package scripts were not written to support windows, and `./node_modules/@uniswap/?(v3-core|v3-periphery)/artifacts/contracts/**/*.json` is not a valid shell expansion for you. It will need to be split up to separate commands to be more compatible.

Thanks for the explanations, got it.

meiskam:
`touch` also doesn't exist on windows .. add `touch-for-windows` package maybe?

Thorhus:
> for windows update scripts section of package.json to this and run with yarn install
> 
> ```
> "scripts": {
>     "contracts:compile:abi": "typechain --target ethers-v5 --out-dir src/abis/types ./src/abis/**/*.json",
>     "contracts:compile:v3": "typechain --target ethers-v5 --out-dir src/types/v3 ./node_modules/@uniswap/v3-core/artifacts/contracts/**/*.json && typechain --target ethers-v5 --out-dir src/types/v3 ./node_modules/@uniswap/v3-periphery/artifacts/contracts/**/*.json",
>     "contracts:compile": "yarn contracts:compile:abi && yarn contracts:compile:v3",
>     "graphql:generate": "graphql-codegen --config codegen.yml",
>     "prei18n:extract": "touch src/locales/en-US.po",
>     "i18n:extract": "lingui extract --locale en-US",
[...]

Issue #3690, opened 2022-04-08
Unable to select Network on mobile device after clicking Connect Wallet


brunocrosier:
**Bug Description**
On the  Swap page, it is impossible to open the Network selector while on mobile if you have previously opened and closed a Currency Input component or the Connect Wallet dialog 

**Steps to Reproduce**

1. Go to https://app.uniswap.org/#/swap?chain=polygon on a device with no hover support (i.e. a mobile)
2. Click on MATIC
3. Either select a currency or just close the dialog
4. Try to open the Network Select. You won't be able to

**Expected Behavior**
Should be able to open and close the Network Select at all times

**Additional Context**
I have investigated this and will create a PR for it. The cause of the issue is that the `NetworkSelect` component is currently only ever toggled open / closed based on hover, which does not work in the same way for mobiles. PR will make it possible to toggle the state based on a click as well as hover.

jbeanz:
can you assign this to me?

brunocrosier:
> can you assign this to me? i've worked on similar issues

thanks! I have a PR which fixes this, which I will push in a couple of hours. Will tag you in it for review :)

brunocrosier:
Hi @jbeanz please feel free to take a look at https://github.com/Uniswap/interface/pull/3698

PR #3693, opened 2022-04-08
chore: Remove Portis

MarkCarbajal:
closes #3680

PR #3698, opened 2022-04-09
NetworkSelect mobile toggle bug

brunocrosier:
Fixes: [[3690]](https://github.com/Uniswap/interface/issues/3690)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -47,7 +47,12 @@
     "simple-import-sort/imports": "error",
     "simple-import-sort/exports": "error",
     "@typescript-eslint/explicit-function-return-type": "off",
-    "prettier/prettier": "error",
+    "prettier/prettier": [
+      "error",
+      {
+        "endOfLine": "auto"
+      }
+    ],
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/ban-ts-comment": "off",
     "@typescript-eslint/ban-ts-ignore": "off",
--- a/package.json
+++ b/package.json
@@ -166,7 +166,6 @@
     "contracts:compile:v3": "typechain --target ethers-v5 --out-dir src/types/v3 \"./node_modules/@uniswap/**/artifacts/contracts/**/*.json\"",
     "contracts:compile": "yarn contracts:compile:abi && yarn contracts:compile:v3",
     "graphql:generate": "graphql-codegen --config codegen.yml",
-    "prei18n:extract": "touch src/locales/en-US.po",
     "i18n:extract": "lingui extract --locale en-US",
     "i18n:compile": "yarn i18n:extract && lingui compile",
     "i18n:pseudo": "lingui extract --locale pseudo && lingui compile",
--- a/src/components/Header/NetworkSelector.tsx
+++ b/src/components/Header/NetworkSelector.tsx
@@ -2,14 +2,13 @@ import { Trans } from '@lingui/macro'
 import { CHAIN_INFO } from 'constants/chainInfo'
 import { CHAIN_IDS_TO_NAMES, SupportedChainId } from 'constants/chains'
 import useActiveWeb3React from 'hooks/useActiveWeb3React'
-import { useOnClickOutside } from 'hooks/useOnClickOutside'
 import useParsedQueryString from 'hooks/useParsedQueryString'
 import usePrevious from 'hooks/usePrevious'
 import { ParsedQs } from 'qs'
 import { useCallback, useEffect, useRef } from 'react'
 import { ArrowDownCircle, ChevronDown } from 'react-feather'
 import { useHistory } from 'react-router-dom'
-import { useModalOpen, useToggleModal } from 'state/application/hooks'
+import { useCloseModal, useModalOpen, useOpenModal, useToggleModal } from 'state/application/hooks'
 import { addPopup, ApplicationModal } from 'state/application/reducer'
 import styled from 'styled-components/macro'
 import { ExternalLink, MEDIA_WIDTHS } from 'theme'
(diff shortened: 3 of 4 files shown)


.eslintrc.json
```diff
@@ -47,7 +47,12 @@
     "simple-import-sort/imports": "error",
     "simple-import-sort/exports": "error",
     "@typescript-eslint/explicit-function-return-type": "off",
-    "prettier/prettier": "error",
+    "prettier/prettier": [
+      "error",
+      {
+        "endOfLine": "auto"
```
brunocrosier, line 53: Fixes an issue for developers working on Windows, where line breaks would falsely be shown as a linting error.

More info: https://stackoverflow.com/a/53769213/1843091

package.json
```diff
@@ -166,7 +166,6 @@
     "contracts:compile:v3": "typechain --target ethers-v5 --out-dir src/types/v3 \"./node_modules/@uniswap/**/artifacts/contracts/**/*.json\"",
     "contracts:compile": "yarn contracts:compile:abi && yarn contracts:compile:v3",
     "graphql:generate": "graphql-codegen --config codegen.yml",
-    "prei18n:extract": "touch src/locales/en-US.po",
```
brunocrosier, line 169: Fixes an issue for developers working on Windows. The `touch` command does not exist in Windows, so running `yarn` will fail. 

From what I can see, it is not necessary to actually run this command, as `src/locales/en-US.po` is already checked into the repo and therefore doesn't need to be created via `touch`. If there is a specific reason why this command does need to be run as part of the install process, I would recommend something like this:

```json
"prei18n:extract": "node prei18n-extract.js",
```

```js
// prei18n-extract.js
import { exec } from 'child_process'
const isWindows = process.platform === 'win32' || /^(msys|cygwin)$/.test(process.env.OSTYPE)

if (isWindows) {
  exec(`type nul > src/locales/en-US.po`)
} else {
  exec(`touch src/locales/en-US.po`)
}
```

Closes: https://github.com/Uniswap/interface/issues/2442

src/components/Header/NetworkSelector.tsx
```diff
@@ -242,11 +241,11 @@ export default function NetworkSelector() {
   const parsedQs = useParsedQueryString()
   const { urlChain, urlChainId } = getParsedChainId(parsedQs)
   const prevChainId = usePrevious(chainId)
-  const node = useRef<HTMLDivElement>()
+  const node = useRef<HTMLDivElement>(null)
```
brunocrosier, line 244: Means we no longer need to assert as `any` 

```jsx
<SelectorWrapper ref={node as any} ... >
```

becomes

```jsx
<SelectorWrapper ref={node} ... >
```

src/components/Header/NetworkSelector.tsx
```diff
@@ -308,7 +307,7 @@ export default function NetworkSelector() {
   }
 
   return (
-    <SelectorWrapper ref={node as any} onMouseEnter={toggle} onMouseLeave={toggle}>
+    <SelectorWrapper ref={node} onMouseEnter={openModal} onMouseLeave={closeModal} onClick={toggle}>
```
brunocrosier, line 310: The reason why we can't have just do:

```jsx
 <SelectorWrapper ref={node} onMouseEnter={toggle} onMouseLeave={toggle} onClick={toggle}>
```

is that we could end up in this state:

1. hover to open
2. click to close (at this point, mouse is still hovering over element)

this would mean that when the mouse leaves, the `toggle` function would be called again, thereby reopening the modal (which is not the desired behaviour)

brunocrosier:
Hey @willhennessy 

Hmm that's strange. I was able to run `yarn build` locally with no issues, and also was able to deploy to Vercel (see here: https://interface-sooty.vercel.app/#/swap?chain=mainnet)

The strange thing is, I didn't make any changes to `src/components/WalletModal/index.tsx` in this PR

I think there's a chance you may be looking at a different PR by mistake? I can see that there is another open PR which does modify `src/components/WalletModal/index.tsx` - and there is indeed a missing space at line 16, column 29

I have commented that file with a suggested fix here: https://github.com/Uniswap/interface/pull/3693/files#r847601265

image

Any questions, let me know!

willhennessy:
Sorry! You're right, I was reviewing PRs and mixed up the tabs. Great detective work. 

We're getting an engineer to review the PR asap. Thanks bruno.

package.json
JFrankfurt: @zzmp thoughts?

.eslintrc.json
zzmp: Could you move your windows-specific fixes to a separate PR? I think it would be great to get them in, but they shouldn't be mixed with other bug fixes (for ease of review and isolation in case of introduced bugs).

JFrankfurt: I think you might want to look at updating your git config instead? https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

I work on windows fairly frequently and have not had any issues with prettier. 

my machine:
```
$ git config -l
...
core.autocrlf=<boolean>
...
```

JFrankfurt (changes requested):
I think you should 
- drop the package.json changes and add them in a new pr w/ a windows compatible script.
- look into a local compatibility solution (prob git config) for the prettier issue

brunocrosier:
@zzmp @JFrankfurt Thanks for your comments. I have removed the Windows changes, and will create a separate PR for that!

JFrankfurt (approved):

package.json
brunocrosier: Added in https://github.com/Uniswap/interface/pull/3728

.eslintrc.json
brunocrosier: Added in https://github.com/Uniswap/interface/pull/3730

vm (approved):

Issue #3677, opened 2022-04-06
Loading spinner disappears at narrow widget


willhennessy:
Desktop. At normal window width, an animated spinner svg appears while the app is fetching a quote form the Auto Router

At more narrow window widths, the animated spinner gets hidden. However there is still plenty of space in the swap box to show the spinner. In fact the swap box doesn't change at all at this width, so there's no reason to hide the spinner.

Here's the relevant line of code:  https://github.com/Uniswap/interface/blob/main/src/components/swap/SwapDetailsDropdown.tsx#L144

<img width="532" alt="Screen Shot 2022-04-06 at 6 47 33 PM" src="https://user-images.githubusercontent.com/5455577/162085181-0df7047c-93fa-4a9e-bf4b-56028225d8ff.png">
<img width="465" alt="Screen Shot 2022-04-06 at 6 47 24 PM" src="https://user-images.githubusercontent.com/5455577/162085179-a33991b8-a6d4-44d2-903c-81264fea5469.png">

PR #3699, opened 2022-04-09
fix: show loading spinner on mobile

brunocrosier:
Fixes: https://github.com/Uniswap/interface/issues/3677


JFrankfurt:
I think we don't show this spinner because we want to save space and not render the tooltip info circle on mobile. 

image


image

If we merge this as-is we get some kinda not-so-great layout thrashing where the conversion rate jumps left and right due to the spinner rendering in and out.

JFrankfurt:
curious what @ppphhhiiilll or @maggievalentine think about the design implications here. Do we add both to mobile?

PR #3707, opened 2022-04-11
fix: update hook deps to improve ref equality checks

zzmp:
- Prevents re-rendering of swap info when no values have actually changed

### DIFF
--- a/src/lib/components/TokenImg.tsx
+++ b/src/lib/components/TokenImg.tsx
@@ -26,16 +26,17 @@ function TokenImg({ token, ...rest }: TokenImgProps) {
     setAttempt((attempt) => ++attempt)
   }, [])
 
-  return useMemo(() => {
+  const src = useMemo(() => {
     // Trigger a re-render when an error occurs.
     void attempt
 
-    const src = srcs.find((src) => !badSrcs.has(src))
-    if (!src) return <MissingToken color="secondary" {...rest} />
+    return srcs.find((src) => !badSrcs.has(src))
+  }, [attempt, srcs])
 
-    const alt = tokenInfo.name || tokenInfo.symbol
-    return <img src={src} alt={alt} key={alt} onError={onError} {...rest} />
-  }, [attempt, onError, rest, srcs, tokenInfo.name, tokenInfo.symbol])
+  if (!src) return <MissingToken color="secondary" {...rest} />
+
+  const alt = tokenInfo.name || tokenInfo.symbol
+  return <img src={src} alt={alt} key={alt} onError={onError} {...rest} />
 }
 
 export default styled(TokenImg)<{ size?: number }>`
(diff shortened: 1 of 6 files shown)


src/lib/components/TokenImg.tsx
```diff
@@ -26,16 +26,17 @@ function TokenImg({ token, ...rest }: TokenImgProps) {
...
     void attempt
 
-    const src = srcs.find((src) => !badSrcs.has(src))
-    if (!src) return <MissingToken color="secondary" {...rest} />
+    return srcs.find((src) => !badSrcs.has(src))
+  }, [attempt, srcs])
 
-    const alt = tokenInfo.name || tokenInfo.symbol
-    return <img src={src} alt={alt} key={alt} onError={onError} {...rest} />
-  }, [attempt, onError, rest, srcs, tokenInfo.name, tokenInfo.symbol])
```
zzmp, line 38: rest is instantiated each time, so this `useMemo` is ineffective.

src/lib/hooks/swap/useSwapInfo.tsx
```diff
@@ -121,13 +123,16 @@ export default function useSwapInfo(): SwapInfo {
   // swapInfo will lag behind swapAtom by a frame, because its update is triggered by swapAtom
```
ianlapham, line 123: do we need this comment still if context is merged? Also does that change remove need to do this

src/lib/hooks/useSlippage.ts
```diff
@@ -27,6 +29,9 @@ export default function useSlippage(trade: InterfaceTrade<Currency, Currency, Tr
     const auto = shouldUseAutoSlippage || !maxSlippage
     const allowed = shouldUseAutoSlippage ? autoSlippage : maxSlippage ?? autoSlippage
     const warning = auto ? undefined : getSlippageWarning(allowed)
+    if (auto && allowed === DEFAULT_AUTO_SLIPPAGE) {
+      return DEFAULT_SLIPPAGE
```
ianlapham, line 33: Does this constant remove a re-render?

ianlapham (commented):
Small questions

src/lib/hooks/swap/useSwapInfo.tsx
zzmp: We will not need this comment after #3710 is merged, but #3710 removes it (and all of the logic in this hook).

src/lib/hooks/useSlippage.ts
zzmp: yes - it removes a re-render on the first frame with a returned result.

src/lib/hooks/swap/useSwapInfo.tsx
ianlapham: great

ianlapham (approved):
Looks good

PR #3708, opened 2022-04-11
fix: use context for block number

zzmp:
Use Context for block number to avoid waterfall re-rendering pattern from using jotai atoms.
Reduces re-renders from ~300 when fetching a swap to ~30, which - along with being more performant - will help facilitate more deterministic testing.


src/lib/hooks/useBlockNumber.tsx
```diff
@@ -0,0 +1,71 @@
...
+    () => ({
+      blockNumber: chainId === activeChainId ? block : undefined,
+      fastForward: (block: number) => setChainBlock({ chainId: activeChainId, block }),
+    }),
+    [activeChainId, block, chainId]
+  )
+  return <ChainBlockContext.Provider value={value}>{children}</ChainBlockContext.Provider>
+}
+
+/** Requires that BlockUpdater be installed in the DOM tree. */
```
justindomingue, line 64: testing that the context is not null is probably a better api than a comment saying it has to be mounted, but nbd

justindomingue (approved):

src/lib/hooks/useBlockNumber.tsx
zzmp: Added a `const MISSING_PROVIDER = Symbol()` to test against.

PR #3710, opened 2022-04-11
chore: simplify swap info

zzmp:
Simplifies swap info by using Context instead of a jotai atom. This avoids the issue of requiring multiple renders from chaining jotai atoms, as a Context's update will trigger all of its children in the same render cycle.

Along with simplifying the code, this makes swap info less error-prone by using Context over atom, and avoiding the possibility of stale state.


ianlapham (approved):
Great fix

PR #3711, opened 2022-04-11
fix: activation frames

zzmp:
Updates @web3-react so that `chainId` stays up-to-date with `active`. This prevents misaligned UI updates when a provider first connects to web3 (ie this makes the wallet, toolbar, and defaults update in sync, instead of over 2 frames).


NoahZinsmeister (approved):
lgtm!

PR #3713, opened 2022-04-12
feat: display connecting state

zzmp:
Displays "Connecting" in the toolbar.

https://user-images.githubusercontent.com/5403956/163024388-f2c9053d-f0fe-4cad-92e1-8d2272152fbe.mov

_NB: The copy was been changed to "Connecting..." after this screencap was made._

### DIFF
--- a/src/lib/components/Swap/Toolbar/index.tsx
+++ b/src/lib/components/Swap/Toolbar/index.tsx
@@ -18,7 +18,7 @@ const ToolbarRow = styled(Row)`
 `
 
 export default memo(function Toolbar({ disabled }: { disabled?: boolean }) {
-  const { chainId } = useActiveWeb3React()
+  const { active, activating, chainId } = useActiveWeb3React()
   const {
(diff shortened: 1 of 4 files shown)


src/lib/components/Swap/Toolbar/index.tsx
```diff
@@ -18,7 +18,7 @@ const ToolbarRow = styled(Row)`
 `
 
 export default memo(function Toolbar({ disabled }: { disabled?: boolean }) {
```
JFrankfurt, line 20: I think you can drop disabled entirely w/ this logic. I don't recall off the top of my head what this disabled state is usually caused by, so you might want to double check and make sure we don't *ever* want a `disabled`-contingent UI state from the toolbar.

zzmp: It's caused by an `!useActiveWeb3React().active`, but that's already been baked into the Toolbar. Removed!

JFrankfurt (approved):

PR #3714, opened 2022-04-12
feat(compliance): risk screening

JFrankfurt:
block addresses from using the hosted interface if their risk value surpasses a threshold

image


src/components/AccountDetails/Copy.tsx
```diff
@@ -43,7 +51,7 @@ export default function CopyHelper(props: { toCopy: string; children?: React.Rea
           <Copy size={'16'} />
         </TransactionStatusText>
       )}
-      {isCopied ? '' : props.children}
+      {isCopied ? '' : children}
```
zzmp, line 54: Add a space between the icon and the content

PR #3715, opened 2022-04-12
fix: Reload the app when there is a javascript error and a new version of the app

leggechr:
The initial implementation by @moodysalem was done in: https://github.com/Uniswap/interface/pull/3435. The problem there was if there was no update from the app it would get stuck refreshing infinitely.

I copied that PR but added a check to ensure that there is actually a new version being returned from the update and only in that case will the app update. I also had to add a few more lines that appear to be required to make the refresh happen properly which I copied from: https://github.com/facebook/create-react-app/issues/5316#issuecomment-591075209

### DIFF
--- a/src/components/ErrorBoundary/index.tsx
+++ b/src/components/ErrorBoundary/index.tsx
@@ -49,6 +49,13 @@ type ErrorBoundaryState = {
 
 const IS_UNISWAP = window.location.hostname === 'app.uniswap.org'
 
+async function updateServiceWorker(): Promise<ServiceWorkerRegistration> {
+  const ready = await navigator.serviceWorker.ready
+  // the return type of update is incorrectly indicated as Promise<void>
+  // https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/update
+  return ready.update() as unknown as Promise<ServiceWorkerRegistration>
+}
+
 export default class ErrorBoundary extends React.Component<unknown, ErrorBoundaryState> {
   constructor(props: unknown) {
     super(props)
@@ -56,6 +63,26 @@ export default class ErrorBoundary extends React.Component<unknown, ErrorBoundar
   }
 
   static getDerivedStateFromError(error: Error): ErrorBoundaryState {
+    updateServiceWorker()
+      .then(async (registration) => {
+        // We want to refresh only if we detect a new service worker is
+        // waiting to be activated.
+        // Details about it: https://developers.google.com/web/fundamentals/primers/service-workers/lifecycle
+        if (registration?.waiting) {
+          await registration.unregister()
+
+          // Makes Workbox call skipWaiting()
+          registration.waiting.postMessage({ type: 'SKIP_WAITING' })
+
+          // Once the service worker is unregistered, we can reload the page to let
+          // the browser download a fresh copy of our app (invalidating the cache)
+          console.log('reloading waiting', registration)
+          window.location.reload()
+        }
+      })
+      .catch((error) => {
+        console.error('Failed to update service worker', error)
+      })
     return { error }
   }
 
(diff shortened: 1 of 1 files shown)


NoahZinsmeister (approved):
this looks good! i know you said you were able to test this locally - how hard do you think it would be to add a unit test? or do you think it makes more sense to try to land tests once the fully featured service worker flow lands?

leggechr:
@NoahZinsmeister the local testing I did involved building and serving the binary and then using the devtools to force an update so it doesn't translate easily to unit tests. I'd prefer to wait until we decide what to do with the rest of the service worker and then I can implement unit tests for the whole thing.

src/components/ErrorBoundary/index.tsx
```diff
@@ -49,13 +49,39 @@ type ErrorBoundaryState = {
 
 const IS_UNISWAP = window.location.hostname === 'app.uniswap.org'
 
+async function updateServiceWorker(): Promise<ServiceWorkerRegistration> {
+  const ready = await navigator.serviceWorker.ready
+  // the return type of update is incorrectly indicated as Promise<void>
```
zzmp, line 54: nit: "incorrectly typed as"

src/components/ErrorBoundary/index.tsx
```diff
@@ -49,13 +49,39 @@ type ErrorBoundaryState = {
...
+  // https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/update
```
zzmp, line 55: nit: "See <url>"

src/components/ErrorBoundary/index.tsx
```diff
@@ -49,13 +49,39 @@ type ErrorBoundaryState = {
...
 export default class ErrorBoundary extends React.Component<unknown, ErrorBoundaryState> {
   constructor(props: unknown) {
     super(props)
     this.state = { error: null }
   }
 
   static getDerivedStateFromError(error: Error): ErrorBoundaryState {
+    updateServiceWorker()
+      .then(async (registration) => {
+        // We want to refresh only if we detect a new service worker is
```
zzmp, line 68: Our [printWidth is set to 120](https://github.com/Uniswap/interface/blob/main/.prettierrc#L4), so this comment can be a single line.

src/components/ErrorBoundary/index.tsx
```diff
@@ -49,13 +49,39 @@ type ErrorBoundaryState = {
...
+        // waiting to be activated.
+        // Details about it: https://developers.google.com/web/fundamentals/primers/service-workers/lifecycle
```
zzmp, line 70: nit: "See <url>"

src/components/ErrorBoundary/index.tsx
```diff
@@ -49,13 +49,39 @@ type ErrorBoundaryState = {
...
+        if (registration?.waiting) {
+          await registration.unregister()
+
+          // Makes Workbox call skipWaiting()
```
zzmp, line 74: Can you link to documentation in this comment? This isn't a typed API, so there's no real inline documentation.

src/components/ErrorBoundary/index.tsx
zzmp, line 68: Can you update this comment to include the rationale behind using skipWaiting (below), or put that in a comment below?

zzmp (commented):
The code looks great, can you just clean up the comments? Normally I wouldn't be this nitpicky but SW is a little esoteric, so I think it's important to document inline and link out to supporting docs (most of which you've already done 😄).

src/components/ErrorBoundary/index.tsx
leggechr: I included a link to the workbox documentation that talks about skipWaiting and why its necessary. I think this should address this comment and the one below.

zzmp (approved):

PR #3719, opened 2022-04-12
fix: initial transitions

zzmp:
These changes remove any jarring UX on initial provider connection. It adds a transition to any element that is already rendered initially (eg input and token buttons). It intentionally does _not_ add transitions to those elements which were not already rendered (eg "Connect your wallet").

- Cleans up the action button (including fixing a bug that allowed stale swaps)
- Adds fade in to reverse and token select buttons when transitioning from disabled state
- Adds fade in to input text

https://user-images.githubusercontent.com/5403956/163095872-b30cbf7e-41de-4614-8ac8-8eef1e4709b5.mov

### DIFF
--- a/src/lib/components/Button.tsx
+++ b/src/lib/components/Button.tsx
@@ -1,5 +1,5 @@
 import { Icon } from 'lib/icons'
-import styled, { Color, css, keyframes } from 'lib/theme'
+import styled, { Color, css } from 'lib/theme'
 import { ComponentProps, forwardRef } from 'react'
 
 export const BaseButton = styled.button`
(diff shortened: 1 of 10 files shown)


src/lib/components/TokenSelect/TokenButton.tsx
```diff
@@ -42,10 +39,37 @@ interface TokenButtonProps {
...
+  }, [row, shouldTransition])
+
   return (
-    <StyledTokenButton onClick={onClick} empty={!value} color={buttonBackgroundColor} disabled={disabled}>
+    <StyledTokenButton
+      onClick={onClick}
+      color={buttonBackgroundColor}
+      disabled={disabled}
+      bleedIn={shouldTransition}
+      onAnimationEnd={() => setShouldTransition(false)}
```
zzmp, line 67: Alternatively, `Button` could be wrapped so that it exposes its own `onBleedIn` handler (at which point they could be renamed to something like `animateEnable`/`onAnimateEnable`), but that seemed like overkill for this one-off.

src/lib/components/Widget.tsx
```diff
@@ -31,7 +31,7 @@ const WidgetWrapper = styled.div<{ width?: number | string }>`
   font-size: 16px;
   font-smooth: always;
   font-variant: none;
-  height: 356px;
+  height: 360px;
```
zzmp, line 34: The extra 4px comes from always showing the TokenButton borders due to the changes in Button. Always showing borders makes handling styles easier, as prior to this borders were special-cased, and required explicit css (eg `calc(1em - 1px)`) to keep box sizes consistent with/without borders.

src/lib/components/Button.tsx
```diff
@@ -14,28 +14,46 @@ export const BaseButton = styled.button`
   line-height: inherit;
   margin: 0;
   padding: 0;
+  transition: filter 0.125s linear;
 
   :disabled {
     cursor: initial;
     filter: saturate(0) opacity(0.4);
+    transition: none;
```
JFrankfurt, line 22: will this stop transitions out of a disabled state?, and is [filter](https://developer.mozilla.org/en-US/docs/Web/CSS/filter) the only property you want to animate?

JFrankfurt:
need to test this later, but I'm a little wary of the `:enabled` and `:disabled` wrapped transitions

src/lib/components/Button.tsx
JFrankfurt: Should the border and background colors below also use this transition?

zzmp: That's much simpler, yes. I've updated the Button to use transition instead of animation.

JFrankfurt (approved):
lgtm 😃

PR #3724, opened 2022-04-13
feat: add 1bp fee tier to polygon

leggechr:
The fee tier is showing up on the UI and I was able to add liquidity for that fee tier. 

@justindomingue  helped me to test this by making sure a swap would go through the 1bp tier: 
image (1)

The resulting transaction was https://mumbai.polygonscan.com/tx/0x744bb33f9f6086cc6b19c29ba604d3ad6bba6c3d9ab1d137bf27b19fc082b92c which we _think_ confirms that my swap went through that tier properly.

### DIFF
--- a/src/components/FeeSelector/shared.tsx
+++ b/src/components/FeeSelector/shared.tsx
@@ -10,7 +10,7 @@ export const FEE_AMOUNT_DETAIL: Record<
   [FeeAmount.LOWEST]: {
     label: '0.01',
     description: <Trans>Best for very stable pairs.</Trans>,
-    supportedChains: [SupportedChainId.MAINNET],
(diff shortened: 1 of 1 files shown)


justindomingue (approved):

leggechr:
My only concern is that there is nothing showing up in the unclaimed fees UI. Is this because the number is so small or because the swap didn't go through that fee tier?
image

snreynolds:
> My only concern is that there is nothing showing up in the unclaimed fees UI. Is this because the number is so small or because the swap didn't go through that fee tier? image

Looks good! Routing through pool `0x85bc6cbfb02a110d5839b344545b02fb6cb70cd5` which has fee `100`

PR #3727, opened 2022-04-13
fix: missing token img

zzmp:
Stores the local version of a src as a badSrc on error, instead of using `e.target.src`, which may be normalized differently.

### DIFF
--- a/src/lib/components/TokenImg.tsx
+++ b/src/lib/components/TokenImg.tsx
@@ -21,18 +21,21 @@ function TokenImg({ token, ...rest }: TokenImgProps) {
   const srcs = useCurrencyLogoURIs(tokenInfo)
 
   const [attempt, setAttempt] = useState(0)
-  const onError = useCallback((e) => {
-    if (e.target.src) badSrcs.add(e.target.src)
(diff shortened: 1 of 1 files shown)


src/lib/components/TokenImg.tsx
```diff
@@ -21,18 +21,21 @@ function TokenImg({ token, ...rest }: TokenImgProps) {
...
     // Trigger a re-render when an error occurs.
     void attempt
 
     return srcs.find((src) => !badSrcs.has(src))
   }, [attempt, srcs])
-
+  const onError = useCallback(
+    (e) => {
+      if (src) badSrcs.add(src)
+      setAttempt((attempt) => ++attempt)
```
JFrankfurt, line 33: any particular reason you do this as a mutation of the local scope before the setter returns the new value?

JFrankfurt (approved):

src/lib/components/TokenImg.tsx
zzmp: `attempt` needs to be updated to force `src` to update, because `badSrcs` is external to React's state. This is just incrementing `attempt` to ensure it is updated.

PR #3728, opened 2022-04-13
fix: add crossplatform `prei18n-extract` script

brunocrosier:
Previous `prei18n-extract` script used `touch` which does not exist on Windows.

Implement `prei18n-extract.js`, which runs the correct command based on the platform

Logic for `isWindows` is borrowed from the popular `cross-env` package: https://github.com/kentcdodds/cross-env/blob/master/src/is-windows.js

Fixes https://github.com/Uniswap/interface/issues/2442

Screenshot of the command working on my Windows machine:

image


zzmp (changes requested):
Can you simplify this (and avoid adding the helper file) by using a x-platform to like [shx](https://www.npmjs.com/package/shx)?

brunocrosier:
> Can you simplify this (and avoid adding the helper file) by using a x-platform to like [shx](https://www.npmjs.com/package/shx)?

Good idea, updated accordingly!

zzmp (approved):

zzmp (changes requested):
Sorry, but for security, we're not adding cross-platform-only packages to our dependencies (like `shx`).

zzmp:
sorry to have gone back and forth on this

PR #3730, opened 2022-04-13
style: `endOfLine` error on Windows 

brunocrosier:
By default, on Windows, line-endings are terminated with a combination of a carriage return (ASCII 0x0d or \r) and a newline(\n), also referred to as CR/LF.

Without making this change to the `.eslintrc.json` file

```json
"prettier/prettier": [
      "error",
      {
        "endOfLine": "auto"
      }
    ],
```

the default Windows development experience would be to see these errors:

image

Although it is possible to work around this in VS Code by editing the `Change end of Line Sequence`  setting, I think it would be best if it worked "out of the box" for all prospective developers

image

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -47,7 +47,12 @@
     "simple-import-sort/imports": "error",
     "simple-import-sort/exports": "error",
     "@typescript-eslint/explicit-function-return-type": "off",
-    "prettier/prettier": "error",
+    "prettier/prettier": [
+      "error",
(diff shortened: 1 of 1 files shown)


NoahZinsmeister:
right now windows doesn't have first-class support for local development (e.g. we only run CI against ubuntu); since we'd like line-endings to be consistent across the codebase, unfortunately i'm going to have to recommend that windows developers make local config changes to solve this issue

PR #3735, opened 2022-04-14
fix: always-fresh service worker cache

zzmp:


src/service-worker.ts
```diff
@@ -48,4 +58,33 @@ registerRoute(({ request, url }: { request: Request; url: URL }) => {
...
+    // If the cache is still fresh, cancel any pending response. The preloadResponse is cancelled
+    // automatically by returning before it is settled.
+    controller.abort()
+  } else {
+    cachedAppShell = response
+  }
+  // TODO(leggechr): Send beacons with metrics to track usage/load times. This should probably be done on the client, not the serviceworker.
+  // Cached responses can have a header added to indicate that they were served from cache.
+
+  const clone = cachedAppShell.clone()
```
leggechr, line 88: How come we have to clone here?

src/service-worker.ts
```diff
@@ -48,4 +58,33 @@ registerRoute(({ request, url }: { request: Request; url: URL }) => {
...
+  // NB: If the cached response is used, the preloadResponse may be cancelled before it is settled.
+  // This will log an error to the console, but it can be ignored - it *should* be cancelled.
+  const { preloadResponse } = event as unknown as { preloadResponse: Promise<Response | undefined> }
+  const controller = new AbortController()
+  const response = (await preloadResponse) || (await fetch(APP_SHELL_HTML, { signal: controller.signal }))
+
+  // Only serve from the cache if the cache is still fresh (checked using etag comparison). The etag
+  // header can be queried before the response has streamed, so this still reduces network time.
+  // (In manual testing, etag lookup took only 2-20% of the total response download time.)
+  const etag = response.headers.get('etag')
```
leggechr, line 76: So this doesn't wait for the whole response to be fetched? It just waits for the headers to come back? Or how does this come back faster than the actual page?

leggechr (commented):
A couple questions for my understanding!

src/service-worker.ts
zzmp: Responses are one-off - once they are read they are no longer valid. Clone allows copying them without invalidating them, so it's necessary for a cache, where the same responses may be read many times. Without it, you'd get a TypeError. 

See https://developer.mozilla.org/en-US/docs/Web/API/Request/clone (particularly the second paragraph).

src/service-worker.ts
zzmp: This JavaScript API is poorly documented in terms of its performance characteristics. 

HTTP responses are text, starting with the headers, then the body, then trailing headers. Response bodies are streamed (by JS APIs), and are only available as promises off of the settled Response promise.

I'm making an assumption, based on this, that the headers are available before the body (because it _is_ available first, to the underlying program, due to the way the response is received over tcp/http). I verified this assumption by timing how long it took to read the etag header vs how long it took to read the full response, and saw that the etag header was available in 20% of the time.

leggechr:
Closing in favor of a user refresh popup.

PR #3739, opened 2022-04-14
style: Enforce `endOfLine` to be `LF` on Windows

brunocrosier:
Alternate (hopefully better!) solution to https://github.com/Uniswap/interface/pull/3730 

Currently, Windows machines will see the following:

image

By changing `endOfLine` to `LF` in `.prettierrc` everything should work correctly for Windows machines without forcing the user to make local git config changes.


zzmp:
The documentation for prettier [1] includes a direction to update .gitattributes. Is this necessary? I'd prefer not to (because we don't currently have a .gitattributes file) but I want to make sure this works as expected.

[1]: https://prettier.io/docs/en/options.html#end-of-line

zzmp:
Adding @NoahZinsmeister because of his prior review of #3730.

brunocrosier:
> The documentation for prettier [1](https://prettier.io/docs/en/options.html#end-of-line) includes a direction to update .gitattributes. Is this necessary? I'd prefer not to (because we don't currently have a .gitattributes file) but I want to make sure this works as expected.

Interesting.. What I can say for sure is that this change (without `.gitattributes`) fixes the issue on my Windows machine + VS Code.

zzmp (approved):

brunocrosier:
Apologies, I just realized that in order for this to work, both `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` need to be updated.

zzmp (commented):
I'll need to re-review now that deps are being updated to new major versions.

package.json
```diff
@@ -80,8 +80,8 @@
     "@types/testing-library__cypress": "^5.0.5",
     "@types/ua-parser-js": "^0.7.35",
     "@types/wcag-contrast": "^3.0.0",
-    "@typescript-eslint/eslint-plugin": "^4.1.0",
-    "@typescript-eslint/parser": "^4.1.0",
+    "@typescript-eslint/eslint-plugin": "^5.19.0",
```
NoahZinsmeister, line 83: just for completeness, why is this version bump required? as you can see from the yarn.lock we're now installing 2 major versions of these 2`@typescript-eslint` packages, which is a bit unfortunate (though i know it only affects local dev builds not prod code)

NoahZinsmeister (commented):
i'm good with this approach, left a comment!

brunocrosier:
Hmm.. unfortunately this option seems to break existing ESLint config. Apologies, I'm going to close this for now.

Would be nice to find a solution which allows devs using Windows machines to get set up, but I think it will require updating a few packages and configs. Happy to give this another shot if there's support for it - if not, might be worth adding something in the CONTRIBUTING.md about it?

PR #3758, opened 2022-04-20
refactor: move state/transactions to createSlice

leggechr:
As an exercise in learning more about redux toolkit I migrated state/transactions to use `createSlice`. There is no change in behavior. `createSlice` removes the need for creating actions in a separate file.


src/state/transactions/reducer.ts
```diff
@@ -1,17 +1,179 @@
...
-  clearAllTransactions,
-  finalizeTransaction,
-  SerializableTransactionReceipt,
-  TransactionInfo,
-} from './actions'
+import { VoteOption } from '../governance/types'
 
 const now = () => new Date().getTime()
 
+interface SerializableTransactionReceipt {
```
justindomingue, line 9: i think we'd typically put these at `transactions/types.ts` for readability

src/state/transactions/reducer.ts
```diff
@@ -66,13 +218,32 @@ export default createReducer(initialState, (builder) =>
...
+        })
+      })
     })
-)
+  },
+})
+
+export const { addTransaction, clearAllTransactions, checkedTransaction, finalizeTransaction } =
+  transactionSlice.actions
+export default transactionSlice.reducer
```
justindomingue, line 249: i'm surprised we're doing a default export here but seems like we already were?

justindomingue (approved):

src/state/transactions/reducer.ts
leggechr: It was what was there before so I kept it to prevent changes in files that import this. Should I remove the default?

src/state/transactions/reducer.ts
leggechr: Done.

justindomingue (approved):

src/state/transactions/reducer.ts
justindomingue: nbd

zzmp (approved):

PR #3759, opened 2022-04-20
feat: upgrade to web3-react v8

vm:
feature additions:
- adds button to disconnect from MetaMask/Fortmatic (Coinbase Wallet coming soon)
- auto-connects to previously connected wallet without a timeout
- extends adding token to Coinbase Wallet
- resolves Coinbase Wallet and MetaMask conflicts when both installed
- uses wallet’s RPC instead of our Infura by default
- allows opening wallet modal when on an invalid network
- handles switching to an invalid network from outside of the UI
- adds tests for connecting/disconnect an injected wallet

web3-react changes:
- errors stored locally instead of in web3-react state
- errors are thrown inline via `connector.activate` Promise
- pass in onError to connectors to handle out-of-band errors
- implements Fortmatic connector locally
- moves ethers logic for switching chains and adding token to web3-react codebase
- uses config object instead of ordered arguments for `initializeConnector`
- removes usage of `eagerlyConnect` in `initializeConnector`


src/components/AccountDetails/index.tsx
```diff
@@ -265,7 +264,7 @@ export default function AccountDetails({
               <AccountGroupingRow>
                 {formatConnectorName()}
                 <div>
-                  {connector !== injected && connector !== walletlink && (
+                  {connector !== injected && connector !== coinbaseWallet && (
```
vm, line 267: use instanceOf

src/components/WalletModal/index.tsx
```diff
@@ -194,32 +199,27 @@ export default function WalletModal({
     setPendingWallet(connector) // set wallet for pending view
     setWalletView(WALLET_VIEWS.PENDING)
 
-    // if the connector is walletconnect and the user has already tried to connect, manually reset the connector
-    if (connector instanceof WalletConnectConnector) {
-      connector.walletConnectProvider = undefined
+    //     // if the connector is walletconnect and the user has already tried to connect, manually reset the connector
+    //     if (connector instanceof WalletConnect) {
+    //       connector.walletConnectProvider = undefined
```
vm, line 204: i couldn't figure out what this was doing at all

src/components/WalletModal/index.tsx
```diff
@@ -194,32 +199,27 @@ export default function WalletModal({
...
+      try {
+        connector.activate()
+        logMonitoringEvent({ account })
+      } catch (error) {
+        setPendingError(true)
+      }
     }
-
-    connector &&
-      activate(connector, undefined, true)
```
vm, line 203: idk what these arguments are

src/components/WalletModal/index.tsx
```diff
@@ -194,32 +199,27 @@ export default function WalletModal({
...
-      connector.walletConnectProvider = undefined
+    //     // if the connector is walletconnect and the user has already tried to connect, manually reset the connector
+    //     if (connector instanceof WalletConnect) {
+    //       connector.walletConnectProvider = undefined
+    //     }
+
+    if (connector) {
+      try {
+        connector.activate()
+        logMonitoringEvent({ account })
```
vm, line 210: unclear if `connector.getAccount` is same as `account` in v8

src/components/WalletModal/index.tsx
```diff
@@ -194,32 +199,27 @@ export default function WalletModal({
...
-        })
   }
 
-  // close wallet modal if fortmatic modal is active
-  useEffect(() => {
-    fortmatic.on(OVERLAY_READY, () => {
-      toggleWalletModal()
-    })
-  }, [toggleWalletModal])
+  // // close wallet modal if fortmatic modal is active
```
vm, line 217: commented for now

src/components/WalletModal/index.tsx
```diff
@@ -310,10 +310,10 @@ export default function WalletModal({
             <CloseColor />
           </CloseIcon>
           <HeaderRow>
-            {error instanceof UnsupportedChainIdError ? <Trans>Wrong Network</Trans> : <Trans>Error connecting</Trans>}
+            {error instanceof ChainIdNotAllowedError ? <Trans>Wrong Network</Trans> : <Trans>Error connecting</Trans>}
```
vm, line 313: make sure these are the same error

src/components/earn/StakingModal.tsx
```diff
@@ -144,7 +144,7 @@ export default function StakingModal({ isOpen, onDismiss, stakingInfo, userLiqui
   }, [maxAmountInput, onUserInput])
 
   async function onAttemptToApprove() {
-    if (!pairContract || !library || !deadline) throw new Error('missing dependencies')
+    if (!pairContract || !connector || !deadline) throw new Error('missing dependencies')
```
vm, line 147: should this check provider?

src/connectors/Fortmatic.ts
```diff
@@ -1,46 +1,46 @@
...
-
-    const [account] = await Promise.all([
-      provider.enable().then((accounts: string[]) => accounts[0]),
-      pollForOverlayReady,
-    ])
-
-    return { provider: this.fortmatic.getProvider(), chainId: (this as any).chainId, account }
-  }
-}
+// import { FortmaticConnector as FortmaticConnectorCore } from 'web3-react-fortmatic-connector'
```
vm, line 1: dunno if we need this after noah implements inside v8 codebase

src/connectors/NetworkConnector.ts
```diff
@@ -1,215 +0,0 @@
-import invariant from 'tiny-invariant'
```
vm, line 1: there's a network provider in v8 now so maybe that's all we need?

src/hooks/useActiveWeb3React.ts
```diff
@@ -1,23 +1,6 @@
-/* eslint-disable react-hooks/rules-of-hooks */
-import { Web3Provider } from '@ethersproject/providers'
-import { default as useWidgetsWeb3React } from 'lib/hooks/useActiveWeb3React'
-import { useWeb3React } from 'web3-react-core'
-
-import { NetworkContextName } from '../constants/misc'
+import { useWeb3React, Web3ReactProvider } from '@web3-react/core'
 
 export default function useActiveWeb3React() {
-  if (process.env.REACT_APP_IS_WIDGET) {
```
vm, line 9: commented widget out for now

src/hooks/useActiveWeb3React.ts
```diff
@@ -1,23 +1,6 @@
...
-    return useWidgetsWeb3React()
-  }
-
-  const interfaceContext = useWeb3React<Web3Provider>()
```
vm, line 13: we need to handle not having the network provider at all times now

src/hooks/web3.ts
```diff
@@ -12,22 +13,22 @@ export function useEagerConnect() {
...
-          activate(gnosisSafe, undefined, true).catch(() => {
-            setTriedSafe(true)
-          })
-        } else {
-          setTriedSafe(true)
-        }
-      })
-    }
-  }, [activate, setTriedSafe, triedSafe])
+  // useEffect(() => {
```
vm, line 19: commented out gnosis safe stuff for now until added to v8

src/utils/switchToNetwork.ts
```diff
@@ -1,12 +1,12 @@
...
 import { hexStripZeros } from '@ethersproject/bytes'
-import { ExternalProvider } from '@ethersproject/providers'
+import { Web3ReactProvider } from '@web3-react/core'
 import { CHAIN_INFO } from 'constants/chainInfo'
 import { SupportedChainId } from 'constants/chains'
 import { INFURA_NETWORK_URLS } from 'constants/infura'
 
 interface SwitchNetworkArguments {
-  provider: ExternalProvider
+  provider: Web3ReactProvider
```
vm, line 9: not sure about this?

src/components/WalletModal/index.tsx
```diff
@@ -137,12 +139,12 @@ export default function WalletModal({
   ENSName?: string
 }) {
   // important that these are destructed from the account-specific web3-react context
-  const { active, account, connector, activate, error } = useWeb3React()
```
vm, line 140: i'm not sure how to port over activate into v8 properly?

src/components/WalletModal/index.tsx
NoahZinsmeister: see other comment

src/components/WalletModal/index.tsx
NoahZinsmeister: hm bit weird that we used `getAccount`, probably shouldn't be that way haha - in v8 you can't query connectors for the account(s) - gotta get it from the store

src/components/WalletModal/index.tsx
NoahZinsmeister: should be!

src/components/earn/StakingModal.tsx
NoahZinsmeister: yes should be provider!

src/connectors/index.ts
```diff
@@ -1,49 +1,64 @@
...
-})
-
-export const walletlink = new WalletLinkConnector({
-  url: INFURA_NETWORK_URLS[SupportedChainId.MAINNET],
-  appName: 'Uniswap',
-  appLogoUrl: UNISWAP_LOGO_URL,
-  supportedChainIds: ALL_SUPPORTED_CHAIN_IDS,
-})
+export const [injected, injectedHooks] = initializeConnector<MetaMask>(
+  (actions) => new MetaMask(actions, { supportedChainIds: ALL_SUPPORTED_CHAIN_IDS })
```
NoahZinsmeister, line 28: hm this doesn't look quite right - 1st argument to `initializeConnector` is the (actions) => function, but the MetaMask connector constructor takes the following args: https://github.com/NoahZinsmeister/web3-react/blob/3e0a0785f27a1a0413a2f81d2b0669c017bd4bf0/packages/metamask/src/index.ts#L36

also, the supported chain ids argument should be passed as the 2nd arg, see https://github.com/NoahZinsmeister/web3-react/blob/3e0a0785f27a1a0413a2f81d2b0669c017bd4bf0/packages/core/src/hooks.ts#L28

src/components/Web3ReactManager/index.tsx
```diff
@@ -19,24 +19,24 @@ const Message = styled.h2`
 `
 
 export default function Web3ReactManager({ children }: { children: JSX.Element }) {
-  const { active } = useWeb3React()
-  const { active: networkActive, error: networkError, activate: activateNetwork } = useWeb3React(NetworkContextName)
+  const { isActive } = useWeb3React()
+  const { isActive: networkActive, error: networkError, activate: activateNetwork } = useWeb3React(NetworkContextName)
 
   // try to eagerly connect to an injected provider, if it exists and has granted access already
   const triedEager = useEagerConnect()
```
NoahZinsmeister, line 26: ah, so this is another piece of code that's unnecessary - in v8 eager connections can be automatic - for example if we pass the `connectEagerly` flag to the metamask provider (https://github.com/NoahZinsmeister/web3-react/blob/3e0a0785f27a1a0413a2f81d2b0669c017bd4bf0/packages/metamask/src/index.ts#L36) this logic will happen automatically on page load. there's some nuance if you're doing SSR, but we're not, so no worries there.

you should be able to pass this `connectEager` flag to any connector which supports the pattern

NoahZinsmeister (commented):
i think this is a good start! left some comments, more than happy to offer more clarifications or even dive into the code to get it into a working state

src/components/AccountDetails/index.tsx
vm: might be good to have a `type` per connector and then export an Enum or something?

src/components/WalletModal/index.tsx
vm: yea i think we'd need to handle this activate logic differently

src/hooks/useActiveWeb3React.ts
vm: i think we can actually permanently remove this...?

src/components/Web3ReactManager/index.tsx
vm: i think it's just metamask for eager, but lmk if otherwise

Liutauras68 (approved):

src/hooks/useFetchListCallback.ts
```diff
@@ -6,26 +6,21 @@ import resolveENSContentHash from 'lib/utils/resolveENSContentHash'
...
   const ensResolver = useCallback(
     async (ensName: string) => {
-      if (!library || chainId !== 1) {
-        const networkLibrary = getNetworkLibrary()
-        const network = await networkLibrary.getNetwork()
-        if (networkLibrary && network.chainId === 1) {
-          return resolveENSContentHash(ensName, networkLibrary)
-        }
-        throw new Error('Could not construct mainnet ENS resolver')
+      if (provider) {
```
vm, line 17: need to fix this

src/components/WalletModal/index.tsx
```diff
@@ -143,6 +146,13 @@ export default function WalletModal({
 
   const isNetworkConnector = connector === network
 
+  // on error, disconnect the wallet override
```
vm, line 149: the error does not seem to propagate?

src/components/AccountDetails/index.tsx
```diff
@@ -265,11 +265,12 @@ export default function AccountDetails({
               <AccountGroupingRow>
                 {formatConnectorName()}
                 <div>
-                  {connector !== injected && connector !== walletlink && (
+                  {connector !== injected && (
                     <WalletAction
                       style={{ fontSize: '.825rem', fontWeight: 400, marginRight: '8px' }}
                       onClick={() => {
-                        ;(connector as any).close()
+                        dispatch(setWalletOverride({ wallet: undefined }))
```
vm, line 272: i'm open to moving this somewhere more async, since i don't love that it's coupled. i've used redux-observable or redux-saga for this type of stuff before.

src/hooks/useActiveWeb3React.ts
```diff
@@ -1,23 +1,5 @@
-/* eslint-disable react-hooks/rules-of-hooks */
```
vm, line 1: we can kill this file after merge

src/index.tsx
```diff
@@ -46,26 +56,149 @@ function Updaters() {
   )
 }
 
+const Wrapper2 = () => {
```
vm, line 59: I can pair with you to explain this code if helpful..

package.json
```diff
@@ -204,11 +197,16 @@
     "@uniswap/token-lists": "^1.0.0-beta.27",
     "@uniswap/v2-sdk": "^3.0.1",
     "@uniswap/v3-sdk": "^3.8.2",
-    "@web3-react/core": "^8.0.23-beta.0",
-    "@web3-react/eip1193": "^8.0.18-beta.0",
-    "@web3-react/empty": "^8.0.12-beta.0",
-    "@web3-react/types": "^8.0.12-beta.0",
-    "@web3-react/url": "^8.0.17-beta.0",
+    "@web3-react/coinbase-wallet": "^8.0.30-beta.0",
```
zzmp, line 200: @NoahZinsmeister at what point do these come out of beta?

src/state/user/reducer.ts
```diff
@@ -202,5 +212,6 @@ export const {
   updateUserExpertMode,
   updateUserLocale,
   updateUserSlippageTolerance,
+  setWalletOverride,
```
zzmp, line 215: nit: `updateWalletOverride` for consistency.

src/constants/wallet.ts
```diff
@@ -56,20 +90,12 @@ export const SUPPORTED_WALLETS: { [key: string]: WalletInfo } = {
...
     name: 'Open in Coinbase Wallet',
+    wallet: Wallet.COINBASE_WALLET,
     iconURL: COINBASE_ICON_URL,
     description: 'Open in Coinbase Wallet app.',
     href: 'https://go.cb-w.com/mtUDhEZPy1',
     color: '#315CF5',
     mobile: true,
     mobileOnly: true,
   },
-  FORTMATIC: {
```
zzmp, line 66: Why are you deleting these wallets?

src/state/user/reducer.ts
```diff
@@ -50,6 +51,9 @@ export interface UserState {
   showSurveyPopup: boolean | undefined
 
   showDonationLink: boolean
+
+  walletOverride: Wallet | null
+  walletOverrideBackfilled: boolean
```
zzmp, line 56: Is this meant to store whether the walletOverride has been set or not?
If so, can you instead set `walletOverride: Wallet | null | undefined`? It's unclear to me what `Backfilled` specifically does.

src/connectors/index.ts
```diff
@@ -1,49 +1,59 @@
...
+        url: INFURA_NETWORK_URLS[SupportedChainId.MAINNET],
+        appName: 'Uniswap',
+        appLogoUrl: UNISWAP_LOGO_URL,
+      },
+      true
+    ),
+  ALL_SUPPORTED_CHAIN_IDS
+)
+
+// this is an ordered priority list. network connector should be at the top because we want to always
```
zzmp, line 51: nits:
- Use full sentences (eg capitalize).
- "list ordered by priority" instead of "ordered priority list", to avoid a reader mistaking it for a data structure (like a "priority queue")

src/connectors/index.ts
```diff
@@ -1,49 +1,59 @@
...
+// pass in a walletOverride manually for the connected wallet
+export const connectors: [GnosisSafe | MetaMask | WalletConnect | CoinbaseWallet | Network, Web3ReactHooks][] = [
```
zzmp, line 53: Could this be simplified by typing it as `[Connector, Web3ReactHooks][]`, or would that prevent narrowing in other files?

src/connectors/index.ts
```diff
@@ -1,49 +1,59 @@
...
-  appName: 'Uniswap',
-  appLogoUrl: UNISWAP_LOGO_URL,
-  supportedChainIds: ALL_SUPPORTED_CHAIN_IDS,
-})
+export const [network, networkHooks] = initializeConnector<Network>(
+  (actions) => new Network(actions, INFURA_NETWORK_URLS, true, 1),
+  Object.keys(INFURA_NETWORK_URLS).map((chainId) => Number(chainId))
+)
+
+export const [injected, injectedHooks] = initializeConnector<MetaMask>(
```
zzmp, line 17: Can you rename this to `[metamask, metamaskHooks]` instead of `injected`?

src/connectors/index.ts
```diff
@@ -1,49 +1,59 @@
...
-})
-
-export const walletlink = new WalletLinkConnector({
-  url: INFURA_NETWORK_URLS[SupportedChainId.MAINNET],
-  appName: 'Uniswap',
-  appLogoUrl: UNISWAP_LOGO_URL,
-  supportedChainIds: ALL_SUPPORTED_CHAIN_IDS,
-})
+export const [network, networkHooks] = initializeConnector<Network>(
+  (actions) => new Network(actions, INFURA_NETWORK_URLS, true, 1),
```
zzmp, line 13: nit: for arguments where the intent is unclear (ie booleans), annotate with an inline comment:

```
  (actions) => new Network(actions, INFURA_NETWORK_URLS, /* connectEagerly= */ true, /* defaultChainId= */ 1),
```

If you write methods like this, use an option pack instead so that the parameters may be named explicitly.

src/index.tsx
```diff
@@ -46,26 +56,154 @@ function Updaters() {
   )
 }
 
+const Web3Wrapper = () => {
```
zzmp, line 59: Move this into its own file.

src/index.tsx
```diff
@@ -46,26 +56,154 @@ function Updaters() {
...
+    isInjectedEagerlyConnecting,
+    isCoinbaseWalletEagerlyConnecting,
+    isWalletConnectEagerlyConnecting,
+    setIsInjectedEagerlyConnecting,
+    setIsCoinbaseWalletEagerlyConnecting,
+    setIsWalletConnectEagerlyConnecting,
+  ])
+
+  return (
+    <Blocklist>
```
zzmp, line 177: This should be passed in as props.children and then rendered, otherwise it couples all of the app state together.

src/index.tsx
```diff
@@ -46,26 +56,154 @@ function Updaters() {
...
+
+  const injectedIsActivating = hooks.useSelectedIsActivating(injected)
+  const coinbaseWalletIsActivating = hooks.useSelectedIsActivating(coinbaseWallet)
+  const walletConnectIsActivating = hooks.useSelectedIsActivating(walletConnect)
+
+  const previousInjectedIsActivating = usePrevious(injectedIsActivating)
+  const previousCoinbaseWalletIsActivating = usePrevious(coinbaseWalletIsActivating)
+  const previousWalletConnectIsActivating = usePrevious(walletConnectIsActivating)
+
+  const [isInjectedEagerlyConnecting, setIsInjectedEagerlyConnecting] = useReducer(
```
zzmp, line 80: I think these would be simpler as `useState`s, because there are no side effects or computations.

src/index.tsx
```diff
@@ -46,26 +56,154 @@ function Updaters() {
...
+        previousIsActivating,
+        isEagerlyConnecting,
+        setIsEagerlyConnecting,
+      } = state
+
+      if (!isActivating && previousIsActivating) {
+        setIsEagerlyConnecting(false)
+      }
+
+      if (isActive === false && previousIsActive === undefined && isActivating) {
```
zzmp, line 138: Why is `previousIsActivating` relevant?

src/index.tsx
```diff
@@ -46,26 +56,154 @@ function Updaters() {
...
+      isEagerlyConnecting: isWalletConnectEagerlyConnecting,
+      setIsEagerlyConnecting: setIsWalletConnectEagerlyConnecting,
+    }
+    const isActiveMap = new Map<Wallet, ConnectorState>([
+      [Wallet.INJECTED, injectedState],
+      [Wallet.COINBASE_WALLET, coinbaseWalletState],
+      [Wallet.WALLET_CONNECT, walletConnectState],
+    ])
+
+    isActiveMap.forEach((state: ConnectorState, wallet: Wallet) => {
```
zzmp, line 124: I'm unsure why the previous states are relevant. Can you leave some comments explaining this logic?

src/components/WalletModal/index.tsx
```diff
@@ -135,105 +130,85 @@ export default function WalletModal({
...
-
-  // close on connection, when logged out before
-  useEffect(() => {
-    if (account && !previousAccount && walletModalOpen) {
-      toggleWalletModal()
-    }
-  }, [account, previousAccount, toggleWalletModal, walletModalOpen])
-
-  // always reset to account view
+  const connectorPrevious = usePrevious(connector)
```
zzmp, line 148: nit: previousConnector, not connectorPrevious.

src/components/WalletModal/index.tsx
```diff
@@ -135,105 +130,85 @@ export default function WalletModal({
...
+      setWalletView(WALLET_VIEWS.OPTIONS)
     }
-  }, [setWalletView, active, error, connector, walletModalOpen, activePrevious, connectorPrevious])
+  }, [connector])
 
-  const tryActivation = async (connector: AbstractConnector | undefined) => {
+  const tryActivation = async (connector: Connector) => {
     let name = ''
-    Object.keys(SUPPORTED_WALLETS).map((key) => {
+    Object.keys(SUPPORTED_WALLETS).forEach((key) => {
```
zzmp, line 163: forEach is innapropriate here - can you use a for-loop so that it short-circuits.

src/components/Identicon/StatusIcon.tsx
```diff
@@ -1,22 +1,18 @@
...
   switch (connector) {
     case injected:
       return <Identicon />
-    case walletconnect:
+    case walletConnect:
       return <img src={WalletConnectIcon} alt={'WalletConnect'} />
-    case walletlink:
+    case coinbaseWallet:
       return <img src={CoinbaseWalletIcon} alt={'Coinbase Wallet'} />
-    case fortmatic:
```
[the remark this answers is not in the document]
vm: removed for now until web3-react supports it, which requires them to implement a few EIPs

src/utils/switchToNetwork.ts
```diff
@@ -36,48 +29,13 @@ function getRpcUrls(chainId: SupportedChainId): [string] {
...
-      } catch (error) {
-        console.debug('Added network but could not switch chains', error)
-      }
-    } else {
-      throw error
-    }
-  }
+export async function switchToNetwork(connector: Connector, chainId: SupportedChainId): Promise<null | void> {
+  const info = CHAIN_INFO[chainId]
+  connector.activate({
```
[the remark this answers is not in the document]
vm: https://github.com/NoahZinsmeister/web3-react/blob/main/packages/metamask/src/index.ts#L120 docs here

src/components/AccountDetails/index.tsx
vm: we actually wouldn't need anything beyond basic redux now that i'm thinking more. eventually it'd be good to just pass an onClick into a generic WaleltModal component

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -97,9 +98,27 @@ function TransactionSubmittedContent({
...
+
+  const [success, setSuccess] = useState<boolean | undefined>()
+
+  const addToken = useCallback(() => {
+    token &&
+      token.symbol &&
+      connector.watchAsset &&
+      connector
+        .watchAsset({
+          address: token.address,
```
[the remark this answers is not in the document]
vm: i tried this, but then i'd be passing in extra props that aren't allowed

src/components/Web3Status/index.tsx
```diff
@@ -213,9 +211,7 @@ export default function Web3Status() {
   return (
     <>
       <Web3StatusInner />
-      {(contextNetwork.active || active) && (
-        <WalletModal ENSName={ENSName ?? undefined} pendingTransactions={pending} confirmedTransactions={confirmed} />
-      )}
+      <WalletModal ENSName={ENSName ?? undefined} pendingTransactions={pending} confirmedTransactions={confirmed} />
```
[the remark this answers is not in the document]
vm: i don't think it actually makes sense to gate the modal based on active wallet. if you're connected to coinbase wallet and it's inactive you should be able to switch out of it?

src/state/user/reducer.ts
vm: i can change walletOverride to `Wallet | undefined`, it doesn't need to be null

src/connectors/index.ts
vm: i'm actually opposed because theoretically injected can be used for wallets other than metamask (see this pr https://github.com/Uniswap/interface/pull/3820). imo MetaMask should be renamed to Injected in web3-react.

src/hooks/useFetchListCallback.ts
vm: i can't really do that with the way web3-react works. it used to work with v6 though... i'd like to implement caching the ENS and balance though

src/constants/wallet.ts
vm: commented above, but waiting on fortmatic to implement some EIPs

src/components/Web3Wrapper/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  const injectedIsActivating = hooks.useSelectedIsActivating(injected)
+  const coinbaseWalletIsActivating = hooks.useSelectedIsActivating(coinbaseWallet)
+  const walletConnectIsActivating = hooks.useSelectedIsActivating(walletConnect)
+
+  const [isInjectedEagerlyConnecting, setIsInjectedEagerlyConnecting] = useState(false)
+  const [isCoinbaseWalletEagerlyConnecting, setIsCoinbaseWalletEagerlyConnecting] = useState(false)
+  const [isWalletConnectEagerlyConnecting, setIsWalletConnectEagerlyConnecting] = useState(false)
+
+  useEffect(() => {
```
zzmp, line 42: Could you define these objects directly? Because you are always defining these objects, I think it is ok to run a forEach and ignore the rule of hooks and omit a lot of the boilerplate - I think that may make it more readable / easier to understand.

src/hooks/useFetchListCallback.ts
zzmp: Can't you have a separate web3-react network connector that only connects to mainnet? It doesn't need to go through the Web3ReactProvider.

src/index.tsx
```diff
@@ -46,26 +46,34 @@ function Updaters() {
   )
 }
 
+const ReduxWrapper = () => {
+  const walletOverride = useAppSelector((state) => state.user.walletOverride)
```
zzmp, line 50: Back out this change. Get walletOverride/connectorOverride in the separate Web3Provider component (see overall comments).

src/utils/switchToNetwork.ts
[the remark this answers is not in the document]
zzmp: Can you cast the arguments to ensure that they are correct? I don't think the type checker guards you here.

zzmp (changes requested):
`Web3Wrapper` should be renamed `Web3Provider` and wrap the `Web3ReactProvider` inside of it. This way the Web3 logic is self-contained in a single provider, eg:

```
function Web3Updater() {
  // Move the logic from Web3Wrapper here.
}

export default function Web3Provider({children}) {
  ...
  return (
    <Web3ReactProvider connectors={connectors} connectorOverride={connectorOverride}>
      <Web3Updater />
    </Web3ReactProvider>
  )
}
```

src/components/AccountDetails/index.tsx
```diff
@@ -265,11 +265,13 @@ export default function AccountDetails({
               <AccountGroupingRow>
                 {formatConnectorName()}
                 <div>
-                  {connector !== injected && connector !== walletlink && (
+                  {/* coinbase wallet reloads the page right now, which breaks the walletOverride from being set properly on localStorage */}
```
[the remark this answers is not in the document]
vm: image
some more context here. felt it'd be best to wait for them to support this instead of adding specific logic now. i will fix comment

src/hooks/useFetchListCallback.ts
vm: oh i see. yeah let me try that out.

src/components/Header/NetworkSelector.tsx
```diff
@@ -238,7 +238,7 @@ const getChainNameFromId = (id: string | number) => {
 }
 
 export default function NetworkSelector() {
-  const { chainId, library } = useActiveWeb3React()
+  const { chainId, provider, connector } = useActiveWeb3React()
```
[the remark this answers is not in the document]
vm: i'm a little too scared to remove all the checks that make sure provider (previously library) exists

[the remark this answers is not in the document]
zzmp: Fair. You could also make a case that it's scope creep for this PR to clean up `provider` usages.

src/state/lists/updater.ts
```diff
@@ -48,7 +48,7 @@ export default function Updater(): null {
         fetchList(listUrl).catch((error) => console.debug('list added fetching error', error))
       }
     })
-  }, [dispatch, fetchList, library, lists])
+  }, [dispatch, fetchList, provider, lists])
```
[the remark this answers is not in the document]
zzmp: Yes. Let's call it out of scope for this PR.

src/components/Header/NetworkSelector.tsx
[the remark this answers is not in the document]
vm: that's mostly what i mean. i think they're probably unnecessary and we happen to be doing a lot of defensive checks that should be removed. at the same time i don't want to do it in here and incraese the scope

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -97,9 +96,25 @@ function TransactionSubmittedContent({
...
+  const { connector } = useActiveWeb3React()
 
-  const { addToken, success } = useAddTokenToMetamask(currencyToAdd)
+  const token = currencyToAdd?.wrapped
+  const logoURL = useCurrencyLogoURIs(token)[0]
+
+  const [success, setSuccess] = useState<boolean | undefined>()
+
+  const addToken = useCallback(() => {
+    if (!token?.symbol || !connector.watchAsset) return
```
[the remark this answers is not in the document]
vm: gave this a shot and couldn't really get it to work. honestly not convinced it makes things easier to read so just leaving for now

cypress/integration/wallet.test.ts
```diff
@@ -0,0 +1,35 @@
...
+    cy.get('[data-cy=option-grid]').should('exist')
+  })
+
+  it('selects injected wallet option', () => {
+    cy.contains('Injected').click()
+    cy.get('#web3-account-identifier-row').contains(TEST_ADDRESS_NEVER_USE_SHORTENED)
+  })
+
+  it('disconnects the wallet', () => {
+    cy.get('[data-cy=wallet-disconnect]').click()
```
zzmp, line 27: There's no test here; you should look for some change after you click the button.

cypress/tsconfig.json
```diff
@@ -1,7 +1,5 @@
 {
   "compilerOptions": {
-    "strict": true,
```
zzmp, line 3: Why not `strict`?

cypress/tsconfig.json
```diff
@@ -1,7 +1,5 @@
 {
```
zzmp, line 1: Do you know where this tsconfig is actually used?

src/components/AccountDetails/index.tsx
```diff
@@ -265,12 +265,16 @@ export default function AccountDetails({
               <AccountGroupingRow>
                 {formatConnectorName()}
                 <div>
-                  {connector !== injected && connector !== walletlink && (
+                  {/* Coinbase Wallet disconnection causes a page reload from their SDK today. This leads to a race condition where the local storage value for walletOverride may not be set in time. */}
```
zzmp, line 268: nit: omit "today".

src/components/AccountDetails/index.tsx
```diff
@@ -265,12 +265,16 @@ export default function AccountDetails({
...
+                  {/* @TODO(vm): Add disconnect for Coinbase Wallet once the team allows preventing reload via SDK */}
```
zzmp, line 269: nit: I think `@TODO` is a Java idiom. You can just use `TODO(vm): ...`.

src/components/AccountDetails/index.tsx
zzmp, line 268: Does this mean that Coinbase Wallet cannot be disconnected from within the interface page?

src/utils/switchToNetwork.ts
```diff
@@ -36,48 +29,13 @@ function getRpcUrls(chainId: SupportedChainId): [string] {
...
+    chainId,
+    chainName: info.label,
+    rpcUrls: getRpcUrls(chainId),
+    nativeCurrency: info.nativeCurrency,
+    blockExplorerUrls: [info.explorer],
+  })
```
zzmp, line 40: Add `as AddEthereumChainParameter`.

src/index.tsx
```diff
@@ -46,26 +42,30 @@ function Updaters() {
   )
 }
 
+const ReduxWrapper = () => {
+  return (
+    <HashRouter>
```
zzmp, line 47: Inline this. There's no need for a separate `ReduxWrapper` const.

src/components/WalletModal/index.tsx
```diff
@@ -135,109 +130,95 @@ export default function WalletModal({
...
-  }, [toggleWalletModal])
-
   // get wallets user can switch too, depending on device/browser
   function getOptions() {
     const isMetamask = !!window.ethereum?.isMetaMask
     return Object.keys(SUPPORTED_WALLETS).map((key) => {
       const option = SUPPORTED_WALLETS[key]
+      const isActive = option.connector === connector
+
+      const baseProps = {
```
zzmp, line 200: nit: rename to `optionProps`.

src/components/Web3Provider/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+interface ConnectorState {
+  isActive: boolean
+  previousIsActive: boolean | undefined
+  isActivating: boolean
+  isEagerlyConnecting: boolean
+  setIsEagerlyConnecting(connecting: boolean): void
+}
+
+const Web3Updater = () => {
```
zzmp, line 21: Add a doc comment explaining what `Web3Updater` does (because it's not obvious and hard to read through the logic).

zzmp (changes requested):

cypress/tsconfig.json
vm: oops i didn't mean to change this, i'll remove changes. but it gets used locally when you run cypress

src/components/AccountDetails/index.tsx
vm: that's true on prod also, but yeah that would be the case for now

zzmp (approved):
Update the `'disconnects the wallet'` test in `cypress/integration/wallet.test.ts`, and add a doc comment for `Web3Updater`, and this lgtm!

src/connectors/index.ts
```diff
@@ -1,49 +1,130 @@
...
+  (actions) =>
+    new WalletConnect(actions, {
+      rpc: INFURA_NETWORK_URLS,
+      qrcode: true,
+    }),
+  ALL_SUPPORTED_CHAIN_IDS
+)
+
+const fortmaticProvider = new Fortmatic(process.env.REACT_APP_FORTMATIC_KEY ?? 'pk_live_357F77728B8EB880').getProvider()
+fortmaticProvider.on = () => {
```
vm, line 94: they need to implement the Event spec from eip1193, for now i'm stubbing it out

src/connectors/index.ts
```diff
@@ -1,49 +1,126 @@
...
+  (actions) =>
+    new WalletConnect(actions, {
+      rpc: INFURA_NETWORK_URLS,
+      qrcode: true,
+    }),
+  ALL_SUPPORTED_CHAIN_IDS
+)
+
+export const [fortmatic, fortmaticHooks] = initializeConnector<EIP1193>(
+  (actions) => new EIP1193(actions, new Fortmatic(process.env.REACT_APP_FORTMATIC_KEY).getProvider())
```
NoahZinsmeister, line 94: the one thing that's a little unfortunate about this approach is that it loads the fortmatic sdk for all users rather than dynamically on import like all the other connectors. not a huge deal as it's prety small: https://bundlephobia.com/package/fortmatic@2.4.0

one easy but somewhat hacky fix would be make a fake 1193 connector that dynamically imports the fortmatic sdk on the first request and then caches it, but that's optional

src/components/WalletModal/index.tsx
```diff
@@ -126,101 +131,104 @@ export default function WalletModal({
...
+  // close on connection/disconnection
   useEffect(() => {
-    if (account && !previousAccount && walletModalOpen) {
+    if (walletModalOpen && account !== previousAccount) {
       toggleWalletModal()
     }
   }, [account, previousAccount, toggleWalletModal, walletModalOpen])
 
-  // always reset to account view
+  useEffect(() => {
```
zzmp, line 167: Is this `useEffect` necessary, or is it duplicating logic that comes in the `useEffect` immeidately after it?

src/components/Web3Provider/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  const [eagerlyConnectingWallets, setEagerlyConnectingWallets] = useState(new Set())
+
+  useEffect(() => {
+    if (error) {
+      console.error(`web3-react error: ${error}`)
+    }
+  }, [error])
+
+  useEffect(() => {
```
zzmp, line 51: Add a comment that this is only meant to run on mount (as it has no deps).

src/components/Web3Provider/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+        // When a user manually sets their new connection, set a wallet override.
+        // Also set an override when they were a user prior to this state being introduced.
+        if (!isEagerlyConnecting || !walletOverrideBackfilled) {
+          dispatch(updateWalletOverride({ wallet }))
+        }
+
+        // Reset the eagerly connecting state.
+        if (isEagerlyConnecting) {
+          eagerlyConnectingWallets.delete(wallet)
+          setEagerlyConnectingWallets(new Set([...eagerlyConnectingWallets]))
```
zzmp, line 100: Once we have one eagerly connected wallet, do we need to continue trying to connect to the others? Or can we replace this with an empty set?

src/connectors/index.ts
```diff
@@ -1,49 +1,126 @@
...
-})
-
-export const walletlink = new WalletLinkConnector({
-  url: INFURA_NETWORK_URLS[SupportedChainId.MAINNET],
-  appName: 'Uniswap',
-  appLogoUrl: UNISWAP_LOGO_URL,
-  supportedChainIds: ALL_SUPPORTED_CHAIN_IDS,
-})
+export const createOrderedConnectors = (walletOverride: Wallet | undefined) => {
+  const connectors: ConnectorListItem[] = [{ connector: gnosisSafe, hooks: gnosisSafeHooks }]
```
zzmp, line 113: Have you tested with gnosis?

src/hooks/useFetchListCallback.ts
```diff
@@ -6,26 +6,21 @@ import resolveENSContentHash from 'lib/utils/resolveENSContentHash'
...
+        return resolveENSContentHash(ensName, provider)
+      } else {
+        return Promise.resolve('')
```
zzmp, line 20: Could you have a separate mainnet connector (or provider) to do this? It doesn't necessarily need to go through @web3-react - it _could_ just be an ethers url provider for mainnet.

src/components/WalletModal/index.tsx
```diff
@@ -126,101 +131,104 @@ export default function WalletModal({
...
-  const [pendingError, setPendingError] = useState<boolean>()
+  const [pendingConnector, setPendingConnector] = useState<Connector | undefined>()
+  // Need to pass network as a default case because useSelectedError requires a connector
+  const pendingError = hooks.useSelectedError(pendingConnector || network)
 
   const walletModalOpen = useModalOpen(ApplicationModal.WALLET)
   const toggleWalletModal = useWalletModalToggle()
 
+  const previousConnector = usePrevious(connector)
   const previousAccount = usePrevious(account)
```
zzmp, line 154: Because the `account` is linked to the `connector`, can you omit `previousAccount` and use `connection` and `previousConnection` instead?

src/components/Web3Provider/index.tsx
zzmp, line 100: This may cause recursive calls, because eagerlyConnectingWallets is in the deps.

zzmp: Is there a reason the eagerly connecting state needs to be reset?

src/components/Header/NetworkSelector.tsx
```diff
@@ -251,37 +250,17 @@ export default function NetworkSelector() {
...
-            history.replace({ search: replaceURLParam(history.location.search, 'chain', getChainNameFromId(chainId)) })
-          }
+  const onSelectChain = useCallback(
+    async (targetChain: number, skipToggle?: boolean) => {
+      if (!connector) return
 
-          if (!skipToggle) {
-            toggle()
-          }
+      switchChain(connector, targetChain)
```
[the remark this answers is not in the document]
vm: @NoahZinsmeister example of the confusion

unfortunately `connector.activate` doesn't throw an error, it provides the error via the hook. to do this we'd need to change how web3-react works a bit

src/components/Web3Provider/index.tsx
vm: https://github.com/Uniswap/interface/pull/3759/files#diff-427a7fa283d9878f0fdf72649b3468c9b117e5136f4484cbdf6bf52c05cee799R93 i need it for this check

src/components/Header/NetworkSelector.tsx
```diff
@@ -251,37 +250,17 @@ export default function NetworkSelector() {
...
 
-          dispatch(addPopup({ content: { failedSwitchNetwork: targetChain }, key: `failed-network-switch` }))
-        })
+      if (!skipToggle) {
+        toggle()
```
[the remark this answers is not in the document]
vm: since we await inside of `switchChain`, this will happen after the switch

src/hooks/useFetchListCallback.ts
vm: good point, this is doable

src/connectors/index.ts
vm: yup, working 👍

src/components/Header/index.tsx
```diff
@@ -76,11 +77,6 @@ const HeaderElement = styled.div`
     margin-left: 0.5em;
   }
 
-  /* addresses safari's lack of support for "gap" */
```
vm, line 79: add this back

src/hooks/useFetchListCallback.ts
```diff
@@ -1,39 +1,32 @@
...
-        if (networkLibrary && network.chainId === 1) {
-          return resolveENSContentHash(ensName, networkLibrary)
-        }
-        throw new Error('Could not construct mainnet ENS resolver')
-      }
-      return resolveENSContentHash(ensName, library)
-    },
-    [chainId, library]
-  )
+  const provider = networkHooks.useProvider()
```
zzmp, line 21: This will not work if the network was switched from MAINNET. There should either be another network provider that _only_ connects to MAINNET (which would allow this to work with no wallet, regardless of the active network), or this should be switched back to MAINNET when a wallet is connected (which would mean this would not work with no wallet connected when not connected to MAINNET).

src/hooks/useFetchListCallback.ts
```diff
@@ -1,39 +1,32 @@
...
 import getTokenList from 'lib/hooks/useTokenList/fetchTokenList'
 import resolveENSContentHash from 'lib/utils/resolveENSContentHash'
 import { useCallback } from 'react'
 import { useAppDispatch } from 'state/hooks'
 
-import { getNetworkLibrary } from '../connectors'
 import { fetchTokenList } from '../state/lists/actions'
 
+const ensResolver = async (ensName: string, provider: Web3Provider | undefined, isActive: boolean) => {
+  if (!provider || !isActive) {
```
zzmp, line 13: This is missing a check that the provider is on MAINNET.

src/state/wallet/reducer.ts
```diff
@@ -0,0 +1,33 @@
+import { createSlice } from '@reduxjs/toolkit'
```
zzmp, line 1: Why does each wallet need its own error state?

src/state/walletOverride/reducer.ts
```diff
@@ -0,0 +1,30 @@
+import { createSlice } from '@reduxjs/toolkit'
+import { Wallet } from 'connectors'
+
+export interface WalletOverrideState {
+  // We want the user to be able to define which wallet they want to use, even if there are multiple connected wallets via web3-react.
+  // If a user had previously connected a wallet but didn't have a wallet override set (because they connected prior to this field being added),
+  // we want to handle that case by backfilling them manually. Once we backfill, we set the backfilled field to `true`.
+  // After some period of time, our active users will have this property set so we can likely remove the backfilling logic.
+  walletOverrideBackfilled: boolean
+  walletOverride?: Wallet
```
zzmp, line 10: Rather than have `walletOverrideBackfilled`, can you just have `wallet?: Wallet | true`. If undefined, then it should be backfilled - if `true`, then no wallet has been overridden, and none need be eagerly tried. This simplifies the logic, and gets rid of a confusing parameter.

src/state/index.ts
```diff
@@ -16,13 +16,17 @@ import { routingApi } from './routing/slice'
...
 
-const PERSISTED_KEYS: string[] = ['user', 'transactions', 'lists']
+const PERSISTED_KEYS: string[] = ['user', 'transactions', 'lists', 'walletOverride']
 
 const store = configureStore({
   reducer: {
     application,
     user,
+    wallet,
+    walletOverride,
```
zzmp, line 29: `walletOverride` is confusingly named. What do you think about moving it to `user.wallet`?

src/connectors/index.ts
```diff
@@ -1,49 +1,157 @@
...
+    case Wallet.FORTMATIC:
+      return fortmaticHooks
+    case Wallet.NETWORK:
+      return networkHooks
+    case Wallet.GNOSIS_SAFE:
+      return gnosisSafeHooks
+  }
+}
+
+export const [network, networkHooks] = initializeConnector<Network>(
```
zzmp, line 103: A lot of boilerplate could be simplified by grouping these into a single object and referencing that object everywhere. eg

```
function createConnector<T>(initializer: PassThroughToInitializeConnector) {
  const [connector, hooks] = initializeConnector<T>(initializer)
  return { connector, hooks }
}
```

Then you can `export const network = createConnector<Network>(...` and use `network.connector` and `network.hooks` throughout (eg for `WalletModal/index.tsx`, as well as returning the full object from `getConnectorFromWallet`, and using this new object directly for your ordered connectors list).

src/components/WalletModal/index.tsx
```diff
@@ -126,101 +133,118 @@ export default function WalletModal({
   confirmedTransactions: string[] // hashes of confirmed
   ENSName?: string
 }) {
-  // important that these are destructed from the account-specific web3-react context
-  const { account, connector, activate, error } = useWeb3React()
+  const dispatch = useAppDispatch()
+  const { connector } = useWeb3React()
+  const isActiveMap: Partial<Record<Wallet, boolean>> = {
```
zzmp, line 138: Why is this `Partial`? Should it be all `ModalWallet`s?

src/components/Web3Provider/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+  ])
+
+  return null
+}
+
+interface Props {
+  children: JSX.Element
+}
+
+export default function Web3Provider({ children }: Props) {
```
zzmp, line 137: nit: if just children, then just inline the props: `{ children }: { children: ReactNode }` (with `import { ReactNode } from 'react'`

src/components/Web3Provider/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+} from 'connectors'
+import usePrevious from 'hooks/usePrevious'
+import { useEffect, useState } from 'react'
+import { useAppDispatch, useAppSelector } from 'state/hooks'
+import { updateWalletOverride } from 'state/walletOverride/reducer'
+
+const connect = async (connector: Connector) => {
+  try {
+    if (connector.connectEagerly) {
+      await connector.connectEagerly()
```
zzmp, line 23: I thought that `connectEagerly` meant it would attempt connection as soon as it is initialized. Is that wrong?

src/components/Web3Provider/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+  previousIsActive: boolean | undefined
+}
+
+// This component handles state changes in web3-react. It eagerly connects to all wallets.
+// It also checks for Coinbase Wallet, Wallet Connect Fortmatic or Injected wallets to become active.
+function Web3Updater() {
+  const dispatch = useAppDispatch()
+
+  const walletOverride = useAppSelector((state) => state.walletOverride.walletOverride)
+  const walletOverrideBackfilled = useAppSelector((state) => state.walletOverride.walletOverrideBackfilled)
```
zzmp, line 43: If you set the type of `walletOverride?: Wallet | true`, then you can simplify to `const backfilled = Boolean(walletOverride)`.

src/components/Web3Provider/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+  const walletConnectIsActive = walletConnectHooks.useIsActive()
+  const previousWalletConnectIsActive = usePrevious(walletConnectIsActive)
+
+  const fortmaticIsActive = fortmaticHooks.useIsActive()
+  const previousFortmaticIsActive = usePrevious(fortmaticIsActive)
+
+  const [isEagerlyConnecting, setIsEagerlyConnecting] = useState(false)
+
+  // The dependency list is empty so this is only run once on mount
```
zzmp, line 59: This belongs above the eslint-disabling comment as a justification for it.

src/components/Web3Provider/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+  useEffect(() => {
+    connect(gnosisSafe)
+    connect(network)
+
+    if (walletOverride) {
```
zzmp, line 64: If you set the type of `walletOverride?: Wallet | true`, this will need to be more specific, maybe to `typeof walletOverride === 'string'`?

src/components/Web3Provider/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+      connect(getConnectorForWallet(walletOverride))
+      setIsEagerlyConnecting(true)
+    } else if (!walletOverrideBackfilled) {
+      MODAL_WALLETS.filter((wallet) => wallet !== Wallet.FORTMATIC)
```
zzmp, line 68: Why is FORTMATIC excluded if it is a modal wallet? Add a comment.

src/components/Web3Provider/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+// This component handles state changes in web3-react. It eagerly connects to all wallets.
+// It also checks for Coinbase Wallet, Wallet Connect Fortmatic or Injected wallets to become active.
+function Web3Updater() {
+  const dispatch = useAppDispatch()
+
+  const walletOverride = useAppSelector((state) => state.walletOverride.walletOverride)
+  const walletOverrideBackfilled = useAppSelector((state) => state.walletOverride.walletOverrideBackfilled)
+
+  const injectedIsActive = injectedHooks.useIsActive()
```
zzmp, line 45: Tighten up all the logic from L45-L55 and L76-113 by avoiding using as much state.

```
const injectedIsActive = injectedHooks.useIsActive()
const coinbaseWalletIsActive = coinbaseWalletHooks.useIsActive()
const walletConnectIsActive = walletConnectHooks.useIsActive()
const fortmaticIsActive = fortmaticHooks.useIsActive()
const activeMap = useMemo(() => ({
  [Wallet.INJECTED]:, injectedIsActive,
  [Wallet.COINBASE_WALLET]:, coinbaseWalletIsActive,
  [Wallet.WALLET_CONNECT]: walletConnectIsActive,
  [Wallet.FORTMATIC]: fortmaticIsActive,
}), [injectedIsActive, coinbaseWalletIsActive, walletConnectIsActive, fortmaticIsActive])
const previousActiveMap = usePrevious(activeMap)

...

useEffect(() => {
  Object.entries(activeMap).forEach(([wallet, active]) => {
    if (active && !previousActiveMap[wallet]) {
      if (isEagerlyConnecting) {
        setIsEagerlyConnecting(false)
      } else if (!walletOverrideBackfilled) {
        // When a user manually sets their new connection, set a wallet override.
        // Also set an override when they were a user prior to this state being introduced.
        dispatch(updateWalletOverride({ wallet }))
      }
    }
  })
}, [...])
```

src/components/WalletModal/index.tsx
zzmp: This should be memoized, or else the object is re-created every render cycle. See the comment starting with "Tighten up all the logic" for suggested logic to build this map.

src/components/WalletModal/index.tsx
```diff
@@ -126,101 +133,118 @@ export default function WalletModal({
...
-    if (walletModalOpen) {
-      resetAccountView()
+    if (pendingConnector && walletView !== WALLET_VIEWS.PENDING) {
+      updateWalletError({ wallet: getWalletForConnector(pendingConnector), error: undefined })
+      setPendingConnector(undefined)
     }
-  }, [walletModalOpen, resetAccountView])
+  }, [pendingConnector, walletView])
+
+  const tryActivation = async (connector: Connector) => {
```
zzmp, line 173: Should this be a `useCallback`?

src/components/WalletModal/index.tsx
```diff
@@ -126,101 +133,118 @@ export default function WalletModal({
...
-      toggleWalletModal()
-    })
-  }, [toggleWalletModal])
+      setPendingConnector(connector)
+      setWalletView(WALLET_VIEWS.PENDING)
+
+      if (isActiveMap[wallet]) {
+        await connector.activate()
+        setWalletView(WALLET_VIEWS.ACCOUNT)
+        dispatch(updateWalletOverride({ wallet }))
```
zzmp, line 200: Why is this only done if the wallet was already active?

zzmp (changes requested):
I think it'd make sense to move `walletOverride`, which is not immediately clear, to `user.wallet`, which I think is more clear for what it stores. You can also include `true` in the type to eliminate `walletOverrideBackfilled` altogether. This should simplify state lookup in a lot of the wallet component logic.

Smaller thing:
- In general, it's cleaner to use named functions (`function abc() {`) than anonymous functions if they are at the top level, because it makes it more obvious that they are functions and not variables when reading through. It also changes their hoisting behavior, forcing you to order things in a specific way (and preventing you from using some otherwise circular constructs).

src/components/AccountDetails/index.tsx
```diff
@@ -265,12 +265,16 @@ export default function AccountDetails({
...
                 {formatConnectorName()}
                 <div>
-                  {connector !== injected && connector !== walletlink && (
+                  {/* Coinbase Wallet reloads the page right now, which breaks the walletOverride from being set properly on localStorage */}
+                  {connector !== coinbaseWallet && (
                     <WalletAction
                       style={{ fontSize: '.825rem', fontWeight: 400, marginRight: '8px' }}
                       onClick={() => {
-                        ;(connector as any).close()
+                        connector.deactivate ? connector.deactivate() : connector.resetState()
```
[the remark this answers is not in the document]
vm: deactivate will reset state internally basically. i think the naming is slightly confusing on the api level

src/components/Header/NetworkSelector.tsx
```diff
@@ -340,10 +344,15 @@ export default function NetworkSelector() {
             <FlyoutHeader>
               <Trans>Select a network</Trans>
             </FlyoutHeader>
-            <Row onSelectChain={handleChainSwitch} targetChain={SupportedChainId.MAINNET} />
-            <Row onSelectChain={handleChainSwitch} targetChain={SupportedChainId.POLYGON} />
-            <Row onSelectChain={handleChainSwitch} targetChain={SupportedChainId.OPTIMISM} />
-            <Row onSelectChain={handleChainSwitch} targetChain={SupportedChainId.ARBITRUM_ONE} />
+            <Row onSelectChain={onSelectChain} targetChain={SupportedChainId.MAINNET} />
+            {/* Formatic is only supported on mainnet, so we hide the other chains */}
```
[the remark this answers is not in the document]
vm: good point

src/components/Header/index.tsx
```diff
@@ -76,7 +77,7 @@ const HeaderElement = styled.div`
...
+  /* addresses safaris lack of support for "gap" */
```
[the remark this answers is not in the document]
vm: i just did it because it broke all syntax highlighting in my editor haha

src/components/Web3Status/index.tsx
```diff
@@ -140,9 +139,10 @@ function WrappedStatusIcon({ connector }: { connector: AbstractConnector | Conne
 }
 
 function Web3StatusInner() {
-  const { account, connector, error } = useWeb3React()
+  const { account, connector, chainId, ENSName } = useWeb3React()
+  const error = useAppSelector((state) => state.wallet.errorByWallet[getWalletForConnector(connector)])
```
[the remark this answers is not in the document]
vm: we completely removed the error from web3-react now, so you have to catch errors in app instead

src/hooks/useFetchListCallback.ts
vm: yea you're right, will fix

src/components/Web3Provider/index.tsx
vm: connectEagerly isn’t an argument on the connector anymore, need to handle it manually. any connector that doesn’t show a prompt you just use activate (like network connector)

src/state/wallet/reducer.ts
vm: explained above, but the reason is it's easier to store it this way that to have a concept of a "pendingError" and "currentError"

src/connectors/index.ts
vm: why would it be dramatic? it's a tiny package. happy to do it tho

src/components/Header/index.tsx
```diff
@@ -265,7 +268,7 @@ export default function Header() {
   const {
     infoLink,
     nativeCurrency: { symbol: nativeCurrencySymbol },
-  } = CHAIN_INFO[chainId ? chainId : SupportedChainId.MAINNET]
+  } = CHAIN_INFO[!chainId || chainNotAllowed ? SupportedChainId.MAINNET : chainId]
```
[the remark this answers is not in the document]
vm: i think it's better to show a "network not supported error" and allow them to switch to mainnet themselves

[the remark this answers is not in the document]
vm: i think it's better to show a "network not supported error" and allow them to switch to mainnet themselves

src/components/Web3Status/index.tsx
```diff
@@ -157,7 +157,25 @@ function Web3StatusInner() {
   const hasSocks = useHasSocks()
   const toggleWalletModal = useWalletModalToggle()
 
-  if (account) {
+  if (chainNotAllowed) {
+    return (
+      <Web3StatusError onClick={toggleWalletModal}>
+        <NetworkIcon />
+        <Text>
+          <Trans>Wrong Network</Trans>
```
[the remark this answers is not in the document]
vm: leaving for now since there is a design change to be done here tracked in a sep jira ticket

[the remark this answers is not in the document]
vm: leaving for now since there is a design change to be done here tracked in a sep jira ticket

src/connectors/index.ts
NoahZinsmeister: https://bundlephobia.com/package/fortmatic@2.4.0

vm: i need to manually make a fortmatic connector to implement connectEagerly properly anyways, so will handle it when i do that

src/utils/switchChain.ts
```diff
@@ -0,0 +1,64 @@
...
+    case gnosisSafe:
+      return ALL_SUPPORTED_CHAIN_IDS.includes(chainId)
+    default:
+      return false
+  }
+}
+
+export const switchChain = async (connector: Connector, chainId: number) => {
+  if (!isChainAllowed(connector, chainId)) {
+    throw Error('chain not supported on this connector')
```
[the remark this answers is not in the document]
vm: agree that there should be at least an id passed into web3-react @NoahZinsmeister

src/components/WalletModal/index.tsx
```diff
@@ -126,101 +119,109 @@ export default function WalletModal({
...
 
-    // if the connector is walletconnect and the user has already tried to connect, manually reset the connector
-    if (connector instanceof WalletConnectConnector) {
-      connector.walletConnectProvider = undefined
-    }
+      try {
+        // Fortmatic opens it's own modal on activation to log in. This modal has a tabIndex
+        // collision into the WalletModal, so we special case by closing the modal.
+        if (connector === fortmatic) {
+          toggleWalletModal()
```
zzmp, line 171: Replace this with the old logic so there is not a delay between closing our modal and opening theirs.

```
fortmatic.on(OVERLAY_READY, toggleWalletModal)
```

src/components/WalletModal/index.tsx
```diff
@@ -126,101 +119,109 @@ export default function WalletModal({
...
-    connector &&
-      activate(connector, undefined, true).catch((error) => {
-        if (error instanceof UnsupportedChainIdError) {
-          activate(connector) // a little janky...can't use setError because the connector isn't set
-        } else {
-          setPendingError(true)
+        dispatch(updateSelectedWallet({ wallet }))
+        dispatch(updateWalletError({ wallet, error: undefined }))
+      } catch (error) {
+        if (
```
zzmp, line 182: Did we special-case this before? I'm not sure it's necessary considering that it adds code complexity. Up to you.

src/components/WalletModal/index.tsx
```diff
@@ -126,101 +119,109 @@ export default function WalletModal({
...
+        await connector.activate()
 
-    connector &&
-      activate(connector, undefined, true).catch((error) => {
-        if (error instanceof UnsupportedChainIdError) {
-          activate(connector) // a little janky...can't use setError because the connector isn't set
-        } else {
-          setPendingError(true)
+        dispatch(updateSelectedWallet({ wallet }))
+        dispatch(updateWalletError({ wallet, error: undefined }))
```
zzmp, line 180: The wallet error should be reset before attempting to activate, or old errors will persist in the UI during attempted activation. Move L180 to L176.

src/components/WalletModal/index.tsx
```diff
@@ -361,32 +311,30 @@ export default function WalletModal({
         <CloseIcon onClick={toggleWalletModal}>
           <CloseColor />
         </CloseIcon>
-        {walletView !== WALLET_VIEWS.ACCOUNT ? (
+        {connector === network ? (
```
zzmp, line 314: This should still be `walletView !== WALLET_VIEWS.ACCOUNT`, or else the "Connect a wallet" header will not be shown, and a left-arrow will be shown on the options menu that does nothing.

src/components/WalletModal/index.tsx
```diff
@@ -361,32 +311,30 @@ export default function WalletModal({
...
           <HeaderRow color="blue">
-            <HoverText onClick={resetAccountView}>
+            <HoverText onClick={() => setWalletView(WALLET_VIEWS.OPTIONS)}>
               <ArrowLeft />
             </HoverText>
           </HeaderRow>
         ) : (
           <HeaderRow>
-            <HoverText>
+            <HoverText onClick={() => setWalletView(WALLET_VIEWS.ACCOUNT)}>
```
zzmp, line 322: This should be reverted (see previous comment).

zzmp (changes requested):

src/connectors/index.ts
```diff
@@ -1,49 +1,147 @@
...
+      options: {
+        url: INFURA_NETWORK_URLS[SupportedChainId.MAINNET],
+        appName: 'Uniswap',
+        appLogoUrl: UNISWAP_LOGO_URL,
+      },
+      onError,
+    })
+)
+
+interface ConnectorListItem {
```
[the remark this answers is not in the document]
vm: tried removing this but then it'll expect only a gnosis safe item vs any generic connector

src/connectors/index.ts
```diff
@@ -1,49 +1,147 @@
...
+  connector: Connector
+  hooks: Web3ReactHooks
+}
+
+export function useConnectors(selectedWallet: Wallet | undefined) {
+  return useMemo(() => {
```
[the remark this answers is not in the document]
vm: i find this more confusing tbh

src/components/WalletModal/index.tsx
vm: this adds so much logic that i found confusing when i was learning the codebase that i think the delay is okay... especially with how few people even use fortmatic. longer term i think we should deprecate fortmatic anyways

zzmp (changes requested):
There's 2-3 regressions in the `WalletModal`.

Missing header in `WalletModal`:
- Open a new page (with no wallet connections)
- Click "Connect Wallet"
- The WalletModal header should read "Connect a wallet"

Non-functional back arrow in `WalletModal`:
- Connect a wallet
- Click the wallet
- Click "Change"
- Click the back arrow - it should take you back to your wallet view

Possible third regression (possible duplicate of #1):
- Connect a wallet
- Click the wallet
- Click "Disconnect"
- The WalletModal header should read "Connect a wallet"

zzmp (approved):

PR #3763, opened 2022-04-21
refactor: remove hideRouteDiagram prop

anasyusef:
`hideRouteDiagram` prop doesn't seem to be used at all in this component.


vm (approved):

vm:
thanks for the pr! i was going to make a similar pr and saw you already had one

Issue #3762, opened 2022-04-20
Crash report: `Error: No V2 factory address on this chain`


Elham3030:
# URL
  
https://app.uniswap.org/#/pool/v2?chain=mainnet

## Error

```
Error: No V2 factory address on this chain
```

## Stacktrace

```
Error: No V2 factory address on this chain
    at L (https://app.uniswap.org/static/js/main.e43a79ef.chunk.js:1:274135)
    at https://app.uniswap.org/static/js/main.e43a79ef.chunk.js:1:663495
    at Array.map (<anonymous>)
    at https://app.uniswap.org/static/js/main.e43a79ef.chunk.js:1:663445
    at Object.xa [as useMemo] (https://app.uniswap.org/static/js/3.f05ab84b.chunk.js:2:2098937)
    at https://app.uniswap.org/static/js/3.f05ab84b.chunk.js:2:2013478
    at Tj (https://app.uniswap.org/static/js/main.e43a79ef.chunk.js:1:663423)
    at oa (https://app.uniswap.org/static/js/3.f05ab84b.chunk.js:2:2094204)
    at za (https://app.uniswap.org/static/js/3.f05ab84b.chunk.js:2:2103771)
    at Yo (https://app.uniswap.org/static/js/3.f05ab84b.chunk.js:2:2147495)
```

## Device data

```json
{
  "ua": "Mozilla/5.0 (Linux; Android 5.1.1; SM-J111F) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/95.0.4638.74 Mobile Safari/537.36",
  "browser": {
    "name": "Chrome",
    "version": "95.0.4638.74",
    "major": "95"
  },
  "engine": {
    "name": "Blink",
    "version": "95.0.4638.74"
  },
  "os": {
    "name": "Android",
    "version": "5.1.1"
  },
  "device": {
    "vendor": "Samsung",
    "model": "SM-J111F",
    "type": "mobile"
  },
  "cpu": {}
}
```

moodysalem:
can repro this by just clicking polygon in the network dropdown from the pool page

behnammohammadkhani:
@moodysalem also, we have the same issue for other networks (arbitrum, optimism)

behnammohammadkhani:
@moodysalem 
I fixed the issues by this PR:
[PR #3777](https://github.com/Uniswap/interface/pull/3777)

ChenEranranS:
[...]

PR #3777, opened 2022-04-25
fix: No V2 factory address on this chain #3762

0xbeny:
the issue is fixed by adding Supported Chain Ids
Added :
-POLYGON
-ARBITRUM_ONE
-OPTIMISM

### DIFF
--- a/src/utils/constructSameAddressMap.ts
+++ b/src/utils/constructSameAddressMap.ts
@@ -6,6 +6,9 @@ const DEFAULT_NETWORKS = [
   SupportedChainId.RINKEBY,
   SupportedChainId.GOERLI,
   SupportedChainId.KOVAN,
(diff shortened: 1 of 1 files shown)


moodysalem:
there is no v2 on these networks, so the proper fix is to just handle it on the pages

behnammohammadkhani:
you're right, so what is the approach here. do you mean disabling some components or showing alert, when changing network. 
please mention what exactly you want, i'm ready to fix

moodysalem:
> you're right, so what is the approach here. do you mean disabling some components or showing alert, when changing network. please mention what exactly you want, i'm ready to fix

the proper fix would be, when connected to an unsupported network, show a message to that effect on the v2 pages

vm:
https://github.com/Uniswap/interface/pull/3913 resolved in this pr

PR #3787, opened 2022-04-29
feat: Ambire Wallet batch transaction skipping gas estimation

0xdex18:
This PR implements batching for ERC20 approvals with Ambire, which is a smart wallet that allows for transaction batching, smart recovery, gas abstractions and others. Rather than having to sign & send the ERC20 approval transaction separately, with this PR, it will be batched together with the specific action when the wallet used is Ambire.

The UX impact is immense, as the approval step is essentially permanently eliminated, without compromising security (the same calls are being done, just in one tx).

While this PR implements some logic custom to Ambire, this logic can be easily modified into an implementation of the batching EIP that’s currently being worked on by Gnosis Safe and Ambire, turning the code into a general implementation of txn batching.

This implementation uses altered SwapCall structure to skip the swap gas estimation, therefore no error is thrown. It can also be future proof, for smart wallets that calculate the gas independently from the dapp.

This PR is similar to the custom logic for Argent with wc_multiCall

### DIFF
--- a/src/hooks/useIsAmbireWC.ts
+++ b/src/hooks/useIsAmbireWC.ts
@@ -1,0 +1,8 @@
+import useActiveWeb3React from 'hooks/useActiveWeb3React'
+import { WalletConnectConnector } from 'web3-react-walletconnect-connector'
+
+export default function useIsAmbireWC(): boolean {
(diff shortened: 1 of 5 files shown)


mattbspector:

PR #3810, opened 2022-05-06
chore: rm widget code

zzmp:
Removes widget code. It has been moved to [Uniswap/widgets](https://github.com/Uniswap/widgets).

Removes code without altering or restructuring (except as necessary), to avoid adding complexity / introducing errors into this change. Code that was shared between widgets and interface still resides in `src/lib`, and can be re-integrated into `src` as future work.


src/lib/hooks/useTokenList/index.tsx
```diff
@@ -1,113 +1,3 @@
...
-  const { chainId } = useActiveWeb3React()
-  const chainTokenMap = useChainTokenMapContext()
-  const tokenMap = chainId && chainTokenMap?.[chainId]
-  return useMemo(() => {
-    if (!tokenMap) return []
-    return Object.values(tokenMap).map(({ token }) => token)
-  }, [tokenMap])
-}
 
 export type TokenMap = { [address: string]: Token }
```
vm, line 3: if this is the only thing in this file should we move it?

zzmp: Done.

There are a few other files that have been gutted in a similar way, but I'd prefer to refactor them in a future PR, so that this PR can be isolated to widget code deletion only.

vm: no prob, just pointing out the one i noticed. maybe good to list those other things so we have them somewhere, but nbd

vm (approved):
lgtm but would wait for jordan's review

JFrankfurt (commented):
iirc some of these `dependencies` might no longer be necessary or we may be able to move them back to `devDependencies`

JFrankfurt (approved):

PR #3820, opened 2022-05-09
feat: Adding Tally Ho wallet with name & logo for Tally Ho Users

DhruvJain1122:
This PR replaces PR: https://github.com/Uniswap/interface/pull/3548  that we've closed and. This new PR has a reduced scope. We hope this reduced scope could be an option to consider from your side.

The result of the new PR is that for users that have Tally Ho as their default wallet, the standard 'Injected' changes to 'Tally' 
A user that has MetaMask  installed will see (no changes)
image

A user that has no inject wallet installed will see (no changes)
image

A user that has both MetaMask and Tally Ho installed, and has MetaMask as their default wallet, will see (no changes)
image

A user that has another injected wallet installed, will see (no changes)
image

A user that has only Tally Ho installed, or has Tally Ho as it's default wallet will see this if this PR is merged
image

Instead of currently seeing:
image


vm:
I need to test this still, then I'll merge later this week

PR #3821, opened 2022-05-09
style: add learn more about wallets link

clayton1110:


vm:
I think we'll need to get translations for the copy, I'll find out how we do that

src/components/WalletModal/index.tsx
```diff
@@ -410,6 +410,19 @@ export default function WalletModal({
                 <ArrowRight size={16} />
               </RowBetween>
             </LinkCard>
+            <Row justify="center">
+              <ThemedText.Black fontSize={14}>
+                <Trans>
```
vm, line 415: You'll want to put Trans inside the Link children rather than outside

src/components/WalletModal/index.tsx
```diff
@@ -410,6 +410,19 @@ export default function WalletModal({
                 <ArrowRight size={16} />
               </RowBetween>
             </LinkCard>
+            <Row justify="center">
+              <ThemedText.Black fontSize={14}>
```
vm, line 414: This would also go inside the link children. Do we have a theme for links?

src/components/WalletModal/index.tsx
```diff
@@ -410,6 +410,19 @@ export default function WalletModal({
...
+                <Trans>
+                  <ExternalLink
+                    href="https://help.uniswap.org/en/articles/5391525-what-is-a-wallet"
+                    style={{ display: 'flex', alignItems: 'center' }}
```
vm, line 418: i think we try to avoid inline styles like this. Can the Row above will handle centering?

vm:
```tsx
            <ThemedText.Black fontSize={14}>
              <ExternalLink href="https://help.uniswap.org/en/articles/5391525-what-is-a-wallet">
                <Row justify="center" alignItems="center">
                  <Trans>Learn more about wallets</Trans>
                  <ArrowRight size={16} />
                </Row>
              </ExternalLink>
            </ThemedText.Black>```

Issue #3088, opened 2022-01-11
Network selector does not prompt to switch networks while connected to a wrong network


willhennessy:
**Bug Description**
A user reached out because they are confused how to proceed in this Wrong Network state. "Hi any idea what this means when I press conect wallet on uniswap:  Wrong Network Please connect to the appropriate Ethereum network. I cant even get to the part where you can scan the bar code"

I realized that while you are connected to the Wrong Network, clicking a network in the selector does NOT prompt to switch your network. It only changes the app UI, which actually confuses the user more because it appears to have switched to the right network but is still broken.

**Steps to Reproduce**

1. Connect to Binance chain https://chainlist.org/
2. Load app.uniswap.org
3. Open the dropdown selector

**Expected Behavior**
When you click on a network in the dropdown, the app should prompt the wallet to connect to the network.

**Additional Context**
* Metamask and Chrome
* maybe relevant:  I recently re-installed metamask so none of the networks are even added to my wallet yet.

willhennessy:
Also, the network selector should show an invalid network as being currently selected.  

Right now, you see Ethereum as the selected network next to a "wrong network" label. That's very confusing because Ethereum is a supported network. It doesn't communicate your current state (e.g. Binance chain)

Suggestion:  The network selector should display an error icon and "Unsupported network
![Uniswap-Interface](https://user-images.githubusercontent.com/5455577/148973531-211e7b6d-7508-495f-895d-cc424f9fa62f.png)
"

willhennessy:
Update, maggie designed the solution here: https://github.com/Uniswap/interface/issues/2627#issuecomment-999154401

JFrankfurt:
Adding to this

Some users are reporting that they are getting the "wrong network" error message while connected to a supported network. (Polygon)

JFrankfurt:
see also: #3092

willhennessy:
hi @leggechr, confirming this is a different issue than #2627.

2627 is mostly a cosmetic change for the Unsupported Network state

This issue is a functionality bug in the Unsupported Network state. While connected to an unsupported network, the in-app network selector fails to trigger a network switch prompt in metamask. Here's a [video](https://drive.google.com/file/d/1pYc-EYsFFcdnTF8ihauXu9NVBvibNwpA/view?usp=sharing) illustrating the intended behavior first (trigger a metamask popup) vs the unintended behavior while connected to an unsupported network.

[...]

PR #3826, opened 2022-05-09
fix: Unsupported network UI redesign -  #2627

Tott0:
Related to #2627 This pr moves the Network error prompt to the NetworkSelector component

image

---

on the other changes I noticed that the dev state is a little bit different from the app at `https://app.uniswap.org` Swap panel has a connect wallet button instead of the Sweap shown in the initial issue, leads me to believe someone already did a bit of work on this?

----

this process has a lot of small issues for starters when there's a chain error the `useWeb3` hook stops providing the `account` string directly, I did a small workaround for it through the connector but as shown it's not syncronous so it looks a bit weird, plus the missing account breaks other things, accountBalance counter in `Header`. and the StatusIcon is not displayed correctly even though `connector` is an instane of `InjectedConnector` (probably because its defined with some supported chainIds)

also related to #3088 the network selector does not do anything when trying to change while in a network error state, plus refreshing the website actually treats the app as if no wallet was connected, it will ask to connect your wallet again


Tott0:
Made a new push to tackle the part of hiding Polling info when the app is in a Chain error state

Is it enough to hide it visually? currently it should still do the queries on background, is just returning `null` for the component itself

Tott0:
after working with Vig on it for a while and pulling the v8 changes, it seems that expected changes are a bit outdated networkselector for example is no longer in the UI and there's another way to force a network change, prompt now asks to refresh
so not sure what's the expected process now

probably will have to wait for v8 changes to be finalized to focus on these

willhennessy:
let's see if Jira linking works:  WEB-100 and WEB-115

Issue #2520, opened 2021-10-08
Standardize USD decimals price format


willhennessy:

Tott0:
@willhennessy taking a look at this, is easy to change that specific USD text but changing every USD text in the app might be a bit difficult, I don't think there's an easy way to find out if we are working with USD text because the formatting is not done in an explicit manner but with some functions of `uniswap/sdk-core` a set of utilities for it. So probably it will be necessary to actually search for the USD texts that need change because some of them are fine, like in that photo the gas price to the right is actually working with 2 decimal places

Tott0:
@willhennessy about point 2, only if its between 0.95-1.05 ?   for example 1.95-2.05 (if possible) should be fine with 2 decimal places?

also need any rounding? right now is not rounding I see that the app can round like this `Rounding.ROUND_DOWN] /  [ROUND_HALF_UP] / [ROUND_UP]`

willhennessy:
thanks! I agree if the price is <$1.05 it's also useful to show 4 decimal places.  For number above that (even your example of 

Updated spec:
* Scope:  USD values in (1) the unit price (to the right of the info 'i' icon) (2) the input and amount USD value as you highlighted [above](https://github.com/Uniswap/interface/issues/2520#issuecomment-1128980066)
* All USD values in the app should use two decimal places.
* With one exception: if the USD price is < $1.05, then use 4 decimal places.
* No rounding

PR #3849, opened 2022-05-17
feat: changes usdcPrice formatting to fixed decimals

Tott0:
For usd pricing in swap modal card, uses toFixed formatting to force 2 decimal places (only exception is when price is between $ 0.95-1.05 in which case it uses 4 decimal places)

image

image


src/components/swap/TradePrice.tsx
```diff
@@ -33,6 +33,11 @@ export default function TradePrice({ price, showInverted, setShowInverted }: Tra
   const theme = useContext(ThemeContext)
 
   const usdcPrice = useUSDCPrice(showInverted ? price.baseCurrency : price.quoteCurrency)
+  /*
+   * calculate needed amount of decimal prices, for prices between 0.95-1.05 use 4 decimal places
+   */
+  const p = Number(usdcPrice?.toFixed(3))
```
leggechr, line 39: Nit: can you just use `toFixed()` here with the default value?

Also if the value is < 0.95 wouldn't we also want to show 4?

Tott0: ah good catch, wasn't sure if toFixed could be used without parameters, will do thanks!

according to the issue
`With one exception: if the USD price is between $0.95-1.05, then use 4 decimal places.`

https://github.com/Uniswap/interface/issues/2520

leggechr: Okay confirmed with Will that you can use 4 decimals for anything < 0.95.

leggechr: Oh sorry I got it backwards. I mean < 1.05

Tott0:
updating PR changing another USD on the fiatValue text with the same changes as the exchange price below

image

leggechr (approved):

src/components/swap/TradePrice.tsx
Tott0: Changed!

PR #3851, opened 2022-05-17
feat: scroll to top when navigating

clayton1110:

- feat: A new feature
- fix: A bug fix

Example commit messages:

- feat: adds support for gnosis safe wallet
- fix: removes a polling memory leak
- chore: bumps redux version

Other things to note:

### DIFF
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -1,4 +1,5 @@
 import Loader from 'components/Loader'
+import ScrollToTop from 'components/ScrollToTop'
 import TopLevelModals from 'components/TopLevelModals'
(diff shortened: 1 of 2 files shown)


src/pages/App.tsx
```diff
@@ -78,6 +79,7 @@ export default function App() {
             <Popups />
             <Polling />
             <TopLevelModals />
+            <ScrollToTop />
```
zzmp, line 82: This isn't presentational, so it should be implemented as a hook, and not a component. It's also top-level, so it doesn't need to be in it's own file.

Can you implement it inline in the App component as a hook?

zzmp (changes requested):

vm:
This was implemented in a hook in App.tsx in a separate PR 👍

PR #3857, opened 2022-05-20
fix: handle non existing inputCurrency from url #3056

### LINKED ISSUE #3056: Clear currency input component when invalid currency address is in url parameters
**Bug Description**
When users add an invalid token address to the url params, the "select a token" option disappears. This also happens when you import a token that exists on one chain (e.g. $BSTICK on Polygon), but does not appear on another, and you switch your network in the app later.

image

**Steps to Reproduce**

Either:
1. Set your wallet network to ETH
2. Go to https://app.uniswap.org/#/swap?inputCurrency=0x8e915a77e301ad12ad1f62012734cf7557eee81a

- or -
1. Set your wallet network to Polygon (correct network for $BSTICK)
2. Go to https://app.uniswap.org/#/swap?inputCurrency=0x8e915a77e301ad12ad1f62012734cf7557eee81a
3. Switch network to Ethereum

**Expected Behavior**
The token selector should be in its default state when the token address is invalid

**Additional Context**
Add any other context about the problem here (screenshots, whether the bug only occurs only in certain mobile/desktop/browser environments, etc.)

skv-headless:
closes #3056
opening an url swap?inputCurrency=not_existing_address
cause a state where you are not able to change input currency

same technique used here https://github.com/Uniswap/interface/blob/main/src/pages/AddLiquidity/index.tsx#L586

how it looks now:
Screenshot 2022-05-20 at 22 14 45

### DIFF
--- a/src/pages/Swap/index.tsx
+++ b/src/pages/Swap/index.tsx
@@ -426,7 +426,7 @@ export default function Swap({ history }: RouteComponentProps) {
                 }
                 value={formattedAmounts[Field.INPUT]}
                 showMaxButton={showMaxButton}
-                currency={currencies[Field.INPUT]}
+                currency={currencies[Field.INPUT] ?? null}
                 onUserInput={handleTypeInput}
                 onMax={handleMaxInput}
                 fiatValue={fiatValueInput ?? undefined}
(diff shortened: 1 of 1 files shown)


JFrankfurt:
@tinaszheng I'm having trouble replicating the original bug on main. Can you take a look at this?

JFrankfurt (approved):

tinaszheng:
@JFrankfurt i can still repro the bug on main using the second set of steps in the original issue

@skv-headless this looks good, thank you!

tinaszheng (approved):

PR #3867, opened 2022-05-26
fix: center NetworkSelector circles

vm:
before:
image

after:
image


src/components/Header/NetworkSelector.tsx
```diff
@@ -93,6 +93,13 @@ const FlyoutRowActiveIndicator = styled.div`
   height: 9px;
   width: 9px;
 `
+
+const CircleContainer = styled.div`
+  width: 20px;
+  display: flex;
```
lynnshaoyu, line 99: nit: can alphabetize css. although this is not enforced in other areas of the codebase so it doesn't matter as much. but would be helpful if new css started being alphabetical :D

PR #3876, opened 2022-06-01
Update graph link

Rachel-Eichenberger:
Found this to be the proper Optimism graph link. info site uses this one, and I have tested in local environment, and it showed the graph while adding LQ
image


ianlapham (approved):
Looks good

Issue #1828, opened 2021-06-08
on/off labels do not work well in non-english languages


moodysalem:
**Bug Description**
<img width="196" alt="Screen Shot 2021-06-08 at 8 57 08 AM" src="https://user-images.githubusercontent.com/7897876/121198237-98f6ce00-c837-11eb-8735-ddab50714803.png">
<img width="158" alt="Screen Shot 2021-06-08 at 8 57 13 AM" src="https://user-images.githubusercontent.com/7897876/121198260-9d22eb80-c837-11eb-88ce-dfb5cf65e3ab.png">

In languages with longer text for on/off, the toggles look very different. Replace with something more responsive maybe

Assigning to Maggie in case this is already being worked on

willhennessy:
hi @maggievalentine what would you recommend?

maggievalentine:
agh... we might have to shift to an on/off toggle that uses color / iconography rather than text. can design some options

infredible:
# Design guidance for this bug

## Overview
These changes should effect:
- transaction settings
- the token list
- position detail page
- positions page


### Remove text/copy from toggle
<img width="642" alt="CleanShot 2022-05-19 at 13 41 10@2x" src="https://user-images.githubusercontent.com/3887562/169400918-9d6437df-94f6-49f8-bc1f-e98b2e64b544.png">

### Make switch knob appear on right for "on" consistently throughout app
Currently, toggle switches appear inconsistently with the knob on either side to indicate "on" (current implementation)
<img width="260" alt="CleanShot 2022-05-19 at 13 42 04@2x" src="https://user-images.githubusercontent.com/3887562/169401057-c86ec7bf-10ab-4316-9be8-ad78bd7e83d0.png">
<img width="327" alt="CleanShot 2022-05-19 at 13 43 21@2x" src="https://user-images.githubusercontent.com/3887562/169401269-8fcc08f4-906d-4397-ad4e-6ac907a55a92.png">

The knob should be on the _right_ side for "on":
<img width="354" alt="CleanShot 2022-05-19 at 13 46 21@2x" src="https://user-images.githubusercontent.com/3887562/169401729-b8ad6cb2-cad7-4ff3-9e0f-d2c452a25163.png">
<img width="356" alt="CleanShot 2022-05-19 at 13 46 37@2x" src="https://user-images.githubusercontent.com/3887562/169401767-9ef512a3-ac33-476b-84e4-c3df039064a8.png">
[...]

PR #3858, opened 2022-05-22
fix: Mobile layout: hide closed positions disappears #3344

ghgoodreau:
This change adds the 'Hide closed positions' button to the mobile Pool page view. As referenced in #3344 , without this, the pool page can accumulate closed positions and the request can even time out with a large number of positions. We shouldn't limit user functionality due to mobile view whenever possible.

By default, the button is identical to the font size of the 'Your positions'. This supports widths up to 340px.
image

The smallest mobile view that I typically support is 320px (iPhone 5). With this, the bar overflows and doesn't look great.
image

With a font size of 14px on 'Hide closed positions', we are able to fit the 'Your positions' + 'Hide closed positions' in one line on screen widths as small as 320px.
image

PR #3884, opened 2022-06-03
fix: use simple toggle instead of toggle with text init commit

lynnshaoyu:
Fix bug: https://github.com/Uniswap/interface/issues/1828

Simplify toggles and remove on/off text that doesn't look good when translated to other languages (due to differing text lengths). 

I also refactored and simplified the toggle code. Deleted ListToggle and SimpleToggle (from https://github.com/Uniswap/interface/pull/3858) to unify into one configurable toggle for all our needs. Simplified and removed unnecessary "isToggleOn" state in original implementation as well.


lynnshaoyu:
Video image of toggles post fix:
https://user-images.githubusercontent.com/41491154/171948305-21c79261-1a27-4652-816f-93aca43cd268.mp4

Screenshots: 
Screen Shot 2022-06-03 at 4 35 08 PM
Screen Shot 2022-06-03 at 4 34 38 PM
Screen Shot 2022-06-03 at 4 34 32 PM

ianlapham (approved):
Looks good!

infredible:
Looking good! One nitpick: could we add a transition animation? The current behavior looks a bit jarring without it. 

An ease-out, 100ms transition would be great.

https://user-images.githubusercontent.com/3887562/171957409-cca3fc59-c9e6-4ea2-a484-9be6aedf0d86.mp4

lynnshaoyu:
Added 0.1 second ease-in animations!

https://user-images.githubusercontent.com/41491154/172074251-8284bcd3-db0b-4201-82db-ca30115817ce.mp4

willhennessy:
thanks lynn this looks great!  one reallly tiny nit:  when the page first loads, the toggle slides over to the right position. This feels like a form of page thrash.  Is it possible to set an initial value for the toggle?

oddly sometimes the value is initial to off (show closed positions on the pool page), and sometimes to on (collect as WETH on the position page)

PR #3886, opened 2022-06-06
fix: remove animation from toggles on initial page load

lynnshaoyu:
Respond to @willhennessy 's comment on https://github.com/Uniswap/interface/pull/3884 and remove animation from toggle on initial page load to reduce thrash.


willhennessy (approved):
lgtm

Issue #3865, opened 2022-05-26
Keyboard navigation in token selector 


lynnshaoyu:
Token selector can only be accessed using a mouse currently. 

Should improve accessibility by enabling tabbing through options and hitting enter on keyboard to select.

PR #3887, opened 2022-06-06
fix: add keyboard accessibility to token selector

lynnshaoyu:
Make token selector accessible via keyboard navigation to resolve https://github.com/Uniswap/interface/issues/3865

https://user-images.githubusercontent.com/41491154/172245742-3685a03b-8a25-487b-afeb-9e6a1c310596.mp4

### DIFF
--- a/src/components/SearchModal/CommonBases.tsx
+++ b/src/components/SearchModal/CommonBases.tsx
@@ -50,6 +50,7 @@ export default function CommonBases({
           return (
             <BaseWrapper
               tabIndex={0}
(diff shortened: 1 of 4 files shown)


callil:
Looking great, only thing I noticed was that hitting the enter key on the suggested pairs doesn't "submit" the selection.

lynnshaoyu:
Ah missed that! Thanks for catching. Fixed in the latest commit.

willhennessy:
this is so cool! looks great lynn

PR #3892, opened 2022-06-07
feat: abstract analytics logging

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-149

Move google analytics functions to single class to prepare for migration to new analytics framework.


zzmp (changes requested):
If the goal is to abstract our analytics, export it under an alias from `/analytics`, so that instead of use cases importing `GoogleAnalyticsProvider from 'analytics/GoogleAnalyticsProvider`, they can import a function from `/analytics`, eg `import { sendAnalyticsEvent } from `/analytics`.

This way, if the provider/reporter are swapped out, the only thing that will change is the implementation in `/analytics`, and none of the use cases will need to change.

JFrankfurt:
I agree with @zzmp 's suggestion.

fwiw (no need to do this here if you don't want to) I would be supportive of updating the api to not directly pass through GA's arguments. The `action`, `label`, `category` names have never really made much sense to me.

zzmp:
It's fine (and probably good) to keep the `GoogleAnalyticsProvider` class, because it is specific to GA, but that should be an implementation detail, ie it should not ever be used outside of the analytics folder.
I think the easiest way to do this is to have `src/analytics` import the `GoogleAnalyticsProvider`, instantiate it, and then expose free functions that are exported, so that in your use cases you don't need to worry about class instantiation.

I also agree with Jordan that the details specific to GA should be abstracted away (ideally by your free methods).

lynnshaoyu:
Hey @zzmp , followed your suggestions as discussed here and on Slack in the latest commit.

This PR covers 1 and 2 of the below, will sort out 3 in a separate PR (@JFrankfurt 's suggestions):
1. Put analytics in /src/analytics
2. Abstract out GoogleAnalytics to free methods exported from /src/analytics
3. Type the free methods more strictly, using typed analytics labels

src/components/analytics/index.ts
```diff
@@ -1,20 +1,42 @@
...
+import GoogleAnalyticsProvider from './GoogleAnalyticsProvider'
 
 export const GOOGLE_ANALYTICS_CLIENT_ID_STORAGE_KEY = 'ga_client_id'
 const GOOGLE_ANALYTICS_ID: string | undefined = process.env.REACT_APP_GOOGLE_ANALYTICS_ID
 
 const storedClientId = window.localStorage.getItem(GOOGLE_ANALYTICS_CLIENT_ID_STORAGE_KEY)
 
+const googleAnalytics = new GoogleAnalyticsProvider()
+
+const sendEvent = (event: string | UaEventOptions, params?: any) => googleAnalytics.sendEvent(event, params)
```
zzmp, line 17: export this inline, and as a defined function (not anonymous):

```
export function sendEvent(...) {
  return googleAnalytics.sendEvent(event, params)
}
```

Same with other wrapped functions.

src/components/analytics/index.ts
```diff
@@ -23,5 +45,44 @@ if (typeof GOOGLE_ANALYTICS_ID === 'string') {
...
-  ReactGA.initialize('test', { gtagOptions: { debug_mode: true } })
+  googleAnalytics.initialize('test', { gtagOptions: { debug_mode: true } })
+}
+
+function reportWebVitals({ name, delta, id }: Metric) {
+  gaCommandSendTiming('Web Vitals', name, Math.round(name === 'CLS' ? delta * 1000 : delta), id)
 }
+
+// tracks web vitals and pageviews
+export function GoogleAnalyticsReporter({ location: { pathname, search } }: RouteComponentProps): null {
```
zzmp, line 56: This should not be exported. The Google-specific stuff stays in `/analytics`.

src/components/analytics/index.ts
```diff
@@ -1,20 +1,42 @@
...
+const outboundLink = (
+  {
+    label,
+  }: {
+    label: string
+  },
+  hitCallback: () => unknown
+) => googleAnalytics.outboundLink({ label }, hitCallback)
+
+const gaCommandSendTiming = (timingCategory: any, timingVar: any, timingValue: any, timingLabel: any) =>
```
zzmp, line 28: This should not be google-specific. Rename to `sendTiming`.

src/components/analytics/index.ts
zzmp: If you need initialization, you could instead export a general `Reporter` that wraps a google analytics reporter, which seems to be what this is. It may be enough to just rename this to something generic.

zzmp, line 56: This is not presentational, it just runs hooks. Instead of adding it as a component, make it a hook that is run on the root component.

zzmp (changes requested):

src/components/analytics/index.ts
lynnshaoyu: done

src/components/analytics/index.ts
lynnshaoyu: sgtm done

lynnshaoyu: Sounds good done.

zzmp (approved):

Issue #1728, opened 2021-05-28
scroll to top when navigating / remember scroll when going back


moodysalem:
we need to fix scroll behavior when navigating between pages

moodysalem:
chrome actually does this automatically, and some testing showed it works pretty well
https://developers.google.com/web/updates/2015/09/history-api-scroll-restoration

ios safari, not so good

willhennessy:
Steps to reproduce.

Browser:  Safari iOS mobile browser

### Bug:  scroll to top
1. visit https://app.uniswap.org/#/vote?chain=mainnet
2. scroll down to an old proposal such as proposal 1.1
3. Click on proposal 1.1

Intended behavior:  navigate to the proposal 1.1 page, with the scroll position at the top of the page so the user can read the proposal from the start.
Actual behavior:  navigates to the proposal 1.1 page, but the scroll position is further down in the middle of the page

### Remember scroll position when going back.
I cannot reproduce this.

lynnshaoyu:
Hey @moodysalem, are we trying to scroll to top on every page navigation in the app, or just for the voting proposal page Will mentioned? 

Also, I believe the app is already remembering original scroll position when going back (click back button on browser) for both chrome and safari.

PR #3893, opened 2022-06-07
fix: fix scroll on page navigation

lynnshaoyu:
fixes https://github.com/Uniswap/interface/issues/1728

### DIFF
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -25,6 +25,7 @@ import PoolV2 from './Pool/v2'
 import PoolFinder from './PoolFinder'
 import RemoveLiquidity from './RemoveLiquidity'
 import RemoveLiquidityV3 from './RemoveLiquidity/V3'
+import ScrollToTop from './ScrollToTop'
 import Swap from './Swap'
 import { OpenClaimAddressModalAndRedirectToSwap, RedirectPathToSwapOnly, RedirectToSwap } from './Swap/redirects'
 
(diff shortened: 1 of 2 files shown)


src/pages/ScrollToTop.js
```diff
@@ -0,0 +1,19 @@
+import { useEffect } from 'react'
+import { withRouter } from 'react-router-dom'
+
+// eslint-disable-next-line react/prop-types
+function ScrollToTop({ history }) {
```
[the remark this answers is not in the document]
lynnshaoyu: whoops sorry this was a WIP not ready for review yet - intended to refactor it into App.tsx. My latest commit is in line w these comments 👍

lynnshaoyu:
After fix, when user clicks on a proposal page opens with scroll set to top of voting proposal: 

https://user-images.githubusercontent.com/41491154/172495317-57d5651e-c749-408c-a4a4-f4eba66d8220.mp4

Verified works in Safari too.

src/pages/App.tsx
```diff
@@ -64,6 +66,21 @@ const Marginer = styled.div`
 `
 
 export default function App() {
+  const history = useHistory()
+
+  const ScrollToTop = () => {
```
zzmp, line 71: Does this need to be wrapped in a functional component? Because this isn't presentational, I think it would be best implemented as a hook (ie remove lines 71, 80-82, 99).

zzmp (changes requested):

src/pages/App.tsx
lynnshaoyu: Ah you're right it's unnecessary. Removed in the latest commit.

zzmp (approved):

PR #3896, opened 2022-06-08
fix: add jest coverage config to package.json

lynnshaoyu:
Configure package.json to show test coverage. Target coverage thresholds set to 70% currently. 

To run tests with coverage info: 
yarn test --coverage --watchAll

Run only tests affected by your changes:
yarn test --coverage --watch

Also added testing instructions to CONTRIBUTING.md in this PR.


CONTRIBUTING.md
```diff
@@ -15,10 +14,17 @@ Thank you for your interest in contributing to the Uniswap interface! 🦄
 1. `yarn install`
 1. `yarn build`
 
+## Running tests
+
+1. Run unit tests: `yarn test`
```
zzmp, line 19: These are all distinct, so they should be bullets, not numbered (`-`).

CONTRIBUTING.md
```diff
@@ -15,10 +14,17 @@ Thank you for your interest in contributing to the Uniswap interface! 🦄
...
+1. Run unit tests and show test coverage info for all tests: `yarn test --coverage --watchAll`
```
zzmp, line 20: I don't think this is necessary - it's more for a maintainer than a contributor.

CONTRIBUTING.md
```diff
@@ -15,10 +14,17 @@ Thank you for your interest in contributing to the Uniswap interface! 🦄
...
+1. Run unit tests and show test coverage info only for tests affected by your changes: `yarn test --coverage --watch`
```
zzmp, line 21: This should be the standard command. Is there any reason to not include `--coverage` in `yarn test`?
Also, `--watch` is done by default through jest, and does not need to be included.

CONTRIBUTING.md
```diff
@@ -15,10 +14,17 @@ Thank you for your interest in contributing to the Uniswap interface! 🦄
...
+1. Run e2e cypress tests with visual interface: `yarn cypress open`
```
zzmp, line 22: In order to run cypress tests, you'll need to a build to test them off of.
You'll need to `yarn build:e2e` in order to get a build with infura keys that will work on your localhost (vs `yarn build`, which is for production), and then `yarn test:e2e` will open cypress (it aliases `cypress open`) with a server backing the tests. CI uses `yarn test:e2e:ci` to run without the cypress UI.

package.json
```diff
@@ -156,6 +156,27 @@
...
+      "src/lib/hooks/**/*.ts*", 
+      "src/lib/state/**/*.ts*", 
+      "src/lib/utils/**/*.ts*", 
+      "src/pages/**/*.ts*",
+      "src/state/**/*.ts*",
+      "src/utils/**/*.ts*"
+    ],
+    "coverageThreshold": {
+      "global": {
+        "branches": 70,
```
zzmp, line 173: These may be too high, considering where we are currently at.

zzmp (changes requested):

CONTRIBUTING.md
lynnshaoyu: Done

package.json
lynnshaoyu: True, changed to 25

CONTRIBUTING.md
lynnshaoyu: Thanks! I added these instructions for running integration tests

CONTRIBUTING.md
lynnshaoyu: I was thinking maybe adding these would encourage contributors to pay attention to / add their own unit tests when submitting PRs. The commands are pretty standard but for some newer contributors they might not be familiar with reading the scripts section of package.json and the coverage watchAll feature of jest.

package.json
lynnshaoyu: I lowered them to just under where we are so our unit test check in the git automated checks don't block submissions.

package.json
```diff
@@ -16,11 +16,32 @@
...
+      "src/hooks/**/*.ts*", 
+      "src/lib/hooks/**/*.ts*", 
+      "src/lib/state/**/*.ts*", 
+      "src/lib/utils/**/*.ts*", 
+      "src/pages/**/*.ts*",
+      "src/state/**/*.ts*",
+      "src/utils/**/*.ts*"
+    ],
+    "coverageThreshold": {
+      "global": {
```
zzmp, line 37: oof

zzmp (approved):

PR #3897, opened 2022-06-08
feat: service worker with etag cache

zzmp:
- Updates the service worker to always serve the freshest document, using etag comparison
- Leverages existing workbox defaults to minimize custom codepaths
- Unit tests document matcher and handler
- Integration tests the `__isDocumentCached`, which is injected into the served document if it is served from the cache


src/serviceWorker/document.ts
```diff
@@ -0,0 +1,124 @@
...
+    navigationPreload.enable()
+    super(matchDocument, handleDocument.bind({ offlineDocument }), 'GET')
+  }
+}
+
+/**
+ * A cache-specific version of the document.
+ * This document sets the local `__isDocumentCached` variable to true.
+ */
+// TODO(xtine): Send a GA beacon from the client to record cache usage / metrics.
```
leggechr, line 100: Remove this

leggechr (approved):

leggechr (approved):

PR #3899, opened 2022-06-09
feat: locally store fiatValueOfFees

guil-lambert:
- feat: store the computed fiatValueOfFees using localStorage

Notes:

- Sometimes the static calls fail when fetching the uncollected fees for a position. 
- When it fails, the "Unclaimed fees" value on PositionPage shows `-`, which is annoying
- This feature stores the `fiatValueOfFees` locally using `localStorage` every time `fiatValueOfFees` is successfully computed
- This is better for LPs: rather than seeing `-` and having to reload the page until an actual value shows up, unclaimed fees now shows the latest value that was computed..


vm:
Think better to avoid this as the value could be stale / misleading, but lmk if you feel otherwise.

PR #3900, opened 2022-06-09
feat: adding Rabby wallet logo on connection options

vvvvvv1vvvvvv:
This PR add Rabby wallet logo when user installed Rabby and nothing changed when user not installed Rabby

When Rabby installed:
image

Rabby not installed:
image

Both Rabby and MetaMask not installed:
image


vm:
We're trying to avoid custom logic around injected wallets for now. We may have a better solution around this in the future with a more generic web3-react connector for non-MetaMask injected wallets, and have a way to specify logos/names. But unclear if/when that will happen.

PR #3908, opened 2022-06-13
test: update cypress

zzmp:
- Updates to cypress v10
- Disables `chromeWebSecurity` to ensure working tests despite reloads
  - Without disabling `chromeWebSecurity`, tests fail for either unknown reasons, or reasons related to service worker / cypress integration. Leaving it disabled will also help to test the service worker (follow-up PR)
  - This requires an additional workaround to retain the `origin` header for infura calls
- Updates `CONTRIBUTING.md` to use `cypress` explicitly


cypress.config.ts
```diff
@@ -0,0 +1,15 @@
+import { defineConfig } from 'cypress'
+
+export default defineConfig({
+  projectId: 'yp82ef',
+  video: false,
+  defaultCommandTimeout: 10000,
+  chromeWebSecurity: false,
+  e2e: {
```
vm, line 8: what does this section do?

vm (approved):

cypress.config.ts
zzmp: `setupNodeEvents` was added by the cypress migration tool. It replaces plugins, but it's not used. I'll see if I can remove it.

Issue #1712, opened 2021-05-26
Show spot price on token select


callil:
We should consider fetching and displaying the spot price of the pair as soon as a token is selected rather than relying on the use to input a value. This will be useful for people who are checking the current market price of a token pair and reduce clicks/actions.

justindomingue:
Would need to make currencyAmount?.currency non-null to get a price.

https://github.com/Uniswap/uniswap-interface/blob/main/src/hooks/useUSDCPrice.ts#L59

moodysalem:
> Would need to make currencyAmount?.currency non-null to get a price.
> 
> https://github.com/Uniswap/uniswap-interface/blob/main/src/hooks/useUSDCPrice.ts#L59

You can just use the other hook, `useUSDCPrice` (ignoring performance of fetching this twice in one component)

MStreet3:
Is there a design of what this should look like?  not clear from the issue where the spot price should appear.

justindomingue:
@MStreet3 see the screenshots below from app.uniswap.org. The spot price for ETH only appears after I input 1.0. The spot price is independent of the input amount so we could always show it when a token is selected.

### No input amount
(change requested is to make it show the spot price)
![image](https://user-images.githubusercontent.com/1284993/120536181-b4298f80-c398-11eb-942e-f7fcf2b5c452.png)

### With input amount
![image](https://user-images.githubusercontent.com/1284993/120536160-ae33ae80-c398-11eb-9c3f-1f559b085799.png)

MStreet3:
Hi, I'm trying to resolve this issue via this PR https://github.com/Uniswap/uniswap-interface/pull/1816

prakashsanker:
@justindomingue  is this issue still open? Can I take a stab at it?

ghgoodreau:
Took a stab at this since it was dead for a bit. Please take a look! #3909

PR #3909, opened 2022-06-13
fix: shows spot price on token select

ghgoodreau:
https://github.com/Uniswap/interface/issues/1712

Adds spot price on token select when a user hasn't entered an amount yet. I wasn't sure how to display this, so I went with ~${spotPrice}. I can change this as needed as I expect some input on how we want to display this. 

Some examples: 
image
image
image


ghgoodreau:
Bumping this for visibility! Would love to have someone take a look next week.

just-toby:
sorry nobody got around to reviewing this last year - closing now as this is out of date

PR #3912, opened 2022-06-14
fix(L2): removes network-specific polling

JFrankfurt:


src/utils/getLibrary.ts
```diff
@@ -18,12 +10,5 @@ export default function getLibrary(provider: any): Web3Provider {
       : 'any'
   )
   library.pollingInterval = 15_000
```
zzmp, line 12: Use the ms macro to mark this as 15s more human-readably.

zzmp (approved):

zzmp (approved):

PR #3922, opened 2022-06-16
build: optimize github actions

zzmp:
- Parallelizes cypress CI runs
- Cleans up CI workflow files
- Aggressively cache node_modules to improve _all_ build times

### DIFF
--- a/.github/workflows/crowdin.yaml
+++ b/.github/workflows/crowdin.yaml
@@ -11,23 +11,19 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v3
 
-      - name: Set up node
-        uses: actions/setup-node@v3
+      - uses: actions/setup-node@v3
         with:
           node-version: 14
           registry-url: https://registry.npmjs.org
           cache: 'yarn'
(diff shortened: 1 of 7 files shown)


.github/workflows/crowdin.yaml
```diff
@@ -11,23 +11,26 @@ jobs:
...
         with:
           node-version: 14
           registry-url: https://registry.npmjs.org
           cache: 'yarn'
 
-      - name: Install dependencies
-        run: yarn install --frozen-lockfile
+      - uses: actions/cache@v3
+        id: install-cache
+        with:
```
vm, line 24: i dont really follow what's going on here mostly

vm (approved):

.github/workflows/crowdin.yaml
zzmp: This explicitly puts node_modules in a cache, instead of the global yarn cache. SO that if this yarn.lock has already been installed, we can just copy node_modules over from cache. This works because we run our CI on identical machines, with identical versions of node, so that we can just copy over node_modules across builds. Effectively, this saves about 30s per CI action, because it lets us skip installing node_modules.

vm: smart

Issue #3747, opened 2022-04-18
The crowdin jobs(uploading, downloading translations) are failing.


octocat512:
**Bug Description**
The crowdin jobs(uploading, downloading translations) are failing to run. 

**Steps to Reproduce**

1. Go to the `Actions` tab on `Github`
2. You can find failed jobs
   ...

**Expected Behavior**
jobs need to be executed successfully.

**Additional Context**
Bumping the version solves the problem.

PR #3927, opened 2022-06-16
fix: Crowdin translations download

andrii-bodnar:
Hey everyone,

Since the `main` branch is protected, the download translations workflow is failing. Just updated the workflow to push translations into the separate branch and create a PR with the translations instead of direct pushing them into the `main`.

#3747


zzmp:
Do you know if we can pass in a token to give the workflow permission to push directly, instead of having to go through a PR process?

andrii-bodnar:
@zzmp just checked the branch protection rules configuration and it seems like GitHub doesn't allow such exclusions.

andrii-bodnar:
It's better to go through a PR process because you can review the translations before merging and avoid some possible issues.

PR #3954, opened 2022-06-22
Explore use top tokens

JFrankfurt:


src/hooks/useTopTokens.ts
```diff
@@ -0,0 +1,158 @@
...
+    setError(null)
+    fetchTopTokens()
+      .then((data) => {
+        if (data) setData(data)
+      })
+      .catch((e) => setError(e))
+      .finally(() => setLoading(false))
+  }, [])
+
+  return { data, error, loading }
```
vm, line 155: is there a way to refetch the data exposed?

vm (approved):

kayleegeorge (approved):

src/hooks/useTopTokens.ts
vm: i think you could just have a `refetch` function passed as part of the hook

justindomingue: `pollingInterval`

PR #3956, opened 2022-06-22
Kg/explore token list UI

kayleegeorge:
Add Time

### DIFF
--- a/src/components/Explore/TimeSelector.tsx
+++ b/src/components/Explore/TimeSelector.tsx
@@ -1,0 +1,177 @@
+import { useOnClickOutside } from 'hooks/useOnClickOutside'
+import { useRef } from 'react'
+import React, { useState } from 'react'
+import { Check, ChevronDown, ChevronUp } from 'react-feather'
+import { useModalOpen, useToggleModal } from 'state/application/hooks'
+import { ApplicationModal } from 'state/application/reducer'
+import styled, { css } from 'styled-components/macro'
+
+const TIME_OPTIONS: { [key: string]: string } = {
+  hour: '1H',
+  day: '1D',
+  week: '1W',
+  month: '1M',
+  year: '1Y',
+}
+
+export enum FlyoutAlignment {
+  LEFT = 'LEFT',
+  RIGHT = 'RIGHT',
+}
+
+const InternalMenuItem = styled.div`
+  flex: 1;
+  padding: 0.5rem 0.5rem;
+  color: ${({ theme }) => theme.text2};
+  :hover {
+    color: ${({ theme }) => theme.text1};
+    cursor: pointer;
+    text-decoration: none;
+  }
+  > svg {
+    margin-right: 8px;
+  }
+`
+
+const InternalLinkMenuItem = styled(InternalMenuItem)`
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  padding: 0.5rem 0.5rem;
+  justify-content: space-between;
+  text-decoration: none;
+  :hover {
+    color: ${({ theme }) => theme.text1};
+    cursor: pointer;
+    text-decoration: none;
+  }
+`
+const MenuTimeFlyout = styled.span<{ flyoutAlignment?: FlyoutAlignment }>`
+  min-width: 150px;
+  max-height: 350px;
+  overflow: auto;
+  background-color: ${({ theme }) => theme.bg1};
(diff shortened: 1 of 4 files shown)


src/components/Explore/TimeSelector.tsx
```diff
@@ -0,0 +1,179 @@
...
+  year: '1Y',
+}
+
+const enum TimeOption {
+  Hour = 'hour',
+  Day = 'day',
+  Week = 'week',
+  Month = 'month',
+  Year = 'year',
+}
```
JFrankfurt, line 23: prob best to just use one of these (pref TimeOptions enum, or exactly the one used in the mock data hooks)

src/components/Explore/TimeSelector.tsx
```diff
@@ -0,0 +1,179 @@
...
+const TIMES = [TimeOption.Hour, TimeOption.Day, TimeOption.Week, TimeOption.Month, TimeOption.Year]
+
+export enum FlyoutAlignment {
```
JFrankfurt, line 26: doesn't need to be exported

JFrankfurt: generally better to only export once something is used, that way you know in future refactors if you remove all usage in a file, the non-exported type can be removed.

src/components/Explore/TimeSelector.tsx
```diff
@@ -0,0 +1,179 @@
...
+  LEFT = 'LEFT',
+  RIGHT = 'RIGHT',
+}
+
+const InternalMenuItem = styled.div`
+  flex: 1;
+  padding: 0.5rem 0.5rem;
```
JFrankfurt, line 33: We generally try to avoid `rem`, since it can have unexpected cascading size effects. Prefer `em` (or px values that are divisible by 4). I tend to do the % 4 px value closest to what's specified in Figma, and let the designer know if it's not perfectly divisible.

src/components/Explore/TimeSelector.tsx
```diff
@@ -0,0 +1,179 @@
...
+const MenuTimeFlyout = styled.span<{ flyoutAlignment?: FlyoutAlignment }>`
+  min-width: 150px;
+  max-height: 350px;
+  overflow: auto;
+  background-color: ${({ theme }) => theme.bg1};
+  box-shadow: 0px 0px 1px rgba(0, 0, 0, 0.01), 0px 4px 8px rgba(0, 0, 0, 0.04), 0px 16px 24px rgba(0, 0, 0, 0.04),
+    0px 24px 32px rgba(0, 0, 0, 0.01);
+  border: 1px solid ${({ theme }) => theme.bg0};
+  border-radius: 12px;
+  padding: 0.5rem;
```
JFrankfurt, line 67: see rem comment above

src/components/Explore/TimeSelector.tsx
```diff
@@ -0,0 +1,179 @@
...
+// TODO: change this to reflect data pipeline
+export default function TimeSelector() {
+  const node = useRef<HTMLDivElement>()
+  const open = useModalOpen(ApplicationModal.TIME_SELECTOR)
+  const toggleMenu = useToggleModal(ApplicationModal.TIME_SELECTOR)
+  useOnClickOutside(node, open ? toggleMenu : undefined)
+  const [activeTime, setTime] = useState(TimeOption.Day)
+
+  return (
+    <StyledMenu ref={node as any}>
```
JFrankfurt, line 151: Is the `as any` necessary here? I think you might be able to drop it if you do this above:
```
const node = useRef<HTMLDivElement | null>(null)
```

src/pages/Explore/index.tsx
```diff
@@ -1,9 +1,13 @@
+// import TimeSelector from 'components/Explore/TimeSelector'
```
JFrankfurt, line 1: drop comment

src/pages/Explore/index.tsx
```diff
@@ -1,9 +1,13 @@
...
+import TimeSelector from 'components/Explore/TimeSelector'
 import AppBody from 'pages/AppBody'
 
 const Explore = () => {
   return (
-    <AppBody>
-      <div>asdf</div>
+    <AppBody background-color="gray">
+      <div>
```
JFrankfurt, line 8: I think you can drop this div wrapper

vm (approved):

PR #3964, opened 2022-06-23
feat: fix metamask mobile browser connection

vm:
Eager connection for MetaMask connector hangs in web3-react. I will fix it in there later, but for now we special case to activate if we know we're in MetaMask's browser.

### DIFF
--- a/src/components/Web3Provider/index.tsx
+++ b/src/components/Web3Provider/index.tsx
@@ -4,9 +4,14 @@ import { BACKFILLABLE_WALLETS, getConnectorForWallet, gnosisSafe, network, useCo
 import { ReactNode, useEffect } from 'react'
 import { useAppSelector } from 'state/hooks'
 
+import { isMobile } from '../../utils/userAgent'
+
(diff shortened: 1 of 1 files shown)


src/components/Web3Provider/index.tsx
```diff
@@ -22,14 +24,18 @@ export default function Web3Provider({ children }: { children: ReactNode }) {
...
     connect(network)
 
-    if (selectedWallet) {
+    if (isMobile && isMetaMask) {
+      connect(injected, true)
+    } else if (selectedWallet) {
       connect(getConnectorForWallet(selectedWallet))
     } else if (!selectedWalletBackfilled) {
-      BACKFILLABLE_WALLETS.map(getConnectorForWallet).forEach(connect)
+      BACKFILLABLE_WALLETS.map(getConnectorForWallet).forEach((connector) => connect(connector))
```
JFrankfurt, line 38: this change shouldn't be necessary, right?

JFrankfurt (approved):

PR #3984, opened 2022-06-28
feat: add token table loading state

kayleegeorge:
feat: add loading state for token table when data is still being fetched 

Screen Shot 2022-06-28 at 1 58 30 PM

### DIFF
--- a/src/components/Explore/TokenRow.tsx
+++ b/src/components/Explore/TokenRow.tsx
@@ -4,7 +4,7 @@ import { useCurrency, useToken } from 'hooks/Tokens'
 import useTheme from 'hooks/useTheme'
 import { TimePeriod, TokenData } from 'hooks/useTopTokens'
 import { darken } from 'polished'
-import React from 'react'
+import React, { ReactNode } from 'react'
 import { ArrowDownRight, ArrowUpRight, Heart } from 'react-feather'
 import { ArrowDown, ArrowUp } from 'react-feather'
 import styled from 'styled-components/macro'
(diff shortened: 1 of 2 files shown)


src/components/Explore/TokenTable.tsx
```diff
@@ -25,7 +25,17 @@ export default function TokenTable() {
   const { data, error, loading } = useTopTokens()
   const timePeriod = TimePeriod.day
   if (loading) {
-    return <GridContainer>Top Token Data Loading</GridContainer>
+    const loadingRows = Array(10)
```
vm, line 28: i think you can make this once as a constant, or memoize this?

vm: since otherwise you create this every render

src/components/Explore/TokenRow.tsx
```diff
@@ -207,6 +236,39 @@ function HeaderCell({
   return <Trans>{getHeaderDisplay(category, timeframe)}</Trans>
 }
 
+export function LoadingRow({ key }: { key: string }) {
+  return (
```
vm, line 240: do you think we should do this inline or make a separate row? my opinion is do it inline in the main row

vm: another option would be to make a `Row` component and then have `LoadedRow` and `LoadingRow` that use it. then the logic isn't coupled but also is abstracted

src/components/Explore/TokenRow.tsx
```diff
@@ -174,6 +174,35 @@ const VolumeCell = styled(Cell)`
...
+const IconLoadingBubble = styled(LoadingBubble)`
+  border-radius: 50%;
+  width: 24px;
+`
+
+const SparkLineLoadingBubble = styled(LongLoadingBubble)`
+  height: 4px;
+`
+/*
+function LoadingBubble({ width, height }: { width: number; height: number }) {
```
vm, line 201: rm

src/components/Explore/TokenRow.tsx
kayleegeorge: tokenRow, LoadedRow, and LoadingRow

src/components/Explore/TokenRow.tsx
```diff
@@ -207,61 +230,128 @@ function HeaderCell({
...
+  price: JSX.Element | string
+  percentChange: JSX.Element
+  marketCap: JSX.Element | string
+  volume: JSX.Element | string
+  sparkLine: JSX.Element | null
+  swap: JSX.Element | null
+}) {
+  const headerRow = header ? 'header' : undefined
+  return (
+    <TokenRowWrapper key={key} className={headerRow}>
```
vm, line 261: make header a prop

src/components/Explore/TokenRow.tsx
```diff
@@ -207,61 +230,128 @@ function HeaderCell({
...
+  price,
+  percentChange,
+  marketCap,
+  volume,
+  sparkLine,
+  swap,
+}: {
+  header: boolean
+  key: string
+  favorited: JSX.Element | null
```
vm, line 249: ReactNode for all these

PR #3988, opened 2022-06-28
fix: add mobile responsiveness break point

kayleegeorge:
added mobile responsiveness break point for token table
Screen Shot 2022-06-28 at 5 23 30 PM


vm (approved):
delete the loading pr but otherwise good!

PR #3992, opened 2022-06-28
chore: upgrade to react 18

lynnshaoyu:
Upgrade to React 18.
https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html

Note: react hooks testing library doesn't support react 18 yet. Therefore, we use equivalent methods from react testing library version 13.1 instead, as per guidance here: https://github.com/testing-library/react-hooks-testing-library


src/components/SearchModal/CurrencyList/index.tsx
```diff
@@ -219,7 +219,15 @@ export default function CurrencyList({
   }, [currencies, otherListTokens])
 
   const Row = useCallback(
-    function TokenRow({ data, index, style }) {
+    function TokenRow({
+      data,
+      index,
+      style,
+    }: {
```
cmcewen, line 226: generally like to define a separate interface `TokenRowProps`. also as i'm reading this code i have a feeling there is a cleaner way we could do this list, but we can leave it for another time

cmcewen (commented):
overall this looks good to me. this is the type of change that could break things in unexpected ways though, so i wonder if we should do a little bug bash, write some more cypress tests, try other browsers, etc to derisk landing it

src/index.tsx
```diff
@@ -42,28 +42,30 @@ function Updaters() {
...
-          </Web3Provider>
-        </LanguageProvider>
-      </HashRouter>
-    </Provider>
-  </StrictMode>,
-  document.getElementById('root')
-)
+const container = document.getElementById('root')
+
+if (container)
```
zzmp, line 47: If there is no container then this should fail (by throwing an error), which I'd expect React to do on it's own. I think this is the rare case where you should assert that container exists on line 45

    const container = document.getElementById('root')!

src/pages/Swap/index.tsx
```diff
@@ -147,11 +147,9 @@ export default function Swap({ history }: RouteComponentProps) {
     () => [!trade?.swaps, TradeState.LOADING === tradeState, TradeState.SYNCING === tradeState],
     [trade, tradeState]
   )
-  // show price estimates based on wrap trade
-  const inputValue = showWrap ? parsedAmount : trade?.inputAmount
-  const outputValue = showWrap ? parsedAmount : trade?.outputAmount
-  const fiatValueInput = useUSDCValue(inputValue)
-  const fiatValueOutput = useUSDCValue(outputValue)
+
+  const fiatValueInput = useUSDCValue(trade?.inputAmount)
```
zzmp, line 151: I think this changes the behavior. Can you revert L151-152?

zzmp (changes requested):
Mostly good, just one potential regression.

The lockfile has a large change, so I'm going to pull it and check it locally before approving the PR.

yarn.lock
```diff
@@ -8428,6 +8447,11 @@ dom-accessibility-api@^0.5.6:
   resolved "https://registry.yarnpkg.com/dom-accessibility-api/-/dom-accessibility-api-0.5.6.tgz#3f5d43b52c7a3bd68b5fb63fa47b4e4c1fdf65a9"
   integrity sha512-DplGLZd8L1lN64jlT27N9TVSESFR5STaEJvX+thCby7fuCHonfPpAlodYc3vuUYbDuDec5w8AMP7oCM5TWFsqw==
 
+dom-accessibility-api@^0.5.9:
```
zzmp, line 8450: This can be coalesced with the dep above it. You can do it manually by deleting lines 8445-8449 and changing 8450 to

    dom-accessibility-api@^0.5.6, dom-accessibility-api@^0.5.9:

You can also do it via CLI with `npx yarn-deduplicate --packages dom-accesibility-api`

yarn.lock
```diff
@@ -4210,7 +4210,7 @@
   dependencies:
     "@types/react" "*"
 
-"@types/react@*", "@types/react@>=16.9.0", "@types/react@^17.0.2":
+"@types/react@*":
```
zzmp, line 4213: This can be coalesced with the type below it by deleting this entire entry and prepending `"types/react@*, ` to L4222, or by running `npx yarn-deduplicate --packages @types/react`.

yarn.lock
```diff
@@ -3965,7 +3967,12 @@
   resolved "https://registry.yarnpkg.com/@types/history/-/history-4.7.9.tgz#1cfb6d60ef3822c589f18e70f8b12f9a28ce8724"
   integrity sha512-MUc6zSmU3tEVnkQ78q0peeEjKWPUADMlC/t++2bI8WnAG2tvYRPIgHG8lWkXwqc8MsUF6Z2MOf+Mh5sazOmhiQ==
 
-"@types/hoist-non-react-statics@*", "@types/hoist-non-react-statics@^3.3.0":
+"@types/history@^4.7.11":
```
zzmp, line 3970: This can be coalesced with the entry above it (see other comments for how to do so).

src/index.tsx
lynnshaoyu: done, just used type assertion since non-null postfix operator is disabled in lint checks (generates error).

package.json
```diff
@@ -110,7 +110,7 @@
     "@fontsource/ibm-plex-mono": "^4.5.1",
     "@fontsource/inter": "^4.5.1",
     "@lingui/core": "^3.9.0",
-    "@lingui/macro": "^3.9.0",
+    "@lingui/macro": "^3.14.0",
```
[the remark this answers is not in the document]
lynnshaoyu: we need it. a lot of <Trans /> tags were causing TS errors without this update.

yarn.lock
lynnshaoyu: done ty

yarn.lock
lynnshaoyu: done ty

src/components/AccountDetails/index.tsx
```diff
@@ -3,10 +3,11 @@ import { useWeb3React } from '@web3-react/core'
 import CopyHelper from 'components/AccountDetails/Copy'
 import { coinbaseWalletConnection, injectedConnection } from 'connection'
 import { getConnection } from 'connection/utils'
-import { useCallback, useContext } from 'react'
+import React, { useCallback, useContext } from 'react'
```
zzmp, line 6: Import `Context` directly to avoid importing all of `React`.

src/components/AddressInputPanel/index.tsx
```diff
@@ -2,8 +2,8 @@ import { Trans } from '@lingui/macro'
 // eslint-disable-next-line no-restricted-imports
 import { t } from '@lingui/macro'
 import { useWeb3React } from '@web3-react/core'
-import { ReactNode, useCallback, useContext } from 'react'
-import styled, { ThemeContext } from 'styled-components/macro'
+import React, { ChangeEvent, ReactNode, useCallback, useContext } from 'react'
```
zzmp, line 5: Import `Context` directly to avoid importing all of `React`.

src/components/Menu/index.tsx
```diff
@@ -311,7 +311,7 @@ export default function Menu() {
 
 interface NewMenuProps {
   flyoutAlignment?: FlyoutAlignment
-  ToggleUI?: React.FunctionComponent
+  ToggleUI?: React.FunctionComponent<React.PropsWithChildren<unknown>>
```
zzmp, line 314: Import these directly to avoid importing all of `React`.

src/pages/Swap/index.tsx
```diff
@@ -12,12 +12,12 @@ import { MouseoverTooltip } from 'components/Tooltip'
 import { useSwapCallback } from 'hooks/useSwapCallback'
 import useTransactionDeadline from 'hooks/useTransactionDeadline'
 import JSBI from 'jsbi'
-import { useCallback, useContext, useEffect, useMemo, useState } from 'react'
+import React, { useCallback, useContext, useEffect, useMemo, useState } from 'react'
```
zzmp, line 15: Import `Context` directly to avoid importing all of `React`.

src/components/AccountDetails/index.tsx
lynnshaoyu: done

src/components/Menu/index.tsx
lynnshaoyu: done

cmcewen (approved):

PR #3993, opened 2022-06-28
feat: disconnect for coinbase wallet

vm:
Adds disconnect for Coinbase Wallet. Preventing reload was a feature added in the coinbase wallet sdk v3.3.0


vm:
Dependent on their SDK emitting disconnect to the provider to make this work with web3-react.

src/components/AccountDetails/index.tsx
```diff
@@ -265,19 +265,29 @@ export default function AccountDetails({
...
+                          connector.resetState()
+                        }
+                      } else {
+                        connector.resetState()
+                      }
+
+                      dispatch(updateSelectedWallet({ wallet: undefined }))
+                      openOptions()
+                    }}
+                    data-cy="wallet-disconnect"
```
zzmp, line 287: Is this intentional, or can it be removed?

zzmp (approved):

src/components/AccountDetails/index.tsx
vm: good point i don't need this

PR #3995, opened 2022-06-29
feat: show injected options in wallet browsers

vm:
If we are in Coinbase Wallet browser then we need to show the Coinbase Wallet option in the WalletModal. Similar logic for MetaMask.

In the future, it would be good to figure out how to avoid special-casing this type of logic.

To reproduce the current issue:
- Open Metamask Mobile browser
- Navigate to app.uniswap.org
- Decline the signature request
- Click the connect button

You'll see a modal with no options like this:
IMG_0593

After:
IMG_0594

### DIFF
--- a/src/components/WalletModal/index.test.tsx
+++ b/src/components/WalletModal/index.test.tsx
@@ -53,7 +53,7 @@ test('Loads Wallet Modal on mobile', async () => {
   global.window.web3 = undefined
 
   render(<WalletModal pendingTransactions={[]} confirmedTransactions={[]} />)
-  // expect(screen.getByText('Open in Coinbase Wallet app.')).toBeInTheDocument()
+  expect(screen.getByText('Open in Coinbase Wallet app.')).toBeInTheDocument()
   expect(screen.getByText('WalletConnect')).toBeInTheDocument()
   expect(screen.getByText('Fortmatic')).toBeInTheDocument()
 })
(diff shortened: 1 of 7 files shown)


JFrankfurt:
It seems like we've had a lot of inconsistent behavior around these options rendering correctly. Can you write a test for this that makes assertions on which options should be rendering in certain contexts?

JFrankfurt (changes requested):
It seems like we've had a lot of inconsistent behavior around these options rendering correctly. Can you write a test for this that makes assertions on which options should be rendering in certain contexts?

src/components/WalletModal/index.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+  UserAgentMock.isMobile = true
+
+  render(<WalletModal pendingTransactions={[]} confirmedTransactions={[]} />)
+  expect(screen.getByText('Open in Coinbase Wallet')).toBeInTheDocument()
+  expect(screen.getByText('WalletConnect')).toBeInTheDocument()
+  expect(screen.getByText('Fortmatic')).toBeInTheDocument()
+  expect(screen.getAllByTestId('wallet-modal-option')).toHaveLength(3)
+})
+
+test('Loads Wallet Modal on MetaMask browser', async () => {
```
lynnshaoyu, line 71: Can we change these test names to things like "it('loads wallet modal on MetaMask browser'....". The 'it' is an alias for test and its supposed to be a naming convention like "it ... is supposed to behave like this". 
All the other tests in the repo use 'it' currently so just want to keep it consistent.

lynnshaoyu (approved):
LGTM over all. Approved, just pls change the tests naming before submit. Ty

Issue #3997, opened 2022-06-29
Crash report: `Error: invalid BigNumber string (argument="value", value="test", code=INVALID_ARGUMENT, version=bignumber/5.6.0)`


BilalMir135:
## URL
  
https://app.uniswap.org/#/pool/test?chain=rinkeby

## Error

```
Error: invalid BigNumber string (argument="value", value="test", code=INVALID_ARGUMENT, version=bignumber/5.6.0)
```

## Stacktrace

```
Error: invalid BigNumber string (argument="value", value="test", code=INVALID_ARGUMENT, version=bignumber/5.6.0)
    at e.value (https://app.uniswap.org/static/js/2.cdb10fed.chunk.js:2:71432)
    at e.value (https://app.uniswap.org/static/js/2.cdb10fed.chunk.js:2:71576)
    at e.value (https://app.uniswap.org/static/js/2.cdb10fed.chunk.js:2:71655)
    at Function.value (https://app.uniswap.org/static/js/2.cdb10fed.chunk.js:2:178446)
    at ap (https://app.uniswap.org/static/js/main.355ff957.chunk.js:1:341752)
    at fa (https://app.uniswap.org/static/js/2.cdb10fed.chunk.js:2:2754298)
    at za (https://app.uniswap.org/static/js/2.cdb10fed.chunk.js:2:2763865)
    at Yf (https://app.uniswap.org/static/js/2.cdb10fed.chunk.js:2:2807589)
    at Cs (https://app.uniswap.org/static/js/2.cdb10fed.chunk.js:2:2793839)
    at Ts (https://app.uniswap.org/static/js/2.cdb10fed.chunk.js:2:2793767)
```

## Device data

```json
{
  "ua": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/100.0.4896.88 Safari/537.36",
  "browser": {
    "name": "Chrome",
    "version": "100.0.4896.88",
    "major": "100"
  },
  "engine": {
    "name": "Blink",
    "version": "100.0.4896.88"
  },
  "os": {
    "name": "Linux",
    "version": "x86_64"
  },
  "device": {},
  "cpu": {
    "architecture": "amd64"
  }
}
```

PR #3998, opened 2022-06-29
fix: add tokenId validation on position page

BilalMir135:
Add `tokenId` validation check on position page to avoid DApp crashing on invalid Ids.
Fixed => Issue #3997


JFrankfurt (changes requested):
I like the idea, but I think it would be better to surface this in the current error UX as an explicit "invalid tokenId in url" message

PR #4000, opened 2022-06-29
chore(extensions): add extension recommendations

JFrankfurt:


.vscode/extensions.json
```diff
@@ -0,0 +1,10 @@
+{
+  "recommendations": [
+    "dbaeumer.vscode-eslint",
+    "eamodio.gitlens",
+    "esbenp.prettier-vscode",
```
zzmp, line 5: I thought we ran prettier through eslint. Does it need its own extension? What does this add?

.vscode/extensions.json
```diff
@@ -0,0 +1,10 @@
...
+    "jpoissonnier.vscode-styled-components",
+    "rvest.vs-code-prettier-eslint"
```
zzmp, line 7: How come prettier isn't already enabled through vscode-eslint? What does this add?

.vscode/extensions.json
```diff
@@ -0,0 +1,10 @@
+{
+  "recommendations": [
+    "dbaeumer.vscode-eslint",
+    "eamodio.gitlens",
```
zzmp, line 4: What does this add to _your_ workflow above the native vs code git integrations?

JFrankfurt:
@grabbou we can drop this

I made the PR before we had any extension recs, and I'm fine with just using what we have now

PR #4005, opened 2022-06-30
fix: resolves network switch race condition

vm:
We have a hook that currently handles two cases:
- If the network is switched outside of the app, make sure we change the URL accordingly
- If the a chain is added as a param in the URL, change the network in the app

There are two bugs:
- If the chainId has changed, confirm that the urlChainId is stale before switching the history
- If There is a urlChainId, make sure that it's actually changed before automatically selecting it in the app

On prod reproduce the bug by having no wallet selected, then switching from ETH to Polygon. It will switch back to ETH.

### DIFF
--- a/src/components/Header/NetworkSelector.tsx
+++ b/src/components/Header/NetworkSelector.tsx
@@ -292,27 +292,14 @@ export default function NetworkSelector() {
         if (!skipToggle) {
           toggle()
         }
-        history.replace({
-          search: replaceURLParam(history.location.search, 'chain', getChainNameFromId(targetChain)),
-        })
       } catch (error) {
         console.error('Failed to switch networks', error)
-
-        // we want app network <-> chainId param to be in sync, so if user changes the network by changing the URL
-        // but the request fails, revert the URL back to current chainId
-        if (chainId) {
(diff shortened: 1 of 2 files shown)


src/components/Header/NetworkSelector.tsx
```diff
@@ -292,40 +293,27 @@ export default function NetworkSelector() {
         if (!skipToggle) {
           toggle()
         }
-        history.replace({
```
vm, line 295: this gets handled by the hook which checks for chainId changing

vm:
As part of this, I will add some tests for switching networks in a separate PR.

src/components/Header/NetworkSelector.tsx
```diff
@@ -292,9 +293,6 @@ export default function NetworkSelector() {
...
-          search: replaceURLParam(history.location.search, 'chain', getChainNameFromId(targetChain)),
-        })
       } catch (error) {
         console.error('Failed to switch networks', error)
```
vm, line 298: the logic here to revert url param i'd like to move as part of the hook instead, by returning success/failure from this async function. later!

JFrankfurt (approved):

PR #4014, opened 2022-07-01
build: simplify actions

zzmp:
- Simplifies github actions (CI/CD)
- Introduces reusable actions (to be extracted to a separate infra repo)


.github/workflows/check-pr-title.yaml
```diff
@@ -8,8 +8,7 @@ on:
       - synchronize
 
 jobs:
-  check-pr-title:
-    name: Check PR Title
+  semantic-commit:
```
vm, line 11: not sure what semantic-commit is

.github/workflows/release.yaml
```diff
@@ -1,62 +1,48 @@
...
   workflow_dispatch:
 
 jobs:
-  bump_version:
-    name: Bump Version
+  wait-on-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: jitterbit/await-check-suites@v1
+        onlyFirstCheckSuite: true
```
vm, line 13: can you tell me what this flag does?

.github/workflows/test.yml
```diff
@@ -7,33 +7,33 @@ on:
   pull_request:
     branches:
       - main
+  # manual trigger
+  workflow_dispatch:
```
vm, line 11: not sure what this is

vm (approved):

.github/workflows/check-pr-title.yaml
zzmp: this should be [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/), because that is what it is checking.

.github/workflows/release.yaml
zzmp: this only waits for the first check suite. this is meant to wait for the test suite generated on merge or push, as opposed to including any suites that are run afterwards, which may include other things (crowdin, or other failed releases).

.github/workflows/test.yml
zzmp: this gives maintainers the ability to run the test check suite manually, so that it can be run against branches besides main on a case-by-case basis.

PR #4015, opened 2022-07-01
feat: don't allow disconnect/change on injected mobile wallets

vm:
Image from iOS

If on MetaMask or Coinbase Wallet browser, you shouldn't be able to change into a different wallet.

### DIFF
--- a/src/components/AccountDetails/index.tsx
+++ b/src/components/AccountDetails/index.tsx
@@ -233,6 +233,7 @@ export default function AccountDetails({
 
   const isMetaMask = !!window.ethereum?.isMetaMask
   const isCoinbaseWallet = !!window.ethereum?.isCoinbaseWallet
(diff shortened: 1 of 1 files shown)


src/components/AccountDetails/index.tsx
```diff
@@ -230,6 +231,9 @@ export default function AccountDetails({
   const theme = useContext(ThemeContext)
   const dispatch = useAppDispatch()
 
+  const isMetaMask = !!window.ethereum?.isMetaMask
+  const isCoinbaseWallet = !!window.ethereum?.isCoinbaseWallet
```
zzmp, line 235: Make another variable to use in the tsx, so that the intent is clearer in code:

const isMobileWallet

vm: good point

zzmp (approved):

zzmp (approved):

PR #4016, opened 2022-07-01
fix: fiat row fix

kayleegeorge:
fixed height for fiat row so button doesn't jump

Screen Shot 2022-07-01 at 3 24 12 PM


kayleegeorge:
Whoops wrong branch

PR #4019, opened 2022-07-01
fix: sync chain query parameter

vm:
4 main cases to consider:
- If there is no chain query param, set it to the current chain
- If the chain changed but the query param is stale, update to the current chain
- If the query param changed, and the chain didn't change, then activate the new chain
- If we activate a chain and it fails, reset the query param to the current chain id


vm:
I did my best to name the constants in the effects to explain it, but what I can do is name the hooks themselves.

vm:
i'm going to write tests for this

src/components/Header/NetworkSelector.tsx
```diff
@@ -242,9 +242,9 @@ function Row({
 
 const getParsedChainId = (parsedQs?: ParsedQs) => {
```
zzmp, line 243: Can you make this a function, and explicitly type the return type? It looks like it's changed, and can now be `void | string`?

    function getChainIdFromParsedQs(parsedQs?: ParsedQs): string | undefined {

src/components/Header/NetworkSelector.tsx
```diff
@@ -242,9 +242,9 @@ function Row({
...
   const chain = parsedQs?.chain
-  if (!chain || typeof chain !== 'string') return { urlChain: undefined, urlChainId: undefined }
+  if (!chain || typeof chain !== 'string') return
 
-  return { urlChain: chain.toLowerCase(), urlChainId: getChainIdFromName(chain) }
+  return getChainIdFromName(chain)
 }
 
 const getChainIdFromName = (name: string) => {
```
zzmp, line 250: Can you make this a function, too, and be more explicit with the return type?

    function getChainIdFromName(name: string): ChainId {

src/components/Header/NetworkSelector.tsx
```diff
@@ -267,20 +267,37 @@ const NETWORK_SELECTOR_CHAINS = [
 
 export default function NetworkSelector() {
   const dispatch = useAppDispatch()
-  const { chainId, provider, connector } = useWeb3React()
-  const previousChainId = usePrevious(chainId)
+
+  const { chainId, provider, connector, isActive } = useWeb3React()
+  const [previousChainId, setPreviousChainId] = useState<number | undefined>(undefined)
+
+  // Can't use `usePrevious` because `chainId` can be undefined while activating.
```
zzmp, line 274: Could you use `useLast`? Or add something like https://github.com/Uniswap/widgets/pull/56?

src/components/Header/NetworkSelector.tsx
```diff
@@ -295,37 +312,42 @@ export default function NetworkSelector() {
...
 
       if (!skipToggle) {
         toggle()
       }
     },
-    [connector, toggle, dispatch]
+    [connector, toggle, dispatch, replaceURLChainParam]
   )
 
+  // If there is no chain query param, set it to the current chain
```
zzmp, line 327: Could these be combined into a single useEffect?

cmcewen (approved):

PR #4023, opened 2022-07-01
feat: remove Tally Ho specific rendering

vm:
We are having difficulty supporting custom Injected wallets beyond MetaMask. Longer-term we'd like to have a separate connector for `injected` and `metaMask` in `web3-react` so they can be resolved separately. However, checking for specific wallets like `window.ethereum.isTally` is not a solution that makes sense for us right now.

For now, the injected wallet should will connect for Tally Ho when it's installed for a user already, except without custom images/copy.


greg-nagy:
Hey, 

This is only visible to ppl who have tally installed and set as the default wallet. 
I don't see the benefit of removing it, so I would like to ask you not to do it.

vm:
@greg-nagy, I added some context above

I think if we need custom connectors in the app to support Tally Ho that doesn't make sense for the interface right now https://github.com/Uniswap/interface/pull/4023

However, I noticed there was some code added on your end to support Uniswap: https://github.com/tallycash/extension/pull/1726

We'll likely add a generic Injected connector in `web3-react` so the MetaMask-specific logic isn't forced on other Injected wallets.

greg-nagy:
@vm it makes me a bit sad because with this decision you are enforcing the `window.ethereum` fight and a lot of non-standard solutions. Basically, you are forcing wallet providers to adopt whatever MM has done, even if it doesn't make sense. 

Anyhow to be on the constructive side of things: Can I help with the generic injected connector for `web3-react`? This topic has been a thorn in my side for the last year and I think there is an easy and backward-compatible fix, we just need wide adoption.

PR #4028, opened 2022-07-05
feat: configure blocks per fetch for l2s

lynnshaoyu:
- update redux multicall package to latest version 1.1.5
- set arbitrum blocks per fetch to 15
-  set optimism blocks per fetch to 15


src/lib/state/multicall.tsx
```diff
@@ -10,9 +11,31 @@ export const store = createStore(reducer)
...
+      return 1
+  }
+}
+
 export function MulticallUpdater() {
   const { chainId } = useWeb3React()
   const latestBlockNumber = useBlockNumber()
   const contract = useInterfaceMulticall()
-  return <multicall.Updater chainId={chainId} latestBlockNumber={latestBlockNumber} contract={contract} />
+  const listenerOptions: ListenerOptions = {
```
zzmp, line 29: Wrap this in useMemo to avoid recreating the object in every render.

zzmp: nit: order these the same as you order the props.

zzmp (changes requested):

src/lib/state/multicall.tsx
```diff
@@ -10,9 +12,34 @@ export const store = createStore(reducer)
 
 export default multicall
 
+function getBlocksPerFetchForChainId(chainId: number | undefined): number {
+  switch (chainId) {
+    case SupportedChainId.ARBITRUM_ONE:
+      return 15
```
zzmp, line 18: Omit this line - intentionally fall through to codify that they will use the same value.

zzmp (approved):

PR #4040, opened 2022-07-06
refactor: extract Web3Provider hooks, create internal Connection representation

vm:
- Moves hooks related to `web3-react` into their own files to simplify understanding this section of the codebase. This sets us up to add some more hooks coming up, like logging errors and falling back to different network providers.
- Renames `Wallet` enum to `ConnectionType`.
- Renames `wallet` reducer to `connection`.
- Creates a `Connection` interface to use internally for `web3-react` that has a `type`, `connector`, and `hooks`.
- generic `getConnection` function which takes a `Connector` or `ConnectionType`.


zzmp:
If you intend to add more hooks, could you name the files after the hooks they will include?

src/hooks/useConnectors.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+function getConnectionForConnectionType(connectionType: ConnectionType): Connection {
+  return {
+    connector: getConnectorForConnectionType(connectionType),
+    hooks: getHooksForConnectionType(connectionType),
+  }
+}
+
+export default function useConnectors() {
```
vm, line 23: only change is i added some comments

src/connectors/utils.ts
```diff
@@ -0,0 +1,69 @@
+import { Connector } from '@web3-react/types'
```
vm, line 1: moved from connectors/index

src/components/AccountDetails/index.tsx
```diff
@@ -241,7 +241,8 @@ export default function AccountDetails({
     const name = Object.keys(SUPPORTED_WALLETS)
       .filter(
         (k) =>
-          SUPPORTED_WALLETS[k].connector === connector && (connector !== injected || isMetaMask === (k === 'METAMASK'))
```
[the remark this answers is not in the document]
vm: generally agree. was trying to avoid touching this code too much since it's a bit of a rabbithole... the SUPPORTED_WALLETS constant is what's powering this file and WalletModal, but it could use a bit of a re-work. i'd like to move to using `connectionType` as the source of truth like you said, but not in this PR.

src/components/WalletModal/index.tsx
```diff
@@ -144,39 +147,39 @@ export default function WalletModal({
 
   useEffect(() => {
     if (pendingConnector && walletView !== WALLET_VIEWS.PENDING) {
-      updateWalletError({ wallet: getWalletForConnector(pendingConnector), error: undefined })
+      updateConnectionError({ connectionType: getConnectionForConnector(pendingConnector).type, error: undefined })
```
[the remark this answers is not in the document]
vm: mostly because the other stuff in connection (hooks, connector) should be unnecessary for updating the error

src/components/Identicon/StatusIcon.tsx
```diff
@@ -1,22 +1,42 @@
...
+    width: ${({ size }) => (size ? size + 'px' : '32px')};
   }
+  ${({ theme }) => theme.mediaWidth.upToMedium`
+    align-items: flex-end;
+  `};
+`
+
+export default function StatusIcon({ connectionType }: { connectionType: ConnectionType }) {
+  let image
+  switch (connectionType) {
```
cmcewen, line 26: 👨‍🍳 💋

cmcewen (approved):
LGTM! still think we can continue to clean this up over time but this is a big step

PR #4054, opened 2022-07-07
fix: unsupported chain displays message instead of crash

cartcrom:
An unsupported network message appears rather than an error display when navigating to the pools page on an unsupported chain.

image


src/pages/Pool/index.tsx
```diff
@@ -126,15 +137,55 @@ function PositionsLoadingPlaceholder() {
   )
 }
 
+function WrongNetworkCard() {
+  const theme = useContext(ThemeContext)
+  return (
+    <>
+      <PageWrapper>
+        <SwapPoolTabs active={'pool'} />
```
vm, line 145: `="pool"`

src/pages/Pool/index.tsx
```diff
@@ -89,6 +90,16 @@ const NoLiquidity = styled.div`
   max-width: 300px;
   min-height: 25vh;
 `
+
+const WrongNetwork = styled.div`
```
vm, line 94: this styling seems very related to the `NoLiquidity` above. Maybe you can rename that to `ErrorContainer` or something?

vm:
Ideally, good to use `styled-components` instead of inline styles if possible

cmcewen:
@cartcrom there are also some crashes from trying to destructure into an object that doesn't exist (CHAIN_INFO does not have keys for all possible chainId values): https://github.com/Uniswap/interface/blob/main/src/pages/Pool/CTACards.tsx#L97

could you just make sure that crash can't happen in that part of the code too?

cartcrom:
> @cartcrom there are also some crashes from trying to destructure into an object that doesn't exist (CHAIN_INFO does not have keys for all possible chainId values): https://github.com/Uniswap/interface/blob/main/src/pages/Pool/CTACards.tsx#L97
> 
> could you just make sure that crash can't happen in that part of the code too?

@cmcewen the way it works now, the CTA cards are only rendered if the current chain is supported. Should I add a second check in CTAcards anyways?

src/pages/Pool/index.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+      return { loading: false, positions: undefined }
+    })
+
+    act(() => {
+      const { asFragment } = render(
+        <Router>
+          <Pool />
+        </Router>
+      )
+      expect(asFragment()).toMatchSnapshot()
```
vm, line 31: i think better to test whether it shows error text here vs using a snapshot

src/pages/Pool/index.test.tsx
```diff
@@ -0,0 +1,50 @@
...
```
vm, line 47: same feedback as above

src/pages/Pool/CTACards.tsx
```diff
@@ -93,8 +94,9 @@ const ResponsiveColumn = styled(AutoColumn)`
 `
 
 export default function CTACards() {
-  const { chainId } = useWeb3React()
-  const { infoLink } = CHAIN_INFO[chainId ? chainId : SupportedChainId.MAINNET]
+  const { chainId, connector } = useWeb3React()
+  const { infoLink } = CHAIN_INFO[chainId && isChainAllowed(connector, chainId) ? chainId : SupportedChainId.MAINNET]
```
vm, line 98: To avoid this in the future, lets make a function called `getChainInfo` and then don't export `CHAIN_INFO`. this function should throw an error if the chainId isn't supported, which we'll be forced to catch.

src/pages/Pool/index.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+    jest.spyOn(useV3Positions, 'useV3Positions').mockImplementation(() => {
+      return { loading: false, positions: undefined }
+    })
+
+    render(
+      <Router>
+        <Pool />
+      </Router>
+    )
+    expect(screen.getAllByTestId('pools-unsupported-err')).toHaveLength(1)
```
vm, line 30: better to check for the text to display rather than checking for the id

vm:
in a separate PR, lets create a `getChainInfo` function and stop exporting CHAIN_INFO

src/pages/Pool/index.test.tsx
cartcrom: screen.getByText wasn't working for me here because there was another element (the link) embedded into the Trans the displayed text was in. I messed around for a while trying to match the text and link but ended up going with the id, can definitely change it if there is a proper way to use getByText

vm (approved):

PR #4061, opened 2022-07-08
feat: Web 214 implement the main submit swap event

lynnshaoyu:
Implement "Swap Submitted" event. 

Screen Shot 2022-07-21 at 5 38 10 PM
Screen Shot 2022-07-21 at 5 38 06 PM

left a TODO for one last event property: "duration_from_first_quote_to_swap_submission_seconds", will implement this along with https://uniswaplabs.atlassian.net/browse/WEB-623?atlOrigin=eyJpIjoiM2U5Y2VlMTEyNmRhNDAwMGI0ZGJkYTQyNTdkYjU5MjciLCJwIjoiaiJ9


src/components/swap/SwapModalFooter.tsx
```diff
@@ -1,36 +1,100 @@
...
         >
-          <Text fontSize={20} fontWeight={500}>
-            <Trans>Confirm Swap</Trans>
-          </Text>
-        </ButtonError>
+          <ButtonError
+            onClick={onConfirm}
+            disabled={disabledConfirm}
+            style={{ margin: '10px 0 0 0' }}
+            className={ElementName.CONFIRM_SWAP_BUTTON}
```
vm, line 90: why do we need a classname and id? can't tell if this is a new addition

src/components/swap/SwapModalFooter.tsx
```diff
@@ -1,36 +1,100 @@
...
+  allowedSlippage: Percent,
+  transactionDeadlineSecondsSinceEpoch: number | undefined,
+  isAutoSlippage: boolean,
+  tokenInAmountUsd: string | undefined,
+  tokenOutAmountUsd: string | undefined
+) => ({
+  estimated_network_fee_usd: trade.gasUseEstimateUSD ? parseFloat(trade.gasUseEstimateUSD?.toFixed(2)) : undefined,
+  transaction_hash: txHash,
+  transaction_deadline_seconds: transactionDeadlineSecondsSinceEpoch
+    ? /** durationInSeconds= */ transactionDeadlineSecondsSinceEpoch - new Date().getTime() / 1000
```
vm, line 29: what does this comment mean?

vm (approved):
two comments

src/components/swap/SwapModalFooter.tsx
lynnshaoyu: I'm trying to indicate that it's calculating the duration (in seconds) till the transaction deadline, by subtracting current time (expressed in seconds since epoch) from the transaction deadline (in seconds since epoch) in the future. do you think a different comment would make this clearer?

vm: yeah i would say that in a sentence rather than a variable

src/components/swap/SwapModalFooter.tsx
```diff
@@ -1,36 +1,104 @@
...
+  tokenInAmountUsd: string | undefined,
+  tokenOutAmountUsd: string | undefined,
+  lpFeePercent: Percent
+) => ({
+  estimated_network_fee_usd: trade.gasUseEstimateUSD ? parseFloat(trade.gasUseEstimateUSD?.toFixed(2)) : undefined,
+  transaction_hash: txHash,
+  transaction_deadline_seconds: transactionDeadlineSecondsSinceEpoch
+    ? /** durationInSeconds= */ transactionDeadlineSecondsSinceEpoch - new Date().getTime() / 1000
+    : undefined,
+  token_in_amount_usd: tokenInAmountUsd ? parseFloat(tokenInAmountUsd) : undefined,
```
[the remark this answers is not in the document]
lynnshaoyu: why? what makes them dangerous?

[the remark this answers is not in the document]
cmcewen: we use BigNumbers a lot and the conversion back and forth can be lossy https://stackoverflow.com/questions/7988827/parse-float-has-a-rounding-limit-how-can-i-fix-this

src/components/swap/SwapModalFooter.tsx
lynnshaoyu: created a helper function with better naming to reduce confusion: getDurationTillTimestampSinceEpoch

src/components/swap/SwapModalFooter.tsx
```diff
@@ -1,36 +1,104 @@
...
+  trade: InterfaceTrade<Currency, Currency, TradeType>,
+  txHash: string | undefined,
+  allowedSlippage: Percent,
+  transactionDeadlineSecondsSinceEpoch: number | undefined,
+  isAutoSlippage: boolean,
+  tokenInAmountUsd: string | undefined,
+  tokenOutAmountUsd: string | undefined,
+  lpFeePercent: Percent
+) => ({
+  estimated_network_fee_usd: trade.gasUseEstimateUSD ? parseFloat(trade.gasUseEstimateUSD?.toFixed(2)) : undefined,
```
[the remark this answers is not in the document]
lynnshaoyu: yeah sounds good, added getFormattedNumber helper

src/components/swap/AdvancedSwapDetails.tsx
```diff
@@ -59,7 +63,7 @@ export function AdvancedSwapDetails({
     if (!trade) return { expectedOutputAmount: undefined, priceImpact: undefined }
     const expectedOutputAmount = trade.outputAmount
     const realizedLpFeePercent = computeRealizedLPFeePercent(trade)
-    const priceImpact = trade.priceImpact.subtract(realizedLpFeePercent)
+    const priceImpact = getPriceImpact(realizedLpFeePercent, trade)
```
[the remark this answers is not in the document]
lynnshaoyu: woohoo :D

src/components/swap/SwapModalFooter.tsx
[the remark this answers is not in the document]
lynnshaoyu: ooohhh that's fair....that reminds me I should ask about significant digits in product-analytics thread...

src/components/swap/SwapModalFooter.tsx
```diff
@@ -83,6 +91,7 @@ export default function SwapModalFooter({
             allowedSlippage,
             transactionDeadlineSecondsSinceEpoch,
             isAutoSlippage,
+            /** isAutoRouterApi= */ !clientSideRouter,
```
vm, line 94: i think you should make the argument to `formatAnalyticsEventProperties` a single object, that way all the parameters are named as keys

lynnshaoyu: sgtm done

src/components/swap/SwapModalFooter.tsx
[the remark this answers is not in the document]
lynnshaoyu: Ok I changed it again. we were actually logging the wrong numbers initially for token amounts (yikes) - we were logging the NUMBER of decimal places available, not the full number.

I modified it to log all available decimal numbers for the token amounts.

src/components/swap/SwapModalFooter.tsx
```diff
@@ -1,36 +1,130 @@
...
+    : undefined,
+  transaction_hash: txHash,
+  transaction_deadline_seconds: getDurationTillTimestampSinceEpoch(transactionDeadlineSecondsSinceEpoch),
+  token_in_amount_usd: tokenInAmountUsd ? parseFloat(tokenInAmountUsd) : undefined,
+  token_out_amount_usd: tokenOutAmountUsd ? parseFloat(tokenOutAmountUsd) : undefined,
+  token_in_address: trade.inputAmount.currency.isToken ? trade.inputAmount.currency.address : undefined,
+  token_out_address: trade.outputAmount.currency.isToken ? trade.outputAmount.currency.address : undefined,
+  token_in_symbol: trade.inputAmount.currency.symbol,
+  token_out_symbol: trade.outputAmount.currency.symbol,
+  token_in_amount: getNumberFormattedToDecimalPlace(trade.inputAmount, trade.inputAmount.currency.decimals),
```
vm, line 63: can you use the same function as we have in `Swap/index.tsx`? look for `formattedAmounts`

src/components/swap/SwapModalFooter.tsx
```diff
@@ -1,36 +1,130 @@
...
+  txHash: string | undefined
+  allowedSlippage: Percent
   onConfirm: () => void
   swapErrorMessage: ReactNode | undefined
   disabledConfirm: boolean
 }) {
+  const transactionDeadlineSecondsSinceEpoch = useTransactionDeadline()?.toNumber() // in seconds since epoch
+  const isAutoSlippage = useUserSlippageTolerance() === 'auto'
+  const [clientSideRouter] = useClientSideRouter()
+  const tokenInAmountUsd = useStablecoinValue(trade.inputAmount)?.toFixed(2)
```
vm, line 94: same thing, i think we'll want to use the same logic we use around the rest of the App (sometimes having decimals beyond 2 matters even for USD)

src/components/swap/SwapModalFooter.tsx
lynnshaoyu: that isn't an exportable function, it uses hooks and has a lot of dependencies in the Swap/index.tsx page. I don't think it is appropriate to replicate all of that logic here (seems like overkill)

src/components/swap/SwapModalFooter.tsx
lynnshaoyu: same as above.

cmcewen (approved):
lgtm! i agree that it doesn't make sense to import that other function for formatting, but if you need to use this formatting as you instrument elsewhere just add it to a utils file

lynnshaoyu:
> lgtm! i agree that it doesn't make sense to import that other function for formatting, but if you need to use this formatting as you instrument elsewhere just add it to a utils file

sweet, i'll clean up all the helper functions after initial instrumentation is done into a separate utils file probably. 

also note on the commit after approval: increase decimal places for percentages to 4 to capture granularity to basis point level.

PR #4066, opened 2022-07-08
feat: add more and incorporate dummy data

kayleegeorge:
changes for demo: add more dummy data, fix network selector color, fixed name cell (ellipses if exceeding name cell space), routed correct Protocol Info link (tied to Info Charts token address)

Screen Shot 2022-07-08 at 12 52 10 PM

Screen Shot 2022-07-08 at 5 44 21 PM


src/components/Explore/BalanceSummary.tsx
```diff
@@ -1,8 +1,9 @@
+import { TEXT_COLORS } from 'components/NetworkAlert/NetworkAlert'
```
vm, line 1: put this in a shared file

vm (approved):

src/components/Explore/BalanceSummary.tsx
kayleegeorge: because it's not in the same component file, where do you suggest i put this? (ie this isn't in components/Explore)

kayleegeorge: I guess I could just put a floater file in components but it doesn't seem like that's best practice? thoughts?

PR #4067, opened 2022-07-08
feat: implement token selector events

lynnshaoyu:
Implement all token selector events (currently in taxonomy): 

1. Token Selector Opened
Example:
Screen Shot 2022-07-12 at 5 57 00 PM

2. Token Imported
Example:
Screen Shot 2022-07-12 at 5 53 22 PM

3. Token Selected
Examples:
Screen Shot 2022-07-12 at 5 52 20 PM
Screen Shot 2022-07-12 at 5 52 34 PM


zzmp (approved):
Is this a simple few-line change that the differ couldn't figure out, or did I miss something?

vm:
@zzmp use hide whitespace and the diff is correct

src/components/SearchModal/CurrencySearch.tsx
```diff
@@ -185,76 +187,78 @@ export function CurrencySearch({
...
-            value={searchQuery}
-            ref={inputRef as RefObject<HTMLInputElement>}
-            onChange={handleInput}
-            onKeyDown={handleEnter}
-          />
-        </Row>
-        {showCommonBases && (
-          <CommonBases chainId={chainId} onSelect={handleCurrencySelect} selectedCurrency={selectedCurrency} />
+    <Trace eventName={EventName.TOKEN_SELECTOR_OPENED} section={ModalName.TOKEN_SELECTOR} logImpression={true}>
+      <ContentWrapper className={ModalName.TOKEN_SELECTOR}>
```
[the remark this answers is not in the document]
lynnshaoyu: removed, wanted to label the divs with classNAmes that matched the context names in trace, but not rly necessary

src/components/SearchModal/CommonBases.tsx
```diff
@@ -47,19 +53,39 @@ export default function CommonBases({
       <AutoRow gap="4px">
         {bases.map((currency: Currency) => {
           const isSelected = selectedCurrency?.equals(currency)
+          const token_address = currency instanceof Token ? currency?.address : undefined
```
cmcewen, line 56: i wonder if similar to having `Props` for components we should create a `formatEventProperties` helper function for each component - just to keep. component logic clean?

src/components/SearchModal/CurrencyList/index.tsx
```diff
@@ -249,6 +264,19 @@ export default function CurrencyList({
           <ImportRow style={style} token={token} showImportView={showImportView} setImportToken={setImportToken} dim />
         )
       } else if (currency) {
+        const eventProperties = {
+          token_symbol: token?.symbol,
+          token_address: token?.address,
+          is_suggested_token: false,
+          is_selected_from_list: true,
+          scroll_position: '',
+          token_list_index: index + 1, // 1-indexed
```
cmcewen, line 273: is there a reason to 1-index? as an engineer i always default to 0 index

src/components/SearchModal/ImportToken.tsx
```diff
@@ -42,6 +44,13 @@ export function ImportToken(props: ImportProps) {
   if (intersection.size > 0) {
     return <BlockedToken onBack={onBack} onDismiss={onDismiss} blockedTokens={Array.from(intersection)} />
   }
+
+  const eventPropertiesList = tokens.map((token) => ({
```
cmcewen, line 48: this feels like it could be very large, and i don't love this abstraction tbh - also we lose that this list of tokens was a complete unit by firing a bunch of events. could we do an array of addresses and one event?

src/components/AmplitudeAnalytics/TraceEvent.tsx
```diff
@@ -8,6 +8,7 @@ type TraceEventProps = {
   events: Event[]
   name: EventName
   properties?: Record<string, unknown>
+  propertiesList?: Record<string, unknown>[]
```
cmcewen, line 11: after reading through this i'm not sure i agree a propertiesList should exist. are there other cases where you would want to send many events from one interaction?

lynnshaoyu: yeah I think you're right, this was the main use case I can think of, but I agree w you that sending one event w a token list is better than multiple events, so we can see the aggregation.

src/components/SearchModal/CurrencyList/index.tsx
lynnshaoyu: hmm for PM's and research folks using this data :P i'm kind of indifferent though. let's see what Will thinks

src/components/SearchModal/CommonBases.tsx
lynnshaoyu: good call, added helper function.

src/components/SearchModal/CurrencyList/index.tsx
lynnshaoyu: true, changed to 0 indexed, Will agreed.

cmcewen (approved):
looks nice and clean!

PR #4074, opened 2022-07-11
build: wait on tests

zzmp:
Use a different action to wait on tests, as the old action could not disambiguate check suites.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -6,17 +6,24 @@ on:
   workflow_dispatch:
 
 jobs:
-  wait-on-test:
+  wait-on-tests:
     runs-on: ubuntu-latest
     steps:
-      - uses: jitterbit/await-check-suites@v1
+      - id: unit-tests
+        uses: fountainhead/action-wait-for-check@v1.0.0
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -6,17 +6,24 @@ on:
...
+        uses: fountainhead/action-wait-for-check@v1.0.0
         with:
-          timeoutSeconds: 600
-          # Only wait for the first check suite, so that only tests are awaited, and not eg crowdin.
-          onlyFirstCheckSuite: true
+          token: ${{ secrets.GITHUB_TOKEN }}
+          checkName: unit-tests
+      - id: cypress-tests
+        uses: fountainhead/action-wait-for-check@v1.0.0
+        with:
```
vm, line 19: do `token` and `uses: fountainhead/action-wait-for-check@v1.0.0` need to be repeated like this?

vm (approved):

.github/workflows/release.yaml
zzmp: the `checkName`s are different - this action is repeated for unit and cypress tests.

PR #4079, opened 2022-07-11
feat: header hover states and favorite active state styling

kayleegeorge:
insert token table hover states and favorite button active state styling

Screen Shot 2022-07-11 at 4 48 45 PM

Screen Shot 2022-07-11 at 4 54 01 PM

### DIFF
--- a/src/components/Explore/FavoriteButton.tsx
+++ b/src/components/Explore/FavoriteButton.tsx
@@ -4,7 +4,7 @@ import { darken } from 'polished'
 import { Heart } from 'react-feather'
 import styled from 'styled-components/macro'
 
-import { showFavoritesAtom } from './TokenTable'
+import { showFavoritesAtom } from './state'
 
 const FavoriteButtonContent = styled.div`
   display: flex;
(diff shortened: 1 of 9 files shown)


src/components/Explore/FavoriteButton.tsx
```diff
@@ -10,28 +11,34 @@ const FavoriteButtonContent = styled.div`
   align-items: center;
   gap: 8px;
 `
-const StyledFavoriteButton = styled.button`
+const StyledFavoriteButton = styled.button<{ active: any }>`
```
vm, line 14: boolean

src/components/Explore/FavoriteButton.tsx
```diff
@@ -1,37 +1,45 @@
...
   gap: 8px;
 `
-const StyledFavoriteButton = styled.button`
+const StyledFavoriteButton = styled.button<{ active: boolean }>`
   padding: 0px 16px;
   border-radius: 12px;
   background-color: ${({ theme }) => theme.bg0};
-  border: 1px solid ${({ theme }) => theme.bg0};
-  color: ${({ theme }) => theme.text1};
+  border: 1px solid ${({ theme, active }) => (active ? '#869EFF' : theme.bg1)};
```
cmcewen, line 19: could you add this color to the theme file and use it as a constant from there? https://github.com/Uniswap/interface/blob/main/src/theme/index.tsx

src/components/Explore/FavoriteButton.tsx
```diff
@@ -1,37 +1,45 @@
...
   const theme = useTheme()
   const [showFavorites, setShowFavorites] = useAtom(showFavoritesAtom)
   return (
-    <StyledFavoriteButton onClick={() => setShowFavorites(!showFavorites)}>
+    <StyledFavoriteButton onClick={() => setShowFavorites(!showFavorites)} active={showFavorites}>
       <FavoriteButtonContent>
-        <Heart size={17} color={theme.text1} fill={theme.text1} /> Favorites
+        <Heart
+          size={17}
+          color={showFavorites ? '#869EFF' : theme.text1}
```
cmcewen, line 39: sometimes you might hear people say DRY (https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) - which basically just means, if you have the same block of code, write it once instead of twice. so you could make this a variable above

src/components/Explore/TokenRow.tsx
```diff
@@ -183,13 +195,19 @@ const PercentChangeInfoCell = styled(Cell)`
...
   min-width: 80px;
+  padding-right: 4px;
 
   @media only screen and (max-width: ${MOBILE_MEDIA_BREAKPOINT}) {
     min-width: max-content;
   }
+
+  &:hover {
+    color: ${({ theme, sortable }) => sortable && theme.text1};
+    background-color: ${({ theme, sortable }) => sortable && darken(0.08, theme.bg0)};
```
cmcewen, line 209: i see darken a lot - wonder if we should ask the design team whether to create a new color for our theme

src/components/Explore/TokenRow.tsx
```diff
@@ -386,7 +410,7 @@ export function HeaderRow({ timeframe }: { timeframe: string }) {
           category={Category.price}
           sortDirection={SortDirection.Decreasing}
           isSorted={sortedBy === Category.price}
-          sortable={false}
+          sortable={true}
```
vm, line 413: when it's `true` it can just be `sortable` with no `={true}`

src/components/Explore/TokenTable.tsx
```diff
@@ -46,6 +45,7 @@ const LOADING_ROWS = Array(10)
 
 // TODO: add to shared location and possibly make a set
 export const favoritesAtom = atomWithStorage<string[]>('favorites', [])
+export const showFavoritesAtom = atomWithStorage<boolean>('showFavorites', false)
```
vm, line 48: i think it's time to put this in a separate file. you can have a file called `state.ts`

src/components/Explore/TokenRow.tsx
kayleegeorge: Yup! one of our tickets is the design theme file for Explore

vm (approved):

PR #4082, opened 2022-07-12
build: wait an hour for IPFS release

zzmp:
- Increases the wait time for IPFS to be available during release to an hour.
  The current setup is functional, but eg https://github.com/Uniswap/interface/runs/7298181685?check_suite_focus=true#step:12:69 took 27 (of 31) tries, so the number of retries needs to be increased.
- Documents IPFS initial request delays.
  As part of this PR, the proxy read timeout should be extended to 8m on cloudflare.

Essentially, IPFS content only propagates to a new node when it is requested by a client, so a cloudflare data center's first load will always have some delay. Anecdotally, that delay is about 5 minutes.

We seem to be able to mitigate that delay for US nodes by pre-requesting the asset through cloudflare's public gateway (although we have no guarantees about this). However, this does not mitigate the delay for international nodes.

I think this leaves us with 3 choices:
1. Increase the proxy read timeout so that initial requests to IPFS do not time out.
2. Maintain our own internationally placed IPFS cluster to intentionally propagate and make the content more available.
3. Abandon IPFS as our primary hosting service.

This PR documents (1) an increased proxy read timeout to 8 minutes, which will need to be made separately using cloudflare's API.

### DIFF
--- a/cypress/release.ts
+++ b/cypress/release.ts
@@ -4,17 +4,12 @@ describe(
   'Release',
   {
     pageLoadTimeout: ONE_MINUTE,
-    retries: 30,
+    retries: 60,
   },
   () => {
     it('loads swap page', () => {
       // We *must* wait in order to space out the retry attempts.
-      cy.wait(ONE_MINUTE)
-        .visit('/', {
-          retryOnStatusCodeFailure: true,
-          retryOnNetworkFailure: true,
-        })
-        .get('#swap-page')
+      cy.wait(ONE_MINUTE).visit('/').get('#swap-page')
     })
   }
 )
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -48,7 +48,9 @@ jobs:
       - run: yarn prepare
       - run: yarn build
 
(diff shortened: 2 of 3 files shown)


cypress/release.ts
```diff
@@ -4,17 +4,12 @@ describe(
...
   {
     pageLoadTimeout: ONE_MINUTE,
-    retries: 30,
+    retries: 60,
   },
   () => {
     it('loads swap page', () => {
       // We *must* wait in order to space out the retry attempts.
-      cy.wait(ONE_MINUTE)
-        .visit('/', {
```
vm, line 13: are these params defaulted or something?

zzmp: They're defaulted. I thought that changing the defaults would force retries, but I think it instead made retries occur without so much wait time between them, so I'm putting them back to defaults.

https://docs.cypress.io/api/commands/visit#Arguments

vm (approved):

vm:
I think this is close-able now?

cmcewen:
don't think we need this anymore

PR #4085, opened 2022-07-12
chore: updates web3-react, adds key for changing connector order

vm:
Latest version of web3-react requires the implementor to provider a key when the connectors array changes.

We change the connectors array when we prioritize the currently selected wallet's connector above others. The alternative option would be to use the prop `connectorOverride`, but this would break usage of `gnosisSafe` as that must come first.

I'm open minded on a different way to create this key, and also whether we should memoize it as it's near the top of the heirarchy.


src/hooks/useOrderedConnections.ts
```diff
@@ -25,10 +23,6 @@ export default function useConnectors() {
     // Add network connection last as it should be the fallback.
     orderedConnectionTypes.push(ConnectionType.NETWORK)
 
-    // Convert to web3-react's representation of connectors.
-    const web3Connectors: [Connector, Web3ReactHooks][] = orderedConnectionTypes
-      .map(getConnection)
-      .map(({ connector, hooks }) => [connector, hooks])
-    return web3Connectors
+    return orderedConnectionTypes.map(getConnection)
```
zzmp, line 26: I preferred the add'l mapping in here. Why did you move it?

src/connection/utils.ts
```diff
@@ -42,3 +42,20 @@ export function getConnection(c: Connector | ConnectionType) {
     }
   }
 }
+
+export function getConnectionName(connectionType: ConnectionType, isMetaMask?: boolean) {
```
zzmp, line 46: If this is just for keys, is it still necessary?

yarn.lock
```diff
@@ -1698,6 +1713,13 @@
   dependencies:
     "@ethersproject/logger" "^5.6.0"
 
+"@ethersproject/networks@5.6.4":
```
zzmp, line 1716: can be deduped with above entry.

zzmp (changes requested):
I'm unsure why the code changes are necessary. Can they be backed out?

The package.json looks good but the new lockfile needs to be de-duped.

NoahZinsmeister (approved):
lgtm

PR #4086, opened 2022-07-12
style: removes Trans in NetworkSelector

vm:
Removes translations in NetworkSelector, as we don't want to translate proper nouns.

### DIFF
--- a/src/components/Header/NetworkSelector.tsx
+++ b/src/components/Header/NetworkSelector.tsx
@@ -146,7 +146,8 @@ const SelectorWrapper = styled.div`
 const StyledChevronDown = styled(ChevronDown)`
   width: 16px;
 `
-const BridgeLabel = ({ chainId }: { chainId: SupportedChainId }) => {
+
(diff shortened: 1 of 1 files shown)


src/components/Header/NetworkSelector.tsx
```diff
@@ -146,40 +146,42 @@ const SelectorWrapper = styled.div`
 const StyledChevronDown = styled(ChevronDown)`
   width: 16px;
 `
-const BridgeLabel = ({ chainId }: { chainId: SupportedChainId }) => {
+
+function getBridgeLabel(chainId: SupportedChainId) {
```
vm, line 150: or maybe we wanna make it translate the `Bridge` text but keep the chain untranslated?

zzmp: If we don't translate bridge we still need to use translations for ordering of the words.

zzmp (commented):
Can you mark a word as not translatable within a phrase?

PR #4089, opened 2022-07-12
feat: implement-page-viewed-event-for-all-main-pages-of-app

lynnshaoyu:
Implement page viewed events for: 

- Swap
- Vote
- Pool
- Charts


cmcewen (approved):
woot! just a note you can do `shouldLogImpression` instead of `shouldLogImpression={true}` in the future if you prefer

PR #4091, opened 2022-07-12
feat: favorite token on tokenDetail page

kayleegeorge:
ability to favorite/unfavorite a token directly from its token detail page

### DIFF
--- a/src/components/Explore/TokenDetail.tsx
+++ b/src/components/Explore/TokenDetail.tsx
@@ -1,7 +1,7 @@
 import CurrencyLogo from 'components/CurrencyLogo'
 import { useCurrency, useToken } from 'hooks/Tokens'
 import { TimePeriod } from 'hooks/useTopTokens'
-import { atom, useAtom } from 'jotai'
+import { useAtom } from 'jotai'
 import { darken } from 'polished'
 import { ReactNode, useState } from 'react'
 import { ArrowDownRight, ArrowLeft, ArrowUpRight, Copy, Heart, Share } from 'react-feather'
@@ -55,6 +55,15 @@ const ChartHeader = styled.div`
   color: ${({ theme }) => theme.text1};
   gap: 4px;
   margin-bottom: 24px;
+`
+const ClickFavorited = styled.span`
+  display: flex;
+  align-items: center;
+  cursor: pointer;
+
+  &:hover {
+    color: ${({ theme }) => theme.primary1};
+  }
 `
 const ContractAddress = styled.button`
   display: flex;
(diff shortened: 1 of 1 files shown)


src/components/Explore/state.ts
```diff
@@ -1,4 +1,26 @@
...
+export const toggleFavoriteToken = ({
+  tokenAddress,
+  favoriteTokens,
+  updateFavoriteTokens,
+}: {
+  tokenAddress: string
+  favoriteTokens: string[]
+  updateFavoriteTokens: (update: SetStateAction<string[]>) => void
+}) => {
+  let updatedFavoriteTokens
```
vm, line 17: can't you use the atom in this function instead of passing stuff in?

kayleegeorge: yes i thought that too but useAtom is a hook -- so then it would need to either be called in a react component or function (not super familiar with that part of React so could that just be useEffect or..?)

vm: make a hook called `useToggleFavorite`

src/components/Explore/TokenDetail.tsx
```diff
@@ -205,8 +206,11 @@ export default function LoadedTokenDetail({ address }: { address: string }) {
   const token = useToken(address)
   const currency = useCurrency(address)
   const [favoriteTokens] = useAtom(favoritesAtom)
-  const isFavorited = atom<boolean>(favoriteTokens.includes(address))
+  // const isFavorited = atom<boolean>(favoriteTokens.includes(address))
```
vm, line 209: rm comment

src/components/Explore/state.ts
```diff
@@ -1,4 +1,23 @@
+import { useAtom } from 'jotai'
 import { atomWithStorage } from 'jotai/utils'
+import { useCallback } from 'react'
 
 export const favoritesAtom = atomWithStorage<string[]>('favorites', [])
 export const showFavoritesAtom = atomWithStorage<boolean>('showFavorites', false)
+
+/* handle favorite token logic */
```
vm, line 8: don't need this comment

src/components/Explore/state.ts
```diff
@@ -1,4 +1,23 @@
...
+export function useToggleFavorite(tokenAddress: string): () => void {
```
vm, line 9: rm the return type, should be inferred

.vscode/settings.json
```diff
@@ -5,7 +5,7 @@
   "editor.formatOnSaveMode": "file",
   "editor.tabCompletion": "on",
   "editor.tabSize": 2,
-  "editor.formatOnSave": true,
+  "editor.formatOnSave": false,
```
vm, line 8: prob back out of this

src/components/Header/NetworkSelector.tsx
```diff
@@ -132,7 +132,7 @@ const SelectorControls = styled.div<{ interactive: boolean }>`
   justify-content: space-between;
   padding: 6px 8px;
 `
-const SelectorLogo = styled(Logo)<{ interactive?: boolean }>`
+const SelectorLogo = styled(Logo) <{ interactive?: boolean }>`
```
vm, line 135: bane of our existence

PR #4097, opened 2022-07-13
test: enforce deps deduplication

zzmp:
- Deduplicates yarn.lock using both `highest` and `fewest` strategies
- Enforces deduplication with a CI check


.github/workflows/test.yml
```diff
@@ -25,21 +33,25 @@ jobs:
...
       - run: yarn prepare
-
       - run: yarn build
       - uses: actions/upload-artifact@v2
         with:
           name: build
           path: build
           if-no-files-found: error
 
+  cypress-build:
```
vm, line 50: why does the cypress build need to change?

zzmp: it doesn't, I was just prepping for another test for deps size.

.github/workflows/test.yml
```diff
@@ -17,6 +17,17 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup
       - run: yarn lint
+  
+  deps-tests:
```
vm, line 21: we may wanna explain to users what to do when this fails somehow

vm (approved):

PR #4098, opened 2022-07-13
test: enforce package size

zzmp:
Fails a CI check if an entrypoint file's gzipped size grows by more than 100B.


cmcewen (commented):
what will happen if someone adds a large new component? also could you add something to the readme about this too?

vm:
I think it best to hold off on this for now. The other PR that forces `npx deduplicate` is probably gonna get more of the benefits regardless.

vm:
If we're gonna merge this now is a good time as all the NFT stuff has been merged in

vm (approved):

PR #4103, opened 2022-07-14
fix: update styled-components in package.json to latest to remove react invalid hook call warnings

lynnshaoyu:
Vig found invalid hook call warnings after react 18 upgrade PR (https://github.com/Uniswap/interface/pull/3992) : https://reactjs.org/warnings/invalid-hook-call-warning.html in console on local build. 

Fixed by updating styled-components to latest version, which apparently is needed by React 18. See thread here: https://github.com/facebook/react/issues/21848

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -41,8 +41,10 @@
     "prettier/@typescript-eslint",
     "plugin:prettier/recommended"
   ],
-  "plugins": ["simple-import-sort", "unused-imports"],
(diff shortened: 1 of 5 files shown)


vm:
I need to explain this somewhere, but you can run `npx yarn-deduplicate strategy=fewer` `npx yarn-deduplicate strategy=latest` to fix the yarn.lock without needing to do it manually

vm (approved):

PR #4104, opened 2022-07-14
feat: add hook for multi-network token balances

JFrankfurt:


JFrankfurt:
This also updates portions of the UI to support the new hook

src/components/Explore/BalanceSummary.tsx
```diff
@@ -63,32 +70,34 @@ export default function BalanceSummary({ address }: { address: string }) {
...
-          <NetworkBalance
-            logoUrl={logoUrl2}
-            balance={'3.3'}
-            tokenSymbol={tokenSymbol ?? 'XXX'}
-            fiatValue={3200}
-            label={'Polygon'}
-            networkColor={TEXT_COLORS['137']}
-          />
+          {data &&
+            Object.entries(data).map(([chainId, amount]) => {
```
vm, line 74: issue with this option is the keys won't be ordered, so it could change per load. might be better to iterate over a list of chainIds and look into the data object to see if there's anything there

src/components/Explore/BalanceSummary.tsx
```diff
@@ -63,32 +70,34 @@ export default function BalanceSummary({ address }: { address: string }) {
...
+              const fiatValue = 1010.12 // for testing purposes
+              const chain = parseInt(chainId)
+              if (!fiatValue || !isChainAllowed(connector, chain)) return null
```
vm, line 77: why do we need to check if the chain is allowed if there's data already?

JFrankfurt: defensive safety check since the UI doesn't control the data source. 

i.e., if the data service started returning Celo balances before the UI had a `chainInfo` for it things could explode

kayleegeorge (approved):

PR #4109, opened 2022-07-14
build: upgrade prettier to v2.7.1

vm:
This is prettier for all the files in the repo so we remove flakes.

Probably people are running a newer version locally of prettier than what we had on the repo. We should find out how to require a certain prettier version (or make it a dev dep?)


cbachmeier (approved):
LGTM

PR #4111, opened 2022-07-14
feat: implement connect wallet category events

lynnshaoyu:
1. Connect Wallet Button Clicked

2. Wallet Selected

3. Wallet Connected

### DIFF
--- a/src/pages/RemoveLiquidity/index.tsx
+++ b/src/pages/RemoveLiquidity/index.tsx
@@ -4,6 +4,8 @@ import { TransactionResponse } from '@ethersproject/providers'
 import { Trans } from '@lingui/macro'
 import { Currency, Percent } from '@uniswap/sdk-core'
 import { useWeb3React } from '@web3-react/core'
+import { ElementName, Event, EventName } from 'components/AmplitudeAnalytics/constants'
+import { TraceEvent } from 'components/AmplitudeAnalytics/TraceEvent'
 import { sendEvent } from 'components/analytics'
 import { useV2LiquidityTokenPermit } from 'hooks/useV2LiquidityTokenPermit'
 import { useCallback, useContext, useMemo, useState } from 'react'
@@ -625,9 +627,16 @@ export default function RemoveLiquidity({
             )}
             <div style={{ position: 'relative' }}>
               {!account ? (
-                <ButtonLight onClick={toggleWalletModal}>
-                  <Trans>Connect Wallet</Trans>
-                </ButtonLight>
+                <TraceEvent
+                  events={[Event.onClick]}
+                  name={EventName.CONNECT_WALLET_BUTTON_CLICKED}
+                  properties={{ received_swap_quote: false }}
+                  element={ElementName.CONNECT_WALLET_BUTTON}
+                >
+                  <ButtonLight onClick={toggleWalletModal}>
+                    <Trans>Connect Wallet</Trans>
+                  </ButtonLight>
+                </TraceEvent>
               ) : (
                 <RowBetween>
                   <ButtonConfirmed
--- a/src/components/AmplitudeAnalytics/constants.ts
+++ b/src/components/AmplitudeAnalytics/constants.ts
@@ -5,11 +5,14 @@
  * and logged.
  */
 export enum EventName {
+  CONNECT_WALLET_BUTTON_CLICKED = 'Connect Wallet Button Clicked',
   PAGE_VIEWED = 'Page Viewed',
   SWAP_SUBMITTED = 'Swap Submitted',
   TOKEN_IMPORTED = 'Token Imported',
   TOKEN_SELECTED = 'Token Selected',
   TOKEN_SELECTOR_OPENED = 'Token Selector Opened',
+  WALLET_CONNECTED = 'Wallet Connected',
+  WALLET_SELECTED = 'Wallet Selected',
   // alphabetize additional event names.
 }
 
@@ -49,9 +52,11 @@ export const enum ModalName {
 export const enum ElementName {
   COMMON_BASES_CURRENCY_BUTTON = 'common-bases-currency-button',
   CONFIRM_SWAP_BUTTON = 'confirm-swap-or-send',
+  CONNECT_WALLET_BUTTON = 'connect-wallet-button',
   IMPORT_TOKEN_BUTTON = 'import-token-button',
   SWAP_BUTTON = 'swap-button',
   TOKEN_SELECTOR_ROW = 'token-selector-row',
+  WALLET_TYPE_OPTION = 'wallet-type-option',
   // alphabetize additional element names.
 }
 
--- a/src/components/WalletModal/Option.tsx
+++ b/src/components/WalletModal/Option.tsx
@@ -1,3 +1,5 @@
+import { ElementName, Event, EventName } from 'components/AmplitudeAnalytics/constants'
+import { TraceEvent } from 'components/AmplitudeAnalytics/TraceEvent'
 import React from 'react'
 import styled from 'styled-components/macro'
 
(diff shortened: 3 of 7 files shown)


src/components/Header/index.tsx
```diff
@@ -245,8 +248,27 @@ const StyledExternalLink = styled(ExternalLink).attrs({
...
+      return PageName.POOL_PAGE
+    default:
+      return PageName.SWAP_PAGE
+  }
+}
+
 export default function Header() {
   const { account, chainId } = useWeb3React()
+  const location = useLocation()
+  const [currentPage, setCurrentPage] = useState(getCurrentPageFromLocation(location.pathname))
```
cmcewen, line 267: won't `useLocation` re-render? so couldn't you just do `currentPage = getCurrentPageFromLocation(location.pathname)`

src/components/WalletModal/index.tsx
```diff
@@ -150,6 +153,20 @@ export default function WalletModal({
     }
   }, [pendingConnector, walletView])
 
+  // When new wallet is successfully set by the user, trigger logging of Amplitude analytics event.
+  useEffect(() => {
```
cmcewen, line 157: i don't think it's ideal to have state/and effect just for logging but i can't quite think of how to avoid it for now

src/components/Web3Status/index.tsx
```diff
@@ -120,9 +127,22 @@ function Sock() {
...
+}
+
 function Web3StatusInner() {
   const { account, connector, chainId, ENSName } = useWeb3React()
   const connectionType = getConnection(connector).type
+  const {
+    trade: { state: tradeState, trade },
+    inputError: swapInputError,
+  } = useDerivedSwapInfo()
+  const [validSwapQuote, setValidSwapQuote] = useState(getIsValidSwapQuote(trade, tradeState, swapInputError))
```
cmcewen, line 145: you don't need state here - can just use a local variable

src/components/WalletModal/index.tsx
lynnshaoyu: yeah same, spoke w vig at length about this, i think based off of how web3react state/redux is set up this is the best way we've got for now.

src/components/Header/index.tsx
lynnshaoyu: yes you're totally right. done.

src/components/Header/index.tsx
```diff
@@ -245,8 +247,23 @@ const StyledExternalLink = styled(ExternalLink).attrs({
...
 
+function getCurrentPageFromLocation(locationPathname: string): PageName {
+  switch (locationPathname) {
+    case '/swap':
+      return PageName.SWAP_PAGE
+    case '/vote':
+      return PageName.VOTE_PAGE
+    case '/pool':
+      return PageName.POOL_PAGE
+    default:
```
cmcewen, line 258: should this default to unknown or undefined?

cmcewen: or even log an error?

lynnshaoyu: undefined seems reasonable, error could be misleading i think.

src/components/Header/index.tsx
```diff
@@ -245,8 +248,27 @@ const StyledExternalLink = styled(ExternalLink).attrs({
...
```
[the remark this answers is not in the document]
lynnshaoyu: yep connor said the same, changing.

src/components/WalletModal/index.tsx
```diff
@@ -150,6 +153,20 @@ export default function WalletModal({
...
+    if (!!account && account !== lastActiveWalletAddress) {
+      sendAnalyticsEvent(EventName.WALLET_CONNECTED, {
+        result: WALLET_CONNECTION_RESULT.SUCCEEDED,
+        wallet_address: account,
+        wallet_type: getConnectionName(getConnection(connector).type, getIsMetaMask()),
+        // TODO(lynnshaoyu): Send correct is_reconnect value after modifying user state.
+        is_reconnect: false,
```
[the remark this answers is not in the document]
lynnshaoyu: yeah that sounds good.

src/components/WalletModal/index.tsx
```diff
@@ -178,6 +195,13 @@ export default function WalletModal({
       } catch (error) {
         console.debug(`web3-react connection error: ${error}`)
         dispatch(updateConnectionError({ connectionType, error: error.message }))
+
+        // Wallect connection has failed at this point, send failed WALLET_CONNECTED
```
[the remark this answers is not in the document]
lynnshaoyu: yeah fair enough

src/components/WalletModal/index.tsx
```diff
@@ -150,6 +153,20 @@ export default function WalletModal({
     }
   }, [pendingConnector, walletView])
 
+  // When new wallet is successfully set by the user, trigger logging of Amplitude analytics event.
+  useEffect(() => {
+    if (!!account && account !== lastActiveWalletAddress) {
```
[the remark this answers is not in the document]
lynnshaoyu: sgtm

src/components/Header/index.tsx
```diff
@@ -245,8 +248,27 @@ const StyledExternalLink = styled(ExternalLink).attrs({
...
+      return PageName.SWAP_PAGE
+  }
+}
+
 export default function Header() {
   const { account, chainId } = useWeb3React()
+  const location = useLocation()
+  const [currentPage, setCurrentPage] = useState(getCurrentPageFromLocation(location.pathname))
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
lynnshaoyu: already been removed in latest commit

src/components/WalletModal/index.tsx
```diff
@@ -150,6 +153,20 @@ export default function WalletModal({
...
+      sendAnalyticsEvent(EventName.WALLET_CONNECTED, {
+        result: WALLET_CONNECTION_RESULT.SUCCEEDED,
+        wallet_address: account,
+        wallet_type: getConnectionName(getConnection(connector).type, getIsMetaMask()),
```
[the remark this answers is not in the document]
lynnshaoyu: yeah that would be way helpful

src/components/WalletModal/index.tsx
```diff
@@ -178,6 +195,13 @@ export default function WalletModal({
       } catch (error) {
         console.debug(`web3-react connection error: ${error}`)
         dispatch(updateConnectionError({ connectionType, error: error.message }))
+
+        // Wallect connection has failed at this point, send failed WALLET_CONNECTED
+        // Amplitude analytics event.
+        sendAnalyticsEvent(EventName.WALLET_CONNECTED, {
```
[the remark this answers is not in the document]
lynnshaoyu: agree, changed to WALLET_CONNECT_TXN_COMPLETED. im against separate event because i think we should make things properties on events where possible (reduce volume of events).

src/components/WalletModal/index.tsx
```diff
@@ -121,9 +123,10 @@ export default function WalletModal({
   ENSName?: string
 }) {
   const dispatch = useAppDispatch()
-  const { account } = useWeb3React()
+  const { connector, account } = useWeb3React()
 
   const [walletView, setWalletView] = useState(WALLET_VIEWS.ACCOUNT)
+  const [lastActiveWalletAddress, setLastActiveWalletAddress] = useState<string | undefined>(account)
```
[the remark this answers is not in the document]
lynnshaoyu: i googled but seems like we need to define another custom hook usePrevious - I think this is cleaner / easier to understand?

src/components/Header/index.tsx
```diff
@@ -333,7 +350,9 @@ export default function Header() {
                 </Trans>
               </BalanceText>
             ) : null}
-            <Web3Status />
+            <Trace page={currentPage}>
```
vm, line 353: i think i'm not following why this page trace lives in the page header, vs having it live in `App.tsx`

lynnshaoyu: yeah good callout, seems reasonable to me to move it to app.tsx. done.

src/pages/RemoveLiquidity/index.tsx
```diff
@@ -625,9 +627,16 @@ export default function RemoveLiquidity({
             )}
             <div style={{ position: 'relative' }}>
               {!account ? (
-                <ButtonLight onClick={toggleWalletModal}>
-                  <Trans>Connect Wallet</Trans>
-                </ButtonLight>
+                <TraceEvent
+                  events={[Event.onClick]}
+                  name={EventName.CONNECT_WALLET_BUTTON_CLICKED}
+                  properties={{ received_swap_quote: false }}
```
vm, line 633: i don' think you need to have `received_swap_quote` as false everywhere, could it just be undefined? no strong opinion though

cmcewen (commented):
looks good except agree on vig's comment on reusing logic

src/components/Web3Status/index.tsx
```diff
@@ -120,9 +127,22 @@ function Sock() {
   )
 }
 
+function getIsValidSwapQuote(
```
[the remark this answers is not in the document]
lynnshaoyu: yeah good call, seems fair to use the same function and logic. moved it over to swap/index.tsx.

src/pages/RemoveLiquidity/index.tsx
lynnshaoyu: i prefer being explicit here too, even in cases like here where it is not possible for received_swap_quote to ever be true.

cmcewen (approved):

PR #4115, opened 2022-07-16
chore: upgrades react-router-dom, fixes dev-mode linking

chenxsan:
I'm not sure if anyone noticed that the `<Link />` component is broken now under development mode.

E.g., 

CleanShot 2022-07-16 at 14 45 58@2x

It seems to be incompatibility between React 18 StrictMode and React router dom V5, see https://github.com/remix-run/react-router/issues/7870#issuecomment-1085478601 for an example.

And I believe they have fixed it in https://github.com/remix-run/react-router/pull/8831.

In this pull request:

1. updated `react-router-dom` to the latest V5 version to fix the bug in development mode.
2. added a e2e test to ensure the `Link` component works as expected.

Btw, let me know if you want to upgrade `react-router-dom` to v6, I think I can help with a pull request.

### DIFF
--- a/src/pages/Pool/PositionPage.tsx
+++ b/src/pages/Pool/PositionPage.tsx
@@ -588,7 +588,11 @@ export function PositionPage({
           />
           <AutoColumn gap="md">
             <AutoColumn gap="sm">
-              <Link style={{ textDecoration: 'none', width: 'fit-content', marginBottom: '0.5rem' }} to="/pool">
+              <Link
+                data-cy="visit-pool"
+                style={{ textDecoration: 'none', width: 'fit-content', marginBottom: '0.5rem' }}
+                to="/pool"
+              >
                 <HoverText>
                   <Trans>← Back to Pools Overview</Trans>
                 </HoverText>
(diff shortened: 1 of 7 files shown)


src/pages/Pool/PositionPage.tsx
```diff
@@ -588,7 +588,11 @@ export function PositionPage({
           />
           <AutoColumn gap="md">
             <AutoColumn gap="sm">
-              <Link style={{ textDecoration: 'none', width: 'fit-content', marginBottom: '0.5rem' }} to="/pool">
+              <Link
+                data-cy="visit-pool"
```
chenxsan, line 592: See https://docs.cypress.io/guides/references/best-practices,

> The data-cy attribute will not change from CSS style or JS behavioral changes, meaning it's not coupled to the behavior or styling of an element.

vm (approved):

lynnshaoyu:
thank you so so much @chenxsan for fixing this!! really appreciate this help.

vm:
@chenxsan if you're interested in upgrading to v6 that would be helpful for sure. we do use the HashRouter so I hadn't looked into how that will work with the upgrade yet

PR #4119, opened 2022-07-17
fix: fixes Popover arrow positioning

chenxsan:
If we zoom in, we can find the arrow not really an arrow:

CleanShot 2022-07-17 at 15 26 55@2x

CleanShot 2022-07-17 at 15 29 50@2x

### DIFF
--- a/src/components/Popover/index.tsx
+++ b/src/components/Popover/index.tsx
@@ -26,6 +26,7 @@ const Arrow = styled.div`
     position: absolute;
     width: 8px;
     height: 8px;
+    box-sizing: border-box;
     z-index: 9998;
 
     content: '';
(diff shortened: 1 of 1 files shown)


src/components/Popover/index.tsx
```diff
@@ -26,6 +26,7 @@ const Arrow = styled.div`
     position: absolute;
     width: 8px;
     height: 8px;
+    box-sizing: border-box;
```
chenxsan, line 29: Much easier to reason about, otherwise we would have a 9x9 box (because it has two borders) inside a 8x8 box.

chenxsan:
And here's the fixed version:
IMG_BAE9EC4BAD9D-1

vm (approved):

PR #4121, opened 2022-07-18
chore: access router data with hooks

chenxsan:
`react-router-dom` [v5.1.0](https://github.com/remix-run/react-router/releases/tag/v5.1.0) introduced some new hooks to access router data so we no longer need to pass props like `match`, `history`, etc. down to child components.

Besides, [it's advised](https://reacttraining.com/blog/react-router-v5-1/) to use `<Route children>` instead of `<Route component>` or `<Route render>`.

Here're what this pull request includes:

1. Update `<Route component>` or `<Route render>` to `<Route children>`
2. use hooks like `useParams`, `useHistory` etc. to access router data

And it should pave the way to upgrade to `react-router-dom` v6.


src/pages/App.tsx
```diff
@@ -92,45 +96,76 @@ export default function App() {
...
-
-              <Route exact strict path="/add/v2/:currencyIdA?/:currencyIdB?" component={RedirectDuplicateTokenIdsV2} />
               <Route
                 exact
                 strict
-                path="/add/:currencyIdA?/:currencyIdB?/:feeAmount?"
-                component={RedirectDuplicateTokenIds}
-              />
+                path="/create-proposal"
+                render={() => <Redirect to="/vote/create-proposal" />}
```
vm, line 106: it's not possible to redirect as a child anymore?

vm (approved):
I will test this and merge it later today, thank you for the work! Looks great.

PR #4124, opened 2022-07-18
feat: enable 1bp optimism fee tier

NoahZinsmeister:
compare to https://github.com/Uniswap/interface/pull/3724/files, the polygon version of this pr

### DIFF
--- a/src/components/FeeSelector/shared.tsx
+++ b/src/components/FeeSelector/shared.tsx
@@ -1,7 +1,7 @@
 import { Trans } from '@lingui/macro'
 import { FeeAmount } from '@uniswap/v3-sdk'
 import { ALL_SUPPORTED_CHAIN_IDS, SupportedChainId } from 'constants/chains'
-import { ReactNode } from 'react'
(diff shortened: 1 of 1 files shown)


src/components/FeeSelector/shared.tsx
```diff
@@ -1,7 +1,7 @@
 import { Trans } from '@lingui/macro'
 import { FeeAmount } from '@uniswap/v3-sdk'
 import { ALL_SUPPORTED_CHAIN_IDS, SupportedChainId } from 'constants/chains'
-import { ReactNode } from 'react'
+import type { ReactNode } from 'react'
```
leggechr, line 4: revert this line?

leggechr (approved):
small nit + the PR title needs to be updated

src/components/FeeSelector/shared.tsx
vm: i feel like we don't use this syntax elsewhere so probably prefer to avoid fwiw

PR #4132, opened 2022-07-18
build: lock jest-styled-components@7.0.7

vm:
Seems like this PR made some changes in the lock file: https://github.com/Uniswap/interface/pull/4128

Then this PR failed the unit-tests based on snapshots: https://github.com/Uniswap/interface/pull/4125

Then this PR updated the snapshots as an unrelated issue: https://github.com/Uniswap/interface/pull/4130

This issue states that 7.0.8 (latest version) has a bug where the sc-* classNames from styled-components are regenerated randomly, leading the snapshot flakes: https://github.com/styled-components/jest-styled-components/issues/405, https://github.com/styled-components/jest-styled-components/issues/276#issuecomment-1149749113

So this re-adds some potentially necessary pieces of the yarn.lock, and also fixes the jest-styled-components version for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "eslint-plugin-react-hooks": "^4.0.0",
     "eslint-plugin-simple-import-sort": "^7.0.0",
     "eslint-plugin-unused-imports": "^2.0.0",
-    "jest-styled-components": "^7.0.5",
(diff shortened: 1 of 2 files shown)


cmcewen (approved):
woot thanks for looking into this

PR #4135, opened 2022-07-19
test: fix swap test flake

vm:


cypress/e2e/swap.test.ts
```diff
@@ -13,7 +13,7 @@ describe('Swap', () => {
   it('can enter an amount into input', () => {
     cy.get('#swap-currency-input .token-amount-input')
       .clear()
-      .type('0.001', { delay: 200 })
+      .type('0.001', { delay: 1000 })
```
vm, line 16: we probably should figure out why the page reloads but i'm just deciding to move on. i'll add a todo tho

vm:
smh it didn't work

cypress/e2e/swap.test.ts
```diff
@@ -11,34 +11,31 @@ describe('Swap', () => {
   })
 
   it('can enter an amount into input', () => {
-    cy.get('#swap-currency-input .token-amount-input')
-      .clear()
-      .type('0.001', { delay: 200 })
-      .should('have.value', '0.001')
+    cy.get('#swap-currency-input .token-amount-input').clear().type('0.001').should('have.value', '0.001')
```
vm, line 14: in a separate pr i will move all these selectors to `data-cy` rather than using ids and classnames

cypress/e2e/swap.test.ts
```diff
@@ -11,34 +11,31 @@ describe('Swap', () => {
...
   })
 
   it('zero swap amount', () => {
-    cy.get('#swap-currency-input .token-amount-input').clear().type('0.0', { delay: 200 }).should('have.value', '0.0')
+    cy.get('#swap-currency-input .token-amount-input').clear().type('0.0').should('have.value', '0.0')
   })
 
   it('invalid swap amount', () => {
-    cy.get('#swap-currency-input .token-amount-input').clear().type('\\', { delay: 200 }).should('have.value', '')
```
vm, line 25: this allows us to avoid the delays, which is not something tests should have

cypress/e2e/swap.test.ts
```diff
@@ -11,34 +11,31 @@ describe('Swap', () => {
...
 
   it('zero output amount', () => {
-    cy.get('#swap-currency-output .token-amount-input').type('0.0', { delay: 200 }).should('have.value', '0.0')
+    cy.get('#swap-currency-output .token-amount-input').clear().type('0.0').should('have.value', '0.0')
   })
 
   it.skip('can swap ETH for DAI', () => {
     cy.get('#swap-currency-output .open-currency-select-button').click()
     cy.get('.token-item-0xc7AD46e0b8a400Bb3C915120d284AafbA8fc4735').should('be.visible')
-    cy.get('.token-item-0xc7AD46e0b8a400Bb3C915120d284AafbA8fc4735').click({ force: true })
```
vm, line 39: also can skip all the force clicks which also should be avoided

cypress/e2e/swap.test.ts
```diff
@@ -11,73 +11,43 @@ describe('Swap', () => {
...
   it('ETH to wETH is same value (wrapped swaps have no price impact)', () => {
     cy.get('#swap-currency-output .open-currency-select-button').click()
-    cy.get('.token-item-0xc778417E063141139Fce010982780140Aa0cD5Ab').click({ force: true })
-    cy.get('#swap-currency-input .token-amount-input').type('0.01', { force: true, delay: 100 })
+    cy.get('.token-item-0xc778417E063141139Fce010982780140Aa0cD5Ab').click()
+    cy.get('#swap-currency-input .token-amount-input').clear().type('0.01')
     cy.get('#swap-currency-output .token-amount-input').should('have.value', '0.01')
   })
-
-  describe('expert mode', () => {
```
vm, line 58: these don't seem to be on right now and the UI is different enough that i didn't wanna work on this. we can re-write these back later if needed

PR #4138, opened 2022-07-19
style: updating ui on unsupported network

cartcrom:
Implements some of the UX changes requested [here](https://github.com/Uniswap/interface/issues/2627#issuecomment-999154401) while staying in the current UI theme/language

* Moved unsupported network message from web3status component to NetworkSelector component, since the message should also dropdown the network selector
* Disabled token selection, numeric input, arrow, and settings buttons when on unsupported network
* implemented isSupportedChain to use in some places instead of isChainAllowed since isChainAllowed has extra logic for checking if a switch is allowed
  * there is also isSupportedChainId in clientSideSmartOrderRouter... not sure if these should be combined or kept separate


cartcrom:
> code looks good - please show the preview to a designer before merging

pinging Cal 👍

src/components/RoutingDiagram/__snapshots__/RoutingDiagram.test.tsx.snap
```diff
@@ -3,45 +3,45 @@
 exports[`renders multi route 1`] = `
```
vm, line 3: wait why did these change?

cartcrom: I regenerated the snapshots because the html class names had just barely changed... should that not have happened?

vm: it shouldn't have. is this branch up to date against main?

vm: and are your local node_modules up to date?

cartcrom: Oh didn't realize that'd affect them thank you, I've been working in another branch and switched back without yarn installing. I'll reinstall, checkout the original snaps and make sure they run

PR #4139, opened 2022-07-19
feat: add new colors to explore page

kayleegeorge:
Added updated color palette & changed in Explore 

Screen Shot 2022-07-21 at 1 18 40 PM


vm:
before this PR, i would first make a PR which replaces all the deprecated variable use with `deprecrated_blue200` etc

vm (approved):

PR #4141, opened 2022-07-19
feat: add updated theme colors

kayleegeorge:
add updated theme colors from new design spec


src/theme/index.tsx
```diff
@@ -110,6 +111,59 @@ function colors(darkMode: boolean): Colors {
     // dont wanna forget these blue yet
     deprecated_blue4: darkMode ? '#153d6f70' : '#C4D9F8',
     // blue5: darkMode ? '#153d6f70' : '#EBF4FF',
+
+    blue200: ColorsPalette.blue200,
+
+    /* UPDATED */
```
vm, line 117: what is this comment referring to?

src/theme/styled.d.ts
```diff
@@ -57,6 +57,58 @@ export interface Colors {
   deprecated_error: Color
   deprecated_success: Color
   deprecated_warning: Color
+
+  /* UPDATED PALETTE */
```
vm, line 61: same question

vm: does this need to get added here? it looks like the colors are only added to `Palette`?

src/theme/index.tsx
kayleegeorge: oh it's just the Updated colors haha, I can take out since the other colors have deprecated now

src/theme/colors.ts
```diff
@@ -0,0 +1,320 @@
+// Based mostly on https://github.com/Uniswap/interface/blob/main/src/theme/index.tsx
+
+import { opacify } from './colorsUtil'
+
+export interface GlobalPalette {
+  white: string
```
vm, line 6: can you add a type for Color and then change all the string to Color? similar to the deprecated version

src/theme/colorsUtil.tsx
```diff
@@ -0,0 +1,23 @@
+/**
```
vm, line 1: name file `utils.tsx`

vm (approved):

PR #4147, opened 2022-07-20
fix: add celo gas override

Jesse-Sawa:
This circumvents the 'out of gas' error thrown by the multicall


vm:
@Jesse-Sawa can you rebase against main? i also changed this file earlier

vm (approved):

PR #4151, opened 2022-07-20
feat: implement other swap events part 1

lynnshaoyu:
Added "shouldLogImpression" property, defaulting to true, onto TraceEvent component as well, because some events require conditional logging. For example, we want to log when the auto router visualization is opened, but not send the same event again when it is closed.

Events created by this PR: 
1. Screen Shot 2022-07-20 at 4 58 24 PM
2. Screen Shot 2022-07-20 at 4 58 32 PM
3. Screen Shot 2022-07-20 at 4 58 38 PM
4. Screen Shot 2022-07-20 at 9 51 40 PM

### DIFF
--- a/src/components/swap/SwapDetailsDropdown.tsx
+++ b/src/components/swap/SwapDetailsDropdown.tsx
@@ -139,6 +139,7 @@ export default function SwapDetailsDropdown({
           events={[Event.onClick]}
           name={EventName.SWAP_DETAILS_EXPANDED}
           element={ElementName.SWAP_DETAILS_DROPDOWN}
+          shouldLogImpression={!showDetails}
         >
           <StyledHeaderRow onClick={() => setShowDetails(!showDetails)} disabled={!trade} open={showDetails}>
             <RowFixed style={{ position: 'relative' }}>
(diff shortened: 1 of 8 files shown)


src/components/swap/SwapDetailsDropdown.tsx
```diff
@@ -133,69 +135,76 @@ export default function SwapDetailsDropdown({
...
-                        />
-                      </Card>
-                    </ResponsiveTooltipContainer>
-                  }
-                  placement="bottom"
+        <TraceEvent
+          events={[Event.onClick]}
+          name={EventName.SWAP_DETAILS_EXPANDED}
+          element={ElementName.SWAP_DETAILS_DROPDOWN}
+          shouldLogImpression={!showDetails}
```
vm, line 142: why does an event trace need to know whether to log impression?

lynnshaoyu: answered in PR description: "some events require conditional logging. For example, we want to log when the auto router visualization is opened, but not send the same event again when it is closed."

vm (approved):

PR #4154, opened 2022-07-20
build: only test for highest yarn-deduplicate strategy

vm:
Attempting to run both `fewer` and `highest` strategies creates conflicts. We had it so we manually removed specific packages from being tested, which is super fragile.


cbachmeier (approved):
LGTM

PR #4165, opened 2022-07-21
feat: implement swap quote received event

lynnshaoyu:
Screen Shot 2022-07-21 at 6 47 53 PM

also this issue: https://uniswaplabs.atlassian.net/browse/WEB-623?atlOrigin=eyJpIjoiODdlZTM3MDY2MGM5NDZhMDg4ZmVkZjhlYjI1NmY3YTYiLCJwIjoiaiJ9


src/components/swap/SwapDetailsDropdown.tsx
```diff
@@ -11,15 +12,19 @@ import Row, { RowBetween, RowFixed } from 'components/Row'
...
 import styled, { keyframes, useTheme } from 'styled-components/macro'
 import { HideSmall, ThemedText } from 'theme'
+import { computeRealizedLPFeePercent } from 'utils/prices'
 
 import { AdvancedSwapDetails } from './AdvancedSwapDetails'
+import { getPriceImpactPercent } from './AdvancedSwapDetails'
 import GasEstimateBadge from './GasEstimateBadge'
 import { ResponsiveTooltipContainer } from './styleds'
+import { formatPercentInBasisPointsNumber } from './SwapModalFooter'
+import { getNumberFormattedToDecimalPlace } from './SwapModalFooter'
```
cmcewen, line 27: could you move these two functions to a utils file? maybe in the analytics folder? also you should be able to import in one line - VSCode has an "organize imports" setting which makes it nicer

src/components/swap/SwapDetailsDropdown.tsx
```diff
@@ -131,6 +159,12 @@ export default function SwapDetailsDropdown({
   const theme = useTheme()
   const { chainId } = useWeb3React()
   const [showDetails, setShowDetails] = useState(false)
+  const [isFirstPriceFetch, setIsFirstPriceFetch] = useState(true)
+  const lpFeePercent = trade ? computeRealizedLPFeePercent(trade) : undefined
```
cmcewen, line 163: if this is computed from trade, can't this be computed in `formatAnalyticsEventProperties`?

src/components/swap/SwapDetailsDropdown.tsx
```diff
@@ -131,6 +159,12 @@ export default function SwapDetailsDropdown({
...
+
+  useEffect(() => {
```
cmcewen, line 165: again can't think of a better way to do this right now but i do see we are mixing logging concerns with react state, which i think should be separate

src/components/swap/SwapDetailsDropdown.tsx
lynnshaoyu: sounds good done

src/components/swap/SwapDetailsDropdown.tsx
lynnshaoyu: yes! sorry mistakenly thought computeRealizedLPFeePercent was a custom react hook and had to be used in the component (but it's not). moved the calc to formatAnalyticsEventProperties.

src/components/swap/SwapDetailsDropdown.tsx
lynnshaoyu: hmm yeah this was my best attempt at a decent way to capture this, can ask for @zzmp's input too when he's back.

cmcewen (approved):
LGTM!

PR #4166, opened 2022-07-21
feat: different table states

kayleegeorge:
add error table state and no token table state

Screen Shot 2022-07-21 at 3 38 40 PM

Screen Shot 2022-07-21 at 3 32 51 PM

### DIFF
--- a/src/components/Explore/TokenTable.tsx
+++ b/src/components/Explore/TokenTable.tsx
@@ -1,6 +1,7 @@
 import { useAllTokens } from 'hooks/Tokens'
 import useTopTokens, { TimePeriod } from 'hooks/useTopTokens'
 import { useAtom, useAtomValue } from 'jotai'
+import { useMemo } from 'react'
 import { AlertTriangle } from 'react-feather'
 import styled from 'styled-components/macro'
 
@@ -43,13 +44,31 @@ const LOADING_ROWS = Array(10)
     return <LoadingRow key={`${index}`} />
   })
 
+function useFilteredTokens(shownTokens: string[]) {
+  const filterString = useAtomValue(filterStringAtom)
+  const allTokens = useAllTokens()
+  const filteredTokens = shownTokens.filter((tokenAddress) => {
+    const token = allTokens[tokenAddress]
+    const tokenName = token?.name ?? ''
+    const tokenSymbol = token?.symbol ?? ''
+
+    if (!filterString) {
+      return true
+    }
+    const lowercaseFilterString = filterString.toLowerCase()
+    const addressIncludesFilterString = tokenAddress.toLowerCase().includes(lowercaseFilterString)
+    const nameIncludesFilterString = tokenName.toLowerCase().includes(lowercaseFilterString)
+    const symbolIncludesFilterString = tokenSymbol.toLowerCase().includes(lowercaseFilterString)
+    return nameIncludesFilterString || symbolIncludesFilterString || addressIncludesFilterString
+  })
+  return useMemo(() => filteredTokens, [filteredTokens])
+}
+
 export default function TokenTable() {
   const { data, error, loading } = useTopTokens()
   const [favoriteTokens] = useAtom(favoritesAtom)
   const [showFavorites] = useAtom(showFavoritesAtom)
-  const filterString = useAtomValue(filterStringAtom)
   const timePeriod = TimePeriod.day
-  const allTokens = useAllTokens()
 
   /* loading and error state */
   if (loading) {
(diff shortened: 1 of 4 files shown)


src/components/Explore/TokenTable.tsx
```diff
@@ -53,20 +79,30 @@ export default function TokenTable() {
...
         <HeaderRow timeframe={timePeriod} />
-        <NoTokenDisplay>No Favorited Tokens</NoTokenDisplay>
+        <NoTokenDisplay>You have no favorited tokens</NoTokenDisplay>
       </GridContainer>
     )
   }
 
   const topTokenAddresses = Object.keys(data)
   const showTokens = showFavorites ? favoriteTokens : topTokenAddresses
+  const filteredTokens = useFilteredTokens(showTokens)
```
kayleegeorge, line 104: @vm conditionally called here because we want to exit earlier if the data is null

vm: you need to put this at the top before you early exit, and have that hook return `[]`

vm: you also shouldn't pass `showTokens` in, call that hook inside `useFilteredTokens`

kayleegeorge: ay yai captain

src/components/Explore/TokenTable.tsx
```diff
@@ -38,6 +44,26 @@ const LOADING_ROWS = Array(10)
     return <LoadingRow key={`${index}`} />
   })
 
+function useFilteredTokens(shownTokens: string[]) {
+  const filterString = useAtomValue(filterStringAtom)
+  const allTokens = useAllTokens()
+  const filteredTokens = shownTokens.filter((tokenAddress) => {
```
vm, line 50:
```
return useMemo(() => {
   ... all the `filteredTokens` logic here
}, [filterString, allTokens]
```

src/components/Explore/TokenTable.tsx
```diff
@@ -38,11 +44,46 @@ const LOADING_ROWS = Array(10)
     return <LoadingRow key={`${index}`} />
   })
 
+function useFilteredTokens(shownTokens: string[]) {
+  const filterString = useAtomValue(filterStringAtom)
+  const allTokens = useAllTokens()
+
+  return useMemo(() => {
+    const filteredTokens = shownTokens.filter((tokenAddress) => {
```
vm, line 52: use `allTokens` here i think?

src/components/Explore/TokenTable.tsx
```diff
@@ -38,11 +44,47 @@ const LOADING_ROWS = Array(10)
...
+      const addressIncludesFilterString = tokenAddress.toLowerCase().includes(lowercaseFilterString)
+      const nameIncludesFilterString = tokenName.toLowerCase().includes(lowercaseFilterString)
+      const symbolIncludesFilterString = tokenSymbol.toLowerCase().includes(lowercaseFilterString)
+      return nameIncludesFilterString || symbolIncludesFilterString || addressIncludesFilterString
+    })
+    return filteredTokens
+  }, [allTokens, shownTokens, filterString])
+}
+
+function NoTokensState(message: ReactNode, timePeriod: TimePeriod) {
```
vm, line 73: this should be a component

src/hooks/useTopTokens.ts
```diff
@@ -283,7 +283,7 @@ const FAKE_TOP_TOKENS_RESULT = {
 }
 
 const useTopTokens = (): UseTopTokensResult => {
-  const [data, setData] = useState<TokenData | null>(null)
+  const [data, setData] = useState<TokenData>({})
```
vm, line 286: this seems non-ideal, why this change?

kayleegeorge: oh i thought you told me to do this LOL

kayleegeorge: but i understand now ok i will change it back

src/components/Explore/TokenTable.tsx
kayleegeorge: but it uses all the same components as the table grid (since it still maintains the header and overall grid container). so I would need to export those styles which seems slightly redundant

kayleegeorge: like this is technical the table component, just a different token display (instead of token rows, it shows none)

vm: I just mean you should do `NoTokensState message={} timeperiod={}`

kayleegeorge: OH meme my b

vm (approved):

PR #4167, opened 2022-07-21
feat: remove fortmatic

vm:
This should significantly decrease our app's load time.


src/utils/switchChain.ts
```diff
@@ -41,19 +34,8 @@ function getRpcUrls(chainId: SupportedChainId): [string] {
...
-    case injectedConnection.connector:
-    case coinbaseWalletConnection.connector:
-    case walletConnectConnection.connector:
-    case networkConnection.connector:
-    case gnosisSafeConnection.connector:
-      return ALL_SUPPORTED_CHAIN_IDS.includes(chainId)
-    default:
-      return false
-  }
+export function isChainAllowed(_connector: Connector, _chainId: number) {
```
vm, line 37: rm this function now

src/hooks/useEagerlyConnect.ts
```diff
@@ -18,15 +19,26 @@ async function connect(connector: Connector) {
...
+  const dispatch = useAppDispatch()
+
   const selectedWalletBackfilled = useAppSelector((state) => state.user.selectedWalletBackfilled)
   const selectedWallet = useAppSelector((state) => state.user.selectedWallet)
 
+  let selectedConnection: Connection | undefined
+  if (selectedWallet) {
+    try {
+      selectedConnection = getConnection(selectedWallet)
+    } catch {
```
cmcewen, line 31: this is unrelated to the fortmatic removal right? just another improvement

src/lib/hooks/useCurrency.ts
```diff
@@ -105,8 +102,7 @@ export function useCurrencyFromMap(tokens: TokenMap, currencyId?: string | null)
 
   const token = useTokenFromMapOrNetwork(tokens, isNative ? undefined : shorthandMatchAddress ?? currencyId)
 
-  const chainAllowed = chainId && isChainAllowed(connector, chainId)
-  if (currencyId === null || currencyId === undefined || !chainAllowed) return null
+  if (currencyId === null || currencyId === undefined) return null
```
cmcewen, line 105: this also seems unrelated? would be nice to keep the diff clean

src/hooks/useEagerlyConnect.ts
vm: it is related because if FORTMATIC is the saved option this all will error out

src/lib/hooks/useCurrency.ts
vm: i could have the function just return `true` and leave all this if we wanted to

vm: reverted

vm: this actually was a bug, i needed to check ALLOWED_CHAIN_IDS.

vm: i'd like to rename this to `isChainSupported` separately also

cmcewen (approved):
LGTM

vm:
Gonna hold off on this until later because need to write up a doc explaining it for the team

zzmp (approved):

PR #4168, opened 2022-07-21
fix: fix mobile padding on table and show header

kayleegeorge:
Mobile padding and responsiveness

Screen Shot 2022-07-21 at 6 33 03 PM


cmcewen (approved):
this looks good to me but just testing, the page looks a lil funky starting at around 390px so we need to confirm what our minimum screen size supported is

PR #4169, opened 2022-07-22
feat: remaining swap events

lynnshaoyu:
1. Swap Price Update Acknowledged: 
Screen Shot 2022-07-22 at 3 32 44 PM
Screen Shot 2022-07-22 at 3 36 21 PM
4. Swap Transaction Completed: 
Screen Shot 2022-07-22 at 6 05 56 PM


src/components/swap/SwapModalHeader.tsx
```diff
@@ -38,14 +42,42 @@ const ArrowWrapper = styled.div`
...
+  shouldLogModalCloseEvent,
+  setShouldLogModalCloseEvent,
   allowedSlippage,
   recipient,
   showAcceptChanges,
   onAcceptChanges,
 }: {
   trade: InterfaceTrade<Currency, Currency, TradeType>
+  shouldLogModalCloseEvent: boolean
+  setShouldLogModalCloseEvent: (arg0: boolean) => void
```
cmcewen, line 80: nit but nice to name the function types with variables you might use - `shouldLog` vs `arg0`

src/components/swap/SwapModalHeader.tsx
```diff
@@ -54,10 +86,29 @@ export default function SwapModalHeader({
...
+  }, [lastExecutionPrice, setLastExecutionPrice, trade.executionPrice])
+
+  useEffect(() => {
+    if (shouldLogModalCloseEvent && showAcceptChanges)
+      sendAnalyticsEvent(
+        EventName.SWAP_PRICE_UPDATE_ACKNOWLEDGED,
+        formatAnalyticsEventProperties(trade, priceUpdate, SWAP_PRICE_UPDATE_USER_RESPONSE.REJECTED)
+      )
+    setShouldLogModalCloseEvent(false)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
cmcewen, line 109: why do you add this disable? would making `setShouldLogModalCloseEvent` a dep fix this?

src/state/transactions/reducer.ts
```diff
@@ -46,6 +49,12 @@ const transactionSlice = createSlice({
         return
       }
       tx.receipt = receipt
+      if (tx.info.type === TransactionType.SWAP) {
```
cmcewen, line 52: you generally try to avoid mutations in the reducer - they should be pure functions. i think the Updater component would be a better home for this

src/components/swap/SwapModalHeader.tsx
lynnshaoyu: sgtm done

src/components/swap/SwapModalHeader.tsx
lynnshaoyu: yeah added the other dependencies, not really a good reason to exclude, i was debugging initially. ty for catching.

src/state/transactions/reducer.ts
lynnshaoyu: sounds good, thanks for the feedback. i assume you meant interface/src/state/transactions/updater.tsx and not interface/src/lib/hooks/transactions/updater.tsx ? moved it over to the former and verified it works the same on my local.

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -39,22 +39,30 @@ export default function ConfirmSwapModal({
   swapErrorMessage: ReactNode | undefined
   onDismiss: () => void
 }) {
+  const [shouldLogModalCloseEvent, setShouldLogModalCloseEvent] = useState(false)
```
vm, line 42: i don't think i'm following the logic here

vm: just add a comment for what `shouldLogModalCloseEvent` means and why it needs to be passed into a child. i did end up figuring out what's going on.

vm (approved):
Approved based on adding a comment

src/components/swap/ConfirmSwapModal.tsx
lynnshaoyu: sgtm, added.

Issue #3843, opened 2022-05-15
Full range Button only works when token 0 is selected


Rachel-Eichenberger:
**Bug Description**
Currently when trying to make a full range position the UI will only set the full range when the token direction is selected with token 0.

**Steps to Reproduce**

1. Go to https://app.uniswap.org/#/add/ETH/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48/3000?chain=mainnet&minPrice=1040.312550&maxPrice=4161.250200
(random pool)
2. Click on Full range button, UI will not fill in full range
3. Click on USDC top right of add lq window, and now UI will fill in full range values
   ...

**Expected Behavior**
UI would allow full range whichever token direction is selected

The UI would use max low tix and max high tick in proper token0 -> token1 position (my though thinking ui is putting upper/lower tick in wrong position, causing UI to bug out and not set at all

**Additional Context**
clicking USDC (or token 0) top right will allow the UI to set pool to full range. 
![image](https://user-images.githubusercontent.com/60412342/168486570-f4b9c996-77d1-4213-8243-04208e692d5c.png)

PR #4174, opened 2022-07-23
fix: Full range button for both tokens

KevinNeewen:
**Description**
This PR is to fix the issue [3843](https://github.com/Uniswap/interface/issues/3843)

**What was causing the bug?**
When the user clicks on the _Full Range_ button, the action `setFullRange` is dispatched, setting both `leftRangeTypedValue` and `rightRangeTypedValue` to `true`. That causes the useEffect below to execute and since there are always query params of minPrice and maxPrice on first load of the AddLiquidity page
image
The `typeLeftRangeInput` and `typeRightRangeInput` actions will dispatched, therefore setting the min/max price with what's in the query params.

### DIFF
--- a/src/pages/AddLiquidity/index.tsx
+++ b/src/pages/AddLiquidity/index.tsx
@@ -159,6 +159,7 @@ export default function AddLiquidity() {
 
   useEffect(() => {
     if (
+      leftRangeTypedValue !== true &&
       typeof parsedQs.minPrice === 'string' &&
       parsedQs.minPrice !== leftRangeTypedValue &&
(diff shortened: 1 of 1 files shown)


cmcewen:
thanks for this PR @KevinNeewen! would you be willing to write a test for this behavior?

KevinNeewen:
> thanks for this PR @KevinNeewen! would you be willing to write a test for this behavior?

Sure no problem. I'll make a following commit shortly to cover the scenario via jest!

vm:
closing for now, but feel free to re-open if a test is added 👍

PR #4179, opened 2022-07-23
feat: TokenDetailsPagequery

JFrankfurt:


cmcewen (approved):
nice - i like the error randomness. this will change with relay, but i wonder if we can make all our network requests randomly fail in dev sometimes

PR #4180, opened 2022-07-24
fix: scroll to top only when pathname changes

chenxsan:
When I was adding liquidity on my cellphone through the in-app browser of metamask, I noticed that every time I click the +/- buttons, the page would scroll to top:

Here's why it happens: when we click the +/- buttons, the url would change, e.g.,

from: `/#/add/ETH/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48/3000?chain=mainnet&maxPrice=3261.7&minPrice=919.71`
to: `/#/add/ETH/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48/3000?chain=mainnet&maxPrice=3261.7&minPrice=919.71`

I.e., the `history` object changes, hence `scrollTo` would get called every time I click the +/- buttons.

Which is obviously wrong, we shouldn't call `scrollTo` when query strings change because we are still on the same page.

### DIFF
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -6,7 +6,7 @@ import TopLevelModals from 'components/TopLevelModals'
 import ApeModeQueryParamReader from 'hooks/useApeModeQueryParamReader'
 import { lazy, Suspense } from 'react'
 import { useEffect } from 'react'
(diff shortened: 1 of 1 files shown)


vm (approved):
thank you, was about to make the same change

PR #4181, opened 2022-07-25
feat: Add support for mixed routes in the interface

zhongeric:
The new versions of the smart order router (SOR) and routing-api introduce support for a new kind of route called `MixedRoutes`. MixedRoutes are routes which have both V2 and V3 pools in them, ex:

`USDC -[V3]-> ETH -[V2]-> USDT`

We only support returning these routes through the AutoRouter / routingAPI, and so we can pass in a new `protocols` queryParam of `[Protocol.V2, Protocol.V3, Protocol.MIXED]` to direct the auto router to also look for and consider mixedRoutes. Client side, we should still only query for V3 routes using the client side SOR and/or the QuoterV2 fallback.

We also need to edit the `Trade` object we create here to include the new mixed route, in order to generate the correct calldata to submit the swap with.

We also show a new badge label of `V3 + V2` to denote the new mixed routes in the RoutingDiagram


zhongeric:
TODO: do we want to quote mixedRoutes in `useClientSideTrade` hook?

src/components/RoutingDiagram/RoutingDiagram.tsx
```diff
@@ -73,6 +74,10 @@ const ProtocolBadge = styled(Badge)`
   z-index: ${Z_INDEX.sticky + 1};
 `
 
+const MixedProtocolBadge = styled(ProtocolBadge)`
+  width: 60px;
+`
```
zhongeric, line 79: This seems to work to prevent the text from overflowing and line breaking

src/state/routing/slice.ts
```diff
@@ -36,7 +36,7 @@ const DEFAULT_QUERY_PARAMS = {
 export const routingApi = createApi({
   reducerPath: 'routingApi',
   baseQuery: fetchBaseQuery({
-    baseUrl: 'https://api.uniswap.org/v1/',
+    baseUrl: 'https://r6oz6jqnth.execute-api.us-east-2.amazonaws.com/prod',
```
zhongeric, line 39: ignore this

vm:
@zhongeric can you rebase/merge main?

src/hooks/useAutoSlippageTolerance.ts
```diff
@@ -43,8 +45,17 @@ function guesstimateGas(trade: Trade<Currency, Currency, TradeType> | undefined)
...
+            gas += V3_SWAP_BASE_GAS_ESTIMATE + section.length * V3_SWAP_HOP_GAS_ESTIMATE
+          } else if (section.every((pool) => pool instanceof Pair)) {
+            gas += V2_SWAP_BASE_GAS_ESTIMATE + (section.length - 1) * V2_SWAP_HOP_GAS_ESTIMATE
+          }
+        })
       } else {
-        // TODO: Update with better estimates once we have interleaved routes.
+        // fallback general gas estimation
         gas += V3_SWAP_BASE_GAS_ESTIMATE + route.pools.length * V3_SWAP_HOP_GAS_ESTIMATE
       }
```
zhongeric, line 60: Chose to keep an `else` case as a fallback, not really needed tho

zhongeric:
Do we want to query for mixedRoutes in the `getClientSideQuote` function as well? See: https://github.com/Uniswap/interface/pull/4181/files#diff-c9fd69a3658e7e36f7ab6b62fd18589fed617dad15c3184bc7335b9e6ad7ee02R68

src/hooks/useAutoSlippageTolerance.ts
```diff
@@ -43,8 +45,19 @@ function guesstimateGas(trade: Trade<Currency, Currency, TradeType> | undefined)
...
+        const sections = partitionMixedRouteByProtocol(route as MixedRoute<Currency, Currency>)
+        let acc = 0
+        sections.forEach((section) => {
+          if (section.every((pool) => pool instanceof Pool)) {
+            acc += V3_SWAP_BASE_GAS_ESTIMATE + section.length * V3_SWAP_HOP_GAS_ESTIMATE
+          } else if (section.every((pool) => pool instanceof Pair)) {
+            acc += V2_SWAP_BASE_GAS_ESTIMATE + (section.length - 1) * V2_SWAP_HOP_GAS_ESTIMATE
+          }
+        })
+        gas += acc
```
zhongeric, line 58: Had to use another acc var here because of esLint no-func-loop

src/state/routing/slice.ts
```diff
@@ -36,7 +37,7 @@ const DEFAULT_QUERY_PARAMS = {
 export const routingApi = createApi({
   reducerPath: 'routingApi',
   baseQuery: fetchBaseQuery({
-    baseUrl: 'https://api.uniswap.org/v1/',
+    baseUrl: 'https://beta.api.uniswap.org/v1/',
```
snreynolds, line 40: @zhongeric this should still point to prod. we will test on beta this morning and deploy to prod before releasing interface

zhongeric: ok sg

src/state/routing/slice.ts
```diff
@@ -28,7 +28,6 @@ const protocols: Protocol[] = [Protocol.V2, Protocol.V3, Protocol.MIXED]
 
 const DEFAULT_QUERY_PARAMS = {
   protocols: protocols.map((p) => p.toLowerCase()).join(','),
-  forceMixedRoutes: true,
```
vm: this part should be done in a separate PR

vm: that way we know what we can minimally revert if issues arise

zhongeric: makes sense, will cut

zhongeric: #4326

vm: I think all that PR needs to be is `forceMixedRoutes: true`?

src/hooks/useAutoSlippageTolerance.ts
```diff
@@ -43,8 +45,19 @@ function guesstimateGas(trade: Trade<Currency, Currency, TradeType> | undefined)
         // V3 gas costs scale on initialized ticks being crossed, but we don't have that data here.
         // We bake in some tick crossings into the base 100k cost.
         gas += V3_SWAP_BASE_GAS_ESTIMATE + route.pools.length * V3_SWAP_HOP_GAS_ESTIMATE
+      } else if (route.protocol === Protocol.MIXED) {
+        const sections = partitionMixedRouteByProtocol(route as MixedRoute<Currency, Currency>)
+        let acc = 0
```
zzmp, line 50: `acc` is unnecssary - you should instead just have L53 and L55 use `gas +=` to be more explicit about what you're adding up.

You should also use `reduce` to make it more clear what is happening in the code:

    gas += sections.reduce((gas, section) => {
      if (section.every((pool) => pool instanceof Pool)) {
        return gas + V3_SWAP_BASE_GAS_ESTIMATE + section.length * V3_SWAP_HOP_GAS_ESTIMATE
      } else if (section.every((pool) => pool instanceof Pair)) {
        return gas + V2_SWAP_BASE_GAS_ESTIMATE + (section.length - 1) * V2_SWAP_HOP_GAS_ESTIMATE
      } else {
        console.warn('Invalid section')
        return gas
      }
    })

src/state/routing/utils.ts
```diff
@@ -97,6 +112,19 @@ const parsePair = ({ reserve0, reserve1 }: V2PoolInRoute): Pair =>
...
+function isV3Route(route: (V3PoolInRoute | V2PoolInRoute)[]): route is V3PoolInRoute[] {
+  return route.every((pool) => pool.type === 'v3-pool')
+}
+
+function isV2Route(route: (V3PoolInRoute | V2PoolInRoute)[]): route is V2PoolInRoute[] {
+  return route.every((pool) => pool.type === 'v2-pool')
+}
+
+function isMixedRoute(route: (V3PoolInRoute | V2PoolInRoute)[]): route is (V3PoolInRoute | V2PoolInRoute)[] {
+  /// Must have at least one V3 pool and one V2 pool
```
zzmp, line 128: This isn't comprehensive. Should you also check that every route is either V3 or V2?

src/state/routing/utils.ts
```diff
@@ -71,6 +79,13 @@ export function transformRoutesToTrade<TTradeType extends TradeType>(
       route
         ?.filter((r): r is typeof route[0] & { routev3: NonNullable<typeof route[0]['routev3']> } => r.routev3 !== null)
         .map(({ routev3, inputAmount, outputAmount }) => ({ routev3, inputAmount, outputAmount })) ?? [],
+    mixedRoutes:
```
zzmp, line 82: Where is this used?

zzmp (changes requested):

src/hooks/useAutoSlippageTolerance.ts
zhongeric: I like the .reduce

src/state/routing/utils.ts
```diff
@@ -41,8 +42,15 @@ export function computeRoutes(
       }
 
       return {
```
zzmp, line 44: This requires a lot of looping that isn't necessary. Could you improve it by doing something like this?

    getRouteProtocol(route: (V3PoolInRoute | V2PoolInRoute)[]): Protocol {
      if (route.every((pool) => pool.type === 'v2-pool')) return Protocol.V2
      if (route.every((pool) => pool.type === 'v3-pool')) return Protocol.V3
      return Protocol.MIXED
    }

You could optimize it more but this seems sufficient. The old `isV2Route`-style functions were type guards, but the `genericPoolPairParser` makes it so that you no longer need to cast to a type before parsing the route.

zhongeric: Yeah this is a lot more elegant. This also solves the comment above about the mixed routes not exactly being comprehensive

src/state/routing/utils.ts
zhongeric: Should be resolved by new logic in `getRouteProtocol`

src/state/routing/utils.ts
zhongeric: used to build the `InterfaceTrade` which is returned by `useRoutingAPITrade()` and then in `useBestTrade()`

src/state/routing/utils.ts
```diff
@@ -41,16 +41,21 @@ export function computeRoutes(
         throw new Error('Expected both amountIn and amountOut to be present')
       }
 
+      const routeProtocol: Protocol = getRouteProtocol(route)
```
zzmp, line 44: rm `: Protocol`, the typing is unneeded here. It can be inferred from the return value of `getRouteProtocol`.

zzmp (approved):

PR #4182, opened 2022-07-25
style: add updated fonts

kayleegeorge:
Add font file based on updated theme


src/theme/fonts.ts
```diff
@@ -0,0 +1,78 @@
+export const fontFamily = {
```
vm, line 1: don't these need to get added to to `useTheme`?

kayleegeorge: oh wait ya, I need to do the text style differently -- ignore this PR for now :)

src/theme/index.tsx
```diff
@@ -258,6 +268,45 @@ export const ThemedText = {
   DeprecatedError({ error, ...props }: { error: boolean } & TextProps) {
     return <TextWrapper fontWeight={500} color={error ? 'deprecated_red1' : 'deprecated_text2'} {...props} />
   },
+  HeadlineLarge({ error, ...props }: { error: boolean } & TextProps) {
```
vm, line 271: i think we probably want to stop using these `TextWrapper` components. probably can avoid for now? maybe just add the new fontFamily stuff into theme and that's enough

kayleegeorge: Ya I was curious about this too, but I think design wanted to capture the font size / line height too

kayleegeorge:
Deciding not to use this TextWrapper model for updating fonts so not needed

PR #4183, opened 2022-07-25
fix: make all token row content clickable

kayleegeorge:
make all token row content clickable

Screen Shot 2022-07-25 at 11 08 08 AM


JFrankfurt:
image
Doing clickable content this way results in a dead zone in the margins. Can you move the click action up to the parent?

kayleegeorge:
For clickable content, we decided that only the actual content (text) should be clickable and take you to the token details page. Even if you clicked in the background of any part of the row, it shouldn't take you to the token details page (but the whole row should turn a diff color for hover state)

vm:
talked in DM, but basically switch this to have cursor pointed only on the actual content, not the full cell. since that is what is clickable

.vscode/settings.json
```diff
@@ -5,7 +5,7 @@
   "editor.formatOnSaveMode": "file",
   "editor.tabCompletion": "on",
   "editor.tabSize": 2,
-  "editor.formatOnSave": true,
```
vm, line 8: rm haha

vm (approved):

PR #4185, opened 2022-07-25
feat: token details small mobile views (320px min)

kayleegeorge:
make min width for mobile views 320, adjust placeholder text depending on mobile view, adjust favorites button on mobile view

Screen Shot 2022-07-25 at 12 30 06 PM
Screen Shot 2022-07-25 at 12 29 50 PM


src/components/Explore/TokenTable/FavoriteButton.tsx
```diff
@@ -25,16 +26,22 @@ const StyledFavoriteButton = styled.button<{ active: boolean }>`
     background-color: ${({ theme, active }) => !active && theme.backgroundContainer};
   }
 `
+const FavoriteMobile = styled.span`
```
vm, line 29: `FavoriteText`, mobile shouldn't be in the name

src/components/Explore/TokenTable/SearchBar.tsx
```diff
@@ -37,9 +38,10 @@ const SearchInput = styled.input<{ expanded: boolean }>`
     background-color: ${({ theme }) => theme.backgroundContainer};
   }
   ::placeholder {
-    color: ${({ expanded, theme }) => expanded && theme.textSecondary};
+    color: ${({ expanded, theme }) => expanded && theme.textTertiary};
+
     @media only screen and (max-width: ${MEDIUM_MEDIA_BREAKPOINT}) {
-      color: transparent;
+      color: ${({ theme, isMobile }) => !isMobile && theme.none};
```
vm, line 44: why do you need to pass in `isMobile` vs using the breakpoint?

src/components/Explore/TokenTable/SearchBar.tsx
```diff
@@ -61,13 +63,15 @@ const SearchInput = styled.input<{ expanded: boolean }>`
 export default function SearchBar() {
   const [filterString, setFilterString] = useAtom(filterStringAtom)
   const [isExpanded, setExpanded] = useState(false)
+  const placeholderText = isMobile ? 'Search apps, tokens, and NFTs' : 'Search by name or token address'
```
vm, line 66: best to avoid this, can we just use one? if we need to do this it'd be better to set the content in css based on breakpoints

kayleegeorge: let me ask design, they were different in the designs -- also I tried setting the placeholder content using CSS and i couldn't get it to work a while back, can try again though ahaha

vm (approved):

PR #4187, opened 2022-07-25
build: add staging NODE_ENV

vm:
I think we can always assume that the NODE_ENV is correct, and if we need to test we can mock it instead of passing in a different one.

I need to change this so that I can use the experiment SDK properly, since I need to create a variable outside of a function scope. This also means that we don't call initializeAnalytics multiple times.

TODOS:
- create separate Amplitude configs for development, staging, and production.
- create staging env for Infura
- add staging for serviceworker


src/components/AmplitudeAnalytics/index.ts
```diff
@@ -6,9 +6,8 @@ import { Identify, identify, init, track } from '@amplitude/analytics-browser'
  * Uniswap has two Amplitude projects: test and production. You must be a
  * member of the organization on Amplitude to view details.
```
vm, line 7: i also think this file should live in `src/analytics` rather than `src/components/AmplitudeAnalytics`

vm:
this won't work until i switch to adding a `staging` env variable, so can wait until then

zzmp (approved):

PR #4190, opened 2022-07-26
feat: Web 262 user model custom properties first PR

lynnshaoyu:
first PR, implement wallet related custom user properties. second PR coming up will implement the misc other custom properties, like "light_mode".

Note: I changed the last / first seen dates from WALLET last/first seen date to USER last/first seen date, because of the way amplitude Identify API works. hard to track last/first seen date by wallet.

### DIFF
--- a/src/components/WalletModal/index.tsx
+++ b/src/components/WalletModal/index.tsx
@@ -113,6 +113,22 @@ const WALLET_VIEWS = {
   PENDING: 'pending',
 }
 
+const sendAnalyticsEventAndUserInfo = (account: string, walletType: string, chainId: number | undefined) => {
+  const currentDate = new Date().toISOString()
+  sendAnalyticsEvent(EventName.WALLET_CONNECT_TXN_COMPLETED, {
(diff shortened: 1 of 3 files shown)


src/components/WalletModal/index.tsx
```diff
@@ -156,15 +156,23 @@ export default function WalletModal({
   // When new wallet is successfully set by the user, trigger logging of Amplitude analytics event.
   useEffect(() => {
     if (account && account !== lastActiveWalletAddress) {
+      const walletType = getConnectionName(getConnection(connector).type, getIsMetaMask())
```
cmcewen, line 159: in general i think when there's a decent amount of analytics code it's nice to split it out into a separate function to avoid mixing with UI code, but it's more of a judgement thing

cmcewen (approved):

src/components/WalletModal/index.tsx
lynnshaoyu: sounds good, moved amplitude stuff to other function sendAnalyticsEventAndUserInfo

PR #4201, opened 2022-07-27
fix: update colors on token details page

kayleegeorge:
Update colors in token detail pages and add colors for network (no more deprecated!)

Screen Shot 2022-07-27 at 3 29 45 PM
s

### DIFF
--- a/src/components/Explore/TokenDetails/TokenDetail.tsx
+++ b/src/components/Explore/TokenDetails/TokenDetail.tsx
@@ -3,6 +3,7 @@ import CurrencyLogo from 'components/CurrencyLogo'
 import { useCurrency, useToken } from 'hooks/Tokens'
 import { TimePeriod } from 'hooks/useTopTokens'
 import { useAtomValue } from 'jotai/utils'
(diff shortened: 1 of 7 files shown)


src/components/Explore/TokenDetails/TokenDetail.tsx
```diff
@@ -70,7 +71,7 @@ const ContractAddress = styled.button`
   cursor: pointer;
 
   &:hover {
-    color: ${({ theme }) => theme.textSecondary};
+    color: ${({ theme }) => darken(0.1, theme.textPrimary)};
```
vm, line 74: check if this is what they want

vm (approved):

src/components/Explore/TokenDetails/TokenDetail.tsx
kayleegeorge: I'm just gonna push this for now so that Cal can do some UI polish stuff -- they don't have a design for this yet from the Figma, so Cal can add it to the UI feedback when ready

PR #4202, opened 2022-07-27
feat: explore table sort

JFrankfurt:


src/components/Explore/TokenTable/TokenRow.tsx
```diff
@@ -146,17 +146,17 @@ const ListNumberCell = styled(Cell)`
     display: none;
   }
 `
-const HeaderLabelCell = styled(Cell)<{ sortable: boolean }>`
+const DataCell = styled(Cell)<{ sortable: boolean }>`
   justify-content: flex-end;
   min-width: 80px;
-  padding-right: 8px;
+  user-select: ${({ sortable }) => (sortable ? 'none' : 'unset')};
```
kayleegeorge, line 152: this needs to have padding-right, the design isn't centered -- it's right-aligned with padding-right of 8px

src/components/Explore/TokenTable/TokenRow.tsx
```diff
@@ -201,13 +201,14 @@ const SortArrowCell = styled(Cell)`
...
-  cursor: pointer;
-`
-const SortOption = styled.span`
   cursor: pointer;
+  display: flex;
+  justify-content: center;
+
+  height: 48px;
+  line-height: 16px;
+  width: 100%;
```
kayleegeorge, line 211: is this necessary?

JFrankfurt: I was matching the row height info, but I can simplify it to just `height: 100%;`

src/components/Explore/TokenTable/TokenRow.tsx
```diff
@@ -200,14 +200,14 @@ const PriceInfoCell = styled(Cell)`
...
-const SortingCategory = styled.span`
-  display: flex;
+const HeaderCellWrapper = styled.span<{ onClick?: () => void }>`
   align-items: center;
-  justify-content: center;
-  cursor: pointer;
-`
-const SortOption = styled.span`
-  cursor: pointer;
+  cursor: ${({ onClick }) => (onClick ? 'pointer' : 'unset')};
```
vm, line 205:
```
  ${({ onClick }) => onClick && `
    cursor: pointer;
  `}
```

JFrankfurt: pref the current implementation because it allows predictable property organization. It's not clear where you'd expect to find the snippet you shared if all the other properties are alphabetized.

src/components/Explore/TokenTable/TokenTable.tsx
```diff
@@ -44,30 +52,76 @@ const LOADING_ROWS = Array(10)
...
+        if (!tokenData) {
+          return 0
+        }
+        const token1 = tokenData[token1Address] as any
+        const token2 = tokenData[token2Address] as any
+
+        if (!token1 || !token2 || !sortDirection || !sortCategory) {
+          return 0
+        }
+        if (sortCategory === Category.marketCap) {
```
vm, line 99: can you make a function called ```
function compare(sortDirection: SortDirection, a: any, b: any) {
  if (a === b) {
    return 0;
  } else if (a > b) {
    return sortDirection === SortDirection.decreasing ? -1 : 1
  } else {
    return sortDirection === SortDirection.decreasing ? 1 : -1
  }
}
```

then instead of doing branches for each sortCategory, you could just determine what `a` and `b` are based on the sort category.

```
let a;
let b;
switch (sortCategory) {
  Category.marketCap:
    a = token1.marketCap
    b = token2.marketCap
  ...
}
```

and finally it'd be
```
return compare(sortDirection, a, b)
```

PR #4204, opened 2022-07-28
feat: new swap events and properties in taxonomy 

lynnshaoyu:
1. New Event: Wrap Token Transaction Submitted
Screen Shot 2022-07-28 at 1 16 02 AM

2.  New Event: Approve Token Transaction Submitted
Screen Shot 2022-07-28 at 1 29 57 AM
@willhennessy the approve token can come from multiple places in the app though, do we want to log all? currently im just logging the ones coming from swap page. 

jira: https://uniswaplabs.atlassian.net/browse/WEB-623?atlOrigin=eyJpIjoiNDY1YTM0NzVjMWI1NDQ3ZjhjY2U2NjYzZDY2NThkOTYiLCJwIjoiaiJ9


aadams:
The routing data look like exactly what we wanted! Is there anyway that I could see a multi-hop trade just to make sure?

src/pages/Swap/index.tsx
```diff
@@ -643,53 +688,67 @@ export default function Swap() {
...
-                        altDisabledStyle={approvalState === ApprovalState.PENDING} // show solid button while waiting
-                        confirmed={
-                          approvalState === ApprovalState.APPROVED || signatureState === UseERC20PermitState.SIGNED
+                      <TraceEvent
+                        events={[Event.onClick]}
+                        name={EventName.APPROVE_TOKEN_TXN_SUBMITTED}
+                        element={ElementName.APPROVE_TOKEN_BUTTON}
+                        properties={formatApproveTokenTxnSubmittedEventProperties(approvalOptimizedTrade)}
+                        shouldLogImpression={
+                          !(
```
vm, line 697: i find this conditional kinda confusing. maybe you can name it as a variable? basically what is the stuff inside the () and why should you only log if that is false

src/pages/Swap/index.tsx
```diff
@@ -625,15 +658,27 @@ export default function Swap() {
...
+                  <TraceEvent
+                    events={[Event.onClick]}
+                    name={EventName.WRAP_TOKEN_TXN_SUBMITTED}
+                    element={ElementName.WRAP_TOKEN_BUTTON}
+                    properties={formatWrapTokenTxnSubmittedEventProperties(
+                      currencies[Field.INPUT],
+                      currencies[Field.OUTPUT],
+                      parsedAmount
+                    )}
+                    shouldLogImpression={!Boolean(wrapInputError)}
```
vm, line 670: can you check `!error`?

src/pages/Swap/index.tsx
```diff
@@ -92,7 +92,40 @@ export function getIsValidSwapQuote(
...
+    chain_id: approvalOptimizedTrade.inputAmount.currency.chainId,
+    token_symbol: approvalOptimizedTrade.inputAmount.currency.symbol,
+    token_address: approvalOptimizedTrade.inputAmount.currency.isNative
+      ? NATIVE_CHAIN_ADDRESS
+      : approvalOptimizedTrade.inputAmount.currency.address,
+  }
+}
+
+const formatWrapTokenTxnSubmittedEventProperties = (
+  inputCurrency: Currency | null | undefined,
```
vm, line 113: why can it be `null` or `undefined`?

src/pages/Swap/index.tsx
```diff
@@ -92,7 +92,40 @@ export function getIsValidSwapQuote(
...
+  outputCurrency: Currency | null | undefined,
+  parsedAmount: CurrencyAmount<Currency> | undefined
+) => {
+  if (!inputCurrency || !outputCurrency || !parsedAmount) return {}
+  return {
+    token_in_address: inputCurrency.isNative ? NATIVE_CHAIN_ADDRESS : inputCurrency.address,
+    token_out_address: outputCurrency.isNative ? NATIVE_CHAIN_ADDRESS : outputCurrency.address,
+    token_in_symbol: inputCurrency.symbol,
+    token_out_symbol: outputCurrency.symbol,
+    chain_id: inputCurrency.chainId === outputCurrency.chainId ? inputCurrency.chainId : undefined,
```
vm, line 123: how could the `chain_ids` not be the same? feels like that should be handled before here

vm: i think fine to avoid this because you shoudn't need to update the swap code.

src/components/swap/SwapModalFooter.tsx
```diff
@@ -34,6 +37,39 @@ interface AnalyticsEventProps {
   tokenOutAmountUsd: string | undefined
   lpFeePercent: Percent
   swapQuoteReceivedDate: Date | undefined
+  routes: RoutingDiagramEntry[]
+}
+
+const formatRoutesEventProperties = (routes: RoutingDiagramEntry[]) => {
```
vm, line 43: i think if you do the `forEach` as a single loop and skip storing the intermediate variables like `routesPercentages` this would be easier to read.

i'm not stuck on that and it's up to you, but i find this a bit hard to follow right now.

src/pages/Swap/index.tsx
lynnshaoyu: yeah good call thanks. its basically saying dont log when the approve button is disabled.

src/pages/Swap/index.tsx
lynnshaoyu: true, removed check.

src/pages/Swap/index.tsx
lynnshaoyu: no, this is just the inverse of whats inside disabled={} on line 664.

src/pages/Swap/index.tsx
lynnshaoyu: currencies[Field.INPUT] and currencies[Field.OUTPUT] can both be null or undefined based on state of this react component. currencies is from the hook useDerivedSwapInfo().

src/components/swap/SwapModalFooter.tsx
lynnshaoyu: nice catch youre totally right, only one foreach is needed. done.

vm (approved):

aadams (approved):
the route data looks good! I think the only thing we would want more is the pool the swap went through, but we can back out that information from the data saved.

lynnshaoyu:
> the route data looks good! I think the only thing we would want more is the pool the swap went through, but we can back out that information from the data saved.

as per discussion offline, i'm removing the routing data for now in this PR. separated out that code into a separate pr https://github.com/Uniswap/interface/pull/4221 we can commit later on pending decisions.

PR #4206, opened 2022-07-28
feat: second user model PR with wallet properties

lynnshaoyu:
1) User model changes on wallet update:  
Screen Shot 2022-08-01 at 12 53 47 PM

Screen Shot 2022-07-29 at 2 59 44 PM

2) App loaded event and initial screen /browser related data: 
Screen Shot 2022-07-29 at 4 58 49 PM


src/components/AmplitudeAnalytics/constants.ts
```diff
@@ -26,15 +27,33 @@ export enum EventName {
...
-  SCREEN_RESOLUTION = 'screen_resolution',
+  ALL_WALLET_CHAIN_IDS = 'all_wallet_chain_ids',
   BROWSER = 'browser',
-  LIGHT_MODE = 'light_mode',
+  DARK_MODE = 'is_dark_mode',
+  SCREEN_RESOLUTION_HEIGHT = 'screen_resolution_height',
+  SCREEN_RESOLUTION_WIDTH = 'screen_resolution_width',
+  WALLET_ADDRESS = 'wallet_address',
+  WALLET_NATIVE_CURRENCY_BALANCE_USD = 'wallet_native_currency_balance_usd',
+  WALLET_CHAIN_IDS_PREFIX = 'wallet_chain_ids_',
```
vm, line 38: i think prefix/suffix should live separately?

src/components/AmplitudeAnalytics/index.ts
```diff
@@ -67,16 +67,16 @@ class UserModel {
     identify(mutate(new Identify()))
   }
 
-  set(key: string, value: string | number) {
+  set(key: string, value: string | number | boolean | string[] | number[]) {
     this.call((event) => event.set(key, value))
   }
 
-  setOnce(key: string, value: string | number) {
+  setOnce(key: string, value: string | number | boolean | string[] | number[]) {
```
vm, line 74: `type Value = string | number | boolean | string[] | number[]`

src/components/AmplitudeAnalytics/index.ts
```diff
@@ -67,16 +67,16 @@ class UserModel {
...
     this.call((event) => event.setOnce(key, value))
   }
 
-  add(key: string, value: string | number) {
-    this.call((event) => event.add(key, typeof value === 'number' ? value : 0))
+  add(key: string, value: number) {
```
vm, line 78: should this also be `Value` or is it only a number?

src/components/WalletModal/index.tsx
```diff
@@ -113,20 +121,46 @@ const WALLET_VIEWS = {
...
+  const walletTokensAddresses: string[] = []
+  balances.forEach((currencyAmount) => {
+    if (currencyAmount !== undefined) {
+      const tokenBalanceAmount = formatToDecimal(currencyAmount, currencyAmount.currency.decimals)
+      if (tokenBalanceAmount > 0) {
+        const tokenAddress = getTokenAddress(currencyAmount.currency)
+        walletTokensAddresses.push(getTokenAddress(currencyAmount.currency))
+        walletTokensSymbols.push(currencyAmount.currency.symbol ?? '')
+        const tokenPrefix = currencyAmount.currency.symbol ?? tokenAddress
+        user.set(`${tokenPrefix}${CUSTOM_USER_PROPERTIES.WALLET_TOKEN_AMOUNT_SUFFIX}`, tokenBalanceAmount)
```
vm, line 138: oh i see... hm i'm wondering what a better way to do this would be. maybe we can just send a map of all the balances instead of individual user properties?

also, if it's a user properly can we still look at historical values?

src/components/WalletModal/index.tsx
```diff
@@ -113,20 +121,46 @@ const WALLET_VIEWS = {
...
     wallet_type: walletType,
     // TODO(lynnshaoyu): Send correct is_reconnect value after modifying user state.
   })
+  const currentDate = new Date().toISOString()
   user.set(CUSTOM_USER_PROPERTIES.WALLET_ADDRESS, account)
   user.set(CUSTOM_USER_PROPERTIES.WALLET_TYPE, walletType)
-  if (chainId) user.postInsert(CUSTOM_USER_PROPERTIES.WALLET_CHAIN_IDS, chainId)
+  if (chainId) {
+    user.postInsert(CUSTOM_USER_PROPERTIES.ALL_WALLET_CHAIN_IDS, chainId)
+    user.postInsert(`${CUSTOM_USER_PROPERTIES.WALLET_CHAIN_IDS_PREFIX}${account}`, chainId)
```
vm, line 159: i guess i am unconvinced on storing user property keys that are dynamic like this, but not sure what the alternative would be

src/pages/App.tsx
```diff
@@ -80,16 +82,60 @@ function getCurrentPageFromLocation(locationPathname: string): PageName | undefi
   }
 }
 
+// Get browser being used, code comes from: https://developer.mozilla.org/en-US/docs/Web/API/Window/navigator.
```
vm, line 85: can this live in `utils/browser.ts` or something similar?

src/pages/App.tsx
```diff
@@ -80,16 +82,60 @@ function getCurrentPageFromLocation(locationPathname: string): PageName | undefi
...
   }, [pathname])
 
+  useEffect(() => {
+    // TODO(zzmp): add web vitals event properties to app loaded event.
+    sendAnalyticsEvent(EventName.APP_LOADED)
+    user.set(CUSTOM_USER_PROPERTIES.BROWSER, getBrowser())
+    user.set(CUSTOM_USER_PROPERTIES.DARK_MODE, isDarkMode)
+    user.set(CUSTOM_USER_PROPERTIES.SCREEN_RESOLUTION_HEIGHT, window.screen.height)
+    user.set(CUSTOM_USER_PROPERTIES.SCREEN_RESOLUTION_WIDTH, window.screen.width)
+  }, [isDarkMode])
```
vm, line 137: you want a separate hook just for `APP_LOADED`, otherwise it will resent whenever darkMode is changed. i guess i would leave all the screen resolution stuff in this hook for APP_LOADED, then make a new one for dark mode.

src/components/WalletModal/index.tsx
```diff
@@ -174,10 +230,21 @@ export default function WalletModal({
...
     setLastActiveWalletAddress(account)
   }, [lastActiveWalletAddress, account, connector, chainId])
 
+  // Send wallet balance info once it becomes available.
+  useEffect(() => {
+    if (!tokenBalancesIsLoading && logWalletBalances && balances && nativeCurrencyBalanceUsdValue) {
+      const nativeCurrencyBalanceUsd =
+        native && nativeCurrencyBalanceUsdValue ? parseFloat(nativeCurrencyBalanceUsdValue) : 0
+      sendAnalyticsWalletBalanceUserInfo(balances, nativeCurrencyBalanceUsd)
+      setLogWalletBalances(false)
```
vm, line 244: might be better to name this `setWalletBalancesLogged` and set to true? otherwise it's unclear what the value si for

src/components/WalletModal/index.tsx
```diff
@@ -152,6 +187,27 @@ export default function WalletModal({
...
+  }, [tokenBalances, allTokens, tokenBalancesIsLoading])
+  const native = useNativeCurrency()
+
+  const sortedTokensWithETH: Currency[] = useMemo(() => {
+    // Use Celo ERC20 Implementation and exclude the native asset
+    if (!native) {
+      return sortedTokens
+    }
+    // Always bump the native token to the top of the list.
+    return native ? [native, ...sortedTokens.filter((t) => !t.equals(native))] : sortedTokens
```
vm, line 204: looks like native has to be true because otherwise you just return sorted tokens? you could also remove the check above if you wanted

src/components/WalletModal/index.tsx
```diff
@@ -152,6 +187,27 @@ export default function WalletModal({
   const walletModalOpen = useModalIsOpen(ApplicationModal.WALLET)
   const toggleWalletModal = useToggleWalletModal()
 
+  const allTokens = useAllTokens()
+  const [tokenBalances, tokenBalancesIsLoading] = useAllTokenBalances()
+  const sortedTokens: Token[] = useMemo(() => {
+    void tokenBalancesIsLoading // creates a new array once balances load to update hooks
```
vm, line 193: not sure what this void is doing?

src/pages/App.tsx
vm: doesn't amplitude handle this for us? i think they detect browser

src/pages/App.tsx
lynnshaoyu: ah good point / catch, thanks. done.

src/pages/App.tsx
lynnshaoyu: amplitude doesn't, indicated in their documentation (https://help.amplitude.com/hc/en-us/articles/215562387-User-property-definitions) and we also emailed with them early on to verify. moved to utils/browser.ts

src/components/AmplitudeAnalytics/index.ts
lynnshaoyu: sgtm done

src/components/AmplitudeAnalytics/index.ts
lynnshaoyu: only number

src/components/AmplitudeAnalytics/constants.ts
lynnshaoyu: sgtm done

src/components/WalletModal/index.tsx
lynnshaoyu: sounds good done

src/components/WalletModal/index.tsx
lynnshaoyu: oh true, i lifted most of this code from CurrencySearch.tsx, made the simplification there as well.

src/components/WalletModal/index.tsx
lynnshaoyu: lifted logic from CurrencySearch.tsx, I think the original goal was just to have tokenBalancesIsLoading as a dependency in the useMemo hence the void, but I made the logic more explicit now with the ternary. also udpated CurrencySearch.tsx.

src/components/WalletModal/index.tsx
lynnshaoyu: yeah its not ideal. but Will really wants these values, and Kyle (from amplitude) suggested a similar dynamically generated user property for token amount balances, so I think this is honestly the best way given amplitude's existing post processing capabilities (charting etc)

src/components/WalletModal/index.tsx
lynnshaoyu: objects are generally really not preferrable in amplitude. it hinders a lot of charting functions that would otherwise be available. also this one in particular was a direct recommendation from amplitude folks, i'll forward you the email xchange. 

we can look at historical values because we can look at historical events, and all the user properties at that time will be attacahed to the event.

src/components/SearchModal/CurrencySearch.tsx
```diff
@@ -126,7 +126,7 @@ export function CurrencySearch({
     const s = debouncedQuery.toLowerCase().trim()
     if (native.symbol?.toLowerCase()?.indexOf(s) !== -1) {
       // Always bump the native token to the top of the list.
-      return native ? [native, ...filteredSortedTokens.filter((t) => !t.equals(native))] : filteredSortedTokens
+      return [native, ...filteredSortedTokens.filter((t) => !t.equals(native))]
```
vm, line 129: would this potentially return undefined at the front if native is not defined?

src/components/WalletModal/index.tsx
```diff
@@ -159,6 +200,25 @@ export default function WalletModal({
...
+  const sortedTokens: Token[] = useMemo(
+    () => (!tokenBalancesIsLoading ? Object.values(allTokens).sort(tokenComparator.bind(null, tokenBalances)) : []),
+    [tokenBalances, allTokens, tokenBalancesIsLoading]
+  )
+  const native = useNativeCurrency()
+
+  const sortedTokensWithETH: Currency[] = useMemo(
+    () =>
+      // Always bump the native token to the top of the list.
+      native ? [native, ...sortedTokens.filter((t) => !t.equals(native))] : sortedTokens,
```
vm, line 214: i think this logic is also used above? https://github.com/Uniswap/interface/pull/4206/files#diff-a3bebdfee25c1ff21ece35285ea88c59da706d32fa738ae2a1941d2e762c7039R129 perhaps we should make a funciton for this

src/components/SearchModal/CurrencySearch.tsx
lynnshaoyu: if native is undefined this code wouldn't be reached, would return out of the fn on line 123

src/components/WalletModal/index.tsx
lynnshaoyu: its just one line thats the same ([native, ...sortedTokens.filter((t) => !t.equals(native))]) seems a little too small to extract imo.

PR #4213, opened 2022-07-29
feat: Adjust gas estimates for auto slippage

willpote:
Updates the local gas estimation algorithm used to compute auto slippage to be more similar to the one used in the Auto Router.
Adds a 10% buffer to the gas estimates overall. Gas estimates from the routing API sometimes under-estimate which can lead to very tight auto-slippage being set, especially on L2s where gas costs are low anyway

### DIFF
--- a/src/hooks/useAutoSlippageTolerance.ts
+++ b/src/hooks/useAutoSlippageTolerance.ts
@@ -24,11 +24,14 @@ function guesstimateGas(trade: Trade<Currency, Currency, TradeType> | undefined)
   if (!!trade) {
     let gas = 0
     for (const { route } of trade.swaps) {
-      if (route.protocol == Protocol.V2) {
+      if (route.protocol === Protocol.V2) {
         gas += 135000 + route.pools.length * 50_000
-      } else {
+      } else if (route.protocol === Protocol.V3) {
         // V3 gas costs scale on initialized ticks being crossed, but we don't have that data here.
         // We bake in some tick crossings into the base 100k cost.
+        gas += 100_000 + route.pools.length * 70_000
+      } else {
+        // TODO: Update with better estimates once we have interleaved routes.
         gas += 100_000 + route.pools.length * 70_000
       }
     }
--- a/src/hooks/useAutoSlippageTolerance.ts
+++ b/src/hooks/useAutoSlippageTolerance.ts
@@ -1,4 +1,4 @@
-import { Trade } from '@uniswap/router-sdk'
+import { Protocol, Trade } from '@uniswap/router-sdk'
 import { Currency, CurrencyAmount, Percent, TradeType } from '@uniswap/sdk-core'
 import { useWeb3React } from '@web3-react/core'
 import { SUPPORTED_GAS_ESTIMATE_CHAIN_IDS } from 'constants/chains'
@@ -14,6 +14,7 @@ import useStablecoinPrice, { useStablecoinValue } from './useStablecoinPrice'
 const V3_SWAP_DEFAULT_SLIPPAGE = new Percent(50, 10_000) // .50%
 const ONE_TENTHS_PERCENT = new Percent(10, 10_000) // .10%
 export const DEFAULT_AUTO_SLIPPAGE = ONE_TENTHS_PERCENT
+const GAS_ESTIMATE_BUFFER = new Percent(10, 100) // 10%
 
 /**
  * Return a guess of the gas cost used in computing slippage tolerance for a given trade
@@ -21,7 +22,17 @@ export const DEFAULT_AUTO_SLIPPAGE = ONE_TENTHS_PERCENT
  */
 function guesstimateGas(trade: Trade<Currency, Currency, TradeType> | undefined): number | undefined {
   if (!!trade) {
-    return 100_000 + trade.swaps.reduce((memo, swap) => swap.route.pools.length + memo, 0) * 30_000
+    let gas = 0
+    for (const { route } of trade.swaps) {
+      if (route.protocol == Protocol.V2) {
+        gas += 135000 + route.pools.length * 50_000
+      } else {
+        // V3 gas costs scale on initialized ticks being crossed, but we don't have that data here.
+        // We bake in some tick crossings into the base 100k cost.
+        gas += 100_000 + route.pools.length * 70_000
+      }
+    }
+    return gas
   }
   return undefined
 }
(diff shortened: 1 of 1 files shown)


src/hooks/useAutoSlippageTolerance.ts
```diff
@@ -14,14 +14,28 @@ import useStablecoinPrice, { useStablecoinValue } from './useStablecoinPrice'
...
+const GAS_ESTIMATE_BUFFER = new Percent(10, 100) // 10%
 
 /**
  * Return a guess of the gas cost used in computing slippage tolerance for a given trade
  * @param trade the trade for which to _guess_ the amount of gas it would cost to execute
  */
 function guesstimateGas(trade: Trade<Currency, Currency, TradeType> | undefined): number | undefined {
   if (!!trade) {
-    return 100_000 + trade.swaps.reduce((memo, swap) => swap.route.pools.length + memo, 0) * 30_000
+    let gas = 0
```
vm, line 25: if this is copied logic, can you link to the location in auto router api as comment?

src/hooks/useAutoSlippageTolerance.ts
```diff
@@ -14,14 +14,28 @@ import useStablecoinPrice, { useStablecoinValue } from './useStablecoinPrice'
...
+    for (const { route } of trade.swaps) {
+      if (route.protocol === Protocol.V2) {
+        gas += 135000 + route.pools.length * 50_000
+      } else if (route.protocol === Protocol.V3) {
+        // V3 gas costs scale on initialized ticks being crossed, but we don't have that data here.
+        // We bake in some tick crossings into the base 100k cost.
+        gas += 100_000 + route.pools.length * 70_000
+      } else {
```
vm, line 33: does this mean it's `v1`?

src/hooks/useAutoSlippageTolerance.ts
```diff
@@ -61,8 +75,8 @@ export default function useAutoSlippageTolerance(
     // if not, use local heuristic
     const dollarCostToUse =
       chainId && SUPPORTED_GAS_ESTIMATE_CHAIN_IDS.includes(chainId) && trade?.gasUseEstimateUSD
-        ? trade.gasUseEstimateUSD
-        : dollarGasCost
+        ? trade.gasUseEstimateUSD.multiply(GAS_ESTIMATE_BUFFER)
```
cmcewen, line 78: not sure i understand why you're multiplying by 10% here?

src/hooks/useAutoSlippageTolerance.ts
```diff
@@ -14,14 +14,28 @@ import useStablecoinPrice, { useStablecoinValue } from './useStablecoinPrice'
...
  * Return a guess of the gas cost used in computing slippage tolerance for a given trade
  * @param trade the trade for which to _guess_ the amount of gas it would cost to execute
  */
 function guesstimateGas(trade: Trade<Currency, Currency, TradeType> | undefined): number | undefined {
   if (!!trade) {
-    return 100_000 + trade.swaps.reduce((memo, swap) => swap.route.pools.length + memo, 0) * 30_000
+    let gas = 0
+    for (const { route } of trade.swaps) {
+      if (route.protocol === Protocol.V2) {
+        gas += 135000 + route.pools.length * 50_000
```
vm, line 28: can you name variables for `135000` and `50_000`?

vm: i think you can actually do something like this...
```ts
let baseGas = 0
let perRouteGas = 0

if (route.protocol === Protocol.V2) {
  baseGas = 135000
  perRouteGas = 50_000
}

gas += baseGas + route.pools.length * perRouteGas
```

something like that

src/hooks/useAutoSlippageTolerance.ts
willpote: The logic is modified because in the auto router we have additional data to make our guesses based on, but I'll add the link anyway for reference

src/hooks/useAutoSlippageTolerance.ts
willpote: No interleaved routes are a new feature we are launching in the next couple weeks. It means the route can now be like:

ETH --V2pool --> USDC --V3pool--> DAI

I'll be letting Eric know that he needs to update this code when he does the PR for it

vm: ahh i see, thanks!

src/hooks/useAutoSlippageTolerance.ts
willpote: Yeah its all pretty rough but my thinking is:

1/ The gas estimates from the routing API and from the guesstimate function will over-estimate in some cases and under-estimate in other cases. 

2/ Underestimates are bad as we end up setting slippage ultra tight which leads to transaction failures => bad ux

3/ On Arbitrum transaction failures rates are hovering around 30% which feels pretty high https://dune.com/queries/565180/1059058 

4/ This whole auto slippage stuff is ultra rough anyway and adding 10% here or there doesn't really mean much in the grand scheme. It should lean us slightly more towards better UX (less failures) with a very hard to quantify con of user getting a worse trade due to being front run or slippage

src/hooks/useAutoSlippageTolerance.ts
willpote: Each route could use a different protocol so I didn't go with what you suggested. I moved the values into constants though

src/hooks/useAutoSlippageTolerance.ts
vm: that is polygon not arbitrum btw

src/hooks/useAutoSlippageTolerance.ts
vm: works for me

vm (approved):
ideally suggest adding some tests to this so we don't break swapping, but otherwise lgtm

willpote:
> ideally suggest adding some tests to this so we don't break swapping, but otherwise lgtm

Hmm i don't see any existing tests for this (or any of the other hooks), and don't think i have the react skills to set up new tests for this

src/hooks/useAutoSlippageTolerance.ts
ZumZoom: I'll kindly reask the question. From the explanation it is still not clear why is it multiplying by 10% instead of adding extra 10%?

PR #4217, opened 2022-07-29
feat: fetch stablecoin price with SOR, PI warning

cmcewen:


src/state/routing/useRoutingAPITrade.ts
```diff
@@ -35,7 +41,13 @@ export function useRoutingAPITrade<TTradeType extends TradeType>(
     [amountSpecified, otherCurrency, tradeType]
   )
 
-  const [clientSideRouter] = useClientSideRouter()
+  let clientSideRouter: boolean
+  const [clientSideRouterStoredPreference] = useClientSideRouter()
+  if (routerPreference) {
+    clientSideRouter = routerPreference === RouterPreference.CLIENT ? true : false
```
vm, line 47:
```ts
const [clientSideRouterStoredPreference] = useClientSideRouter()
const clientSideRouter = routerPreference  ? routerPreference ===  RouterPreference.CLIENT : clientSideRouterStoredPreference
```

vm: just a preference up to you

src/pages/Swap/index.tsx
```diff
@@ -68,6 +69,7 @@ import { useExpertModeManager } from '../../state/user/hooks'
 import { LinkStyledButton, ThemedText } from '../../theme'
 import { computeFiatValuePriceImpact } from '../../utils/computeFiatValuePriceImpact'
 import { maxAmountSpend } from '../../utils/maxAmountSpend'
+import { computeRealizedPriceImpact, warningSeverity } from '../../utils/prices'
```
JFrankfurt, line 72: bumping the `warningSeverity` is already imported notif below

src/pages/Swap/index.tsx
```diff
@@ -596,6 +605,7 @@ export default function Swap() {
                   allowedSlippage={allowedSlippage}
                 />
               )}
+              {showPriceImpactWarning ? <PriceImpactWarning priceImpact={largerPriceImpact} /> : undefined}
```
vm, line 608: think it needs to be null rather than undefined? you can also do `showPriceImactWarning && <PriceImpactWarning`

src/pages/Swap/index.tsx
```diff
@@ -86,6 +88,17 @@ export function getIsValidSwapQuote(
...
+  if (a && b) {
+    return a.greaterThan(b) ? a : b
+  } else if (a) {
+    return a
+  } else if (b) {
+    return b
+  }
+  return undefined
+}
+
```
JFrankfurt, line 101: nit: I feel kinda meh on this being a fn since it's p simple and only used once.

It shouldn't be exported. It's only used here, and linters won't usually alert you to unused variables if they're exported.

src/hooks/useStablecoinPrice.ts
```diff
@@ -28,11 +27,7 @@ export default function useStablecoinPrice(currency?: Currency): Price<Currency,
   const amountOut = chainId ? STABLECOIN_AMOUNT_OUT[chainId] : undefined
   const stablecoin = amountOut?.currency
 
-  // TODO(#2808): remove dependency on useBestV2Trade
-  const v2USDCTrade = useBestV2Trade(TradeType.EXACT_OUTPUT, amountOut, currency, {
-    maxHops: 2,
-  })
-  const v3USDCTrade = useClientSideV3Trade(TradeType.EXACT_OUTPUT, amountOut, currency)
+  const trade = useRoutingAPITrade(TradeType.EXACT_OUTPUT, amountOut, currency, RouterPreference.CLIENT)
```
JFrankfurt, line 30: can we destructure this to avoid the annoying `trade.trade` thing?

src/components/swap/AdvancedSwapDetails.tsx
```diff
@@ -63,11 +56,10 @@ export function AdvancedSwapDetails({
...
 
   const { expectedOutputAmount, priceImpact } = useMemo(() => {
-    if (!trade) return { expectedOutputAmount: undefined, priceImpact: undefined }
-    const expectedOutputAmount = trade.outputAmount
-    const realizedLpFeePercent = computeRealizedLPFeePercent(trade)
-    const priceImpact = getPriceImpactPercent(realizedLpFeePercent, trade)
-    return { expectedOutputAmount, priceImpact }
+    return {
+      expectedOutputAmount: trade?.outputAmount,
+      priceImpact: trade ? computeRealizedPriceImpact(trade) : undefined,
```
vm, line 61: do we no longer subtract LP fee percentage?

src/hooks/useStablecoinPrice.ts
JFrankfurt, line 30: Does this break client-side-only price impact calculations?

src/components/swap/AdvancedSwapDetails.tsx
JFrankfurt: That happens in computeRealizedPriceImpact

JFrankfurt:
Some nits and questions, but overall lgtm. Biggest q is re: breaking price impact for client-side-only tx

src/pages/Swap/index.tsx
vm: apparently in react 18 can return undefined now so can ignore this if you want

src/pages/Swap/index.tsx
```diff
@@ -205,7 +218,7 @@ export default function Swap() {
   const outputValue = showWrap ? parsedAmount : trade?.outputAmount
   const fiatValueInput = useStablecoinValue(inputValue)
   const fiatValueOutput = useStablecoinValue(outputValue)
-  const priceImpact = useMemo(
+  const stablecoinPriceImpact = useMemo(
```
vm, line 221: why is this variable stableCoin specific now?

src/components/swap/AdvancedSwapDetails.tsx
cmcewen: this happens inside computeRealizePriceImpact. lp percentage is calculated from trade

src/pages/Swap/index.tsx
cmcewen: yeah i tried this and though i was bad at react but actually it just persisted the severity so i needed a new var

src/pages/Swap/index.tsx
cmcewen: so there are actually two different ways to calculate price impact - based on taking input token amount + output token amount and swapping to stablecoins, and comparing those diffs, or by using the swap's impact on the market midprice. i found it confusing that they are both referred to as priceImpact so i tried to differentiate by naming them differently in this file, though it's not an exhaustive rebrand

src/pages/Swap/index.tsx
vm: agree, this function seems a bit strange to me

src/hooks/useStablecoinPrice.ts
cmcewen: why would it do that? this uses the client router so it should only do client calculations. pote didn't want to add more load to the routing api

src/pages/Swap/index.tsx
cmcewen: removed the export. i felt like reading the code i had to think too much about what was going on, so it helped me to just name the function

vm: fair enough to me then

vm (approved):
LGTM, would wait on a Jordan approve also

src/hooks/useStablecoinPrice.ts
JFrankfurt: lol I missed the v obvious `RouterPreference.CLIENT)` and just saw the fn name change

JFrankfurt (approved):

PR #4248, opened 2022-08-01
feat: initial FeatureFlagProvider

vm:
Adds a base FeatureFlagProvider and an example hook.

Next step is to fetch the flags from the amplitude SDK, and then block loading of the app until flags are loaded.


src/featureFlag/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+enum Phase0Flag {
+  Control = 'Control',
+  Enabled = 'Enabled',
+}
+
+export function usePhase0Flag(): Phase0Flag {
+  const context = useContext(FeatureFlagContext)
+  if (!context) throw Error('Feature flag hooks can only be used by children of FeatureFlagProvider.')
+  return Phase0Flag.Control
```
vm, line 22: later this will get the flags state from context and return it

src/featureFlag/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+interface FeatureFlagContextType {
+  isLoaded: boolean
+}
+
+const DEFAULT_STATE = { isLoaded: false }
+const FeatureFlagContext = createContext<FeatureFlagContextType>(DEFAULT_STATE)
+
+export function FeatureFlagProvider({ children }: { children: ReactNode }) {
+  return <FeatureFlagContext.Provider value={{ isLoaded: false }}>{children}</FeatureFlagContext.Provider>
```
vm, line 11: this provider will also own not rendering the app until the flags are fetched (or a timeout has been hit)

yarn.lock
```diff
@@ -17482,6 +17498,11 @@ unc-path-regex@^0.1.2:
   resolved "https://registry.yarnpkg.com/unc-path-regex/-/unc-path-regex-0.1.2.tgz#e73dd3d7b0d7c5ed86fbac6b0ae7d8c6a69d50fa"
   integrity sha1-5z3T17DXxe2G+6xrCufYxqadUPo=
 
+unfetch@4.1.0:
```
lynnshaoyu, line 17501: why is this needed?

src/featureFlag/index.tsx
lynnshaoyu: could you add in typescript documentation above this an example for how developers can use this hook. the example you gave in notion looks good

src/featureFlag/index.tsx
```diff
@@ -0,0 +1,23 @@
+import { createContext, ReactNode, useContext } from 'react'
+
+interface FeatureFlagContextType {
+  isLoaded: boolean
+}
+
+const DEFAULT_STATE = { isLoaded: false }
+const FeatureFlagContext = createContext<FeatureFlagContextType>(DEFAULT_STATE)
```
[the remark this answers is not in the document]
lynnshaoyu: can you add a TODO describing what needs to be changed

yarn.lock
vm: its just a dependency within amplitude's package.json

lynnshaoyu (approved):

PR #4263, opened 2022-08-03
style: import static assets for NFTs

cbachmeier:
Static images/assets for NFT pages


zzmp (approved):
I suspect that some of these are duplicates of assets already in the codebase, but I think it's fine overall.

PR #4274, opened 2022-08-04
refactor: moved token detail price into chart

cartcrom:
Moved the token price & delta percentage from the tokendetails component into the price chart component & overlayed them onto the chart svg so that crosshair will appear when hovering over the whitespace above the chart

### DIFF
--- a/src/components/Charts/PriceChart.tsx
+++ b/src/components/Charts/PriceChart.tsx
@@ -28,6 +28,14 @@ function getDelta(start: number, current: number) {
   return (isPositive ? '+' : '') + delta.toFixed(2)
 }
 
+export const ChartWrapper = styled.div`
+  position: relative;
+`
+
+export const ChartHeader = styled.div`
+  position: absolute;
+`
+
 export const TokenPrice = styled.span`
   font-size: 36px;
   line-height: 44px;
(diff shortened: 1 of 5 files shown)


src/components/Charts/PriceChart.tsx
```diff
@@ -30,13 +28,35 @@ function getDelta(start: number, current: number) {
...
+  font-size: 36px;
+  line-height: 44px;
+`
+export const DeltaContainer = styled.div`
+  height: 16px;
+  display: flex;
+  align-items: center;
+`
+const ArrowCell = styled.div`
+  padding-left: 2px;
```
vm, line 49: technically we don't allow 2px, lmk if we can avoid

src/components/Charts/PriceChart.tsx
```diff
@@ -30,13 +28,35 @@ function getDelta(start: number, current: number) {
...
+`
+
 interface PriceChartProps {
   width: number
   height: number
 }
 
 export function PriceChart({ width, height }: PriceChartProps) {
-  const margin = { top: 40, bottom: 50 }
+  const margin = { top: 80, bottom: 20, crosshair: 70 }
```
vm, line 59: can you do 72 haha

src/components/Charts/PriceChart.tsx
cartcrom: I copied this over from TokenDetail

vm: ah ok

src/components/Charts/PriceChart.tsx
vm: if copied all good

vm (approved):

PR #4276, opened 2022-08-04
build: enable flags on staging

vm:
Allows all feature flags to be enabled on Vercel for now, eventually will add a better way to toggle them!


vm:
decided against this because this will mean vercel is no longer a true production preview

PR #4278, opened 2022-08-04
refactor: separate files for feature flags, use `BaseVariant`

vm:
Instead of needing to reimplement the `useFlag` each time, and create a new enum when it's simply `CONTROL | ENABLED`, we made a base hook and a base enum.

I also made a file per flag, that way we have less conflicts between PRs and can delete a whole file once an experiment is complete.


cartcrom (approved):
Looks good

PR #4279, opened 2022-08-04
fix: theme add hover state and flyout colors

kayleegeorge:
fix: theme add hover state and flyout colors


vm (approved):
wait for vercel and then can merge

PR #4283, opened 2022-08-04
chore: remove selectedWalletBackfilled

vm:
Removes the backfilling logic. Now that we've had the code merged in main for 2 months our active users will have had their `selectedWallet` properties set. This means that we only need to lazily load the various connectors after this is merged for new users, which should reduce load time.

We want to remove Fortmatic in a separate PR still, which has a very large load time and is not lazily loaded currency.

### DIFF
--- a/src/hooks/useEagerlyConnect.ts
+++ b/src/hooks/useEagerlyConnect.ts
@@ -2,7 +2,6 @@ import { Connector } from '@web3-react/types'
 import { gnosisSafeConnection, networkConnection } from 'connection'
 import { getConnection } from 'connection/utils'
 import { useEffect } from 'react'
-import { BACKFILLABLE_WALLETS } from 'state/connection/constants'
 import { useAppSelector } from 'state/hooks'
 
 async function connect(connector: Connector) {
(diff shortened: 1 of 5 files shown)


cypress/e2e/add-liquidity.test.ts
```diff
@@ -3,6 +3,10 @@ import { CyHttpMessages } from 'cypress/types/net-stubbing'
 import { aliasQuery, hasQuery } from '../utils/graphql-test-utils'
 
 describe('Add Liquidity', () => {
+  before(() => {
+    window.localStorage.set('selectedWallet', 'INJECTED')
```
zzmp, line 7: I think we want this to be a part of the visit override, and not done in each individual test. See cypress/support/e2e.ts: `Cypress.Commands.overwrite('visit', ...)`.

src/hooks/useEagerlyConnect.ts
```diff
@@ -27,11 +25,6 @@ export default function useEagerlyConnect() {
 
     if (selectedWallet) {
       connect(getConnection(selectedWallet).connector)
-    } else if (!selectedWalletBackfilled) {
-      BACKFILLABLE_WALLETS.map(getConnection)
-        .map((connection) => connection.connector)
-        .forEach(connect)
-    }
-    // The dependency list is empty so this is only run once on mount
+    } // The dependency list is empty so this is only run once on mount
```
zzmp, line 28: If eslint allows it, this comment should be on the next line

zzmp (changes requested):

src/hooks/useEagerlyConnect.ts
vm: I can't because that's where i do the disable-line

zzmp (approved):

PR #4284, opened 2022-08-04
feat: log is expert mode in amplitude user model

lynnshaoyu:
requested by @willhennessy to start logging expert mode in user model.

### DIFF
--- a/src/state/user/hooks.tsx
+++ b/src/state/user/hooks.tsx
@@ -1,6 +1,8 @@
 import { Percent, Token } from '@uniswap/sdk-core'
 import { computePairAddress, Pair } from '@uniswap/v2-sdk'
 import { useWeb3React } from '@web3-react/core'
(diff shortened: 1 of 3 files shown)


src/state/user/hooks.tsx
```diff
@@ -101,6 +103,7 @@ export function useExpertModeManager(): [boolean, () => void] {
 
   const toggleSetExpertMode = useCallback(() => {
     dispatch(updateUserExpertMode({ userExpertMode: !expertMode }))
+    user.set(CUSTOM_USER_PROPERTIES.EXPERT_MODE, !expertMode)
```
vm, line 106: if you're gonna do this in the effect above, then don't need this.

lynnshaoyu: ah yep youre right removed thanks.

vm (approved):

PR #4285, opened 2022-08-04
feat: Phase 0 NavBar

cbachmeier:
- NavBar for token explore
- Features old wallet
- Hidded behind Feature Flag
- Requires the correct env variables as mentioned in slack
- There are a couple of difference between this NavBar and the one merged into interface-nft, notably some style changes as well as hiding NFT specific components
- Code in the /nft/components folder are some shared wrapper components used across NFT design including an overlay, createPortal wrapper, and Box, Row, Column, which import vanilla extract sprinkles directly


src/nft/components/Box.ts
```diff
@@ -0,0 +1,47 @@
...
+  })
+
+  return React.createElement(as, {
+    className: clsx(atomicClasses, className),
+    ...nativeProps,
+    ref,
+  })
+})
+
+export const AnimatedBox = animated(Box)
```
vm, line 43: can we avoid adding spring-spring until AnimatedBox is used?

src/components/NavBar/ChainSwitcher.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const previousUrlChainId = usePrevious(urlChainId)
+
+  const ref = useRef<HTMLDivElement>(null)
+  useOnClickOutside(ref, isOpen ? toggleOpen : undefined)
+
+  const navigate = useNavigate()
+  const location = useLocation()
+  const info = chainId ? getChainInfo(chainId) : undefined
+
+  const onSelectChain = useCallback(
```
vm, line 103: this logic is shared with the existing NetworkSelector. can we move it into a separate hook so it can be re-used? https://github.com/Uniswap/widgets/pull/70 example in here

src/components/NavBar/ChainSwitcher.tsx
```diff
@@ -0,0 +1,178 @@
...
+      }
+
+      if (!skipToggle) {
+        toggleOpen()
+      }
+    },
+    [connector, toggleOpen, dispatch]
+  )
+
+  useEffect(() => {
```
vm, line 125: i think we should move these into `App.tsx` instead, so then you won't need to re-do them here.

vm:
This PR has a lot of features at once, could we split it up into...

- Update the FeatureFlagModal vh
- Move chainId query hooks into App.tsx
- Factor out `onSelectChain` into a hook, and use that in the existing NetworkSelector
- PR for just ChainSwitcher, flag that on the existing Header component

Then after that would be...:
- Add new Nav styled with no other components
- Add MobileNav conditional with no content
- Add Menu
- Add SearchBar

This is just a quick listing, and open-minded on how to split it up better. Will also make it so we can split who reviews the PRs.

src/nft/components/Box.ts
cbachmeier: turns out react-spring was already imported for use in `src/components/Popups/PopupItem.tsx` so I removed the extra import and used the existing one

src/components/NavBar/ChainSwitcher.tsx
```diff
@@ -85,67 +60,19 @@ interface ChainSwitcherProps {
...
-    if (chainId && !urlChainId) {
-      navigate({ search: replaceURLParam(location.search, 'chain', getChainNameFromId(chainId)) }, { replace: true })
-    }
-  }, [chainId, location, urlChainId, urlChain])
+  const selectChain = useSelectChain()
+  useSyncChainQuery()
 
-  if (!chainId || !info || !provider) {
+  // TODO: need to handle unsupported chain
+  if (!chainId || !info) {
```
vm, line 75: need to render a state for unsupported chains

PR #4287, opened 2022-08-05
test: update snapshots based on color change

vm:
.


cartcrom (approved):
insane color accuracy

PR #4288, opened 2022-08-05
build: declare d3-curve-circlecorners types

zzmp:
Use d3-curve-circlecorners from the registry, by supplying types in the global definitions file.


cartcrom (approved):
Looks great, thank you!

PR #4297, opened 2022-08-06
refactor: removes unused useFastForwardBlockNumber

enzoferey:
I was looking through the updater implementation and I noticed this hook is actually not being used.


cmcewen:
thanks @enzoferey! looking at this i don't see us use fastForward anywhere, so could you remove it completely?

enzoferey:
@cmcewen yes of course ! 👍🏻

enzoferey:
@cmcewen done, let me done if that's not what you had in mind as a full clean up related to the `fastForward` block number feature 👍🏻

vm (approved):

enzoferey:
Cheers 👍🏻

PR #4307, opened 2022-08-08
feat: added token sparklines to explore page

cartcrom:
replaced static svg with sparkline chart component. Abstracted linechart out of pricechart, so that sparklinechart & pricechart can use same line

### DIFF
--- a/src/components/Charts/LineChart.tsx
+++ b/src/components/Charts/LineChart.tsx
@@ -2,6 +2,7 @@ import { Group } from '@visx/group'
 import { LinePath } from '@visx/shape'
 import { CurveFactory } from 'd3'
 import { radius } from 'd3-curve-circlecorners'
+import React from 'react'
 import { ReactNode } from 'react'
 import { useTheme } from 'styled-components/macro'
 import { Color } from 'theme/styled'
@@ -19,7 +20,7 @@ interface LineChartProps<T> {
   height: number
 }
 
-export default function LineChart<T>({
+function LineChart<T>({
   data,
   getX,
   getY,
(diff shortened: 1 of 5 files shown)


src/components/Charts/LineChart.tsx
```diff
@@ -0,0 +1,50 @@
...
+  marginTop: number
+  curve?: CurveFactory
+  color?: Color
+  strokeWidth: number
+  children?: ReactNode
+  width: number
+  height: number
+}
+
+export default function LineChart<T>({
```
cmcewen, line 22: consider memoizing this component since it may be expensive to render: https://reactjs.org/docs/react-api.html#reactmemo

src/components/Explore/TokenDetails/ShareButton.tsx
```diff
@@ -13,6 +13,7 @@ const ShareButtonDisplay = styled.div`
   display: flex;
   cursor: pointer;
   position: relative;
+  z-index: 1;
```
cmcewen, line 16: could you use a z index value from here? https://github.com/Uniswap/interface/blob/main/src/theme/index.tsx#L29

src/components/Charts/SparklineChart.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const max = Math.max(...prices)
+  return [min, max]
+}
+
+interface SparklineChartProps {
+  width: number
+  height: number
+}
+
+export default function SparklineChart({ width, height }: SparklineChartProps) {
```
cmcewen, line 21: also probably want to react.memo this

cmcewen (approved):

PR #4308, opened 2022-08-08
feat: remove unnecessary user properties

lynnshaoyu:
remove O(n) properties so user properties are O(1)

### DIFF
--- a/src/components/WalletModal/index.tsx
+++ b/src/components/WalletModal/index.tsx
@@ -5,7 +5,6 @@ import { Connector } from '@web3-react/types'
 import { sendAnalyticsEvent, user } from 'components/AmplitudeAnalytics'
 import {
   CUSTOM_USER_PROPERTIES,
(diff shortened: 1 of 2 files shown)


src/components/WalletModal/index.tsx
```diff
@@ -163,16 +162,12 @@ const sendAnalyticsEventAndUserInfo = (
     wallet_type: walletType,
     is_reconnect: isReconnect,
   })
-  const currentDate = new Date().toISOString()
```
zzmp, line 166: Will the date now be captured by Amplitude as metadata for the property (eg when it was set), or do we lose the date by no longer setting it?

lynnshaoyu: no we lose the date, its not in the metadata from what I understand. we just see the latest values for users. we can see when user model values changed for each event, and use the dates from the events though.

zzmp (approved):

PR #4320, opened 2022-08-10
feat: integrate relay

cartcrom:
to generate relay queries: yarn relay (or just yarn install)


package.json
```diff
@@ -57,6 +63,7 @@
   "devDependencies": {
     "@ethersproject/experimental": "^5.4.0",
     "@graphql-codegen/cli": "1.21.5",
+    "@graphql-codegen/schema-ast": "^2.5.1",
     "@graphql-codegen/typescript": "1.22.3",
     "@graphql-codegen/typescript-operations": "^1.18.2",
     "@graphql-codegen/typescript-rtk-query": "^1.1.1",
```
JFrankfurt, line 69: It looks like you've removed these from the codegen file--I was surprised to see that. Do we not need `@graphql-codegen/typescript` or  `@graphql-codegen/typescript-operations`?

src/fetchGraphQL.ts
```diff
@@ -0,0 +1,53 @@
...
+const fetchQuery = (operation: any, variables: Variables): ObservableFromValue<GraphQLResponse> => {
+  // TODO: figure out why this returns null
+  const chainId = (store.getState() as AppState).application.chainId
+
+  const subgraphUrl = CHAIN_SUBGRAPH_URL[1]
+
+  const body = JSON.stringify({
+    query: operation.text, // GraphQL text from input
+    variables,
+  })
```
JFrankfurt, line 42: looking at the first example link there's a query name provided here. Why is that not relevant here?

src/fetchGraphQL.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+const headers = {
+  Accept: 'application/json',
+  'Content-type': 'application/json',
+}
+
+// Define a function that fetches the results of a request (query/mutation/etc)
+// and returns its results as a Promise:
+const fetchQuery = (operation: any, variables: Variables): ObservableFromValue<GraphQLResponse> => {
```
JFrankfurt, line 33: How should `operation` be typed here?

src/hooks/usePoolTickData.ts
```diff
@@ -162,9 +155,7 @@ function useTicksFromSubgraph(
         )
       : undefined
 
-  return useAllV3TicksQuery(poolAddress ? { poolAddress: poolAddress?.toLowerCase(), skip: 0 } : skipToken, {
-    pollingInterval: ms`30s`,
-  })
+  return useAllV3TicksQuery(poolAddress, 0)
```
JFrankfurt, line 158: looks like we're loosing a polling feature here. Is there a straightforward way to preserve this?

codegen.yml
```diff
@@ -2,10 +2,6 @@ overrideExisting: true
 schema: 'https://api.thegraph.com/subgraphs/name/uniswap/uniswap-v3'
 documents: 'src/**/!(*.d).{ts,tsx}'
 generates:
-  ./src/state/data/generated.ts:
+  ./schema/schema.graphql:
```
JFrankfurt, line 5: The old path was intentionally omitted from version control w/ .gitignore

It's generally best-practice to omit generated code from git. You should update the path to ignore the generated files here as well.

package.json
cartcrom: Relay generates the types, so now we only need codegen to grab the schema

src/fetchGraphQL.ts
cartcrom: Comment next to the name variable in that example says "//used by graphql mock on tests", I don't think the backend has any use of a name variable, that repo probably just has mocks implemented that return based on name

src/hooks/graphql/AllV3TicksQuery.ts
```diff
@@ -0,0 +1,52 @@
...
+`
+
+type TickData = { error: any | null; isLoading: boolean; data: AllV3TicksQuery$data | null }
+
+export default function useAllV3TicksQuery(poolAddress: string | undefined, skip: number, interval: number) {
+  const [data, setData] = useState<TickData>({ error: null, isLoading: true, data: null })
+  const chainId = useAppSelector((state) => state.application.chainId)
+  const environment = useRelayEnvironment()
+
+  const refreshData = useCallback(() => {
```
vm, line 30: don't think you need to `useCallback` something that's passed into `useEffect`, since it's gonna get re-run anyways. that's for making sure the value doesn't change as a prop.

src/fetchGraphQL.ts
```diff
@@ -0,0 +1,53 @@
...
+
+const headers = {
+  Accept: 'application/json',
+  'Content-type': 'application/json',
+}
+
+// Define a function that fetches the results of a request (query/mutation/etc)
+// and returns its results as a Promise:
+const fetchQuery = (params: RequestParameters, variables: Variables): ObservableFromValue<GraphQLResponse> => {
+  // TODO: figure out why this returns null
```
JFrankfurt, line 34: is this todo still outstanding?

src/fetchGraphQL.ts
```diff
@@ -0,0 +1,53 @@
...
+  const chainId = (store.getState() as AppState).application.chainId
+
+  const subgraphUrl = CHAIN_SUBGRAPH_URL[chainId ?? SupportedChainId.MAINNET]
```
JFrankfurt, line 37: what happens when chainid is a value that The Graph doesn't support? (i.e., optimistic kovan?)

src/hooks/graphql/AllV3TicksQuery.ts
```diff
@@ -0,0 +1,52 @@
...
+      price1
+    }
+  }
+`
+
+type TickData = { error: any | null; isLoading: boolean; data: AllV3TicksQuery$data | null }
+
+export default function useAllV3TicksQuery(poolAddress: string | undefined, skip: number, interval: number) {
+  const [data, setData] = useState<TickData>({ error: null, isLoading: true, data: null })
+  const chainId = useAppSelector((state) => state.application.chainId)
```
JFrankfurt, line 27: double checking--is this the right chainId to use?

src/hooks/graphql/AllV3TicksQuery.ts
```diff
@@ -0,0 +1,52 @@
...
+      })
+    }
+  }
+
+  useEffect(refreshData, [refreshData, poolAddress, skip])
+
+  useInterval(
+    () => {
+      refreshData()
+    },
```
JFrankfurt, line 47: p sure you don't need to wrap this call in the anon fn. Can just be `useInterval(refreshData, interval, true)`

src/hooks/graphql/FeeTierDistributionQuery.ts
```diff
@@ -0,0 +1,73 @@
...
+type FeeTierDistribitionData = { error: any | null; isLoading: boolean; data: FeeTierDistributionQuery$data | null }
+
+export default function useFeeTierDistributionQuery(
+  token0: string | undefined,
+  token1: string | undefined,
+  interval: number
+) {
+  const [data, setData] = useState<FeeTierDistribitionData>({ error: null, isLoading: true, data: null })
+  const environment = useRelayEnvironment()
+  const chainId = useAppSelector((state) => state.application.chainId)
```
JFrankfurt, line 49: same q re: appropriate chainId

src/hooks/graphql/FeeTierDistributionQuery.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  useEffect(refreshData, [refreshData, token0, token1])
+
+  useInterval(
+    () => {
+      refreshData()
+    },
+    interval,
+    true
+  )
```
JFrankfurt, line 71: same comment on the anon fn wrapper

src/utils/computeSurroundingTicks.ts
```diff
@@ -10,7 +10,7 @@ export default function computeSurroundingTicks(
   token0: Token,
   token1: Token,
   activeTickProcessed: TickProcessed,
-  sortedTickData: TickData[],
+  sortedTickData: readonly TickData[],
```
JFrankfurt, line 13: what's the purpose of this change?

JFrankfurt (changes requested):

src/fetchGraphQL.ts
cartcrom: When on an unsupported chain, the pools page will not render inputs that lead to the API being called. This functions the same way the redux data store's calls functioned before, but I'll add an additional check to the function

src/hooks/graphql/AllV3TicksQuery.ts
cartcrom: Oh duh, thanks

src/utils/computeSurroundingTicks.ts
cartcrom: relay's generated data type AllV3TicksQuery$data returns a readonly. I deleted this type and refactored to use the generated types instead

JFrankfurt (approved):

src/RelayEnvironment.ts
```diff
@@ -0,0 +1,9 @@
+import { Environment, Network, RecordSource, Store } from 'relay-runtime'
```
vm, line 1: wonder if all this should live under `src/relay` rather than spread out

vm: Screen Shot 2022-08-16 at 2 52 55 PM

PR #4322, opened 2022-08-10
fix: update time option menu for token details and colors to match theme file

kayleegeorge:
Update time selector based on figma, rename backgroundAction to backgroundInteractive and backgroundContainer to backgroundModule

before:
Screen Shot 2022-08-10 at 4 14 01 PM

after:
Screen Shot 2022-08-10 at 4 13 36 PM


cartcrom:
@kayleegeorge  I think the buttons should be a little taller, they look a little squished compared to figma:
image

You have the right 6px padding, but might need to increase the button text's line height

src/components/Charts/PriceChart.tsx
```diff
@@ -83,18 +83,28 @@ const ArrowCell = styled.div`
...
-  height: 36px;
+  background-color: ${({ theme, active }) => (active ? theme.backgroundInteractive : 'transparent')};
+  font-weight: 600;
+  font-size: 16px;
+  padding: 6px 12px;
   border-radius: 12px;
+  line-height: 20px;
   border: none;
   cursor: pointer;
   color: ${({ theme }) => theme.textPrimary};
```
cartcrom, line 110: Color should be text secondary if button is not selected

cartcrom (approved):
looks good!

PR #4323, opened 2022-08-10
feat: use filterTimeAtom for PriceChart

vm:
Defaults to the selected time period for the price chart data! Ideally, I'd like to consolidate the time period display stuff later on.


cartcrom (approved):
Looks perfect

PR #4326, opened 2022-08-11
feat: Enable queries for mixed routes 

zhongeric:
Adding `Protocol.MIXED` to the `protocols` array that is sent to the routing-api and clientSide smart order router.

Essentially a feature flag for the mixed routes launch


vm:
I think this PR can just be `forceMixedRoutes: true` on the routing api config?

zhongeric:
`forceMixedRoutes` should only be used for testing. That forces every single swap to use a mixed route. By including it in the protocols array, we give the router the option to select a mixed route if it's better than others.

zzmp:
> This will fail until routing-api is deployed to prod because of query param change

does this mean that turning it off on the server will cause the client to crash?

zhongeric:
> > This will fail until routing-api is deployed to prod because of query param change
> 
> does this mean that turning it off on the server will cause the client to crash?

Not anymore since we deployed new routing-api

vm:
can you rebase against main instead of merging? or reset and just add the last commit you need, otherwise this commit log is confusing

zzmp (approved):

vm:
Moved to new PR

PR #4329, opened 2022-08-11
refactor: removing duplicate AccountDetails code

cartcrom:
Simplied logic for displaying ENS name instead of wallet address

### DIFF
--- a/src/components/AccountDetails/index.tsx
+++ b/src/components/AccountDetails/index.tsx
@@ -129,7 +129,7 @@ const AccountControl = styled.div`
   }
 `
 
-const AddressLink = styled(ExternalLink)<{ hasENS: boolean; isENS: boolean }>`
+const AddressLink = styled(ExternalLink)`
   font-size: 0.825rem;
   color: ${({ theme }) => theme.deprecated_text3};
   margin-left: 1rem;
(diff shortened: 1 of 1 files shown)


src/components/AccountDetails/index.tsx
```diff
@@ -281,59 +281,21 @@ export default function AccountDetails({
...
-                              <Trans>View on Explorer</Trans>
-                            </span>
-                          </AddressLink>
-                        )}
-                      </div>
-                    </AccountControl>
-                  </>
-                )}
+                <AccountControl>
+                  <div>
```
vm, line 285: could we remove this extra div? i know it was there before

src/components/AccountDetails/index.tsx
```diff
@@ -129,7 +129,7 @@ const AccountControl = styled.div`
   }
 `
 
-const AddressLink = styled(ExternalLink)<{ hasENS: boolean; isENS: boolean }>`
+const AddressLink = styled(ExternalLink)`
```
vm, line 132: were we not using these params?

vm (approved):

src/components/AccountDetails/index.tsx
cartcrom: nope they were just sitting there

src/components/AccountDetails/index.tsx
cartcrom: removing it messes up the spacing for some reason

cartcrom: maybe this code should get style-updated later, there some inline stuff and weird structuring

vm: okay cool, yea feel free later, just wanted to see if it was quick

src/components/AccountDetails/index.tsx
vm: weird

PR #4330, opened 2022-08-11
test: skip fetchTokenList tests

vm:
The tests are not comprehensive and are also using actual fetches to validate responses.

I'd be open to mocking the fetch calls (there are a few) and testing the other error cases if we feel that's a good use of time.


zzmp (changes requested):
I think we should disable these tests for now (instead of deleting them), and update them in the near term to mock the fetches instead of going to the network.

vm:
I can do that

zzmp (approved):

PR #4347, opened 2022-08-14
refactor: split phase0 flag into redesign, explore, tokenSafety

vm:
Splits out the large feature flag into smaller pieces!


src/pages/Swap/index.tsx
```diff
@@ -526,7 +526,7 @@ export default function Swap() {
   return (
     <Trace page={PageName.SWAP_PAGE} shouldLogImpression>
       <>
-        {phase0Flag === Phase0Variant.Enabled ? (
+        {redesignFlag === RedesignVariant.Enabled ? (
```
cartcrom, line 529: redesignFlagEnabled here instead

cartcrom (approved):

PR #4348, opened 2022-08-14
refactor: split ChainId useEffects

vm:
- Combines 3 effects into one (for syncing the chainId query param)
- Moves syncing query param logic into App.tsx
- Creates `useSelectChain` useCallback

### DIFF
--- a/src/components/Header/NetworkSelector.tsx
+++ b/src/components/Header/NetworkSelector.tsx
@@ -1,23 +1,16 @@
 import { Trans } from '@lingui/macro'
 import { useWeb3React } from '@web3-react/core'
-import { getConnection } from 'connection/utils'
 import { getChainInfo } from 'constants/chainInfo'
-import { CHAIN_IDS_TO_NAMES, SupportedChainId } from 'constants/chains'
-import useParsedQueryString from 'hooks/useParsedQueryString'
-import usePrevious from 'hooks/usePrevious'
+import { SupportedChainId } from 'constants/chains'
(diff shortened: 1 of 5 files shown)


src/components/Header/NetworkSelector.tsx
```diff
@@ -425,7 +328,14 @@ export default function NetworkSelector() {
             </FlyoutHeader>
             {NETWORK_SELECTOR_CHAINS.map((chainId: SupportedChainId) =>
               isChainAllowed(connector, chainId) ? (
-                <Row onSelectChain={onSelectChain} targetChain={chainId} key={chainId} />
+                <Row
+                  onSelectChain={(targetChainId: SupportedChainId) => {
+                    selectChain(targetChainId)
+                    closeModal()
```
zzmp, line 334: Should this wait until the selectChain callback resolves?

src/pages/App.tsx
```diff
@@ -114,6 +115,8 @@ export default function App() {
   useAnalyticsReporter()
   initializeAnalytics()
 
+  useSyncChainQuery()
```
zzmp, line 118: Can this still be in the network selector so that it's not global, but grouped with like logic?

zzmp (approved):
lgtm, just some minor questions

src/components/Header/NetworkSelector.tsx
vm: yeah i can change that

src/pages/App.tsx
vm: yeah i can move it back

PR #4361, opened 2022-08-15
feat: add L2 network label to Explore token row

kayleegeorge:
Add L2 network badge to token rows on Explore based on network filter dropdown

Screen Shot 2022-08-16 at 2 31 11 PM


vm (commented):
should this live in chaininfo?

kayleegeorge:
> should this live in chaininfo?

normally I'd say yes, but this is like a 'secondary logo'? so wasn't sure -- there's a different logo that is used for the network switcher for example

vm (changes requested):
Rename `Arbitrum.png` to `arbitrumCircle.png` (and for other chains)
Rename `arbitrumSecondaryLogoUrl` to `arbitrumCircleLogoUrl`

vm (approved):

PR #4365, opened 2022-08-15
feat: add a feature flagged blank navbar

cbachmeier:
Adds a placeholder file onto which to build the phase 0 navbar and toggles it using the navbar feature flag

### DIFF
--- a/src/components/NavBar/Navbar.css.ts
+++ b/src/components/NavBar/Navbar.css.ts
@@ -22,7 +22,7 @@ export const nav = style([
 export const mobileWalletContainer = style([
   sprinkles({
     position: 'fixed',
-    display: { mobile: 'flex', desktopXl: 'none' },
+    display: 'flex',
     bottom: '0',
     right: '1/2',
     marginY: '0',
--- a/src/components/NavBar/Navbar.css.ts
+++ b/src/components/NavBar/Navbar.css.ts
@@ -12,10 +12,8 @@ export const nav = style([
     borderStyle: 'solid',
   }),
   {
-    borderColor: '#00000018',
     borderWidth: '0.5px',
     backdropFilter: 'blur(24px)',
-    background: '#1B213018',
   },
 ])
 
--- a/src/components/NavBar/Navbar.css.ts
+++ b/src/components/NavBar/Navbar.css.ts
@@ -1,0 +1,34 @@
+import { style } from '@vanilla-extract/css'
+
+import { sprinkles } from '../../nft/css/sprinkles.css'
+
+export const nav = style([
+  sprinkles({
+    paddingX: '20',
+    paddingY: '12',
+    width: 'full',
+    height: '72',
+    zIndex: '2',
+    borderStyle: 'solid',
+  }),
+  {
+    borderColor: '#00000018',
+    borderWidth: '0.5px',
+    backdropFilter: 'blur(24px)',
+    background: '#1B213018',
+  },
+])
+
+export const mobileWalletContainer = style([
+  sprinkles({
+    position: 'fixed',
+    display: { mobile: 'flex', desktopXl: 'none' },
+    bottom: '0',
+    right: '1/2',
+    marginY: '0',
+    marginX: 'auto',
+  }),
+  {
+    transform: 'translate(50%,-50%)',
+  },
+])
--- a/src/components/NavBar/Navbar.tsx
+++ b/src/components/NavBar/Navbar.tsx
@@ -1,3 +1,4 @@
+import Web3Status from 'components/Web3Status'
 import { useWindowSize } from 'hooks/useWindowSize'
 
 import { breakpoints } from '../../nft/css/sprinkles.css'
@@ -8,7 +9,7 @@ const MobileNavbar = () => {
     <>
       <nav className={styles.nav} />
       <div className={styles.mobileWalletContainer}>
-        <Wallet />
+        <Web3Status />
       </div>
     </>
   )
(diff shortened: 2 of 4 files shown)


src/components/NavBar/Navbar.css.ts
```diff
@@ -0,0 +1,34 @@
+import { style } from '@vanilla-extract/css'
```
vm, line 1: We should be using `styled-components` here since this is shared code. I think using `vanilla-extract` for NFT-specific pages is okay since it's isolated and we're punting on deciding whether to migrate or not.

src/components/NavBar/Navbar.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { useWindowSize } from 'hooks/useWindowSize'
+
+import { breakpoints } from '../../nft/css/sprinkles.css'
+import * as styles from './Navbar.css'
+
+const MobileNavbar = () => {
+  return (
+    <>
+      <nav className={styles.nav} />
+      <div className={styles.mobileWalletContainer}>
```
vm, line 11: think you can just conditionally add this in `Navbar` without needing a separate `MobileNavbar`?

src/components/NavBar/Navbar.tsx
```diff
@@ -0,0 +1,28 @@
...
+        <Web3Status />
+      </div>
+    </>
+  )
+}
+
+const Navbar = () => {
+  const { width: windowWidth } = useWindowSize()
+
+  if (windowWidth && windowWidth < breakpoints.desktopXl) {
```
vm, line 21: this check should happen inside the styled component, with visibility hidden under a certain breakpoint

vm (changes requested):
commented above

src/components/NavBar/Navbar.tsx
cbachmeier: There are a lot more differences between the two when content is added. I figured it would be best to separate now for easier addition later.

src/components/NavBar/Navbar.css.ts
cbachmeier: I disagree here. Seeing as they're both just css in js solutions there's no functionality differences. NFT devs are more comfortable using the design system we've already been using, but if other engs want to re-write it in styled components I wouldn't be opposed.

src/components/NavBar/Navbar.tsx
cbachmeier: what would be the benefits of doing this over current implementation?

vm: In general, we should be handling breakpoints in CSS rather than separating them into components. say we want to have 5 different breakpoints, it would be messy to have 5 different components to support that.

vm: also technically you could hit this breakpoint on web by making the screen small enough, so it's not mobile specific

src/components/NavBar/Navbar.tsx
vm: ah in that case then i think we can leave it this way

src/components/NavBar/Navbar.css.ts
```diff
@@ -0,0 +1,34 @@
...
+  sprinkles({
+    paddingX: '20',
+    paddingY: '12',
+    width: 'full',
+    height: '72',
+    zIndex: '2',
+    borderStyle: 'solid',
+  }),
+  {
+    borderColor: '#00000018',
```
vm, line 15: can we use some type of theme file for these? otherwise it won't work in light mode i think

src/components/NavBar/Navbar.tsx
vm: i think it's already doing this right? https://github.com/Uniswap/interface/pull/4365/files#diff-366a560da382b959d44ac866e3febea7ad155874ff9bd316ad4820edceec8ef1R25

src/components/NavBar/Navbar.css.ts
cbachmeier: The lightmode for this color/component hasn't been designed yet

vm: in general can we avoid having hex colors like this though? if we can the `useTheme` or some other way to determine the color variables

vm: also chatted with @infredible, the light theme should be finalized also

cbachmeier: Our general practice has been if this is a one-off case (which is true for this border color) then you can hard code the hex, if it's used more than once you name and add it to the shared style file `src/nft/css/sprinkles.css.ts`

src/components/NavBar/Navbar.tsx
cbachmeier: We usually handle breakpoints in css however in this case there is a major design difference between the navbar seen on mobile vs desktop where several tabs are removed, the chain switcher swaps sides, the wallet drops to the bottom, and the searchbar becomes a collapsible icon. This is best seen on the full phase 0 navbar PR. This is why we did this major split between the two components.

src/components/NavBar/Navbar.css.ts
infredible: Here are [the designs](https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-App-Spec-(Phase-0-%26-Phase-1)-%5BWIP%5D?node-id=4760%3A101171) for light/dark mode in the nav bar

In these updated designs, we shouldn't be using any hard-coded one-off colors

vm (approved):

PR #4366, opened 2022-08-16
feat: send web vitals

zzmp:
Creates a new event, `Web Vitals`, with properties for each web vital (CLS, FCP, FID, LCP). Sends each vital in a separate call because they are available at different times, through callbacks. This may result in multiple events, but [should be parseable in Amplitude using the median](https://community.amplitude.com/instrumentation-and-data-management-57/exploiting-event-properties-when-the-event-contains-several-properties-to-be-reported-core-web-vitals-292).

### DIFF
--- a/src/components/AmplitudeAnalytics/constants.ts
+++ b/src/components/AmplitudeAnalytics/constants.ts
@@ -23,6 +23,7 @@ export enum EventName {
   TOKEN_SELECTOR_OPENED = 'Token Selector Opened',
   WALLET_CONNECT_TXN_COMPLETED = 'Wallet Connect Transaction Completed',
   WALLET_SELECTED = 'Wallet Selected',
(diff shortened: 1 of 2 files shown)


lynnshaoyu (approved):

src/components/AmplitudeAnalytics/constants.ts
```diff
@@ -23,6 +23,7 @@ export enum EventName {
   TOKEN_SELECTOR_OPENED = 'Token Selector Opened',
   WALLET_CONNECT_TXN_COMPLETED = 'Wallet Connect Transaction Completed',
   WALLET_SELECTED = 'Wallet Selected',
+  WEB_VITALS = 'Web Vitals',
```
vm, line 26: This should be `Web Vitals Measured` or something, not to be too picky

PR #4367, opened 2022-08-16
feat(widget): light and dark themes

JFrankfurt:
image
image

### DIFF
--- a/src/theme/token-details-widget-theme.ts
+++ b/src/theme/token-details-widget-theme.ts
@@ -1,0 +1,34 @@
+import { colors } from './colors'
+
+export const LIGHT_THEME = {
+  // surface
+  container: '#EDEFF7',
+  interactive: '#EDEFF7',
+  module: '#E1E4EE',
+  accent: colors.pink400,
+  dialog: '#E1E4EE',
+  // text
+  primary: colors.gray900,
+  secondary: colors.gray500,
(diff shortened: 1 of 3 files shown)


src/theme/token-details-widget-theme.ts
```diff
@@ -0,0 +1,34 @@
+import { colors } from './colors'
+
+export const LIGHT_THEME = {
```
vm, line 3: I think if you do `useTheme`, might get the dark/light theme switching for free?

JFrankfurt: That was my first thought too, but `useTheme` gives you the current context value for the `styled-components` theme, and we need the (significantly different) widget theme here only.

vm: ah okay, np

vm: similar comment around whether this is token-details specific as below

src/utils/token-details-widget-config.ts
```diff
@@ -0,0 +1,8 @@
+import { SupportedChainId } from 'constants/chains'
```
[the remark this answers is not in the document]
JFrankfurt: **distant** future we might replace the /swap interface with this

vm (approved):
personally recommend renaming to rm token-details, but looks good to me!

PR #4368, opened 2022-08-16
feat: Wallet p0

aballerr:

- feat: A new feature
- fix: A bug fix

Example commit messages:

- feat:  Adds in a wallet to match the new wallet designs

Designs can be found here
https://www.figma.com/file/bCN7G6oArm7pPts03jvJ5x/Unification?node-id=8595%3A81590


src/components/NavBar/Navbar.css.ts
```diff
@@ -0,0 +1,34 @@
+import { style } from '@vanilla-extract/css'
```
aballerr, line 1: ignore all navbar changes, this was just to get the wallet showing

src/pages/App.tsx
```diff
@@ -5,6 +5,7 @@ import Loader from 'components/Loader'
 import TopLevelModals from 'components/TopLevelModals'
```
aballerr, line 5: ignore this file, not part of pr

src/components/WalletDropdown/DefaultMenu.tsx
```diff
@@ -0,0 +1,262 @@
...
+    explorer,
+  } = getChainInfoOrDefault(chainId ? chainId : SupportedChainId.MAINNET)
+
+  const { address: parsedAddress } = useENS(account)
+  const unclaimedAmount: CurrencyAmount<Token> | undefined = useUserUnclaimedAmount(parsedAddress)
+  const connectionType = getConnection(connector).type
+  const nativeCurrency = useNativeCurrency()
+  const nativeCurrencyPrice = useStablecoinPrice(nativeCurrency ?? undefined) || 0
+
+  const amountUSD = useMemo(() => {
```
aballerr, line 119: is there a hook that does this for me, I couldn't find it

src/components/WalletDropdown/TransactionMenu.tsx
```diff
@@ -0,0 +1,11 @@
+import { Trans } from '@lingui/macro'
```
aballerr, line 1: this is a placeholder for now, will open another pr with all transaction logic later

src/components/WalletDropdown/DefaultMenu.tsx
```diff
@@ -0,0 +1,258 @@
+import { Trans } from '@lingui/macro'
```
aballerr, line 1: Wallet went through a number of revisions, code bases, styling, etc....

I tried to use all uni stuff since this is a uni component.  Please leave as much feedback to simplify and improve to match uni standards so I can do better in future. thanks

src/components/Header/index.tsx
```diff
@@ -347,6 +350,11 @@ export default function Header() {
             ) : null}
             <Web3Status />
           </AccountElement>
+          {walletFlag === WalletVariant.Enabled && (
+            <div style={{ position: 'absolute', top: 75, right: 20 }}>
```
vm, line 354: Can you make a styled component for this div?

src/components/WalletDropdown/DefaultMenu.tsx
```diff
@@ -0,0 +1,258 @@
...
+            <Trans>Claim</Trans> {unclaimedAmount?.toFixed(0, { groupSeparator: ',' } ?? '-')} <Trans>reward</Trans>
+          </UNIbutton>
+        )}
+      </Column>
+    </>
+  )
+}
+
+const PendingBadge = styled.span`
+  background-color: #4c82fb3d;
```
vm, line 180: you'll want to use the theme colors from `useTheme`. That handles dark/light mode for you

src/components/WalletDropdown/DefaultMenu.tsx
```diff
@@ -0,0 +1,258 @@
...
+  color: #869eff;
+  font-weight: 600;
+  padding: 4px 6px;
```
vm, line 183: should avoid any px units that aren't a multiple of 4

src/components/WalletDropdown/DefaultMenu.tsx
```diff
@@ -0,0 +1,258 @@
...
+const IconWrap = styled.span`
+  margin-top: auto;
+  margin-bottom: auto;
+  margin-left: 4px;
+  height: 16px;
+  width: 16px;
+`
+
+const DefaultText = styled(Text)`
+  font-size: 14px;
```
vm, line 196: we have font sizes in `useTheme` as well

src/components/WalletDropdown/DefaultMenu.tsx
```diff
@@ -0,0 +1,258 @@
...
+  font-weight: 600;
+  padding: 4px 6px;
+  border-radius: 5px;
+`
+
+const IconWrap = styled.span`
+  margin-top: auto;
+  margin-bottom: auto;
+  margin-left: 4px;
+  height: 16px;
```
vm, line 191: does this need a height/width if the inner item is fixed size?

src/components/WalletDropdown/DefaultMenu.tsx
```diff
@@ -0,0 +1,258 @@
...
+        <span style={{ display: 'flex' }}>
+          <span style={{ marginTop: 'auto', marginBottom: 'auto' }}>
+            <DefaultText>{ISO}</DefaultText>
+          </span>
+          <IconWrap>
+            <ChevronRight size={16} strokeWidth={3} />
+          </IconWrap>
+        </span>
+      </ToggleMenuItem>
+      <ToggleMenuItem onClick={toggleDarkMode}>
```
vm, line 250: you'll want to add one more for feature flags

src/components/WalletDropdown/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+  background-color: ${({ theme }) => theme.backgroundSurface};
+  padding-top: 16px;
+  padding-bottom: 16px;
+`
+
+const WalletDropdown = () => {
+  const { account } = useWeb3React()
+  const [menu, setMenu] = useState<'DEFAULT' | 'LANGUAGE' | 'TRANSACTIONS'>('DEFAULT')
+  const isAuthenticated = !!account
+  const height = isAuthenticated ? '324px' : '200px'
```
vm, line 27: why this specific height? could we just not set a height here and use padding instead?

src/components/WalletDropdown/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+  top: 60px;
+  right: 70px;
+  background-color: ${({ theme }) => theme.backgroundSurface};
+  padding-top: 16px;
+  padding-bottom: 16px;
+`
+
+const WalletDropdown = () => {
+  const { account } = useWeb3React()
+  const [menu, setMenu] = useState<'DEFAULT' | 'LANGUAGE' | 'TRANSACTIONS'>('DEFAULT')
```
vm, line 25: `enum MenuState`

src/components/WalletDropdown/SlideOutMenu.tsx
```diff
@@ -0,0 +1,62 @@
...
+  children: JSX.Element | Array<JSX.Element>
+  clear?: () => void
+}) => {
+  return (
+    <Menu>
+      <BackSection>
+        <ChevronLeft style={{ cursor: 'pointer' }} onClick={close} size={24} />
+        <Text style={{ position: 'absolute', left: '50%', transform: 'translateX(-50%)' }}>{title}</Text>
+        {clear && (
+          <Text
```
vm, line 47: would like to avoid mixing rebass with styled components for now if possible, can basically make this a styled component

src/components/WalletDropdown/LanguageMenu.tsx
```diff
@@ -0,0 +1,72 @@
...
+const InternalLinkMenuItem = styled(InternalMenuItem)`
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  padding: 12px 16px;
+  justify-content: space-between;
+  text-decoration: none;
+  color: ${({ theme }) => theme.textPrimary};
+  :hover {
+    color: ${({ theme }) => theme.textPrimary};
```
vm, line 33: don't need this line

src/components/WalletDropdown/LanguageMenu.tsx
```diff
@@ -0,0 +1,72 @@
...
+    cursor: pointer;
+    text-decoration: none;
```
vm, line 35: don't need

src/components/WalletDropdown/LanguageMenu.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { SlideOutMenu } from './SlideOutMenu'
+
+const InternalMenuItem = styled(Link)`
+  flex: 1;
+  padding: 0.5rem 0.5rem;
+  color: ${({ theme }) => theme.textTertiary};
+  :hover {
+    color: ${({ theme }) => theme.textTertiary};
+    cursor: pointer;
+    text-decoration: none;
```
vm, line 20: i think you don't need this and the color under hover

src/components/WalletDropdown/IconButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+  position: absolute;
+  left: 50%;
+  top: 50%;
+  transform: translate(-50%, -50%);
+  width: 16px;
+  height: 16px;
+`
+
+interface IconButtonProps {
+  text: JSX.Element
```
vm, line 53: think you can use `React.ReactNode` here

src/components/WalletDropdown/IconButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+const IconBlock = styled.div<PropsWithChildren<BadgeProps>>`
+  background-color: ${({ theme }) => theme.backgroundInteractive};
+  border-radius: 12px;
+  display: inline-block;
+  position: relative;
+  height: 32px;
+  width: 32px;
+  color: ${({ theme }) => theme.textPrimary};
+  :hover {
+    background-color: rgba(173, 188, 255, 0.48);
```
vm, line 30: can we confirm that a special case color in needed here?

src/components/WalletDropdown/IconButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+    -webkit-transition: background-color 200ms linear;
+    cursor: pointer;
+    ${IconHoverText} {
+      opacity: 1;
+    }
+  }
+  :active {
+    background-color: ${({ theme }) => theme.backgroundSurface};
+    -webkit-transition: background-color 50ms linear;
```
vm, line 39: we have `autoprefixer` in our build process, so maybe don't need webkit specific overrides?

src/components/WalletDropdown/LanguageMenu.tsx
```diff
@@ -0,0 +1,72 @@
...
+    </InternalLinkMenuItem>
+  )
+}
+
+export const LanguageMenu = ({ close }: { close: () => void }) => {
+  const activeLocale = useActiveLocale()
+
+  return (
+    <SlideOutMenu title={<Trans>Language</Trans>} close={close}>
+      <div style={{ marginTop: 16 }}>
```
vm, line 65: styled component for this

src/components/WalletDropdown/SlideOutMenu.tsx
```diff
@@ -0,0 +1,62 @@
+import { ChevronLeft } from 'react-feather'
+import { Text } from 'rebass'
+import styled from 'styled-components/macro'
+import { colors } from 'theme/colors'
+
+const BackSection = styled.div`
+  position: relative;
+  display: flex;
+  align-items: center;
+  text-align: center;
```
vm, line 10: do you need both align-items and text-align?

src/components/WalletDropdown/SlideOutMenu.tsx
```diff
@@ -0,0 +1,62 @@
...
+  padding-left: 16px;
```
vm, line 11: padding: 0 16px

src/components/WalletDropdown/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+  border-radius: 12px;
+  width: 320px;
+  height: ${(props) => props.height};
+  display: flex;
+  flex-direction: column;
+  font-size: 16px;
+  top: 60px;
+  right: 70px;
+  background-color: ${({ theme }) => theme.backgroundSurface};
+  padding-top: 16px;
```
vm, line 19: padding: 16px 0; idea is it makes it clear that it's supposed to be fixed between the two

src/components/WalletDropdown/DefaultMenu.tsx
```diff
@@ -0,0 +1,258 @@
...
+  const allTransactions = useAllTransactions()
+  const toggleWalletModal = useToggleWalletModal()
+
+  const pendingTransactions = useMemo(
+    () => Object.values(allTransactions).filter((tx) => !tx.receipt),
+    [allTransactions]
+  )
+
+  return (
+    <div style={{ paddingLeft: 16, paddingRight: 16, width: '100%', height: '100%' }}>
```
vm, line 215: styled component for this

src/components/WalletDropdown/index.tsx
aballerr: The design team requested the height of the wallet always stay the same when switch to other menus (ie language/transactions).  So, I hard coded it here

src/components/WalletDropdown/DefaultMenu.tsx
aballerr: Screen Shot 2022-08-17 at 2 15 51 PM


Feature flags are being moved here, so I don't think we need them in wallet

src/components/WalletDropdown/LanguageMenu.tsx
```diff
@@ -0,0 +1,67 @@
...
+    <InternalLinkMenuItem active={active} onClick={onClick} to={to}>
+      <Text fontSize={16} fontWeight={400} lineHeight="24px">
+        {LOCALE_LABEL[locale]}
+      </Text>
+      {active && <Check color={theme.accentAction} opacity={1} size={20} />}
+    </InternalLinkMenuItem>
+  )
+}
+
+export const LanguageMenu = ({ close }: { close: () => void }) => {
```
vm, line 55: use `onClose` for handler props

src/components/WalletDropdown/LanguageMenu.tsx
```diff
@@ -0,0 +1,67 @@
...
+`
+
+function LanguageMenuItem({ locale, active }: { locale: SupportedLocale; active: boolean }) {
+  const { to, onClick } = useLocationLinkProps(locale)
+  const theme = useTheme()
+
+  if (!to) return null
+
+  return (
+    <InternalLinkMenuItem active={active} onClick={onClick} to={to}>
```
vm, line 46: use `isActive` for boolean props

src/components/WalletDropdown/SlideOutMenu.tsx
```diff
@@ -0,0 +1,62 @@
...
+    text-decoration: none;
+  }
+`
+
+const Menu = styled.div`
+  width: 100%;
+  height: 100%;
+  font-size: 16px;
+  overflow-y: scroll;
+  &::-webkit-scrollbar {
```
vm, line 21: rm webkit specific

src/components/WalletDropdown/SlideOutMenu.tsx
```diff
@@ -0,0 +1,62 @@
...
+    display: none;
```
vm, line 22: why display none?

src/components/WalletDropdown/SlideOutMenu.tsx
vm: unless there's a specific reason to keep this, since we don't really do this elsewhere?

src/components/WalletDropdown/SlideOutMenu.tsx
```diff
@@ -0,0 +1,62 @@
...
+export const SlideOutMenu = ({
+  children,
+  close,
+  title,
+  clear,
+}: {
+  close: () => void
+  title: JSX.Element
+  children: JSX.Element | Array<JSX.Element>
+  clear?: () => void
```
vm, line 35: similar to above, `onClear`? it's actually unclear what difference between clear and close is just by looking at the prop names.

src/components/WalletDropdown/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { opacify } from 'theme/utils'
+
+import DefaultMenu from './DefaultMenu'
+import { LanguageMenu } from './LanguageMenu'
+import { TransactionHistoryMenu } from './TransactionMenu'
+
+const WalletWrapper = styled.div<{ height: string }>`
+  border-radius: 12px;
+  width: 320px;
+  height: ${(props) => props.height};
```
vm, line 13: the heights should be set in here with a prop specifying which version to use. so `authenticated ? ... : ...`

vm: that's the separation of concerns between the view layer (in here) and the logic (in the component)

src/components/WalletDropdown/index.tsx
vm: but why is it larger if authenticated vs not? i think that's a bit brittle because it would need to change if the ui changed at all. generally this is why i don't love fixed heights. it'd be better if you had fixed heights on the children maybe?

src/components/WalletDropdown/DefaultMenu.tsx
```diff
@@ -0,0 +1,264 @@
...
+          <div style={{ display: 'flex' }}>
+            <StatusIcon connectionType={connectionType} size={24} />
+            <Text fontSize={16} fontWeight={600} marginTop="2.5px">
+              {account && shortenAddress(account, 2, 4)}
+            </Text>
+          </div>
+        </div>
+        <IconContainer>
+          <IconButton onClick={copy} Icon={Copy} text={isCopied ? <Trans>Copied!</Trans> : <Trans>Copy</Trans>} />
+          <IconButton
```
vm, line 144: can you make this an <a> rather than having an onclick to open window?

vm: `<a>` my formatting broke sorry

src/components/WalletDropdown/DefaultMenu.tsx
```diff
@@ -0,0 +1,263 @@
...
+  & > a:last-child {
+    margin-right: 0px;
+    ${IconHoverText}:last-child {
+      left: 0px;
+    }
+  }
+`
+
+const Divider = styled.div`
+  border: 0.5px solid ${({ theme }) => theme.backgroundOutline};
```
vm, line 75: somewhat concerned with using a .5px border, could you just confirm with fred that this is expected?

src/components/WalletDropdown/DefaultMenu.tsx
```diff
@@ -0,0 +1,263 @@
...
+  align-items: center;
+  padding: 8px 0px;
+  justify-content: space-between;
+  font-size: 14px;
+  font-weight: 400;
+  width: 100%;
+  margin-bottom: 8px;
+  color: ${({ theme }) => theme.textSecondary};
+  :hover {
+    cursor: pointer;
```
vm, line 96: cursor pointer should live outside the hover state

src/components/WalletDropdown/DefaultMenu.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+  const amountUSD = useMemo(() => {
+    const price = parseFloat(nativeCurrencyPrice.toFixed(5))
+    const balance = parseFloat(balanceString || '0')
+    return price * balance
+  }, [balanceString, nativeCurrencyPrice])
+
+  return (
+    <>
+      <div style={{ marginBottom: 12 }}>
```
vm, line 136: can't tell if this was just copied from existing stuff, but otherwise should use styled-components for these inline styles. might be able to consolidate the parent divs a bit but i didn't look closely

src/components/WalletDropdown/DefaultMenu.tsx
```diff
@@ -0,0 +1,263 @@
...
+            <Text fontSize={16} fontWeight={600} marginTop="2.5px">
+              {account && shortenAddress(account, 2, 4)}
+            </Text>
+          </div>
+        </div>
+        <IconContainer>
+          <IconButton onClick={copy} Icon={Copy} text={isCopied ? <Trans>Copied!</Trans> : <Trans>Copy</Trans>} />
+          <IconButton href={`${explorer}address/${account}`} Icon={ExternalLink} text={<Trans>Explore</Trans>} />
+          <IconButton
+            onClick={() => {
```
vm, line 149: can you do `useCallback` for this? again i know maybe it was copied but should be helpful for future readers

src/components/WalletDropdown/IconButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+  border-radius: 12px;
+  display: inline-block;
+  position: relative;
+  height: 32px;
+  width: 32px;
+  color: ${({ theme }) => theme.textPrimary};
+  :hover {
+    background-color: ${opacify(48, colors.blue200)};
+    transition: background-color 200ms linear;
+    cursor: pointer;
```
vm, line 29: same cursor pointer comment

src/components/WalletDropdown/IconButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+  top: 50%;
+  transform: translate(-50%, -50%);
+  width: 16px;
+  height: 16px;
+`
+
+interface IconButtonProps {
+  text: React.ReactNode
+  Icon: Icon
+  onClick?: () => void | undefined
```
vm, line 52: i think can just do void? lmk if not bc i'm curious

src/components/WalletDropdown/DefaultMenu.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+const BalanceWrapper = styled.div`
+  padding: 16px 0;
+`
+
+const HeaderWrapper = styled.div`
+  margin-bottom: 12px;
+`
+
+const AuthenticatedHeader = () => {
```
vm, line 126: nit: i would move this into it's own file

src/components/WalletDropdown/IconButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+interface IconButtonProps {
+  text: React.ReactNode
+  Icon: Icon
+  onClick?: () => void
+  href?: string
+}
+
+const IconButton = ({ Icon, onClick, text, href }: IconButtonProps) => (
+  <IconBlock onClick={onClick} href={href} target="_blank">
```
vm, line 56: i think we want this to be a button when it's an onClick, or an `<a>` otherwise?

src/components/WalletDropdown/LanguageMenu.tsx
```diff
@@ -0,0 +1,67 @@
...
+    <InternalLinkMenuItem isActive={isActive} onClick={onClick} to={to}>
+      <Text fontSize={16} fontWeight={400} lineHeight="24px">
+        {LOCALE_LABEL[locale]}
+      </Text>
+      {isActive && <Check color={theme.accentAction} opacity={1} size={20} />}
+    </InternalLinkMenuItem>
+  )
+}
+
+export const LanguageMenu = ({ onClose }: { onClose: () => void }) => {
```
vm, line 55: export default

src/components/WalletDropdown/SlideOutMenu.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export const SlideOutMenu = ({
+  children,
+  onClose,
+  title,
+  onClear,
+}: {
+  onClose: () => void
+  title: JSX.Element
+  children: JSX.Element | Array<JSX.Element>
```
vm, line 48: i think ReactChild is a thing?

src/components/WalletDropdown/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+import styled from 'styled-components/macro'
+
+import DefaultMenu from './DefaultMenu'
+import { LanguageMenu } from './LanguageMenu'
+import { TransactionHistoryMenu } from './TransactionMenu'
+
+const WalletWrapper = styled.div<{ isAuthenticated: boolean }>`
+  border-radius: 12px;
+  width: 320px;
+  height: ${({ isAuthenticated }) => (isAuthenticated ? '324px' : '200px')};
```
vm, line 12: I think we want to set the height of everything except the Authenticated section as 200. then the authenticated header has a height of 124px. this way you won't need a conditional here

vm (commented):
can ignore nits if you want, but only a couple other things. looks great

src/components/WalletDropdown/IconButton.tsx
aballerr: I'll split them up.  I'm going to use a div though, button requires extra styling and for some reason it doesn't align with the other elements.

src/components/WalletDropdown/index.tsx
aballerr: I had to change this slightly just now to account for when the claim button renders.  I think what I have here is the simplest given the requirements that the wallet always stays the same height when switching between the various menus.  Otherwise i have to keep track of height in multiple components

src/components/WalletDropdown/index.tsx
```diff
@@ -10,7 +10,7 @@ import { TransactionHistoryMenu } from './TransactionMenu'
 const WalletWrapper = styled.div<{ isAuthenticated: boolean; hasUnclaimed: boolean }>`
   border-radius: 12px;
   width: 320px;
-  height: ${({ isAuthenticated, hasUnclaimed }) => (isAuthenticated ? (hasUnclaimed ? ' 376px' : '324px') : '200px')};
+  max-height: 376px;
```
vm, line 11: there are two max-heights now btw

src/components/WalletDropdown/AuthenticatedHeader.tsx
```diff
@@ -0,0 +1,152 @@
...
+  display: inline-block;
+  margin-top: 4px;
+`
+
+const BalanceWrapper = styled.div`
+  padding: 16px 0;
+`
+
+const HeaderWrapper = styled.div`
+  margin-bottom: 12px;
```
vm, line 82:
```
display: flex;
```

then remove `float: right` on IconContainer i think?

vm (approved):

PR #4372, opened 2022-08-16
feat: remove v2, v3 swap router

vm:
Removes usage of v2 and v3 routers in favor of just the new swap router. Also removes usage of V2Trade and V3Trade to just use Trade.


src/constants/addresses.ts
```diff
@@ -34,19 +34,6 @@ export const V3_CORE_FACTORY_ADDRESSES: AddressMap = {
...
-    SupportedChainId.ARBITRUM_ONE,
-    SupportedChainId.ARBITRUM_RINKEBY,
-    SupportedChainId.POLYGON,
-    SupportedChainId.POLYGON_MUMBAI,
-  ]),
-  [SupportedChainId.CELO]: CELO_V3_ROUTER_ADDRESS,
-  [SupportedChainId.CELO_ALFAJORES]: CELO_V3_ROUTER_ADDRESS,
-}
-
 export const V3_MIGRATOR_ADDRESSES: AddressMap = {
```
NoahZinsmeister, line 37: oh unrelated, but seeing this reminds that we can probably remove (or at the very least lazily load) the entire migration flow

src/constants/addresses.ts
```diff
@@ -34,19 +34,6 @@ export const V3_CORE_FACTORY_ADDRESSES: AddressMap = {
...
-export const V3_ROUTER_ADDRESS: AddressMap = {
-  ...constructSameAddressMap('0xE592427A0AEce92De3Edee1F18E0157C05861564', [
-    SupportedChainId.OPTIMISM,
-    SupportedChainId.OPTIMISTIC_KOVAN,
-    SupportedChainId.ARBITRUM_ONE,
-    SupportedChainId.ARBITRUM_RINKEBY,
-    SupportedChainId.POLYGON,
-    SupportedChainId.POLYGON_MUMBAI,
-  ]),
-  [SupportedChainId.CELO]: CELO_V3_ROUTER_ADDRESS,
```
NoahZinsmeister, line 46: this variable should be renamed `CELO_SWAP_ROUTER_ADDRESS` - i verified that it's SwapRouter02, which is v2 + v3, and even though celo doesn't have v2, we decided for calldata parity it's worthwhile to use the same code

src/pages/Swap/index.tsx
```diff
@@ -305,13 +303,8 @@ export default function Swap() {
...
   )
 
-  const approvalOptimizedTrade = useApprovalOptimizedTrade(trade, allowedSlippage)
-  const approvalOptimizedTradeString =
-    approvalOptimizedTrade instanceof V2Trade
-      ? 'V2SwapRouter'
-      : approvalOptimizedTrade instanceof V3Trade
-      ? 'V3SwapRouter'
-      : 'SwapRouter'
+  const approvalOptimizedTrade = trade
```
NoahZinsmeister, line 306: should remove and rename all uses of `approvalOptimizedTrade`

src/pages/Swap/index.tsx
```diff
@@ -305,13 +303,8 @@ export default function Swap() {
...
+  const approvalOptimizedTradeString = 'SwapRouter'
```
NoahZinsmeister, line 307: nit: move to outermost scope and maybe add comment for future readers

NoahZinsmeister (approved):
lgtm!

PR #4374, opened 2022-08-16
feat: add routing functionality to navbar

cbachmeier:
- Adds tabs to route to the different pages in phase 0. Make sure you have the explore feature flag toggle on to see that path. 
- Adds sidebar menu that holds routes when on mobile and other routes such as socials, info, and governance

### DIFF
--- a/src/components/NavBar/NavDropdown.tsx
+++ b/src/components/NavBar/NavDropdown.tsx
@@ -1,0 +1,37 @@
+import { Box } from 'nft/components/Box'
+import { ReactNode } from 'react'
+
+import * as styles from './NavDropdown.css'
+
+interface NavDropdownProps {
+  top: number
+  right?: number
+  leftAligned?: boolean
+  horizontalPadding?: boolean
+  centerHorizontally?: boolean
+  children: ReactNode
+}
+
+export const NavDropdown = ({
+  top,
+  centerHorizontally,
+  leftAligned,
+  horizontalPadding,
+  children,
+}: NavDropdownProps) => {
+  return (
+    <Box
+      paddingX={horizontalPadding ? '16' : undefined}
+      style={{
+        top: `${top}px`,
+        left: centerHorizontally ? '50%' : leftAligned ? '0px' : 'auto',
+        right: centerHorizontally || leftAligned ? 'auto' : '10px',
+        transform: centerHorizontally ? 'translateX(-50%)' : 'unset',
+        zIndex: 3,
+      }}
+      className={styles.NavDropdown}
+    >
+      {children}
+    </Box>
+  )
+}
--- a/src/components/NavBar/Navbar.tsx
+++ b/src/components/NavBar/Navbar.tsx
@@ -60,16 +60,17 @@ const MobileNavbar = () => {
 const Navbar = () => {
   const { width: windowWidth } = useWindowSize()
   const { pathname } = useLocation()
+
+  if (windowWidth && windowWidth < breakpoints.desktopXl) {
+    return <MobileNavbar />
+  }
+
   const isPoolActive =
     pathname.startsWith('/pool') ||
     pathname.startsWith('/add') ||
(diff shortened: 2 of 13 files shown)


src/components/NavBar/Navbar.tsx
```diff
@@ -1,28 +1,106 @@
...
 import * as styles from './Navbar.css'
 
+interface NavLinkItemProps {
+  href: string
+  id?: NavLinkProps['id']
+  isActive?: boolean
+  children: ReactNode
+}
+
+const NavLinkItem = ({ href, id, isActive, children }: NavLinkItemProps) => {
```
zzmp, line 21: I don't think `close` will be propagated.

src/components/NavBar/Navbar.tsx
```diff
@@ -1,28 +1,106 @@
...
+  return (
+    <NavLink
```
zzmp, line 23: This seems like a thin wrapper - could you just use `NavLink` directly?

zzmp: Is this done to consolidate the styles to use menuItem/activeMenuItem? If so,rename the component to MenuItem.

src/components/NavBar/Navbar.tsx
```diff
@@ -1,28 +1,106 @@
...
         <Web3Status />
-      </div>
+      </Box>
     </>
   )
 }
 
 const Navbar = () => {
   const { width: windowWidth } = useWindowSize()
+  const { pathname } = useLocation()
```
zzmp, line 62: Move this to after the early return.

src/components/NavBar/Navbar.tsx
```diff
@@ -1,28 +1,106 @@
...
-  return <nav className={styles.nav} />
+  return (
+    <nav className={styles.nav}>
+      <Box display="flex" height="full" flexWrap="nowrap" alignItems="stretch">
+        <Box className={styles.leftSideContainer}>
+          <Box as="a" href="#/swap" className={styles.logoContainer}>
+            <UniIcon width="48" height="48" className={styles.logo} />
+          </Box>
+          <Row gap="8">
+            <NavLinkItem href={'/swap'} isActive={pathname.startsWith('/swap')}>
```
zzmp, line 82: Here and throughout - pass strings directly as `"/swap"`, not `{'/swap'}`

src/components/NavBar/NavDropdown.tsx
```diff
@@ -0,0 +1,37 @@
...
+export const NavDropdown = ({
+  top,
+  centerHorizontally,
+  leftAligned,
+  horizontalPadding,
+  children,
+}: NavDropdownProps) => {
+  return (
+    <Box
+      paddingX={horizontalPadding ? '16' : undefined}
```
zzmp, line 24: Why isn't this in the `.css.ts` file?

zzmp (changes requested):

src/components/NavBar/Navbar.tsx
cbachmeier: Could you add some context?

src/components/NavBar/Navbar.tsx
cbachmeier: Calling the hook conditionally breaks our eslint rules: `React Hook "useLocation" is called conditionally. React Hooks must be called in the exact same order in every component render.`

zzmp: 👍 . Could you move the `isPoolActive` logic after the early return, as it's part of the subsequent logic?

src/components/NavBar/NavDropdown.tsx
cbachmeier: Because as a style choice we don't pass variables, ie `horizontalPadding` in this case, into the css files. This way logic is kept in the tsx files.

PR #4375, opened 2022-08-16
refactor: removes BodyWrapper padding

JackShort:
Bodywrapper cannot have padding for phase1 nft designs due to the fact that the banner on the NFT pages need to extend to the edge of screen.


vm:
I'm unclear what this breaks within the existing app, can you post screenshots?

JackShort:
oh sorry the old version of this PR changed the padding across the board - however this new version will only affect NFT pages. Changed the title

vm (approved):
after running prettier

PR #4376, opened 2022-08-16
feat: add the phase0 chain switcher

cbachmeier:
Adds the phase0 chain switcher. Note the alignment/spacing will be off until more components get added to fill out navbar


src/components/NavBar/ChainSwitcher.tsx
```diff
@@ -0,0 +1,112 @@
...
+  )
+}
+
+const ChainDetails = ({ children }: { children: ReactNode }) => <Row>{children}</Row>
+
+const NETWORK_SELECTOR_CHAINS = [
+  SupportedChainId.MAINNET,
+  SupportedChainId.POLYGON,
+  SupportedChainId.OPTIMISM,
+  SupportedChainId.ARBITRUM_ONE,
```
vm, line 55: missing Celo

src/components/NavBar/ChainSwitcher.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  const ref = useRef<HTMLDivElement>(null)
+  useOnClickOutside(ref, isOpen ? toggleOpen : undefined)
+
+  const info = chainId ? getChainInfo(chainId) : undefined
+
+  const selectChain = useSelectChain()
+  useSyncChainQuery()
+
+  // TODO: need to handle unsupported chain
```
vm, line 74: can we handle this?

src/components/NavBar/ChainSwitcher.css.ts
```diff
@@ -0,0 +1,49 @@
+import { style } from '@vanilla-extract/css'
+
+import { sprinkles } from '../../nft/css/sprinkles.css'
+
+export const ChainSwitcher = style([
```
vm, line 5: how does light/dark mode get handled in vanilla extract?

src/components/NavBar/ChainSwitcher.tsx
```diff
@@ -0,0 +1,112 @@
+/* eslint-disable react-hooks/exhaustive-deps */
```
vm, line 1: what's this for?

src/components/NavBar/ChainSwitcher.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { NavDropdown } from './NavDropdown'
+
+const ChainRow = ({
+  targetChain,
+  onSelectChain,
+}: {
+  targetChain: SupportedChainId
+  onSelectChain: (targetChain: number) => void
+}) => {
+  const { provider, chainId } = useWeb3React()
```
vm, line 26: hm i think we can skip this check

src/components/NavBar/ChainSwitcher.css.ts
```diff
@@ -0,0 +1,49 @@
...
+    border: 'none',
+    color: 'blackBlue',
+    justifyContent: 'space-between',
+    paddingX: '16',
+    paddingY: '12',
+    cursor: 'pointer',
+  }),
+  {
+    lineHeight: '24px',
+    width: '308px',
```
vm, line 27: think we want the width to be specified by the chain switcher container, then this would be width 100%

src/components/NavBar/ChainSwitcher.css.ts
cbachmeier: Colors that have dark mode and light mode variations are defined in `themeContractValues` in the file `sprinkles.css.ts`. Then there is a `darkTheme.ts` file and a `lightTheme.ts` file which provide the hex values for the colors  defined in `themeContractValues`. Then if you just use the centrally defined variable names from the sprinkles file, for example `lightGrayContainer`, it will automatically be swapped to the appropriate hex color according to the theme.

src/components/NavBar/ChainSwitcher.css.ts
cbachmeier: this is the container for the dropdown which has a different width than ChainSwitcher modal in the navbar

src/components/NavBar/ChainSwitcher.tsx
cbachmeier: outdated and no longer needed, thanks for the catch

src/components/NavBar/ChainSwitcher.tsx
cbachmeier: Handling unsupported chains in not part of the current product spec. Additionally this is the same logic as is in `src/components/Header/NetworkSelector.tsx` ln297. I'll remove the comment since it just causes confusion

vm: ah sounds good

PR #4377, opened 2022-08-16
feat: add phase0 searchbar

cbachmeier:
- adds the phase 0 searchbar
- spacing for mobile is off as not all components have been added


src/components/NavBar/SearchBar.tsx
```diff
@@ -0,0 +1,556 @@
...
+    toggleOpen()
+  }
+
+  // Close the modal on escape
+  useEffect(() => {
+    const keyDownHandler = (event: KeyboardEvent) => {
+      if (event.key === 'Enter' && isHovered) {
+        event.preventDefault()
+        // TODO update with correct token explore URI
+        navigate(`tokens/${token.address}`)
```
vm, line 145: you'll want to do `/tokens/...` with react-router-v6

src/components/NavBar/SearchBar.tsx
```diff
@@ -0,0 +1,556 @@
...
+        event.preventDefault()
+        navigate(`/nft/collection/${collection.address}`)
+        handleClick()
+      }
+    }
+    document.addEventListener('keydown', keyDownHandler)
+    return () => {
+      document.removeEventListener('keydown', keyDownHandler)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
vm, line 67: think it's an anti-pattern to avoid this. was there a bug to list them all?

src/components/NavBar/SearchBar.tsx
```diff
@@ -0,0 +1,556 @@
...
+  }, [toggleOpen, history, isHovered, collection])
+
+  return (
+    <Row
+      as="a"
+      href={`/#/nft/collection/${collection.address}`}
```
vm, line 73: we'll want to use `react-router` link here

src/components/NavBar/SearchBar.tsx
```diff
@@ -0,0 +1,556 @@
...
+
+interface SearchBarDropdownProps {
+  toggleOpen: () => void
+  tokens: FungibleToken[]
+  collections: GenieCollection[]
+  hasInput: boolean
+}
+
+export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }: SearchBarDropdownProps) => {
+  const [hoveredIndex, setHoveredIndex] = useState<number | undefined>(undefined)
```
vm, line 295: why is hoveredIndex needed vs using :hover?

src/components/NavBar/SearchBar.tsx
```diff
@@ -0,0 +1,556 @@
...
+
+  const totalSuggestions = hasInput
+    ? tokens.length + collections.length
+    : Math.min(searchHistory.length, 2) +
+      (isNFTPage || !isTokenPage ? trendingCollections?.length ?? 0 : 0) +
+      (isTokenPage || !isNFTPage ? trendingTokens?.length ?? 0 : 0)
+
+  // Close the modal on escape
+  useEffect(() => {
+    const keyDownHandler = (event: KeyboardEvent) => {
```
vm, line 351: would love to move the keyboard hotkey stuff out since you have a few in here. perhaps a `useHotKeyHandler` hook?

src/components/NavBar/SearchBar.tsx
```diff
@@ -0,0 +1,556 @@
...
+    ['searchTokens', debouncedSearchValue],
+    () => fetchSearchTokens(debouncedSearchValue),
+    {
+      refetchOnWindowFocus: false,
+      refetchOnMount: false,
+      refetchOnReconnect: false,
+    }
+  )
+
+  const isNFTPage = location.hash.includes('/nft')
```
vm, line 464: use `useLocation` for this

src/components/NavBar/SearchBar.tsx
```diff
@@ -0,0 +1,556 @@
...
+
+export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }: SearchBarDropdownProps) => {
+  const [hoveredIndex, setHoveredIndex] = useState<number | undefined>(undefined)
+  const searchHistory = useSearchHistory(
+    (state: { history: (FungibleToken | GenieCollection)[] }) => state.history
+  ).slice(0, 2)
+  const isNFTPage = location.hash.includes('/nft')
+  const isTokenPage = location.hash.includes('/token')
+
+  const tokenSearchResults =
```
vm, line 302: is there a way to move the business logic into a utils file in this folder? hard to follow the UI logic with the algorithm mixed in

src/components/NavBar/SearchBar.tsx
```diff
@@ -0,0 +1,556 @@
...
+    {
+      refetchOnWindowFocus: false,
+      refetchOnMount: false,
+      refetchOnReconnect: false,
+    }
+  )
+
+  const isNFTPage = location.hash.includes('/nft')
+
+  // If not an nft page show up to 5 tokens, else up to 3. Max total suggestions of 8
```
vm, line 466: same comment here about business logic

src/components/NavBar/SearchBar.tsx
cbachmeier: A couple of reasons:
1. This allows us to track the user across the various sections in the suggestions dropdown including past search history, trending tokens, trending collection, token search results, and collection search results
2. This handles for both mouse and keyboard event tracking

src/components/NavBar/SearchBar.tsx
cbachmeier: We need some the keyboard logic to be localized to the suggestion row components, especially for tracking hovered/focused state, which is why we cant abstract to a generalized hook. You can read more on our decision when this was developed in this pr: https://github.com/Uniswap/interface-nft/pull/26

src/components/NavBar/SearchBar.tsx
cbachmeier: This case is just a conditional checking if there are search results otherwise show the blank state, I dont think a util function is necessary for this case

src/components/NavBar/SearchBar.tsx
cbachmeier: I've moved this to a util function

src/components/NavBar/SearchBar.tsx
```diff
@@ -0,0 +1,546 @@
...
+  const searchRef = useRef<HTMLDivElement>(null)
+  const location = useLocation()
+  const { width: windowWidth } = useWindowSize()
+
+  useOnClickOutside(searchRef, () => {
+    isOpen && toggleOpen()
+  })
+
+  // TODO NFT Search Results implmented here
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
vm, line 448: rm?

cbachmeier: This is needed for initializing a blank collection array. This won't be needed when we've added collection search results

src/components/NavBar/SearchBar.tsx
```diff
@@ -0,0 +1,546 @@
...
+  hasInput: boolean
+}
+
+export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }: SearchBarDropdownProps) => {
+  const [hoveredIndex, setHoveredIndex] = useState<number | undefined>(undefined)
+  const searchHistory = useSearchHistory(
+    (state: { history: (FungibleToken | GenieCollection)[] }) => state.history
+  ).slice(0, 2)
+  const location = useLocation()
+  const isNFTPage = location.hash.includes('/nfts')
```
vm, line 296: [`useLocation`](https://v5.reactrouter.com/web/api/Hooks/uselocation)

vm, line 296: can't we use path?

cbachmeier: whats the advantage of using path over location hash?

src/components/NavBar/SearchBar.tsx
```diff
@@ -0,0 +1,546 @@
...
+        startingIndex={isNFTPage ? 0 : tokens.length}
+        setHoveredIndex={setHoveredIndex}
+        toggleOpen={toggleOpen}
+        suggestions={collections}
+        header={'NFT Collections'}
+      />
+    ) : null
+
+  // TODO Trending NFT Results implmented here
+  const trendingCollections = [] as TrendingCollection[]
```
vm, line 326: i suggest moving this to a util called `getTrendingCollections`

src/components/NavBar/SearchBar.tsx
```diff
@@ -0,0 +1,546 @@
...
+        event.preventDefault()
+        navigate(`/tokens/${token.address}`)
+        handleClick()
+      }
+    }
+    document.addEventListener('keydown', keyDownHandler)
+    return () => {
+      document.removeEventListener('keydown', keyDownHandler)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
vm, line 151: why the exhaustive deps ignore here?

src/components/NavBar/SearchBar.tsx
```diff
@@ -0,0 +1,546 @@
...
+
+interface TokenRowProps {
+  token: FungibleToken
+  isHovered: boolean
+  setHoveredIndex: (index: number | undefined) => void
+  toggleOpen: () => void
+  index: number
+}
+
+const TokenRow = ({ token, isHovered, setHoveredIndex, toggleOpen, index }: TokenRowProps) => {
```
vm, line 125: move TokenRow to it's own file

src/components/NavBar/SearchBar.tsx
```diff
@@ -0,0 +1,546 @@
...
+          <Box className={styles.secondaryText} color={token.price24hChange >= 0 ? 'green400' : 'red400'}>
+            {token.price24hChange.toFixed(2)}%
+          </Box>
+        )}
+      </Column>
+    </Link>
+  )
+}
+
+const SkeletonRow = () => {
```
vm, line 211: skeleton row can go in the tokenrow file if you want, or a separate file as well

src/components/NavBar/SearchBar.tsx
cbachmeier: this is temporary before we add in the query for fetching collection search results

src/components/NavBar/SearchBar.tsx
```diff
@@ -0,0 +1,546 @@
...
+  toggleOpen: () => void
+  suggestions: (GenieCollection | FungibleToken)[]
+  header: string
+  headerIcon?: JSX.Element
+  hoveredIndex: number | undefined
+  startingIndex: number
+  setHoveredIndex: (index: number | undefined) => void
+}
+
+export const SearchBarDropdownSection = ({
```
vm, line 241: own file for this also

src/components/NavBar/SearchBar.tsx
vm: we may not use hash if we switch away from hashrouter, so that's implementation detail

src/components/NavBar/SearchBar.tsx
vm: that's a good reason to separate it imo, bc we would likely move most of the logic into a hook that's separated

vm: optional

src/components/NavBar/SearchBar.tsx
vm: optional

src/components/NavBar/SearchBar.tsx
vm: optional

src/components/NavBar/SearchBar.css.ts
```diff
@@ -0,0 +1,159 @@
...
+  }),
+  {
+    lineHeight: '20px',
+  },
+])
+
+export const imageHolder = style([
+  suggestionImage,
+  {
+    background: 'radial-gradient(167.86% 167.86% at -21.43% -50%, #4C82FB 0%, #09265E 100%)',
```
vm, line 125: could you confirm with fred about the hardcoded background values here?

src/components/NavBar/SearchBar.tsx
```diff
@@ -0,0 +1,545 @@
...
+            as="img"
+            src={collection.imageUrl}
+            alt={collection.name}
+            className={styles.suggestionImage}
+            onError={() => setBrokenImage(true)}
+            onLoad={() => setLoaded(true)}
+            style={{
+              background: loaded
+                ? 'none'
+                : 'radial-gradient(167.86% 167.86% at -21.43% -50%, #4C82FB 0%, #09265E 100%)',
```
vm, line 91: this seems to be the same background as in the sprinkles? can we make a variable for it

src/components/NavBar/SearchBar.tsx
vm: optional

src/components/NavBar/SearchBar.css.ts
cbachmeier: this was taken directly from the Figma

src/components/NavBar/SearchBar.tsx
cbachmeier: made the classname a conditional to condense useage

src/lib/utils/searchBar.ts
```diff
@@ -0,0 +1,20 @@
+import { FungibleToken, GenieCollection } from 'nft/types'
```
vm, line 1: nit: would bet better if this could live within the NavBar component dir imo

vm (approved):

src/lib/utils/searchBar.ts
cbachmeier: We like to have all datatypes in the `types` folder for organization. GenieCollection is used in several places outside of the NavBar

src/components/NavBar/SearchBar.css.ts
cbachmeier: Talked with design and they actually wanted to update this to be the same color as the skeleton loading state

PR #4380, opened 2022-08-17
chore: remove unused exports

vm:
I used this while working on https://github.com/Uniswap/interface/pull/4372.

`npx dead-exports`


NoahZinsmeister (approved):
maybe worth throwing this in CI

vm:
agree. will talk to zach about that

PR #4383, opened 2022-08-17
feat: adds banner section (carousel only) for NFT Explore

gbugyis:
Adds a page route to /nfts which contains the Banner components and carousel interval. Activity feed will come in a later PR.


src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,202 @@
...
+      refetchOnWindowFocus: false,
+      refetchOnMount: false,
+    }
+  )
+
+  useEffect(() => {
+    /* Rotate through Top 5 Collections on 15 second interval */
+    const interval = setInterval(async () => {
+      if (collections) {
+        const nextCollectionIndex = current === collections.length - 1 ? 0 : current + 1
```
vm, line 38: think you can do `(current + 1) % collections.length`

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,202 @@
...
+        if (!hovered) setCurrent(nextCollectionIndex)
+      }
+    }, 15000)
+    return () => clearInterval(interval)
+  }, [current, collections, hovered])
+
+  return (
+    <Box
```
vm, line 46: could this be a `Link` or does that break things?

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,202 @@
...
+      <Box>
+        {collection.volumeChange ? (
+          <Box as="span" color={collection.volumeChange > 0 ? 'green200' : 'error'} marginLeft="4">
+            {collection.volumeChange > 0 && '+'}
+            {formatChange(collection.volumeChange)}%
+          </Box>
+        ) : null}
+      </Box>
+    </Row>
+    <Box
```
vm, line 157: similarly think we want this as a Link

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+/* Carousel Progress indicators */
+const CarouselProgress = ({
+  length,
+  currentIndex,
+  setCurrent,
+}: {
+  length: number
+  currentIndex: number
+  setCurrent: Dispatch<SetStateAction<number>>
```
vm, line 177: in general suggest making this take an `onClick` instead, so it doesn't need to deal with the implementation details of the parent component. so `onClick(index: number): void`

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,202 @@
...
+            e.preventDefault()
+            e.stopPropagation()
+            setCurrent(carouselIndex)
+          }}
+          key={carouselIndex}
+        >
+          <Box
+            as="span"
+            display="inline-block"
+            className={clsx(styles.carouselIndicator, carouselIndex === currentIndex && styles.carouselIndicatorActive)}
```
vm, line 197: nit: can be easier to read as `clsx(styles.carouselIndicator, { [styles.carouselIndicatorActive]: carouselIndex === currentIndex }`

src/nft/pages/explore/index.tsx
```diff
@@ -0,0 +1,19 @@
+//import TrendingCollections from '../../components/explore/TrendingCollections'
```
vm, line 1: i think can just remove the commented stuff for now to avoid conflicts

src/pages/App.tsx
```diff
@@ -206,6 +209,7 @@ export default function App() {
                   <Route path="migrate/v2/:address" element={<MigrateV2Pair />} />
 
                   <Route path="*" element={<RedirectPathToSwapOnly />} />
+                  {phase1Flag === Phase1Variant.Enabled && <Route path="/nfts" element={<NftExplore />} />}
```
vm, line 212: could we rename `phase1` to `NFT`?

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,202 @@
...
+import { useQuery } from 'react-query'
+import { useNavigate } from 'react-router-dom'
+
+import * as styles from './Explore.css'
+
+const Banner = () => {
+  const navigate = useNavigate()
+  /* Sets initially displayed collection to random number between 1 and 5  */
+  const [current, setCurrent] = useState(Math.floor(Math.random() * 5 + 1))
+  const [hovered, toggleHover] = useReducer((state) => !state, false)
```
vm, line 21: could you explain why hovered is being controlled vs using CSS alone?

src/pages/App.tsx
cbachmeier: This is outside of the scope of this PR

src/nft/components/explore/Banner.tsx
gbugyis: the hover state (on the entire banner) is used to pause the carousel rotation, and is also passed down to trigger an active state on the ‘Explore Collection’ button (even if the button itself is not hovered, per product spec). I think trying to both of those with CSS states would end up with more convoluted code

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,195 @@
...
+  collection,
+  rank,
+  hovered,
+}: {
+  collection: TrendingCollection
+  rank: number
+  hovered: boolean
+}) => (
+  <Column className={styles.collectionDetails} paddingTop="16">
+    <div>
```
vm, line 107: nit: can you remove the inner space and put the styles on the div maybe?

src/pages/App.tsx
```diff
@@ -43,6 +43,7 @@ import Tokens from './Tokens'
 
 const TokenDetails = lazy(() => import('./TokenDetails'))
 const Vote = lazy(() => import('./Vote'))
+const NftExplore = lazy(() => import('nft/pages/explore'))
```
zzmp, line 46: `lazy` 😍

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,195 @@
...
+import { fetchTrendingCollections } from 'nft/queries'
+import { TimePeriod, TrendingCollection } from 'nft/types'
+import { formatEthPrice } from 'nft/utils/currency'
+import { putCommas } from 'nft/utils/putCommas'
+import { formatChange, toSignificant } from 'nft/utils/toSignificant'
+import { ReactNode, useEffect, useReducer, useState } from 'react'
+import { useQuery } from 'react-query'
+import { Link } from 'react-router-dom'
+
+import * as styles from './Explore.css'
```
zzmp, line 16: Do you intend to share styles across components? If not, can you rename this to `./Banner.css.ts` for consistency?

src/nft/components/explore/Explore.css.ts
```diff
@@ -0,0 +1,120 @@
...
+export const carouselIndicator = style([
+  sprinkles({
+    backgroundColor: 'explicitWhite',
+    width: '36',
+    height: '4',
+    marginRight: '6',
+    borderRadius: 'round',
+  }),
+  {
+    opacity: '0.4',
```
zzmp, line 116: I think opacity is meant to be baked into the design system - can you check in with the design team on whether this is ok to do?

src/nft/components/explore/Explore.css.ts
```diff
@@ -0,0 +1,120 @@
...
+export const volumeRank = style([
+  sprinkles({
+    paddingTop: '8',
+    paddingBottom: '8',
+    paddingRight: '16',
+    paddingLeft: '16',
+    color: 'green200',
+  }),
+  {
+    background: 'rgba(92, 254, 157, 0.12)',
```
zzmp, line 88: This should be a designated color, not a custom value.

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+const Banner = () => {
+  /* Sets initially displayed collection to random number between 1 and 5  */
```
zzmp, line 19: Why is this 1-indexed and not 0-indexed?

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,195 @@
...
+      if (collections) {
+        const nextCollectionIndex = (current + 1) % collections.length
+        if (!hovered) setCurrent(nextCollectionIndex)
+      }
+    }, 15000)
+    return () => clearInterval(interval)
+  }, [current, collections, hovered])
+
+  return (
+    <Box className={styles.fullWidth} onMouseEnter={toggleHover} onMouseLeave={toggleHover} cursor="pointer">
```
zzmp, line 46: This can get into the inverse state.
It's safer to define `hovered` using `useState` and explicitly call `setHover(true)`/`setHover(false)`.

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,195 @@
...
+      refetchOnWindowFocus: false,
+      refetchOnMount: false,
+    }
+  )
+
+  useEffect(() => {
+    /* Rotate through Top 5 Collections on 15 second interval */
+    const interval = setInterval(async () => {
+      if (collections) {
+        const nextCollectionIndex = (current + 1) % collections.length
```
zzmp, line 38: This can be inlined. Either way, the whole thing should be guarded by `!hovered`, so it's more obvious which logic is actually being executed.

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,195 @@
...
+        if (!hovered) setCurrent(nextCollectionIndex)
```
zzmp, line 39: As an async function, this should check that the component is not stale before setting any state.

      if (!stale && !hovered) {
        setCurrent((current + 1) % collections.length)
      }
    ...
    let stale = false
    return () => {
      stale = true
      clearInterval(interval)
    }

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,195 @@
...
+      }
+    }, 15000)
```
zzmp, line 41: nit: `15_000`

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,195 @@
...
+    return () => clearInterval(interval)
+  }, [current, collections, hovered])
+
+  return (
+    <Box className={styles.fullWidth} onMouseEnter={toggleHover} onMouseLeave={toggleHover} cursor="pointer">
+      {collections ? (
+        collections.map((collection: TrendingCollection, index: number) => (
+          <Link to={`/nfts/collection/${collections[current].address}`} key={index} style={{ textDecoration: 'none' }}>
```
zzmp, line 49: Shouldn't this be `collection`, and not `collections[current]`?

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,195 @@
...
+                    as="section"
+                    className={section}
+                    display="flex"
+                    flexDirection="row"
+                    flexWrap="nowrap"
+                    paddingTop="40"
+                  >
+                    <CollectionDetails collection={collection} hovered={hovered} rank={index + 1} />
+                  </Box>
+                  <CarouselProgress length={collections.length} currentIndex={index} onClick={setCurrent} />
```
zzmp, line 70: What will this set current to?

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,195 @@
...
+    return () => clearInterval(interval)
+  }, [current, collections, hovered])
+
+  return (
+    <Box className={styles.fullWidth} onMouseEnter={toggleHover} onMouseLeave={toggleHover} cursor="pointer">
+      {collections ? (
+        collections.map((collection: TrendingCollection, index: number) => (
+          <Link to={`/nfts/collection/${collections[current].address}`} key={index} style={{ textDecoration: 'none' }}>
+            <Box
+              visibility={index === current ? 'visible' : 'hidden'}
```
zzmp, line 51: Why not filter the `collections` to only include `current`, instead of rendering a bunch of hidden elements?

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+/* Carousel Progress indicators */
+const CarouselProgress = ({
+  length,
+  currentIndex,
+  onClick,
+}: {
+  length: number
+  currentIndex: number
+  onClick(index: number): void
```
zzmp, line 168: This is confusing, because it's a built-in event being overridden. Could you retain it with no signature, and bind the index when you render, so it's more clear what is happening?

zzmp (changes requested):

src/nft/components/explore/Banner.tsx
gbugyis: There will be other styles here for the value prop and activity feed, but I'll rename for now.

src/nft/components/explore/Banner.tsx
JFrankfurt: or `ms`

src/nft/pages/explore/index.tsx
```diff
@@ -0,0 +1,11 @@
+import Banner from '../../components/explore/Banner'
+
+const NftExplore = () => {
+  return (
+    <>
+      <Banner />
+    </>
+  )
+}
```
JFrankfurt, line 9: There's going to be more content here soon, right? This isn't just a passthrough wrapper?

src/nft/components/explore/Explore.css.ts
```diff
@@ -0,0 +1,120 @@
...
+    borderRadius: '64px',
+  },
+])
+
+export const exploreCollection = style([
+  {
+    width: '174px',
```
JFrankfurt, line 95: We do mod 4 px values in almost all cases--if this is spec'd to 174 it's likely a goof from design, should double check this.

src/nft/components/explore/Explore.css.ts
```diff
@@ -0,0 +1,120 @@
...
+    width: 'full',
+    backgroundColor: 'grey900',
+    left: '0',
+    top: '0',
+  }),
+])
+
+export const bannerContent = sprinkles({
+  position: 'relative',
+  zIndex: '1',
```
JFrankfurt, line 49: confirming this is necessary--if so, should prefer a theme spec'd layer

src/nft/pages/explore/index.tsx
gbugyis: Right, I'm migrating things over in pieces to have smaller PRs

src/nft/components/explore/Explore.css.ts
```diff
@@ -0,0 +1,120 @@
...
+    width: 'full',
+    maxWidth: 'full',
+    position: 'absolute',
+  }),
+  {
+    top: '72px',
+    marginLeft: '-50vw',
+    marginRight: '-50vw',
+  },
+])
```
JFrankfurt, line 18: does 100vw positioned at left:0 not work here? This looks at first glance like it's a convoluted center/fill.

src/nft/components/explore/Banner.tsx
gbugyis: I refactored the carousel progress indicators a bit so hopefully the behavior is more clear

src/nft/components/explore/Explore.css.ts
gbugyis: iirc the zIndex was necessary before the navbar changes - no longer needed

src/nft/pages/explore/index.tsx
```diff
@@ -0,0 +1,11 @@
+import Banner from '../../components/explore/Banner'
```
zzmp, line 1: This would be easier to follow as `'nft/components/export/Banner'`

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,185 @@
...
+                    />
+                  ))}
+              </Center>
+            </CollectionWrapper>
+          </Box>
+        </Link>
+      ) : (
+        <>
+          {/* TODO: Improve Loading State */}
+          <p>Loading</p>
```
zzmp, line 88: Is this spec'ed? Should it use the loading component?
https://github.com/Uniswap/interface/tree/main/src/components

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,185 @@
...
+      refetchOnReconnect: false,
+      refetchOnWindowFocus: false,
+      refetchOnMount: false,
+    }
+  )
+
+  useEffect(() => {
+    /* Rotate through Top 5 Collections on 15 second interval */
+    const interval = setInterval(async () => {
+      if (collections && !hovered && !stale) {
```
zzmp, line 37: I think it makes sense to add a guard statement with hovered and stale:
```
if (hovered || stale) return
if (collections) {
  ...
}
```

wdyt?

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,185 @@
...
+      to={`/nfts/collection/${collection.address}`}
+      style={{ textDecoration: 'none', backgroundColor: `${hovered ? vars.color.blue400 : vars.color.grey700}` }}
+    >
+      Explore collection
+    </Link>
+  </Column>
+)
+
+/* Carousel Progress Indicators */
+const CarouselIndicator = ({ active, onClick }: { active: boolean; onClick: () => void }) => (
```
zzmp, line 168: Should this be a separate component?

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,185 @@
...
+        </>
+      )}
+    </Box>
+  )
+}
+
+export default Banner
+
+/* Collection Wrapper: applies background image to entire banner */
+const CollectionWrapper = ({ bannerImageUrl, children }: { bannerImageUrl: string; children: ReactNode }) => (
```
zzmp, line 98: The CollectionWrapper should be a part of the CollectionDetails. Keeping it separate only complicates the markdown for the Banner.

zzmp (changes requested):

src/nft/components/explore/Banner.tsx
gbugyis: agree - cleaner that way

src/nft/components/explore/Banner.tsx
gbugyis: planning to add that in a future PR along with other design updates

src/nft/components/explore/Banner.tsx
```diff
@@ -2,14 +2,15 @@ import clsx from 'clsx'
 import { Box } from 'nft/components/Box'
 import { Center, Column, Row } from 'nft/components/Flex'
 import { VerifiedIcon } from 'nft/components/icons'
-import { bodySmall, buttonMedium, header1, section } from 'nft/css/common.css'
+import { bodySmall, buttonMedium, header1 } from 'nft/css/common.css'
```
zzmp, line 5: This import line should include `section`

src/nft/components/explore/Banner.tsx
```diff
@@ -33,13 +34,14 @@ const Banner = () => {
 
   useEffect(() => {
     /* Rotate through Top 5 Collections on 15 second interval */
+    let stale = false
+    if (hovered || stale) return
```
zzmp, line 38: Ideally the guard statement is the first thing in the function (before declaring `stale`), to make it easier to read. You should move `let stale = false` back above the `return`, and move the guard after the comment.

src/nft/components/explore/Banner.tsx
```diff
@@ -56,30 +58,36 @@ const Banner = () => {
...
-                <CollectionDetails collection={collections[current]} hovered={hovered} rank={current + 1} />
-              </Box>
-              <Center marginTop="16" position="relative">
+            <div
+              className={styles.bannerWrap}
+              style={{ backgroundImage: `url(${collections[current].bannerImageUrl})` }}
+            >
+              <Box className={styles.bannerOverlay} width="full" />
+              <CollectionDetails collection={collections[current]} hovered={hovered} rank={current + 1} />
+              <Center marginTop="32" position="relative">
```
zzmp, line 67: Can you refactor this whole element out as `CarouselIndicator`?

src/nft/components/explore/Banner.tsx
```diff
@@ -112,74 +112,64 @@ const CollectionDetails = ({
...
-      )}
-    </Row>
-    <Row className={bodySmall} marginTop="12" color="explicitWhite">
-      <Box>
-        <Box as="span" color="darkGray" marginRight="4">
-          Floor:
+  <Box
+    as="section"
+    className={section}
+    display="flex"
```
zzmp, line 118: Why aren't these defined in the css for `section`?

zzmp (changes requested):

src/nft/components/explore/Banner.tsx
gbugyis: then you get an eslint no-use-before-define error

src/nft/components/explore/Banner.tsx
gbugyis: `section` was previously used in other parts of `/nfts` and needed to be kept simple. Now its only used here so I moved it out and combined with these styles.

src/nft/components/explore/Banner.tsx
gbugyis: The moved out the individual indicators, but am still leaving the `.map` in the Banner component based on other feedback above.

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,175 @@
...
+      stale = true
+      clearInterval(interval)
+    }
+  }, [current, collections, hovered])
+
+  return (
+    <Box onMouseEnter={() => setHover(true)} onMouseLeave={() => setHover(false)} cursor="pointer" width="full">
+      {collections && collections[current] ? (
+        <Link to={`/nfts/collection/${collections[current].address}`} style={{ textDecoration: 'none' }}>
+          <Box style={{ height: '386px' }}>
```
JFrankfurt, line 54: - can you avoid inline styles here?
- is this height as spec'd? it's not divisible by 4

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,175 @@
...
+        <Box>
+          <Box as="span" color="darkGray" marginRight="4">
+            Floor:
+          </Box>
+          {collection.floor ? formatEthPrice(collection.floor.toString()) : '--'} ETH
+        </Box>
+        <Box>
+          {collection.floorChange ? (
+            <Box as="span" color={collection.floorChange > 0 ? 'green200' : 'error'} marginLeft="4">
+              {collection.floorChange > 0 && '+'}
```
JFrankfurt, line 110: should there be a `-` here when it's less than 0?

src/nft/components/explore/Banner.tsx
```diff
@@ -0,0 +1,175 @@
...
+              {collection.volumeChange > 0 && '+'}
+              {formatChange(collection.volumeChange)}%
+            </Box>
+          ) : null}
+        </Box>
+      </Row>
+      <Link
+        className={clsx(buttonMedium, styles.exploreCollection)}
+        to={`/nfts/collection/${collection.address}`}
+        style={{ textDecoration: 'none', backgroundColor: `${hovered ? vars.color.blue400 : vars.color.grey700}` }}
```
JFrankfurt, line 133: is this link hover behavior unique to this one usage?

JFrankfurt (approved):
some comments and questions, but overall lgtm

zzmp (approved):

src/nft/components/explore/Banner.tsx
gbugyis: API returns a negative valu
negative_floor
e

src/nft/components/explore/Banner.tsx
gbugyis: This was built against a previous Figma version (before common design system values were applied), and will be updated in a future PR along with a bunch of other layout/style updates. This PR is more of a first step towards migration than the final result.

src/nft/components/explore/Banner.tsx
gbugyis: same as above in this case - From a glance at Figma this button will no longer exist in the next iteration.

PR #4386, opened 2022-08-17
chore: rename flag explore -> tokens

vm:
For consistency, keeping the names aligned.

### DIFF
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -4,9 +4,9 @@ import { Trace } from 'components/AmplitudeAnalytics/Trace'
 import Loader from 'components/Loader'
 import TopLevelModals from 'components/TopLevelModals'
 import { useFeatureFlagsIsLoaded } from 'featureFlags'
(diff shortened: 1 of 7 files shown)


src/pages/App.tsx
```diff
@@ -156,7 +156,7 @@ export default function App() {
             <Suspense fallback={<Loader />}>
               {isLoaded ? (
                 <Routes>
-                  {exploreFlag === ExploreVariant.Enabled && (
+                  {tokensFlag === TokensVariant.Enabled && (
                     <>
                       <Route path="/tokens" element={<Explore />} />
```
kayleegeorge, line 161: should we change the component name too here

vm: yes separately

cbachmeier (commented):
Can you also have the links in the phase0 navbar now route to /tokens instead of /explore?

kayleegeorge (approved):

PR #4388, opened 2022-08-17
chore: rename Explore directory to Tokens

kayleegeorge:
rename Explore directory and page to Tokens


src/components/Charts/PriceChart.tsx
```diff
@@ -3,7 +3,7 @@ import { localPoint } from '@visx/event'
 import { EventType } from '@visx/event/lib/types'
```
vm, line 3: this should live under `Tokens`

src/components/Tokens/TokenDetails/PriceChart.tsx
```diff
@@ -22,8 +22,8 @@ import {
   weekFormatter,
 } from 'utils/formatChartTimes'
 
-import data from './data.json'
-import LineChart from './LineChart'
+import data from '../../Charts/data.json'
+import LineChart from '../../Charts/LineChart'
```
vm, line 26: use a fixed import not a relative import for this

vm (approved):

PR #4391, opened 2022-08-17
feat: adding collection stats to collection page

JackShort:
* adds collection stats to the collection page
* note currently we have the code split between mobile and desktop that might change in the future if I see that its possible, until then they will stay separated

### Goal is to import entire collection page - breakdown is as follows
1. pr for routing (done)
2. pr that imports collection page info (this)
3. pr that imports the assets
4. pr that imports filters

access the collection page by going to `nfts/collection/<NFT_CONTRACT_ADDRESS>?chain=mainnet`

`nfts/collection/0xbc4ca0eda7647a8ab7c2061c2e118a18a936f13d?chain=mainnet`


src/nft/pages/collection/index.tsx
```diff
@@ -1,5 +1,35 @@
...
+
+import { useIsMobile } from '../../hooks/useIsMobile'
+import { CollectionStatsFetcher } from '../../queries'
+import { CollectionDesktop } from './CollectionDesktop'
+import { CollectionMobile } from './CollectionMobile'
+
+const Collection = () => {
+  const { contractAddress } = useParams()
+
+  const isMobile = useIsMobile((state) => state.isMobile)
```
zzmp, line 15: You can elide L15-16 into one line: `const { isMobile, setMobileWidth } = useIsMobile()`

src/nft/pages/collection/index.tsx
```diff
@@ -1,5 +1,35 @@
...
+  const setMobileWidth = useIsMobile((state) => state.setMobileWidth)
+  const location = useLocation()
+  const isActivityToggled = useMemo(() => location.pathname.includes('/activity'), [location])
```
zzmp, line 18: It's not obvious from the code what this is for. Can you either move this into the desktop and mobile components (so it is adjacent to the code for toggling activity), or define the callback to toggle activity in this file?

Otherwise, by splitting it, it's harder to understand - to borrow React terminology: it's like a semi-controlled component.

src/nft/pages/collection/index.tsx
```diff
@@ -1,5 +1,35 @@
...
+    CollectionStatsFetcher(contractAddress as string)
+  )
+
+  const { width } = useWindowSize()
+  useEffect(() => {
+    setMobileWidth(width ?? 0)
+  }, [width, setMobileWidth])
+
+  if (isMobile)
+    return <CollectionMobile {...{ collectionStats: collectionStats ?? ({} as GenieCollection), isActivityToggled }} />
```
zzmp, line 30: `{}` does not satisfy `GenieCollection`, so you shouldn't coerce it, as it may lead to problems in future refactors. Instead, can you have the `CollectionStats` check for `undefined` and return `null` in that case?

src/nft/pages/collection/Components.tsx
```diff
@@ -0,0 +1,29 @@
+import { Box } from 'nft/components/Box'
+import { Row } from 'nft/components/Flex'
+import * as styles from 'nft/pages/collection/common.css'
+
+interface ActivitySwitcherProps {
+  showActivity: boolean
+  toggleActivity: () => void
+}
+
+export const ActivitySwitcher = ({ showActivity, toggleActivity }: ActivitySwitcherProps) => {
```
zzmp, line 10: This file should be named `ActivitySwitcher`.

As mentioned in an earlier comment, the definitions for showActivity and toggleActivity should be adjacent for clarity. Another option for that is to make this an uncontrolled component, and have it define those things itself.

zzmp, line 10: This doesn't seem so much like an activity switcher as a general tab manager. Should it be refactored to reflect that?

src/nft/pages/collection/CollectionMobile.tsx
```diff
@@ -0,0 +1,33 @@
...
+      </Box>
+
+      <Row paddingLeft="32" paddingRight="32">
+        <CollectionStats stats={collectionStats} isMobile={true} />
+      </Row>
+      <Row paddingLeft="32" paddingRight="32">
+        <ActivitySwitcher showActivity={isActivityToggled} toggleActivity={() => undefined} />
+      </Row>
+      <Row alignItems="flex-start" position="relative" paddingLeft="32" paddingRight="32">
+        {/* // @ts-ignore */}
```
zzmp, line 28: What is being ignored here? Can this line be removed?

src/nft/pages/collection/CollectionDesktop.tsx
```diff
@@ -0,0 +1,33 @@
...
+      </Box>
+
+      <Row paddingLeft="32" paddingRight="32">
+        <CollectionStats stats={collectionStats} isMobile={false} />
+      </Row>
+      <Row paddingLeft="32" paddingRight="32">
+        <ActivitySwitcher showActivity={isActivityToggled} toggleActivity={() => undefined} />
+      </Row>
+      <Row alignItems="flex-start" position="relative" paddingLeft="32" paddingRight="32">
+        {/* // @ts-ignore */}
```
zzmp, line 28: What is being ignored here? Can this line be removed?

src/nft/components/collection/CollectionStats.tsx
```diff
@@ -0,0 +1,328 @@
...
+        {totalSupplyStr}
+      </StatsItem>
+      {numOwnersStr ? (
+        <StatsItem label="Owners" isMobile={isMobile ?? false}>
+          {numOwnersStr}
+        </StatsItem>
+      ) : null}
+      {stats.floorPrice ? (
+        <StatsItem label="Floor Price" isMobile={isMobile ?? false}>
+          {floorPriceStr} ETH
```
zzmp, line 262: Is this only meant for mainnet (with ETH hardcoded), or should it be updated to reflect the native token for any chain?

src/nft/components/layout/Marquee.tsx
```diff
@@ -0,0 +1,59 @@
+import { Box } from 'nft/components/Box'
+import * as styles from 'nft/components/layout/Marquee.css'
+import { ReactNode, useEffect, useRef, useState } from 'react'
+
+export const Marquee = ({ children, speed = 20 }: { children: ReactNode; speed?: number }) => {
+  const [containerWidth, setContainerWidth] = useState(0)
```
zzmp, line 6: The widths do not need to be included in state - they are used immediately after being computed.

src/nft/components/layout/Marquee.tsx
```diff
@@ -0,0 +1,59 @@
...
+    }
+  })
+
+  useEffect(() => {
+    setIsMounted(true)
+  }, [])
+
+  return (
+    <>
+      {!isMounted ? null : (
```
zzmp, line 47: Why is there an explicit check for mounting? This seems like an antipattern.

src/nft/components/layout/Marquee.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export const Marquee = ({ children, speed = 20 }: { children: ReactNode; speed?: number }) => {
+  const [containerWidth, setContainerWidth] = useState(0)
+  const [marqueeWidth, setMarqueeWidth] = useState(0)
+  const [duration, setDuration] = useState(0)
+  const [isMounted, setIsMounted] = useState(false)
+  const containerRef = useRef<HTMLDivElement | null>(null)
+  const marqueeRef = useRef<HTMLDivElement | null>(null)
+
+  const calculateWidth = () => {
```
zzmp, line 13: This isn't exactly what it does - can you rename it: `updateScrollDuration()`?

zzmp (changes requested):

src/nft/components/collection/CollectionStats.tsx
```diff
@@ -0,0 +1,328 @@
...
+        ref={descriptionRef}
+        className={clsx(styles.description, styles.nameText, readMore && styles.descriptionOpen)}
+      >
+        <ReactMarkdown
+          source={description}
+          allowedTypes={['link', 'paragraph', 'strong', 'code', 'emphasis', 'text']}
+          renderers={{ paragraph: 'span' }}
+        />
+      </Box>
+      <Box as="span" display={showReadMore ? 'inline' : 'none'} className={styles.readMore} onClick={toggleReadMore}>
```
zzmp, line 225: Should the entire description be the click target here?

src/nft/pages/collection/CollectionMobile.tsx
JackShort: there is a bug in tslint that causes it to show an error about like infinite props for this component

src/nft/pages/collection/CollectionDesktop.tsx
JackShort: see above

src/nft/components/collection/CollectionStats.tsx
JackShort: currently the specs are that we are only supporting NFTs on mainnet

src/nft/pages/collection/index.tsx
JackShort: honestly activity is out of the scope of this pr so I am just going to remove it until I add in the activity feature

src/nft/pages/collection/Components.tsx
JackShort: removing activity until its implemented later

JackShort: see comments above

src/nft/components/collection/CollectionStats.tsx
JackShort: no the spec is for the user to click show more and the description expands

src/nft/components/layout/Marquee.tsx
```diff
@@ -0,0 +1,49 @@
...
+    let containerWidth = 0
+    let marqueeWidth = 0
+
+    try {
+      if (marqueeRef.current && containerRef.current) {
+        containerWidth = containerRef.current.getBoundingClientRect().width
+        marqueeWidth = marqueeRef.current.getBoundingClientRect().width
+      }
+    } catch (e) {
+      console.error('Refs became unexpectedly null', e)
```
zzmp, line 20: Do you want to bail out, or continue to setDuration(0)?

zzmp, line 20: I don't know that anything needs to be logged here. Can you remove the `console.error`?

src/nft/pages/collection/index.tsx
```diff
@@ -1,5 +1,30 @@
...
+  const { data: collectionStats } = useQuery(['collectionStats', contractAddress], () =>
+    CollectionStatsFetcher(contractAddress as string)
+  )
+
+  const { width } = useWindowSize()
+  useEffect(() => {
+    setMobileWidth(width ?? 0)
+  }, [width, setMobileWidth])
+
+  if (isMobile) return <CollectionMobile {...{ collectionStats }} />
```
zzmp, line 25: Change this to the more canonical `collectionStats={collectionStats}`?

src/nft/pages/collection/index.tsx
```diff
@@ -1,5 +1,30 @@
...
+
+  const { isMobile, setMobileWidth } = useIsMobile()
+
+  const { data: collectionStats } = useQuery(['collectionStats', contractAddress], () =>
+    CollectionStatsFetcher(contractAddress as string)
+  )
+
+  const { width } = useWindowSize()
+  useEffect(() => {
+    setMobileWidth(width ?? 0)
```
zzmp, line 22: I'm surprised this isn't handled at a higher level, considering that it's global state.
Is this set anywhere else? Does it _need_ to be set here? _Could_ it be set at a higher level, to avoid duplicating this logic in components that shouldn't include this logic?

src/nft/pages/collection/CollectionDesktop.tsx
zzmp: I don't see that locally. Can you try removing it and see if the CI fails?

zzmp (changes requested):

src/nft/components/layout/Marquee.tsx
JackShort: solved by setting duration to 0

src/nft/pages/collection/index.tsx
JackShort: updated the isMobile hook

src/nft/hooks/useIsMobile.ts
```diff
@@ -1,25 +1,28 @@
...
-      setMobileWidth: (width: number) =>
-        set(() => ({
-          width,
-          isMobile: width < breakpoints.tabletSm,
-        })),
-    }),
-    { name: 'isMobile' }
-  )
-)
+export function useIsMobile() {
```
zzmp, line 10: The type is not immediately apparent. Can you annotate this with a return type, `: boolean`?

src/nft/pages/collection/index.tsx
```diff
@@ -11,20 +9,15 @@ import { CollectionMobile } from './CollectionMobile'
...
 
   const { data: collectionStats } = useQuery(['collectionStats', contractAddress], () =>
     CollectionStatsFetcher(contractAddress as string)
   )
 
-  const { width } = useWindowSize()
-  useEffect(() => {
-    setMobileWidth(width ?? 0)
-  }, [width, setMobileWidth])
+  if (isMobile) return <CollectionMobile collectionStats={collectionStats} />
```
zzmp, line 18: nit: this can be simplified with a ternary operator, `return isMobile ? ...`

zzmp (approved):

PR #4392, opened 2022-08-17
fix: width of widget and network balances on token details

kayleegeorge:
fix: widget width and network balances align
Screen Shot 2022-08-17 at 2 46 01 PM

### DIFF
--- a/src/components/Tokens/TokenDetails/BalanceSummary.tsx
+++ b/src/components/Tokens/TokenDetails/BalanceSummary.tsx
@@ -9,15 +9,14 @@ import styled, { useTheme } from 'styled-components/macro'
 import NetworkBalance from './NetworkBalance'
 
 const BalancesCard = styled.div`
-  width: 284px;
+  width: 320px;
   height: fit-content;
   color: ${({ theme }) => theme.textPrimary};
   font-size: 12px;
-  line-height: 20px;
+  line-height: 16px;
   padding: 20px;
   background-color: ${({ theme }) => theme.backgroundSurface};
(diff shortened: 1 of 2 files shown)


src/components/Tokens/TokenDetails/BalanceSummary.tsx
```diff
@@ -9,15 +9,14 @@ import styled, { useTheme } from 'styled-components/macro'
 import NetworkBalance from './NetworkBalance'
 
 const BalancesCard = styled.div`
-  width: 284px;
+  width: 320px;
```
vm, line 12: think you can make a container that wraps the balances and widget, set the width to 320px, then do width 100% for all items inside

kayleegeorge: is that necessary tho

kayleegeorge: bc i think widget you specify the height

kayleegeorge: meh ok ill change

src/pages/TokenDetails/index.tsx
```diff
@@ -53,6 +53,7 @@ const RightPanel = styled.div`
   display: flex;
   flex-direction: column;
   gap: 20px;
+  width: 320px;
```
vm, line 56: then now you can make this a var at the top of the file

vm (approved):
after comment

PR #4394, opened 2022-08-17
fix: add missing colon to end of drop shadow color

lynnshaoyu:
left off a colon when defining a new drop shadow color


vm (approved):
plz add description of how it occurred / what could have caught it

kayleegeorge (approved):

PR #4398, opened 2022-08-18
feat: Wallet v1 transactions part 1

aballerr:
There is another pr with the initial wallet work here: https://github.com/Uniswap/interface/pull/4368
This Pr only contains 3 transactions in order to reduce the size and make testing easier.  I will open another pr with the remaining transactions once this is merged.

I left all the wallet pr code in here because it's difficult to access the transactions otherwise.  What needs to be reviewed is the following

**Files to Review**
src/components/AccountDetailsV2/*
src/components/WalletDropdown/TransactionMenu.tsx
src/pages/Pool/PositionPage.tsx
src/state/transactions/types.ts

**To View**
You need to enable the wallet feature flag.  Then complete a transaction, I would recommend the polygon network for testing purposes

You can get a feel for what it looks like with a preloaded state in this older view (view only, not actionable)
https://deploy-preview-68--projectsmith.netlify.app/#/swap?chain=polygon

**What This Code does**

In Short, shows transactions for: Swap Transactions, Adding Liquidity, Removing Liquidity

There are more transactions NOT listed here.  I will open that in a separate pr to help keep things a bit simpler and easier to review.

The larger spec is here: https://www.notion.so/uniswaplabs/Transaction-State-History-Web-0ea7b872504148cfaab557088bcd8314
Jira is here: https://uniswaplabs.atlassian.net/browse/WEB-736

The wallet will hold transaction information under the transaction tabs.  There are 8 different transactions, but this version only contains 3: Swap, Add liquidity, Remove Liquidity. 

**Design**
https://www.figma.com/file/bCN7G6oArm7pPts03jvJ5x/Unification?node-id=9165%3A83071

**Logic**
The base logic for the pr was borrowed from here: 
src/components/AccountDetails/TransactionSummary.tsx
src/components/AccountDetails/Transaction.tsx

**At a high level**
Each transaction has three parts.  The logos, the text body and the success/pending/failure icons.  My code essentially does the following
1. Get the logos based on the transaction type
2. Get the body based on the transaction type
3. Pass the logos and body to a main wrapper to render the transaction


src/components/WalletDropdown/TransactionMenu.tsx
```diff
@@ -0,0 +1,159 @@
...
+const getConfirmedTransactions = (confirmedTransactions: Array<TransactionDetails>) => {
+  const ONE_DAY = 1000 * 60 * 60 * 24
+  const THIRTY_DAYS = ONE_DAY * 30
+  const NOW = new Date().getTime()
+
+  const TODAY: Array<TransactionDetails> = []
+  const CURRENT_WEEK: Array<TransactionDetails> = []
+  const LAST_30_DAYS: Array<TransactionDetails> = []
+  const CURRENY_YEAR: Array<TransactionDetails> = []
+  const yearMap: { [key: string]: Array<TransactionDetails> } = {}
```
aballerr, line 59: The logic is as follows, found in notion here: https://www.notion.so/uniswaplabs/Transaction-State-History-Web-0ea7b872504148cfaab557088bcd8314


1. **Pending section:**  any tx that were submitted by this wallet and are currently pending.
2. **Today:**  any tx that were completed (success or fail) today
3. **This week:**  all tx that were completed (successfully) this week
4. **Past 30 days:**  any tx that were completed in the last 30 days
5. **Current year (2022):**  all tx that were completed in the current calendar year
6. **Subsequent past year sections (2021, 2020):**  group tx by the calendar year in which they were completed. This is important

src/components/AccountDetailsV2/LogoView.tsx
```diff
@@ -0,0 +1,61 @@
...
+      {!isCentered && (
+        <CurrencyLogoWrapTwo>
+          <CurrencyLogo size="24px" currency={currency1} />
+        </CurrencyLogoWrapTwo>
+      )}
+    </CurrencyWrap>
+  )
+}
+
+export const getLogoView = ({ info }: { info: TransactionInfo }) => {
```
aballerr, line 49: Logos essentially have to views

Single currency: IE approve (not in this pr)
Two currencies: IE Swap

src/components/AccountDetailsV2/LogoView.tsx
```diff
@@ -0,0 +1,61 @@
...
+  switch (info.type) {
```
aballerr, line 50: There is not consistency in the info body of the transaction and as a result we need this get function that can handle all of the various transaction types.  I didn't feel comfortable rewriting transaction info to be consistent.  Might be best for a follow up pr in the future

src/components/AccountDetailsV2/TransactionBody.tsx
```diff
@@ -0,0 +1,171 @@
...
+  transactionState: TransactionState
+}) => {
+  const { createPool, quoteCurrencyId, baseCurrencyId } = info
+  const action = [<Trans key="1">Adding</Trans>, <Trans key="2">Added</Trans>, <Trans key="3">Add</Trans>][
+    transactionState
+  ]
+
+  return (
+    <>
+      {createPool ? (
```
aballerr, line 94: Copied over from original logic.  Creating is included here

aballerr: don't worry about this for this pr

src/components/AccountDetailsV2/TransactionBody.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+const HighlightText = styled.span`
+  color: ${({ theme }) => theme.textPrimary};
+  font-weight: 600;
+`
+
+const formatAmount = (amountRaw: string, decimals: number, sigFigs: number): string =>
+  new Fraction(amountRaw, JSBI.exponentiate(JSBI.BigInt(10), JSBI.BigInt(decimals))).toSignificant(sigFigs)
+
+const getFailedText = (transactionState: TransactionState) =>
```
aballerr, line 25: any time text fails, it takes on the form of

 Swap [TOKEN] for [TOKEN] failed
 
It was a lot easier for me to wrap this in a function and append it at the end of every  transaction

src/components/AccountDetailsV2/TransactionBody.tsx
```diff
@@ -0,0 +1,171 @@
...
+}
+
+const SwapSummary = ({
+  info,
+  transactionState,
+}: {
+  info: ExactInputSwapTransactionInfo | ExactOutputSwapTransactionInfo
+  transactionState: TransactionState
+}) => {
+  const action = [<Trans key="1">Swapping</Trans>, <Trans key="2">Swapped</Trans>, <Trans key="3">Swap</Trans>][
```
aballerr, line 53: 1. I had separate text for all 3 states.  So I adopted this format for consistency
2.  Linter complains about keys.  I tried mapping and it complained still.  So, it looks like this

src/components/AccountDetailsV2/LogoView.tsx
```diff
@@ -0,0 +1,61 @@
...
+  width: 36px;
+  height: 36px;
+`
+
+const CurrencyWrapStyles = css`
+  position: absolute;
+  height: 24px;
+`
+
+const CurrencyLogoWrap = styled.span<{ isCentered: boolean }>`
```
vm, line 18: Perhaps `InputCurrency` and `OutputCurrency` for the other?

src/components/AccountDetailsV2/LogoView.tsx
```diff
@@ -0,0 +1,61 @@
...
+  right: 0px;
+`
+
+export const LogoView = ({ currencyId0, currencyId1 }: { currencyId0: string; currencyId1?: string }) => {
+  const currency0 = useCurrency(currencyId0)
+  const currency1 = useCurrency(currencyId1)
+  const isCentered = !(currency0 && currency1)
+
+  return (
+    <CurrencyWrap>
```
vm, line 36: can this be flex justify space between?

src/components/AccountDetailsV2/LogoView.tsx
```diff
@@ -0,0 +1,61 @@
...
+      <CurrencyLogoWrap isCentered={isCentered}>
```
vm, line 37: if so don't need the is centered and the bottom/right/top/left stuff i think?

src/components/WalletDropdown/TransactionMenu.tsx
```diff
@@ -1,9 +1,159 @@
...
+      <TransactionTitle>{title}</TransactionTitle>
+      {transactions.map((transactionDetails, i) => (
+        <TransactionSummary key={i} transactionDetails={transactionDetails} />
+      ))}
+    </TransactionListWrapper>
+  )
+}
+
+const getConfirmedTransactions = (confirmedTransactions: Array<TransactionDetails>) => {
+  const ONE_DAY = 1000 * 60 * 60 * 24
```
vm, line 51: these should live outside the function closure as they as constants

src/components/WalletDropdown/TransactionMenu.tsx
```diff
@@ -1,9 +1,159 @@
...
+      CURRENY_YEAR.push(transaction)
+    } else {
+      const year = getYear(addedTime)
+
+      if (!yearMap[year]) yearMap[year] = [transaction]
+      else yearMap[year].push(transaction)
+    }
+  })
+
+  const allTransactions: Array<TransactionInformation> = [
```
vm, line 80: ALL_TRANSACTIONS should live outside also?

src/components/WalletDropdown/TransactionMenu.tsx
```diff
@@ -1,9 +1,159 @@
...
+const TransactionTitle = styled.span`
+  padding-bottom: 8px;
+  padding-top: 20px;
+  padding-left: 12px;
+  padding-right: 12px;
+  font-weight: 600;
+  color: ${({ theme }) => theme.textTertiary};
+`
+
+const renderTransactions = (transactionInformation: TransactionInformation) => {
```
vm, line 37: Would make this a component called `TransactionList`. generally don't want a function that returns JSX as you lose the React perf benefits

src/components/AccountDetailsV2/TransactionBody.tsx
vm: why can't you use a switch statement for this?

src/components/AccountDetailsV2/TransactionBody.tsx
```diff
@@ -0,0 +1,171 @@
...
+    transactionState
+  ]
+
+  return (
+    <>
+      {action}{' '}
+      <FormattedCurrencyAmount
+        rawAmount={
+          info.tradeType === TradeType.EXACT_INPUT ? info.inputCurrencyAmountRaw : info.expectedInputCurrencyAmountRaw
```
vm, line 62: `function getRawAmount` which takes info imo, then re-use below

src/components/WalletDropdown/TransactionMenu.tsx
```diff
@@ -1,9 +1,159 @@
...
+    </TransactionListWrapper>
+  )
+}
+
+const getConfirmedTransactions = (confirmedTransactions: Array<TransactionDetails>) => {
+  const ONE_DAY = 1000 * 60 * 60 * 24
+  const THIRTY_DAYS = ONE_DAY * 30
+  const NOW = new Date().getTime()
+
+  const TODAY: Array<TransactionDetails> = []
```
vm, line 55: this should not be capitalized as they are not constants. I would also name it `today_transactions`

src/components/WalletDropdown/TransactionMenu.tsx
vm, line 80: TransactionInformation should be TransactionGroup maybe?

src/components/WalletDropdown/TransactionMenu.tsx
```diff
@@ -1,9 +1,159 @@
...
+      title: 'This year',
+      transactions: CURRENY_YEAR,
+    },
+  ]
+
+  const sortedYears = Object.keys(yearMap)
+    .sort((a, b) => parseInt(b) - parseInt(a))
+    .map((year) => ({ title: year, transactions: yearMap[year] }))
+
+  allTransactions.push(...sortedYears)
```
vm, line 103: rename to `transactionGroups`?

src/components/AccountDetailsV2/TransactionBody.tsx
```diff
@@ -0,0 +1,171 @@
...
+}
+
+const RemoveLiquidityV3Summary = ({
+  info: { baseCurrencyId, quoteCurrencyId, expectedAmountBaseRaw, expectedAmountQuoteRaw },
+  transactionState,
+}: {
+  info: RemoveLiquidityV3TransactionInfo
+  transactionState: TransactionState
+}) => {
+  const action = [<Trans key="1">Removing</Trans>, <Trans key="2">Removed</Trans>, <Trans key="3">Remove</Trans>][
```
vm, line 116: similar as above, could be a switch

src/components/AccountDetailsV2/LogoView.tsx
aballerr: the first currency need to be top left and the second bottom right

vm: ah ok all good

JFrankfurt:
image
a few questions here:

- Is now the time to remove these toast pop ups?
- there appear to be no element `:hover` styles applied to these menus, but design (cc @pp-hh-ii-ll ) says there should be

image
- I was initially confused here by the distinction between the "TODAY" transactions and the currently pending transaction (also today! mere moments ago! why is it not "today"??) @pp-hh-ii-ll help pl0x.
- I think it feels bad to have an input or output token symbol wrap to a different line than its associated amount.
- The scroll bar here shouldn't be the whole container. It should just apply to the list elements, and it shouldn't be visible when the content doesn't actually need to (or is unable to) scroll.

image
- header/control content should not scroll

JFrankfurt (changes requested):
see prior comment

.eslintrc.json
```diff
@@ -30,7 +30,8 @@
     ".idea/",
     ".vscode/",
     "package-lock.json",
-    "yarn.lock"
+    "yarn.lock",
+    "src/*"
```
JFrankfurt, line 34: shoudn't be ignoring `src/*`

aballerr: I ignore while developing.  These rules should be applied precomit to improve the development experience

.eslintrc.json
```diff
@@ -31,6 +31,7 @@
     ".vscode/",
     "package-lock.json",
     "yarn.lock"
+
```
JFrankfurt, line 34: nit: drop newline

src/components/AccountDetailsV2/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+      ? TransactionState.Pending
+      : success
+      ? TransactionState.Success
+      : TransactionState.Failed
+
+    return transactionState
+  }, [receipt, tx])
+
+  // const body = getTransactionBody({ info, transactionState })
+  const link = `${explorer}tx/${hash}`
```
JFrankfurt, line 63: is this comment necessary still?

src/components/WalletDropdown/TransactionMenu.tsx
```diff
@@ -1,9 +1,167 @@
...
+import { TransactionSummary } from '../AccountDetailsV2'
 import { SlideOutMenu } from './SlideOutMenu'
 
-export const TransactionHistoryMenu = ({ onClose }: { onClose: () => void }) => (
-  <SlideOutMenu onClose={onClose} onClear={undefined} title={<Trans>Transactions</Trans>}>
-    <div />
-  </SlideOutMenu>
-)
+const ONE_DAY = 1000 * 60 * 60 * 24
+const THIRTY_DAYS = ONE_DAY * 30
```
JFrankfurt, line 15: use `ms` to do this:
```
ms`1 day`
```
```
ms`30 days`
```

src/components/WalletDropdown/TransactionMenu.tsx
```diff
@@ -1,9 +1,167 @@
...
+
+const TransactionList = ({ transactionInformation }: { transactionInformation: TransactionInformation }) => {
+  const { title, transactions } = transactionInformation
+
+  return (
+    <TransactionListWrapper key={title}>
+      <TransactionTitle>{title}</TransactionTitle>
+      {transactions.map((transactionDetails, i) => (
+        <TransactionSummary key={i} transactionDetails={transactionDetails} />
+      ))}
```
JFrankfurt, line 48: use the tx hash as the key here

JFrankfurt (changes requested):

src/components/WalletDropdown/TransactionMenu.tsx
vm: in general want to avoid key = index as jordan says, as it will cause issues when the list changes

src/components/WalletDropdown/DefaultMenu.tsx
```diff
@@ -47,7 +43,7 @@ const ToggleMenuItem = styled.button`
   margin-bottom: 8px;
   color: ${({ theme }) => theme.textSecondary};
   :hover {
-    text-decoration: none;
+    color: ${({ theme }) => theme.textPrimary};
   }
```
JFrankfurt, line 47: this should have a `transition: 250ms color ease;`

src/components/WalletDropdown/SlideOutMenu.tsx
```diff
@@ -38,6 +37,14 @@ const ClearAll = styled.div`
   margin-bottom: auto;
 `
 
+const StyledChevron = styled(ChevronLeft)`
+  cursor: pointer;
+
+  &:hover {
+    color: ${({ theme }) => theme.textPrimary};
+  }
```
JFrankfurt, line 45: add a color transition w/ 250ms

src/components/WalletDropdown/TransactionMenu.tsx
```diff
@@ -1,9 +1,167 @@
...
+  padding-left: 12px;
+  padding-right: 12px;
+  color: ${({ theme }) => theme.textSecondary};
+`
+
+export const TransactionHistoryMenu = ({ onClose }: { onClose: () => void }) => {
+  const allTransactions = useAllTransactions()
+  const { chainId } = useWeb3React()
+  const dispatch = useAppDispatch()
+  const transactionGroupsInformation = []
```
JFrankfurt, line 126: not *suuuper* important, but I'd put the building of this array in a `useMemo` and move its declaration down to the bottom where you've got the `if` logic happening on line 144

JFrankfurt (approved):

PR #4403, opened 2022-08-18
fix: fix add liquidity select tokens alignment

lynnshaoyu:
before:
Screen Shot 2022-08-18 at 1 54 25 PM

after: 
Screen Shot 2022-08-18 at 1 54 16 PM

### DIFF
--- a/src/components/CurrencyInputPanel/index.tsx
+++ b/src/components/CurrencyInputPanel/index.tsx
@@ -11,6 +11,7 @@ import { RedesignVariant, useRedesignFlag } from 'featureFlags/flags/redesign'
 import { darken } from 'polished'
 import { ReactNode, useCallback, useState } from 'react'
 import { Lock } from 'react-feather'
+import { useLocation } from 'react-router-dom'
 import styled, { useTheme } from 'styled-components/macro'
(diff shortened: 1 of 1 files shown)


src/components/CurrencyInputPanel/index.tsx
```diff
@@ -151,18 +151,6 @@ const FiatRow = styled(LabelRow)<{ redesignFlag: boolean }>`
   height: ${({ redesignFlag }) => !redesignFlag && '24px'};
 `
 
-const NoBalanceState = styled.div`
```
kayleegeorge, line 154: oh wait i think this stuff is needed for swap -- i totally forgot this component is also used in liquidity select

vm (approved):

src/components/CurrencyInputPanel/index.tsx
kayleegeorge: discussed in DMs: only display on swap, not on liquidity context

lynnshaoyu: modified in latest commit to only remove in the add liquidity screen. this one shows what swap looks like: 
Screen Shot 2022-08-18 at 11 19 12 PM

PR #4405, opened 2022-08-18
fix search

kayleegeorge:
fix search bar based on design, no more expanding

Screen Shot 2022-08-18 at 12 50 11 PM

Screen Shot 2022-08-18 at 12 50 06 PM


kayleegeorge:
for some reason, search is slow -- not sure how to fix

vm (approved):

PR #4409, opened 2022-08-18
fix: fix skeleton hover state bug

cbachmeier:
Fixes unintentional flicker when mouse out on skeleton loading state
skeletonHoverState

### DIFF
--- a/src/components/NavBar/SearchBar.css.ts
+++ b/src/components/NavBar/SearchBar.css.ts
@@ -71,7 +71,7 @@ export const suggestionRow = style([
   {
     ':hover': {
       cursor: 'pointer',
-      background: vars.color.lightGrayContainer,
+      background: vars.color.lightGrayButton,
     },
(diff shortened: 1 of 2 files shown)


JFrankfurt (approved):
While we're here, can you slap one of these on there too?
`transition: 250ms ease background-color;`

It looks like we don't have a specified transition time in the theme, but 250ms is pretty standard in our codebase.

cbachmeier:
> While we're here, can you slap one of these on there too? `transition: 250ms ease background-color;`
> 
> It looks like we don't have a specified transition time in the nft theme, but 250ms is pretty standard in our codebase.

@JFrankfurt  One benefit of VanillaExtract is that I can add that common style very simply: https://github.com/Uniswap/interface/pull/4409/files#diff-26a8e1810e548f7716e47ee3bb13ce70385367c47c5174e14de08f64ed69ec46R70
transition

PR #4413, opened 2022-08-18
feat: enable NFT search results

cbachmeier:
enable NFT collection search results when feature flag is enabled

### DIFF
--- a/src/components/NavBar/SearchBar.tsx
+++ b/src/components/NavBar/SearchBar.tsx
@@ -148,14 +148,21 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }:
       .slice(0, isNFTPage ? 3 : 2)
   }, [isNFTPage, trendingCollectionResults])
 
+  const showTrendingTokens: Boolean = useMemo(
+    () => (trendingCollections?.length ?? 0) > 0 && !isTokenPage && phase1Flag === Phase1Variant.Enabled,
+    [trendingCollections?.length, isTokenPage, phase1Flag]
+  )
+
   const { data: trendingTokenResults } = useQuery([], () => fetchTrendingTokens(4), {
     refetchOnWindowFocus: false,
     refetchOnMount: false,
     refetchOnReconnect: false,
   })
 
+  const trendingTokensLength = phase1Flag === Phase1Variant.Enabled ? (isTokenPage ? 3 : 2) : 4
+
   const trendingTokens = useMemo(() => {
-    return trendingTokenResults?.slice(0, phase1Flag === Phase1Variant.Enabled ? (isTokenPage ? 3 : 2) : 4)
+    return trendingTokenResults?.slice(0, trendingTokensLength)
   }, [trendingTokenResults, isTokenPage, phase1Flag])
 
   const totalSuggestions = hasInput
@@ -232,7 +239,7 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }:
               headerIcon={<TrendingArrow />}
             />
           )}
-          {(trendingCollections?.length ?? 0) > 0 && !isTokenPage && phase1Flag === Phase1Variant.Enabled && (
+          {showTrendingTokens && (
             <SearchBarDropdownSection
               hoveredIndex={hoveredIndex}
               startingIndex={searchHistory.length + (isNFTPage ? 0 : trendingTokens?.length ?? 0)}
--- a/src/components/NavBar/SearchBar.tsx
+++ b/src/components/NavBar/SearchBar.tsx
@@ -1,4 +1,4 @@
-import { Phase1Variant, usePhase1Flag } from 'featureFlags/flags/phase1'
+import { NftVariant, useNftFlag } from 'featureFlags/flags/nft'
 import useDebounce from 'hooks/useDebounce'
 import { useOnClickOutside } from 'hooks/useOnClickOutside'
 import { useWindowSize } from 'hooks/useWindowSize'
@@ -96,7 +96,7 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }:
   const { pathname } = useLocation()
   const isNFTPage = pathname.includes('/nfts')
   const isTokenPage = pathname.includes('/tokens')
-  const phase1Flag = usePhase1Flag()
+  const phase1Flag = useNftFlag()
 
   const tokenSearchResults =
     tokens.length > 0 ? (
@@ -113,7 +113,7 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }:
     )
 
   const collectionSearchResults =
-    phase1Flag === Phase1Variant.Enabled ? (
+    phase1Flag === NftVariant.Enabled ? (
       collections.length > 0 ? (
         <SearchBarDropdownSection
           hoveredIndex={hoveredIndex}
@@ -149,7 +149,7 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }:
   }, [isNFTPage, trendingCollectionResults])
 
   const showTrendingCollections: boolean = useMemo(
-    () => (trendingCollections?.length ?? 0) > 0 && !isTokenPage && phase1Flag === Phase1Variant.Enabled,
+    () => (trendingCollections?.length ?? 0) > 0 && !isTokenPage && phase1Flag === NftVariant.Enabled,
     [trendingCollections?.length, isTokenPage, phase1Flag]
   )
 
@@ -159,11 +159,11 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }:
     refetchOnReconnect: false,
   })
 
-  const trendingTokensLength = phase1Flag === Phase1Variant.Enabled ? (isTokenPage ? 3 : 2) : 4
+  const trendingTokensLength = phase1Flag === NftVariant.Enabled ? (isTokenPage ? 3 : 2) : 4
 
   const trendingTokens = useMemo(() => {
     return trendingTokenResults?.slice(0, trendingTokensLength)
-  }, [trendingTokenResults, isTokenPage, phase1Flag])
+  }, [trendingTokenResults, trendingTokensLength])
 
   const totalSuggestions = hasInput
     ? tokens.length + collections.length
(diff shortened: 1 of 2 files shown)


src/components/NavBar/SearchBar.tsx
```diff
@@ -131,10 +155,10 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }:
   })
 
   const trendingTokens = useMemo(() => {
-    // TODO reimplement this logic with NFT search
-    // return trendingTokenResults?.slice(0, isTokenPage ? 3 : 2)
-    return trendingTokenResults?.slice(0, 4)
-  }, [trendingTokenResults])
+    return phase1Flag === Phase1Variant.Enabled
+      ? trendingTokenResults?.slice(0, isTokenPage ? 3 : 2)
+      : trendingTokenResults?.slice(0, 4)
```
JFrankfurt, line 160: can you simplify this boolean logic down just to determining the `end` number?

src/components/NavBar/SearchBar.tsx
```diff
@@ -210,14 +234,14 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }:
               headerIcon={<TrendingArrow />}
             />
           )}
-          {(trendingCollections?.length ?? 0) > 0 && !isTokenPage && (
+          {(trendingCollections?.length ?? 0) > 0 && !isTokenPage && phase1Flag === Phase1Variant.Enabled && (
```
JFrankfurt, line 237: can you collapse this to a descriptive variable and `Boolean()` cast it for `render` usage?

src/components/NavBar/SearchBar.tsx
```diff
@@ -210,14 +234,14 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }:
...
             <SearchBarDropdownSection
               hoveredIndex={hoveredIndex}
               startingIndex={searchHistory.length + (isNFTPage ? 0 : trendingTokens?.length ?? 0)}
               setHoveredIndex={setHoveredIndex}
               toggleOpen={toggleOpen}
               suggestions={trendingCollections as unknown as GenieCollection[]}
-              header={'Trending NFT collections'}
+              header={'Popular NFT collections'}
```
JFrankfurt, line 244: Should use `<Trans />` on string literal values

src/components/NavBar/SearchBar.tsx
cbachmeier: I'm apprehensive about creating a boolean only used here that obfuscates a lot of the context especially when 1/3 of the conditions will be removed when we remove the feature flag

src/components/NavBar/SearchBar.tsx
```diff
@@ -131,10 +155,8 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }:
   })
 
   const trendingTokens = useMemo(() => {
-    // TODO reimplement this logic with NFT search
-    // return trendingTokenResults?.slice(0, isTokenPage ? 3 : 2)
-    return trendingTokenResults?.slice(0, 4)
-  }, [trendingTokenResults])
+    return trendingTokenResults?.slice(0, phase1Flag === Phase1Variant.Enabled ? (isTokenPage ? 3 : 2) : 4)
```
JFrankfurt, line 158: can you un-nest this logic so it's a little easier to follow?

src/components/NavBar/SearchBar.tsx
cbachmeier: Localization was not in the scope of this work and will be added later

src/components/NavBar/SearchBar.tsx
```diff
@@ -312,8 +341,8 @@ export const SearchBar = () => {
           </Box>
           <Box
             as="input"
-            placeholder="Search tokens"
-            width={isOpen ? 'full' : '120'}
+            placeholder={`Search tokens${phase1Flag === Phase1Variant.Enabled ? ' and NFT collections' : ''}`}
```
JFrankfurt, line 344: rendered string values should be translated

src/components/NavBar/SearchBar.tsx
```diff
@@ -238,15 +260,22 @@ export const SearchBar = () => {
...
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  const collections = [] as GenieCollection[]
-  const collectionsAreLoading = false
+  const { data: collections, isLoading: collectionsAreLoading } = useQuery(
+    ['searchCollections', debouncedSearchValue],
+    () => fetchSearchCollections(debouncedSearchValue),
+    {
+      refetchOnWindowFocus: false,
+      refetchOnMount: false,
+      refetchOnReconnect: false,
```
JFrankfurt, line 275: double checking these values

src/components/NavBar/SearchBar.tsx
cbachmeier: So like how I had it before?
```
return phase1Flag === Phase1Variant.Enabled
      ? trendingTokenResults?.slice(0, isTokenPage ? 3 : 2)
      : trendingTokenResults?.slice(0, 4)
```
or are you thinking something different?

src/components/NavBar/SearchBar.tsx
```diff
@@ -110,19 +113,40 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }:
...
+    return trendingCollectionResults
+      ?.map((collection) => {
+        return {
+          ...collection,
+          collectionAddress: collection.address,
+          floorPrice: formatEthPrice(collection.floor.toString()),
+          stats: {
+            total_supply: collection.totalSupply,
+            one_day_change: collection.floorChange,
+          },
```
JFrankfurt, line 145: double checking there shouldn't be a `{ ...collection.stats, ... }` here

src/components/NavBar/SearchBar.tsx
cbachmeier: localization is not in the scope of this work and will be a follow up PR

src/components/NavBar/SearchBar.tsx
```diff
@@ -110,19 +113,40 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }:
...
+        <SearchBarDropdownSection
+          hoveredIndex={hoveredIndex}
+          startingIndex={isNFTPage ? 0 : tokens.length}
+          setHoveredIndex={setHoveredIndex}
+          toggleOpen={toggleOpen}
+          suggestions={collections}
+          header={'NFT Collections'}
+        />
+      ) : (
+        <Box className={styles.notFoundContainer}>No NFT collections found.</Box>
```
JFrankfurt, line 127: translate these rendered strings

src/components/NavBar/SearchBar.tsx
cbachmeier: this is correct

JFrankfurt (changes requested):
see comments

src/components/NavBar/SearchBar.tsx
cbachmeier: this is the desired behaviour to avoid overloading api with unnecessary calls

src/components/NavBar/SearchBar.tsx
cbachmeier: localization is not in the scope of this work and is to be addressed in a follow up PR

src/components/NavBar/SearchBar.tsx
JFrankfurt: We've always enforced localization at the same time as we've written the UI, but I'm not personally opposed to changing that if there is a card already assigned/prioritized to go out with this release. Does that exist somewhere?

src/components/NavBar/SearchBar.tsx
cbachmeier: the extra information is not needed in suggestion preview

src/components/NavBar/SearchBar.tsx
JFrankfurt: I mean a single call to `slice`, and un-nested logic for determination of the `end` argument. 
```
let end = 4
if (phase1Flag === Phase1Variant.Enabled) {
  end = isTokenPage ? 3 : 2
}
return trendingTokenResults?.slice(0, end)
```
pref current implementation to former for the single `slice` call, but the bit about `end` determination logic structure is a nit.

src/components/NavBar/SearchBar.tsx
JFrankfurt: The risk here is that you can end up accidentally rendering truthy values into your template when you didn't intend to. 

i.e., 
```
const val1 = 1
const val2 = false
return val1 && val2 && <div>some text</div>
```
will render the digit ` to your template because of the sequential boolean execution

if you do this instead, you get a more descriptive template and no risk of ever accidentally rendering truthy values:
```
const val1 = 1
const val2 = false
const theTextShouldRender = Boolean(val1 && val2) // false
return theTextShouldRender  && <div>some text</div>
```

src/components/NavBar/SearchBar.tsx
cbachmeier: I think `let` overwrites should be used as infrequently as possible vs const when we know from init what the value should be set as. What do you think of something like this:
```
  const trendingTokensLength = phase1Flag === Phase1Variant.Enabled ? (isTokenPage ? 3 : 2) : 4

  const trendingTokens = useMemo(() => {
    return trendingTokenResults?.slice(0, trendingTokensLength)
  }, [trendingTokenResults, isTokenPage, phase1Flag])

File: src/components/NavBar/SearchBar.tsx (hunk 2 above)
Comment (cbachmeier) [reply src/components/NavBar/SearchBar.tsx:235]: In this case all condition are explicitly bools to avoid the above issue, notably the # of collections is explicitly compared to 0 as opposed to assuming `0` will be treated as false.

Is there some standard practice for when to collapse conditions to a `shouldShowX`? 3 or more conditions? Even if 1 of those 3 conditions is temporary when we remove the feature flag?

Comment (cbachmeier) [reply src/components/NavBar/SearchBar.tsx]: I've gone with:
```
const trendingTokensLength = phase1Flag === NftVariant.Enabled ? (isTokenPage ? 3 : 2) : 4

  const trendingTokens = useMemo(() => {
    return trendingTokenResults?.slice(0, trendingTokensLength)
  }, [trendingTokenResults, trendingTokensLength])

JFrankfurt (approved):

PR #4414, opened 2022-08-18
refactor: rename phase1flag to nftflag

cbachmeier:
-rename phase1flag to nftflag
  - I specifically did not pluralize to `nfts` in this case as the capitalization gets tricky for an acronym so this was the most straightforward


vm (approved):
fire

PR #4416, opened 2022-08-19
fix: Fix unresponsive image in vote page

tainguyen-kms:
Fixes unresponsive image in description section of the vote page

before
after

### DIFF
--- a/src/pages/Vote/VotePage.tsx
+++ b/src/pages/Vote/VotePage.tsx
@@ -62,14 +62,6 @@ const PageWrapper = styled(AutoColumn)`
   `};
 `
 
-const MarkDownDescription = styled(ReactMarkdown)`
-  & > p > img {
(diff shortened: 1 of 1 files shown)


vm:
I think you can try
```
    <Markdown
      renderers={{
        image: (props) => <img {...props} style={{ maxWidth: "100%" }} alt="" />
      }}
    >
```

might need to mess with the styles a bit, and haven't looked at the docs closely yet. lmk if that works?

tainguyen-kms:
> I think you can try
> 
> ```
>     <Markdown
>       renderers={{
>         image: (props) => 
>       }}
>     >
> ```
> 
> might need to mess with the styles a bit, and haven't looked at the docs closely yet. lmk if that works?

Thanks for your feedback, I've updated to use renderer for image

vm (approved):

PR #4418, opened 2022-08-19
feat: Removes censorship from the UI.

MicahZoltu:
## Technical

Also swaps out some of the RPC servers used since the Infura ones are known to censor.  Keydonix is currently just CloudFlare's JSON-RPC server under the hood (DNS CNAME).  The reason the public CF server wasn't included was because there aren't public testnet servers, and I would like to ensure CF gets paid for the traffic being sent to it so they continue to offer their service (using the public gateway would result in the cost being entirely theirs).  Also, if CloudFlare censors (or ends up being too expensive), I can redirect Keydonix to QuickNode or somewhere else that is cheaper/uncensored.

The wallet test was disabled because after a day of debugging I was unable to figure out why changing the RPC server for mainnet from Infura to Keydonix results in the test failing.  When I tested the behavior in a production environment everything worked, so I decided to disable the test for now so I could move forward.

## Philosophical

Ethereum is meant to be censorship resistant money, where people can go to freely interact and trade with each other without fear of being kicked out of the financial system we are building by governments, banks, corporations, etc.  When applications are built on top of Ethereum, they should strive to uphold that ethos and they should not be quick to fold at the first sign of any pushback.

Unfortunately, Uniswap integrated TRM at some point and TRM is the opposite of "pushing back against regulatory overreach" as they follow a "block first, ask questions later" approach.  The blacklisting strategies are also entirely opaque to users so there is no way to know if they are following a *reasonable* set of blacklisting rules that the community approves of, or if they are blacklisting journalists and aid workers living under totalitarian regimes.

On top of that, the current RPC provider (infura) is also aggressively censoring and following censorship policies that are opaque to users.  I'm willing to foot the bill to get them removed by providing the JSON-RPC servers necessary to serve Uniswap users uncensored content (assuming all censorship is removed, this is a package deal).

By removing censoring code, we can start working together toward a brighter future where humans have the right to freedom of speech/expression/association (including financial expression and trade) and these freedoms aren't wantonly removed by governments and businesses.


MicahZoltu:
The failing test *appears* to be one that fails intermittently (at least in my fork).  I'm unable to trigger a re-run in this PR though, so I can't get it to go green on my own I don't think.

JFrankfurt:
I wrote the front/back end implementation of our TRM code when we decided to stop relying on the manually hardcoded block list. I volunteered to take the task because I really care about freedom and user privacy; I wanted to ensure we respect our users as much as possible while complying with the law.

My thoughts on blocking and trm:
- We are doxxed individuals in the US. If we offer our services (this website that interacts with the AMM protocol) to sanctioned individuals we are breaking the law and key individuals related to that could go to jail. We choose not to run that risk.
- We  run the blocking check at page load because we didn't want to induce lag on swap/lp interactions.
- **We have blocked 253 addresses to date** via TRM (roughly 4 months). There was a brief period when we were rolling this out when we blocked addresses that were counter-parties to sanctioned addresses, but we scaled that back and now only block addresses that are sanctioned directly or are direct recipients of scam/fraud/hacked funds. As @banteg has been tweeting recently, over-compliance by blocking according to proximity can result in massive swaths of innocent users being incorrectly banned. We are very aware of this and super cautious not to do it. 

If you'd like to validate this, you can directly query the same endpoint we use to check addresses for a dump of the  addresses which have used our interface. These are all publicly known and 

```js
import fetch from "node-fetch";
import { readFile, writeFile } from "fs/promises";
import { resolve } from "path";

const dataFilePath = resolve("./data.json");

async function checkDataLength() {
  const fileData = await readFile(dataFilePath, { encoding: "utf-8" });
  const fileJson = JSON.parse(fileData);
  return Object.values(fileJson).length;
}

async function fetchNext(cursor) {
  const url = new URL(
    "https://screening-worker.uniswap.workers.dev?namespace=BLOCKED"
  );
  if (cursor) {
    url.searchParams.append("cursor", cursor);
  }
  console.log(`fetching ${url}`);

  const response = await fetch(url, {
    headers: { Accept: "application/json" },
  });
  console.log("response.ok: ", response.ok);
  const { data, meta } = await response.json();
  console.log("new cursor: ", meta.cursor);

  const fileData = await readFile(dataFilePath, { encoding: "utf-8" });
  const fileJson = JSON.parse(fileData);
  const newFileJson = Object.assign(fileJson, data);
  await writeFile(dataFilePath, JSON.stringify(newFileJson, null, 2), {
    encoding: "utf-8",
  });

  if (meta.list_complete) {
    console.log("LIST COMPLETE");
    return;
  }

  if (meta.cursor) {
    console.log("============");
    await fetchNext(meta.cursor);
  }

  return;
}

fetchNext()
.then((x) => {
    console.log("done", x);
  })
  .catch((e) => console.error(e));

```

re: RPC change
Is keydonix a domain you own @MicahZoltu? Are you just proxying to the public cloudflare rpc? I don't think it's viable to switch all our default non-signer rpc requests to either a free rpc or an unknown (or personal) endpoint. Is there a summary somewhere of rpc censorship that has been happening the past few weeks? I only know about the flashbots stuff. I think we may be open to switching our default.

MicahZoltu:
> We are doxxed individuals in the US. If we offer our services (this website that interacts with the AMM protocol) to sanctioned individuals we are breaking the law and key individuals related to that could go to jail.

I think it is important to note that this PR doesn't propose that the software operated by Uniswap Labs is changed.  It merely proposes that the source code published here, and by extension the artifacts generated by someone who builds this code, don't censor.  Uniswap Labs could maintain a separate branch that censors which is deployed to app.uniswap.com which would make me sad but would at least make it so it was relatively easy for people to find a legitimate/verifiable copy of the UI that doesn't censor.

> Is keydonix a domain you own @MicahZoltu? Are you just proxying to the public cloudflare rpc? I don't think it's viable to switch all our default non-signer rpc requests to either a free rpc or an unknown (or personal) endpoint. Is there a summary somewhere of rpc censorship that has been happening the past few weeks? I only know about the flashbots stuff. I think we may be open to switching our default.

Infura is censoring (and has been for some time) many users, including entire countries.  In particular, countries where the people are in dire need of access to uncensored finance.

Keydonix is a domain I own, and ipfs.keydonix.com points to CloudFlare's paid gateway (I'm paying for it).  I agree with you about not using an unpaid gateway for the reasons you mentioned.  QuickNode has indicated that they do not censor, so if you don't trust me (reasonable) then they would be another viable paid alternative.

-----

On a more philosophical note, if the Uniswap developers are unable to provide uncensored software to their users, my recommendation would be to walk away and make room for new developers who can.  While I would find it very unfortunate to lose such good developers, I think it is worse to have those good developers working against those of us fighting for freedom and privacy, which is functionally what is happening now.

Every user that uses the censored uniswap.org UI is a user that isn't using, promoting, funding, and supporting an uncensored UI.  By continuing to sink significant development efforts into a censored frontend, you are undermining the efforts of people building end to end censorship resistant software.

Perhaps just focus on writing code and exit the hosting/operating game?  Let others who have a higher risk tolerance or who live in less dictatorial countries do the operating.

qizhou:
I strongly support @MicahZoltu 's proposal of removing censorship, but I agree with @JFrankfurt that this will put uniswap at high risk.  In fact, as long as we are living in the http:// protocol world, we inevitably face the threat of censorship because those domains/servers are owned by centralized entities (e.g., I would expect that CF may join censorship shortly).

I have been thinking about the solution for a while, but I believe the key part is that we need some decentralized URL protocol to replace http://:
- For website hosting, we could use ipfs:// for static web objects, and we may further store the websites on chain (and managed by DAO) and access them via eth-web3:// protocol;
- For JSON-RPC, we should have eth-jsrpc:// protocol that serves the same feature as https:// JSON-RPC.

Note that to support these protocols, we will need native integration or extension with browsers, but I feel this should be the direction of fundamentally addressing the censorship problem.

antoncoding:
> Infura is censoring (and has been for some time) many users, including entire countries. In particular, countries where the people are in dire need of access to uncensored finance.
> 
> Keydonix is a domain I own, and ipfs.keydonix.com points to CloudFlare's paid gateway (I'm paying for it). I agree with you about not using an unpaid gateway for the reasons you mentioned. QuickNode has indicated that they do not censor, so if you don't trust me (reasonable) then they would be another viable paid alternative.
> 

Personal opinion: I think changing the endpoint to another personal endpoint does not seem to make the code more censorship resistant, it just kind of re-direct the legal risk to another party (yourself or CF?). Also from the UX perspective, infura has been one of the most stable and reliable RPC providers, so swapping out also mean probably more downtime or delay .. etc, for average users.

I feel like a better solution is to give users the option to swap out RPC endpoints very easily, while your endpoint can be a good alternative on the suggestion list. That way Uniswap fully puts the "choice" in users' hands, and remain neutral as a platform, which I think sounds more decentralised and fair.

MicahZoltu:
> Personal opinion: I think changing the endpoint from Infura to another personal endpoint does not seem to make the code more censorship resistant, it just kind of re-direct the legal risk to another party (yourself or CF?). Also from the UX perspective, infura has been one of the most stable and reliable RPC providers, so swapping out also mean probably more downtime or delay .. etc, for average users. I feel like from Uniswap team's perspective, letting and external person control the downtime of the product might not be an acceptable option.

I'm happy to switch the endpoint to something else that doesn't censor like QuickNode if that is preferred.  For mainnet, the public CF server could be used though that suffers the problem mentioned above by @JFrankfurt.

> I feel like a better solution is to give users the option to swap out RPC endpoints very easily, while your endpoint can be a good alternative on the suggestion list. That way Uniswap fully puts the "choice" in users' hands, and remain neutral as a platform, which I think sounds more decentralised and fair.

I totally agree with this, and I think long term that is the direction that Uniswap should go.  This PR was meant to be the minimum change that can remove censorship *today*.  I think there are a lot of things that could be improved beyond this, with what you mention probably being top of the list.

FWIW, users today *can* change their provider by connecting to a wallet.

paulmillr:
@JFrankfurt why was it closed? Don't you have some closing remarks, at least?

octalmage:
Sad to see this closed without a comment.

PR #4422, opened 2022-08-19
feat: collection asset cards

JackShort:
* added collection asset cards to the collection page


src/nft/components/collection/Card.tsx
```diff
@@ -0,0 +1,383 @@
...
+          />
+        </Box>
+      ) : (
+        <NoContentContainer />
+      )}
+    </>
+  )
+}
+
+Card.Image = CardImage
```
vm, line 133: make this a module instead

vm (changes requested):
some base level changes

src/nft/components/collection/Card.tsx
```diff
@@ -0,0 +1,383 @@
...
+export interface CardContextProps {
+  asset: GenieAsset
+  hovered: boolean
+  selected: boolean
+  href: string
+  setHref: (href: string) => void
+}
+
+const CardContext = createContext<CardContextProps>({
+  asset: {} as GenieAsset,
```
vm, line 34: `Partial<GenieAsset>` i think?

vm: i guess if it's empty like this shouldn't asset just be undefined?

src/nft/components/collection/Card.tsx
```diff
@@ -0,0 +1,383 @@
...
+      toggleHovered,
+      href,
+      setHref,
+    }),
+    [asset, hovered, href]
+  )
+
+  const assetRef = useRef<HTMLDivElement>(null)
+
+  useLayoutEffect(() => {
```
vm, line 71: is this necessary if we have the onMouseEnter and leave stuff below?

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -0,0 +1,82 @@
...
+  )
+
+  const collectionNfts = useMemo(() => {
+    if (!collectionAssets || !AssetsFetchSuccess) return undefined
+
+    return collectionAssets.pages.flat()
+  }, [collectionAssets, AssetsFetchSuccess])
+
+  if (!collectionNfts) {
+    return <div>No CollectionAssets</div>
```
vm, line 53: typo? or is this temp

JackShort: yeah this is just temp

src/nft/components/collection/Card.tsx
JackShort: yes there are states where onMouseEnter doesn't trigger - for instance if the card is loaded underneath the user's cursor

src/nft/components/collection/CollectionNfts.tsx
vm: can you add a todo in code for that

src/nft/components/collection/Card.tsx
JackShort: solved by performing a check that context is used inside of provider

src/nft/components/collection/Card.css.ts
```diff
@@ -0,0 +1,171 @@
...
+  sprinkles({
+    overflow: 'hidden',
+    borderStyle: 'solid',
+    borderWidth: '1px',
+  }),
+  {
+    boxSizing: 'border-box',
+    WebkitBoxSizing: 'border-box',
+    '@media': {
+      [`(max-width: ${breakpoints.tabletSm - 1}px)`]: {
```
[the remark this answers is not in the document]
JackShort: our mobile state doesn't have a set width and since max-width is inclusive we have to do - 1

src/nft/components/collection/Card.tsx
vm: i'm just trying to avoid casting the type here

src/nft/components/collection/Card.css.ts
```diff
@@ -0,0 +1,171 @@
...
+
+export const card = style([
+  sprinkles({
+    overflow: 'hidden',
+    borderStyle: 'solid',
+    borderWidth: '1px',
+  }),
+  {
+    boxSizing: 'border-box',
+    WebkitBoxSizing: 'border-box',
```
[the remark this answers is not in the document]
JackShort: sprinkles does not and I don't think we installed autoprefixing on craco

src/nft/components/collection/Card.tsx
```diff
@@ -0,0 +1,358 @@
...
+const PrimaryInfo = ({ children }: { children: ReactNode }) => {
+  return (
+    <Box
+      overflow="hidden"
+      whiteSpace="nowrap"
+      textOverflow="ellipsis"
+      color="blackBlue"
+      fontWeight="medium"
+      fontSize="14"
+      style={{ lineHeight: '20px' }}
```
vm, line 174: shouldn't line height just be a prop on box?

src/nft/components/collection/CollectionAsset.css.ts
```diff
@@ -0,0 +1,142 @@
...
+  },
+  sprinkles({
+    backgroundColor: 'genieBlue',
+    borderRadius: 'round',
+  }),
+])
+
+export const tokenQuantity = style([
+  {
+    padding: '10px 17px',
```
vm, line 54: nit: i think should be 8px 16px?

vm (approved):

src/nft/components/collection/Card.tsx
JackShort: it should be but it isn't - we need to add that

PR #4424, opened 2022-08-19
feat: remove all token user properties besides usdc, weth, native token

lynnshaoyu:


src/components/WalletModal/index.tsx
```diff
@@ -191,7 +164,9 @@ export default function WalletModal({
   const redesignFlagEnabled = redesignFlag === RedesignVariant.Enabled
   const [walletView, setWalletView] = useState(WALLET_VIEWS.ACCOUNT)
   const [lastActiveWalletAddress, setLastActiveWalletAddress] = useState<string | undefined>(account)
-  const [shouldLogWalletBalances, setShouldLogWalletBalances] = useState(false)
+  const [shouldLogUsdcBalance, setShouldLogUsdcBalance] = useState(false)
```
vm, line 167: can these 3 be a single state?

src/components/WalletModal/index.tsx
```diff
@@ -244,28 +209,48 @@ export default function WalletModal({
...
+        nativeCurrencyBalance,
+        nativeCurrencyBalance.currency.decimals
+      )
+      user.set(CUSTOM_USER_PROPERTIES.WALLET_NATIVE_CURRENCY_AMOUNT, nativeCurrencyBalanceAmount)
+      user.set(CUSTOM_USER_PROPERTIES.WALLET_NATIVE_CURRENCY_BALANCE_USD, nativeCurrencyBalanceUsd)
+      setShouldLogNativeBalance(false)
+    }
+  }, [nativeCurrencyBalanceUsdValue, nativeCurrencyBalance, shouldLogNativeBalance, setShouldLogNativeBalance, native])
+
+  useEffect(() => {
```
vm, line 235: can you generalize this effect to use? very related logic between the three effects here

src/components/WalletModal/index.tsx
lynnshaoyu: i dont think it makes sense to because some of them might not be available / present for certain wallets, so we want to log them when / if a single one becomes available independent of the others

vm: that makes sense

src/components/AmplitudeAnalytics/constants.ts
```diff
@@ -37,14 +41,21 @@ export enum CUSTOM_USER_PROPERTIES {
...
   WALLET_TYPE = 'wallet_type',
+  WALLET_USDC_AMOUNT = 'wallet_usdc_amount',
+  WALLET_USDC_BALANCE_USD = 'wallet_usdc_balance_usd',
+  WALLET_WETH_AMOUNT = 'wallet_weth_amount',
+  WALLET_WETH_BALANCE_USD = 'wallet_weth_balance_usd',
 }
 
-export enum CUSTOM_USER_PROPERTY_SUFFIXES {
-  WALLET_TOKEN_AMOUNT_SUFFIX = '_token_amount',
+const ETH = nativeOnChain(1)
```
vm, line 53: is this and USDC not stored elsewhere as a constant?

lynnshaoyu: no they arent right now

src/components/WalletModal/index.tsx
lynnshaoyu: extracted to custom hook as per offline guidance, hope this looks good?

vm (approved):

PR #4434, opened 2022-08-22
build: enforce node/npm versions

vm:
This is the version we use to build on IPFS and is listed in our .nvmrc. Vanilla-extract also doesn't work on anything over 14 properly, so local builds won't work otherwise.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+engine-strict = true
--- a/package.json
+++ b/package.json
@@ -227,5 +227,9 @@
     "workbox-precaching": "^6.1.0",
     "workbox-routing": "^6.1.0",
     "zustand": "^4.0.0-rc.1"
+  },
+  "engines": {
+    "npm" : "8",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -227,5 +227,9 @@
     "workbox-precaching": "^6.1.0",
     "workbox-routing": "^6.1.0",
     "zustand": "^4.0.0-rc.1"
+  },
+  "engines": {
+    "npm" : "8",
```
zzmp, line 232: We use yarn, so I don't think we should set an `npm` version.

.npmrc
```diff
@@ -0,0 +1 @@
+engine-strict = true
```
vm, line 1: https://github.com/yarnpkg/yarn/issues/4895

vm: https://github.com/yarnpkg/yarn/issues/4895#issuecomment-545644733

zzmp (approved):

PR #4444, opened 2022-08-22
feat: wallet hover styling and moving wallet to correct navbar

aballerr:

This pr is focusing on two things
1. The wallet doesn't have complete hover states, this pr adds hover states to everything that is actionable.  
2. The wallet was part of old navbar and not toggable.  I've made it toggable through the new navbar

Other things to note:


src/components/AccountDetailsV2/index.tsx
```diff
@@ -25,6 +25,11 @@ const Grid = styled.a`
   text-decoration: none;
   border-bottom: ${({ theme }) => `1px solid ${theme.backgroundOutline}`};
   padding: 12px;
+
+  &:hover {
+    background-color: ${({ theme }) => theme.backgroundModule};
+    transition: 250ms background-color ease;
```
aballerr, line 31: Idk if there is a standard transition, but I use this everywhere in the pr

src/components/WalletDropdown/AuthenticatedHeader.tsx
```diff
@@ -81,6 +81,10 @@ const HeaderWrapper = styled.div`
   justify-content: space-between;
 `
 
+const AuthenticatedHeaderWrapper = styled.div`
```
aballerr, line 84: this component and some other where in in a parent wrapper with 16px padding.  In order to get the hover state right, I had to move that padding to the children components.  This is why this has padding now

src/components/WalletDropdown/LanguageMenu.tsx
```diff
@@ -29,6 +29,8 @@ const InternalLinkMenuItem = styled(InternalMenuItem)<{ isActive: boolean }>`
   color: ${({ theme }) => theme.textPrimary};
   :hover {
     cursor: pointer;
+    background-color: ${({ isActive, theme }) => !isActive && theme.backgroundModule};
```
aballerr, line 32: I disabled the hover state on the active language.  I thought it looked really weird imo and technically since it's selected it is not actionable

src/components/WalletDropdown/SlideOutMenu.tsx
```diff
@@ -29,12 +18,16 @@ const Header = styled.span`
...
-  margin-left: auto;
   color: ${({ theme }) => theme.accentAction};
   font-weight: 600;
   font-size: 14px;
   margin-top: auto;
   margin-bottom: auto;
+
+  :hover {
+    opacity: 0.6;
+    transition: 250ms opacity ease;
```
aballerr, line 29: I couldn't find a good style for the clear all hover state.  I reused the current hover state in the app.

src/components/Web3Status/index.tsx
```diff
@@ -215,9 +238,10 @@ export default function Web3Status() {
   const confirmed = sortedRecentTransactions.filter((tx) => tx.receipt).map((tx) => tx.hash)
 
   return (
-    <>
+    <span ref={ref}>
```
aballerr, line 241: I want the wallet to close when clicking outside of the container, this accomplishes that

src/hooks/useWindowSize.ts
```diff
@@ -29,3 +29,8 @@ export function useWindowSize() {
 
   return windowSize
 }
+
+export function useIsDesktopView() {
```
aballerr, line 33: navbar has a 1260 breakpoint.  I couldn't find this in the code, so I made this hook.  If there's something better, lmk

src/components/Header/index.tsx
```diff
@@ -1,11 +1,9 @@
 import { Trans } from '@lingui/macro'
 import useScrollPosition from '@react-hook/window-scroll'
 import { useWeb3React } from '@web3-react/core'
```
aballerr, line 3: wallet was in old header, I removed it.  I am worried we will forget it's in here and the logic to switch between the two isn't simple.  So, it's now only accessible through new nav

src/components/WalletDropdown/DefaultMenu.tsx
```diff
@@ -18,6 +18,8 @@ const ConnectButton = styled(ButtonPrimary)`
   width: 288px;
   font-weight: 600;
   font-size: 16px;
+  margin-left: auto;
```
vm, line 21: in general with these situations I recommend making the parent a flex and using justify-content: center. can you confirm this doesn't break where it's used currently?

src/components/WalletDropdown/LanguageMenu.tsx
vm: is there an `:active` element you could use instead, or is that unrelated?

src/components/WalletDropdown/SlideOutMenu.tsx
```diff
@@ -46,6 +39,26 @@ const StyledChevron = styled(ChevronLeft)`
...
+`
+
+const BackSectionContainer = styled.div`
+  display: flex;
+  justify-content: space-between;
+  position: relative;
+  width: 100%;
+`
+
+const ChildComponentsWrap = styled.div`
```
vm, line 58: ChildrenContainer

src/hooks/useWindowSize.ts
vm: we really should be trying to do this in css alone ideally. in this case why do you need it as a prop?

src/components/Web3Status/index.tsx
```diff
@@ -201,10 +208,26 @@ function Web3StatusInner() {
...
+
 export default function Web3Status() {
   const { ENSName } = useWeb3React()
 
   const allTransactions = useAllTransactions()
+  const ref = useRef<HTMLDivElement>(null)
+  const closeModal = useCloseModal(ApplicationModal.WALLET_DROPDOWN)
+  const isOpen = useIsOpen()
+
+  useOnClickOutside(ref, isOpen ? closeModal : undefined)
```
vm, line 230: fire this works better than what i tried to do, so will close my pr

src/components/Web3Status/index.tsx
```diff
@@ -133,6 +139,11 @@ function Web3StatusInner() {
     inputError: swapInputError,
   } = useDerivedSwapInfo()
   const validSwapQuote = getIsValidSwapQuote(trade, tradeState, swapInputError)
+  const walletFlag = useWalletFlag()
```
vm, line 142: i don't think you need this flag anymore

src/components/WalletDropdown/index.tsx
```diff
@@ -25,15 +29,46 @@ export enum MenuState {
...
+
+const WalletMobile = css`
+  left: 50%;
+  bottom: 45px;
+  transform: translateX(-50%);
+`
+
+const WalletDropdownWrapper = styled.div<{ isMobile: boolean }>`
+  position: absolute;
+  ${({ isMobile }) => (isMobile ? WalletMobile : WalletDesktop)};
```
vm, line 45: yea you can handle this completely in here rather than needing the prop.

vm (changes requested):
some comments

src/hooks/useWindowSize.ts
aballerr: if the breakpoint changes then it will be annoying to back and change this everywhere.  The other option is we set it as a constant and export it, then use a media query like you are suggesting

src/components/WalletDropdown/LanguageMenu.tsx
aballerr: this is unrelated styling.  By active, I mean the users selected language, not when they are clicking on it

src/hooks/useWindowSize.ts
vm: we should use the constant, there are some in Explore. i'm happy to move them into theme.

vm: https://github.com/Uniswap/interface/blob/main/src/components/Tokens/constants.ts

src/components/WalletDropdown/DefaultMenu.tsx
aballerr: I tried and it breaks, but I can do this in future

src/components/WalletDropdown/index.tsx
```diff
@@ -25,15 +26,37 @@ export enum MenuState {
   TRANSACTIONS = 'TRANSACTIONS',
 }
 
+const WalletDropdownWrapper = styled.div`
+  position: absolute;
+
+  @media only screen and (min-width: 1280px) {
```
vm, line 32: you don't need a breakpoint here, can just move this out a level

vm (approved):

PR #4445, opened 2022-08-22
chore: updating navbar padding

JackShort:
- updating navbar padding across token pages
- following figma guideline when available. If no figma guidelines available defaulting to 120px padding on page and none for mobile

### DIFF
--- a/src/pages/AddLiquidity/styled.tsx
+++ b/src/pages/AddLiquidity/styled.tsx
@@ -33,10 +33,14 @@ export const ScrollablePage = styled.div`
   position: relative;
   display: flex;
   flex-direction: column;
+  padding: 68px 0px 0px;
   ${({ theme }) => theme.mediaWidth.upToMedium`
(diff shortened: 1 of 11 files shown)


src/pages/AddLiquidity/styled.tsx
```diff
@@ -33,10 +33,14 @@ export const ScrollablePage = styled.div`
   position: relative;
   display: flex;
   flex-direction: column;
+  padding: 68px 0px 0px;
   ${({ theme }) => theme.mediaWidth.upToMedium`
     max-width: 480px;
     margin: 0 auto;
-    padding: 0px 8px;
+    padding: 32px 8px 0px;
```
vm, line 40: just update padding-top here

vm:
Discussed in DM, but basically need a way to update BodyWrapper since we can't expect every Page to handle this properly. However, NFT needs a banner that stretches the full page.

PR #4448, opened 2022-08-22
feat: add top tokens query for Explore tokens table

kayleegeorge:
top tokens query to fetch all 50 top tokens at once 
TODO: handle error and loading states, update svg, handle sorting

Screen Shot 2022-08-23 at 10 00 00 AM


src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -469,7 +467,7 @@ export default function LoadedRow({
 
   const heartColor = isFavorited ? theme.accentActive : undefined
   // TODO: consider using backend network?
-  const tokenRowData = useTokenRowQuery(tokenAddress, timePeriod, chainIdToChainName(filterNetwork))
+  // const topTokenData: any = useTopTokenQuery(favo, 1)
```
cmcewen, line 470: delete?

src/graphql/data/TopTokenQuery.ts
```diff
@@ -0,0 +1,83 @@
...
+        }
+      }
+    `,
+    {
+      page,
+    }
+  )
+
+  const topTokens: Record<string, TokenData> =
+    topTokenData.topTokenProjects?.reduce((acc, token) => {
```
cmcewen, line 61: i don't love changing the shape of the data on the client like this. ideally the API + objects are well designed so we just take them as they are

cmcewen (approved):
i'd like to merge this to unblock other folks but i think there is a bit of work to do still here

src/graphql/data/TopTokenQuery.ts
vm: I think that's an unrealistic expectation as this is used across different systems. Converting to an internal data representation is quite common.

cmcewen: sometimes, but why do we always index into the first market? should we only fetch one market then `(limit: 1)`? also think this has impact for entity caching as well

vm: ah i haven't used graphql much. yes i think we should just align the query properly, agreed.

kayleegeorge: I think we're doing this in all the other queries too (including price charts) so maybe we can update this and the error/loading states in a separate PR?

vm (approved):

PR #4451, opened 2022-08-23
fix: respect padding in cells with proper hover

cmcewen:
before:
Screen Shot 2022-08-23 at 12 47 43 PM

Screen Shot 2022-08-23 at 12 47 48 PM

after:

Screen Shot 2022-08-23 at 12 45 56 PM

Screen Shot 2022-08-23 at 12 46 02 PM

### DIFF
--- a/src/components/Tokens/TokenTable/TokenRow.tsx
+++ b/src/components/Tokens/TokenTable/TokenRow.tsx
@@ -44,7 +44,7 @@ const Cell = styled.div`
   align-items: center;
   justify-content: center;
 `
-const StyledTokenRow = styled.div`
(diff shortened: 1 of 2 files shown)


src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -54,10 +54,14 @@ const StyledTokenRow = styled.div`
 
   max-width: ${MAX_WIDTH_MEDIA_BREAKPOINT};
   min-width: 390px;
-  padding: 0px 12px;
+  padding-top: ${({ first }) => (first ? '4px' : '0px')};
```
vm, line 57: should be able to use :first-of-type and :last-of-type

vm (approved):

PR #4453, opened 2022-08-23
fix: change top tokens query return type

kayleegeorge:
change top tokens query return type from record to TokenData[], fixes favoriting reload behavior as well


src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -432,10 +432,9 @@ export default function LoadedRow({
   tokenData: TokenData
   timePeriod: TimePeriod
 }) {
-  const token = useToken(tokenAddress)
   const currency = useCurrency(tokenAddress)
-  const tokenName = token?.name ?? ''
-  const tokenSymbol = token?.symbol ?? ''
+  const tokenName = tokenData.name ?? ''
```
vm, line 436: i don't actually like defaulting to empty string here. why is it necessary?

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -48,34 +47,29 @@ const TokenRowsContainer = styled.div`
   width: 100%;
 `
 
-function useFilteredTokens(addresses: string[]) {
+function useFilteredTokens(tokens: TokenData[]) {
```
vm, line 50: should `useFilteredTokens` and `useSortedTokens` be conslidated?

src/graphql/data/TopTokenQuery.ts
```diff
@@ -1,8 +1,37 @@
...
+  YEAR,
+  ALL,
+}
+
+interface IAmount {
+  currency: Currency | null
+  value: number | null
+}
+
+export type TokenData = {
```
vm, line 20: how could it be null or undefined?

src/pages/Tokens/index.tsx
```diff
@@ -91,7 +91,7 @@ const Tokens = () => {
         </FiltersWrapper>
 
         <TokenTableContainer>
-          <TokenTable data={topTokens} error={error} loading={loading} />
+          <TokenTable data={topTokens ?? []} error={error} loading={loading} />
```
vm, line 94: better to allow TokenTable to take undefined for data imo

src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -432,10 +432,9 @@ export default function LoadedRow({
   tokenData: TokenData
   timePeriod: TimePeriod
 }) {
-  const token = useToken(tokenAddress)
   const currency = useCurrency(tokenAddress)
-  const tokenName = token?.name ?? ''
-  const tokenSymbol = token?.symbol ?? ''
+  const tokenName = tokenData.name
+  const tokenSymbol = tokenData.symbol?.toUpperCase()
```
vm, line 437: you can just do the uppercase in css

src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -458,7 +457,7 @@ export default function LoadedRow({
   const exploreTokenSelectedEventProperties = {
     chain_id: filterNetwork,
     token_address: tokenAddress,
-    token_symbol: token?.symbol,
+    token_symbol: tokenData?.symbol,
```
vm, line 460: then use tokenSymbol variable here

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -48,34 +47,29 @@ const TokenRowsContainer = styled.div`
...
+      shownTokens.filter((token) => {
         if (!filterString) {
           return true
         }
         const lowercaseFilterString = filterString.toLowerCase()
-        const addressIncludesFilterString = tokenAddress.toLowerCase().includes(lowercaseFilterString)
-        const nameIncludesFilterString = tokenName.toLowerCase().includes(lowercaseFilterString)
-        const symbolIncludesFilterString = tokenSymbol.toLowerCase().includes(lowercaseFilterString)
+        const addressIncludesFilterString = token.address.toLowerCase().includes(lowercaseFilterString)
+        const nameIncludesFilterString = token?.name?.toLowerCase().includes(lowercaseFilterString)
```
vm, line 64: can token be undefined?

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -48,34 +47,29 @@ const TokenRowsContainer = styled.div`
...
+        const symbolIncludesFilterString = token?.symbol?.toLowerCase().includes(lowercaseFilterString)
```
vm, line 65: same

src/components/Tokens/TokenTable/TokenTable.tsx
vm: and will this work as expected when undefined?

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -94,40 +88,39 @@ function useSortedTokens(addresses: string[], tokenData: Record<string, TokenDat
...
         let b: number
         switch (sortCategory) {
           case Category.marketCap:
-            a = token1.marketCap
-            b = token2.marketCap
-            break
-          case Category.percentChange:
-            a = token1.delta
-            b = token2.delta
+            a = token1.marketCap?.value ?? 0
```
vm, line 104: not convinced defaulting to 0 is correct here?

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -156,9 +149,8 @@ function LoadingTokenTable() {
 export default function TokenTable({ data, error, loading }: UseTopTokensResult) {
   const showFavorites = useAtomValue<boolean>(showFavoritesAtom)
   const timePeriod = useAtomValue<TimePeriod>(filterTimeAtom)
-  const topTokenAddresses = data ? Object.keys(data) : []
-  const filteredTokens = useFilteredTokens(topTokenAddresses)
-  const filteredAndSortedTokens = useSortedTokens(filteredTokens, data)
+  const filteredTokens = useFilteredTokens(data ?? [])
+  const filteredAndSortedTokens = useSortedTokens(filteredTokens)
```
vm, line 153: sortedFilteredTokens

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -156,9 +149,8 @@ function LoadingTokenTable() {
 export default function TokenTable({ data, error, loading }: UseTopTokensResult) {
   const showFavorites = useAtomValue<boolean>(showFavoritesAtom)
   const timePeriod = useAtomValue<TimePeriod>(filterTimeAtom)
-  const topTokenAddresses = data ? Object.keys(data) : []
-  const filteredTokens = useFilteredTokens(topTokenAddresses)
-  const filteredAndSortedTokens = useSortedTokens(filteredTokens, data)
+  const filteredTokens = useFilteredTokens(data ?? [])
```
vm, line 152: i'd allow useFilteredTokens to handle taking undefined

src/components/Tokens/TokenTable/TokenTable.tsx
kayleegeorge: ya same, if it's not there, should we just put it at the bottom? this is for sort right

src/graphql/data/TopTokenQuery.ts
kayleegeorge: that's the type the api returns

src/components/Tokens/TokenTable/TokenTable.tsx
vm: yes

src/graphql/data/TopTokenQuery.ts
vm: it returns `string | null | undefined`? what's the diff between when it returns null vs undefined?

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -48,34 +47,30 @@ const TokenRowsContainer = styled.div`
   width: 100%;
 `
 
-function useFilteredTokens(addresses: string[]) {
+function useFilteredTokens(tokens: TokenData[] | null | undefined) {
```
vm, line 50: why null or undefined? it should be only one of these

src/graphql/data/TopTokenQuery.ts
vm: i dont think any of these will be undefined t'll just be null

kayleegeorge: gives error when don't include undefined for those particular fields that still have it as an option O.o

vm (approved):

PR #4455, opened 2022-08-23
feat: Adding in remaining transactions to transaction history

aballerr:
This is a follow up to this commit
https://github.com/Uniswap/interface/commit/9b07d8ce64878f5becf88cb6d8e77639b3493299

Added the following transactions

Wrap
Collect Fees
Approval
Claim

I tested, this is what it looks like

Screen Shot 2022-08-23 at 4 31 04 PM
Screen Shot 2022-08-23 at 4 31 10 PM
Screen Shot 2022-08-23 at 4 31 13 PM

This is preloaded txs, but what uni and eth look like

Screen Shot 2022-08-23 at 4 38 36 PM

### DIFF
--- a/src/components/AccountDetailsV2/LogoView.tsx
+++ b/src/components/AccountDetailsV2/LogoView.tsx
@@ -1,11 +1,11 @@
+import { useWeb3React } from '@web3-react/core'
+import { UNI_ADDRESS } from 'constants/addresses'
 import { TransactionInfo, TransactionType } from 'state/transactions/types'
 import styled, { css } from 'styled-components/macro'
-import { UNI_ADDRESS } from 'constants/addresses'
 
+import { nativeOnChain } from '../../constants/tokens'
 import { useCurrency } from '../../hooks/Tokens'
 import CurrencyLogo from '../CurrencyLogo'
-import { nativeOnChain } from '../../constants/tokens'
-import { useWeb3React } from '@web3-react/core'
 
 const CurrencyWrap = styled.div`
   position: relative;
@@ -53,7 +53,7 @@ const getCurrency = ({ info, chainId }: { info: TransactionInfo; chainId: number
       return { currencyId0: unwrapped ? wrappedCurrency : base, currencyId1: unwrapped ? base : wrappedCurrency }
     case TransactionType.COLLECT_FEES:
       const { currencyId0, currencyId1 } = info
-      return { currencyId0: currencyId0, currencyId1: currencyId1 }
+      return { currencyId0, currencyId1 }
     case TransactionType.APPROVAL:
       return { currencyId0: info.tokenAddress, currencyId1: undefined }
     case TransactionType.CLAIM:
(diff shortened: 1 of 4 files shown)


src/components/AccountDetailsV2/TransactionBody.tsx
```diff
@@ -180,6 +187,128 @@ const CreateV3PoolSummary = ({
...
+        failed: <Trans>Wrap</Trans>,
+      }
+
+  return (
+    <>
+      <Action {...actionProps} />{' '}
+      <HighlightText>
+        {amount} {from}
+      </HighlightText>{' '}
+      <Trans>to</Trans>{' '}
```
vm, line 303: not sure translating this will work as expected

src/components/AccountDetailsV2/LogoView.tsx
```diff
@@ -32,22 +35,38 @@ interface CurrencyPair {
...
       const { baseCurrencyId, quoteCurrencyId } = info
       return { currencyId0: baseCurrencyId, currencyId1: quoteCurrencyId }
     case TransactionType.SWAP:
       const { inputCurrencyId, outputCurrencyId } = info
       return { currencyId0: inputCurrencyId, currencyId1: outputCurrencyId }
+    case TransactionType.WRAP:
+      const { unwrapped } = info
+      const native = info.chainId ? nativeOnChain(info.chainId) : undefined
+      const base = 'ETH'
+      const wrappedCurrency = native?.wrapped.address ?? 'WETH'
```
vm, line 52: we have a tokens constants file that should have this stuff btw

vm: also how does this work on Polygon?

src/components/AccountDetailsV2/TransactionBody.tsx
```diff
@@ -180,6 +187,128 @@ const CreateV3PoolSummary = ({
...
+    pending: <Trans>Collecting</Trans>,
+    success: <Trans>Collected</Trans>,
+    failed: <Trans>Collect</Trans>,
+  }
+
+  return (
+    <>
+      <Action {...actionProps} />{' '}
+      <FormattedCurrencyAmount rawAmount={expectedCurrencyOwed0} currencyId={currencyId0} sigFigs={2} />{' '}
+      <Trans>and</Trans>{' '}
```
vm, line 209: same issue, not sure translating here will work as expected bc of language differences

vm:
can you add line spacing on the text? they feel kinda tight

aballerr:
> can you add line spacing on the text? they feel kinda tight

I'm not sure I understand, can you be more specific

> also how does this work on Polygon?

This is the current logic implemented in TransactionSummary.tsx

src/components/AccountDetailsV2/TransactionBody.tsx
aballerr: I think you could say that about all translations on our site.  I don't really see a better alternative than having someone dedicated to making sure stuff translates as it is intended

vm (approved):

vm:
Screen Shot 2022-08-23 at 10 53 36 PM
i think this could use vertical spacing on the text

PR #4456, opened 2022-08-23
feat: make currencyInputPanel redesign changes apply only to swap

lynnshaoyu:
revert currency input panel in non-swap use cases. 

for example in add liquidity: 
Screen Shot 2022-08-24 at 1 45 44 AM

reason: redesign should exist in swap use case only, or else the other use cases look messed up.

### DIFF
--- a/src/components/CurrencyInputPanel/index.tsx
+++ b/src/components/CurrencyInputPanel/index.tsx
@@ -7,15 +7,14 @@ import { TraceEvent } from 'components/AmplitudeAnalytics/TraceEvent'
 import { AutoColumn } from 'components/Column'
 import { LoadingOpacityContainer, loadingOpacityMixin } from 'components/Loader/styled'
 import { isSupportedChain } from 'constants/chains'
-import { RedesignVariant, useRedesignFlag } from 'featureFlags/flags/redesign'
 import { darken } from 'polished'
 import { ReactNode, useCallback, useState } from 'react'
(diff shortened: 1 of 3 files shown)


src/components/CurrencyInputPanel/index.tsx
```diff
@@ -26,37 +24,35 @@ import { RowBetween, RowFixed } from '../Row'
...
   justify-content: center;
   z-index: 2;
 `
 
-const Container = styled.div<{ hideInput: boolean; disabled: boolean; redesignFlag: boolean }>`
+const Container = styled.div<{ hideInput: boolean; disabled: boolean }>`
   border-radius: ${({ hideInput }) => (hideInput ? '16px' : '20px')};
-  border: 1px solid ${({ theme, redesignFlag }) => (redesignFlag ? 'transparent' : theme.deprecated_bg0)};
-  background-color: ${({ theme, redesignFlag }) => (redesignFlag ? 'transparent' : theme.deprecated_bg1)};
+  border: 1px solid ${({ theme }) => theme.deprecated_bg0};
```
kayleegeorge, line 52: would we want these to be a different color other than deprecated_bg0 when redesignFlag is turned on though? like background.surface or something?

kayleegeorge: @callil

kayleegeorge (approved):
lgtm :-)

PR #4459, opened 2022-08-24
feat: update search icon

cbachmeier:
- Updates the search icons & fixes alignment

Desktop:
Screen Shot 2022-08-23 at 21 16 30

Mobile:
Screen Shot 2022-08-23 at 21 48 59

Alignment:
Screen Shot 2022-08-23 at 21 16 45

### DIFF
--- a/src/nft/components/icons.tsx
+++ b/src/nft/components/icons.tsx
@@ -964,7 +964,7 @@ export const MagnifyingGlassIcon = (props: SVGProps) => (
   <svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg" {...props}>
     <path
       d="M15 15L11.2439 11.2439M12.3821 6.69106C12.3821 9.83414 9.83414 12.3821 6.69106 12.3821C3.54797 12.3821 1 9.83414 1 6.69106C1 3.54797 3.54797 1 6.69106 1C9.83414 1 12.3821 3.54797 12.3821 6.69106Z"
-      stroke="#99A1BD"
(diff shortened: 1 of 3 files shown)


src/nft/components/icons.tsx
```diff
@@ -1299,8 +1292,8 @@ export const PlusIconLarge = (props: SVGProps) => (
 )
 
 export const ChevronLeftIcon = (props: SVGProps) => (
-  <svg fill="none" xmlns="http://www.w3.org/2000/svg" {...props}>
-    <path d="M15 6L9 12L15 18" stroke="currentColor" strokeWidth="2" strokeLinecap="round" strokeLinejoin="round" />
+  <svg width="8" height="16" viewBox="0 0 8 16" fill="none" xmlns="http://www.w3.org/2000/svg" {...props}>
```
vm, line 1295: i think we these are supposed to be passed in as props.

cbachmeier: these are defaults pulled from the Figma. Height and Width passed in via props would overwrite

JFrankfurt (approved):

vm (approved):

src/nft/components/icons.tsx
vm: Screen Shot 2022-08-24 at 10 00 55 AM
Screen Shot 2022-08-24 at 10 02 19 AM

In this case, it's not an issue since only used in one place, but `{...props}` won't override `height=` in the component. you'd need to explicitly destructure the prop and set a default on it.

PR #4465, opened 2022-08-24
chore: updating breakpoints

JackShort:
* updating breakpoints to new standard based on https://tailwindcss.com/docs/screens
* reason for doing so: improves granularity in responsiveness - also uses updated screen widths
* deprecated all of the old breakpoints

this would be the new way to perform media queries (which I think is much more readable):

Screen Shot 2022-08-24 at 1 30 34 PM

### DIFF
--- a/src/components/AccountDetails/index.tsx
+++ b/src/components/AccountDetails/index.tsx
@@ -24,7 +24,7 @@ const HeaderRow = styled.div`
   padding: 1rem 1rem;
   font-weight: 500;
   color: ${(props) => (props.color === 'blue' ? ({ theme }) => theme.deprecated_primary1 : 'inherit')};
-  ${({ theme }) => theme.mediaWidth.breakpointLg`
+  ${({ theme }) => theme.mediaWidth.deprecated_upToMedium`
     padding: 1rem;
   `};
 `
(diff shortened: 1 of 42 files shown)


src/components/AccountDetails/index.tsx
```diff
@@ -24,7 +24,7 @@ const HeaderRow = styled.div`
   padding: 1rem 1rem;
   font-weight: 500;
   color: ${(props) => (props.color === 'blue' ? ({ theme }) => theme.deprecated_primary1 : 'inherit')};
-  ${({ theme }) => theme.mediaWidth.upToMedium`
```
vm, line 27: i think we want to leave all the existing ones alone. `deprecated_upToMedium`

src/theme/styled.d.ts
```diff
@@ -128,10 +128,16 @@ declare module 'styled-components/macro' {
 
     // media queries
     mediaWidth: {
-      upToExtraSmall: ThemedCssFunction<DefaultTheme>
-      upToSmall: ThemedCssFunction<DefaultTheme>
-      upToMedium: ThemedCssFunction<DefaultTheme>
-      upToLarge: ThemedCssFunction<DefaultTheme>
+      deprecated_upToExtraSmall: ThemedCssFunction<DefaultTheme>
```
vm, line 131: what is this type? is it not just a number?

vm: we can deprecate the entire `mediaWidths` and make a new one called `breakpoints`

vm: turns out its doing some css snippet magic

vm (approved):

PR #4470, opened 2022-08-24
fix: fix decimal formatting of token price on details page

kayleegeorge:
Fix price formatting on token details page: Show 6 decimal points or <0.000001 (i.e. shouldn't show $0.00) - https://uniswaplabs.atlassian.net/browse/WEB-881

before:
image

after:
Screen Shot 2022-08-26 at 10 18 53 AM

### DIFF
--- a/src/components/Tokens/TokenDetails/PriceChart.tsx
+++ b/src/components/Tokens/TokenDetails/PriceChart.tsx
@@ -218,14 +218,11 @@ export function PriceChart({ width, height, token }: PriceChartProps) {
   const [delta, arrow] = getDelta(startingPrice.value, displayPrice.value)
   const crosshairEdgeMax = width * 0.85
   const crosshairAtEdge = !!crosshair && crosshair > crosshairEdgeMax
-  const priceNumDecimalPlaces = displayPrice.value < 1.05 ? 4 : 2
 
   return (
     <>
       <ChartHeader>
-        <TokenPrice>
-          {displayPrice.value < 0.0001 ? '$<0.0001' : displayPrice.value.toFixed(priceNumDecimalPlaces)}
-        </TokenPrice>
+        <TokenPrice>{displayPrice.value < 0.000001 ? '$<0.000001' : displayPrice.value.toFixed(6)}</TokenPrice>
         <DeltaContainer>
           {delta}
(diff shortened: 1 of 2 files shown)


src/components/Tokens/TokenDetails/PriceChart.tsx
```diff
@@ -218,11 +218,14 @@ export function PriceChart({ width, height, token }: PriceChartProps) {
...
   const crosshairEdgeMax = width * 0.85
   const crosshairAtEdge = !!crosshair && crosshair > crosshairEdgeMax
+  const priceNumDecimalPlaces = displayPrice.value < 1.05 ? 4 : 2
 
   return (
     <>
       <ChartHeader>
-        <TokenPrice>${displayPrice.value.toFixed(2)}</TokenPrice>
+        <TokenPrice>
+          {displayPrice.value < 0.0001 ? '$<0.0001' : displayPrice.value.toFixed(priceNumDecimalPlaces)}
```
vm, line 227: could we use `formatDollarAmount`?

vm: also noticed this is .0001 and the other is .001, not sure if intended

kayleegeorge: i was pondering, but i think they wanted to show up to four decimal points when it's < $1.05 and formatDollarAmount does $<0.001 (three decimal points) -- wdyt

vm: i think you should ask design to standardize and then use that util

lynnshaoyu: checkout formatToDecimal here: interface/src/components/AmplitudeAnalytics/utils.ts. 

also I think 6 significant figures for smaller values makes sense over 4 (it would be consistent with other parts of the codebase).

lynnshaoyu (approved):
approving, recommend using helper function formatToDecimal tho and 6 sig figs over 4

src/components/Tokens/TokenDetails/PriceChart.tsx
kayleegeorge: I agree! thanks Lynn

PR #4471, opened 2022-08-24
feat: update Chain Switcher Modal and Active State

cbachmeier:
- Updates design to match latest mocks
  - Adds extra info to active chain such as scan, bridge, helpCenter
- Adds active and unsupported state to network switcher button
- Added lightGrayOverlay to common styles as this is used across the app

Screenshots:
chainSwitcher

Unsupported case:
Screen Shot 2022-08-24 at 09 21 16


cmcewen (approved):

src/components/NavBar/ChainSwitcher.tsx
```diff
@@ -18,32 +24,51 @@ import { NavDropdown } from './NavDropdown'
...
+          <img src={logoUrl} alt={label} className={styles.Icon} />
+          {label}
+        </ChainDetails>
+        {active && <ActiveNetworkIcon />}
+      </Row>
+      {active && (
+        <Column gap="8" paddingBottom="16">
+          <Box className={styles.Separator} />
+          {bridge && <ChainLinkOut externalLink={bridge} label={`${label} bridge`} />}
+          {explorer && <ChainLinkOut externalLink={explorer} label={`${label} scan`} />}
```
vm, line 55: we'll want to name the relevant scanners like `Etherscan` in the CHAIN_INFO I think

src/components/NavBar/ChainSwitcher.tsx
```diff
@@ -18,32 +24,51 @@ import { NavDropdown } from './NavDropdown'
...
+        </Column>
+      )}
+    </Column>
   )
 }
 
 const ChainDetails = ({ children }: { children: ReactNode }) => <Row>{children}</Row>
 
+const ChainLinkOut = ({ externalLink, label }: { externalLink: string; label: string }) => (
+  <Row as="a" href={externalLink} className={styles.ChainInfo} target={'_blank'} rel={'noopener noreferrer'}>
```
JFrankfurt, line 66: I think at one point we had an external link component that added target/rel values automatically. Might be worth adding that logic here, or checking if it's already there.

src/components/NavBar/ChainSwitcher.tsx
```diff
@@ -68,33 +93,46 @@ export const ChainSwitcher = ({ isMobile }: ChainSwitcherProps) => {
...
+        background={isOpen ? 'accentActiveSoft' : 'none'}
+        onClick={toggleOpen}
+      >
+        {!chainId || !info ? (
+          <>
+            <TokenWarningRedIcon fill={themeVars.colors.darkGray} width={24} height={24} />
+            <Box as="span" className={subhead} style={{ lineHeight: '20px' }}>
+              {info?.label ?? 'Unsupported'}
+            </Box>
+          </>
```
JFrankfurt, line 111: If no chain id is provided, that would mean there's no network connection... Is `Unsupported` the right language for that scenario?

JFrankfurt (approved):
lgtm once label and bridge text are fixed

src/components/NavBar/ChainSwitcher.tsx
cbachmeier: That would be a question for @infredible . Unsupported is the language used in the Figma

cbachmeier:
From @infredible : we're going back to the old style and no longer supporting the extra links

src/components/NavBar/ChainSwitcher.tsx
vm: it's more than `chainId` being undefined means web3-react is still loading most likely. as long as that case is handled unsupported won't be necessary. rather, you'll need to use the `isChainSupported` helper to check if supported chainId

cbachmeier:
Current and approved by Design:
noActiveBg

src/components/NavBar/ChainSwitcher.tsx
cbachmeier: Thanks for the clarification, I believe the helper fn is named `isChainAllowed`

src/components/NavBar/ChainSwitcher.tsx
```diff
@@ -72,24 +74,43 @@ export const ChainSwitcher = ({ isMobile }: ChainSwitcherProps) => {
...
+        gap="8"
+        className={styles.ChainSwitcher}
+        background={isOpen ? 'accentActiveSoft' : 'none'}
+        onClick={toggleOpen}
+      >
+        {!isSupported ? (
+          <>
+            <TokenWarningRedIcon fill={themeVars.colors.darkGray} width={24} height={24} />
+            <Box as="span" className={subhead} style={{ lineHeight: '20px' }}>
+              {info?.label ?? 'Unsupported'}
```
vm, line 92: if the chain is unsupported it won't be in info, so can skip that. also should add a translate wrapper on this

cbachmeier: Localization for the Navbar will be handled in a subsequent PR

vm: i think this part does need to change though, since the info label case will never be hit

PR #4474, opened 2022-08-24
chore: updating navbar padding for token pages

JackShort:
* updates appropriate padding across all non nft pages

### DIFF
--- a/src/components/swap/styleds.tsx
+++ b/src/components/swap/styleds.tsx
@@ -8,10 +8,18 @@ import { Z_INDEX } from 'theme'
 
 import { AutoColumn } from '../Column'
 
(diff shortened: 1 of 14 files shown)


vm (approved):
Can you add pictures for all pages (at least just for the nav bar off version) and just confirm with Cal before merge? I'm okay with the copy paste as it they are somewhat unrelated even if similar.

JackShort:
> Can you add pictures for all pages (at least just for the nav bar off version) and just confirm with Cal before merge? I'm okay with the copy paste as they are somewhat unrelated even if similar.

going to send the deploy link to cal

JackShort:
cal approved the design

PR #4477, opened 2022-08-24
feat: Sample transitions pr

aballerr:
Opening this pr to get feedback and decide what theming for this could/should look like

### DIFF
--- a/src/theme/index.tsx
+++ b/src/theme/index.tsx
@@ -42,7 +42,7 @@ const transitions = {
     medium: 250,
     fast: 125,
   },
-  easing: {
+  timing: {
     in: 'ease-in',
     out: 'ease-out',
     inOut: 'ease-in-out',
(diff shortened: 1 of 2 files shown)


src/theme/index.tsx
```diff
@@ -36,6 +36,19 @@ const BREAKPOINTS = {
   xxxl: 1920,
 }
 
+const transitions = {
+  duration: {
+    slow: 500,
+    medium: 250,
+    fast: 125,
+  },
+  easing: {
```
vm, line 45: i would call this `timing` or `timingFunction` since you could also have `linear`, `cubic-bezier(n,n,n,n)`, etc. the css attribute is `transition-timing-function`.

vm (approved):
LGTM if others don't agree on my naming suggestion

src/theme/index.tsx
aballerr: this makes sense to me

vm:
would also name this pr `chore: transition states in theme`

PR #4481, opened 2022-08-25
feat: add token explore promo banner

kayleegeorge:
Add token explore promo banner - https://uniswaplabs.atlassian.net/browse/WEB-570
* uses local storage to store whether the user has interacted with the banner or closed it
* 
Screen Shot 2022-08-25 at 10 49 44 AM

Screen Shot 2022-08-25 at 10 49 52 AM


src/components/Tokens/TokensBanner.tsx
```diff
@@ -0,0 +1,86 @@
...
+  font-size: 12px;
+  line-height: 16px;
+  width: 240px;
+`
+
+export default function TokensBanner() {
+  const theme = useTheme()
+  const { pathname } = useLocation()
+  const currentPage = getCurrentPageFromLocation(pathname)
+  const [showBanner, setShowBanner] = useState(false)
```
vm, line 57: this should be in local storage right? only show once per user?

src/components/Tokens/TokensBanner.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const [tokensBannerDismissed, setTokensBannerDismissed] = useAtom(tokensBannerDismissedAtom)
+
+  useEffect(() => {
```
vm, line 60: think you should just render the TokensBanner inside of the Swap and PoolPage routes instead of needing to check these states

src/components/Tokens/TokensBanner.tsx
```diff
@@ -0,0 +1,86 @@
...
+      setTokensBannerDismissed(true)
+    }
+  }, [currentPage, setTokensBannerDismissed])
+
+  const closeBanner = () => {
+    setShowBanner(false)
+    setTokensBannerDismissed(true)
+  }
+  const clickBanner = () => {
+    window.location.href = 'https://app.uniswap.org/#/tokens'
```
vm, line 72: you'll want to use react-router Link

src/components/Tokens/TokensBanner.tsx
```diff
@@ -0,0 +1,86 @@
...
+  display: ${({ show }) => (show ? 'flex' : 'none')};
+  flex-direction: column;
+  padding: 12px 16px 12px 20px;
+  gap: 8px;
+  bottom: 48px;
+  right: 16px;
+  width: 320px;
+  height: 88px;
+  z-index: 5;
+  background-color: ${({ theme }) => (theme.darkMode ? theme.backgroundScrim : opacify(60, '#FDF0F8'))};
```
vm, line 25: why the specific case for lightMode here?

kayleegeorge: design had dark/light specific designs for this component in particular (noted in figma and the jira ticket)

src/components/Tokens/TokensBanner.tsx
```diff
@@ -0,0 +1,86 @@
...
+  font-weight: 600;
+  font-size: 14px;
+  line-height: 20px;
+  cursor: pointer;
+`
+const Description = styled.span`
+  font-weight: 400;
+  font-size: 12px;
+  line-height: 16px;
+  width: 240px;
```
vm, line 50: don't think you need a width set for this and popup container. would use a percentage width in here and then a fixed width on the banner. also does this handle mobile?

src/components/Tokens/TokensBanner.tsx
vm: i'm not set on this, but the logic in here is a bit brittle, trying to think of how to simplify

src/components/Tokens/state.ts
```diff
@@ -13,6 +13,7 @@ export const filterNetworkAtom = atom<SupportedChainId>(SupportedChainId.MAINNET
 export const filterTimeAtom = atom<TimePeriod>(TimePeriod.DAY)
 export const sortCategoryAtom = atom<Category>(Category.marketCap)
 export const sortDirectionAtom = atom<SortDirection>(SortDirection.decreasing)
+export const tokensBannerDismissedAtom = atomWithStorage<boolean>('tokensBannerDismissed', false)
```
vm, line 16: this should live in App level state... tbh would recommend using redux for this, the user slice

src/components/Tokens/TokensBanner.tsx
kayleegeorge: i think what you said is good tbh

src/components/Tokens/TokensBanner.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const [showTokensPromoBanner, setShowTokensPromoBanner] = useShowTokensPromoBanner()
+
+  const closeBanner = () => {
+    setShowTokensPromoBanner(false)
+  }
+
+  return (
+    <PopupContainer show={showTokensPromoBanner}>
+      <Header>
+        <HeaderText to={'https://app.uniswap.org/#/tokens'} onClick={closeBanner}>
```
cmcewen, line 66: can you make this a relative link? `/#/tokens`?

cmcewen (approved):

PR #4483, opened 2022-08-25
fix: Token Details - move About section below Market Stats

gbugyis:
- Moves About section below Market Stats

flip-about-and-stats-sections


infredible:
we should also switch the layout in the skeleton state

CleanShot 2022-08-25 at 12 51 03@2x

vm (approved):

PR #4489, opened 2022-08-25
fix: bug with network with no info

cbachmeier:
Previously if a network had no info, nothing would be shown. Now it is treated as Unsupported.

Before:
Screen Shot 2022-08-25 at 09 45 02

After:
Screen Shot 2022-08-25 at 09 41 30

### DIFF
--- a/src/components/NavBar/ChainSwitcher.tsx
+++ b/src/components/NavBar/ChainSwitcher.tsx
@@ -73,7 +73,7 @@ export const ChainSwitcher = ({ leftAlign }: ChainSwitcherProps) => {
   const selectChain = useSelectChain()
   useSyncChainQuery()
 
-  if (!chainId || !info) {
+  if (!chainId) {
     return null
   }
 
@@ -83,7 +83,7 @@ export const ChainSwitcher = ({ leftAlign }: ChainSwitcherProps) => {
     <NavDropdown top={54} leftAligned={leftAlign} paddingBottom={8} paddingTop={8}>
       <Column marginX="8">
         {NETWORK_SELECTOR_CHAINS.map((chainId: SupportedChainId) =>
-          isSupported ? (
+          isChainAllowed(chainId) ? (
             <ChainRow
               onSelectChain={async (targetChainId: SupportedChainId) => {
                 await selectChain(targetChainId)
@@ -107,7 +107,7 @@ export const ChainSwitcher = ({ leftAlign }: ChainSwitcherProps) => {
         background={isOpen ? 'accentActiveSoft' : 'none'}
         onClick={toggleOpen}
       >
-        {!isSupported ? (
+        {!isSupported || !info ? (
           <>
             <TokenWarningRedIcon fill={themeVars.colors.darkGray} width={24} height={24} />
             <Box as="span" className={subhead} display={{ sm: 'none', xl: 'flex' }} style={{ lineHeight: '20px' }}>
--- a/src/components/NavBar/ChainSwitcher.tsx
+++ b/src/components/NavBar/ChainSwitcher.tsx
@@ -12,7 +12,6 @@ import { subhead } from 'nft/css/common.css'
 import { themeVars, vars } from 'nft/css/sprinkles.css'
 import { useIsMobile } from 'nft/hooks'
 import { ReactNode, useReducer, useRef } from 'react'
-import { isChainAllowed } from 'utils/switchChain'
 
 import * as styles from './ChainSwitcher.css'
 import { NavDropdown } from './NavDropdown'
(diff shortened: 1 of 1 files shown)


src/components/NavBar/ChainSwitcher.tsx
```diff
@@ -83,7 +83,7 @@ export const ChainSwitcher = ({ leftAlign }: ChainSwitcherProps) => {
     <NavDropdown top={54} leftAligned={leftAlign} paddingBottom={8} paddingTop={8}>
       <Column marginX="8">
         {NETWORK_SELECTOR_CHAINS.map((chainId: SupportedChainId) =>
-          isSupported ? (
+          isChainAllowed(chainId) ? (
```
JFrankfurt, line 86: is this not the same, but with worse perf?

JFrankfurt, line 86: is this necessary? chains here should always be supported

vm: agreed

cbachmeier: removed

JFrankfurt (approved):

src/components/NavBar/ChainSwitcher.tsx
```diff
@@ -107,7 +105,7 @@ export const ChainSwitcher = ({ leftAlign }: ChainSwitcherProps) => {
         background={isOpen ? 'accentActiveSoft' : 'none'}
         onClick={toggleOpen}
       >
-        {!isSupported ? (
+        {!isSupported || !info ? (
```
vm, line 108: there shouldn't be a case where you need to check !info

vm: and inside this you can always render Unsupported vs checking for info.label

vm (changes requested):
above

src/components/NavBar/ChainSwitcher.tsx
cbachmeier: that was the point of this bug, Palm didn't have info so nothing was rendering. This also works as a null check for the info needed on the supported case

vm: if palm didn't have info wouldn't that be an unsupported chain?

vm: if we need to check `!info` for typescript reasons, could skip doing the `isSupported` check instead? they're functionally the same.

vm:
```ts
  const isSupported = !!info

  const dropdown = (
    <NavDropdown top={54} leftAligned={leftAlign} paddingBottom={8} paddingTop={8}>
      <Column marginX="8">
        {NETWORK_SELECTOR_CHAINS.map((chainId: SupportedChainId) => (
          <ChainRow
            onSelectChain={async (targetChainId: SupportedChainId) => {
              await selectChain(targetChainId)
              toggleOpen()
            }}
            targetChain={chainId}
            key={chainId}
          />
        ))}
      </Column>
    </NavDropdown>
  )

  return (
    <Box position="relative" ref={ref}>
      <Row
        as="button"
        gap="8"
        className={styles.ChainSwitcher}
        background={isOpen ? 'accentActiveSoft' : 'none'}
        onClick={toggleOpen}
      >
        {!isSupported ? (
          <>
            <TokenWarningRedIcon fill={themeVars.colors.darkGray} width={24} height={24} />
            <Box as="span" className={subhead} display={{ sm: 'none', xl: 'flex' }} style={{ lineHeight: '20px' }}>
              Unsupported
            </Box>
          </>
        ) : (
          <>
            <img src={info.logoUrl} alt={info.label} className={styles.Image} />
            <Box as="span" className={subhead} display={{ sm: 'none', xl: 'flex' }} style={{ lineHeight: '20px' }}>
              {info.label}
            </Box>
          </>
        )}
        {isOpen ? (
          <NewChevronUpIcon width={16} height={16} color="blackBlue" />
        ) : (
          <NewChevronDownIcon width={16} height={16} color="blackBlue" />
        )}
      </Row>
      {isOpen && (isMobile ? <Portal>{dropdown}</Portal> : <>{dropdown}</>)}
    </Box>
  )
```

cbachmeier: There's no where in `isChainAllowed` that explicitly enforces this condition, which is why if this check is removed we have to do null checks on all the info fields. I've added what this would look like. Right now it is possible that a Supported Chain is added without info which would then cause a a half loaded state where no icon or label would be shown.

vm (approved):

PR #4494, opened 2022-08-25
refactor: remove isChainAllowed

vm:
Since `isChainAllowed` was only being used for checking if Fortmatic could switch to a chain (only Mainnet was supported), we no longer need it. Now we can just check if it's an allowed chain, which `isSupportedChain` handles.

### DIFF
--- a/src/lib/hooks/useCurrency.ts
+++ b/src/lib/hooks/useCurrency.ts
@@ -2,11 +2,11 @@ import { arrayify } from '@ethersproject/bytes'
 import { parseBytes32String } from '@ethersproject/strings'
 import { Currency, Token } from '@uniswap/sdk-core'
 import { useWeb3React } from '@web3-react/core'
+import { isSupportedChain } from 'constants/chains'
 import { useBytes32TokenContract, useTokenContract } from 'hooks/useContract'
 import { NEVER_RELOAD, useSingleCallResult } from 'lib/hooks/multicall'
 import useNativeCurrency from 'lib/hooks/useNativeCurrency'
 import { useMemo } from 'react'
-import { isChainAllowed } from 'utils/switchChain'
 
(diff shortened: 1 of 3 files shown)


src/lib/hooks/useCurrency.ts
```diff
@@ -105,8 +105,8 @@ export function useCurrencyFromMap(tokens: TokenMap, currencyId?: string | null)
 
   const token = useTokenFromMapOrNetwork(tokens, isNative ? undefined : shorthandMatchAddress ?? currencyId)
 
-  const chainAllowed = chainId && isChainAllowed(chainId)
-  if (currencyId === null || currencyId === undefined || !chainAllowed) return null
+  const supportedChain = isSupportedChain(chainId)
+  if (currencyId === null || currencyId === undefined || !supportedChain) return null
```
cbachmeier, line 109: non blocking: we could move this check after the useWeb3React hook to potentially return null before setting isNative, token, etc.

cbachmeier (approved):
LGTM

src/lib/hooks/useCurrency.ts
vm: we can't because all the hooks need to happen without early returns in react unfortunately

cbachmeier: Sorry token would have to be before but it seems like `shorthandMatchAddress` and `isNative` both don't use hooks

cbachmeier: See their reqs in token

vm: `shorthandMatchAddress` uses `useMemo` though. `isNative` is used in the `useTokenFromMapOrNetwork` call so it also can't get moved too much.

PR #4498, opened 2022-08-25
fix: Wallet updates

aballerr:
This addresses these tasks in pending
https://uniswaplabs.atlassian.net/jira/software/c/projects/WEB/boards/9?selectedIssue=WEB-817&assignee=62b22b665479e75d631c4ea6

Change Summary
1. fixes screen resizing so wallet isn't randomly at top for 1280px screen width
2. Updates navbar connect button in disconnect state to include a dropdown option or a connect option
3. Updates scrollbar on chrome to be close to figma specs.  I don't think it's possible to change the size of the scrollbar thumb (that thing that moves)
4. Updates scrollbar on firefox to be close to figma specs.  Firefox doesn't support webkit, only chrome and safari do and it has limited ability to customize.  
5. Scrollbar would show on transactions page by default with certain mac system preferences.  It only now shows if content is scrollable
6. Removes active background-color on selected language
7. Makes selected language always have the same blue check mark
8. Hover state on the default menu doesn't fill screen width anymore
9. Updated hover states to have easing-in at 125ms
10. Add box shadow and border to wallet dropdown per figma specs

Screen Shot 2022-08-25 at 5 06 25 PM


src/components/WalletDropdown/index.tsx
```diff
@@ -17,6 +17,9 @@ const WalletWrapper = styled.div`
   top: 60px;
   right: 70px;
   background-color: ${({ theme }) => theme.backgroundSurface};
+  border: ${({ theme }) => `1px solid ${theme.backgroundOutline}`};
+  box-shadow: 0px 24px 32px rgba(0, 0, 0, 0.04), 0px 16px 24px rgba(0, 0, 0, 0.04), 0px 4px 8px rgba(0, 0, 0, 0.04),
```
aballerr, line 21: there is not theme for this box shadow

vm, line 21: could you confirm with fred about this box-shadow? i feel like we're trying to avoid non color variables recently

src/components/Web3Status/index.tsx
```diff
@@ -54,10 +54,24 @@ const Web3StatusError = styled(Web3StatusGeneric)`
...
 `
 
+const Web3StatusConnectNavbar = styled.button<{ faded?: boolean }>`
+  ${({ theme }) => theme.flexRowNoWrap}
+  background-color: ${({ theme }) => theme.accentActionSoft};
+  border-radius: 12px;
+  border: none;
+  cursor: pointer;
+
+  padding: 10px 12px 10px 12px;
```
vm, line 64: 10px 12px;

src/components/Web3Status/index.tsx
```diff
@@ -195,11 +251,23 @@ function Web3StatusInner() {
         properties={{ received_swap_quote: validSwapQuote }}
         element={ElementName.CONNECT_WALLET_BUTTON}
       >
-        <Web3StatusConnect onClick={toggleWallet} faded={!account}>
-          <Text>
-            <Trans>Connect Wallet</Trans>
-          </Text>
-        </Web3StatusConnect>
+        {navbarFlag === NavBarVariant.Enabled ? (
+          <Web3StatusConnectNavbar faded={!account}>
```
vm, line 255: perhaps could make this flex with align items center so you don't need the margin top/bottom auto? up to you

vm (approved):

src/components/WalletDropdown/index.tsx
aballerr: update on this, it was indeed not correct.  Fred updated to correct and I updated it to match that

PR #4499, opened 2022-08-25
feat: integrate widget settings/value

zzmp:
Updates @uniswap/widgets and integrates the SwapWidget as a controlled widget in Token Details.

- Integrates settings, such that widget settings update the app and vice versa
- Integrates values, such that the app token selector is used (instead of the widget's built-in selector)
  - The widget begins with the detailed token as the output token
  - The widget does not lock down tokens - it allows the user to switch input/output tokens and amounts


src/pages/TokenDetails/Widget.tsx
```diff
@@ -0,0 +1,183 @@
+import { TransactionReceipt } from '@ethersproject/abstract-provider'
```
vm, line 1: thoughts on putting the Widget in a higher level file since theoretically it could be used outside of token details?

src/pages/TokenDetails/Widget.tsx
```diff
@@ -0,0 +1,183 @@
...
+  )
+}
+
+// Integrates the Widget's settings, keeping the widget and app settings in sync.
+function useSyncWidgetSettings() {
+  const [userTtl, setUserTtl] = useUserTransactionTTL()
+  const [ttl, setTtl] = useState<number | undefined>(userTtl / 60)
+  const onTransactionDeadlineChange = useCallback(
+    (ttl: number | undefined) => {
+      setTtl(ttl)
```
vm, line 70: why do ttl and userTtl need to be stored separately?

src/pages/TokenDetails/Widget.tsx
```diff
@@ -0,0 +1,183 @@
...
+      const userTtl = ttl === undefined ? ttl : ttl * 60
+      setUserTtl(userTtl ?? DEFAULT_DEADLINE_FROM_NOW)
+    },
+    [setUserTtl]
+  )
+  const [userSlippage, setUserSlippage] = [useUserSlippageTolerance(), useSetUserSlippageTolerance()]
```
vm, line 76: shouldn't `[useUserSlippageTolerance]`return both of these like useState?

src/pages/TokenDetails/Widget.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const settingsHandlers: SwapEventHandlers = useMemo(
+    () => ({ onSettingsReset, onSlippageChange, onTransactionDeadlineChange }),
+    [onSettingsReset, onSlippageChange, onTransactionDeadlineChange]
+  )
+
+  return { settings: { settings, ...settingsHandlers } }
+}
+
+// Integrates the Widget's controlled value, using the app's CurrencySearchModal for token selection.
+function useSyncWidgetValue(defaultToken: Currency) {
```
vm, line 110: i would use `useSyncWidgetInputs`

src/pages/TokenDetails/Widget.tsx
```diff
@@ -0,0 +1,183 @@
...
+      selectedCurrency={selectingToken}
+      otherSelectedCurrency={otherToken}
+      onCurrencySelect={onTokenSelect}
+    />
+  )
+  return { value: { value, ...valueHandlers }, tokenSelector }
+}
+
+// Integrates the Widget's transactions, showing the widget's transactions in the app.
+function useSyncWidgetTransactions() {
```
vm, line 172: aren't these constant? not sure why they need to be memo'd or even a hook

src/pages/TokenDetails/Widget.tsx
vm: this function is doing a lot. would recommend splitting it into smaller helper hooks or something?

src/pages/TokenDetails/Widget.tsx
```diff
@@ -0,0 +1,183 @@
...
+        {...settings}
+        {...transactions}
+      />
+      {tokenSelector}
+    </>
+  )
+}
+
+// Integrates the Widget's settings, keeping the widget and app settings in sync.
+function useSyncWidgetSettings() {
```
vm, line 65: i think it'd be helpful to move the hooks into their own file

src/pages/TokenDetails/Widget.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const value: SwapController = useMemo(
+    () => ({ type: isExactInput ? TradeType.EXACT_INPUT : TradeType.EXACT_OUTPUT, amount, ...tokens }),
+    [amount, isExactInput, tokens]
+  )
+  const valueHandlers: SwapEventHandlers = useMemo(
+    () => ({ onAmountChange, onSwitchTokens, onTokenSelectorClick }),
+    [onAmountChange, onSwitchTokens, onTokenSelectorClick]
+  )
+
+  const tokenSelector = (
```
vm, line 159: does this need to get memo'd?

src/pages/TokenDetails/Widget.tsx
zzmp: they are different types, and different units, so they must be converted. this acts as the integration layer to eg allow undefined (defaulting) and convert between seconds/minutes.

src/pages/TokenDetails/Widget.tsx
zzmp: This is just a placeholder for a future integration that @JFrankfurt is working on

src/pages/TokenDetails/Widget.tsx
zzmp: i honestly don't know. because it is a component, I think it may be memoized by the react framework already.
I'll do some testing to figure this out.

src/pages/TokenDetails/Widget.tsx
zzmp: fixed the tech debt

src/pages/TokenDetails/Widget.tsx
zzmp: I don't think this needs to be memo-ized, but I don't think it will hurt.
I've memo-ized it by wrapping CurrencySearchModal in React.memo.

vm (approved):

src/components/Widget/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+        hideConnectionUI
+        jsonRpcUrlMap={RPC_URLS}
+        routerUrl={WIDGET_ROUTER_URL}
+        width={WIDGET_WIDTH}
+        locale={locale}
+        theme={theme}
+        provider={provider}
+        {...inputs}
+        {...settings}
+        {...transactions}
```
JFrankfurt, line 43: clever!

src/components/Widget/inputs.tsx
```diff
@@ -0,0 +1,66 @@
+import { Currency, Field, SwapController, SwapEventHandlers, TradeType } from '@uniswap/widgets'
+import CurrencySearchModal from 'components/SearchModal/CurrencySearchModal'
+import { useCallback, useMemo, useState } from 'react'
+
+// Integrates the Widget's controlled value, using the app's CurrencySearchModal for token selection.
+export function useSyncWidgetInputs(defaultToken: Currency) {
+  const [isExactInput, setIsExactInput] = useState(false)
```
JFrankfurt, line 7: vaguely feels to me like this should be using the type enum instead of a bool

src/components/Widget/inputs.tsx
```diff
@@ -0,0 +1,66 @@
...
+        return {
+          [otherField]: token === otherToken ? selectingToken : otherToken,
+          [selectingField]: token,
+        }
+      })
+    },
+    [otherField, otherToken, selectingField, selectingToken]
+  )
+  const value: SwapController = useMemo(
+    () => ({ type: isExactInput ? TradeType.EXACT_INPUT : TradeType.EXACT_OUTPUT, amount, ...tokens }),
```
JFrankfurt, line 47: same issue as above, feel like this could just be `type` from a replacement for `isExactInput` that uses the `TradeType` itself instead of a bool

src/pages/TokenDetails/index.tsx
```diff
@@ -137,22 +119,7 @@ export default function TokenDetails() {
...
-              onTxSubmit={onTxSubmit}
-              onTxSuccess={onTxSuccess}
-              onTxFail={onTxFail}
-              provider={provider}
-              routerUrl={ROUTER_URL}
-              theme={widgetTheme}
-              // tokenList={[]}
-              width={WIDGET_WIDTH}
-            />
+            {token && <Widget defaultToken={token} />}
```
JFrankfurt, line 122: is the conditional render necessary?

src/components/Widget/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+      />
```
JFrankfurt, line 44: I think this should include chainId from the query parameter (though prob not necessary to include in this pr)

JFrankfurt (approved):

src/pages/TokenDetails/Widget.tsx
zzmp: adding a note and prefixing the types to make this more apparent.

src/pages/TokenDetails/index.tsx
zzmp: Widget is typed to require `Currency`, not `Currency | undefined`, because the passed prop only defines the starting currency. If `defaultToken` started as `undefined` and then became defined (as happens when currencies load) then the widget would be initialized with `undefined` and not render the token when it loads. (This is intended, as the widget is spec'ed to be able to swap any token after initialization.) Because of this, the conditional render is necessary; because it's enforced by the type of Widget's props, I don't have a comment there.

zzmp: I'll add a comment to `WidgetProps` explaining this behavior.

src/components/Widget/index.tsx
zzmp: defaultChainId is inferred from the passed provider, so it should not be included in the props.
Adding this as a comment.

src/pages/TokenDetails/index.tsx
zzmp: I've tested, and I believe I was wrong here. I took out the conditional render.

PR #4507, opened 2022-08-26
feat: NavBar I18N

cbachmeier:
- Adds I18N support for the following elements in the NavBar:
  - Page Tabs
  - Searchbar
  - Overflow Menu
  - Chain Switcher
- Fixes a minor bug in the Chain switcher to right align dropdown on desktop


src/components/NavBar/SearchBar.tsx
```diff
@@ -316,6 +320,8 @@ export const SearchBar = () => {
     setSearchValue('')
   }, [pathname])
 
+  const placeholderText = phase1Flag === NftVariant.Enabled ? t`Search tokens and NFT collections` : t`Search tokens`
```
vm, line 323: you would make two separate <Trans> components here instead of importing t

src/constants/chainInfo.ts
```diff
@@ -60,7 +61,7 @@ const CHAIN_INFO: ChainInfoMap = {
     docs: 'https://docs.uniswap.org/',
     explorer: 'https://etherscan.io/',
     infoLink: 'https://info.uniswap.org/#/',
-    label: 'Ethereum',
+    label: t`Ethereum`,
```
vm, line 64: don't need to translate the chain names

src/components/NavBar/ChainSwitcher.tsx
```diff
@@ -79,7 +79,7 @@ export const ChainSwitcher = ({ leftAlign }: ChainSwitcherProps) => {
   const isSupported = !!info
 
   const dropdown = (
-    <NavDropdown top="56" left="0">
+    <NavDropdown top="56" left={leftAlign ? '0' : 'auto'} right={leftAlign ? 'auto' : '0'}>
```
vm, line 82: this seems unrelated?

cbachmeier: I noticed a small bug while working on this. Given that it was a two line fix it didn't seem like a valuable use of time to separate into its own PR

src/components/NavBar/SearchBar.tsx
cbachmeier: I can't set placeholder text with a JSX Element. Using `t` was previously used for placeholder text in `src/components/SearchModal/CurrencySearch.tsx` and `src/components/SearchModal/ManageLists.tsx` and `src/pages/CreateProposal/ProposalEditor.tsx`

src/constants/chainInfo.ts
cbachmeier: removed

src/components/NavBar/SearchBar.tsx
vm: check what i did in `SearchBar` in TokenTable?

vm: https://github.com/Uniswap/interface/blob/main/src/components/Tokens/TokenTable/SearchBar.tsx#L62

JFrankfurt: @vm why do you prefer not using `t` here?

src/constants/chainInfo.ts
```diff
@@ -2,8 +2,7 @@ import celoCircleLogoUrl from 'assets/images/celoCircle.png'
 import ethereumLogoUrl from 'assets/images/ethereum-logo.png'
 import optimismCircleLogoUrl from 'assets/images/optimismCircle.png'
 import polygonCircleLogoUrl from 'assets/images/polygonCircle.png'
-import arbitrumLogoUrl from 'assets/svg/arbitrum_logo.svg'
-import arbitrumCircleLogoUrl from 'assets/svg/arbitrum_logo.svg'
+import { default as arbitrumCircleLogoUrl, default as arbitrumLogoUrl } from 'assets/svg/arbitrum_logo.svg'
```
JFrankfurt, line 5: - why the diff here?
- why is the same thing imported twice as two different variable names?

cbachmeier: I made an edit and then undid it which caused some prettier styling to take place, it has no functional difference.

src/components/NavBar/SearchBar.tsx
cbachmeier: Also curious about not using `t`. Tried to implement the logic in the above file which causes a lot of bloat due to the feature flag conditional. Here's the code:
```
const SearchInput = () => {
    return phase1Flag === NftVariant.Enabled ? (
      <Trans
        render={({ translation }) => (
          <Box
            as="input"
            placeholder={`${translation}`}
            width="full"
            onChange={(event: ChangeEvent<HTMLInputElement>) => {
              !isOpen && toggleOpen()
            }}
            className={styles.searchBarInput}
            value={searchValue}
          />
        )}
      >
        Search tokens and NFT collections
      </Trans>
    ) : (
      <Trans
        render={({ translation }) => (
          <Box
            as="input"
            placeholder={`${translation}`}
            width="120"
            onChange={(event: ChangeEvent<HTMLInputElement>) => {
              !isOpen && toggleOpen()
            }}
            className={styles.searchBarInput}
            value={searchValue}
          />
        )}
      >
        Search tokens
      </Trans>
    )
  }

cbachmeier: I found an unrestricted replacement for t looking through the [lingui docs](https://lingui.js.org/ref/macro.html): `i18n`

vm (approved):
good after reverting last change, can use t for now

src/components/NavBar/SearchBar.tsx
vm: was just trying to avoid a restricted import. if it's actually okay we should unrestrict in the rest of the app imo.

PR #4508, opened 2022-08-26
fix: Fix to mobile wallet experience

aballerr:
The issue

The mobile wallet would close when clicking on or interacting with it.  This occurred because the wallet is coded to close any time a click occurs outside the web3status ref.  In mobile, we were using a portal, so the wallet was technically outside of this ref.  The reason wallet is under the status ref is due to a tricky edge case.  Clicking the status should toggle the wallet close and clicking outside the wallet should also close it.  If the ref only lives on the wallet and not status, this would lead to an edge case that would close and then immediately reopen the wallet.  I put the wallet under status so that when clicking on it, you are not technically clicking outside of wallet.  This results in the proper behavior of closing the wallet when clicking status and closing when clicking outside.

The solution
I made useClickOnOutside a bit more extensible.  It now take as many ref elements as you want for other elements that should not affect if it closes or not

### DIFF
--- a/src/components/WalletDropdown/index.tsx
+++ b/src/components/WalletDropdown/index.tsx
@@ -36,13 +36,13 @@ const WalletDropdownWrapper = styled.div`
   position: absolute;
   top: 65px;
   right: 20px;
+  z-index: 2;
 
   @media only screen and (max-width: ${({ theme }) => `${theme.breakpoint.sm}px`}) {
     top: unset;
     left: 0;
     right: 0;
     bottom: 56px;
-    z-index: 1;
   }
 `
 
--- a/src/components/WalletDropdown/index.tsx
+++ b/src/components/WalletDropdown/index.tsx
@@ -1,5 +1,6 @@
 import { useState } from 'react'
 import styled from 'styled-components/macro'
+import { Z_INDEX } from 'theme'
 
 import { useModalIsOpen } from '../../state/application/hooks'
 import { ApplicationModal } from '../../state/application/reducer'
@@ -36,7 +37,7 @@ const WalletDropdownWrapper = styled.div`
   position: absolute;
   top: 65px;
   right: 20px;
-  z-index: 2;
+  z-index: ${Z_INDEX.dropdown};
 
   @media only screen and (max-width: ${({ theme }) => `${theme.breakpoint.sm}px`}) {
     top: unset;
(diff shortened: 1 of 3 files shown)


aballerr:
I'm going to do some visual follow up on this, but wanted this logic to be separate here

src/components/WalletDropdown/index.tsx
```diff
@@ -36,13 +36,13 @@ const WalletDropdownWrapper = styled.div`
   position: absolute;
   top: 65px;
   right: 20px;
+  z-index: 2;
```
cmcewen, line 39: we have a `Z_INDEX` enum in the theme folder - i'm okay adding 2 in this case for it as a deprecated value if the other refactor would be a lot of work, but in the future hoping to move everything there 🤞

src/hooks/useOnClickOutside.ts
```diff
@@ -2,18 +2,26 @@ import { RefObject, useEffect, useRef } from 'react'
 
 export function useOnClickOutside<T extends HTMLElement>(
   node: RefObject<T | undefined>,
-  handler: undefined | (() => void)
+  handler: undefined | (() => void),
+  additionalNodes: Array<RefObject<T | undefined>> = []
```
cmcewen, line 6: ignoredNodes?

src/hooks/useOnClickOutside.ts
```diff
@@ -2,18 +2,26 @@ import { RefObject, useEffect, useRef } from 'react'
...
   const handlerRef = useRef<undefined | (() => void)>(handler)
+
   useEffect(() => {
     handlerRef.current = handler
   }, [handler])
 
   useEffect(() => {
     const handleClickOutside = (e: MouseEvent) => {
-      if (node.current?.contains(e.target as Node) ?? false) {
+      const additionalNodeClicked = additionalNodes.reduce(
```
cmcewen, line 16: for readability, i might do something like this:

```
const clickedNode = e.target as Node
const ignoredNodeClicked = ...
```

src/hooks/useOnClickOutside.ts
```diff
@@ -2,18 +2,27 @@ import { RefObject, useEffect, useRef } from 'react'
...
   const handlerRef = useRef<undefined | (() => void)>(handler)
+
   useEffect(() => {
     handlerRef.current = handler
   }, [handler])
 
   useEffect(() => {
     const handleClickOutside = (e: MouseEvent) => {
-      if (node.current?.contains(e.target as Node) ?? false) {
+      const nodeClicked = node.current?.contains(e.target as Node)
```
cmcewen, line 16: not blocking - but my previous comment was just to create a new variable `const clickedNode = e.target as Node` since you use it lower, and also it helps the reader understand the function a little easier

cmcewen (approved):

PR #4515, opened 2022-08-26
fix: wallet mobile ui changes

aballerr:
1. no border  radius on bottom of wallet for  mobile ui
2. no box shadow for mobile ui
3. Increased connect button width on mobile for disconnect
3. Stick back section was sticking out on language and transaction history menu.  I will come up with a better solution, but want this fixed for bug bash so I set width to 99% for now

### DIFF
--- a/src/components/WalletDropdown/SlideOutMenu.tsx
+++ b/src/components/WalletDropdown/SlideOutMenu.tsx
@@ -67,7 +67,7 @@ const StyledChevron = styled(ChevronLeft)`
 const BackSection = styled.div`
   position: absolute;
   background-color: ${({ theme }) => theme.backgroundSurface};
-  width: 100%;
+  width: 99%;
   padding: 0 16px 16px 16px;
   color: ${({ theme }) => theme.textSecondary};
   cursor: default;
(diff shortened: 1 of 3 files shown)


src/components/WalletDropdown/SlideOutMenu.tsx
```diff
@@ -67,7 +67,7 @@ const StyledChevron = styled(ChevronLeft)`
 const BackSection = styled.div`
   position: absolute;
   background-color: ${({ theme }) => theme.backgroundSurface};
-  width: 100%;
+  width: 99%;
```
aballerr, line 70: I will come up with something better later today, but I want wallet to look right for bug bash

aballerr: this juts out, it's a real eye sore rn

cmcewen (approved):

src/components/WalletDropdown/SlideOutMenu.tsx
aballerr: okay, this is jutting out b/c I added a border to the wallet.  The backsection overlays on top of the border

PR #4517, opened 2022-08-26
fix: added eth to explore

cartcrom:
Might need to change later depending on whether we have wETH and ETH separate

### DIFF
--- a/src/graphql/data/TopTokenQuery.ts
+++ b/src/graphql/data/TopTokenQuery.ts
@@ -34,6 +34,8 @@ export interface UseTopTokensResult {
   loading: boolean
 }
 
+const WRAPPED_ETH_ADDRESS = '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2'
+
 export function useTopTokenQuery(page: number) {
   const topTokenData = useLazyLoadQuery<TopTokenQueryType>(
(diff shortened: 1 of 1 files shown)


aballerr:
> Might need to change later depending on whether we have wETH and ETH separate

I would create a task in jira or a todo comment if this is important to make sure we account for this somewhere

src/graphql/data/TopTokenQuery.ts
```diff
@@ -115,11 +117,16 @@ export function useTopTokenQuery(page: number) {
     }
   )
 
-  const topTokens: TokenData[] | undefined = topTokenData.topTokenProjects?.map((token) =>
-    token?.tokens?.[0].address
+  const topTokens: TokenData[] | undefined = topTokenData.topTokenProjects?.map((token) => {
+    let address = token?.tokens?.[0].address
```
zzmp, line 121: Can this instead be compared to existing constants? See src/constants/tokens.ts specifically nativeOnChain (note you can also get this as a hook through useNativeCurrency).

You can probably compare this with ExtendedEther.onChain(chain) to make it more cross-chain compatible.

zzmp (changes requested):

vm:
This is close-able now?

PR #4525, opened 2022-08-29
fix: [TokenDetails] Glyph placement on line chart

gbugyis:
- Uses d3 curveCardinalOpen instead of d3-curve-circlecorners to fix inconsistency between glyph placement and chart line on steep data points.

### DIFF
--- a/package.json
+++ b/package.json
@@ -175,7 +175,6 @@
     "clsx": "^1.1.1",
     "copy-to-clipboard": "^3.2.0",
     "d3": "^7.6.1",
-    "d3-curve-circlecorners": "^0.1.6",
     "ethers": "^5.1.4",
     "firebase": "^9.1.3",
     "focus-visible": "^5.2.0",
--- a/src/components/Charts/LineChart.tsx
+++ b/src/components/Charts/LineChart.tsx
@@ -1,7 +1,6 @@
 import { Group } from '@visx/group'
 import { LinePath } from '@visx/shape'
 import { CurveFactory } from 'd3'
-import { radius } from 'd3-curve-circlecorners'
 import React from 'react'
 import { ReactNode } from 'react'
 import { useTheme } from 'styled-components/macro'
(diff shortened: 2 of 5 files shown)


src/components/Charts/LineChart.tsx
```diff
@@ -37,7 +36,7 @@ function LineChart<T>({
     <svg width={width} height={height}>
       <Group top={marginTop}>
         <LinePath
-          curve={curve ?? radius(0.25)}
+          curve={curve}
```
cartcrom, line 39: curve should potentially be a required prop here now, or we can choose a new default

src/components/Tokens/TokenDetails/PriceChart.tsx
```diff
@@ -246,8 +249,7 @@ export function PriceChart({ width, height, token }: PriceChartProps) {
         getX={(p: PricePoint) => timeScale(p.timestamp)}
         getY={(p: PricePoint) => rdScale(p.value)}
         marginTop={margin.top}
-        /* Default curve doesn't look good for the ALL chart */
-        curve={timePeriod === TimePeriod.ALL ? curveBasis : undefined}
+        curve={curveCardinalOpen.tension(curveTension)}
```
cartcrom, line 252: awesome

src/components/Charts/LineChart.tsx
gbugyis: ah, good catch. I made it required now

package.json
```diff
@@ -175,7 +175,6 @@
     "clsx": "^1.1.1",
     "copy-to-clipboard": "^3.2.0",
     "d3": "^7.6.1",
-    "d3-curve-circlecorners": "^0.1.6",
```
cartcrom, line 178: One last thing, could you remove the d3-curve-circlecorners from react-app-env.d.ts

cartcrom (changes requested):

cartcrom (approved):

PR #4528, opened 2022-08-29
fix: add null check to collection floor price

cbachmeier:
The trending collections endpoint is experiencing a bug and is returning collections with a null floor price. While this is being investigated by data team this allows for more graceful handling.

### DIFF
--- a/src/components/NavBar/SearchBar.tsx
+++ b/src/components/NavBar/SearchBar.tsx
@@ -141,7 +141,7 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }:
         return {
           ...collection,
           collectionAddress: collection.address,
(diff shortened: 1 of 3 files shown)


src/components/NavBar/SearchBar.tsx
```diff
@@ -141,7 +141,7 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }:
         return {
           ...collection,
           collectionAddress: collection.address,
-          floorPrice: formatEthPrice(collection.floor.toString()),
+          floorPrice: formatEthPrice(collection.floor?.toString()),
```
JackShort, line 144: can you type formatEthPrice so it accepts `undefined`?

JackShort (approved):
just type `formatEthPrice`

PR #4529, opened 2022-08-29
fix: catch invalid address token details

JackShort:
* catches invalid address error on token details page and redirects to token explore


aballerr (approved):

aballerr:
tested and works for me

PR #4532, opened 2022-08-29
feat: adding token safety article link

cartcrom:


src/constants/tokenSafety.tsx
```diff
@@ -2,8 +2,7 @@ import { Plural, Trans } from '@lingui/macro'
 
 import WarningCache, { TOKEN_LIST_TYPES } from './TokenSafetyLookupTable'
 
-// TODO: Replace this with Steph's article when it is available.
-export const TOKEN_SAFETY_ARTICLE = 'https://help.uniswap.org/en/'
+export const TOKEN_SAFETY_ARTICLE = 'https://support.uniswap.org/hc/en-us/articles/8723118437133'
```
aballerr, line 5: is this link correct? it doesn't direct me to any thing

cartcrom: @aballerr I thought the same thing, but the support team told me the article just currently isn't published, but the public will have access when we launch the features

aballerr: okay, makes sense to me.

aballerr (approved):

PR #4534, opened 2022-08-29
feat: token data cache

cartcrom:
Created a cache to store token PriceHistory for price charts.


src/components/Tokens/TokenDetails/PriceChart.tsx
```diff
@@ -162,20 +161,18 @@ export function PriceChart({ width, height, token }: PriceChartProps) {
   const theme = useTheme()
 
   // TODO: Add network selector input, consider using backend type instead of current front end selector type
-  const pricePoints: PricePoint[] = useTokenPriceQuery(token.address, timePeriod, 'ETHEREUM').filter(
-    (p): p is PricePoint => Boolean(p && p.value)
-  )
+  const { error, isLoading, data } = useTokenPriceQuery(token.address, 'ETHEREUM', timePeriod)
+
+  const pricePoints: PricePoint[] = data.filter((p): p is PricePoint => Boolean(p && p.value))
```
cmcewen, line 166: you should memo this so the array reference is constant

src/components/Tokens/TokenDetails/TokenDetail.tsx
```diff
@@ -218,10 +218,8 @@ export default function LoadedTokenDetail({ address }: { address: string }) {
     currency = nativeOnChain(connectedChainId)
   }
 
-  const tokenName = isWrappedNativeToken && currency ? currency.name : tokenDetailData.name
-  const defaultTokenSymbol = tokenDetailData.tokens?.[0]?.symbol ?? token.symbol
-  const tokenSymbol = isWrappedNativeToken && currency ? currency.symbol : defaultTokenSymbol
-
+  const tokenName = tokenDetailData.name
+  const tokenSymbol = tokenDetailData.tokens?.[0]?.symbol?.toUpperCase() ?? token.symbol
```
cmcewen, line 222: i think vig has mentioned uppercasing in CSS before - would vote to be consistent for whatever we choose

src/pages/TokenDetails/index.tsx
```diff
@@ -95,6 +98,14 @@ export default function TokenDetails() {
     return chainIds
   }, [connectedChainId])
 
+  const [timePeriod] = useAtom(filterTimeAtom)
+
+  // Preload token price data on first render
+  useMemo(() => {
+    if (token?.address) fillTokenPriceCache(token.address, 'ETHEREUM', timePeriod)
```
cmcewen, line 105: consider not running on first render: https://stackoverflow.com/a/57941438

src/graphql/data/TokenPrice.ts
```diff
@@ -0,0 +1,173 @@
...
+        address
+        symbol
+        decimals
+      }
+    }
+  }
+`
+
+export function useTokenPriceQuery(address: string, chain: Chain, timePeriod: TimePeriod) {
+  const [prices, setPrices] = useState<PriceHistory>(TokenAPICache.checkPriceHistory(address, timePeriod))
```
cmcewen, line 90: so relay is designed to handle caching itself: https://medium.com/jusbrasil-tech/caching-fields-on-relay-modern-with-hooks-30d76502e404. i think unfortunately my understanding of the backend implementation is that ids are not stable, so we won't be able to use that. if you confirm that i think this solution works for now but just want you to be aware

src/components/Tokens/TokenDetails/PriceChart.tsx
cmcewen: also - can you not do this as part of the query hook itself?

src/graphql/data/TokenPrice.ts
```diff
@@ -0,0 +1,173 @@
...
+  useEffect(() => {
+    const cached = TokenAPICache.checkPriceHistory(address, timePeriod)
+    if (cached) {
+      setPrices(cached)
+    } else {
+      fetchData()
+    }
+  }, [address, timePeriod, fetchData])
+
+  return { error, isLoading, data: prices ?? [] }
```
cmcewen, line 130: you'll return a new array for data every time - so just be mindful of object references. undefined might be better than an empty array

src/graphql/data/TokenPrice.ts
```diff
@@ -0,0 +1,183 @@
...
+      duration: toHistoryDuration(timePeriod),
+    }).subscribe({
+      start: () => setIsLoading(true),
+      next: updatePrices,
+      error: setError,
+      complete: () => setIsLoading(false),
+    })
+  }, [setIsLoading, address, chain, timePeriod, updatePrices, setError])
+
+  useEffect(() => {
```
cmcewen, line 131: so in this case, the thing you want to be careful of is triggering multiple fetches when you are already mid-fetch. which i don't think this actually helps with. your cache should be cheap to access, and then if you don't have a hit you want to fetch the data, and while you're fetching not trigger any other fetches.

this use effect doesn't help much with changing the address or time period because `checkPriceHistory` is cheap. you also may get to a stale cache by not expiring it in this function

cmcewen: so long story short - i don't think you need a useEffect here, but i do think you need some notion of `isFetching`

src/pages/TokenDetails/index.tsx
cmcewen: why can't you fetch all the data in one query?

src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -454,29 +455,30 @@ export function LoadingRow() {
...
+  const tokenSymbol = tokenData?.tokens?.[0].symbol
   const isFavorited = useIsFavorited(tokenAddress)
   const toggleFavorite = useToggleFavorite(tokenAddress)
   const filterString = useAtomValue(filterStringAtom)
   const filterNetwork = useAtomValue(filterNetworkAtom)
   const L2Icon = getChainInfo(filterNetwork).circleLogoUrl
-  const delta = tokenData.percentChange?.[timePeriod]?.value
-  const arrow = delta ? getDeltaArrow(delta) : null
-  const formattedDelta = delta ? formatDelta(delta) : null
+  //const delta = tokenData.percentChange?.[timePeriod]?.value
```
cmcewen, line 477: stray comment?

src/graphql/data/RelayEnvironment.ts
```diff
@@ -1,9 +1,37 @@
-import { Environment, Network, RecordSource, Store } from 'relay-runtime'
+import ms from 'ms.macro'
+import { Variables } from 'react-relay'
+import { Environment, Network, RecordSource, RequestParameters, Store } from 'relay-runtime'
+import RelayQueryResponseCache from 'relay-runtime/lib/network/RelayQueryResponseCache'
 
 import fetchGraphQL from './fetchGraphQL'
 
+export const cache = new RelayQueryResponseCache({ size: 250, ttl: ms`5m` })
```
cmcewen, line 8: did you figure out why 250 here?

src/graphql/data/Token.ts
```diff
@@ -0,0 +1,353 @@
...
+
+  cachedTopTokens = topTokens
+
+  return topTokens
+}
+
+type CachedTopToken = NonNullable<NonNullable<TokenTopQuery$data>['topTokenProjects']>[number]
+export function checkCachedTopToken(address: string): CachedTopToken {
+  let cachedTokenData: CachedTopToken = null
+  if (cachedTopTokens) {
```
cmcewen, line 131: why do this processing every time? why not forEach it once and then store the result?

src/pages/TokenDetails/index.tsx
```diff
@@ -137,15 +142,17 @@ export default function TokenDetails() {
       })
     : null
 
-  if (token === undefined) {
-    return <Navigate to={{ ...location, pathname: '/tokens' }} replace />
-  }
+  // TODO: Fix this logic to not automatically redirect on refresh, yet still catch invalid addresses
```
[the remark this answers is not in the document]
cartcrom: @JackShort pushed this code & knows about the automatic redirect on refresh issue. I commented it out cause I wanted to save his code but disabled it for now since it's hard to test/demo caching when refresh redirects

src/graphql/data/RelayEnvironment.ts
cartcrom: Yes, 250 means "max number of request in cache, least-recently updated entries purged first (default: 100)" from https://github.com/relay-tools/react-relay-network-modern. I'll add a comment explaining

cartcrom: I'm not sure what the best number to choose is here, 250 was just in their sample. Very open to suggestions

src/components/Tokens/state.ts
```diff
@@ -48,8 +49,11 @@ export function useSetSortCategory(category: Category) {
   }, [category, sortCategory, setSortCategory, sortDirection, setDirectionCategory])
 }
 
-export function useIsFavorited(tokenAddress: string) {
+export function useIsFavorited(tokenAddress: string | null | undefined) {
   const favoritedTokens = useAtomValue<string[]>(favoritesAtom)
 
-  return useMemo(() => favoritedTokens.includes(tokenAddress.toLocaleLowerCase()), [favoritedTokens, tokenAddress])
+  return useMemo(
```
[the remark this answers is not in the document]
cartcrom: I didn't add the memo, but agree. Should creating an object for the be a sep PR tho?

src/graphql/data/RelayEnvironment.ts
cmcewen: idk i'd use the default of 100 then if you don't have a clear reason why 250

src/graphql/data/Token.ts
```diff
@@ -0,0 +1,360 @@
...
+      const address = current?.tokens?.[0].address
+      if (address) acc[address] = current
+      return acc
+    }, {} as Record<string, CachedTopToken>) ?? {}
+  console.log(cachedTopTokens)
+
+  return topTokens
+}
+
+// export function checkCachedTopToken(address: string): CachedTopToken {
```
cmcewen, line 135: commented out?

cmcewen (approved):

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -47,33 +49,37 @@ const TokenRowsContainer = styled.div`
...
-        const lowercaseFilterString = filterString.toLowerCase()
-        const addressIncludesFilterString = token?.address?.toLowerCase().includes(lowercaseFilterString)
-        const nameIncludesFilterString = token?.name?.toLowerCase().includes(lowercaseFilterString)
-        const symbolIncludesFilterString = token?.symbol?.toLowerCase().includes(lowercaseFilterString)
-        return nameIncludesFilterString || symbolIncludesFilterString || addressIncludesFilterString
-      }),
-    [shownTokens, filterString]
+      data.topTokenProjects
+        ?.filter(
+          (token) => !showFavorites || (token?.tokens?.[0].address && favorites.includes(token?.tokens?.[0].address))
```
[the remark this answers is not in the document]
JFrankfurt: is the `[0]` indexing a data structure artifact, or is there some other significance to the first in the list? If the former, is there a straightforward way to resolve that so it's not accessed this way?

... Is this the list of (potentially multi-chain) tokens that @mzywang was telling us about? if so, should probable check if favorites includes *any* of the addresses, not just check against the first.

PR #4535, opened 2022-08-29
fix: missing logo redesign

lynnshaoyu:
Screen Shot 2022-08-29 at 3 47 25 PM
Screen Shot 2022-08-30 at 1 14 46 AM
Screen Shot 2022-08-30 at 1 17 25 AM

https://uniswaplabs.atlassian.net/browse/WEB-982?atlOrigin=eyJpIjoiY2M4ODFhY2ExZjcxNDVjNWIyMDMxYjFkNjliOWQ1NDIiLCJwIjoiaiJ9

### DIFF
--- a/src/components/CurrencyLogo/index.tsx
+++ b/src/components/CurrencyLogo/index.tsx
@@ -36,6 +36,7 @@ export default function CurrencyLogo({
     alt: `${currency?.symbol ?? 'token'} logo`,
     size,
     srcs: useCurrencyLogoURIs(currency),
+    symbol: currency?.symbol,
     style,
     ...rest,
   }
--- a/src/components/CurrencyLogo/index.tsx
+++ b/src/components/CurrencyLogo/index.tsx
@@ -24,11 +24,13 @@ const StyledNativeLogo = styled(StyledLogo)`
 
 export default function CurrencyLogo({
   currency,
+  symbol,
   size = '24px',
   style,
   ...rest
 }: {
   currency?: Currency | null
+  symbol?: string | null
   size?: string
   style?: React.CSSProperties
 }) {
@@ -36,7 +38,7 @@ export default function CurrencyLogo({
     alt: `${currency?.symbol ?? 'token'} logo`,
     size,
     srcs: useCurrencyLogoURIs(currency),
-    symbol: currency?.symbol,
+    symbol: symbol ?? currency?.symbol,
     style,
     ...rest,
   }
--- a/src/components/CurrencyLogo/index.tsx
+++ b/src/components/CurrencyLogo/index.tsx
@@ -30,7 +30,7 @@ export default function CurrencyLogo({
   ...rest
 }: {
   currency?: Currency | null
-  symbol?: string
+  symbol?: string | null
   size?: string
   style?: React.CSSProperties
 }) {
--- a/src/components/CurrencyLogo/index.tsx
+++ b/src/components/CurrencyLogo/index.tsx
@@ -30,7 +30,7 @@ export default function CurrencyLogo({
   ...rest
 }: {
   currency?: Currency | null
-  symbol?: string | null
+  symbol?: string
   size?: string
   style?: React.CSSProperties
 }) {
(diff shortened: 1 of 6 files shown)


aballerr:
tested and everything works well for me

src/components/CurrencyLogo/index.tsx
```diff
@@ -24,18 +24,21 @@ const StyledNativeLogo = styled(StyledLogo)`
 
 export default function CurrencyLogo({
   currency,
+  symbol,
   size = '24px',
   style,
   ...rest
 }: {
   currency?: Currency | null
+  symbol?: string | null
```
zzmp, line 33: I think this can be narrowed to `symbol?: string`, with no `undefined`.
It has no loading state, which I think is what we use `null` to signify.

src/components/Logo/index.tsx
```diff
@@ -34,5 +60,9 @@ export default function Logo({ srcs, alt, style, ...rest }: LogoProps) {
     )
   }
 
-  return <Slash {...rest} style={{ ...style, color: theme.deprecated_bg4 }} />
+  return (
+    <MissingImageLogo size={size} getFontSize={getFontSize}>
+      {symbol?.toUpperCase().replace('$', '').replace(/\s+/g, '').slice(0, 3)}
```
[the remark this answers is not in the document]
zzmp: I think the double replace makes this more readable.

Is the `slice` necessary? Does it need to be restricted to `3` characters?

src/components/Logo/index.tsx
```diff
@@ -1,22 +1,48 @@
...
 
 const BAD_SRCS: { [tokenAddress: string]: true } = {}
 
 interface LogoProps extends Pick<ImageProps, 'style' | 'alt' | 'className'> {
   srcs: string[]
+  symbol?: string
+  size?: string
 }
 
+const getFontSize = (size?: string) => {
```
zzmp, line 13: Is there a less bespoke way to do this? Can the font size always be a quarter of the size to standardize this?

zzmp (changes requested):

src/components/Logo/index.tsx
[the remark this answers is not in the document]
lynnshaoyu: yeah we do need the two replaces (i think), not sure how else I'd do this. also only first 3 chars is necessary, as determined with fred, because it looks cramped when more than 3 chars are permitted.

src/components/CurrencyLogo/index.tsx
lynnshaoyu: done

lynnshaoyu: actually line 505 in TokenRow complains (and the build breaks) if we remove null.

src/components/Logo/index.tsx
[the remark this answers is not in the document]
zzmp: Can you leave a comment that "it looks cramped when more than 3 chars are used"? Otherwise a future eng may call this out as a bug.

src/components/Logo/index.tsx
```diff
@@ -1,22 +1,36 @@
...
 }
 
+const getFontSize = (size?: string) =>
+  size ? Math.floor(parseInt(size.replace('px', '')) / 3).toString() + 'px' : '12px'
+
+const MissingImageLogo = styled.div<{ size?: string; getFontSize: (size?: string) => string }>`
+  border-radius: 100px;
+  color: ${({ theme }) => theme.textPrimary};
+  background-color: ${({ theme }) => theme.backgroundInteractive};
+  font-size: ${({ getFontSize, size }) => getFontSize(size)};
```
zzmp, line 20: Instead of passing a JS function can you use CSS `calc` so there's less logic being passed around?

zzmp: If you can't use `floor` in CSS, can you instead pass `fontSize` directly (you may not even need to go through `styled` in that case)? Calculating it in the CSS is confusing, as you're splitting where the logic executes.

zzmp (approved):
two changes, but no need to get re-approval if they're both made

src/components/Logo/index.tsx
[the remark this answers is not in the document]
lynnshaoyu: done

src/components/Logo/index.tsx
lynnshaoyu: done, used calc divide by 3 as an approx, looks the same visually.

PR #4536, opened 2022-08-29
chore: Cypress utility function for selecting feature flags and walletdrop down cypress tests

aballerr:
There is no task for this, but we will need to write cypress tests for new features soon

1. This pr contains a utility function to select a feature flag.  This will be helpful long term since new features should ideally have cypress tests preproduction.  
2. This pr tests the wallet dropdown in the disconnected state.  There seems to be some issues switching to the rinkby network after turning on the navbar flag.  
3. All tests pass

All tests pass locally
Screen Shot 2022-08-31 at 4 46 33 PM


cypress/utils/index.ts
```diff
@@ -0,0 +1,11 @@
+export const getTestSelector = (selectorId: string) => `[data-cy=${selectorId}]`
+
+// 1. Go to featureFlags/index.tsx
+// 2. Find the FeatureFlag enum
+// 3. Copy the string that correspond with the flag you want to enable and pass it here
```
aballerr, line 5: I would just import this enum, but I can't import from a .tsx file and idk if we should be importing cross libraries

cypress/e2e/wallet-dropdown.test.ts
```diff
@@ -0,0 +1,34 @@
+import { getTestSelector } from '../utils'
+
+describe('Wallet Dropdown', () => {
+  before(() => {
+    cy.visit('/')
+  })
+
+  it.skip('should change the theme', () => {
```
aballerr, line 8: flag isn't selectable, so I'm going to just skip these until we have a branch where we can comment out.

cypress/utils/index.ts
zzmp: It's fine to import here, as it's a coupled test library.
You should factor the flag into a separate TS file, so it can be included in this file's compilation.

cypress/utils/index.ts
```diff
@@ -0,0 +1,11 @@
...
+export const selectFeatureFlag = (featureFlag: string) => {
+  cy.get(getTestSelector('menu')).click()
+  cy.get(getTestSelector('feature-flags')).click()
+  cy.get(`#${featureFlag}`).select('enabled')
```
zzmp, line 9: This may not work, because loading a window clears the local storage (see cypress/support/e2e.ts?L=37).

Could this instead be added to the cypress/support/e2e.ts file, as part of the `visit` command?
Note that you can augment the `VisitOptions` type in that file, in L19 (as is done for `serviceWorker` tests).

cypress/e2e/wallet-dropdown.test.ts
zzmp: Why isn't it selectable?

zzmp (changes requested):

cypress/e2e/wallet-dropdown.test.ts
aballerr: I'm guessing feature flags is bit conditionally on the build type and the build type probably disables feature flags.  I looked at the video that from the failing tests and the feature flag option wasn't there

aballerr: Screen Shot 2022-08-31 at 4 29 11 PM


as you can see here, it times out b/c the flag doesn't exist.  I tested and ran this all locally.

cypress/utils/index.ts
aballerr: it does persist and this reload is consistent with the current logic for enabling widgets.  I think this is okay as is

vm (dismissed):
pretty sure we want to mock the flag in the hooks rather than update the states via this modal. this ties the UI to our testing which isn't ideal

aballerr:
> pretty sure we want to mock the flag in the hooks rather than update the states via this modal. this ties the UI to our testing which isn't ideal

This doesn't tie our ui to functional tests because our current cypress testing suite doesn't render a ui with the flags enabled.  In general, we aren't going to be writing functional tests for a feature that isn't done.  In this case, it makes sense to have them simply skipped until we are actually turning this feature on.

vm:
should be able to update the flags without using the flag modal by mocking the state within the hook. I agree that testing before features are released is a good idea, I'm more concerned about the cypress utility

src/featureFlags/index.tsx
```diff
@@ -1,13 +1,14 @@
...
 import { createContext, ReactNode, useCallback, useContext } from 'react'
+export { FeatureFlag } from './flags/featureFlags'
 
 interface FeatureFlagsContextType {
   isLoaded: boolean
   flags: Record<string, string>
 }
 
-const FeatureFlagContext = createContext<FeatureFlagsContextType>({ isLoaded: false, flags: {} })
+export const FeatureFlagContext = createContext<FeatureFlagsContextType>({ isLoaded: false, flags: {} })
```
zzmp, line 11: Why are you exporting this context?

cypress/utils/index.ts
zzmp: You should remove these comments now that you've exported the enum.

cypress/utils/index.ts
```diff
@@ -0,0 +1,13 @@
+import { FeatureFlag } from '../../src/featureFlags/flags/featureFlags'
+
+export const getTestSelector = (selectorId: string) => `[data-cy=${selectorId}]`
```
zzmp, line 3: Avoid using `data-cy` as it requires (this) custom implementation and introduces a second way to do something that's already covered.

Instead, use `data-testid`, and use the [`*ByTestId` selectors](https://testing-library.com/docs/queries/bytestid) from `@testing-library`.

cypress/e2e/wallet-dropdown.test.ts
zzmp: Either the env vars should be updated so that testing can cover this case, or the exported Context should be updated so that the flag hooks can be used in testing.
These shouldn't just be skipped - they're valuable tests.

zzmp (changes requested):

cypress/utils/index.ts
aballerr: I tried using that function and it turns and html element when cypress expects a string

src/components/WalletDropdown/IconButton.tsx
```diff
@@ -64,18 +64,19 @@ interface IconButtonProps {
   Icon: Icon
   onClick?: () => void
   href?: string
+  dataTestId?: string
```
zzmp, line 67: Can this be `['data-testId']` to keep it consistent?

cypress/support/e2e.ts
```diff
@@ -17,6 +19,7 @@ declare global {
     }
     interface VisitOptions {
       serviceWorker?: true
+      featureFlag?: FeatureFlag
```
zzmp, line 22: Amend to `featureFlags?: FeatureFlag[]`, so that it is possible to test with multiple flags at once.
You'll also need to update the logic in L44:

    const featureFlags = options.featureFlags.reduce((flags, flag) => {
      ...flags,
      [flag]: 'enabled',
    }, {})
    win.localStorage.setItem('featureFlags', featureFlags)

zzmp (approved):

PR #4537, opened 2022-08-29
fix: Fixing border overlap and reducing button size

aballerr:
https://uniswaplabs.atlassian.net/browse/WEB-988

1. The divider border top/bottom were overlapping causing it to be too big.  Setting only border-bottom fixes this
2. Request for button size to be reduced slightly

Screen Shot 2022-08-29 at 6 26 29 PM
Screen Shot 2022-08-29 at 6 26 34 PM

### DIFF
--- a/src/components/WalletDropdown/DefaultMenu.tsx
+++ b/src/components/WalletDropdown/DefaultMenu.tsx
@@ -27,7 +27,7 @@ const ConnectButton = styled(ButtonPrimary)`
 `
 
 const Divider = styled.div`
-  border: 0.5px solid ${({ theme }) => theme.backgroundOutline};
+  border-bottom: 1px solid ${({ theme }) => theme.backgroundOutline};
   margin-top: 16px;
   margin-bottom: 16px;
 `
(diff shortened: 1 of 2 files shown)


src/components/WalletDropdown/DefaultMenu.tsx
```diff
@@ -27,7 +27,7 @@ const ConnectButton = styled(ButtonPrimary)`
 `
 
 const Divider = styled.div`
-  border: 1px solid ${({ theme }) => theme.backgroundOutline};
+  border-bottom: 1px solid ${({ theme }) => theme.backgroundOutline};
```
zzmp, line 30: This is a duplicated component, with Transactionmenu.tsx (which already correctly uses border-bottom). Can you make a new `HorizontalRule.tsx` component (and have it export an aliased `Divider` element to make it more searchable) and use that?

zzmp: Fine to do this as a follow-up, too.

zzmp (approved):

PR #4547, opened 2022-08-31
fix: modify chart axis to to match crosshairs in year view

lynnshaoyu:
for year view we want to consistently show the 1st of the month, not vary between 1st and end of month on diff ticks. 

before:
Screen Shot 2022-08-31 at 2 12 49 AM

after:
Screen Shot 2022-08-31 at 2 06 15 AM

### DIFF
--- a/src/components/Tokens/TokenDetails/PriceChart.tsx
+++ b/src/components/Tokens/TokenDetails/PriceChart.tsx
@@ -18,7 +18,7 @@ import {
   dayHourFormatter,
   hourFormatter,
   monthDayFormatter,
-  monthFormatter,
+  monthTickFormatter,
   monthYearDayFormatter,
   monthYearFormatter,
   weekFormatter,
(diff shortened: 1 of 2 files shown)


src/components/Tokens/TokenDetails/PriceChart.tsx
```diff
@@ -117,11 +117,24 @@ const TimeButton = styled.button<{ active: boolean }>`
...
+function getTicks(startTimestamp: number, endTimestamp: number, forceStartOfMonthTicks = false, numTicks = 5) {
+  const ticks = Array.from(
     { length: numTicks },
     (v, i) => endTimestamp - ((endTimestamp - startTimestamp) / (numTicks + 1)) * (i + 1)
   )
+  if (!forceStartOfMonthTicks) return ticks
+
+  // when a tick maps to a date not on the first of the month, modify the tick to the closest
+  // first of month date. For example, Dec 31 becomes Jan 1, and Dec 5 becomes Dec 1.
+  return ticks.map((tick) => {
```
cartcrom, line 129: Would it be possible to instead put this logic in monthFormatter, or create a monthTickFormatter and pass that in at line 152?

lynnshaoyu: good idea, separated out into a different function getStartOfMonthTicks to make lines 152 and 156 more readable.

cartcrom (approved):
Looks perfect!

PR #4553, opened 2022-08-31
feat: integrate widget tx states

JFrankfurt:
image


src/components/Widget/transactions.ts
```diff
@@ -1,17 +1,76 @@
...
+    [addTransaction, allowedSlippage, chainId]
+  )
+
   const txHandlers: TransactionEventHandlers = useMemo(
     () => ({
-      onTxSubmit: (hash: string, tx: unknown) => console.log('onTxSubmit'),
-      onTxSuccess: (hash: string, receipt: TransactionReceipt) => console.log('onTxSuccess'),
-      onTxFail: (hash: string, receipt: TransactionReceipt) => console.log('onTxFail'),
+      onTxSubmit,
+      onTxSuccess: (hash: string, receipt: TransactionReceipt) => console.log('onTxSuccess', hash, receipt),
```
zzmp, line 70: If they're unneeded, rm `onTxSuccess`/`onTxFail` handlers

src/components/Widget/transactions.ts
```diff
@@ -1,17 +1,76 @@
...
+
+      if (!type || !response) {
+        return
+      }
+      if (type === TransactionType.WRAP || type === TransactionType.UNWRAP) {
+        const { amount } = transaction.info
+
+        addTransaction(response, {
+          type: TransactionType.WRAP,
+          unwrapped: false,
```
zzmp, line 34: This should be dependent on `TransactionType`, maybe:

    unwrapped: type === TransactionType.UNWRAP

src/components/Widget/transactions.ts
```diff
@@ -1,17 +1,76 @@
...
+      const { type, response } = transaction.info
+
+      if (!type || !response) {
+        return
+      }
+      if (type === TransactionType.WRAP || type === TransactionType.UNWRAP) {
+        const { amount } = transaction.info
+
+        addTransaction(response, {
+          type: TransactionType.WRAP,
```
zzmp, line 33: This must be the app's TransactionType

    import { TransactionType as AppTransactionType } from 'state/transactions/types'

Then you can drop the `unknown` conversion.

src/components/Widget/transactions.ts
```diff
@@ -1,17 +1,76 @@
...
+          unwrapped: false,
+          currencyAmountRaw: amount.quotient.toString(),
+          chainId,
+        } as unknown as WrapTransactionInfo)
+      }
+      if (type === TransactionType.SWAP) {
+        const { trade, tradeType } = transaction.info
+        const baseTxInfo = {
+          type: TransactionType.SWAP,
```
zzmp, line 42: This must be the app's TransactionType

    import { TransactionType as AppTransactionType } from 'state/transactions/types'

Then you can drop the `unknown` conversion.

src/components/Widget/transactions.ts
```diff
@@ -1,17 +1,76 @@
...
+import { useUserSlippageTolerance } from 'state/user/hooks'
+import { currencyId } from 'utils/currencyId'
 
 /** Integrates the Widget's transactions, showing the widget's transactions in the app. */
 export function useSyncWidgetTransactions() {
-  // TODO(jfrankfurt): Integrate widget transactions with app transaction tracking.
+  const { chainId } = useWeb3React()
+  const addTransaction = useTransactionAdder()
+  const [appSlippage] = useUserSlippageTolerance()
+  const allowedSlippage = appSlippage === 'auto' ? V3_SWAP_DEFAULT_SLIPPAGE : appSlippage
```
zzmp, line 20: The slippage is actually defined at the time of the transaction by the widget, and the default may not match the interface default. To keep the definition in one reliable place, I think it's best to include this in the trade object emitted by the widget (which will need to be added to it).

zzmp (changes requested):

src/components/Widget/transactions.ts
JFrankfurt: agree--this default value is what I felt most uncomfortable about in the pr

src/components/Widget/transactions.ts
```diff
@@ -26,20 +32,20 @@ export function useSyncWidgetTransactions() {
...
-          unwrapped: false,
+          type: AppTransactionType.WRAP,
+          unwrapped: type === WidgetTransactionType.UNWRAP,
           currencyAmountRaw: amount.quotient.toString(),
           chainId,
-        } as unknown as WrapTransactionInfo)
+        } as WrapTransactionInfo)
       }
-      if (type === TransactionType.SWAP) {
+      if (type === WidgetTransactionType.SWAP) {
```
zzmp, line 45: nit: `else if`

zzmp (changes requested):

src/components/Widget/transactions.ts
zzmp: See https://github.com/Uniswap/widgets/pull/187

JFrankfurt: done

zzmp (approved):

PR #4557, opened 2022-09-01
feat: add maxHeight to CurrencySearchModal

gbugyis:
Adjusts heights on CurrencySearchModal to cap height at 680px (per design) and prevent lists from becoming unreasonably long. 

Note: the Modal component expects `vh` units, so 680 is converted using the useWindowSize hook.


src/components/SearchModal/CurrencySearchModal.tsx
```diff
@@ -97,11 +98,17 @@ export default memo(function CurrencySearchModal({
     [setModalView, prevView]
   )
 
+  const { height } = useWindowSize()
   // change min height if not searching
   let minHeight: number | undefined = 80
```
zzmp, line 103: Can these be combined into a single `modalHeight`?

src/components/SearchModal/CurrencySearchModal.tsx
```diff
@@ -97,11 +98,17 @@ export default memo(function CurrencySearchModal({
...
+  let maxHeight: number | undefined = 80
   let content = null
   switch (modalView) {
     case CurrencyModalView.search:
+      if (height) {
+        minHeight = Math.min(Math.round((680 / height) * 100), 80)
```
zzmp, line 109: Add an inline comment that this converts from `px` to `vh`.

zzmp (changes requested):

src/components/SearchModal/CurrencySearchModal.tsx
```diff
@@ -97,11 +98,16 @@ export default memo(function CurrencySearchModal({
     [setModalView, prevView]
   )
 
+  const { height } = useWindowSize()
```
JFrankfurt, line 101: nit: `const { height: windowHeight } = useWindowSize()`

JFrankfurt: since you're dealing with a few different `height`s here

PR #4559, opened 2022-09-01
fix: change from flex to inline to fix safari bugs

cbachmeier:
- Fixes issues seen on Safari desktop and iOS where searchbar and magnifying glass icon would shift around
- Bug was caused by absolute positioning and flexbox being handled differently(and inconsistently) by Safari vs Chrome/Firefox
- Fixed by changing searchbar's container from flex to `inline-block`
- Undoes fix #4556 in favor of this solution

Screenshots:
Safari desktop
searchBarSafariFixed

Chrome on iOS
https://user-images.githubusercontent.com/11512321/187941529-5c642f33-a865-492c-9f16-f73607f01b9f.MP4

### DIFF
--- a/src/components/NavBar/SearchBar.css.ts
+++ b/src/components/NavBar/SearchBar.css.ts
@@ -27,6 +27,7 @@ export const searchBarContainer = style([
     right: '0',
     top: '0',
     zIndex: '3',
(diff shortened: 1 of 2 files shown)


JackShort (approved):
lgtm

PR #4560, opened 2022-09-01
feat: switch to using The Graph for all data

JFrankfurt:


src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -149,14 +161,15 @@ export function LoadingTokenTable() {
   )
 }
 
-export default function TokenTable({ data }: { data: TokenData[] | undefined }) {
+export default function TokenTable() {
   const showFavorites = useAtomValue<boolean>(showFavoritesAtom)
   const timePeriod = useAtomValue<TimePeriod>(filterTimeAtom)
+  const data = useLazyLoadQuery<TopTokenQueryType>(topTokensQuery, {})
```
cartcrom, line 167: Curious what your thoughts are on best practices / best hook to use here in the future when we implement refreshing of the data. I'm thinking useQueryLoaded called further up in the component tree, calling the returned loadQuery function on an interval, and passing the queryReference to TokenTable and calling usePreloadedQuery on it here

PR #4566, opened 2022-09-02
fix: Fixing key usage

aballerr:
Potential fix

https://uniswaplabs.atlassian.net/browse/WEB-1081


src/components/WalletDropdown/AuthenticatedHeader.tsx
```diff
@@ -110,7 +112,8 @@ const AuthenticatedHeader = () => {
     }
     connector.resetState()
     dispatch(updateSelectedWallet({ wallet: undefined }))
-  }, [connector, dispatch])
+    dispatch(removeConnectedWallet({ account, walletType }))
```
vm, line 115: what is this connected wallet? seems like duplicated state

PR #4567, opened 2022-09-02
fix: only initialize using chain query

vm:
This makes it so the ?chain=polygon query parameter is used on initialization or update (when user manually changes the url). When it maps to a supported chain id, we will prompt web3-react to switch chains.

However, we no longer will keep the chain synced in the query parameters to prevent against race conditions. For example, this solves the bug where ?chain=polygon is completely ignored on load currently.

I think this is a reasonable trade-off for now, and if we decide to move to a place where more of the page state is represented in the URL vs in redux/jotai state then we can re-evaluate a better way of architecting the sync.

### DIFF
--- a/src/hooks/useSyncChainQuery.ts
+++ b/src/hooks/useSyncChainQuery.ts
@@ -1,9 +1,8 @@
 import { useWeb3React } from '@web3-react/core'
-import { CHAIN_IDS_TO_NAMES, SupportedChainId } from 'constants/chains'
+import { CHAIN_IDS_TO_NAMES } from 'constants/chains'
 import { ParsedQs } from 'qs'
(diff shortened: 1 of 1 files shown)


vm:
One issue with this path is reload will cause the network to get switched back to mainnet. I think we'd need to store some state for just the network connector case, since wallets will have the latest chainId saved.

cmcewen:
@vm is there no way to mitigate the race conditions? or just it'd require a lot of work or plumbing some state around

zzmp (approved):
I don't like that this changes the state of the page on reload, but I think this is fine for now - you make a good point that the state is not actually stored in the URL, and doing so adds complexity.

PR #4570, opened 2022-09-04
use supported chain ids from sdk-core

JFrankfurt:


yarn.lock
```diff
@@ -4079,6 +4079,18 @@
     tiny-invariant "^1.1.0"
     toformat "^2.0.0"
 
+"@uniswap/sdk-core@^3.0.2":
+  version "3.0.2"
```
vm, line 4083: i'm surprised it's adding a new dep instead of upgrading the existing one?

vm (approved):

yarn.lock
JFrankfurt: image

JFrankfurt: yeah not sure what's happening here, but seems related.

PR #4573, opened 2022-09-06
feat: add pinning with 4everland

saullary:
integrate with 4everland hosting service to obtain distributed pin, and global accelerations


zzmp:
This would require creating a new account (for the token and project id).

PR #4575, opened 2022-09-06
fix: fixing token colors and token select persistence

aballerr:
Jira Tickets
1: https://uniswaplabs.atlassian.net/browse/WEB-1114
2. https://uniswaplabs.atlassian.net/browse/WEB-1109

What I did
1. I refactored the code so they both use the same styled components for the hearts.  They both now have the same color in light and dark mode. 
2.  The liked heart wasn't persisting b/c it was comparing address values with different cases.  I made everything lowercase for consistency
3. I added a hook to determine if an address is liked for consistency

### DIFF
--- a/src/components/Tokens/TokenDetails/TokenDetail.tsx
+++ b/src/components/Tokens/TokenDetails/TokenDetail.tsx
@@ -14,14 +14,14 @@ import { useAtomValue } from 'jotai/utils'
 import { darken } from 'polished'
 import { Suspense, useCallback } from 'react'
 import { useState } from 'react'
-import { ArrowLeft, Heart } from 'react-feather'
+import { ArrowLeft } from 'react-feather'
 import { useNavigate } from 'react-router-dom'
 import styled from 'styled-components/macro'
-import { ClickableStyle, CopyContractAddress } from 'theme'
+import { CopyContractAddress } from 'theme'
 import { formatDollarAmount } from 'utils/formatDollarAmt'
 
-import { favoritesAtom, filterNetworkAtom, useToggleFavorite } from '../state'
-import { ClickFavorited } from '../TokenTable/TokenRow'
+import { filterNetworkAtom, useIsFavorited, useToggleFavorite } from '../state'
+import { ClickFavorited, FavoriteIcon } from '../TokenTable/TokenRow'
 import LoadingTokenDetail from './LoadingTokenDetail'
 import Resource from './Resource'
 import ShareButton from './ShareButton'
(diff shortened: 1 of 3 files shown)


src/components/Tokens/state.ts
```diff
@@ -47,3 +47,9 @@ export function useSetSortCategory(category: Category) {
     }
   }, [category, sortCategory, setSortCategory, sortDirection, setDirectionCategory])
 }
+
+export function useIsFavorited(tokenAddress: string) {
+  const favoritedTokens = useAtomValue<string[]>(favoritesAtom)
+
+  return useMemo(() => favoritedTokens.includes(tokenAddress.toLocaleLowerCase()), [favoritedTokens, tokenAddress])
```
cmcewen, line 54: one thing that will only impact us is whether someone has favorited before you forced lowercase, in which case `favoritedTokens` may have some non lowercased values. because we haven't launched i think this is okay, but in general you just want to be mindful of cached state

cmcewen (approved):
nice

src/components/Tokens/state.ts
aballerr: ah good call out, didn't think of that.  If this pops up again, I'll keep this potential edge case in mind

src/components/Tokens/TokenDetails/TokenDetail.tsx
```diff
@@ -14,14 +14,14 @@ import { useAtomValue } from 'jotai/utils'
...
 import { useNavigate } from 'react-router-dom'
 import styled from 'styled-components/macro'
-import { ClickableStyle, CopyContractAddress } from 'theme'
+import { CopyContractAddress } from 'theme'
 import { formatDollarAmount } from 'utils/formatDollarAmt'
 
-import { favoritesAtom, filterNetworkAtom, useToggleFavorite } from '../state'
-import { ClickFavorited } from '../TokenTable/TokenRow'
+import { filterNetworkAtom, useIsFavorited, useToggleFavorite } from '../state'
+import { ClickFavorited, FavoriteIcon } from '../TokenTable/TokenRow'
```
cartcrom, line 24: Consider exporting a single \<TokenFavoriter /> component from TokenRow to get rid of duplicate code

cartcrom: oh whoops didn't realize this was already merged

aballerr: lmao no worries, if I touch this file again I'll update

PR #4576, opened 2022-09-06
fix: ensure nav bar goes above all other components when scrolling

lynnshaoyu:
pointed out by kenzie in slack: https://uniswapteam.slack.com/archives/CKX7SR7E1/p1660760838891389

Before: 
Screen Shot 2022-09-06 at 2 49 52 PM

After:
Screen Shot 2022-09-06 at 2 49 27 PM

### DIFF
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -12,6 +12,7 @@ import { lazy, Suspense, useEffect } from 'react'
 import { Navigate, Route, Routes, useLocation } from 'react-router-dom'
 import { useIsDarkMode } from 'state/user/hooks'
 import styled from 'styled-components/macro'
+import { Z_INDEX } from 'theme'
 import { SpinnerSVG } from 'theme/components'
 import { getBrowser } from 'utils/browser'
 import { getCLS, getFCP, getFID, getLCP, Metric } from 'web-vitals'
@@ -73,7 +74,7 @@ const HeaderWrapper = styled.div`
   justify-content: space-between;
   position: fixed;
   top: 0;
-  z-index: 9999;
+  z-index: ${Z_INDEX.absoluteTop};
 `
 
 const Marginer = styled.div`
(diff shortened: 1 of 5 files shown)


aballerr:
This worked for me for both old and new navbar in browser and mobile. This also solves another issue with the widget when an error or setting state shows

bugged state
Screen Shot 2022-09-06 at 3 17 38 PM

With preview
Screen Shot 2022-09-06 at 3 12 27 PM

Screen Shot 2022-09-06 at 3 20 08 PM

src/pages/App.tsx
```diff
@@ -73,7 +73,7 @@ const HeaderWrapper = styled.div`
   justify-content: space-between;
   position: fixed;
   top: 0;
-  z-index: 2;
+  z-index: 9999;
```
aballerr, line 76: maybe we should add an enum for this value and that here?
https://github.com/Uniswap/interface/blob/622581ee0aeaa6c72aa8345fe607810278ae5e0b/src/theme/index.tsx#L55

and refactor where this 9999 is used elsewhere in the app, I think it's just a couple of others

aballerr (approved):

lynnshaoyu:
marking as draft to fix this issue jack pointed out, dropdown should be on top and not cutoff by nav bar: 
Screen Shot 2022-09-06 at 3 30 02 PM

src/pages/App.tsx
lynnshaoyu: sounds like a good idea done

lynnshaoyu:
> marking as draft to fix this issue jack pointed out, dropdown should be on top and not cutoff by nav bar: Screen Shot 2022-09-06 at 3 30 02 PM

instead of upping the menu's z index i moved it down a bit so theres no overlap, similar to the chain and overflow menu dropdowns 

Screen Shot 2022-09-06 at 4 53 02 PM

PR #4584, opened 2022-09-08
feat: Add animations for desktop, tablet, and mobile searchbar

cbachmeier:
- ease-in in transition for searchbar appearing
- ease-out transition for searchbar disappearing
- had to re-write a number of styles that could not be animated (ie display)
- added an isTablet hook which is a copy of isMobile but with the correct breakpoints that was needed for these styles

Screenshots:
Desktop
searchAnimationDesktop

Tablet
searchAnimationTablet

Mobile
searchAnimationMobile


src/nft/hooks/useIsTablet.ts
```diff
@@ -0,0 +1,28 @@
+import { breakpoints } from 'nft/css/sprinkles.css'
+import { useEffect, useState } from 'react'
+
+const isClient = typeof window === 'object'
```
zzmp, line 4: Use `window !== undefined` instead of `window === 'object'`.

I'd consider it more idiomatic, because its documented by at least one other framework ([nextjs](https://nextjs.org/docs/migrating/from-create-react-app#safely-accessing-web-apis)).

cbachmeier: Made this change for both isTablet and isMobile

JackShort (approved):
lgtm - just don't need `clsx` in some places

src/components/NavBar/SearchBar.tsx
```diff
@@ -330,53 +331,62 @@ export const SearchBar = () => {
...
           onClick={() => !isOpen && toggleOpen()}
           gap="12"
         >
-          <Box display={{ sm: 'none', md: 'flex' }}>
-            <MagnifyingGlassIcon />
-          </Box>
-          <Box display={{ sm: 'flex', md: 'none' }} color="placeholder" onClick={toggleOpen}>
-            <ChevronLeftIcon />
+          <Box
+            className={clsx(
```
JackShort, line 361: clsx not necessary here

src/components/NavBar/SearchBar.tsx
```diff
@@ -330,53 +331,62 @@ export const SearchBar = () => {
...
           <Box
             as="input"
             placeholder={placeholderText}
-            width={isOpen || phase1Flag === NftVariant.Enabled ? 'full' : '120'}
             onChange={(event: ChangeEvent<HTMLInputElement>) => {
               !isOpen && toggleOpen()
               setSearchValue(event.target.value)
             }}
-            className={styles.searchBarInput}
+            className={clsx(
```
JackShort, line 379: same with here

PR #4586, opened 2022-09-08
feat: lazy load infinite scroll tokens

lynnshaoyu:


src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -149,14 +167,94 @@ export function LoadingTokenTable() {
...
+
+export default function TokenTable() {
   const showFavorites = useAtomValue<boolean>(showFavoritesAtom)
   const timePeriod = useAtomValue<TimePeriod>(filterTimeAtom)
-  const filteredTokens = useFilteredTokens(data)
-  const sortedFilteredTokens = useSortedTokens(filteredTokens)
+
+  const [page, setPage] = useState<number>(1)
+  const [topTokens, setTopTokens] = useState<TokenData[]>(
+    Array.from({ length: MAX_TOKENS_TO_LOAD }).map((_) => ({ loading: true } as TokenData))
```
cmcewen, line 182: can we use this for the `LOADING_ROWS` piece above?

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -149,14 +167,94 @@ export function LoadingTokenTable() {
...
+  const isItemLoaded = (index: number) => index < topTokens.length && topTokens[index].loading === false
+
+  const loadMoreItems = (startIndex: number, stopIndex: number) => {
+    if (stopIndex >= page * PAGE_SIZE && page < MAX_PAGE) setPage(page + 1)
+    return fetchQuery<TopTokenQueryType>(environment, query, {
+      pageSize: PAGE_SIZE,
+      page,
+    })
+      .toPromise()
+      .then((data) => {
```
cmcewen, line 197: i generally prefer async/await vs .then - is this because you have to specifically return a promise?

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -149,14 +167,94 @@ export function LoadingTokenTable() {
...
+        })
+        setTopTokens(currentTopTokens)
+      })
+      .catch((e) => setError(e))
+  }
+
+  const Row = function TokenRow({ data, index, style }: TokenRowProps) {
+    const tokenData = data[index]
+    if (!tokenData || topTokens[index].loading) {
+      return <LoadingRow style={style} key={index} />
```
cmcewen, line 241: is this also a dupe of the loading rows above?

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -169,31 +267,39 @@ export default function TokenTable({ data }: { data: TokenData[] | undefined })
...
-    </Suspense>
+    <GridContainer>
+      <HeaderRow />
+      <TokenDataContainer>
+        <AutoSizer>
+          {({ height, width }) => (
+            <InfiniteLoader isItemLoaded={isItemLoaded} itemCount={MAX_TOKENS_TO_LOAD} loadMoreItems={loadMoreItems}>
+              {({ onItemsRendered, ref }: { onItemsRendered: (props: ListOnItemsRenderedProps) => any; ref: any }) => (
+                <FixedSizeList
+                  className="List"
```
cmcewen, line 287: i think we prefer using `style` vs `className`

src/components/Tokens/TokenTable/TokenTable.tsx
lynnshaoyu: yeah we need to return a promise to the infinite loader

src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -364,6 +364,7 @@ function HeaderCell({
 
 /* Token Row: skeleton row component */
 export function TokenRow({
+  style,
```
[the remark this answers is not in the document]
lynnshaoyu: makes sense done

src/components/Tokens/TokenTable/TokenTable.tsx
lynnshaoyu: not relevant anymore, changed loading table code.

src/components/Tokens/TokenTable/TokenTable.tsx
lynnshaoyu: no this is different. the above loading table is for the suspense, if we dont have that we see the tiny little round spinner again.

src/components/Tokens/TokenTable/TokenTable.tsx
lynnshaoyu: i removed classname

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -43,41 +61,60 @@ const NoTokenDisplay = styled.div`
...
+
+function useFilteredTokens(
+  tokenData: TokenData[],
+  filterString: string,
+  showFavorites: boolean,
+  nonDefaultFiltersExist: boolean
+) {
   const favoriteTokenAddresses = useAtomValue(favoritesAtom)
-  const showFavorites = useAtomValue(showFavoritesAtom)
   const shownTokens =
```
cmcewen, line 78: i think you'll need to memo this for the below memo to work properly because it's a dep of that memo. it does say it creates a shallow copy of a portion of the array (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter), but i believe the array itself will be a new ref

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -92,41 +129,97 @@ function useSortedTokens(tokenData: TokenData[] | null) {
...
-        return sortFn(a, b)
-      }),
-    [tokenData, sortDirection, sortCategory, sortFn, timePeriod]
+      nonDefaultSortingParamsExist
+        ? tokenData &&
+          tokenData.sort((token1, token2) => {
+            if (!tokenData) {
+              return 0
+            }
+            // fix delta/percent change property
```
cmcewen, line 138: i know this was here before, but i'm not sure what this comment means exactly

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -92,41 +129,97 @@ function useSortedTokens(tokenData: TokenData[] | null) {
...
+  topTokens: TokenData[]
+) {
+  fetchQuery<TopTokenQueryType>(environment, query, {
+    pageSize,
+    page,
+  })
+    .toPromise()
+    .then((data) => {
+      const newPageTopTokens: TokenData[] = !!data?.topTokenProjects
+        ? data.topTokenProjects.map((token) =>
```
cmcewen, line 184: could you split this function out into a separate one to make it more readable? `formatTokenProjects` or something?

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -92,41 +129,97 @@ function useSortedTokens(tokenData: TokenData[] | null) {
...
+                  },
+                }
+              : ({ loading: false } as TokenData)
+          )
+        : []
+      const currentTopTokens = [...topTokens]
+      newPageTopTokens.forEach((token, i) => {
+        if (currentTopTokens[startIndex + i].loading) currentTopTokens[startIndex + i] = token
+      })
+      setTopTokens(currentTopTokens)
```
cmcewen, line 218: could this create a race condition if you call it multiple times?

src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -430,7 +431,7 @@ export function HeaderRow() {
 }
 
 /* Loading State: row component with loading bubbles */
-export function LoadingRow() {
+export function LoadingRow({ style }: { style?: CSSProperties }) {
```
JFrankfurt, line 434: When setting up passthroughs like this I generally try to do something like `{...props}` on the styled child instead of individually defining the props here.

src/components/Tokens/TokenTable/TokenTable.tsx
JFrankfurt: +1

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -43,41 +61,60 @@ const NoTokenDisplay = styled.div`
...
+  tokenData: TokenData[],
+  filterString: string,
+  showFavorites: boolean,
+  nonDefaultFiltersExist: boolean
+) {
   const favoriteTokenAddresses = useAtomValue(favoritesAtom)
-  const showFavorites = useAtomValue(showFavoritesAtom)
   const shownTokens =
-    showFavorites && tokens ? tokens.filter((token) => favoriteTokenAddresses.includes(token.address)) : tokens
+    showFavorites && tokenData ? tokenData.filter((token) => favoriteTokenAddresses.includes(token.address)) : tokenData
```
JFrankfurt, line 79: the type signature `useFilteredTokens` says this check shouldn't be necessary and that the `.filter` call will always happen. Is that incorrect? should there be an `undefined` or `?` on the `tokenData` type?

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -136,27 +229,82 @@ function NoTokensState({ message }: { message: ReactNode }) {
...
+  const [page, setPage] = useState<number>(1)
+  const [topTokens, setTopTokens] = useState<TokenData[]>(
+    Array.from({ length: MAX_TOKENS_TO_LOAD }).map((_) => ({ loading: true } as TokenData))
+  )
+  const [error, setError] = useState<any>(undefined)
+
+  const preloadedTopTokens = useTopTokenQuery(MAX_TOKENS_TO_LOAD, 1) ?? []
+  const nonDefaultFiltersExist = hasNonDefaultFilters(filterString, showFavorites)
+  const nonDefaultSortingParamsExist = hasNonDefaultSortingParams(sortCategory, sortDirection, timePeriod)
+  const usePreloadedTokens = nonDefaultFiltersExist || nonDefaultSortingParamsExist
```
JFrankfurt, line 271: `exist` here is a little confusing to me here. Its usage as a flag to skip sort/filtering is also kinda confusing to me.

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -136,27 +229,82 @@ function NoTokensState({ message }: { message: ReactNode }) {
...
+    timePeriod,
+    nonDefaultSortingParamsExist
+  )
+
+  const isItemLoaded = (index: number) => index < topTokens.length && topTokens[index].loading === false
+
+  const loadMoreItems = (startIndex: number, stopIndex: number) => {
+    if (stopIndex >= page * PAGE_SIZE && page < MAX_PAGE) setPage(page + 1)
+    return createFetchTokensPromise(page, PAGE_SIZE, startIndex, setTopTokens, setError, topTokens)
+  }
```
JFrankfurt, line 286: should this be a `useCallback`?

src/components/Tokens/state.ts
```diff
@@ -15,10 +15,11 @@ export const sortCategoryAtom = atom<Category>(Category.marketCap)
 export const sortDirectionAtom = atom<SortDirection>(SortDirection.decreasing)
 
 /* for favoriting tokens */
-export function useToggleFavorite(tokenAddress: string) {
+export function useToggleFavorite(tokenAddress: string | null | undefined) {
```
JFrankfurt, line 18: can this work w/ just a `?`

JFrankfurt (changes requested):

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -136,27 +229,82 @@ function NoTokensState({ message }: { message: ReactNode }) {
...
+    filteredTokens,
+    sortCategory,
+    sortDirection,
+    timePeriod,
+    nonDefaultSortingParamsExist
+  )
+
+  const isItemLoaded = (index: number) => index < topTokens.length && topTokens[index].loading === false
+
+  const loadMoreItems = (startIndex: number, stopIndex: number) => {
```
cartcrom, line 283: This function, along with createFetchTokensPromise, doesn't actually return a promise. Does react window actually need us to use promises then, or can we use relay-hooks

src/components/Tokens/TokenTable/TokenRow.tsx
JFrankfurt: bumping this

src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -492,6 +496,7 @@ export default function LoadedRow({
   // TODO: currency logo sizing mobile (32px) vs. desktop (24px)
   return (
     <StyledLink
+      style={style}
```
JFrankfurt, line 499: same comment re: spreading element styles instead of enumerating each pass-through prop

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -48,26 +57,53 @@ function NoTokensState({ message }: { message: ReactNode }) {
...
 export default function TokenTable() {
   const showFavorites = useAtomValue<boolean>(showFavoritesAtom)
 
   // TODO: consider moving prefetched call into app.tsx and passing it here, use a preloaded call & updated on interval every 60s
   const prefetchedTokens = usePrefetchTopTokens()
   const { loading, tokens, loadMoreTokens } = useTopTokens(prefetchedTokens)
 
+  const isItemLoaded = (index: number) => !loading && !!tokens && index < tokens.length
+
+  const Row = function TokenRow({ index, style }: TokenRowProps) {
```
JFrankfurt, line 91: `const Row = useCallback(...` instead

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -92,15 +128,39 @@ export default function TokenTable() {
...
+                    }) => (
+                      <FixedSizeList
+                        height={height}
+                        width={width}
+                        itemCount={MAX_TOKENS_TO_LOAD}
+                        itemSize={70}
+                        onItemsRendered={onItemsRendered}
+                        ref={ref}
+                      >
+                        {Row}
```
JFrankfurt, line 156: not sure about the `FixedSizeList` api, but I think you should be able to use the function above as a normal component here. <Row />

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -92,15 +128,39 @@ export default function TokenTable() {
     } else {
       return (
         <>
-          <GridContainer>
+          <GridContainer fixedHeight={true}>
```
JFrankfurt, line 131: can drop `={true}` and just name the prop for the same effect

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -48,26 +57,53 @@ function NoTokensState({ message }: { message: ReactNode }) {
...
+  const isItemLoaded = (index: number) => !loading && !!tokens && index < tokens.length
+
+  const Row = function TokenRow({ index, style }: TokenRowProps) {
+    const token = !!tokens && tokens[index]
+    if (!token || loading) {
+      return <LoadingRow style={style} key={index} />
+    }
+    return (
+      <LoadedRow
+        style={style}
```
JFrankfurt, line 98: same comment re: `style` passthrough

JFrankfurt (changes requested):

src/components/Tokens/TokenTable/TokenRow.tsx
lynnshaoyu: done

src/components/Tokens/TokenTable/TokenTable.tsx
lynnshaoyu: sgtm done

src/components/Tokens/TokenTable/TokenTable.tsx
lynnshaoyu: done

src/components/Tokens/TokenTable/TokenTable.tsx
lynnshaoyu: i think this actually needs to be named. it complains when I try to do the same spread done in LoadingRow with "Property 'style' does not exist on type 'IntrinsicAttributes &...."

src/components/Tokens/TokenTable/TokenRow.tsx
lynnshaoyu: responsed in TokenTable comment

src/components/Tokens/TokenTable/TokenTable.tsx
lynnshaoyu: this is going to get changed in fix for this: https://uniswaplabs.atlassian.net/browse/WEB-1149?atlOrigin=eyJpIjoiMWE1NjEzM2Y5M2NjNDBmZmJkMmE0YjQzMTE1ZTFlZjUiLCJwIjoiaiJ9 so will leave this for future PR

src/components/Tokens/TokenTable/TokenTable.tsx
JFrankfurt: Yeah you have to let the react component know that is is a button/div etc in the prop typing with something like:

```
interface ButtonProps
    extends React.HTMLProps<HTMLButtonElement> {}
interface WrapperProps extends React.FunctionComponent<ButtonProps> { ... customProps };
```

JFrankfurt: (copied from rebass types on their default button)

lynnshaoyu: ahh gotcha thanks, i think this latest commit should have the requested changes?

PR #4587, opened 2022-09-08
refactor: move initializing analytics to top level

vm:
Initializing amplitude is required for setting up amplitude experiment. We will use the key returned by `init` to set up the experiment module. Make sure that we error out if the key is undefined.

### DIFF
--- a/src/components/AmplitudeAnalytics/index.ts
+++ b/src/components/AmplitudeAnalytics/index.ts
@@ -7,31 +7,28 @@ import { isProductionEnv } from 'utils/env'
  * Uniswap has two Amplitude projects: test and production. You must be a
  * member of the organization on Amplitude to view details.
  */
-export function initializeAnalytics() {
-  const API_KEY = isProductionEnv() ? process.env.REACT_APP_AMPLITUDE_KEY : process.env.REACT_APP_AMPLITUDE_TEST_KEY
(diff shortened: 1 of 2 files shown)


src/components/AmplitudeAnalytics/index.ts
```diff
@@ -7,32 +7,30 @@ import { isProductionEnv } from 'utils/env'
...
-        ipAddress: false,
-        carrier: false,
-        city: false,
-        region: false,
-        dma: false, // designated market area
-      },
-    }
-  )
+if (typeof API_KEY === 'undefined') {
+  throw new Error(`${AMPLITUDE_KEY_NAME} is undefined, Amplitude analytics will not run.`)
```
cmcewen, line 15: you sure we should throw here?

vm: if we don't have amplitude analytics set up then amplitude experiment set up will fail. we want amplitude experiment to be set up even on local now, so i think we do want this. i could avoid by skipping local usage of experiment, but then we lose default flags.

vm: i feel like we need to start giving people a default set of development keys?

cmcewen: i think a console error will let people know it's not working, but in general i think failing gracefully is always best. if there's no API key for analytics, imo things should still work instead of break

vm: gotcha, i will figure out a better way to handle this

vm:
closing while i re-consider how we think about defaults

PR #4590, opened 2022-09-09
fix: remove $ when formatting amounts

kraikov:
There is already a method to format dollar amounts, so this must be a mistake.

### DIFF
--- a/src/utils/formatDollarAmt.tsx
+++ b/src/utils/formatDollarAmt.tsx
@@ -26,7 +26,7 @@ export const formatAmount = (num: number | undefined, digits = 2) => {
   if (num === 0) return '0'
   if (!num) return '-'
   if (num < 0.001) {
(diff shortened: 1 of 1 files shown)


src/utils/formatDollarAmt.tsx
```diff
@@ -26,7 +26,7 @@ export const formatAmount = (num: number | undefined, digits = 2) => {
   if (num === 0) return '0'
   if (!num) return '-'
   if (num < 0.001) {
-    return '$<0.001'
+    return '<0.001'
```
zzmp, line 29: This fn is actually unused. Could you update your PR to just remove it (L24 and on)?

zzmp (changes requested):

vm:
We've updated our formatting helpers and this is now handled 👍

PR #4592, opened 2022-09-09
build: ship node version from 14 to 16.

zhyd1997:
## `yarn start`

```shell
time yarn start
```

node v14:
Screen Shot 2022-09-09 at 23 25 52
node v16:
Screen Shot 2022-09-09 at 23 25 58

saved ~ 1/3 time

## `yarn build`
first build time comparison (no meaningful improvement):

Screen Shot 2022-09-09 at 23 28 27


vm:
unfortunately, we need to do more testing here because node 14 is somewhat necessary for our IPFS deployment atm. it's something we're planning to look at sometime later on because as you have shown... build time is faster!

PR #4596, opened 2022-09-09
feat: optimize AlphaRouter usage

zzmp:
- Cache AlphaRouter (instead of the BaseProvider passed to the router), to prevent router re-instantiation
- Adds a new RouterPreference: PRICE
- Adds the ability to explicitly tune each preference: API, CLIENT, PRICE
  -  A PRICE tuning is defined, but unused - this will be activated after determining the performance of router caching

### DIFF
--- a/src/hooks/useBestTrade.test.ts
+++ b/src/hooks/useBestTrade.test.ts
@@ -1,7 +1,9 @@
 import { renderHook } from '@testing-library/react'
 import { CurrencyAmount, TradeType } from '@uniswap/sdk-core'
 import { DAI, USDC_MAINNET } from 'constants/tokens'
+import { RouterPreference } from 'state/routing/slice'
 import { TradeState } from 'state/routing/types'
+import { useClientSideRouter } from 'state/user/hooks'
 
 import { useRoutingAPITrade } from '../state/routing/useRoutingAPITrade'
 import useAutoRouterSupported from './useAutoRouterSupported'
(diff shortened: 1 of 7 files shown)


src/hooks/useBestTrade.test.ts
```diff
@@ -140,7 +149,12 @@ describe('#useBestV3Trade ExactOut', () => {
 
     const { result } = renderHook(() => useBestTrade(TradeType.EXACT_OUTPUT, DAIAmount, USDC_MAINNET))
 
-    expect(mockUseRoutingAPITrade).toHaveBeenCalledWith(TradeType.EXACT_OUTPUT, undefined, USDC_MAINNET)
+    expect(mockUseRoutingAPITrade).toHaveBeenCalledWith(
```
vm, line 152: @Zach Pomerantz i would recommend switching useRoutingApiTrade to take a params object instead of ordered arguments, then allow defaults again (for example, defaulting to CLIENT)

zzmp: That feels like a bigger change, and switching to a params object risks perf issues with hooks memoization.
The RouterPreference is always available, so I'd rather just make that a new required argument and take it out of useRoutingApiTrade.

src/state/routing/slice.ts
```diff
@@ -1,36 +1,67 @@
...
+const API_QUERY_PARAMS = {
   protocols: protocols.map((p) => p.toLowerCase()).join(','),
-  // example other params
-  // forceCrossProtocol: 'true',
-  // minSplits: '5',
+}
+const CLIENT_QUERY_PARAMS = {
+  protocols,
+}
+const PRICE_QUERY_PARAMS = {
```
vm, line 43: maybe a comment of how this was generated?

src/state/routing/slice.ts
```diff
@@ -1,36 +1,68 @@
...
+// routing API quote params: https://github.com/Uniswap/routing-api/blob/main/lib/handlers/quote/schema/quote-schema.ts
+const API_QUERY_PARAMS = {
+  protocols: 'v2,v3,mixed',
+}
+const CLIENT_PARAMS = {
+  protocols: [Protocol.V2, Protocol.V3, Protocol.MIXED],
+}
+// Price queries are tuned down to minimize the required RPCs to respond to them.
+// TODO(zzmp): This will be used after testing router caching.
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
vm, line 43: is this unused?

vm (approved):

PR #4598, opened 2022-09-09
feat: add animation to search skeleton and typing state

cbachmeier:
- Adds search skeletons to loading state for trending items
- Changes search loading behaviour to fade to 30% opacity before showing new results
- Update styles for search skeleton

Screenshots:
Skeleton and Typing transition
searchFullP0


src/components/NavBar/SearchBar.tsx
```diff
@@ -56,8 +58,10 @@ export const SearchBarDropdownSection = ({
         <Box>{header}</Box>
       </Row>
       <Column gap="12">
-        {suggestions?.map((suggestion, index) =>
-          isCollection(suggestion) ? (
+        {(isLoading ? [...Array(suggestions.length)] : suggestions).map((suggestion, index) =>
```
JackShort, line 61: what if we break this thing into a larger ternary operator based around `isLoading`

src/components/NavBar/SearchBar.tsx
```diff
@@ -131,41 +137,45 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }:
...
-          floorPrice: formatEthPrice(collection.floor?.toString()),
-          stats: {
-            total_supply: collection.totalSupply,
-            one_day_change: collection.floorChange,
-          },
-        }
-      })
-      .slice(0, isNFTPage ? 3 : 2)
+      ? trendingCollectionResults
+          .map((collection) => {
```
JackShort, line 148: can change `{` to `(` and remove `return`

src/components/NavBar/SearchBar.tsx
```diff
@@ -131,41 +137,45 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }:
...
+          })
+          .slice(0, isNFTPage ? 3 : 2)
+      : Array<GenieCollection>(isNFTPage ? 3 : 2)
   }, [isNFTPage, trendingCollectionResults])
 
-  const showTrendingCollections: boolean = useMemo(
-    () => (trendingCollections?.length ?? 0) > 0 && !isTokenPage && phase1Flag === NftVariant.Enabled,
-    [trendingCollections?.length, isTokenPage, phase1Flag]
+  const { data: trendingTokenResults, isLoading: trendingTokensAreLoading } = useQuery(
+    [],
```
JackShort, line 164: this should be `["fetchTrendingTokens"]`

src/components/NavBar/SearchBar.tsx
```diff
@@ -201,61 +211,91 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput }:
     }
   }, [toggleOpen, hoveredIndex, totalSuggestions])
 
+  useEffect(() => {
```
JackShort, line 214: instead of doing this and storing the whole rendered state in a node - is there something wrong with cacheing `tokens` and `collections` in a local variable and updating those in a `useEffect` and you can render the state normally in the `render` method?

cbachmeier: That was my original attempt at a solution however I ran into two issues when trying that. 
1. You also need to be able to fade out on the default search results dropdown, ie search history, trending tokens, and trending collections, so tracking the old state for that data adds a layer of complexity.
2. When I attempted to useRef for tracking token and collection search state, when interacting with useQuery, the final re-render would never occur. So I would enter a search query, default results will fade, but then unfade without showing the new results. Then when next text was entered, the expected results from the previous query would render.

src/components/NavBar/SearchBar.tsx
cbachmeier: went with 'trendingTokens' to match convention

src/components/NavBar/SearchBar.tsx
cbachmeier: I moved the logic to make the blank array map compatible higher in the call hierarchy to help with readability

src/components/NavBar/SearchBar.tsx
JackShort: ah ok figured there was reasoning to it

JackShort (approved):

PR #4603, opened 2022-09-12
feat: NFT Explore Value prop row

gbugyis:
- Adds row below main Banner to display marketing copy marketplace images


src/nft/components/explore/Explore.css.ts
```diff
@@ -106,3 +106,68 @@ export const carouselIndicator = sprinkles({
   borderRadius: 'round',
   display: 'inline-block',
 })
+
+/* Value Prop Styles */
+export const valuePropWrap = style([
+  {},
```
cbachmeier, line 112: can remove empty brakets

src/nft/components/explore/Explore.css.ts
```diff
@@ -106,3 +106,68 @@ export const carouselIndicator = sprinkles({
...
+  sprinkles({
+    width: 'full',
+    borderWidth: '1px',
+    borderStyle: 'solid',
+    borderColor: 'grey700',
```
cbachmeier, line 117: does this border change in light mode vs dark mode?

src/nft/components/explore/Explore.css.ts
```diff
@@ -106,3 +106,68 @@ export const carouselIndicator = sprinkles({
...
+  }),
+])
+
+export const valuePropOverlay = style([
+  {
+    opacity: '0.85',
+    height: '135px',
+    '@media': {
+      [`(min-width: ${breakpoints.lg}px)`]: {
+        opacity: '0.2',
```
cbachmeier, line 132: opacities have now been added to sprinkles

src/nft/components/explore/Explore.css.ts
gbugyis: Same in both, but it is an rgba value rather than a simple grey so I added it to the sprinkles.

src/nft/components/explore/Explore.css.ts
gbugyis: nice - thats very helpful

src/nft/components/explore/Explore.css.ts
```diff
@@ -106,3 +106,60 @@ export const carouselIndicator = sprinkles({
...
+      [`(min-width: ${breakpoints.sm}px)`]: {
+        width: '74%',
+        paddingLeft: '36px',
+        fontSize: '28px',
+        lineHeight: '36px',
+        paddingTop: '32px',
+      },
+      // Extra breakpoint for text wrapping
+      [`(min-width: 800px)`]: { width: '60%' },
+      [`(min-width: ${breakpoints.md}px)`]: { width: '58%' },
```
cbachmeier, line 161: the difference between these two breakpoints is 32px, do we need the additional 800px condition?

src/nft/components/explore/Explore.css.ts
```diff
@@ -106,3 +106,60 @@ export const carouselIndicator = sprinkles({
...
+    lineHeight: '28px',
+    '@media': {
+      [`(max-width: 400px)`]: { width: '88%' },
+      [`(min-width: 400px)`]: { width: '67%' },
+      [`(min-width: ${breakpoints.sm}px)`]: {
+        width: '74%',
+        paddingLeft: '36px',
+        fontSize: '28px',
+        lineHeight: '36px',
+        paddingTop: '32px',
```
cbachmeier, line 157: all(or at least most) of these can be sprinkles

src/nft/components/explore/Explore.css.ts
```diff
@@ -106,3 +106,60 @@ export const carouselIndicator = sprinkles({
...
+      },
+      // Extra breakpoint for text wrapping
+      [`(min-width: 800px)`]: { width: '60%' },
+      [`(min-width: ${breakpoints.md}px)`]: { width: '58%' },
+      [`(min-width: ${breakpoints.lg}px)`]: { width: '50%' },
```
cbachmeier, line 162: sprinkle

src/nft/components/explore/ValueProp.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const ValueProp = () => {
+  return (
+    <Box width="full">
+      <Column as="section" className={styles.section}>
+        <Box
+          className={clsx(styles.bannerWrap, styles.valuePropWrap)}
+          style={{
+            height: '135px',
+            background: `url(${MarketplacesImage}) top 0 right 0 no-repeat`,
```
cbachmeier, line 16: When the banner reduces for smaller viewports, I think we should be left instead of right aligned, this way as the screen shrinks, the more well known marketplaces stay visible

src/nft/components/explore/Explore.css.ts
cbachmeier: looks like you added it to sprinkles without changing the bordercolor. Also the rgba value you added corresponds to the same value as the hex for medGray in dark mode

src/nft/components/explore/Explore.css.ts
gbugyis: how so? We can't specify width as a percentage with sprinkles (besides `half` and `full`), can we?

src/nft/components/explore/Explore.css.ts
gbugyis: It was just for text wrapping, but properly not truly needed. Removed it now.

src/nft/components/explore/Explore.css.ts
gbugyis: I fixed the style to use the new `outline` sprinkle color. The opacity in the rgba value will give a different result on light/dark whereas medGray would be the same in each case.

vm:
Screen Shot 2022-09-14 at 10 25 00 AM

the line height may bit too scrunched?

src/nft/components/explore/Explore.css.ts
cbachmeier: specifically for the 50% width breakpoint we can use `half`

gbugyis:
> the line height may bit too scrunched?

@vm the line-height was getting overridden by a duplicate media query. Fixed in last commit.

src/nft/components/explore/Explore.css.ts
gbugyis: Moving that to a sprinkle isn't working - the width gets overridden by the media query. Gonna leave as is.

PR #4606, opened 2022-09-12
fix: animated box type fix

aballerr:
I am currently porting over some components from interface-nft.  AnimatedBox is having some type errors and this pr fixes those types errors.  I split it out into a separate pr since animated box is used in some other code that others might be porting over.


package.json
```diff
@@ -150,6 +150,7 @@
     "@vanilla-extract/css-utils": "^0.1.2",
     "@vanilla-extract/dynamic": "^2.0.2",
     "@vanilla-extract/sprinkles": "^1.4.1",
+    "@react-spring/web": "~9.4.5",
```
aballerr, line 153: anything above this breaks the build

PR #4607, opened 2022-09-12
fix: bug where changing tabs can open the searchbar

cbachmeier:
There was a bug where after pressing enter on the searchbar, changing tabs, and then changing back, the searchbar would open. This fixes that bug.

Screenshots:
searchBarTabFocus


JackShort (approved):
lgtm

PR #4611, opened 2022-09-12
fix: ignore failed quotes

zzmp:
Ignore failed quotes, which propagate an Error object to the store that cannot be serialized.
This is ok, because routingApi is not persisted - only persisted paths _must_ be serialized.

### DIFF
--- a/src/state/index.ts
+++ b/src/state/index.ts
@@ -1,4 +1,4 @@
-import { configureStore } from '@reduxjs/toolkit'
+import { configureStore, isPlain } from '@reduxjs/toolkit'
 import { setupListeners } from '@reduxjs/toolkit/query/react'
 import multicall from 'lib/state/multicall'
 import { load, save } from 'redux-localstorage-simple'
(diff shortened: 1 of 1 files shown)


src/state/index.ts
```diff
@@ -39,7 +39,16 @@ const store = configureStore({
     [routingApi.reducerPath]: routingApi.reducer,
   },
   middleware: (getDefaultMiddleware) =>
-    getDefaultMiddleware({ thunk: true })
+    getDefaultMiddleware({
+      thunk: true,
+      serializableCheck: {
+        isSerializable: (value: any) => {
```
vm, line 45: it feels odd to have a routing api specific detail at the top level state middleware? are there alternatives to this

zzmp:
I think this was occurring because errors are not meant to be thrown from the queryFn, but to be returned as {error}.

zzmp (commented):
Errors from the queryFn should not be thrown - they should be returned as a typed object (`{ error: FetchBaseQueryError }`). Throwing them causes the caching mechanism to throw, whereas returning the typed object will prevent caching without propagating the error outside of rtkq's caching mechanism.

See https://redux-toolkit.js.org/rtk-query/usage-with-typescript#error-result-example

PR #4612, opened 2022-09-12
fix: fetch the usd price from parsed amount

zzmp:
Fetches the stablecoin price without waiting for the quote API to resolve.


cmcewen (approved):
nice

PR #4614, opened 2022-09-13
feat: adds ethers provider tracing

zzmp:
Refactors all JsonRpcProviders to be instantiated in the same file, and adds tracing under the `traceJsonRpc` flag.
- Moves JsonRpcProvider instantiations to constants/providers.ts
- Guards against other instantiations with a new type-only lint rule
- Adds explicit fallback URLs, and includes them in instantiation
- Adds a flag-guarded console log for all calls to the Provider

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -81,6 +81,18 @@
           }
         ]
       }
+    ],
+    "@typescript-eslint/no-restricted-imports": [
+      "error",
+      {
+        "paths": [
+          {
+            "name": "@ethersproject/providers",
+            "message": "Please only use Providers instantiated in constants/providers to improve traceability.",
+            "allowTypeImports": true
+          }
+        ]
+      }
     ]
   }
 }
--- a/src/components/Web3Provider/index.tsx
+++ b/src/components/Web3Provider/index.tsx
@@ -1,10 +1,13 @@
-import { Web3ReactHooks, Web3ReactProvider } from '@web3-react/core'
+import { SupportedChainId } from '@uniswap/widgets'
+import { useWeb3React, Web3ReactHooks, Web3ReactProvider } from '@web3-react/core'
 import { Connector } from '@web3-react/types'
 import { Connection } from 'connection'
 import { getConnectionName } from 'connection/utils'
+import { RPC_PROVIDERS } from 'constants/providers'
+import { TraceJsonRpcVariant, useTraceJsonRpcFlag } from 'featureFlags/flags/traceJsonRpc'
 import useEagerlyConnect from 'hooks/useEagerlyConnect'
 import useOrderedConnections from 'hooks/useOrderedConnections'
-import { ReactNode, useMemo } from 'react'
+import { ReactNode, useEffect, useMemo } from 'react'
 
 export default function Web3Provider({ children }: { children: ReactNode }) {
   useEagerlyConnect()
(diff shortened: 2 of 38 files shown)


.eslintrc.json
```diff
@@ -81,6 +81,18 @@
           }
         ]
       }
+    ],
+    "@typescript-eslint/no-restricted-imports": [
```
vm, line 85: yessssss

src/components/Web3Provider/index.tsx
```diff
@@ -15,7 +18,37 @@ export default function Web3Provider({ children }: { children: ReactNode }) {
 
   return (
     <Web3ReactProvider connectors={connectors} key={key}>
+      <Tracer />
       {children}
     </Web3ReactProvider>
   )
 }
+
+function Tracer() {
```
vm, line 27: can't we just make this a hook instead a full component?

src/utils/switchChain.ts
```diff
@@ -2,36 +2,22 @@ import { Connector } from '@web3-react/types'
...
-      return ['https://polygon-rpc.com/']
-    case SupportedChainId.POLYGON_MUMBAI:
-      return ['https://rpc-endpoints.superfluid.dev/mumbai']
-    case SupportedChainId.CELO:
-      return ['https://forno.celo.org']
-    case SupportedChainId.CELO_ALFAJORES:
-      return ['https://alfajores-forno.celo-testnet.org']
+      return RPC_URLS[chainId][0]
+    // Attempting to add a chain using an infura URL will not work, as the URL will be unreachable from the MetaMask background page.
+    // MetaMask allows switching to any publicly reachable URL, but for novel chains, it will display a warning if it is not on the "Safe" list.
```
vm, line 16: I would explain this comment within the fallback urls definition instead of here maybe?

src/components/Web3Provider/index.tsx
zzmp: We can't access the Web3ReactProvider's context without being a nested JSXElement.
Hooks won't have access to `useWeb3React().provider`, which is necessary to get a (potentially third-party) active provider.

src/utils/switchChain.ts
zzmp: There's a longer explanation (with referenced source code) in the FALLBACK_URLS definition - this is here too so it's apparent why RPC_URLS is invalid in these cases.

src/components/Web3Provider/index.tsx
vm: ah yea you're right

src/components/Web3Provider/index.tsx
```diff
@@ -15,7 +18,37 @@ export default function Web3Provider({ children }: { children: ReactNode }) {
...
+  const { chainId, provider } = useWeb3React()
+  const networkProvider = RPC_PROVIDERS[(chainId || SupportedChainId.MAINNET) as SupportedChainId]
```
vm, line 29: think better not to default the chain here? may lead to weird sideeffects when switching chains unless i'm misunderstanding.

vm (approved):

src/components/Web3Provider/index.tsx
vm: but in that case wouldn't the chainId be passed in rather than using the chainId from web3-react? ex: i'm on Polygon, but use the Mainnet infura provider

vm: i'm pretty sure we want to just debug all the network providers instead of just the chainId we're on

PR #4615, opened 2022-09-13
feat: caching provider

zzmp:
- Adds a block-level caching layer to the JsonRpcProvder subclass


src/constants/providers.ts
```diff
@@ -1,14 +1,49 @@
...
 import { SupportedChainId } from './chains'
 import { RPC_URLS } from './networks'
 
 class AppJsonRpcProvider extends StaticJsonRpcProvider {
+  private _blockCache = new Map<string, Promise<any>>()
+  get blockCache() {
+    // If the blockCache has not yet been initialized this block, do so by
+    // setting a listener to clear it on the next block.
+    if (!this._blockCache.size) {
+      this.once('block', () => this._blockCache.clear())
```
cmcewen, line 16: is the provider already subscribed per block? will this add a new subscription or use the existing one?

src/constants/providers.ts
```diff
@@ -1,14 +1,49 @@
...
 // This is the only file which should instantiate new Providers.
 // eslint-disable-next-line @typescript-eslint/no-restricted-imports
 import { StaticJsonRpcProvider } from '@ethersproject/providers'
+import { isPlain } from '@reduxjs/toolkit'
 
 import { SupportedChainId } from './chains'
 import { RPC_URLS } from './networks'
 
 class AppJsonRpcProvider extends StaticJsonRpcProvider {
+  private _blockCache = new Map<string, Promise<any>>()
```
cmcewen, line 11: do you foresee any perf issues here with the size of the payload or anything?

src/constants/providers.ts
```diff
@@ -1,14 +1,49 @@
...
+    // Only cache eth_call's.
+    if (method !== 'eth_call') return super.send(method, params)
+
+    // Only cache if params are serializable.
+    if (!isPlain(params)) return super.send(method, params)
+
+    const key = `call:${JSON.stringify(params)}`
+    const cached = this.blockCache.get(key)
+    if (cached) {
+      this.emit('debug', {
```
cmcewen, line 35: this is just copied from the send method i'm guessing?

cmcewen (commented):
this tentatively looks good to me but just want to noodle on it a bit to make sure i'm not missing anything

src/constants/providers.ts
zzmp: Based on my read of the source code, the provider is only subscribed if there is a call (through `blockNumber`, `getBlockNumber()`, `on('block', ...)`, or `once('block', ...)`); so this is necessary to be notified when there is a new block, _and_ this will add a new subscription iff one does not already exist.

Caveat: I only skimmed the code - it's possible it will create an add'l subscription, but that doesn't seem bad for perf.

src/constants/providers.ts
zzmp: I do not, because this is a per-block cache.
If this persisted, I would, but any time it is used a listener is set up to clear the cache at the next block.

src/constants/providers.ts
zzmp: With the exception of `id: 'cache'`, which I added to make it explicit when `traceJsonRpc` is enabled that a request is served from cache.

src/constants/providers.ts
zzmp: I hope this is self-documented, by naming it `_blockCache`. There is a comment in the `blockCache` getter, but I don't think add'l doc comments are necessary.

cmcewen (approved):
i'm okay to merge this but i do think we need to continue investing in measurement/tooling to so we can feel a little better about the real world impact here

PR #4616, opened 2022-09-13
style: added transition ease to explore

cartcrom:
Created an ease constant variable called TRANSITION_TIME, used by ClickableStyle and other components.


JFrankfurt:
should prob use `transition-duration` for these time definitions https://developer.mozilla.org/en-US/docs/Web/CSS/transition-duration

vm (approved):
lgtm. i think `transition` is just shorthand for the diff transition attrs so can use either?

cmcewen (approved):
love it

PR #4618, opened 2022-09-13
feat: migrate select page assets

cbachmeier:
- Show a connected wallets nfts
- Allows a user to select nfts
- Can query tokenIds for assets
- Added sparkle animation when loading more wallet nfts on scroll


aballerr (approved):

src/nft/components/common/Loading/LoadingSparkle.css.ts
```diff
@@ -0,0 +1,34 @@
...
+      animation: `0.5s infinite ${pathAnim}`,
+      ...pathAnimCommonProps,
+    },
+    '&:nth-child(2)': {
+      animation: `0.5s infinite ${pathAnim}`,
+      animationDelay: '0.1s',
+      ...pathAnimCommonProps,
+    },
+    '&:nth-child(3)': {
+      animation: `0.5s infinite ${pathAnim}`,
```
vm, line 29: this could go in the common props i think

src/nft/components/sell/select/SelectPage.css.ts
```diff
@@ -0,0 +1,66 @@
...
+export const section = style([
+  sprinkles({
+    paddingLeft: { sm: '16', lg: '0' },
+    paddingRight: { sm: '16', lg: '0' },
+  }),
+  { maxWidth: '1000px', margin: '0 auto' },
+])
+
+export const sellAssetSelected = style({
+  border: `4px solid ${vars.color.genieBlue} !important`,
```
vm, line 14: seems best to avoid !Important, we may have a lint rule but think it's not enabled for sprinkles. are there alternatives in this case?

src/nft/components/sell/select/SelectPage.tsx
```diff
@@ -1,3 +1,380 @@
...
+import { fetchWalletAssets } from 'nft/queries'
+import { fetchMultipleCollectionStats, OSCollectionsFetcher } from 'nft/queries'
+import { SellPageStateType, WalletAsset } from 'nft/types'
+import { Dispatch, FormEvent, SetStateAction, useEffect, useMemo, useReducer, useState } from 'react'
+import InfiniteScroll from 'react-infinite-scroll-component'
+import { useInfiniteQuery, useQuery } from 'react-query'
+
+import * as styles from './SelectPage.css'
+
+const formatEth = (price: number) => {
```
vm, line 18: nit: i believe we have a similar function here: `utils/formatDollarAmount`

src/nft/components/sell/select/SelectPage.tsx
```diff
@@ -1,3 +1,380 @@
...
+    setWalletAssets(ownerAssets?.flat() ?? [])
+  }, [ownerAssets, setWalletAssets])
+
+  useEffect(() => {
+    ownerCollections && setWalletCollections(ownerCollections)
+  }, [ownerCollections, setWalletCollections])
+
+  const listFilter = useWalletCollections((state) => state.listFilter)
+
+  useEffect(() => {
```
vm, line 94: slightly confused about the need for two separate hooks here. this and L103

src/nft/components/sell/select/SelectPage.tsx
```diff
@@ -1,3 +1,380 @@
...
+    }
+  }, [searchText, walletAssets, listFilter, setDisplayAssets])
+
+  useEffect(() => {
+    setDisplayAssets(walletAssets, listFilter)
+  }, [walletAssets, listFilter, setDisplayAssets])
+
+  useEffect(() => {
+    if (ownerCollections?.length && collectionStats?.length) {
+      const ownerCollectionsCopy = [...ownerCollections]
```
vm, line 109: i don't think this will truly copy the collections as it'll still be a reference. if you want to create new things i'd suggest this:

```
ownerCollectionsCopy = ownerCollection.map(collection => {
  return {
    ...collection,
    floorPrice: 1000,
  }
})
```

src/nft/components/sell/select/SelectPage.tsx
```diff
@@ -1,3 +1,380 @@
...
+      for (const collection of ownerCollectionsCopy) {
+        const floorPrice = collectionStats.find((stat) => stat.address === collection.address)?.floorPrice
+        collection.floorPrice = floorPrice ? Math.round(floorPrice * 1000 + Number.EPSILON) / 1000 : 0 //round to at most 3 digits
```
vm, line 112: nit: would suggest making a helpful function for this so the comment isn't necessary

src/nft/components/sell/select/SelectPage.tsx
```diff
@@ -1,3 +1,380 @@
...
+      }
+      setWalletCollections(ownerCollectionsCopy)
+    }
+  }, [collectionStats, ownerCollections, setWalletCollections])
+
+  return (
+    // Column style is temporary while we move over the filters bar that adjust width
+    <Column style={{ width: 'calc(100vw - 32px)' }}>
```
vm, line 120: where does the 32px come from? is that based on a diff page element?

src/nft/components/sell/select/SelectPage.tsx
```diff
@@ -1,3 +1,380 @@
...
+        <Row
+          display={{ sm: 'flex', md: 'none' }}
+          position="fixed"
+          bottom="60"
+          left="16"
+          height="56"
+          borderRadius="12"
+          paddingX="16"
+          paddingY="12"
+          style={{ background: '#0d0e0ef2', width: 'calc(100% - 32px)', lineHeight: '24px' }}
```
vm, line 164: should color be added to the styles file? also notice the same `-32px` logic here so wanna know if that's related to above

src/nft/components/sell/select/SelectPage.tsx
```diff
@@ -1,3 +1,380 @@
...
+          className={subhead}
+        >
+          {sellAssets.length}&nbsp; selected item{sellAssets.length === 1 ? '' : 's'}
```
vm, line 167: nit: we have a `Plural` in lingui/macro for stuff like this

src/nft/components/sell/select/SelectPage.tsx
```diff
@@ -1,3 +1,380 @@
...
+      !sellAssets.find(
+        (x) => x.tokenId === asset.tokenId && x.asset_contract.address === asset.asset_contract.address
+      ) &&
+      !isMobile
+    )
+      toggleCart()
+  }
+
+  return (
+    <Column
```
vm, line 231: should this be a `react-router` Link?

src/nft/components/sell/select/SelectPage.tsx
```diff
@@ -1,3 +1,380 @@
...
+        paddingLeft={{ sm: '0', md: '52' }}
+        paddingRight={{ sm: '0', md: '72' }}
+        paddingTop={{ sm: '16', md: '40' }}
+      >
+        <AnimatedBox paddingX="16" flexShrink="0" width="full">
+          <Row gap="8" flexWrap="nowrap">
+            <CollectionSearch searchText={searchText} setSearchText={setSearchText} />
+            <SelectAllButton />
+          </Row>
+          <InfiniteScroll
```
vm, line 133: nit: could be clearer to follow by make a separate component called `AssetList`. same for `SellableAssetList` below.

imo best to keep Page components very high level around page structure and keep components isolated

src/nft/components/sell/select/SelectPage.tsx
cbachmeier: This color is a one-off only used in this instance

src/nft/components/sell/select/SelectPage.tsx
cbachmeier: As per the comment above this is a temporary style as without the filters menu the sizing of the assets was off. The filters menu will be added in a subsequent PR to help keep the size of PRs down.

src/nft/components/sell/select/SelectPage.tsx
cbachmeier: This does make a deep copy. Evidenced here: https://stackoverflow.com/questions/7486085/copy-array-by-value

> The [ ...myArray ], myArray.splice(0), myArray.slice(), and myArray.concat() techniques can be used to deep copy arrays with literal values

This also handles the functionality for which a copy is needed as evidenced on interface-nft. All of the functionality for this code has already been tested in interface-nft

vm: oh wild, didn't know that

vm: Screen Shot 2022-09-14 at 12 09 34 PM

vm: Screen Shot 2022-09-14 at 12 10 52 PM

vm: i think it works regardless so just a nit but wanted to point it out

src/nft/components/sell/select/SelectPage.tsx
cbachmeier: Good catch, the hook at 103 is extraneous and can be removed

src/nft/components/sell/select/SelectPage.tsx
```diff
@@ -1,3 +1,380 @@
...
+        ownerAddress: address ?? '',
+        collectionAddresses: collectionFilters,
+        pageParam,
+      })
+    },
+    {
+      getNextPageParam: (lastPage, pages) => {
+        return lastPage?.flat().length === 25 ? pages.length : null
+      },
+      refetchOnWindowFocus: false,
```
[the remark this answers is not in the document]
cbachmeier: Both of these params are default true https://tanstack.com/query/v4/docs/reference/useQuery?from=reactQueryV3&original=https://react-query-v3.tanstack.com/reference/useQuery

src/nft/components/sell/select/SelectPage.tsx
```diff
@@ -1,3 +1,380 @@
...
+  const {
+    data: ownerAssetsData,
+    fetchNextPage,
+    hasNextPage,
+    isSuccess,
+  } = useInfiniteQuery(
+    ['ownerAssets', address, collectionFilters],
+    async ({ pageParam = 0 }) => {
+      return await fetchWalletAssets({
+        ownerAddress: address ?? '',
```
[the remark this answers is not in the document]
cbachmeier: This still passes and the fetch returns an empty array. This catch was put in place for eslint compilation as if there is no ownerAddress that means there is an upstream issue with our wallet connection and the 'Connect' button will be shown from sell.tsx

src/nft/components/sell/select/SelectPage.tsx
cbachmeier: None of the functions in formatDollarAmount handle the 1000s -> K formatting, editing those functions to support this could cause other unintended behaviours

src/nft/components/sell/select/SelectPage.tsx
cbachmeier: converted

src/nft/components/sell/select/SelectPage.tsx
cbachmeier: As a general design principle we kept related components colocated on a page provided the file didn't get overly bloated. If after our series of future architecture decisions we decided against this approach for all files, we can revisit afterwards.

src/nft/components/sell/select/SelectPage.tsx
cbachmeier: Yes the width logic is also related to a temporary fix to make the content more viewable while waiting for the filters menu to be added to the page

src/nft/components/sell/select/SelectPage.tsx
JackShort: sorry about my previous comment (deleted it) - read vig's code as charlie's code lmao. Vig is right by doing it `[...obj]` you are copying the object by value but the actual properties are passed by call-by-reference in JS (its dumb). Its weird so the right way to take a copy of an array is vig's method

src/nft/components/sell/select/SelectPage.tsx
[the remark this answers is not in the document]
vm: oh wow that's crazy...

vm (approved):

PR #4623, opened 2022-09-14
fix: Tick collisions on token detail price chart

gbugyis:
On the Token Details price chart, when selecting all the x-axis tick intervals were too short causing overlaps in the tick labels: 

Before: 
before

After: 
after

### DIFF
--- a/src/components/Tokens/TokenDetails/PriceChart.tsx
+++ b/src/components/Tokens/TokenDetails/PriceChart.tsx
@@ -200,7 +200,7 @@ export function PriceChart({ width, height, tokenAddress, priceData }: PriceChar
         return [
           monthYearFormatter(locale),
           monthYearDayFormatter(locale),
-          timeMonth.range(startDate, endDate, 3).map((x) => x.valueOf() / 1000),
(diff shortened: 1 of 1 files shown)


src/components/Tokens/TokenDetails/PriceChart.tsx
```diff
@@ -200,7 +200,7 @@ export function PriceChart({ width, height, tokenAddress, priceData }: PriceChar
         return [
           monthYearFormatter(locale),
           monthYearDayFormatter(locale),
-          timeMonth.range(startDate, endDate, 3).map((x) => x.valueOf() / 1000),
+          timeMonth.range(startDate, endDate, 6).map((x) => x.valueOf() / 1000),
```
vm, line 203: seems fine, though i'm wondering if it's possible to make this dynamic based on width?

vm (approved):

src/components/Tokens/TokenDetails/PriceChart.tsx
gbugyis: Something like that is coming in the next PR for more even spacing of the ticks. This one seemed more urgent though since it's busted and unreadable at the 3mo interval.

PR #4626, opened 2022-09-14
chore: merging loading states

aballerr:
in progress


src/nft/components/collection/CollectionAssetLoading.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { Box } from '../../components/Box'
+import { Row } from '../Flex'
+import * as styles from './CollectionAssetLoading.css'
+
+export const CollectionAssetLoading = () => {
+  return (
+    <Box as="div" className={styles.collectionAssetLoading}>
+      <Box as="div" position="relative" width="full">
+        <Box as="div" className={styles.collectionAssetsImageLoading} />
+        <Box as={'img'} width="full" opacity="0" src={SizingImage} draggable={false} />
```
cbachmeier, line 13: id the loading image the only way to achieve this effect? Could we set a fixed height and width and set flexShrink to 0? I think we should avoid hardcoding in a random nft to use for sizing

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -233,39 +254,25 @@ export const CollectionNfts = ({ contractAddress, collectionStats, rarityVerifie
       <InfiniteScroll
         next={fetchNextPage}
         hasMore={hasNextPage ?? false}
-        loader={hasNextPage ? <p>Loading from scroll...</p> : null}
-        dataLength={collectionNfts?.length ?? 0}
+        loader={hasNextPage ? loadingAssets : null}
+        dataLength={collectionNfts ? collectionNfts.length : 0}
```
cbachmeier, line 258: any reason for this syntax over the previous?

src/nft/css/sprinkles.css.ts
```diff
@@ -14,6 +14,7 @@ const themeContractValues = {
     backgroundModule: '',
     backgroundOutline: '',
     backgroundSurface: '',
+    lightGray: '',
```
cbachmeier, line 17: remove lightGray and replace where it is used with `backgroundSurface`

cbachmeier (changes requested):

src/nft/components/collection/CollectionAssetLoading.tsx
aballerr: The height and width change proportionally to screen size.  Fixed size will not look as good in my opinion.  The alternative to this is making the asset cards height not dependent on the image, which I would prefer not to do.

cbachmeier: I sent a blank 250x250 png to use, also the brackets around `img` can be removed

cbachmeier (approved):
LGTM

PR #4627, opened 2022-09-14
feat: amplitude swap fixes

lynnshaoyu:
1) change "Swap Submitted" to "Swap Submit Button Clicked"
2) Add 'Swap Signed' event when transaction hash is produced after swap approve signing. 
Screen Shot 2022-09-14 at 3 40 23 PM

Notion doc w/ details: https://www.notion.so/uniswaplabs/Amplitude-Router-Discrepancy-60dfd67b736c44ec85c4d5021d038466


src/state/transactions/updater.tsx
```diff
@@ -78,19 +40,7 @@ export default function Updater() {
           },
         })
       )
-      const tx = transactions[chainId]?.[hash]
 
-      if (tx.info.type === TransactionType.SWAP && trade) {
-        sendAnalyticsEvent(
```
cmcewen, line 84: why remove this? it does seem useful to know whether the browser was open right?

lynnshaoyu: i think we agreed to remove it: https://uniswapteam.slack.com/archives/C03JAKVATU6/p1663003407549829?thread_ts=1662855832.386199&cid=C03JAKVATU6

lynnshaoyu: keeping it creates confusion because this gives an incomplete picture

lynnshaoyu: added back as per slack discussion

cmcewen (approved):

PR #4634, opened 2022-09-15
feat: Web 924 explore use real data for the sparkline price charts

lynnshaoyu:
Screen Shot 2022-09-15 at 1 35 22 PM


src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -543,7 +561,20 @@ export default function LoadedRow({
...
+                  getY={(p: PricePoint) => rdScale(p.value)}
+                  curve={curveCardinal.tension(curveTension)}
+                  color={pricePercentChange && pricePercentChange < 0 ? theme.accentFailure : theme.accentSuccess}
+                  strokeWidth={2}
+                  width={width}
+                  height={height}
+                />
+              )}
+            </ParentSize>
           </SparkLine>
```
cartcrom, line 578: Can we wrap all of this in a separate imported component instead, including all the vars from L487-497

src/graphql/data/Token.ts
```diff
@@ -107,7 +107,7 @@ const topTokensQuery = graphql`
     }
   }
 `
-const tokenPricesFragment = graphql`
+export const tokenPricesFragment = graphql`
```
cartcrom, line 110: Does this still need to be exported

src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -543,7 +561,20 @@ export default function LoadedRow({
...
-            <ParentSize>{({ width, height }) => <SparklineChart width={width} height={height} />}</ParentSize>
+            <ParentSize>
+              {({ width, height }) => (
+                <LineChart
+                  data={pricePoints}
+                  getX={(p: PricePoint) => widthScale(p.timestamp)}
+                  getY={(p: PricePoint) => rdScale(p.value)}
+                  curve={curveCardinal.tension(curveTension)}
+                  color={pricePercentChange && pricePercentChange < 0 ? theme.accentFailure : theme.accentSuccess}
+                  strokeWidth={2}
```
cartcrom, line 572: Can we try 1.5px with here? I think that's how figma is

cartcrom (changes requested):

src/graphql/data/Token.ts
lynnshaoyu: nope ur right :P thx for catch

src/components/Tokens/TokenTable/TokenRow.tsx
lynnshaoyu: done ty

src/components/Tokens/TokenTable/TokenRow.tsx
lynnshaoyu: ooh yes moved everything to sparkline charts file

cartcrom (approved):

PR #4643, opened 2022-09-16
feat: Add Select Page Modal

cbachmeier:
- Adds shopping tag modal to Select Page of listing flow
- Width will not be dynamically adjusted until Filters PR is merged in
- Doesn't include content used for Collection Shopping bag or content for the Listing Page
- Shopping bag modal added to the TopLevelModals wrapper


src/nft/components/cart/Cart.css.ts
```diff
@@ -0,0 +1,43 @@
...
+])
+
+export const closeIcon = style({
+  transform: 'rotate(90deg)',
+  float: 'right',
+  paddingTop: '1px',
+})
+
+export const cartFooter = style({
+  borderTop: `1px solid ${themeVars.colors.lightGray}`,
```
JackShort, line 42: some of these styles can be in sprinkles

src/nft/components/cart/Cart.tsx
```diff
@@ -0,0 +1,130 @@
...
+          marginTop={{ sm: '0', md: '4' }}
+          marginRight={{ sm: '0', md: '20' }}
+          className={styles.cartContainer}
+          width={{ sm: 'full', md: '288' }}
+          height={{ sm: 'full', md: 'auto' }}
+          backgroundColor="white"
+          marginLeft="0"
+          justifyContent="flex-start"
+        >
+          {isSell && sellPageState === SellPageStateType.LISTING ? null : (
```
JackShort, line 49: we can remove all of these sell checks because the bag doesn't use them

src/nft/components/cart/Cart.tsx
```diff
@@ -0,0 +1,130 @@
...
+          <CloseDropDownIcon />
+        </Box>
+      ) : null}
+      <Box className={headlineSmall} paddingTop="0" paddingBottom="8" paddingX="0" margin="0">
+        {isSell ? 'Selected items' : 'My Bag'}
+      </Box>
+      {bagQuantity > 0 ? (
+        <Box className={bodySmall} paddingTop="0" paddingBottom="8" paddingX="0" marginY="0" marginX="auto">
+          {bagQuantity} {bagQuantity === 1 ? 'NFT' : 'NFTs'}
+          <span style={{ fontSize: '8px', position: 'relative', top: '-2px', padding: '0 2px 0 4px' }}>&#x2022;</span>
```
JackShort, line 114: this can be changed to `<Box as="span" position="relative" paddingTop="2" paddingLeft="2" paddingRight="4" style={{ fontSize: '8px' }}>•</Box>`

src/nft/components/cart/CartAssetRow.css.ts
```diff
@@ -0,0 +1,152 @@
...
+export const cartAssetName = style([
+  sprinkles({
+    fontWeight: 'medium',
+    overflow: 'hidden',
+    marginRight: 'auto',
+  }),
+  {
+    marginTop: '4px',
+    textOverflow: 'ellipsis',
+    whiteSpace: 'nowrap',
```
JackShort, line 27: a lot of styles in this file are outdated and can be put into sprinkles

src/nft/components/cart/CartAssetRow.css.ts
```diff
@@ -0,0 +1,152 @@
...
+  width: '12px',
+  marginLeft: '2px',
+  boxSizing: 'border-box',
+})
+
+export const removeIcon = style({
+  width: '32px',
+})
+
+export const ethIcon = style({
```
JackShort, line 78: don't think we need any of the styles below this since the old bag is getting deprecated

src/nft/components/cart/CartBalance.css.ts
```diff
@@ -0,0 +1,14 @@
+import { style } from '@vanilla-extract/css'
+import { sprinkles } from 'nft/css/sprinkles.css'
+
+export const orderButton = style([
+  {
+    padding: '12px 0',
```
JackShort, line 6: both of these can be sprinkles

JackShort (approved):
Looks good works well. Nothing blocking just some sprinkles stuff

src/nft/components/cart/Cart.css.ts
cbachmeier: this is borderTop so only the color can be moved

cbachmeier: this style is actually not used by the sell tag and has been removed

src/nft/components/cart/Cart.tsx
cbachmeier: We still need the -2 position and your left and right are swapped so it became: `<Box as="span" position="relative" paddingRight="2" paddingLeft="4" style={{ fontSize: '8px', top: '-2px' }}>`

JackShort: Ah yes you're right - css is dumb I feel like left -> right is way more intuitive than right -> left

cbachmeier: I always remember it as clockwise staring with 0/12 position ie top then going right (3), bottom (6), left (9)

PR #4646, opened 2022-09-16
fix: show copied text where address was clicked

vm:
Instead of defaulting to the center of the parent, we show the text where the button was clicked.

### DIFF
--- a/src/theme/components.tsx
+++ b/src/theme/components.tsx
@@ -2,7 +2,7 @@ import { Trans } from '@lingui/macro'
 import { outboundLink } from 'components/analytics'
 import { MOBILE_MEDIA_BREAKPOINT } from 'components/Tokens/constants'
 import useCopyClipboard from 'hooks/useCopyClipboard'
-import React, { forwardRef, HTMLProps, ReactNode, useCallback, useImperativeHandle } from 'react'
+import React, { forwardRef, HTMLProps, ReactNode, useCallback, useImperativeHandle, useState } from 'react'
 import {
   ArrowLeft,
   CheckCircle,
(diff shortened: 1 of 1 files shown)


src/theme/components.tsx
```diff
@@ -209,12 +209,15 @@ export function ExternalLinkIcon({
   )
 }
 
-const ToolTipWrapper = styled.div<{ isCopyContractTooltip?: boolean }>`
+const TOOLTIP_WIDTH = 60
+
+const ToolTipWrapper = styled.div<{ isCopyContractTooltip?: boolean; tooltipX?: number }>`
```
cmcewen, line 214: is this the x coordinate on the page? maybe tooltipXCoordinate or something?

src/theme/components.tsx
```diff
@@ -209,12 +209,15 @@ export function ExternalLinkIcon({
...
   display: flex;
   flex-direction: column;
   align-items: center;
-  position: ${({ isCopyContractTooltip }) => (isCopyContractTooltip ? 'relative' : 'absolute')};
-  right: ${({ isCopyContractTooltip }) => isCopyContractTooltip && '50%'};
+  position: absolute;
+  left: ${({ isCopyContractTooltip, tooltipX }) =>
+    isCopyContractTooltip && (tooltipX ? `${tooltipX - TOOLTIP_WIDTH / 2}px` : '50%')};
```
cmcewen, line 220: this seems a lil funky to me. is there any other way to do this?

src/theme/components.tsx
vm: its the local x coordinate within it's box

src/theme/components.tsx
vm: there may be a way to do it on the cursor itself i can look into it

vm: i looked into it a while and don't see alternative options. there is no way to position relatively to cursor. i could instead do the calculation outside of the wrapper if we want?

zzmp (approved):

PR #4651, opened 2022-09-16
fix: load correct initial display price and % delta w/o need for scrubbing on token detail chart

lynnshaoyu:
before: display price and % delta only loads after scrubbing

after: display price and % delta correctly set to ending price upon first load of token detail page. 

https://uniswaplabs.atlassian.net/browse/WEB-1113?atlOrigin=eyJpIjoiMDg0NzNlNjI3NjM5NDcxYThhYjA5NWFlNzJlNzBmZTEiLCJwIjoiaiJ9

### DIFF
--- a/src/components/Tokens/TokenDetails/PriceChart.tsx
+++ b/src/components/Tokens/TokenDetails/PriceChart.tsx
@@ -10,7 +10,7 @@ import { useTokenPricesCached } from 'graphql/data/Token'
 import { PricePoint, TimePeriod } from 'graphql/data/Token'
 import { useActiveLocale } from 'hooks/useActiveLocale'
 import { useAtom } from 'jotai'
-import { useCallback, useMemo, useState } from 'react'
+import { useCallback, useEffect, useMemo, useState } from 'react'
 import { ArrowDownRight, ArrowUpRight } from 'react-feather'
 import styled, { useTheme } from 'styled-components/macro'
 import {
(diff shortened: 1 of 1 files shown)


src/components/Tokens/TokenDetails/PriceChart.tsx
```diff
@@ -127,15 +127,26 @@ interface PriceChartProps {
...
   const endingPrice = prices?.[prices.length - 1] ?? DATA_EMPTY
   const [displayPrice, setDisplayPrice] = useState(startingPrice)
+
+  // set display price to ending price once first set of prices is loaded.
+  useEffect(() => {
+    if (prices && !firstLoadedPriceDisplayed) {
+      setDisplayPrice(endingPrice)
+      setFirstLoadedPriceDisplayed(true)
+    }
+  }, [prices, firstLoadedPriceDisplayed, endingPrice])
```
cartcrom, line 149: instead of creating firstLoadedPriceDisplayed, could we maybe just useEffect where every time prices is updated, we check if the display price isn't initialized (i.e. seeing if its equal to DATA_EMPTY), and if so, update it to be endingprice

lynnshaoyu: done, except not checking if its equal to DATA_EMPTY, want to update endingPrice when a new one is available due to changed prices.

cartcrom (approved):

PR #4655, opened 2022-09-16
refactor: move components/AmplitudeAnalytics to analytics

vm:
analytics should be seen more as a module than as a component within our folder structure, so I moved it out. eventually we could potentially make this a package that's shared across repos.


src/analytics/amplitude/index.ts
```diff
@@ -1,15 +1,15 @@
 import { Identify, identify, init, track } from '@amplitude/analytics-browser'
```
vm, line 1: this will get rebased after https://github.com/Uniswap/interface/pull/4653 is merged

cmcewen:
while you're here - imo having the name of the service in the path is an implementation detail, and really it should just be analytics

vm:
i can rename

cmcewen (approved):

PR #4657, opened 2022-09-16
refactor: switch explore over to new queries

cartcrom:


.vscode/settings.json
```diff
@@ -15,5 +15,16 @@
...
+    {
+      "filePath": "src/components/Tokens/TokenDetails",
+      "group": "Default"
+    },
+    {
+      "filePath": "src/graphql/data",
+      "group": "Default"
+    }
+  ],
+  "favorites.sortOrder": "ASC"
```
JFrankfurt, line 29: was this intended to be a local change, not a workspace setting change?

src/components/Tokens/state.ts
```diff
@@ -1,18 +1,17 @@
...
-
 export const favoritesAtom = atomWithStorage<string[]>('favorites', [])
 export const showFavoritesAtom = atomWithStorage<boolean>('showFavorites', false)
 export const filterStringAtom = atomWithReset<string>('')
 export const filterNetworkAtom = atom<SupportedChainId>(SupportedChainId.MAINNET)
 export const filterTimeAtom = atom<TimePeriod>(TimePeriod.DAY)
-export const sortCategoryAtom = atom<Category>(Category.marketCap)
-export const sortDirectionAtom = atom<SortDirection>(SortDirection.decreasing)
+export const sortMethodAtom = atom<TokenSortMethod>(TokenSortMethod.TOTAL_VALUE_LOCKED)
+export const sortAscendingAtom = atom<boolean>(false)
```
JFrankfurt, line 14: why was this switching to `sortAscendingAtom`? 

If it's removed, I think we should probably drop the prior type used here from the codebase if we don't need it anymore. (`SortDirection`)

src/graphql/data/TopTokens.ts
```diff
@@ -0,0 +1,222 @@
...
+    let returnTokens = tokens
+    if (showFavorites) {
+      returnTokens = returnTokens?.filter((token) => (token?.address && favorites.includes(token.address)) ?? false)
+    }
+    if (lowercaseFilterString) {
+      returnTokens = returnTokens?.filter((token) => {
+        const addressIncludesFilterString = token?.address?.toLowerCase().includes(lowercaseFilterString)
+        const nameIncludesFilterString = token?.name?.toLowerCase().includes(lowercaseFilterString)
+        const symbolIncludesFilterString = token?.symbol?.toLowerCase().includes(lowercaseFilterString)
+        return (nameIncludesFilterString || symbolIncludesFilterString || addressIncludesFilterString) ?? false
```
JFrankfurt, line 117: I don't think you need the `?? false` here

src/graphql/data/TopTokens.ts
```diff
@@ -0,0 +1,222 @@
...
+      })
+    }
+    return returnTokens
+  }, [tokens, showFavorites, lowercaseFilterString, favorites])
+}
+
+const PAGE_SIZE = 20
+//const MAX_COUNT = 100
```
JFrankfurt, line 125: remove

src/graphql/data/TopTokens.ts
```diff
@@ -0,0 +1,222 @@
...
+function toContractInput(token: PrefetchedTopToken) {
+  return {
+    address: token?.address ?? '',
+    chain: token?.chain ?? 'ETHEREUM',
+  }
+}
+
+export type TopToken = NonNullable<TopTokens_TokensQuery['response']['tokens']>[number]
+export function useTopTokens(prefetchedTokens: TopTokens100Query['response']) {
+  // TODO: add filtering by favorites and filter string
```
JFrankfurt, line 136: this is done, right? I think you can drop the comment.

src/graphql/data/TopTokens.ts
```diff
@@ -0,0 +1,222 @@
...
+  const duration = toHistoryDuration(useAtomValue(filterTimeAtom))
+  const environment = useRelayEnvironment()
+  const [tokens, setTokens] = useState<TopToken[]>()
+
+  const [page, setPage] = useState(1)
+  const selectedTokens = useFilteredTokens(useSortedTokens(prefetchedTokens.topTokens))
+  const [isFetching, setIsFetching] = useState(true)
```
JFrankfurt, line 143: pref `loading`, `setLoading` if that's what this is being used for

src/graphql/data/TopTokens.ts
```diff
@@ -0,0 +1,222 @@
...
+        { contracts, duration },
+        { fetchPolicy: 'store-or-network' }
+      )
+        .toPromise()
+        .then(onSuccess)
+    },
+    [duration, environment]
+  )
+
+  const loadMoreTokens = () => {
```
JFrankfurt, line 161: make this a `useCallback`, I think

src/graphql/data/TopTokens.ts
```diff
@@ -0,0 +1,222 @@
...
+    const contracts = selectedTokens.slice(page * PAGE_SIZE, (page + 1) * PAGE_SIZE).map(toContractInput)
```
JFrankfurt, line 162: is `contracts` the correct name here? Should it be `contractInputs`?

src/graphql/data/TopTokens.ts
```diff
@@ -0,0 +1,222 @@
...
+    loadTokens(contracts, (data) => {
+      if (data?.tokens) {
+        setTokens([...(tokens ?? []), ...data.tokens])
```
JFrankfurt, line 165: should the `tokens` value here be `selectedTokens`?

src/graphql/data/TopTokens.ts
```diff
@@ -0,0 +1,222 @@
...
+
+  // Reset count when filters are changed
+  useEffect(() => {
+    setIsFetching(true)
+    setTokens([])
+    const contracts = selectedTokens.slice(0, PAGE_SIZE).map(toContractInput)
+    loadTokens(contracts, (data) => {
+      if (data?.tokens) {
+        console.log('D')
+        setTokens([...data.tokens])
```
JFrankfurt, line 179: is this spread necessary? can you just do `setTokens(data.tokens)`?

src/graphql/data/TopTokens.ts
```diff
@@ -0,0 +1,222 @@
...
+        setIsFetching(false)
+        setPage(1)
+      }
+    })
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
JFrankfurt, line 184: if this is necessary, then you should add a comment explaining why the specific variables that are missing were excluded

src/graphql/data/util.ts
```diff
@@ -0,0 +1,49 @@
...
+  [SupportedChainId.CELO_ALFAJORES]: 'CELO',
+  [SupportedChainId.ARBITRUM_ONE]: 'ARBITRUM',
+  [SupportedChainId.ARBITRUM_RINKEBY]: 'ARBITRUM',
+  [SupportedChainId.OPTIMISM]: 'OPTIMISM',
+  [SupportedChainId.OPTIMISTIC_KOVAN]: 'OPTIMISM',
+}
+
+export function useCurrentChainName() {
+  const { chainId } = useWeb3React()
+  return chainId ? CHAIN_IDS_TO_BACKEND_NAME[chainId] : 'ETHEREUM'
```
JFrankfurt, line 48: I think this will crash the app when the connected chain isn't supported by the protocol (but specifically by CHAIN_IDS_TO_BACKEND_NAME)

JFrankfurt (changes requested):

src/components/Tokens/state.ts
cartcrom: I changed to a boolean for simplicity sake since there will only ever be two sorting directions. I believe I did remove SortDirection as well

src/graphql/data/TopTokens.ts
cartcrom: I'm following the general naming scheme of queries using $contracts which are of type ContractInput

src/graphql/data/TopTokens.ts
cartcrom: selectedTokens are the prefetched tokens w/ filters and sorting applied. tokens are the fully-fetched displayable tokens. loading more combines the existing results w/ new ones. I'll rename selectedTokens for clarity

src/components/Tokens/state.ts
JFrankfurt: Ah, you're right. Didn't see the deleted file.

src/graphql/data/TopTokens.ts
cartcrom: data.tokens is readonly. would Array.from(data.tokens) be better here?

JFrankfurt: Oh weird that ts blocks you from doing this. The `tokens` value should (kinda) be read only too.

I'd just assert it's a writable type or do //@ts-ignore if you can't figure out a smooth way to make it work.

My opinion in this kind of edge case is that it's better to manipulate the type system into what we want at runtime than to manipulate our business logic into something the type system will accept.

src/graphql/data/util.ts
cartcrom: fixed

src/graphql/thegraph/schema.graphql
```diff
@@ -18,6 +18,9 @@ type _Block_ {
 
   """The block number"""
   number: Int!
+
+  """Integer representation of the timestamp stored in blocks for the chain"""
+  timestamp: Int
```
JFrankfurt, line 23: why is this diff here?

JFrankfurt (approved):

PR #4666, opened 2022-09-19
fix: updates Widget defaultToken on URL change

vm:
When the URL of a page changes the defaultToken of the Widget should change. Right now that state doesn't get reset on re-renders, so you may see a stale token input.

### DIFF
--- a/src/components/Widget/inputs.tsx
+++ b/src/components/Widget/inputs.tsx
@@ -1,6 +1,5 @@
 import { Currency, Field, SwapController, SwapEventHandlers, TradeType } from '@uniswap/widgets'
 import CurrencySearchModal from 'components/SearchModal/CurrencySearchModal'
-import usePrevious from 'hooks/usePrevious'
 import { useCallback, useEffect, useMemo, useState } from 'react'
 
 /**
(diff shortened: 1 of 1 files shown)


src/components/Widget/inputs.tsx
```diff
@@ -17,6 +18,16 @@ export function useSyncWidgetInputs(defaultToken?: Currency) {
   const [tokens, setTokens] = useState<{ [Field.INPUT]?: Currency; [Field.OUTPUT]?: Currency }>({
     [Field.OUTPUT]: defaultToken,
   })
+
+  const previousDefaultToken = usePrevious(defaultToken)
+  useEffect(() => {
+    if (previousDefaultToken && previousDefaultToken !== defaultToken) {
```
zzmp, line 24: I don't think you need to check the previous token at all (and doing so adds extra state to the hook). Is there a case where you would _not_ want to update `tokens` when `defaultToken` updates? I'd just always reset it (if it's not already set):

    setTokens((tokens) => {
      if (tokens[Field.OUTPUT] === defaultToken) return tokens // this is a nop
      return { [Field.OUTPUT]: defaultToken }
    }, [defaultToken])

zzmp (changes requested):

src/components/Widget/inputs.tsx
vm: ah i see, works for me

vm: probably don't even need the default token check inside actually?

zzmp (approved):

PR #4669, opened 2022-09-19
fix: make token detail chart full width and x axis fixes

lynnshaoyu:
chart changes:
1) make chart full width as per https://uniswaplabs.atlassian.net/browse/WEB-1181?atlOrigin=eyJpIjoiMzYwN2VmOWYwOGJlNGY1ODlkYTMyNGM5N2M4OWFkM2EiLCJwIjoiaiJ9

2) add offsets to start and end time in x axis so the dates dont get cut off (because chart extends full width now) 

3) Limit to 6 ticks on "All" time chart, which has variable time frame, to avoid situations like this: 
Screen Shot 2022-09-19 at 6 03 04 PM

### DIFF
--- a/src/components/Tokens/TokenDetails/PriceChart.tsx
+++ b/src/components/Tokens/TokenDetails/PriceChart.tsx
@@ -116,7 +116,6 @@ const TimeButton = styled.button<{ active: boolean }>`
 
 const margin = { top: 100, bottom: 48, crosshair: 72 }
 const timeOptionsHeight = 44
-const crosshairDateOverhang = 80
 
 interface PriceChartProps {
   width: number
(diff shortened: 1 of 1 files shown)


infredible:
one thing im realizing is that this hover experience would feel a lot better if we removed the hashes on the x-axis intervals like so:
image

i think it'll help not make it feel weird when the crosshair doesn't line up with a hashmark

lynnshaoyu:
> one thing im realizing is that this hover experience would feel a lot better if we removed the hashes on the x-axis intervals like so: image
> 
> i think it'll help not make it feel weird when the crosshair doesn't line up with a hashmark

makes sense, i removed the ticks in the latest commit.

cartcrom (commented):
@lynnshaoyu the tick count and spacing here has changed a lot
image



@infredible do we want the times to be more centered? Also, w/ removing the tick lines under the timestamps, should the timestamps float a little lower/closer to the axis maybe?
image

lynnshaoyu:
> @lynnshaoyu the tick count and spacing here has changed a lot image
> 
> @infredible do we want the times to be more centered? Also, w/ removing the tick lines under the timestamps, should the timestamps float a little lower/closer to the axis maybe? image

yeah so I actually think this is intended behavior. if we up it more, we start seeing the crowded ticks again (in the PR description). the timeTicks function gets as close to the specific number of ticks as possible (in our case, we specified 6 intervals as the desired amount) see documentation here: https://github.com/d3/d3-time/blob/main/README.md#utcTicks:~:text=the%20same%20arguments.-,%23,-d3.utcTicks

However I decreased the offsets at both ends to reduce the size of possible "gaps" without ticks at the ends:
Screen Shot 2022-09-20 at 3 35 24 PM

infredible:
@cartcrom not sure what you mean by "more centered"

And ya, good eye, i think the labels can move down a bit. Should be 4px distance from the bottom border

gbugyis (approved):

PR #4670, opened 2022-09-19
fix: close WalletModal on connection

vm:
Once the new Web3Status is released we can deprecate the Wallet info view in WalletModal as well

### DIFF
--- a/src/components/WalletModal/index.tsx
+++ b/src/components/WalletModal/index.tsx
@@ -8,6 +8,7 @@ import { AutoColumn } from 'components/Column'
 import { AutoRow } from 'components/Row'
 import { getConnection, getConnectionName, getIsCoinbaseWallet, getIsInjected, getIsMetaMask } from 'connection/utils'
 import { RedesignVariant, useRedesignFlag } from 'featureFlags/flags/redesign'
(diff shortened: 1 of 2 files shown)


src/components/WalletModal/index.tsx
```diff
@@ -174,6 +177,12 @@ export default function WalletModal({
     }
   }, [walletModalOpen, setWalletView, account])
 
+  useEffect(() => {
+    if (account && account !== previousAccount && walletModalOpen) {
+      toggleWalletModal()
+    }
```
aballerr, line 183: would it make sense to also set the previous account here?  I guess I'm asking if there is an edge case where ```acount !=== previousAccount``` and  the previous account is never updated, so the wallet toggles back and forth indefinitely

aballerr: well I guess not indefinitely, but everytime you open it

vm: but that would reset previous account on the next render, so it won't lead to an infinite loop. since there will be a new render prior to the next user action

aballerr: ok, makes sense.

aballerr (approved):

PR #4675, opened 2022-09-20
chore: Merging in search, sort

aballerr:
1. Added in search at top of page
2. Added in sort at top of page
3. Added in url filters
4. Fixed logic to show a div when user inputs an invalid address


src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -1,31 +1,178 @@
...
-import { useEffect, useMemo, useState } from 'react'
+import { DropDownOption, GenieAsset, GenieCollection, UniformHeight, UniformHeights } from 'nft/types'
+import qs from 'query-string'
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import InfiniteScroll from 'react-infinite-scroll-component'
 import { useInfiniteQuery } from 'react-query'
+import { useLocation } from 'react-router-dom'
+
+const rarityStatusCache = new Map<string, boolean>()
+function getRarityStatus(id: string, assets?: (GenieAsset | undefined)[]) {
```
JackShort, line 31: `rarityStatusCache` should really be declared inside the function. Also we can put `getRarityStatus` inside of `asset.ts` util file

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -1,31 +1,178 @@
...
+  const hasRarity = assets && Array.from(assets).reduce((reducer, asset) => !!(reducer || asset?.rarity), false)
+
+  if (hasRarity) {
+    rarityStatusCache.set(id, hasRarity)
+  }
+
+  return hasRarity
+}
+
+const urlParamsUtils = {
```
JackShort, line 44: can we move this to its own util file

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -1,31 +1,178 @@
...
+  removeDefaults: (query: any) => {
```
JackShort, line 45: We know the query type is `Record<string, any>`

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -1,31 +1,178 @@
...
+    }
+
+    return clonedQuery
+  },
+
+  // Making values in our URL more state-friendly
+  buildQuery: (query: any, collectionStats: GenieCollection) => {
+    const clonedQuery: Record<string, any> = { ...query }
+
+    ;['traits', 'markets'].forEach((key) => {
```
JackShort, line 75: This can be declared as an array then looped through appropriately without the weird `;`

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -1,31 +1,178 @@
...
+      })
+    } catch (err) {
+      clonedQuery['traits'] = []
+    }
+
+    return clonedQuery
+  },
+}
+
+const trimTraitStr = (trait: string) => {
```
JackShort, line 157: should go in utils function with above function

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -43,18 +190,37 @@ export const CollectionNfts = ({ contractAddress }: CollectionNftsProps) => {
...
+        case SortBy.CommonToRare: {
+          sort = { 'rarity.providers.0.rank': -1 }
+          break
+        }
+        default:
+      }
+
       return await AssetsFetcher({
         contractAddress,
+        sort: sort ?? undefined,
```
JackShort, line 220: couldn't you just declare sort where it will start as undefined instead of assigning it to null?

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -75,23 +241,165 @@ export const CollectionNfts = ({ contractAddress }: CollectionNftsProps) => {
...
+      sort: string
+      sortBy: number
+      all: boolean
+      buyNow: boolean
+      traits: string[]
+      markets: string[]
+    }
+    const modifiedQuery = urlParamsUtils.buildQuery(query, collectionStats)
+
+    requestAnimationFrame(() => {
```
JackShort, line 313: this is weird - the `useCollectionFilters.setState(modifiedQuery as any)` should probably happen in the `useEffect` where this function is called and `modifiedQuery` should be returned from this function

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -75,23 +241,165 @@ export const CollectionNfts = ({ contractAddress }: CollectionNftsProps) => {
...
+    })
+    setMarketCount(marketCount)
+    oldStateRef.current = useCollectionFilters.getState()
+  }, [collectionStats?.marketplaceCount, setMarketCount])
+
+  // Applying filters from URL to local state
+  useEffect(() => {
+    collectionStats?.traits && applyFiltersFromURL()
+
+    collectionStats?.traits &&
```
JackShort, line 395: both of these could fall under the same if-statement

src/nft/components/collection/CollectionSearch.tsx
```diff
@@ -0,0 +1,28 @@
...
+      borderColor={{ default: 'medGray', focus: 'genieBlue' }}
+      borderWidth="1px"
+      borderStyle="solid"
+      borderRadius="12"
+      padding="12"
+      backgroundColor="white"
+      fontSize="16"
+      height="44"
+      color={{ placeholder: 'darkGray', default: 'blackBlue' }}
+      value={searchByNameText}
```
cbachmeier, line 21: useDebounce to wait till after the user stops typing. Can check example with universal searchbar

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -77,21 +104,95 @@ export const CollectionNfts = ({ contractAddress }: CollectionNftsProps) => {
...
+
   useEffect(() => {
     setUniformHeight(UniformHeights.unset)
   }, [contractAddress])
 
-  if (!collectionNfts) {
-    // TODO: collection unavailable page
-    return <div>No CollectionAssets</div>
-  }
+  useEffect(() => {
```
cbachmeier, line 170: We should still have the catch for the unavailable page, right now it's just a black page with the filter menu. You can test by inputting a bad collectionAddress in the url

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -77,21 +104,95 @@ export const CollectionNfts = ({ contractAddress }: CollectionNftsProps) => {
...
   const collectionNfts = useMemo(() => {
     if (!collectionAssets || !AssetsFetchSuccess) return undefined
 
     return collectionAssets.pages.flat()
   }, [collectionAssets, AssetsFetchSuccess])
 
+  const rarityStatusCache = new Map<string, boolean>()
+  const hasRarity = getRarityStatus(rarityStatusCache, collectionStats?.address, collectionNfts)
+
+  const sortDropDownOptions: DropDownOption[] = useMemo(
```
cbachmeier, line 120: Getting the error `dropDownOptions[selectedIndex] is undefined` regularly

cbachmeier (changes requested):

src/nft/utils/urlParams.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  // Applying local state changes to URL
+  const url = window.location.href.split('?')[0]
+  const stringifiedQuery = qs.stringify(modifiedQuery, { arrayFormat: 'comma' })
+
+  // Using pushState on purpose here. router.push() will trigger re-renders & API calls.
+  window.history.pushState({}, ``, `${url}${stringifiedQuery && `?${stringifiedQuery}`}`)
+}
+
+export const applyFiltersFromURL = (location: Location, collectionStats: GenieCollection) => {
```
cbachmeier, line 170: on genie the url filters had a useeffect where they would be cleared on a change of pathname that we should add back, changing pages for me is maintaining the filters

src/nft/pages/collection/index.tsx
```diff
@@ -45,38 +45,48 @@ const Collection = () => {
...
+                width: gridWidthOffset.interpolate((x) => `calc(100% - ${x as number}px)`),
+              }}
+            >
+              <CollectionNfts collectionStats={collectionStats} contractAddress={contractAddress} />
+            </AnimatedBox>
+          </Row>
+        </>
+      ) : (
+        // TODO: Put no collection asset page here
+        !isLoading && <div className={styles.noCollectionAssets}>No collection assets exist at this address</div>
```
aballerr, line 88: I moved this logic here and it will be the only thing that renders now when no collections exists (besides the navbar).  The previous logic wasn't working prior to my changes and also included some filter stuff that isn't relevant.

cbachmeier (approved):
LGTM. Not required but would be nice to use useDebounce for the collection search otherwise create a ticket in the backlog so we can follow up later

PR #4676, opened 2022-09-20
fix: makes sure wallet error is visible

vm:
Since the header height changed we want to make sure the error is always visible. Ideally, this should be done more dynamically but that's outside the scope of this fix.
Screen Shot 2022-09-20 at 3 12 05 PM


cbachmeier (approved):
LGTM

PR #4677, opened 2022-09-20
fix: center SearchBar vertically

vm:
after:
Screen Shot 2022-09-20 at 3 37 02 PM

before:
Screen Shot 2022-09-20 at 3 37 07 PM

Honestly, not sure why this is necessary. I think longer term we'll want to handle this horizontal centering within Nav via `justify-self: center` rather than fixed positioning.

### DIFF
--- a/src/components/NavBar/SearchBar.css.ts
+++ b/src/components/NavBar/SearchBar.css.ts
@@ -34,7 +34,7 @@ export const searchBarContainer = style([
     '@media': {
       [`screen and (min-width: ${breakpoints.lg}px)`]: {
         right: `-${DESKTOP_NAVBAR_WIDTH / 2 - MAGNIFYING_GLASS_ICON_WIDTH}px`,
-        top: -'5px',
+        top: '-5px',
       },
     },
   },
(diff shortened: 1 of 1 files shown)


src/components/NavBar/SearchBar.css.ts
```diff
@@ -34,6 +34,7 @@ export const searchBarContainer = style([
     '@media': {
       [`screen and (min-width: ${breakpoints.lg}px)`]: {
         right: `-${DESKTOP_NAVBAR_WIDTH / 2 - MAGNIFYING_GLASS_ICON_WIDTH}px`,
+        top: -'5px',
```
cbachmeier, line 37: nit: '-' should be inside the quotes?

cbachmeier:
the fixed positioning was so that the navbar can hover over the tabs menu when on tablet width
Screen Shot 2022-09-20 at 12 47 57

cbachmeier (approved):
LGTM, small nit/question

src/components/NavBar/SearchBar.css.ts
vm: oh wow yes, how does this even work

PR #4680, opened 2022-09-20
chore: feature flags token favorites

vm:
I just noticed in the design that the Network Filter goes before the TimeSelector as well

Screen Shot 2022-09-20 at 4 27 12 PM
Screen Shot 2022-09-20 at 4 27 19 PM

### DIFF
--- a/src/components/Tokens/TokenTable/TokenRow.tsx
+++ b/src/components/Tokens/TokenTable/TokenRow.tsx
@@ -5,6 +5,7 @@ import { EventName } from 'analytics/constants'
 import SparklineChart from 'components/Charts/SparklineChart'
 import CurrencyLogo from 'components/CurrencyLogo'
 import { getChainInfo } from 'constants/chainInfo'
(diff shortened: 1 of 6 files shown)


src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -41,11 +42,17 @@ const Cell = styled.div`
...
+  first?: boolean
+  last?: boolean
+  loading?: boolean
+  favoriteTokensEnabled?: boolean
+}>`
   background-color: transparent;
   display: grid;
   font-size: 15px;
-  grid-template-columns: 1fr 7fr 4fr 4fr 4fr 4fr 5fr 1.2fr;
+  grid-template-columns: ${({ favoriteTokensEnabled }) =>
```
zzmp, line 54: Can this just have the extra column even if not enabled?

zzmp (approved):

src/components/Tokens/TokenTable/TokenRow.tsx
vm: i thought so but it adds empty space lol. i am not convinced about using flex grid for this anymore.

PR #4682, opened 2022-09-20
fix: fully round corner on account dropdown

vm:
Screen Shot 2022-09-20 at 5 59 43 PM
Screen Shot 2022-09-20 at 5 59 50 PM

### DIFF
--- a/src/components/Web3Status/index.tsx
+++ b/src/components/Web3Status/index.tsx
@@ -34,6 +34,7 @@ import Loader from '../Loader'
 import { RowBetween } from '../Row'
 import WalletModal from '../WalletModal'
 
+// https://stackoverflow.com/a/31617326
 const FULL_BORDER_RADIUS = 9999
 
 const Web3StatusGeneric = styled(ButtonSecondary)`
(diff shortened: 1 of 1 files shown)


src/components/Web3Status/index.tsx
```diff
@@ -34,12 +34,14 @@ import Loader from '../Loader'
...
 
+const FULL_BORDER_RADIUS = 9999
+
 const Web3StatusGeneric = styled(ButtonSecondary)`
   ${({ theme }) => theme.flexRowNoWrap}
   width: 100%;
   align-items: center;
   padding: 0.5rem;
-  border-radius: 14px;
+  border-radius: ${FULL_BORDER_RADIUS}px;
```
zzmp, line 44: Why can't this just be 100%?

zzmp (changes requested):

src/components/Web3Status/index.tsx
vm: then it becomes an oval

vm: Screen Shot 2022-09-20 at 6 18 46 PM

zzmp: Can you link to the SO post above the const?

zzmp (approved):

PR #4685, opened 2022-09-21
feat: weth check

zzmp:
Throw an error before deposits can be made to the wrapped native currency on the wrong chain.
Includes a deep link in the error message for filing a bug, to encourage reporting.

### DIFF
--- a/src/hooks/useWrapCallback.tsx
+++ b/src/hooks/useWrapCallback.tsx
@@ -108,7 +108,8 @@ export default function useWrapCallback(
                       contract_chain_id: network.chainId,
                       type: WrapType.WRAP,
                     })
-                    const error = new Error('Invalid WETH contract')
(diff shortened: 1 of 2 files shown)


lynnshaoyu (approved):
lgtm. could you document the event + properties in the events tab here: https://docs.google.com/spreadsheets/d/1olyWPN_EGdbvoIH2c-ByJqibO1W7KMywLE5Nnm4c1RQ/edit#gid=45524225 ? ty

src/hooks/useWrapCallback.tsx
```diff
@@ -70,6 +70,11 @@ export default function useWrapCallback(
   )
   const addTransaction = useTransactionAdder()
 
+  // This allows an async error to propagate within the React lifecycle.
```
vm, line 73: not fully following here but seems good if it works

vm (approved):

PR #4687, opened 2022-09-21
fix: update swap box styles

vm:
Screen Shot 2022-09-27 at 1 03 54 PM
Screen Shot 2022-09-27 at 1 04 03 PM


vm:
I've decided to do this differently as SwapCurrencyInputPanel is a bit of a copy/paste situation

lynnshaoyu (approved):

PR #4688, opened 2022-09-21
fix: default empty hoverIndex on SearchBar open

vm:
Previously, onClick the first item would be hovered.


cbachmeier:
The original behaviour of having the first item focused was actually specifically requested from a previous debug session so that when a user presses enter after searching 'ETH' for example, the first option is selected.

vm:
Ah, I misunderstood a JIRA task I had.

PR #4693, opened 2022-09-21
fix: approval pending ux

zzmp:
- Fixes color values so that the icons for token approval are visible
- Transitions to the Loader icon immediately upon clicking the Approve button, without waiting for wallet interaction

### DIFF
--- a/src/pages/Swap/index.tsx
+++ b/src/pages/Swap/index.tsx
@@ -708,7 +708,7 @@ export default function Swap() {
                               approvalState === ApprovalState.APPROVED || signatureState === UseERC20PermitState.SIGNED
                             }
                           >
-                            <AutoRow justify="space-between" style={{ flexWrap: 'nowrap' }}>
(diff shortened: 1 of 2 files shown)


src/pages/Swap/index.tsx
```diff
@@ -714,8 +720,8 @@ export default function Swap() {
                                   </Trans>
                                 )}
                               </span>
-                              {approvalState === ApprovalState.PENDING ? (
-                                <Loader stroke="white" />
+                              {approvalPending || approvalState === ApprovalState.PENDING ? (
```
vm, line 723: is approvalState === PENDING still necessary? can it ever be true while approvalPending is false?

vm (approved):

src/pages/Swap/index.tsx
zzmp: I think it can if you continue to use the UX while waiting for the pending approval to go through

PR #4695, opened 2022-09-22
feat: log token explore search event

gbugyis:
Trigger event when user uses search on Token Explore page.  I'm assuming that device (mobile/tablet/desktop) is passed automatically.


lynnshaoyu (approved):
LGTM!

PR #4696, opened 2022-09-22
Jack/updated bag

JackShort:
* porting over updated bag - made changes that state logic is no longer handled in `useEffect`


cbachmeier (changes requested):

cbachmeier (approved):
LGTM

PR #4700, opened 2022-09-22
fix: track origin

zzmp:


src/analytics/Trace.tsx
```diff
@@ -44,7 +44,8 @@ export const Trace = memo(
 
     useEffect(() => {
       if (shouldLogImpression) {
-        sendAnalyticsEvent(name ?? EventName.PAGE_VIEWED, { ...combinedProps, ...properties })
+        const origin = window.location.origin
```
vm, line 47: should it be on all events or just page viewed

zzmp: just page viewed. other events can refer back to this one.

PR #4703, opened 2022-09-22
fix: we should be logging local/ vercel events to console 

lynnshaoyu:


src/analytics/index.ts
```diff
@@ -35,7 +35,7 @@ export function initializeAnalytics() {
 
 /** Sends an event to Amplitude. */
 export function sendAnalyticsEvent(eventName: string, eventProperties?: Record<string, unknown>) {
-  if (!API_KEY) {
+  if (!isProductionEnv() && API_KEY) {
```
vm, line 38: (!isProductionEnv() || !API_KEY)

src/analytics/index.ts
```diff
@@ -35,9 +35,10 @@ export function initializeAnalytics() {
 
 /** Sends an event to Amplitude. */
 export function sendAnalyticsEvent(eventName: string, eventProperties?: Record<string, unknown>) {
-  if (!isProductionEnv() && API_KEY) {
+  if (!API_KEY) return
+
+  if (!isProductionEnv()) {
```
vm, line 40: i guess i'm not following. what is the test environment for?

PR #4705, opened 2022-09-22
fix: fix explore table filtering and sorting bugs

lynnshaoyu:
1) get rid of data flickering when you change time period 
2) fix search string being ignored if you change the time period you’re viewing
3) make search filter queries reset /disappear when you navigate to a different page and come back. 
4) when user starts typing a search term, show correct number of loading rows (we have this info from prefetched top 100 tokens query), and load actual rows once price history token data becomes available. 
5) add local "cache" of preloaded tokens with price history (if we have already rendered them previously)

### DIFF
--- a/src/components/Tokens/TokenTable/SearchBar.tsx
+++ b/src/components/Tokens/TokenTable/SearchBar.tsx
@@ -59,10 +59,14 @@ const SearchInput = styled.input`
 `
 
 export default function SearchBar() {
-  const currentFilterString = useAtomValue(filterStringAtom)
-  const [localFilterString, setLocalFilterString] = useState(currentFilterString ?? '')
+  const currentString = useAtomValue(filterStringAtom)
+  const [localFilterString, setLocalFilterString] = useState(currentString)
   const setFilterString = useUpdateAtom(filterStringAtom)
   const debouncedLocalFilterString = useDebounce(localFilterString, 300)
+
+  useEffect(() => {
+    setLocalFilterString(currentString)
+  }, [currentString])
 
   useEffect(() => {
     setFilterString(debouncedLocalFilterString)
--- a/src/components/Tokens/TokenTable/SearchBar.tsx
+++ b/src/components/Tokens/TokenTable/SearchBar.tsx
@@ -2,7 +2,7 @@ import { Trans } from '@lingui/macro'
 import searchIcon from 'assets/svg/search.svg'
 import xIcon from 'assets/svg/x.svg'
 import useDebounce from 'hooks/useDebounce'
-import { useUpdateAtom } from 'jotai/utils'
+import { useAtomValue, useUpdateAtom } from 'jotai/utils'
 import { useEffect, useState } from 'react'
 import styled from 'styled-components/macro'
 
@@ -59,7 +59,8 @@ const SearchInput = styled.input`
 `
 
 export default function SearchBar() {
-  const [localFilterString, setLocalFilterString] = useState('')
+  const currentFilterString = useAtomValue(filterStringAtom)
+  const [localFilterString, setLocalFilterString] = useState(currentFilterString ?? '')
   const setFilterString = useUpdateAtom(filterStringAtom)
   const debouncedLocalFilterString = useDebounce(localFilterString, 300)
 
--- a/src/components/Tokens/TokenTable/TokenTable.tsx
+++ b/src/components/Tokens/TokenTable/TokenTable.tsx
@@ -70,6 +70,7 @@ export default function TokenTable() {
 
   // TODO: consider moving prefetched call into app.tsx and passing it here, use a preloaded call & updated on interval every 60s
   const { loading, tokens, tokensWithoutPriceHistoryCount, hasMore, loadMoreTokens } = useTopTokens()
+  const showMoreLoadingRows = Boolean(loading && hasMore)
 
   const observer = useRef<IntersectionObserver>()
   const lastTokenRef = useCallback(
@@ -122,7 +123,7 @@ export default function TokenTable() {
                   ref={index + 1 === tokens.length ? lastTokenRef : undefined}
                 />
               ))}
-              {loading && hasMore && LoadingMoreRows}
+              {showMoreLoadingRows && LoadingMoreRows}
             </TokenDataContainer>
           </GridContainer>
         </>
--- a/src/components/Tokens/TokenTable/TokenTable.tsx
+++ b/src/components/Tokens/TokenTable/TokenTable.tsx
@@ -1,8 +1,6 @@
 import { Trans } from '@lingui/macro'
 import { showFavoritesAtom } from 'components/Tokens/state'
-import { filterTimeAtom } from 'components/Tokens/state'
 import { PAGE_SIZE, useTopTokens } from 'graphql/data/TopTokens'
-import { toHistoryDuration } from 'graphql/data/util'
 import { useAtomValue } from 'jotai/utils'
 import { ReactNode, useCallback, useRef } from 'react'
 import { AlertTriangle } from 'react-feather'
@@ -72,8 +70,6 @@ export default function TokenTable() {
 
   // TODO: consider moving prefetched call into app.tsx and passing it here, use a preloaded call & updated on interval every 60s
   const { loading, tokens, tokenCount, loadMoreTokens } = useTopTokens()
-  const duration = toHistoryDuration(useAtomValue(filterTimeAtom))
-  console.log('duration', duration)
   const hasMore = !tokens || tokens.length < MAX_TOKENS_TO_LOAD
 
   const observer = useRef<IntersectionObserver>()
@@ -120,7 +116,7 @@ export default function TokenTable() {
             <TokenDataContainer>
               {tokens.map((token, index) => (
                 <LoadedRow
-                  key={token?.name}
+                  key={token?.address}
                   tokenListIndex={index}
                   tokenListLength={tokens?.length ?? 0}
                   token={token}
--- a/src/components/Tokens/TokenTable/TokenTable.tsx
+++ b/src/components/Tokens/TokenTable/TokenTable.tsx
@@ -1,7 +1,7 @@
 import { Trans } from '@lingui/macro'
 import { showFavoritesAtom } from 'components/Tokens/state'
 import { filterTimeAtom } from 'components/Tokens/state'
-import { PAGE_SIZE, usePrefetchTopTokens, useTopTokens } from 'graphql/data/TopTokens'
+import { PAGE_SIZE, useTopTokens } from 'graphql/data/TopTokens'
 import { toHistoryDuration } from 'graphql/data/util'
 import { useAtomValue } from 'jotai/utils'
 import { ReactNode, useCallback, useRef } from 'react'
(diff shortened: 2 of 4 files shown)


JFrankfurt:
typed in a search string to filter the results:
- the two resulting list elements flipped to a loading state (I don't think they should have, since they were already rendered)
- two POST requests were sent, but the second was empty

JFrankfurt:
nit: if I text filter for a token here's what I experience:
- full page loading experience (expected)
- loading experience for the rows that result from my search string (weird! we know enough to show those rows, so why are they in a loading state??)
- loaded rows render 👍

It seems like it should be possible to drop the second state from this flow!

JFrankfurt:
not sure how, but I was able to get to a state where it seemed like my filter string was ignored

lynnshaoyu:
> not sure how, but I was able to get to a state where it seemed like my filter string was ignored

ooof can u lmk how to repro if you figure out the steps :O

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -55,13 +54,13 @@ function NoTokensState({ message }: { message: ReactNode }) {
...
-const InitialLoadingRows = Array(ROWS_PER_PAGE_FETCH).fill(<LoadingRow />)
+const LoadingRows = (rowCount?: number) => Array(rowCount ?? PAGE_SIZE).fill(<LoadingRow />)
 
-export function LoadingTokenTable() {
+export function LoadingTokenTable({ rowCount }: { rowCount?: number }) {
   return (
     <GridContainer>
       <HeaderRow />
-      <TokenDataContainer>{InitialLoadingRows}</TokenDataContainer>
+      <TokenDataContainer>{LoadingRows(rowCount)}</TokenDataContainer>
```
JFrankfurt, line 63: <3

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -118,14 +115,14 @@ export default function TokenTable() {
             <TokenDataContainer>
               {tokens.map((token, index) => (
                 <LoadedRow
-                  key={token?.name}
+                  key={token?.address}
                   tokenListIndex={index}
                   tokenListLength={tokens?.length ?? 0}
                   token={token}
-                  ref={tokens.length === index + 1 ? lastTokenRef : undefined}
+                  ref={index + 1 === tokens.length ? lastTokenRef : undefined}
```
JFrankfurt, line 122: why this change?

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -118,14 +115,14 @@ export default function TokenTable() {
...
                 />
               ))}
-              {loading && LoadingMoreRows}
```
JFrankfurt, line 128: compress the compound boolean render conditional to a computed single boolean:
`const showMoreLoadingRows = Boolean(loading && hasMore)`
...
`{showMoreLoadingRows && LoadingMoreRows}`

src/components/Tokens/TokenTable/TokenTable.tsx
lynnshaoyu: lmao because my brain thinks this makes more sense XD 

we are checking if the current index is equal to the tokens length, not vice versa. there is no diff tbh.

src/graphql/data/TopTokens.ts
```diff
@@ -129,63 +130,101 @@ function toContractInput(token: PrefetchedTopToken) {
...
-  loadMoreTokens: () => void
+// Map of key: ${chain} + ${address} and value: TopToken object.
+// Acts as a local cache.
+const tokensWithPriceHistoryCache: Record<string, TopToken> = {}
+
+const checkIfAllTokensCached = (prefetchedSelectedTokens: PrefetchedTopToken[]) => {
+  let everyTokenHasBeenPreloaded = true
+  const preloadedSelectedTokens: TopToken[] = []
+  prefetchedSelectedTokens.map((token) =>
+    !!token && `${token.chain.toString()}${token.address}` in tokensWithPriceHistoryCache
```
JFrankfurt, line 141: I think `.toString` will be called here implicitly, so you don't need to include it.

src/graphql/data/TopTokens.ts
```diff
@@ -129,63 +130,101 @@ function toContractInput(token: PrefetchedTopToken) {
...
-      complete() {
+  const loadMoreTokens = useCallback(() => {
+    setLoading(true)
+    const contracts = prefetchedSelectedTokensWithoutPriceHistory
+      .slice(page * PAGE_SIZE, (page + 1) * PAGE_SIZE)
+      .map(toContractInput)
+    loadTokensWithPriceHistory(contracts, (data) => {
+      if (data?.tokens) {
+        data.tokens.forEach((token) =>
+          !!token ? (tokensWithPriceHistoryCache[`${token.chain.toString()}${token.address}`] = token) : null
```
JFrankfurt, line 192: same comment here re: `.toString`

src/graphql/data/TopTokens.ts
```diff
@@ -129,63 +130,101 @@ function toContractInput(token: PrefetchedTopToken) {
...
+      setLoading(false)
+      return
+    } else {
+      setLoading(true)
+      setTokens([])
+      const contracts = prefetchedSelectedTokensWithoutPriceHistory.slice(0, PAGE_SIZE).map(toContractInput)
+      loadTokensWithPriceHistory(contracts, (data) => {
+        if (data?.tokens) {
+          data.tokens.forEach((token) =>
+            !!token ? (tokensWithPriceHistoryCache[`${token.chain.toString()}${token.address}`] = token) : null
```
JFrankfurt, line 217: same here re: `.toString`

JFrankfurt (changes requested):
see comments

src/graphql/data/TopTokens.ts
lynnshaoyu: done

src/graphql/data/TopTokens.ts
lynnshaoyu: done

src/graphql/data/TopTokens.ts
lynnshaoyu: done

src/graphql/data/TopTokens.ts
```diff
@@ -120,7 +121,7 @@ function useFilteredTokens(tokens: PrefetchedTopToken[]) {
   }, [tokens, showFavorites, lowercaseFilterString, favorites])
 }
 
-const PAGE_SIZE = 20
+export const PAGE_SIZE = 20
```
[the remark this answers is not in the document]
lynnshaoyu: done

src/components/Tokens/TokenTable/SearchBar.tsx
```diff
@@ -59,10 +59,15 @@ const SearchInput = styled.input`
...
 export default function SearchBar() {
-  const [localFilterString, setLocalFilterString] = useState('')
+  const currentString = useAtomValue(filterStringAtom)
+  const [localFilterString, setLocalFilterString] = useState(currentString)
   const setFilterString = useUpdateAtom(filterStringAtom)
   const debouncedLocalFilterString = useDebounce(localFilterString, 300)
 
+  useEffect(() => {
+    setLocalFilterString(currentString)
+  }, [currentString])
```
cmcewen, line 69: where else in the app would be updating this value?

src/components/Tokens/TokenTable/TokenTable.tsx
cmcewen: do these rows need keys?

src/graphql/data/TopTokens.ts
```diff
@@ -8,18 +8,19 @@ import {
...
+import { useCurrentChainName } from './util'
 
-export function usePrefetchTopTokens() {
-  const duration = toHistoryDuration(useAtomValue(filterTimeAtom))
+export function usePrefetchTopTokens(duration: HistoryDuration) {
   const chain = useCurrentChainName()
-  const args = useMemo(() => ({ chain, duration }), [chain, duration])
-  return useLazyLoadQuery<TopTokens100Query>(topTokens100Query, args)
+
+  const top100 = useLazyLoadQuery<TopTokens100Query>(topTokens100Query, { duration, chain })
```
cmcewen, line 22: why not just return this value instead of assigning and returning that?

src/graphql/data/TopTokens.ts
```diff
@@ -129,63 +130,101 @@ function toContractInput(token: PrefetchedTopToken) {
...
+const tokensWithPriceHistoryCache: Record<string, TopToken> = {}
+
+const checkIfAllTokensCached = (prefetchedSelectedTokens: PrefetchedTopToken[]) => {
+  let everyTokenHasBeenPreloaded = true
+  const preloadedSelectedTokens: TopToken[] = []
+  prefetchedSelectedTokens.map((token) =>
+    !!token && `${token.chain.toString()}${token.address}` in tokensWithPriceHistoryCache
+      ? preloadedSelectedTokens.push(tokensWithPriceHistoryCache[`${token.chain.toString()}${token.address}`])
+      : (everyTokenHasBeenPreloaded = false)
+  )
```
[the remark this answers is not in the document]
lynnshaoyu: sounds good, made it clearer.

src/graphql/data/TopTokens.ts
```diff
@@ -129,63 +130,108 @@ function toContractInput(token: PrefetchedTopToken) {
...
+  }, [prefetchedSelectedTokensWithoutPriceHistory, page, loadTokensWithPriceHistory, tokens])
 
-  return { loading, tokens: useFilteredTokens(useSortedTokens(tokens)) as TopToken[], loadMoreTokens }
+  // Reset count when filters are changed
+  useLayoutEffect(() => {
+    const { everyTokenHasBeenPreloaded, preloadedSelectedTokens } = checkIfAllTokensCached(
+      prefetchedSelectedTokensWithoutPriceHistory
+    )
+    if (everyTokenHasBeenPreloaded) {
+      setTokens([...preloadedSelectedTokens])
```
cmcewen, line 207: why do you need to spread here?

src/graphql/data/TopTokens.ts
```diff
@@ -129,63 +130,108 @@ function toContractInput(token: PrefetchedTopToken) {
...
+      setLoading(false)
+      return
+    } else {
+      setLoading(true)
+      setTokens([])
+      const contracts = prefetchedSelectedTokensWithoutPriceHistory.slice(0, PAGE_SIZE).map(toContractInput)
+      loadTokensWithPriceHistory(contracts, (data) => {
+        if (data?.tokens) {
+          data.tokens.forEach((token) =>
```
cmcewen, line 216: why do you mutate the array vs map here? would that fix the TS error?

src/components/Tokens/TokenTable/TokenTable.tsx
lynnshaoyu: done

src/graphql/data/TopTokens.ts
lynnshaoyu: that is a good idea haha, fixed.

src/components/Tokens/TokenTable/SearchBar.tsx
lynnshaoyu: https://github.com/Uniswap/interface/blob/0b9fda5b25adc56fed0a29ca84eb7d38c62ca348/src/pages/Tokens/index.tsx#L71

src/components/Tokens/TokenTable/TokenTable.tsx
lynnshaoyu: dont think so they're all the same

src/graphql/data/TopTokens.ts
lynnshaoyu: good idea, changed to map. that doesn't fix the ts error, but i spread the array instead and it fixed it

src/graphql/data/TopTokens.ts
lynnshaoyu: i dont need to :P removed spread.

src/components/Tokens/TokenTable/SearchBar.tsx
cmcewen: makes sense 👍

cmcewen (approved):

PR #4707, opened 2022-09-23
fix: show real balance in footer summary instead of dummy data

lynnshaoyu:
When balance is non zero: 
Screen Shot 2022-09-22 at 11 45 37 PM

When balance is zero: 
Screen Shot 2022-09-22 at 11 45 56 PM

### DIFF
--- a/src/components/Tokens/TokenDetails/BalanceSummary.tsx
+++ b/src/components/Tokens/TokenDetails/BalanceSummary.tsx
@@ -1,10 +1,6 @@
 import { Trans } from '@lingui/macro'
-import { useWeb3React } from '@web3-react/core'
-import { formatToDecimal } from 'analytics/utils'
 import { useToken } from 'hooks/Tokens'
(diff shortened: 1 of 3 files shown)


src/components/Tokens/TokenDetails/BalanceSummary.tsx
```diff
@@ -49,16 +45,18 @@ const TotalBalanceItem = styled.div`
...
 `
 
-export default function BalanceSummary({ address }: { address: string }) {
+export default function BalanceSummary({
+  address,
+  balanceNumber,
+  balanceUsdNumber,
+}: {
+  address: string
+  balanceNumber?: number
```
vm, line 54: nit: can you name this `balance` and usd would be `balanceValueUsd`

vm (approved):

src/components/Tokens/TokenDetails/BalanceSummary.tsx
lynnshaoyu: done

PR #4708, opened 2022-09-23
chore: Merging Loading states 2

aballerr:
1.  Image banner now has loading state
2.  Collection Profile has a loading state
3. Collection Stats has a loading state
5. Collection name, description and social media icons have a loading state
6. Added a zustand store to keep track of collectionStats loading and collectionNfts Loading.  
7. Added loading states to filter button, sort dropdown and search by name

This is the last pr for loading states


src/nft/components/collection/CollectionSearch.tsx
```diff
@@ -19,7 +23,8 @@ export const CollectionSearch = () => {
       height="44"
       color={{ placeholder: 'textSecondary', default: 'textPrimary' }}
       value={searchByNameText}
-      placeholder={'Search by name'}
+      placeholder={isCollectionNftsLoading ? '' : 'Search by name'}
+      className={clsx(isCollectionNftsLoading && styles.filterButtonLoading)}
```
cbachmeier, line 27: I think this conditional should be based off NFTState loading, it's strange when I type in a collection search query and the searchbar and filter bar become unusable in their loading states

src/nft/components/collection/CollectionStats.css.ts
```diff
@@ -113,3 +114,65 @@ export const statsValue = style([
...
+  sprinkles({
+    width: '60',
+    height: '16',
+  }),
+])
+
+export const descriptionLoading = style([
+  loadingAsset,
+  sprinkles({
+    fontSize: '16',
```
cbachmeier, line 138: why does the loading component need a fontSize?

src/nft/components/collection/CollectionStats.css.ts
```diff
@@ -113,3 +114,65 @@ export const statsValue = style([
...
+  loadingAsset,
+  sprinkles({
+    height: '32',
+  }),
+  {
+    width: 236,
+  },
+])
+
+export const iconsLoading = style([
```
cbachmeier, line 172: there might be some flex shrinking going on as the icon widths seem very squished

src/nft/components/collection/CollectionStats.tsx
```diff
@@ -124,14 +135,15 @@ const CollectionName = ({
   collectionSocialsIsOpen: boolean
   toggleCollectionSocials: () => void
 }) => {
+  const isCollectionStatsLoading = useIsCollectionLoading((state) => state.isCollectionStatsLoading)
+  const nameClass = isCollectionStatsLoading
+    ? styles.nameTextLoading
+    : clsx(isMobile ? headlineMedium : headlineMedium, styles.nameText)
```
cbachmeier, line 141: this clsx conditional doesn't seem to do anything

src/nft/components/collection/CollectionStats.tsx
```diff
@@ -124,14 +135,15 @@ const CollectionName = ({
...
+
   return (
     <Row justifyContent="space-between">
       <Row minWidth="0">
-        <Box
-          marginRight={!isVerified ? '12' : '0'}
-          className={clsx(isMobile ? headlineMedium : headlineMedium, styles.nameText)}
-          style={{ lineHeight: '32px' }}
-        >
+        <Box marginRight={!isVerified ? '12' : '0'} className={nameClass} style={{ lineHeight: '32px' }}>
```
cbachmeier, line 146: lineHeight can be removed as it's now included in headlineMedium

src/nft/components/collection/CollectionStats.tsx
```diff
@@ -143,8 +155,8 @@ const CollectionName = ({
           gap="8"
           height="32"
         >
-          {collectionStats.discordUrl ? (
-            <SocialsIcon href={collectionStats.discordUrl}>
+          {collectionStats.discordUrl || isCollectionStatsLoading ? (
+            <SocialsIcon href={collectionStats.discordUrl ?? ''}>
```
cbachmeier, line 159: Instead of adding these conditionals atomically and having some null check logic, we could do one check for isCollectionStatsLoading and map an array of length 4 to 4 divs with the loading class

src/nft/components/collection/CollectionStats.tsx
```diff
@@ -209,7 +222,11 @@ const CollectionDescription = ({ description }: { description: string }) => {
       setShowReadMore(true)
   }, [descriptionRef, baseRef])
 
-  return (
+  return isCollectionStatsLoading ? (
+    <Box marginTop={{ sm: '12', md: '16' }} className={styles.descriptionLoading}>
+      description
```
cbachmeier, line 227: why do we need this placeholder 'description' text?

src/nft/components/collection/CollectionStats.tsx
```diff
@@ -251,22 +273,22 @@ const StatsRow = ({ stats, isMobile, ...props }: { stats: GenieCollection; isMob
       <StatsItem label="Items" isMobile={isMobile ?? false}>
         {totalSupplyStr}
       </StatsItem>
-      {numOwnersStr ? (
+      {numOwnersStr || isCollectionStatsLoading ? (
```
cbachmeier, line 276: similar comments as above regardling doing 1 check for isCollectionStatsLoading and mapping an array to X number of divs

src/nft/components/common/SortDropdown/SortDropdown.tsx
```diff
@@ -70,51 +72,56 @@ export const SortDropdown = ({
         width={inFilters ? 'full' : 'inherit'}
         onClick={toggleOpen}
         cursor="pointer"
-        className={clsx(isOpen && !mini && styles.activeDropdown)}
+        className={isCollectionNftsLoading ? styles.isLoadingDropdown : clsx(isOpen && !mini && styles.activeDropdown)}
```
cbachmeier, line 75: The filter and search loading components overlap
Screen Shot 2022-09-28 at 11 59 40

src/nft/components/common/SortDropdown/SortDropdown.tsx
```diff
@@ -28,6 +29,7 @@ export const SortDropdown = ({
   const [isOpen, toggleOpen] = useReducer((s) => !s, false)
   const [isReversed, toggleReversed] = useReducer((s) => !s, false)
   const [selectedIndex, setSelectedIndex] = useState(0)
+  const isCollectionNftsLoading = useIsCollectionLoading((state) => state.isCollectionNftsLoading)
```
cbachmeier, line 32: I think this check and searchBar should be off isCollectionStatsLoading

src/nft/pages/collection/index.tsx
```diff
@@ -55,20 +60,27 @@ const Collection = () => {
...
+                  src={collectionStats?.bannerImageUrl}
+                  className={isLoading ? styles.loadingBanner : styles.bannerImage}
+                  background="none"
+                />
+              )}
+            </Box>
           </Box>
           <Column paddingX="32">
-            {collectionStats && <CollectionStats stats={collectionStats} isMobile={isMobile} />}
+            <CollectionStats stats={collectionStats || ({} as GenieCollection)} isMobile={isMobile} />
```
cbachmeier, line 83: Instead of having to do this typecasted null object, can we have the component `CollectionStats` return null if `collectionStats` is undefined?

cbachmeier (changes requested):

src/nft/pages/collection/index.tsx
aballerr: I had to change this because the loading states exist inside of these components.  I will update the logic so they render if it's loading or collection stats exist so it doesn't render in the case the stats do not exist for some reason

src/nft/components/collection/CollectionSearch.tsx
aballerr: I can switch.  Technically the filter count changes, but it is probably a better user experience this way

src/nft/components/collection/CollectionStats.css.ts
```diff
@@ -134,9 +134,6 @@ export const statsLabelLoading = style([
```
cbachmeier, line 134: I believe you can remove the iconsLoading style since the social icons loading state has been removed

src/nft/components/collection/CollectionStats.tsx
```diff
@@ -268,27 +252,39 @@ const StatsRow = ({ stats, isMobile, ...props }: { stats: GenieCollection; isMob
   const totalVolumeStr = ethNumberStandardFormatter(stats.stats?.total_volume)
   const floorPriceStr = ethNumberStandardFormatter(stats.floorPrice)
 
+  const statItems = new Array(5).fill(
```
cbachmeier, line 255: nit: I would give this a more meaningful var name such as statsLoadingSkeleton

src/nft/components/common/SortDropdown/SortDropdown.tsx
```diff
@@ -51,7 +53,7 @@ export const SortDropdown = ({
       borderBottomLeftRadius={isOpen ? '0' : undefined}
       borderBottomRightRadius={isOpen ? '0' : undefined}
       height="44"
-      style={{ width: inFilters ? 'full' : mini ? 'min' : maxWidth ? maxWidth : '300px' }}
+      style={{ width: isCollectionStatsLoading ? 220 : width }}
```
cbachmeier, line 56: nit: you could add the loading conditional to the `width` const. Also, do you need to add `px` to the end of `200`?

cbachmeier (approved):
LGTM had some nonblocking comments

PR #4710, opened 2022-09-23
feat: explore chain switching

cartcrom:
Adds the ability to view to non-mainnet tokens by switching networks on explore and token details, following @infredible's network switching audit. Switching wallet chain or switching the network-filter's chain will update the data displayed

The TokenDetails query was also refactored to use the new schema & provide the logoUrl (rather than relying on useCurrency for CurrencyLogo).

**explore urls are now:**
/tokens/\<chain>
_and_
/tokens/\<chain>/\<address>
_or_
/tokens/\<chain>/NATIVE

**Note:** hooks like useCurrency() and useToken() currently only retrieve data on the same chain as the connected wallet. Since explore pages' chain can differ from a wallets chain, this means these hooks don't work well on explore. I was able to remove their usage on explore (simplifying these pages to just use the backend-served data instead), but tokenSafety warning still use these hooks and need to be updated to work with chain switching


src/components/CurrencyLogo/index.tsx
```diff
@@ -27,17 +27,20 @@ export default function CurrencyLogo({
...
   size?: string
   style?: React.CSSProperties
+  src?: string | null
 }) {
+  const logoURIs = useCurrencyLogoURIs(currency)
   const props = {
     alt: `${currency?.symbol ?? 'token'} logo`,
     size,
-    srcs: useCurrencyLogoURIs(currency),
+    srcs: src ? [src] : logoURIs,
```
cmcewen, line 43: this will create a new array on every render

src/components/Tokens/TokenDetails/ChartSection.tsx
```diff
@@ -51,55 +52,52 @@ const TokenActions = styled.div`
...
   const toggleFavorite = useToggleFavorite(token.address)
-  const chainInfo = getChainInfo(token?.chainId)
-  const networkLabel = chainInfo?.label
-  const networkBadgebackgroundColor = chainInfo?.backgroundColor
-  const warning = checkWarning(token.address)
-
-  let currency = useCurrency(token.address)
+  const chainId = CHAIN_NAME_TO_CHAIN_ID[token.chain]
+  const L2Icon = getChainInfo(chainId).circleLogoUrl
+  const warning = checkWarning(token.address ?? '')
```
cmcewen, line 80: why did you add the `?? ''` here? it looks like the func signature didn't change

src/components/Tokens/TokenDetails/ChartSection.tsx
```diff
@@ -51,55 +52,52 @@ const TokenActions = styled.div`
...
       <TokenInfoContainer>
         <TokenNameCell>
-          <CurrencyLogo currency={currency} size={'32px'} symbol={tokenSymbol} />
-          {tokenName ?? <Trans>Name not found</Trans>}
-          <TokenSymbol>{tokenSymbol ?? <Trans>Symbol not found</Trans>}</TokenSymbol>
+          <LogoContainer>
+            <CurrencyLogo src={logoSrc} size={'32px'} symbol={nativeCurrency?.symbol ?? token.symbol} />
+            <L2NetworkLogo networkUrl={L2Icon} size={'16px'} />
+          </LogoContainer>
+          {nativeCurrency?.name ?? token.name ?? <Trans>Name not found</Trans>}
```
cmcewen, line 94: what does this failure case look like? if you're landing on a page with an address but we can't fetch the token name or symbol, doesn't that mean the token doesn't exist? so how could you get there?

src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -314,18 +313,18 @@ const SparkLineLoadingBubble = styled(LongLoadingBubble)`
   height: 4px;
 `
 
-const L2NetworkLogo = styled.div<{ networkUrl?: string }>`
-  height: 12px;
-  width: 12px;
+export const L2NetworkLogo = styled.div<{ networkUrl?: string; size?: string }>`
+  height: ${({ size }) => size ?? '12px'};
```
cmcewen, line 317: just to double check - this is the only way to do a default value in style components? ideally we could set the fallback value to size to be 12 and then you don't have to have these ternary expressions

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -116,15 +119,18 @@ export default function TokenTable() {
...
-                  tokenListLength={tokens?.length ?? 0}
-                  token={token}
-                  ref={tokens.length === index + 1 ? lastTokenRef : undefined}
-                />
-              ))}
+              {tokens.map(
+                (token, index) =>
+                  token && (
+                    <LoadedRow
+                      key={token?.name}
```
cmcewen, line 126: is this guaranteed unique?

cmcewen (commented):
few things to start on - i will take another pass

src/components/Tokens/TokenDetails/ChartSection.tsx
cartcrom: token was a Token() before as in a result of useToken(), now its just token data from the api. every element from the api is potentially null, but I'm planning at looking into the @required directive to see if we can get rid of some repetitive ternarys and nullchecking https://relay.dev/docs/next/guides/required-directive/

src/components/Tokens/TokenDetails/ChartSection.tsx
cartcrom: I think this is definitely something to make another PR/ticket for if there isn't already, figure out which data points we require else return. hopefully relay's @required also helps here tho, as it should throw out results that don't have required fields

src/components/Tokens/TokenTable/TokenRow.tsx
cartcrom: It looks like my only alternative is https://styled-components.com/docs/advanced, but i don't see any other usage of defaultProps in our repo

cartcrom: repetitive ternarys in styled components are really annoying

src/pages/TokenDetails/index.tsx
```diff
@@ -90,14 +95,41 @@ function NetworkBalances(tokenAddress: string | undefined) {
...
+    tokenAddress = WRAPPED_NATIVE_CURRENCY[chainId]?.address?.toLowerCase()
+  }
+
   const tokenWarning = tokenAddress ? checkWarning(tokenAddress) : null
   const isBlockedToken = tokenWarning?.canProceed === false
+
+  const timePeriod = useAtomValue(filterTimeAtom)
+  const currentChainName = validateUrlChainParam(chainName)
+  const token = useTokenQuery(tokenAddress ?? '', currentChainName, timePeriod).tokens?.[0]
+  console.log(token)
```
cmcewen, line 113: console.log here

cmcewen (approved):

PR #4717, opened 2022-09-24
fix: Comparing formattedData reference with [] doesn't work as expected

ICONationDevTeam:
This condition will always return 'false' since JavaScript compares objects by reference, not value. ts(2839)


src/components/LiquidityChartRangeInput/index.tsx
```diff
@@ -176,7 +176,7 @@ export default function LiquidityChartRangeInput({
           message={<Trans>Liquidity data not available.</Trans>}
           icon={<CloudOff size={56} stroke={theme.deprecated_text4} />}
         />
-      ) : !formattedData || formattedData === [] || !price ? (
```
zzmp, line 179: This is a referential equality check, so it will _always_ return `false`.

zzmp (approved):

PR #4718, opened 2022-09-26
feat: [Collection Stats] update listings copy and change value to percentage

gbugyis:
- updates Collection Stats 'Listings' copy text ('Listings' -> 'Listed') and changes the value from a raw number to percentage of total supply.


src/nft/components/collection/CollectionStats.tsx
```diff
@@ -240,7 +240,9 @@ const StatsItem = ({ children, label, isMobile }: { children: ReactNode; label:
 const StatsRow = ({ stats, isMobile, ...props }: { stats: GenieCollection; isMobile?: boolean } & BoxProps) => {
   const numOwnersStr = stats.stats ? putCommas(stats.stats.num_owners) : 0
   const totalSupplyStr = stats.stats ? putCommas(stats.stats.total_supply) : 0
-  const totalListingsStr = stats.stats ? putCommas(stats.stats.total_listings) : 0
+  const listedPercentageStr = stats.stats
+    ? ((stats.stats.total_listings / stats.stats.total_supply) * 100).toFixed(1)
```
cbachmeier, line 244: need a check to make sure `stats.stats.total_listings` is greater than 0

src/nft/components/collection/CollectionStats.tsx
```diff
@@ -267,8 +269,8 @@ const StatsRow = ({ stats, isMobile, ...props }: { stats: GenieCollection; isMob
         </StatsItem>
       ) : null}
       {stats.stats?.total_listings ? (
```
cbachmeier, line 271: Can you check with product to see if we want to show when there's `0%` listed?

src/nft/components/collection/CollectionStats.tsx
```diff
@@ -267,8 +269,8 @@ const StatsRow = ({ stats, isMobile, ...props }: { stats: GenieCollection; isMob
...
-        <StatsItem label="Listings" isMobile={isMobile ?? false}>
-          {totalListingsStr}
+        <StatsItem label="Listed" isMobile={isMobile ?? false}>
+          {listedPercentageStr} %
```
cbachmeier, line 273: I think we should get rid of the space before the `%` which would be in parity with our percentage behaviour elsewhere on the app

cbachmeier (changes requested):

src/nft/components/collection/CollectionStats.tsx
gbugyis: Hiding it for now until I get confirmation

cbachmeier (approved):
LGTM

PR #4721, opened 2022-09-26
fix: error when hitting enter on search

cartcrom:


src/graphql/data/util.ts
```diff
@@ -76,3 +76,12 @@ export function isValidBackendChainName(chainName: string | undefined): chainNam
   }
   return false
 }
+
+export function getTokenDetailsURL(address: string, chainName: Chain) {
```
lynnshaoyu, line 80: i think you could combine  getTokenDetailsURL and  getTokenDetailsURLWithId into one function possibly for ease of maintenance in the future since they're doing the same thing, just with different input params. like you can have chainName and chainId both be optional params, if both are undefined return ''.

lynnshaoyu (approved):
optional suggestion on combining the two getTokenDetailsUrl functions

PR #4723, opened 2022-09-26
refactor: remove unnecessary auth

cartcrom:
removed AWS authentication which is no longer needed after backend updates.

A private api key is no longer needed to fetch schema as well, so schema has been removed from the repo, and will be fetched automatically by yarn build now


.gitignore
```diff
@@ -10,6 +10,7 @@
 
 # generated graphql types
 __generated__/
+schema.graphql
```
vm, line 13: why can this be ignored now?

fetch-schema.js
```diff
@@ -8,11 +8,5 @@ const thegraphConfig = require('./relay_thegraph.config')
...
 
-const API_URL = process.env.REACT_APP_GQL_API_URL
-const API_KEY = process.env.REACT_APP_GQL_API_KEY
-
-if (API_URL && API_KEY) {
-  exec(`get-graphql-schema ${API_URL} --h X-API-KEY=${API_KEY} > ${dataConfig.schema}`)
-} else {
-  console.log('REACT_APP_GQL_API_URL or REACT_APP_GQL_API_KEY is missing from env.local')
-}
+const DATA_API_URL = 'https://api.uniswap.org/v1/graphql'
```
vm, line 11: nit: should this use the process.env variable still?

vm (approved):

.gitignore
cartcrom: now it gets fetched before build

fetch-schema.js
cartcrom: fair point, fixed

PR #4730, opened 2022-09-27
feat: add deep shadow on card hover

cbachmeier:
Add deep shadow as detailed in Figma specs
Add deep shadow to nft cards on hover


JackShort (approved):
lgtm

PR #4737, opened 2022-09-27
fix: resolve bug where changing duration when filter string is applied does not change data (original duration is still applied)

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-1371?atlOrigin=eyJpIjoiZGRhNjMxMmMyZGE4NGI4MTgyOGQyMTgwODI4ZTUxNjAiLCJwIjoiaiJ9

### DIFF
--- a/src/graphql/data/TopTokens.ts
+++ b/src/graphql/data/TopTokens.ts
@@ -132,18 +132,18 @@ function toContractInput(token: PrefetchedTopToken) {
   }
 }
 
-// Map of key: ${chain} + ${address} and value: TopToken object.
+// Map of key: ${HistoryDuration} and value: another Map, of key:${chain} + ${address} and value: TopToken object.
 // Acts as a local cache.
 const tokensWithPriceHistoryCache: Record<string, Record<string, TopToken>> = {}
 
-const checkIfAllTokensCached = (tokens: PrefetchedTopToken[]) => {
+const checkIfAllTokensCached = (duration: string, tokens: PrefetchedTopToken[]) => {
   let everyTokenInCache = true
   const cachedTokens: TopToken[] = []
 
   const checkCache = (token: PrefetchedTopToken) => {
     const tokenCacheKey = !!token ? `${token.chain}${token.address}` : ''
-    if (tokenCacheKey in tokensWithPriceHistoryCache) {
-      cachedTokens.push(tokensWithPriceHistoryCache[tokenCacheKey])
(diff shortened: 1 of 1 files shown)


src/graphql/data/TopTokens.ts
```diff
@@ -132,18 +132,18 @@ function toContractInput(token: PrefetchedTopToken) {
...
 }
 
-// Map of key: ${chain} + ${address} and value: TopToken object.
+// Map of key: ${HistoryDuration} and value: another Map, of key:${chain} + ${address} and value: TopToken object.
 // Acts as a local cache.
-const tokensWithPriceHistoryCache: Record<string, TopToken> = {}
+const tokensWithPriceHistoryCache: Record<string, Record<string, TopToken>> = {}
 
-const checkIfAllTokensCached = (tokens: PrefetchedTopToken[]) => {
+const checkIfAllTokensCached = (duration: string, tokens: PrefetchedTopToken[]) => {
```
JFrankfurt, line 139: iirc there is a HistoryDuration type somewhere that we could use here

lynnshaoyu: yeah although typescript  forces us to predefine the key values and include '%future added value'. I guess this works fine though? (changed it)

src/graphql/data/TopTokens.ts
```diff
@@ -132,18 +132,27 @@ function toContractInput(token: PrefetchedTopToken) {
...
+
+const checkIfAllTokensCached = (duration: HistoryDuration, tokens: PrefetchedTopToken[]) => {
   let everyTokenInCache = true
   const cachedTokens: TopToken[] = []
 
   const checkCache = (token: PrefetchedTopToken) => {
     const tokenCacheKey = !!token ? `${token.chain}${token.address}` : ''
-    if (tokenCacheKey in tokensWithPriceHistoryCache) {
-      cachedTokens.push(tokensWithPriceHistoryCache[tokenCacheKey])
+    if (duration in tokensWithPriceHistoryCache && tokenCacheKey in tokensWithPriceHistoryCache[duration]) {
```
JFrankfurt, line 154: will `duration in tokensWithPriceHistoryCache` always return true now that you've got a default value?

lynnshaoyu: yes true, fixed here and line 215. ty for catching.

cmcewen (approved):

PR #4744, opened 2022-09-28
refactor: split out SearchBarDropdown

vm:
The SearchBar file was pretty large so wanted to split it to simplify.


src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -0,0 +1,291 @@
...
+                headerIcon={<TrendingArrow />}
+                isLoading={trendingCollectionsAreLoading}
+              />
+            )}
+          </Column>
+        )
+
+      setResultsState(currentState)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
cmcewen, line 266: do you know why this is disabled here?

cmcewen (approved):

src/components/NavBar/SearchBarDropdown.tsx
vm: it's missing `collectionSearchResults` and `tokenSearchResults`. i'm gonna make a pr to add those in, generally not a good thing to ignore

PR #4745, opened 2022-09-28
feat: animate in token details line chart

cmcewen:
part 1 of WEB-911


src/components/Charts/AnimatedInLineChart.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { LineChartProps } from './LineChart'
+
+const config = {
+  duration: 800,
+  easing: easeCubicInOut,
+}
+
+// code reference: https://airbnb.io/visx/lineradial
+
+function LineChart<T>({
```
zzmp, line 18: Name this `AnimatedInlineChart` to match the filename.
Rename the filename `AnimatedInlineChart` (inline is one word).

src/components/Charts/AnimatedInLineChart.tsx
```diff
@@ -0,0 +1,90 @@
...
+  marginTop,
+  curve,
+  color,
+  strokeWidth,
+  width,
+  height,
+  children,
+}: LineChartProps<T>) {
+  const lineRef = useRef<SVGPathElement>(null)
+  const [lineLength, setLineLength] = useState<number>(0)
```
zzmp, line 31: nit: `useState`'s should have implied types - setting the generic type is redundant.

src/components/Charts/AnimatedInLineChart.tsx
```diff
@@ -0,0 +1,90 @@
...
+                  fill="none"
+                  stroke={lineColor}
+                />
+                {shouldAnimate && (
+                  <animated.path
+                    d={d}
+                    strokeWidth={strokeWidth}
+                    fill="none"
+                    stroke={lineColor}
+                    strokeDashoffset={spring.frame.interpolate((v: unknown) => (v as number) * lineLength)}
```
zzmp, line 76: does this require an invariant for the typecase?

src/components/Charts/AnimatedInLineChart.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  return (
+    <svg width={width} height={height}>
+      <Group top={marginTop}>
+        <LinePath curve={curve} x={getX} y={getY}>
+          {({ path }) => {
+            const d = path(data) || ''
+            return (
+              <>
+                <animated.path
```
zzmp, line 62: Why can't you use a single path and change the stroke's opacity, dashoffset, and dasharray with animation? I think that would let you reduce state and simplify.

zzmp (changes requested):

zzmp:
Please include instructinos for testing.

src/components/Charts/AnimatedInLineChart.tsx
cmcewen: it is an "animated in line-chart" as opposed to an "animated inline chart" haha

src/components/Charts/AnimatedInLineChart.tsx
cmcewen: you need to measure the length of the line in order to animate along it, so basically you need to do a render first. the code is from here https://airbnb.io/visx/lineradial which references this https://www.visualcinnamon.com/2016/01/animating-dashed-line-d3/

cmcewen:
@zzmp to test, just visit any token details page and you should see the animation on the chart itself

src/components/Charts/AnimatedInLineChart.tsx
cmcewen: spring should always exist and it will always be a number. there is a later version of react-spring which has nicer stuff including types but didn't want to mix concerns

src/components/Charts/AnimatedInLineChart.tsx
zzmp: is this initial render causing the flash on hard refresh, as noted by @infredible ?

zzmp: Can you explain this in a comment, where you currently have the reference?

zzmp (approved):
lgtm, but please explicitly explain the reason for duplicated <animated.path>'s in the initial comment.

PR #4746, opened 2022-09-28
fix: remove ignore exhaustive deps in SearchBarDropdown

vm:
Generally, we should not be ignoring exhaustive deps checks as it can lead to subtle bugs. I believe the component is working still while adding this back in.

The core issue is we should not be storing React elements in state from a useEffect. We should be caching just the data and re-rendering the components as needed. If that is expensive then we should be memoizing the component. I think for now we can do this, but should re-think the architecture a bit potentially.

### DIFF
--- a/src/components/NavBar/SearchBarDropdown.tsx
+++ b/src/components/NavBar/SearchBarDropdown.tsx
@@ -94,38 +94,6 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput, i
   const phase1Flag = useNftFlag()
   const [resultsState, setResultsState] = useState<ReactNode>()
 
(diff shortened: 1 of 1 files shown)


cbachmeier (approved):
Seems to be working as expected.

As per the data storage, that was my original attempt at a solution however I ran into two issues when trying that.

    - You also need to be able to fade out on the default search results dropdown, ie search history, trending tokens, and trending collections, so tracking the old state for that data adds a layer of complexity.
    - When I attempted to useRef for tracking token and collection search state, when interacting with useQuery, the final re-render would never occur. So I would enter a search query, default results will fade, but then unfade without showing the new results. Then when next text was entered, the expected results from the previous query would render.

PR #4751, opened 2022-09-28
fix: left align SearchBar

vm:
There is a task to left align the search bar when unfocused while there is a search query. Looking at this code, it looks like we'll left align it when NFT is released. My opinion is it'd be better to remove this center alignment logic so we can be consistent on the NFT launch, and not need further edge casing.


vm:
This isn't an acceptable design tradeoff after discussion so I'm closing

PR #4756, opened 2022-09-28
chore: updating pool tooltip

JackShort:
* updating tooltip to "This NFT is part of a liquidity pool. Buying this will increase the price of the remaining pooled NFTs."

### DIFF
--- a/src/nft/components/collection/Card.tsx
+++ b/src/nft/components/collection/Card.tsx
@@ -595,7 +595,11 @@ const Suspicious = () => {
 const Pool = () => {
   return (
     <MouseoverTooltip
-      text={<Box fontSize="14">This item is part of an NFT liquidity pool. Price increases as supply decreases.</Box>}
+      text={
+        <Box fontSize="14">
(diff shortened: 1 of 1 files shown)


src/nft/components/collection/Card.tsx
```diff
@@ -595,7 +595,11 @@ const Suspicious = () => {
 const Pool = () => {
   return (
     <MouseoverTooltip
-      text={<Box fontSize="14">This item is part of an NFT liquidity pool. Price increases as supply decreases.</Box>}
+      text={
+        <Box fontSize="14">
```
zzmp, line 599: Should the text prop include markup (`<Box>`)?

JackShort: I think so because it lets us control the box size and the text prop takes in a ReactNode

cbachmeier: We could also swap fontSize '14' to be className `bodySmall` but this is non-blocking

cbachmeier (approved):

PR #4757, opened 2022-09-28
chore: remove extra decimals on cards

JackShort:
* removing extra decimals on cards

### DIFF
--- a/src/nft/utils/currency.ts
+++ b/src/nft/utils/currency.ts
@@ -53,11 +53,10 @@ export const ethNumberStandardFormatter = (
 
   if (amountInDecimals < 0.0001) return `< ${conditionalDollarSign}0.00001`
   if (amountInDecimals < 1) return `${conditionalDollarSign}${amountInDecimals.toFixed(3)}`
-  const formattedPrice = amountInDecimals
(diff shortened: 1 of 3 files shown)


src/nft/utils/currency.ts
```diff
@@ -41,24 +41,26 @@ export const numberToWei = (amount: number) => {
...
-    amountInDecimals
-      .toFixed(2)
-      .toString()
-      .replace(/\B(?=(\d{3})+(?!\d))/g, ',')
-  )
+  const formattedPrice = amountInDecimals
+    .toFixed(2)
+    .toString()
+    .replace(/\B(?=(\d{3})+(?!\d))/g, ',')
+  return conditionalDollarSign + (removeZeroes ? formattedPrice.replace(/.00/g, '') : formattedPrice)
```
cbachmeier, line 60: Do we also want to remove any trailing zeros not just if there's double behind the decimal? For numbers like `3.40` for example

JackShort: design agreed - updated across app besides bag

cbachmeier (approved):
LGTM

PR #4758, opened 2022-09-28
chore: changing last_sale to camelCase

JackShort:
* changing `last_sale` to camelCase

### DIFF
--- a/src/nft/queries/genie/WalletAssetsFetcher.ts
+++ b/src/nft/queries/genie/WalletAssetsFetcher.ts
@@ -38,10 +38,10 @@ export const fetchWalletAssets = async ({
     return {
       ...asset,
       collectionIsVerified: asset.asset_contract.isVerified,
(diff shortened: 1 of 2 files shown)


src/nft/queries/genie/WalletAssetsFetcher.ts
```diff
@@ -38,10 +38,10 @@ export const fetchWalletAssets = async ({
     return {
       ...asset,
       collectionIsVerified: asset.asset_contract.isVerified,
-      lastPrice: asset.last_sale && formatEther(asset.last_sale.total_price),
+      lastPrice: asset.lastSale && formatEther(asset.lastSale.total_price),
```
cbachmeier, line 41: this new param is only on the v3 endpoint so in this fetcher we also need to update the url to be `REACT_APP_GENIE_V3_API_URL`

cbachmeier (changes requested):

PR #4759, opened 2022-09-28
fix: use font-feature-settings in AppWrapper

vm:
Moves font-feature-settings directly into AppWrapper, instead of being located in index.html and various other parts of the App. However, this doesn't seem to fix the slashed zeros yet.

### DIFF
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -56,7 +56,7 @@ const AppWrapper = styled.div<{ redesignFlagEnabled: boolean }>`
   flex-flow: column;
   align-items: flex-start;
   font-feature-settings: ${({ redesignFlagEnabled }) =>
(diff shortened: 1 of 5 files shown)


src/pages/App.tsx
```diff
@@ -50,10 +51,12 @@ const Collection = lazy(() => import('nft/pages/collection'))
 const Sell = lazy(() => import('nft/pages/sell/sell'))
 const Asset = lazy(() => import('nft/pages/asset/Asset'))
 
-const AppWrapper = styled.div`
+const AppWrapper = styled.div<{ redesignFlagEnabled: boolean }>`
   display: flex;
   flex-flow: column;
   align-items: flex-start;
+  font-feature-settings: ${({ redesignFlagEnabled }) =>
+    redesignFlagEnabled ? "'ss01' on, 'ss02' off, 'cv01' on, 'cv03' on" : "'ss01' on, 'ss02' on, 'cv01' on, 'cv03' on"};
```
vm, line 59: @callil to update these values as needed

callil (approved):

PR #4764, opened 2022-09-29
feat: animate in details text

cmcewen:
created a new animations file to separate concerns and also allow for static bundle since the values won't change based on dark mode. visit any token details page to see a (very fast) fade in


cmcewen:
creating PR into main

PR #4765, opened 2022-09-29
feat: add empty wallet state

cbachmeier:
Adds state where user has no NFTs, includes button to go to explore nfts page


src/nft/components/sell/select/SelectPage.tsx
```diff
@@ -246,68 +247,72 @@ export const SelectPage = () => {
...
-              loader={
-                hasNextPage ? (
-                  <Center>
-                    <LoadingSparkle />
-                  </Center>
-                ) : null
-              }
-              dataLength={displayAssets.length}
-              style={{ overflow: 'unset' }}
+      {walletAssets.length === 0 ? (
```
JackShort, line 250: might be better to add a `isLoading` to the wallet assets fetcher and here you can check if its `!isLoading && walletAssets.length === 0` to avoid this initial flashing of no wallet assets

JackShort (approved):
lgtm - one non blocking comment

src/nft/components/sell/select/SelectPage.tsx
cbachmeier: Good callout, we have another story to add a loading state so I will add that check as part of that story.

PR #4766, opened 2022-09-29
fix: shorten SearchBar height

vm:
Screen Shot 2022-09-30 at 1 26 25 PM
Screen Shot 2022-09-30 at 1 26 32 PM


cmcewen (approved):
could you include before/after screenshots?

PR #4770, opened 2022-09-30
chore: upgrade react-spring version

cmcewen:
I ran into a bug developing some animations that requires upgrading. A side benefit is that the first render of modals doesn't hang in dev mode.


cbachmeier (approved):
Tested all NFT components work as expected

PR #4773, opened 2022-09-30
feat: fade in text on details page

cmcewen:
part 2 of WEB-911

### DIFF
--- a/src/theme/animations.ts
+++ b/src/theme/animations.ts
@@ -1,0 +1,28 @@
+import { css, keyframes } from 'styled-components/macro'
+
+const transitions = {
+  duration: {
+    slow: '500ms',
+    medium: '250ms',
+    fast: '125ms',
+  },
+  timing: {
+    ease: 'ease',
+    in: 'ease-in',
+    out: 'ease-out',
+    inOut: 'ease-in-out',
+  },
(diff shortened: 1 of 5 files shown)


src/theme/index.tsx
```diff
@@ -36,6 +36,7 @@ const BREAKPOINTS = {
   xxxl: 1920,
 }
 
+// deprecated - please use the animations.ts file
```
zzmp, line 39: Deprecate this with jsdoc (`/** @deprecated */`; https://jsdoc.app/tags-deprecated.html) so that it will be more apparent in modern editors.

src/theme/animations.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const transitions = {
+  duration: {
+    slow: '500ms',
+    medium: '250ms',
+    fast: '125ms',
+  },
+  timing: {
+    ease: 'ease',
+    in: 'ease-in',
```
zzmp, line 11: Unless we are never going to use another function (like linear), I'd strongly prefer these be named `easeIn`, `easeOut`, `easeInOut`.

zzmp (approved):
some changes requested, but no rereview needed.

PR #4774, opened 2022-09-30
style: filling background color for collection header

JackShort:
* solid background color for collection stats icon


cbachmeier (approved):
LGTM

PR #4777, opened 2022-09-30
fix: memoize tokens in TokenDetails

zzmp:
Fixes an issue where the widget was being cleared because the token was reinstantiated in every render.
Also fixes the same issue with balance checking.


cartcrom (approved):
Thank you for the refactors! everything looks so much better

cmcewen (approved):

PR #4781, opened 2022-10-03
fix: make number of loading rows make more sense when filtering, changing duration etc

lynnshaoyu:
after fix loom: https://www.loom.com/share/a0f5b0bb59f2435081297f2328571b78

Notice: 
- when you search something, it already knows how many rows there will be, so we show that number of loading rows
- when you change duration (with a filter applied), the same number of rows will exist with new duration. so we show that number of loading rows.

### DIFF
--- a/src/components/Tokens/TokenTable/TokenTable.tsx
+++ b/src/components/Tokens/TokenTable/TokenTable.tsx
@@ -95,7 +95,7 @@ export default function TokenTable() {
 
   /* loading and error state */
   if (loading && (!tokens || tokens?.length === 0)) {
-    return <LoadingTokenTable rowCount={PAGE_SIZE} />
+    return <LoadingTokenTable rowCount={Math.min(PAGE_SIZE, maxFetchable)} />
   } else {
     if (error || !tokens) {
       return (
--- a/src/components/Tokens/TokenTable/TokenTable.tsx
+++ b/src/components/Tokens/TokenTable/TokenTable.tsx
@@ -95,7 +95,6 @@ export default function TokenTable() {
 
   /* loading and error state */
   if (loading && (!tokens || tokens?.length === 0)) {
-    console.log('loading count', loadingRowsCount)
     return <LoadingTokenTable rowCount={loadingRowsCount} />
   } else {
     if (error || !tokens) {
--- a/src/components/Tokens/TokenTable/TokenTable.tsx
+++ b/src/components/Tokens/TokenTable/TokenTable.tsx
@@ -75,7 +75,7 @@ export default function TokenTable() {
 
   // TODO: consider moving prefetched call into app.tsx and passing it here, use a preloaded call & updated on interval every 60s
   const chainName = validateUrlChainParam(useParams<{ chainName?: string }>().chainName)
-  const { error, loading, tokens, hasMore, loadMoreTokens, tokenListLength, loadingRowsCount } = useTopTokens(chainName)
+  const { error, loading, tokens, hasMore, loadMoreTokens, loadingRowsCount } = useTopTokens(chainName)
   const showMoreLoadingRows = Boolean(loading && hasMore)
 
   const observer = useRef<IntersectionObserver>()
@@ -126,7 +126,7 @@ export default function TokenTable() {
                     <LoadedRow
                       key={token?.address}
                       tokenListIndex={index}
-                      tokenListLength={tokenListLength}
+                      tokenListLength={tokens.length}
                       token={token}
                       ref={index + 1 === tokens.length ? lastTokenRef : undefined}
                     />
--- a/src/components/Tokens/TokenTable/TokenTable.tsx
+++ b/src/components/Tokens/TokenTable/TokenTable.tsx
@@ -75,7 +75,7 @@ export default function TokenTable() {
 
   // TODO: consider moving prefetched call into app.tsx and passing it here, use a preloaded call & updated on interval every 60s
   const chainName = validateUrlChainParam(useParams<{ chainName?: string }>().chainName)
-  const { error, loading, tokens, hasMore, loadMoreTokens, maxFetchable } = useTopTokens(chainName)
+  const { error, loading, tokens, hasMore, loadMoreTokens, tokenListLength, loadingRowsCount } = useTopTokens(chainName)
   const showMoreLoadingRows = Boolean(loading && hasMore)
 
   const observer = useRef<IntersectionObserver>()
@@ -95,7 +95,8 @@ export default function TokenTable() {
 
   /* loading and error state */
   if (loading && (!tokens || tokens?.length === 0)) {
-    return <LoadingTokenTable rowCount={Math.min(PAGE_SIZE, maxFetchable)} />
+    console.log('loading count', loadingRowsCount)
+    return <LoadingTokenTable rowCount={loadingRowsCount} />
   } else {
     if (error || !tokens) {
       return (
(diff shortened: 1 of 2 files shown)


src/graphql/data/TopTokens.ts
```diff
@@ -176,11 +176,11 @@ export function useTopTokens(chain: Chain): UseTopTokensReturnValue {
   const [tokens, setTokens] = useState<TopToken[]>()
   const [page, setPage] = useState(0)
   const [error, setError] = useState<Error | undefined>()
-  const [prefetchedData, setPrefetchedData] = useState<PrefetchedTopToken[]>([])
+  const [prefetchedData, setPrefetchedData] = useState<PrefetchedTopToken[]>()
   const prefetchedSelectedTokensWithoutPriceHistory = useFilteredTokens(useSortedTokens(prefetchedData))
   const maxFetchable = useMemo(
-    () => prefetchedSelectedTokensWithoutPriceHistory.length,
-    [prefetchedSelectedTokensWithoutPriceHistory]
+    () => (!!prefetchedData ? prefetchedSelectedTokensWithoutPriceHistory.length : 100),
```
zzmp, line 182: Does this need the `!!`?

src/graphql/data/TopTokens.ts
```diff
@@ -59,7 +59,7 @@ export enum TokenSortMethod {
 
 export type PrefetchedTopToken = NonNullable<TopTokens100Query['response']['topTokens']>[number]
 
-function useSortedTokens(tokens: TopTokens100Query['response']['topTokens']) {
+function useSortedTokens(tokens: TopTokens100Query['response']['topTokens'] | undefined) {
```
zzmp, line 62: Why are you changing this type?

lynnshaoyu: cuz its now defaulting to undefined (see line 179) so we can make the check on line 182

src/graphql/data/TopTokens.ts
```diff
@@ -176,11 +176,11 @@ export function useTopTokens(chain: Chain): UseTopTokensReturnValue {
   const [tokens, setTokens] = useState<TopToken[]>()
   const [page, setPage] = useState(0)
   const [error, setError] = useState<Error | undefined>()
-  const [prefetchedData, setPrefetchedData] = useState<PrefetchedTopToken[]>([])
+  const [prefetchedData, setPrefetchedData] = useState<PrefetchedTopToken[]>()
   const prefetchedSelectedTokensWithoutPriceHistory = useFilteredTokens(useSortedTokens(prefetchedData))
   const maxFetchable = useMemo(
```
zzmp, line 181: Can this be renamed to `tokenListSize`?

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -95,7 +95,7 @@ export default function TokenTable() {
 
   /* loading and error state */
   if (loading && (!tokens || tokens?.length === 0)) {
-    return <LoadingTokenTable rowCount={PAGE_SIZE} />
+    return <LoadingTokenTable rowCount={Math.min(PAGE_SIZE, maxFetchable)} />
```
zzmp, line 98: It is unclear that TokenDataContainer is capped at PAGE_SIZE, so this limitation looks ad-hoc. Can you point out where that is done? Will tokens.length be maxFetchable, or will it be Math.min(PAGE_SIZE, maxFetchable)? Can maxFetchable be capped to PAGE_SIZE, so that this logic can be silo'ed to useTopTokens and not need to be duplicated on this line?

src/graphql/data/TopTokens.ts
lynnshaoyu: true not needed, removed

src/graphql/data/TopTokens.ts
```diff
@@ -176,11 +176,11 @@ export function useTopTokens(chain: Chain): UseTopTokensReturnValue {
...
-    () => prefetchedSelectedTokensWithoutPriceHistory.length,
-    [prefetchedSelectedTokensWithoutPriceHistory]
+    () => (prefetchedData ? prefetchedSelectedTokensWithoutPriceHistory.length : 100),
```
zzmp, line 182: Should 100 be PAGE_SIZE?

lynnshaoyu: yes done

src/graphql/data/TopTokens.ts
lynnshaoyu: yes, changed to tokenListLength because thats what we use in TokenTable.tsx. see line 130

src/components/Tokens/TokenTable/TokenTable.tsx
lynnshaoyu: I think I get what you mean, the ad-hoc nature makes it unclear why we are doing this. 

I split out this logic into "loadingRowsCount" useMemo fn in TopToken.ts and added a comment. hope it makes it more clear now?

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -95,7 +95,8 @@ export default function TokenTable() {
 
   /* loading and error state */
   if (loading && (!tokens || tokens?.length === 0)) {
-    return <LoadingTokenTable rowCount={PAGE_SIZE} />
+    console.log('loading count', loadingRowsCount)
```
zzmp, line 98: rm console.log

src/graphql/data/TopTokens.ts
```diff
@@ -167,7 +167,8 @@ interface UseTopTokensReturnValue {
   tokens: TopToken[] | undefined
   hasMore: boolean
   loadMoreTokens: () => void
-  maxFetchable: number
+  tokenListLength: number
+  loadingRowsCount: number
```
zzmp, line 171: Can you omit `loadingRowsCount` and just use `tokenListLength`?

src/graphql/data/TopTokens.ts
```diff
@@ -176,12 +177,18 @@ export function useTopTokens(chain: Chain): UseTopTokensReturnValue {
...
+  const [prefetchedData, setPrefetchedData] = useState<PrefetchedTopToken[]>()
   const prefetchedSelectedTokensWithoutPriceHistory = useFilteredTokens(useSortedTokens(prefetchedData))
-  const maxFetchable = useMemo(
+  const tokenListLength = useMemo(
     () => prefetchedSelectedTokensWithoutPriceHistory.length,
     [prefetchedSelectedTokensWithoutPriceHistory]
   )
+  // loadingRowsCount defaults to PAGE_SIZE when no prefetchedData is available yet because the initial load
+  // count will always be PAGE_SIZE.
+  const loadingRowsCount = useMemo(
```
zzmp, line 188: Return this value as `tokenListLength`

src/graphql/data/TopTokens.ts
zzmp: I think splitting it into two vars adds unnecessary complexity.

zzmp (changes requested):

src/graphql/data/TopTokens.ts
lynnshaoyu: can't, see line 130 on TokenTable. there would be information loss

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -75,7 +75,7 @@ export default function TokenTable() {
 
   // TODO: consider moving prefetched call into app.tsx and passing it here, use a preloaded call & updated on interval every 60s
   const chainName = validateUrlChainParam(useParams<{ chainName?: string }>().chainName)
-  const { error, loading, tokens, hasMore, loadMoreTokens, maxFetchable } = useTopTokens(chainName)
+  const { error, loading, tokens, hasMore, loadMoreTokens, loadingRowsCount } = useTopTokens(chainName)
```
zzmp, line 78: nit: `loadingRowCount` (not `Rows`) to match the prop that it is being passed in to.

PR #4784, opened 2022-10-04
fix: fixing issue with decimal in price range

aballerr:
1. If you try and put a decimal in the price range then nothing happens.  This fixes that issue


src/nft/components/collection/PriceRange.tsx
```diff
@@ -2,7 +2,6 @@ import { Row } from 'nft/components/Flex'
 import { NumericInput } from 'nft/components/layout/Input'
```
aballerr, line 2: So, I looked at this code and it does verify if the value is a number here.  If it isn't a number, the value is set to an empty string ```''```.  I looked at the previous genie code and I don't actually know how it worked properly.  We call a ```parseFloat``` on the value, which when called on the value "0." should always just be a zero.  Currently, we don't actually do any math with this value, so it is fine as a string.

cbachmeier (approved):
LGTM

PR #4785, opened 2022-10-04
feat: search attempts route to current chain

cartcrom:


src/components/NavBar/SuggestionRow.tsx
```diff
@@ -123,7 +134,9 @@ export const TokenRow = ({ token, isHovered, setHoveredIndex, toggleOpen, traceE
     traceEvent()
   }, [addToSearchHistory, toggleOpen, token, traceEvent])
 
-  const tokenDetailsPath = getTokenDetailsURL(token.address, undefined, token.chainId)
+  const [bridgedAddress, bridgedChain, L2Icon] = useBridgedAddress(token)
+  console.log(bridgedAddress)
```
cbachmeier, line 138: remove this logging

src/components/NavBar/SuggestionRow.tsx
```diff
@@ -100,6 +102,15 @@ export const CollectionRow = ({
   )
 }
 
+function useBridgedAddress(token: FungibleToken): [string | undefined, number | undefined, string | undefined] {
```
cbachmeier, line 105: I think we should move this helper fn to the same util files as `getTokenDetailsURL`

cbachmeier (changes requested):

callil (approved):
LGTM - adding the icon helps immensely.

src/components/NavBar/SuggestionRow.tsx
cartcrom: I'm gonna do some refactoring of where that util lives later. Right now, that util is in the graphql section of the code, it'd probably be weird to have this FungibleToken data structure used there too since it's from a different endpoint

cbachmeier (approved):
LGTM

PR #4787, opened 2022-10-04
feat: explore category info icons

cartcrom:
Added info icons to TVL and Volume on explore and token details
image

### DIFF
--- a/src/components/Tokens/TokenDetails/InfoTip.tsx
+++ b/src/components/Tokens/TokenDetails/InfoTip.tsx
@@ -7,7 +7,7 @@ const InfoTipContainer = styled.div`
   display: flex;
   position: relative;
   align-items: center;
-  cursor: pointer;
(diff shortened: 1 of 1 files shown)


src/components/Tokens/TokenDetails/InfoTip.tsx
```diff
@@ -0,0 +1,34 @@
+import Tooltip from 'components/Tooltip'
+import { ReactNode, useCallback, useState } from 'react'
+import { Info } from 'react-feather'
+import styled from 'styled-components/macro'
+
+const InfoTipContainer = styled.div`
+  display: flex;
+  position: relative;
+  align-items: center;
+  cursor: pointer;
```
JFrankfurt, line 10: nit: (don't think we do this elsewhere) - since it's not clickable, we prob should do `cursor: help` instead of `pointer`

PR #4788, opened 2022-10-04
fix(IconButton): improves code quality and fixes a console error re: incorrect data- attribute syntax

JFrankfurt:
image

### DIFF
--- a/src/components/WalletDropdown/IconButton.tsx
+++ b/src/components/WalletDropdown/IconButton.tsx
@@ -58,31 +58,28 @@ const IconWrapper = styled.span`
   width: 16px;
   height: 16px;
 `
-
-interface IconButtonProps {
-  text: React.ReactNode
+interface BaseProps {
   Icon: Icon
-  onClick?: () => void
-  href?: string
-  dataTestId?: string
+}
+interface IconLinkProps extends React.ComponentPropsWithoutRef<'a'>, BaseProps {}
+interface IconButtonProps extends React.ComponentPropsWithoutRef<'button'>, BaseProps {}
+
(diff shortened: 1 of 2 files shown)


src/components/WalletDropdown/IconButton.tsx
```diff
@@ -58,31 +58,28 @@ const IconWrapper = styled.span`
...
-        <IconHoverText>{text}</IconHoverText>
-      </IconWrapper>
-    </IconBlockButton>
-  )
+const IconBlock = (props: React.ComponentPropsWithoutRef<'a' | 'button'>) => {
+  if ('href' in props) {
+    return <IconBlockLink {...props} />
+  }
+  // ignoring 'button' 'type' conflict between React and styled-components
+  // @ts-ignore
```
vm, line 72: i don't think we want to ignore here. can we just deconstruct the props and send the right ones?

aballerr (approved):

src/components/WalletDropdown/IconButton.tsx
JFrankfurt: Since it's currently passing through all valid `button` attributes, that list would be like 80 entries long. This is just a way of saying "all things that are valid on a button element are valid here" so future devs don't need to go explicitly enumerate all these: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attributes

the issue is that styled-components wants `"submit" | "reset" | "button" | undefined` for `type`, but React is more permissive w/ `string | undefined`. 

It looks like `styled-components`'s types are actually [technically more accurate](https://html.spec.whatwg.org/multipage/form-elements.html#attr-button-type).

PR #4793, opened 2022-10-04
feat: added widget skeleton to loading state

cartcrom:
added new widget skeleton component to loading state

### DIFF
--- a/src/components/Widget/index.tsx
+++ b/src/components/Widget/index.tsx
@@ -19,8 +19,7 @@ export interface WidgetProps {
 
 export default function Widget({ defaultToken, onReviewSwapClick }: WidgetProps) {
   const locale = useActiveLocale()
-  const darkMode = useIsDarkMode()
-  const theme = darkMode ? DARK_THEME : LIGHT_THEME
+  const theme = useIsDarkMode() ? DARK_THEME : LIGHT_THEME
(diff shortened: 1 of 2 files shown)


src/components/Widget/index.tsx
```diff
@@ -49,3 +49,9 @@ export default function Widget({ defaultToken, onReviewSwapClick }: WidgetProps)
     </>
   )
 }
+
+export function WidgetSkeleton() {
+  const darkMode = useIsDarkMode()
+  const theme = useMemo(() => (darkMode ? DARK_THEME : LIGHT_THEME), [darkMode])
+  return <SwapWidgetSkeleton theme={theme} width={WIDGET_WIDTH} />
```
zzmp, line 56: `DARK_THEME` and `LIGHT_THEME` are referentially stable, so you do not need to memoize `theme`.

    return <SwapWidgetSkeleton theme={darkMode ? DARK_THEME : LIGHT_THEME} widget={WIDGET_WIDTH} />

zzmp (changes requested):

src/components/Widget/index.tsx
cartcrom: I was copying over what was in <Widget/>, I'll remove from both

zzmp (approved):

PR #4804, opened 2022-10-05
fix: Adjust verified icons to accentAction

callil:
Fixes: https://uniswaplabs.atlassian.net/browse/WEB-1443?filter=10071

Verified badges now use accentAction instead of a the fixed blue.

Before:
Screen Shot 2022-10-05 at 11 52 47 AM

After:
Screen Shot 2022-10-05 at 11 54 30 AM

Does not touch NFT verified icon.


src/components/TokenSafety/TokenSafetyIcon.tsx
```diff
@@ -11,6 +12,7 @@ const VerifiedContainer = styled.div`
 export const VerifiedIcon = styled(Verified)<{ size?: string }>`
   width: ${({ size }) => size ?? '1em'};
   height: ${({ size }) => size ?? '1em'};
+  color: ${colors.magentaVibrant};
```
vm, line 15: you should use `theme.accentAction`

vm (approved):

PR #4808, opened 2022-10-05
perf: improve token details query performance

cartcrom:
TokenDetails now uses one hook to fetch priceHistory data and token data in tandem for faster render time. This also prevents the page from ever re-suspending/rendering upon switching time durations


lynnshaoyu:
build / deployment error?

lynnshaoyu (approved):

PR #4809, opened 2022-10-05
feat: profile listing sidebar

cbachmeier:
- Use the styling from Bag for the listing sidebar on profile page
  - Move header and content into their own files
  - Update header, empty state, and content if on profile page 
- deprecate listing tag
- re-arrange listing modal files to be in the profile folder of Bag
- update lineheight for subhead to be 24


src/nft/components/bag/Bag.tsx
```diff
@@ -322,57 +287,52 @@ const Bag = () => {
...
-              <Column display={priceChangedAssets.length > 0 || unavailableAssets.length > 0 ? 'flex' : 'none'}>
-                {unavailableAssets.length > 0 && (
-                  <UnavailableAssetsHeaderRow
-                    assets={unavailableAssets}
-                    usdPrice={fetchedPriceData}
-                    clearUnavailableAssets={() => setItemsInBag(availableItems)}
-                    didOpenUnavailableAssets={didOpenUnavailableAssets}
-                    setDidOpenUnavailableAssets={setDidOpenUnavailableAssets}
-                    isMobile={isMobile}
+            {!(isProfilePage && profilePageState === ProfilePageStateType.LISTING) ? (
```
JackShort, line 290: maybe move this to a variable named `isOnProfileListingPage` just so logically its easier to decipher when being read

JackShort (approved):
works pretty well - we discussed future refactorings over slack but for now it is good

src/nft/components/bag/Bag.tsx
cbachmeier: I personally prefer this setup as this var would only be use in 1 spot and I think this is very readable since there's only the two conditions: "I'm on the profile page and currently listing"

PR #4813, opened 2022-10-05
fix: import widget font

cmcewen:
WEB-1444

### DIFF
--- a/src/components/Widget/index.tsx
+++ b/src/components/Widget/index.tsx
@@ -1,3 +1,6 @@
+// eslint-disable-next-line
+import '@uniswap/widgets/dist/fonts.css'
+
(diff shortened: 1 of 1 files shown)


vm (approved):

src/components/Widget/index.tsx
```diff
@@ -1,3 +1,6 @@
+// eslint-disable-next-line
```
zzmp, line 1: This should include a comment preceding the eslint directive explaining why it is necessary.
Less importantly, it should also explicitly disable only the necessary lint rule.

```
// Import fonts.css for the side-effect of loading fonts for @uniswap/widgets.
// eslint-disable-next-line i-don't-know-the-rule
```

PR #4818, opened 2022-10-06
fix: refactors the way chainId is accessed in some places

JFrankfurt:
this is a code organization/QoL patch

`useGlobalChainName` is currently exported from `src/graphql/data/utils` and accesses the connected chainId from redux (non-standard).

the `useGlobalChainName` doesn't need to be a hook and is more useful if it accepts chainId as a parameter. 

I've renamed it to `chainIdToBackendName`, which better explains the purpose of the util in the graphql context and allows more flexible usage.


zzmp:
Can you include the rationale for this refactor in the PR description?

cartcrom (approved):

PR #4826, opened 2022-10-06
test: use goerli

zzmp:
Skips the tests that require the Goerli test environment to work properly for now, which we can add back later.


vm (approved):
Skips the tests that require the Goerli test environment to work properly for now, which we can add back later.

PR #4827, opened 2022-10-06
feat: lazy load sparklines

cartcrom:


src/components/Charts/SparklineChart.tsx
```diff
@@ -51,5 +73,3 @@ function SparklineChart({ width, height, tokenData, pricePercentChange, timePeri
     />
   )
 }
-
-export default React.memo(SparklineChart)
```
cmcewen, line 55: y no memo?

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -75,73 +79,51 @@ export function LoadingTokenTable({ rowCount }: { rowCount?: number }) {
...
-        </>
-      )
-    }
+    return (
+      <GridContainer>
+        <HeaderRow />
+        <TokenDataContainer>
+          {tokens.map(
+            (token, index) =>
+              token && (
```
cmcewen, line 115: where would token be null/undefined here?

src/graphql/data/TopTokens.ts
```diff
@@ -121,240 +128,43 @@ function useFilteredTokens(tokens: PrefetchedTopToken[]) {
...
-  const hasMore = !tokens || tokens.length < prefetchedSelectedTokensWithoutPriceHistory.length
   const environment = useRelayEnvironment()
+  const [sparklines, setSparklines] = useState<SparklineMap>({})
+  useMemo(() => {
+    fetchQuery<TopTokensSparklineQuery>(environment, tokenSparklineQuery, {
+      duration,
+      chain,
+    }).subscribe({
+      next(data) {
+        const map: SparklineMap = {}
```
cmcewen, line 149: is there ever a case where the data will be the same and we trigger a re-render without needing to?

src/components/Charts/SparklineChart.tsx
cartcrom: Didn't mean to remove, can put back in

src/components/Tokens/TokenTable/TokenTable.tsx
cartcrom: Relay makes everything nullable, and the backend might return null if there's no data available

src/graphql/data/TopTokens.ts
cartcrom: The query cache works here just fine, it should only rerequest data if it is stale

cmcewen (approved):

src/components/Charts/SparklineChart.tsx
```diff
@@ -1,30 +1,53 @@
...
   const theme = useTheme()
   // for sparkline
-  const pricePoints = filterPrices(tokenData?.market?.priceHistory) ?? []
-  const hasData = pricePoints.length !== 0
-  const startingPrice = hasData ? pricePoints[0] : DATA_EMPTY
-  const endingPrice = hasData ? pricePoints[pricePoints.length - 1] : DATA_EMPTY
+  const pricePoints = tokenData?.address ? sparklineMap[tokenData.address] : null
+
+  // Don't display if there's one or less pricepoints
+  if (!pricePoints || pricePoints.length <= 1) {
```
vm, line 41: should this be an empty state rather than loading state?

vm (commented):
should we log the row number backwards when it's descending?

src/components/Charts/SparklineChart.tsx
cmcewen: the failure case is that there is an error with fetching the data right?

awruble:
yes - I think we should log the row number backwards when it's descending

PR #4828, opened 2022-10-06
fix: issues with NFT listing

cbachmeier:
a combination of refactors and updates to existing 3rd party apis have left out ability to list broken

- [x] List on OS
- [x] List on LR
- [x] List on X2Y2
  - [x] new listings
  - Updating listings for x2y2 will be handled in a subsequent PR, have contact x2y2 team to help fix


src/nft/queries/genie/WalletAssetsFetcher.ts
```diff
@@ -34,11 +34,12 @@ export const fetchWalletAssets = async ({
     },
   })
   const data = await r.json()
+  console.log(data.data.assets)
```
gbugyis, line 37: remove the console log

gbugyis (changes requested):
I'm seeing a couple warnings in the browser console from missing key props. Other than that, and the comment above the flow is behaving as expected for me.

gbugyis (approved):

PR #4829, opened 2022-10-06
fix(token-details): nits from design re: headers, spacing, layout

JFrankfurt:
https://uniswaplabs.atlassian.net/browse/WEB-1456

image


src/components/Tokens/TokenDetails/About.tsx
```diff
@@ -24,8 +25,8 @@ const TokenDescriptionContainer = styled.div`
 const TruncateDescriptionButton = styled.div`
   color: ${({ theme }) => theme.textSecondary};
   font-weight: 400;
-  font-size: 14px;
-  padding-top: 14px;
+  font-size: 0.85em;
```
cmcewen, line 28: i do want to settle what we should do here with units but cool with this for now

cmcewen (approved):

src/components/Tokens/TokenDetails/About.tsx
JFrankfurt: haha, thought of you as I wrote this

infredible:
Some issues:

### Swap button is gone in mobile sizes
**before**
CleanShot 2022-10-06 at 16 25 05@2x
**now**
CleanShot 2022-10-06 at 16 24 51@2x


### Links and contract address subheads
- should be 14px
- should be 500 weight

image

### Spacing
CleanShot 2022-10-06 at 16 16 41@2x

PR #4833, opened 2022-10-07
chore: remove phase0 flags

vm:
.


cypress/e2e/wallet.test.ts
```diff
@@ -1,31 +0,0 @@
-import { TEST_ADDRESS_NEVER_USE_SHORTENED } from '../support/ethereum'
```
vm, line 1: this was replaced by `wallet-dropdown.test.ts`

cypress/e2e/lists.test.ts
```diff
@@ -1,11 +0,0 @@
-describe('Lists', () => {
```
vm, line 1: token list is no longer editable by the user in the interface

cmcewen (approved):

PR #4836, opened 2022-10-07
fix(token-details): make breakpoints more consistent and use only theme.breakpoint values

JFrankfurt:
There's some inconsistency with the nft codebase's use of breakpoints that don't match those in `src/theme/index.ts`. I've tried to get as close to the figma's spec as I can without touching the nft team's navbar code, as I don't feel confident adjusting their work there w/o a significant amount of debugging/communications work. 

image

image

image

image

image

### DIFF
--- a/src/components/Tokens/TokenDetails/BalanceSummary.tsx
+++ b/src/components/Tokens/TokenDetails/BalanceSummary.tsx
@@ -91,6 +91,7 @@ export default function BalanceSummary({ tokenAmount, nativeCurrencyAmount, isNa
     tokenAmount.currency.address.toLowerCase() === nativeCurrencyAmount.currency.wrapped.address.toLowerCase()
 
   if (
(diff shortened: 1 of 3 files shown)


src/components/Tokens/TokenDetails/BalanceSummary.tsx
```diff
@@ -130,8 +131,8 @@ export default function BalanceSummary({ tokenAmount, nativeCurrencyAmount, isNa
     <BalancesCard>
       <TotalBalanceSection>
         <Trans>Your balance</Trans>
-        {currencies.map((props) => (
-          <BalanceRow {...props} key={props.currency.wrapped.address} />
+        {currencies.map((props, i) => (
+          <BalanceRow {...props} key={props.currency.wrapped.address + i} />
```
cmcewen, line 135: when do we have things with the same address?

cmcewen (approved):

PR #4837, opened 2022-10-07
style: updated crosshair light/dark color and stroke width

infredible:


infredible:
before/after
CleanShot 2022-10-07 at 08 48 46@2x
CleanShot 2022-10-07 at 08 48 52@2x

cmcewen (approved):

PR #4839, opened 2022-10-07
fix: animation on mount wasn't happening

cmcewen:
The chart also doesn't need to have the children passed to it so I moved them around


JFrankfurt:
It looks like you could take this further and remove the children prop setup from `LineChart` itself. The only usage is in the sparkline component, and the `children` prop is used w/ the type is only present in `AnimatedInLineChart`. Could drop `children` from `LineChart` and do `interface AnimatedInLIneChart extends PropsWithChildren<LineChartProps> {}` or similar

JFrankfurt (approved):

cmcewen:
@JFrankfurt yup agreed just wanted to keep this diff small but i intend to do that

PR #4849, opened 2022-10-10
fix: aligns token ranking column

vm:


vm:
Screen Shot 2022-10-10 at 12 20 15 PM

PR #4853, opened 2022-10-10
chore: use the same font as the main app in the widget

cmcewen:
Re-use app font for widgets theming, for consistency and to prevent redundant bandwidth usage.


src/theme/widget.ts
```diff
@@ -1,5 +1,12 @@
 import { colorsDark, colorsLight } from './colors'
 
+const fonts = {
+  fontFamily: {
+    font: "'Inter custom', sans-serif;",
```
zzmp, line 5: Should this be `"'Inter custom';"`, without `sans-serif`? Based on https://github.com/Uniswap/interface/blob/main/public/index.html#L40.

src/theme/widget.ts
```diff
@@ -1,5 +1,12 @@
...
+    variable: "'Inter custom', sans-serif;",
```
zzmp, line 6: Omit `;` from both properties. Based on https://github.com/Uniswap/widgets/blob/main/src/components/Widget.tsx#L47-L55.

zzmp (changes requested):

src/theme/widget.ts
```diff
@@ -1,5 +1,12 @@
 import { colorsDark, colorsLight } from './colors'
 
+const fonts = {
+  fontFamily: {
```
zzmp, line 4: If they are the same, this should be simplified to just use a string:

    const fontFamily = 'Inter custom'

The widget will check for `typeof fonts === 'string'` and use the value directly.

zzmp (approved):
You can simplify with the requested change, but no re-review required.

PR #4854, opened 2022-10-10
fix: keep search history synced with trending

zzmp:
Syncs searchHistory with trendingTokenResults, which were otherwise (potentially) fetching more updated results than the searchHistory cache.

WEB-1508

### DIFF
--- a/src/nft/hooks/useSearchHistory.ts
+++ b/src/nft/hooks/useSearchHistory.ts
@@ -5,6 +5,7 @@ import { devtools, persist } from 'zustand/middleware'
 interface SearchHistoryProps {
   history: (FungibleToken | GenieCollection)[]
   addItem: (item: FungibleToken | GenieCollection) => void
+  updateItem: (update: FungibleToken | GenieCollection) => void
 }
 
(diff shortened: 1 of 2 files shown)


src/nft/hooks/useSearchHistory.ts
```diff
@@ -18,6 +19,16 @@ export const useSearchHistory = create<SearchHistoryProps>()(
           return { history: historyCopy }
         })
       },
+      updateItem: (update: FungibleToken | GenieCollection) => {
```
cmcewen, line 22: is it worth ensuring that this item has more recent data before overwriting? would we ever have the opposite case happen?

zzmp: I don't think that is possible, because CollectionRow (which calls addItem) is populated by the same query that would be calling updateItem. It may be possible in the future, if this codepath grows and/or is reused.

However, even if it were possible, I think it'd be preferable to always update searchHistory when requested, so that if multiple queries are being used, searchHistory is forced to be in sync, and there is no disparity possible.

cmcewen (approved):

PR #4855, opened 2022-10-11
fix: map wrapped tokens to native tokens for top/trending

zzmp:
- Maps trending tokens from wrapped to native token
- Maps sparklined top tokens from wrapped to native token
- Closes out subscriptions and memoizes top tokens for performance
- Updates useCurrencyLogoURIs to be usable with token data from queries

WEB-1504


awruble:
Preview looks good except clicking on ETH in both Search and Token Explore should take you to the ETH token detail page, and both are still taking me to WETH.

zzmp:
> Preview looks good except clicking on ETH from both Search and Token Explore should take you to the ETH token detail page, and both are still taking me to WETH.

Fixed! good catch

cmcewen:
Screen Shot 2022-10-10 at 10 02 34 PM

we lost the ETH logo 😢

src/nft/queries/genie/TrendingTokensFetcher.ts
```diff
@@ -1,6 +1,18 @@
+import { nativeOnChain, WRAPPED_NATIVE_CURRENCY } from 'constants/tokens'
+
 import { FungibleToken } from '../../types'
 
+function replaceWrappedWithNative(token: FungibleToken) {
```
cmcewen, line 5: i can't quite articulate the way to combine this and the other replace wrapped function so i'm okay leaving them for now

src/graphql/data/TopTokens.ts
```diff
@@ -135,16 +130,31 @@ interface UseTopTokensReturnValue {
   sparklines: SparklineMap
 }
 
+function replaceWrappedWithNative<T extends { address: string | null } | null>(chainId: number, token: T): T {
```
JFrankfurt, line 133: I feel like this util should be somewhere else. I get the sense from the (admittedly sparse) Relay docs that this file is intended to be scoped specifically to queries and maaaybe their primary access hooks. Utils related to the data from those queries maybe should go in `src/graphql/utils/<Query>`?

src/graphql/data/TopTokens.ts
```diff
@@ -135,16 +130,31 @@ interface UseTopTokensReturnValue {
...
-    }).subscribe({
+  const query = useMemo(
+    () =>
+      fetchQuery<TopTokensSparklineQuery>(environment, tokenSparklineQuery, { duration, chain }).map((data) => ({
+        topTokens: data.topTokens?.map((token) => replaceWrappedWithNative(chainId, token)),
+      })),
+    [chain, chainId, duration, environment]
+  )
+  useEffect(() => {
+    const subscription = query.subscribe({
```
JFrankfurt, line 157: this memo -> useEffect link feels off

They're tightly coupled, but the code doesn't really make that clear until you take a sec to read it. Would putting the query construction currently in useMemo in the useEffect below have any negative effect? All else equal I'd prefer the query stays with its usage.

src/graphql/data/TopTokens.ts
```diff
@@ -153,18 +163,19 @@ export function useTopTokens(chain: Chain): UseTopTokensReturnValue {
         setSparklines(map)
       },
     })
-  }, [chain, duration, environment])
+    return () => subscription.unsubscribe()
```
JFrankfurt, line 166: ty

src/graphql/data/TopTokens.ts
cartcrom: I wrote the original useMemo here. Since useLazyLoadQuery suspends, the first render triggering a useEffect wouldn't run until that query finishes. For perf, I put it in a memo so the two queries run in parallel. I agree that it looks wonky tho

src/nft/queries/genie/TrendingTokensFetcher.ts
zzmp: They're in wildly different parts of the codebase, otherwise I'd throw it in a shared utils.

zzmp: Moved to src/utils/unwrapToken.ts.

src/graphql/data/TopTokens.ts
zzmp: I agree - this is an artifact of refactoring.

src/lib/hooks/useCurrencyLogoURIs.ts
```diff
@@ -54,15 +53,27 @@ export function getTokenLogoURI(address: string, chainId: SupportedChainId = Sup
...
+      }
+    | null
+    | undefined
+): string[] {
+  const locations = useHttpLocations(currency?.logoURI)
   return useMemo(() => {
     const logoURIs = [...locations]
     if (currency) {
-      if (currency.isNative) {
+      if (currency.isNative || currency.address === 'NATIVE') {
```
cmcewen, line 72: think there is a const in analytics/constants - we should ideally use it everywhere to prevent misspellings

src/nft/queries/genie/TrendingTokensFetcher.ts
```diff
@@ -1,6 +1,9 @@
+import { unwrapToken } from 'graphql/data/util'
+
 import { FungibleToken } from '../../types'
 
 export const fetchTrendingTokens = async (numTokens?: number): Promise<FungibleToken[]> => {
+  // TODO: WETH->ETH
```
cmcewen, line 6: this todo is not very descriptive tbh

cmcewen (approved):

src/nft/queries/genie/TrendingTokensFetcher.ts
zzmp: That was a personal TODO so I could easily find the relevant codepaths :)

PR #4858, opened 2022-10-11
chore: remove unused TokensLoading

vm:
Looks this isn't necessary as we do the loading states inline now.


zzmp (approved):
always down for removing code

PR #4866, opened 2022-10-11
fix: use widget network connection

zzmp:


JFrankfurt:
I'm not sure this is the right approach.
see:
```
 * Fallback JSON-RPC endpoints.
 * These are used if the integrator does not provide an endpoint, or if the endpoint does not work.
```

We ideally want the integrator to provide a network connector that we can use instead of using the fallbacks.

JFrankfurt (approved):

PR #4869, opened 2022-10-11
feat: widget analytics

lynnshaoyu:
has everything except SWAP_QUOTE_RECEIVED event, because widget handler logic (onInitialSwapQuote) is broken...we are repeatedly fetching the swap quote, not just the first one as intended. need to fix widget then add back interface logging logic. to be done in follow up PR, i left a TODO for myself in the code. 

i verified everything else is being logged correctly on my local.


src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -18,11 +18,11 @@ import TransactionConfirmationModal, {
 import SwapModalFooter from './SwapModalFooter'
 import SwapModalHeader from './SwapModalHeader'
 
-const formatAnalyticsEventProperties = ({
+export const formatSwapSignedAnalyticsEventProperties = ({
```
zzmp, line 21: Move this to analytics/utils

zzmp (approved):

PR #4879, opened 2022-10-12
fix: handle backspace out of /tokens

vm:
We don't need the logic to specifically route to the chain in the /tokens url as the redirect happens in Tokens page component.

We also need to replace the url instead of redirecting, so that backspace will work.


cartcrom (approved):
Looks great

PR #4885, opened 2022-10-12
fix: updates outputCurrency link in mobile balance footer

vm:
Uses the same address creation logic from the parent


cartcrom (approved):
Looks good

PR #4888, opened 2022-10-12
revert: add back phase0 bug fixes

vm:
Changes for after deploy today

### DIFF
--- a/src/components/Tokens/TokenDetails/MobileBalanceSummaryFooter.tsx
+++ b/src/components/Tokens/TokenDetails/MobileBalanceSummaryFooter.tsx
@@ -82,7 +82,8 @@ export default function MobileBalanceSummaryFooter({
   tokenAmount,
   nativeCurrencyAmount,
   isNative,
-}: BalanceSummaryProps) {
+  tokenAddress,
+}: BalanceSummaryProps & { tokenAddress?: string }) {
   const balanceUsdValue = useStablecoinValue(tokenAmount)
   const nativeBalanceUsdValue = useStablecoinValue(nativeCurrencyAmount)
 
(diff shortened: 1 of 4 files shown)


JFrankfurt (approved):

src/components/Tokens/TokenDetails/MobileBalanceSummaryFooter.tsx
```diff
@@ -82,7 +82,8 @@ export default function MobileBalanceSummaryFooter({
   tokenAmount,
   nativeCurrencyAmount,
   isNative,
-}: BalanceSummaryProps) {
+  tokenAddress,
+}: BalanceSummaryProps & { tokenAddress?: string }) {
```
JFrankfurt, line 86: you should already have `tokenAddress` available on `tokenAmount`

JFrankfurt (changes requested):

src/components/Tokens/TokenDetails/MobileBalanceSummaryFooter.tsx
vm: good pt, fixed

JFrankfurt: image

it appears to be working in the case I just tested... 🤔

JFrankfurt: image

I've tested: (on ethereum w/ a connected wallet) ETH ("NATIVE"), WETH, USDT (no balance), DAI (w/ balance) and all worked as expected

vm: hmmmmm

src/components/Tokens/TokenDetails/MobileBalanceSummaryFooter.tsx
```diff
@@ -123,7 +121,7 @@ export default function MobileBalanceSummaryFooter({
           </BalanceTotal>
         </BalanceInfo>
       )}
-      <SwapButton to={`/swap?outputCurrency=${outputTokenAddress}`}>
+      <SwapButton to={`/swap?outputCurrency=${tokenAmount?.currency.address}`}>
```
JFrankfurt, line 124: I feel like we shouldn't leave the option open for a swap button rendered w/o a valid url. Could do either: 
- don't render w/o a valid url including outputCurrency
- render, but with only `/swap` and no query parameters

vm:
the issue is the tokenamount should be created with 0 when there is no wallet connected potentially? i am not sure if that's the best solution bc to me 0 balance and no balance are different.

PR #4890, opened 2022-10-12
fix: remove old background-color flash

JFrankfurt:
since we needed to check feature flags before we showed phase0, the old `background` value was always the first thing to render


vm:
might be the same as in https://github.com/Uniswap/interface/pull/4888?

PR #4893, opened 2022-10-12
fix: use correct optimism icon in explore

cartcrom:
used preferred design confirmed by op team


src/components/Tokens/TokenTable/NetworkFilter.tsx
```diff
@@ -106,7 +106,7 @@ export default function NetworkFilter() {
       <NetworkFilterOption onClick={toggleMenu} aria-label={`networkFilter`} active={open}>
         <StyledMenuContent>
           <NetworkLabel>
-            <Logo src={circleLogoUrl ?? logoUrl} /> {label}
```
vm, line 109: should we just kill the circle urls?

zzmp: What else uses circle url?

src/constants/chainInfo.ts
```diff
@@ -114,7 +113,8 @@ const CHAIN_INFO: ChainInfoMap = {
     infoLink: 'https://info.uniswap.org/#/optimism/',
     label: 'Optimism',
     logoUrl: optimismLogoUrl,
-    circleLogoUrl: optimismCircleLogoUrl,
+    // Optimism perfers same icon for both
+    circleLogoUrl: optimismLogoUrl,
```
zzmp, line 117: Considering that your updating the circleLogoUrl, I don't think you need to make anny changes to NetworkFilter.tsx

src/graphql/data/util.ts
```diff
@@ -66,7 +66,7 @@ export const CHAIN_NAME_TO_CHAIN_ID: { [key: string]: SupportedChainId } = {
   OPTIMISM: SupportedChainId.OPTIMISM,
 }
 
-export const BACKEND_CHAIN_NAMES: Chain[] = ['ARBITRUM', 'CELO', 'ETHEREUM', 'OPTIMISM', 'POLYGON']
+export const BACKEND_CHAIN_NAMES: Chain[] = ['ETHEREUM', 'POLYGON', 'OPTIMISM', 'ARBITRUM', 'CELO']
```
zzmp, line 69: back out this change - it looks unrelated.

zzmp (changes requested):

src/graphql/data/util.ts
cartcrom: This does affect the order in which the ui displays them. I updated it so the explore network filter lists chains in the same order as the global network switcher. Since this doesn't affect the data components but makes UI better I think this is a relevant change. Ultimately a better solution could be to update chainInfo to include a backend chain name field instead of having this list, perhaps in sep PR

src/components/Tokens/TokenTable/NetworkFilter.tsx
cartcrom: logoUrl is for displaying chain switching options, like a main logo. CircleLogo provides circular image that is more visible laid over a token icon, used in the token table & details. LogoUrls don't look good overlaid on token icons because parts of them are transparent

src/constants/chainInfo.ts
cartcrom: The filter needed to be updated to use logos instead of circleLogos, to match how the global network switcher displays

zzmp (approved):

src/graphql/data/util.ts
awruble: ah yes I had a sep ticket for this network ordering in the UI

PR #4895, opened 2022-10-12
fix: make background opaque

JFrankfurt:
## before:
image

## after: 
image


src/components/Tokens/TokensBanner.tsx
```diff
@@ -71,26 +75,28 @@ export default function TokensBanner() {
...
-          color={theme.textSecondary}
-          onClick={(e) => {
-            e.preventDefault()
-            e.stopPropagation()
-            setShowTokensPromoBanner(false)
-          }}
-          style={{ cursor: 'pointer' }}
-        />
-      </Header>
+    <BackgroundColor>
```
cmcewen, line 78: won't this persist if the banner isn't shown?

cmcewen (changes requested):

src/components/Tokens/TokensBanner.tsx
JFrankfurt: image

gah! good catch! I assumed it was being conditionally rendered by the parent

cmcewen (approved):

PR #4901, opened 2022-10-12
feat: revised profile card designs

cbachmeier:
- updates 721 profile card styles
- for both listed and unlisted nfts
- fixes wrong color for accentActionSoft in Dark Mode
- updates width of shopping bag when calculating how much to reduce screen size on profile page


gbugyis (approved):
Overall LGTM, just wondering if some of these should be moved to styled-components here? Also could be done separately together with the other Cards. 

Seeing one more key prop warning from the loading skeleton but nothing blocking.

PR #4904, opened 2022-10-12
feat: load token from query data

zzmp:
- Prevents balance fetching cross-chain. Balances are fetched on-chain, so cross-chain balances would be invalid.
- Adds an onSwitchChain handler to the widget to allow it to control the global chain.
  - This requires upgrade of the widget.
  - In order to make this visually clean, also adds hysteresis to the widget:
    - A loading token does not update the widget's output token.
    - The widget will continue to render as a skeleton until it has a token specified.
- Loads the Token Details token (the initial widget token) using query data.
  - This is faster (fewer on-chain calls), only requiring a call for decimals.
  - Uses the RPC provider for the cross-chain, if applicable, to ensure valid data.
  
---

  REVIEWER: These changes are split across specific commits. You may review this as one PR, but you will probably find it easier to look at each commit individually.

### DIFF
--- a/src/components/Widget/index.tsx
+++ b/src/components/Widget/index.tsx
@@ -2,13 +2,20 @@
 // eslint-disable-next-line no-restricted-imports
 import '@uniswap/widgets/dist/fonts.css'
 
-import { Currency, EMPTY_TOKEN_LIST, OnReviewSwapClick, SwapWidget, SwapWidgetSkeleton } from '@uniswap/widgets'
+import {
+  AddEthereumChainParameter,
+  Currency,
+  EMPTY_TOKEN_LIST,
+  OnReviewSwapClick,
+  SwapWidget,
+  SwapWidgetSkeleton,
+} from '@uniswap/widgets'
 import { useWeb3React } from '@web3-react/core'
-import { networkConnection } from 'connection'
-import { RPC_PROVIDERS } from 'constants/providers'
 import { useActiveLocale } from 'hooks/useActiveLocale'
+import { useCallback } from 'react'
 import { useIsDarkMode } from 'state/user/hooks'
 import { DARK_THEME, LIGHT_THEME } from 'theme/widget'
+import { switchChain } from 'utils/switchChain'
 
 import { useSyncWidgetInputs } from './inputs'
 import { useSyncWidgetSettings } from './settings'
@@ -32,19 +39,28 @@ export default function Widget({ defaultToken, onReviewSwapClick }: WidgetProps)
   const { settings } = useSyncWidgetSettings()
   const { transactions } = useSyncWidgetTransactions()
 
+  const onSwitchChain = useCallback(
+    ({ chainId }: AddEthereumChainParameter) => switchChain(connector, Number(chainId)),
+    [connector]
+  )
+
+  if (!inputs.value.INPUT && !inputs.value.OUTPUT) {
+    return <WidgetSkeleton />
+  }
+
   return (
     <>
       <SwapWidget
         disableBranding
         hideConnectionUI
-        jsonRpcUrlMap={RPC_PROVIDERS}
         routerUrl={WIDGET_ROUTER_URL}
         width={WIDGET_WIDTH}
         locale={locale}
         theme={theme}
         onReviewSwapClick={onReviewSwapClick}
         // defaultChainId is excluded - it is always inferred from the passed provider
-        provider={connector === networkConnection.connector ? null : provider} // use jsonRpcUrlMap for network providers
+        provider={provider}
+        onSwitchChain={onSwitchChain}
         tokenList={EMPTY_TOKEN_LIST} // prevents loading the default token list, as we use our own token selector UI
         {...inputs}
         {...settings}
--- a/src/components/Widget/index.tsx
+++ b/src/components/Widget/index.tsx
@@ -40,7 +40,8 @@ export default function Widget({ defaultToken, onReviewSwapClick }: WidgetProps)
   const { transactions } = useSyncWidgetTransactions()
 
   const onSwitchChain = useCallback(
-    ({ chainId }: AddEthereumChainParameter) => switchChain(connector, Number(chainId)),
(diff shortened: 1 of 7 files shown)


zzmp:
I ran the cypress tests manually and they are passing.

src/components/Widget/inputs.tsx
```diff
@@ -21,10 +21,13 @@ export function useSyncWidgetInputs(defaultToken?: Currency) {
   })
 
   useEffect(() => {
-    setTokens({
-      [Field.OUTPUT]: defaultToken,
-    })
-    setAmount(EMPTY_AMOUNT)
+    // Add some hysteresis - if there is no new default, do not update the display.
```
JFrankfurt, line 24: nit: prob don't need this comment

JFrankfurt:
I haven't figured out why yet, but this change results in WETH being set as the widget output token when the user visits /tokens/ethereum/NATIVE. It should be set to ETH in that case.

JFrankfurt:
Possibly related to the prior comment:
the current app sets the widget output currency to ETH when it should be DAI on this url: 
https://app.uniswap.org/#/tokens/arbitrum/0xda10009cbd5d07dd0cecc66161fc93d7c9000da1

but your diff here correctly sets the output to Arbitrum DAI.

src/components/Widget/inputs.tsx
zzmp: I'd prefer to keep it to avoid someone trying to clean up the code and deleting the if-statement.

src/components/Widget/index.tsx
```diff
@@ -32,19 +39,29 @@ export default function Widget({ defaultToken, onReviewSwapClick }: WidgetProps)
   const { settings } = useSyncWidgetSettings()
   const { transactions } = useSyncWidgetTransactions()
 
+  const onSwitchChain = useCallback(
+    // TODO: Widget should not break if this rejects - upstream the catch to ignore it.
+    ({ chainId }: AddEthereumChainParameter) => switchChain(connector, Number(chainId)).catch(() => undefined),
+    [connector]
+  )
+
+  if (!inputs.value.INPUT && !inputs.value.OUTPUT) {
```
cmcewen, line 48: where would this happen? i think this makes sense on token details but not sure it makes sense elsewhere (not that we would use it elsewhere?)

src/components/Widget/inputs.tsx
cmcewen: we have the usePrevValue hook somewhere for looking for state changes - is that more clear?

src/lib/hooks/useCurrency.ts
```diff
@@ -2,16 +2,90 @@ import { arrayify } from '@ethersproject/bytes'
...
+  symbol?: string | null
+  name?: string | null
+  project?: { logoUrl?: string | null } | null
+} = {}): Token | null | undefined {
+  const { chainId: activeChainId } = useWeb3React()
+  const address = isAddress(tokenAddress)
+  const [decimals, setDecimals] = useState<number | null | undefined>(null)
+
+  const tokenContract = useTokenContract(chainId === activeChainId ? (address ? address : undefined) : undefined, false)
+  const { loading, result: [decimalsResult] = [] } = useSingleCallResult(
```
cmcewen, line 43: ❤️

src/lib/hooks/useCurrency.ts
```diff
@@ -29,10 +103,7 @@ function parseStringOrBytes32(str: string | undefined, bytes32: string | undefin
  * Returns null if token is loading or null was passed.
  * Returns undefined if tokenAddress is invalid or token does not exist.
  */
-export function useTokenFromNetwork(
-  tokenAddress: string | null | undefined,
-  tokenChainId?: number
-): Token | null | undefined {
+export function useTokenFromNetwork(tokenAddress: string | undefined): Token | null | undefined {
```
cmcewen, line 106: i don't think this is properly named now. this behavior is `useTokenFromCurrentWeb3ReactNetwork` or something like that right?

src/lib/hooks/useCurrency.ts
```diff
@@ -2,16 +2,90 @@ import { arrayify } from '@ethersproject/bytes'
...
+    return () => {
+      stale = true
+    }
+  }, [activeChainId, address, chainId, decimalsResult, loading])
+
+  return useMemo(() => {
+    if (!chainId || !address) return undefined
+    if (decimals === null || decimals === undefined) return decimals
+    if (!symbol || !name) {
+      return new Token(chainId, address, decimals, symbol ?? undefined, name ?? undefined)
```
cmcewen, line 81: morally, i still believe we should have token cache. no reason to create new objects given each token with the same chainId and address is the same

src/pages/TokenDetails/index.tsx
```diff
@@ -160,7 +151,11 @@ export default function TokenDetails() {
             <AddressSection address={tokenQueryData.address ?? ''} />
           </LeftPanel>
           <RightPanel>
-            <Widget defaultToken={widgetToken} onReviewSwapClick={onReviewSwap} />
+            <Widget
+              // A null token is still loading, and should not be overridden.
```
cmcewen, line 155: i don't think this is well documented but i do think we should make explicit if this is a common pattern of null vs undefined

src/components/Widget/index.tsx
zzmp: Right now the widget is specialized to token details, and is not used elsewhere.
I'd expect to break it into its own component before using it elsewhere.

src/lib/hooks/useCurrency.ts
zzmp: at this point I agree, but think it should be a follow-up for complexity.

src/pages/TokenDetails/index.tsx
zzmp: I think it's pretty common with the token codepaths

cmcewen (approved):

PR #4909, opened 2022-10-13
feat: use widget validate function

cmcewen:
we can use the precompiled validate function now, and also add a CSP


zzmp:
This feels more dangerous for fast follow - could it wait until after?

cmcewen:
@zzmp i don't mind waiting - what would make you feel more confident about it?

zzmp:
Just general manual testing around changing the csp

public/index.html
```diff
@@ -17,6 +17,7 @@
 
   <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
   <meta name="theme-color" content="#FC72FF" />
+  <meta http-equiv="Content-Security-Policy" content="script-src 'self' 'unsafe-inline'" />
```
JFrankfurt, line 20: A CSP crash course is more than I have capacity to digest rn. I've read through https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP twice now and I'm not certain I fully understand the implications of this change yet. 

@AndrewMohawk -- any thoughts on what (if anything) could break with this change?

JFrankfurt (approved):
image

looks to me like the CSP won't block anything, so lgtm 👍

Issue #4900, opened 2022-10-12
Crash report: `TypeError: Cannot read properties of undefined (reading 'circleLogoUrl')`


St0nersdash:
## URL
  
https://app.uniswap.org/#/tokens/ethereum_goerli

## Error

```
TypeError: Cannot read properties of undefined (reading 'circleLogoUrl')
```

## Stacktrace

```
TypeError: Cannot read properties of undefined (reading 'circleLogoUrl')
    at https://app.uniswap.org/static/js/main.86b651fc.chunk.js:1:793006
    at Sa (https://app.uniswap.org/static/js/9.11782dee.chunk.js:2:3689420)
    at Of (https://app.uniswap.org/static/js/9.11782dee.chunk.js:2:3699530)
    at Ss (https://app.uniswap.org/static/js/9.11782dee.chunk.js:2:3748695)
    at gu (https://app.uniswap.org/static/js/9.11782dee.chunk.js:2:3735568)
    at mu (https://app.uniswap.org/static/js/9.11782dee.chunk.js:2:3735496)
    at vu (https://app.uniswap.org/static/js/9.11782dee.chunk.js:2:3735359)
    at ou (https://app.uniswap.org/static/js/9.11782dee.chunk.js:2:3732139)
    at iu (https://app.uniswap.org/static/js/9.11782dee.chunk.js:2:3730690)
    at O (https://app.uniswap.org/static/js/9.11782dee.chunk.js:2:3757270)
```

## Device data

```json
{
  "ua": "Mozilla/5.0 (Linux; Android 10; Android SDK built for x86 Build/OSM1.180201.023) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/77.0.3865.92 Mobile Safari/537.36",
  "browser": {
    "name": "Chrome",
    "version": "77.0.3865.92",
    "major": "77"
  },
  "engine": {
    "name": "Blink",
    "version": "77.0.3865.92"
  },
  "os": {
    "name": "Android",
    "version": "10"
  },
  "device": {
    "model": "Android SDK",
    "type": "mobile"
  },
  "cpu": {}
}
```

Issue #4897, opened 2022-10-12
Crash report: `TypeError: Cannot read properties of undefined (reading 'circleLogoUrl')`


Pochetes:
## URL
  
https://app.uniswap.org/#/swap

## `swap` state
    
```json
{
  "INPUT": {
    "currencyId": "ETH"
  },
  "OUTPUT": {
    "currencyId": null
  },
  "independentField": "INPUT",
  "typedValue": "",
  "recipient": null
}
```

## Error

```
TypeError: Cannot read properties of undefined (reading 'circleLogoUrl')
```

## Stacktrace

```
TypeError: Cannot read properties of undefined (reading 'circleLogoUrl')
    at https://app.uniswap.org/static/js/main.86b651fc.chunk.js:1:793006
    at Sa (https://app.uniswap.org/static/js/9.11782dee.chunk.js:2:3689420)
    at Of (https://app.uniswap.org/static/js/9.11782dee.chunk.js:2:3699530)
    at Ss (https://app.uniswap.org/static/js/9.11782dee.chunk.js:2:3748695)
    at gu (https://app.uniswap.org/static/js/9.11782dee.chunk.js:2:3735568)
    at mu (https://app.uniswap.org/static/js/9.11782dee.chunk.js:2:3735496)
    at vu (https://app.uniswap.org/static/js/9.11782dee.chunk.js:2:3735359)
    at ou (https://app.uniswap.org/static/js/9.11782dee.chunk.js:2:3732139)
    at su (https://app.uniswap.org/static/js/9.11782dee.chunk.js:2:3732530)
    at Hi (https://app.uniswap.org/static/js/9.11782dee.chunk.js:2:3673141)
```

## Device data

```json
{
  "ua": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/105.0.0.0 Safari/537.36",
  "browser": {
    "name": "Chrome",
    "version": "105.0.0.0",
    "major": "105"
  },
  "engine": {
    "name": "Blink",
    "version": "105.0.0.0"
  },
  "os": {
    "name": "Mac OS",
    "version": "10.15.7"
  },
  "device": {},
  "cpu": {}
}
```

Issue #3611, opened 2022-03-27
Add sentry support.


zhyd1997:
**Is your feature request related to a problem? Please describe.**

I'm always frustrated when errors occurred but hard to debug.

**Describe the solution you'd like**

Integrate the app with Sentry, but replace user IP address with wallet account.

**Describe alternatives you've considered**

If we can refactor the codebase with Nextjs, it will be better for uploading source maps to Sentry performance and app running performance.

**Additional context**

PR #4916, opened 2022-10-13
fix: handled undefined circleLogoUrl and failed fetchQuery

zhyd1997:
Fixes: #4897 
Fixes: #4900 
Fixes: #4908 
Fixes: #4912 
Fixes: #4926
Fixes: #4927
Fixes: #4929
Fixes: #4930


src/components/Tokens/TokenTable/NetworkFilter.tsx
```diff
@@ -99,7 +99,7 @@ export default function NetworkFilter() {
   const { chainName } = useParams<{ chainName?: string }>()
   const currentChainName = validateUrlChainParam(chainName)
 
-  const { label, circleLogoUrl, logoUrl } = getChainInfo(CHAIN_NAME_TO_CHAIN_ID[currentChainName])
+  const { label, circleLogoUrl = '', logoUrl } = getChainInfo(CHAIN_NAME_TO_CHAIN_ID[currentChainName])
```
zzmp, line 102: I don't know that this is strictly necessary. Do you?

zzmp (changes requested):
Could you add:
- src/components/Tokens/TokenTable/TokenRow.tsx line 486

Also, the root cause of this is that `CHAIN_NAME_TO_CHAIN_ID[token.chain]` may be returning undefined. Can you replace any instances of `CHAIN_NAME_TO_CHAIN_ID[chain]` with a new method, `getChainNameForChainId(chainId: number): SupportedChainId | undefined`? That will fix the typing, so that typescript does not even _let_ us access circleLogoUrl without the optional operator.

zzmp:
> Could you add:
> 
> * src/components/Tokens/TokenTable/TokenRow.tsx line 486
> 
> Also, the root cause of this is that `CHAIN_NAME_TO_CHAIN_ID[token.chain]` may be returning undefined. Can you replace any instances of `CHAIN_NAME_TO_CHAIN_ID[chain]` with a new method, `getChainNameForChainId(chainId: number): SupportedChainId | undefined`? That will fix the typing, so that typescript does not even _let_ us access circleLogoUrl without the optional operator.

I think it'd be enough to add TokenRow.tsx to get this merged.
Fixing the root cause can be done in a follow-up (by you or by us).

zhyd1997:
@zzmp 

Apologized for this PR change, previously I thought this is just an optional chaining issue, so I set it up on Gitpod, but there is a CORS error for me to fetch the data:
Screen Shot 2022-10-14 at 08 32 18

I found that all the PRs deployed here have no such issues, so I attempted to push it here, unfortunately, the PR not solved the `undefined` issue.

I will try to clone it on my M1 mac and see if localhost has CORS error 👀

src/graphql/data/fetchGraphQL.ts
```diff
@@ -16,7 +16,12 @@ const fetchQuery = (params: RequestParameters, variables: Variables): Promise<Gr
     variables,
   })
 
-  return fetch(URL, { method: 'POST', body, headers }).then((res) => res.json())
+  return fetch(URL, { method: 'POST', body, headers })
+    .then((res) => res.json())
+    .catch((e) => {
+      console.error(e)
+      return { data: [] }
+    })
```
zhyd1997, line 24: Also, I add a `catch` statement to prevent the app crashed:
Screen Shot 2022-10-14 at 08 39 40

I also suggest that the team can [set up Sentry support ](https://github.com/Uniswap/interface/issues/3611) to debug all of the errors in the production environment.

src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -483,7 +483,7 @@ export const LoadedRow = forwardRef((props: LoadedRowProps, ref: ForwardedRef<HT
 
   const lowercaseChainName = useParams<{ chainName?: string }>().chainName?.toUpperCase() ?? 'ethereum'
   const filterNetwork = lowercaseChainName.toUpperCase()
-  const L2Icon = getChainInfo(CHAIN_NAME_TO_CHAIN_ID[filterNetwork]).circleLogoUrl
+  const L2Icon = getChainInfo(CHAIN_NAME_TO_CHAIN_ID[filterNetwork])?.circleLogoUrl
```
zhyd1997, line 486: @zzmp 

You're right !!! 🚀

zhyd1997:
> Can you replace any instances of `CHAIN_NAME_TO_CHAIN_ID[chain]` with a new method, `getChainNameForChainId(chainId: number): SupportedChainId | undefined`?

Would this be `getChainIdForChainName(chainName: string)`? 👀

zhyd1997:
Hi @zzmp 

I just marked the PR as ready for review, `getChainIdForChainName(chainName: string)` will not solve the problem because  the `getChainInfo` function signature would be changed to this line if `chainId` is `undefined`:

https://github.com/Uniswap/interface/blob/332843f428b308e071b7ba8324aa50e4b8046aa2/src/constants/chainInfo.ts#L224-L226

so that we still need optional chaining for `undefined`.

src/graphql/data/fetchGraphQL.ts
zzmp: This should return { error: ErrorObject }, but it can be fixed in a follow-up. This should do a good job of fixing the crash for now.

zzmp (approved):

PR #4931, opened 2022-10-15
fix: add padding to LoadingRow

vm:
This makes sure that there is no jump once the tokens are loaded

### DIFF
--- a/src/components/Tokens/TokenTable/TokenRow.tsx
+++ b/src/components/Tokens/TokenTable/TokenRow.tsx
@@ -348,7 +348,6 @@ function HeaderCell({
   category,
 }: {
   category: TokenSortMethod // TODO: change this to make it work for trans
-  sortable: boolean
 }) {
   const theme = useTheme()
   const sortAscending = useAtomValue(sortAscendingAtom)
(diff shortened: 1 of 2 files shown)


vm:
This is the intended behavior, so this PR is reviewable.

src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -346,7 +346,6 @@ export const HEADER_DESCRIPTIONS: Record<TokenSortMethod, ReactNode | undefined>
 /* Get singular header cell for header row */
 function HeaderCell({
   category,
-  sortable,
```
cmcewen, line 349: why did you remove it from the destructuring and not the type?

src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -441,7 +440,7 @@ export function HeaderRow() {
 }
 
 /* Loading State: row component with loading bubbles */
-export function LoadingRow() {
+export function LoadingRow(props: { first?: boolean; last?: boolean }) {
```
cmcewen, line 443: can you use CSS rules for this?

vm: i gave it a few tries and couldn't figure something out

cmcewen (approved):

PR #4932, opened 2022-10-16
fix: TokenRow hover state shouldn't include padding

vm:
This is due to padding existing on the top and bottom cells on the table. If we include the hover state with padding then the top element is larger than the rest.


cartcrom (changes requested):
I wouldn't merge this unless design wants this. From @infredible ~week ago:

image

vm:
@infredible do you want the space to be gone completely? so we would remove the padding between the header and the first element.

vm:
Screen Shot 2022-10-17 at 11 27 02 AM
otherwise we have a single element on top which is sized differently

vm:
https://github.com/Uniswap/interface/pull/4942 give this a try as well

PR #4940, opened 2022-10-17
fix: hover state shows for full header cell

vm:
Since the clickable area is the entire cell, we should make sure the hover state shows for any part of the cell.


cbachmeier (approved):
LGTM

PR #4941, opened 2022-10-17
chore: remove navBarFlag

vm:
removes all usage of this flag


src/pages/Pool/PositionPage.tsx
```diff
@@ -49,18 +48,18 @@ import { calculateGasMargin } from '../../utils/calculateGasMargin'
...
+const PageWrapper = styled.div`
+  padding: 68px 8px 0px;
 
   min-width: 800px;
   max-width: 960px;
 
   @media only screen and (max-width: ${({ theme }) => `${theme.breakpoint.md}px`}) {
-    padding: ${({ navBarFlag }) => (navBarFlag ? '48px 8px 0px' : '0px 8px 0px')};
+    padding: 48px 8px 0px;
   }
```
cbachmeier, line 60: non blocking nit: but this PageWrapper is re-declared a lot and it might be worth making into a common reusable style

cbachmeier (approved):
LGTM

src/pages/Pool/PositionPage.tsx
vm: agree. i'm not sure why we did it this way... will make a task

PR #4952, opened 2022-10-18
fix: don't crash on unexpected chain

vm:
We may want to consider a better error state here, but at least this will not cause a crash.


src/components/Tokens/TokenTable/NetworkFilter.tsx
```diff
@@ -130,7 +130,7 @@ export default function NetworkFilter() {
                 }}
               >
                 <NetworkLabel>
-                  <Logo src={chainInfo.logoUrl ?? chainInfo.circleLogoUrl} />
+                  <Logo src={chainInfo.logoUrl} />
```
zzmp, line 133: Should this be optional?

zzmp (changes requested):

zzmp (approved):

PR #4954, opened 2022-10-18
fix: show blocked tokens warning

cartcrom:
Moved token warning to allow it to appear even when backend returns no data for a token

i.e. Grump Cat at http://localhost:3000/#/tokens/ethereum/0x93B2FfF814FCaEFFB01406e80B4Ecd89Ca6A021b
image

the unsupported tokenlist also appeared to be failing to skip validation locally so I also modified the list updater code. (These lists need to skip validation because many blocked tokens have long names that don't fit validation criteria)

### DIFF
--- a/src/state/lists/updater.ts
+++ b/src/state/lists/updater.ts
@@ -25,9 +25,11 @@ export default function Updater(): null {
   const fetchList = useFetchListCallback()
   const fetchAllListsCallback = useCallback(() => {
     if (!isWindowVisible) return
-    Object.keys(lists).forEach((url) =>
-      fetchList(url).catch((error) => console.debug('interval list fetching error', error))
-    )
+    Object.keys(lists).forEach((url) => {
+      // Skip validation on unsupported lists
+      const isUnsupportedList = UNSUPPORTED_LIST_URLS.includes(url)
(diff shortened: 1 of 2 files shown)


src/state/lists/updater.ts
```diff
@@ -25,9 +25,11 @@ export default function Updater(): null {
   const fetchList = useFetchListCallback()
   const fetchAllListsCallback = useCallback(() => {
     if (!isWindowVisible) return
-    Object.keys(lists).forEach((url) =>
-      fetchList(url).catch((error) => console.debug('interval list fetching error', error))
-    )
+    Object.keys(lists).forEach((url) => {
+      // Skip validation on unsupported lists
+      const isUnsupportedList = UNSUPPORTED_LIST_URLS.includes(url)
+      fetchList(url, false, isUnsupportedList).catch((error) => console.debug('interval list fetching error', error))
```
JFrankfurt, line 31: nit: I personally prefer not including debug/error logs that aren't necessary in prod code

JFrankfurt (approved):
lgtm

src/state/lists/updater.ts
cartcrom: agree, this was already there but maybe we can take this one and some of the other logs out separately

PR #4955, opened 2022-10-18
chore: remove favoriteTokens flag

vm:
There was some discussion of eventually adding favorited tokens back in, but I believe by that point the architecture of this feature would be quite different. There is also a class of bugs that can be caused by keeping this feature in code during that time. It is simple enough that re-implementing later is fine, and can always reference this PR if needed.


cartcrom (approved):
looks good

PR #4956, opened 2022-10-18
fix: truncate search token name earlier

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-1560?atlOrigin=eyJpIjoiNDNlZTBiNjk1NTdlNDYyZTg5MzI4OGRlODRmMjg3OTIiLCJwIjoiaiJ9

before:
Screen Shot 2022-10-18 at 11 37 37 AM

after:
Screen Shot 2022-10-18 at 11 37 05 AM

### DIFF
--- a/src/components/NavBar/SearchBar.css.ts
+++ b/src/components/NavBar/SearchBar.css.ts
@@ -103,8 +103,10 @@ export const suggestionImage = sprinkles({
 export const suggestionPrimaryContainer = style([
   sprinkles({
     alignItems: 'flex-start',
-    width: 'almostFull',
   }),
+  {
+    width: '90%',
+  },
 ])
 
 export const suggestionSecondaryContainer = sprinkles({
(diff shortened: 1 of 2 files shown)


src/components/NavBar/SearchBar.css.ts
```diff
@@ -103,7 +103,7 @@ export const suggestionImage = sprinkles({
 export const suggestionPrimaryContainer = style([
   sprinkles({
     alignItems: 'flex-start',
-    width: 'full',
```
cbachmeier, line 106: We only want to add sprinkles if they're used frequently. Since this is a one-off, can you instead after the closing bracket of `sprinkles` create a new obj with the 90% width style. It will look like this:
```
export const suggestionPrimaryContainer = style([
  sprinkles({
    alignItems: 'flex-start',
  }),
  {
    width: '90%',
  }
])  
```

cbachmeier (changes requested):

src/components/NavBar/SearchBar.css.ts
lynnshaoyu: thank you! makes sense, fixed.

cbachmeier (approved):
LGTM

PR #4961, opened 2022-10-18
feat: updated token safety copy

cartcrom:


src/constants/tokenSafety.tsx
```diff
@@ -14,17 +14,36 @@ export function getWarningCopy(warning: Warning | null, plural = false) {
...
-        <Plural
-          value={plural ? 2 : 1}
-          _1="You can't trade this token using the Uniswap App."
-          other="You can't trade these tokens using the Uniswap App."
-        />
-      )
+    switch (warning.level) {
+      case WARNING_LEVEL.MEDIUM:
+        heading = (
+          <Plural
```
vm, line 20: if you have `plural`, why do you need to pass it into this component? can't you just do `plural ? text1 : text2?`

vm (approved):

src/constants/tokenSafety.tsx
cartcrom: Plural is a replacement for using ternary \<Trans> logic

cartcrom:
will merge this branch in tandem with some other safety changes after approval from legal

src/constants/tokenSafety.tsx
vm: ah

PR #4964, opened 2022-10-19
chore: remove redesign flag in swap

vm:
Necessary so we can remove the redesign flag from base components.


yyip-dev (approved):
Lgtm! Caveat: still new and my first review here(!) but this cleanup looks straightforward; I clicked around on the preview and didn't see any issues.

PR #4958, opened 2022-10-18
chore: remove remaining redesign flag usage

vm:
one piece of removing the redesign flag within the codebase

PR #4965, opened 2022-10-19
chore: remove liquidity redesign flags

vm:
Removing the flags from the liquidity features separately (this was the only usage)


vm:
consolidated in https://github.com/Uniswap/interface/pull/4958

PR #4966, opened 2022-10-19
chore: rm redesign flags from CurrencyInputPanel

vm:
We probably should re-consolidate these two panels soon as well, but seems like a separate problem.

### DIFF
--- a/src/components/CurrencyInputPanel/SwapCurrencyInputPanel.tsx
+++ b/src/components/CurrencyInputPanel/SwapCurrencyInputPanel.tsx
@@ -105,7 +105,7 @@ const CurrencySelect = styled(ButtonGray)<{
   visibility: ${({ visible }) => (visible ? 'visible' : 'hidden')};
 `
 
-const InputRow = styled.div<{ selected: boolean }>`
+const InputRow = styled.div`
   ${({ theme }) => theme.flexRowNoWrap}
   align-items: center;
   justify-content: space-between;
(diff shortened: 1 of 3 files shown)


src/components/CurrencyInputPanel/SwapCurrencyInputPanel.tsx
```diff
@@ -25,153 +24,110 @@ import { RowBetween, RowFixed } from '../Row'
...
-  background-color: ${({ theme, redesignFlag }) => (redesignFlag ? 'transparent' : theme.deprecated_bg2)};
   display: flex;
   align-items: center;
   justify-content: center;
   z-index: 2;
 `
 
-const Container = styled.div<{ hideInput: boolean; disabled: boolean; redesignFlag: boolean }>`
-  min-height: ${({ redesignFlag }) => redesignFlag && '44px'};
+const Container = styled.div<{ hideInput: boolean; disabled: boolean }>`
```
JackShort: I think you can remove `disabled`

src/components/CurrencyInputPanel/SwapCurrencyInputPanel.tsx
```diff
@@ -25,153 +24,110 @@ import { RowBetween, RowFixed } from '../Row'
...
-    `}
+  &:active:before {
+    background-color: ${({ theme }) => theme.stateOverlayPressed};
+  }
 
   visibility: ${({ visible }) => (visible ? 'visible' : 'hidden')};
 `
 
-const InputRow = styled.div<{ selected: boolean; redesignFlag: boolean }>`
+const InputRow = styled.div<{ selected: boolean }>`
```
JackShort, line 108: `selected` also not needed

JackShort (approved):
lgtm - just another var that isn't used

PR #4967, opened 2022-10-19
feat: show wrapped native asset at top of token selector

cartcrom:
Configured token selector to show wrapped native assets at top of token selector list by default

Also fixed bug where ETH wouldn't show up when searching "ether"
+ small performance improvements in sort methods

WEB-1176

### DIFF
--- a/src/components/SearchModal/CurrencySearch.tsx
+++ b/src/components/SearchModal/CurrencySearch.tsx
@@ -104,20 +104,24 @@ export function CurrencySearch({
   const filteredSortedTokens = useSortTokensByQuery(debouncedQuery, sortedTokens)
 
   const native = useNativeCurrency()
+  const wrapped = native.wrapped
 
-  const filteredSortedTokensWithETH: Currency[] = useMemo(() => {
-    // Use Celo ERC20 Implementation and exclude the native asset
-    if (!native) {
-      return filteredSortedTokens
-    }
-
+  const displayCurrencies: Currency[] = useMemo(() => {
     const s = debouncedQuery.toLowerCase().trim()
-    if (native.symbol?.toLowerCase()?.indexOf(s) !== -1) {
-      // Always bump the native token to the top of the list.
-      return [native, ...filteredSortedTokens.filter((t) => !t.equals(native))]
+    let tokens: Currency[] = filteredSortedTokens
+    if (wrapped.symbol?.toLowerCase()?.indexOf(s) !== -1) {
+      // Always bump the wrapped token to the top of the list.
+      tokens = [wrapped, ...tokens.filter((t) => !t.equals(wrapped))]
     }
-    return filteredSortedTokens
-  }, [debouncedQuery, native, filteredSortedTokens])
+    if (
+      (!disableNonToken && native.symbol?.toLowerCase()?.indexOf(s) !== -1) ||
+      native.name?.toLowerCase()?.indexOf(s) !== -1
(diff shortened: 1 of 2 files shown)


src/lib/hooks/useTokenList/sorting.ts
```diff
@@ -50,11 +50,14 @@ export function useSortTokensByQuery<T extends Token | TokenInfo>(query: string,
     const symbolSubtrings: T[] = []
     const rest: T[] = []
 
+    const trimmedQuery = query.toLowerCase().trim()
```
zzmp, line 53: nit: mv this between the comment and the `tokens.map`.

src/components/SearchModal/CurrencySearch.tsx
```diff
@@ -104,20 +104,24 @@ export function CurrencySearch({
...
   const native = useNativeCurrency()
+  const wrapped = native.wrapped
 
-  const filteredSortedTokensWithETH: Currency[] = useMemo(() => {
-    // Use Celo ERC20 Implementation and exclude the native asset
-    if (!native) {
-      return filteredSortedTokens
-    }
-
+  const displayCurrencies: Currency[] = useMemo(() => {
```
zzmp, line 109: nit: call this `searchCurrencies` to match `CurrencySearch.tsx`.

src/components/SearchModal/CurrencySearch.tsx
```diff
@@ -104,20 +104,24 @@ export function CurrencySearch({
   const filteredSortedTokens = useSortTokensByQuery(debouncedQuery, sortedTokens)
 
   const native = useNativeCurrency()
+  const wrapped = native.wrapped
```
zzmp, line 107: Is there a case (celo?) where `native.equals(native.wrapped)`? In this case, you should omit `wrapped` (ie `const wrapped: Token | undefined = native.wrapped.equals(native) ? undefined : native.wrapped`).

src/components/SearchModal/CurrencySearch.tsx
```diff
@@ -104,20 +104,24 @@ export function CurrencySearch({
...
-    if (!native) {
-      return filteredSortedTokens
-    }
-
+  const displayCurrencies: Currency[] = useMemo(() => {
     const s = debouncedQuery.toLowerCase().trim()
-    if (native.symbol?.toLowerCase()?.indexOf(s) !== -1) {
+    let tokens: Currency[] = filteredSortedTokens
+    if (wrapped.symbol?.toLowerCase()?.indexOf(s) !== -1) {
+      // Always bump the wrapped token to the top of the list.
```
zzmp, line 113: Simplify by moving this out of the condition. Instead of L111, you can have

    const tokens = filteredSortedTokens.filter((t) => {
      if (t.equals(wrapped) || disableNonToken && t.isNative) return false
      return true
    })
    // This would be an alternative to the check mentioned for L107.
    const natives = native.equals(wrapped) ? [native] : [native, wrapped]
    return [...natives, ...tokens]

src/components/SearchModal/CurrencySearch.tsx
zzmp: it looks like this works in celo in the preview, but I'm not sure how...
maybe it uses disableNonToken?

zzmp (changes requested):

zzmp (approved):

PR #4968, opened 2022-10-19
feat: update token safety / lists / verification

cartcrom:
Removes verification icons, updates token warning language, and removes "active lists" from state. tokens from uniswap-default-list is now the only default list viewable in the selector w/ no search input.

Deleted no longer relevant tests for active lists


src/components/SearchModal/CurrencySearch.tsx
```diff
@@ -96,7 +96,10 @@ export function CurrencySearch({
 
   const [balances, balancesAreLoading] = useAllTokenBalances()
   const sortedTokens: Token[] = useMemo(
-    () => (!balancesAreLoading ? [...filteredTokens].sort(tokenComparator.bind(null, balances)) : []),
+    () =>
+      !balancesAreLoading
+        ? [...filteredTokens].sort(tokenComparator.bind(null, balances)).sort(verificationComparator)
```
cmcewen, line 101: why are you sorting twice? can't you just use one comparator?

src/components/TokenSafety/index.tsx
```diff
@@ -255,10 +255,9 @@ export default function TokenSafety({
           <ShortColumn>
             <SafetyLabel warning={displayWarning} />
           </ShortColumn>
-          <ShortColumn>{heading && <InfoText fontSize="20px">{heading}</InfoText>}</ShortColumn>
           <ShortColumn>
             <InfoText>
-              {description}{' '}
+              {heading} {description}{' '}
```
cmcewen, line 260: could you make the spaces more explicit with a template literal? just feels a bit brittle like this

src/state/lists/hooks.ts
```diff
@@ -66,25 +66,20 @@ function useCombinedTokenMapFromUrls(urls: string[] | undefined): TokenAddressMa
...
-
 export function useInactiveListUrls(): string[] {
   const lists = useAllLists()
-  const allActiveListUrls = useActiveListUrls()
   return useMemo(
-    () => Object.keys(lists).filter((url) => !allActiveListUrls?.includes(url) && !UNSUPPORTED_LIST_URLS.includes(url)),
-    [lists, allActiveListUrls]
+    () =>
+      Object.keys(lists).filter(
+        (url) => !DEFAULT_ACTIVE_LIST_URLS?.includes(url) && !UNSUPPORTED_LIST_URLS.includes(url)
```
cmcewen, line 74: you can define this function as a constant outside of here since the lists are constant

KillariDev:
Is there some reasoning why this was done?

PR #4969, opened 2022-10-19
feat: NFT GraphQL Feature Flag

cbachmeier:
- Adds a feature flag for integrating NFT features with the in-progress graphql endpoint
- As the endpoint is currently in beta and not part of the main gql endpoint I moved the env provider to the App level so that I can conditionally set the environment based off the feature flag
  - Tested this locally and Token queries continued to work when the flag was disabled as expected
  - Merged with the in progress branch for pulling nft assets: when the flag was enabled, I was able to retrieve assets as expected
  
screenshots:
Screen Shot 2022-10-19 at 12 28 33


cartcrom (approved):
Looks great, thx for implementing metadata switch

PR #4971, opened 2022-10-20
feat: Collection page log events

gbugyis:
Adds events for page itself, plus Filters (menu, and options), and Activity tab


src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -279,12 +281,19 @@ export const CollectionNfts = ({ contractAddress, collectionStats, rarityVerifie
...
-            <FilterButton
-              isMobile={isMobile}
-              isFiltersExpanded={isFiltersExpanded}
-              onClick={() => setFiltersExpanded(!isFiltersExpanded)}
-              collectionCount={collectionNfts?.[0]?.totalCount ?? 0}
-            />
+            <TraceEvent
+              events={[Event.onClick]}
+              element={ElementName.NFT_FILTER_BUTTON}
+              name={EventName.NFT_FILTER_OPENED}
```
cbachmeier, line 287: Can we change this to NFT_FILTER_TOGGLED as this fires whenever the filter is opened/closed. Alternatively only log the event if the Filter was closed prior to click

src/nft/components/collection/MarketplaceSelect.tsx
```diff
@@ -46,6 +48,7 @@ const MarketplaceItem = ({
       removeMarket(value)
       setCheckboxSelected(false)
     }
+    sendAnalyticsEvent(EventName.NFT_FILTER_SELECTED, { marketplace: value, checked: !isCheckboxSelected })
```
cbachmeier, line 51: Did you run into any ordering issues here? This runs after the `setCheckboxSelected` so on paper it doesn't seem like you would need to negate the `isCheckboxSelected` unless you called the analytics event before the set method

src/nft/pages/collection/index.tsx
```diff
@@ -60,83 +62,85 @@ const Collection = () => {
   }
 
   return (
-    <>
-      <Column width="full">
-        {contractAddress ? (
-          <>
-            {' '}
-            <Box width="full" height="160">
+    <Trace page={PageName.NFT_COLLECTION_PAGE} properties={{ contractAddress }} shouldLogImpression>
```
cbachmeier, line 65: Right now this wraps the collection page but also the error `no collection page exists here` page. Wanted to see if this is intended or if this wrapper should go after the `contractAddress ?` condition

cbachmeier (dismissed):

src/nft/pages/collection/index.tsx
gbugyis: I removed the Trace event here as its no longer needed (fired at App level for all routes)

src/nft/components/collection/CollectionNfts.tsx
gbugyis: at the time I wasn't aware of `shouldLogImpression` - using that now to conditionally fire events

src/nft/components/collection/MarketplaceSelect.tsx
gbugyis: true, it doesn't seem like it should be necessary but somehow, even after the set method, negation was necessary to get the right value passed (same for comment below).

lynnshaoyu:
pls add some screenshots of events being logged to console🙏

lynnshaoyu (changes requested):
need screenshots of events being logged as evidence of local testing

gbugyis:
> need screenshots of events being logged as evidence of local testing

Page View: 
page_view

Filter Opened:
filter_opened

Filter Selected (Price Range):
filter_selected-PriceRange

Filter Selected (Trait):
filter_selected-Trait

Filter Selected (Marketplace):
filter_selected-Marketplace

Activity Selected:
activity_selected

src/analytics/constants.ts
```diff
@@ -14,6 +14,9 @@ export enum EventName {
   PAGE_VIEWED = 'Page Viewed',
   NAVBAR_SEARCH_SELECTED = 'Navbar Search Selected',
   NAVBAR_SEARCH_EXITED = 'Navbar Search Exited',
+  NFT_ACTIVITY_SELECTED = 'NFT Activity Selected',
+  NFT_FILTER_OPENED = 'NFT Collection Filter Opened',
+  NFT_FILTER_SELECTED = 'NFT Filter selected',
```
lynnshaoyu, line 19: pls capitalize Selected

src/nft/components/collection/CollectionStats.tsx
```diff
@@ -277,7 +277,7 @@ const StatsRow = ({ stats, isMobile, ...props }: { stats: GenieCollection; isMob
   )
 
   return (
-    <Row gap={{ sm: '20', md: '60' }} {...props}>
+    <Row gap={{ sm: '36', md: '60' }} {...props}>
```
lynnshaoyu, line 280: is this change intentional? unrelated to analytics?

src/nft/components/collection/MarketplaceSelect.tsx
```diff
@@ -48,6 +50,7 @@ const MarketplaceItem = ({
       removeMarket(value)
       setCheckboxSelected(false)
     }
+    sendAnalyticsEvent(EventName.NFT_FILTER_SELECTED, { filter_type: 'Marketplace' })
```
lynnshaoyu, line 53: can we define a FILTER_TYPES enum for constants like 'Marketplace', 'Price Range' etc?

lynnshaoyu (changes requested):

src/nft/components/collection/CollectionStats.tsx
gbugyis: Yes, intentional. This was a nit on a different PR and I merged that one before I saw the comment.

src/nft/pages/collection/index.tsx
```diff
@@ -61,75 +63,81 @@ const Collection = () => {
...
   return (
     <>
-      <Column width="full">
-        {contractAddress ? (
-          <>
-            {' '}
-            <Box width="full" height="160">
+      <Trace
+        page={PageName.NFT_COLLECTION_PAGE}
+        properties={{ collection_address: contractAddress }}
```
lynnshaoyu, line 68: can we add chain_id as an event property here? i know it's only on eth right now, but that might not be the case forever so want to make sure we start tracking now.

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -356,12 +358,20 @@ export const CollectionNfts = ({ contractAddress, collectionStats, rarityVerifie
...
-                isFiltersExpanded={isFiltersExpanded}
-                onClick={() => setFiltersExpanded(!isFiltersExpanded)}
-                collectionCount={collectionNfts?.[0]?.totalCount ?? 0}
-              />
+              <TraceEvent
+                events={[Event.onClick]}
+                element={ElementName.NFT_FILTER_BUTTON}
+                name={EventName.NFT_FILTER_OPENED}
+                shouldLogImpression={!isFiltersExpanded}
+                properties={{ collection_address: contractAddress }}
```
lynnshaoyu, line 366: pls add chain_id here too. anywhere we log collection_address should include chain_id in general

lynnshaoyu (approved):

PR #4972, opened 2022-10-20
feat: nft sweep

JackShort:
* implemented sweep both mobile and desktop


package.json
```diff
@@ -195,6 +195,7 @@
     "polyfill-object.fromentries": "^1.0.1",
     "popper-max-size-modifier": "^0.2.0",
     "qs": "^6.9.4",
+    "rc-slider": "^10.0.1",
```
cbachmeier, line 198: @aballerr used a different slider library for Price range, can two confirm which slider library we want to use moving forward?

src/nft/hooks/useBag.ts
```diff
@@ -79,7 +82,11 @@ export const useBag = create<BagState>()(
       addAssetToBag: (asset) =>
```
cbachmeier, line 82: since the `fromSweep` is optional and by default is false, you should be able to remove this now overridden function definition `addAssetToBag: (asset: UpdatedGenieAsset) => void`

src/nft/hooks/useBag.ts
```diff
@@ -104,6 +142,29 @@ export const useBag = create<BagState>()(
           return { itemsInBag: itemsCopy }
         })
       },
+      removeAssetsFromBag: (assets) => {
+        set(({ itemsInBag }) => {
+          if (get().isLocked) return { itemsInBag: get().itemsInBag }
+          if (itemsInBag.length === 0) return { itemsInBag: [] }
+          const itemsCopy = itemsInBag.filter((item) => !assets.some((asset) => asset.id === item.asset.id))
+          return { itemsInBag: itemsCopy }
+        })
```
cbachmeier, line 151: Do we need a seperate function or can we either call `removeAsset` multiple times or replace calls to `removeAsset` with `removeAsset` and an array of length 1?

src/nft/queries/genie/SweepFetcher.ts
```diff
@@ -24,22 +36,37 @@ export const fetchSweep = async ({
...
   }
 
   if (markets) {
     payload.markets = markets
   }
-
+  const numberOfTraits = traits?.filter((trait) => trait.trait_type === 'Number of traits')
+  if (numberOfTraits) {
+    payload.filters.numTraits = numberOfTraits.map((el) => ({ traitCount: el.trait_value }))
+  }
```
cbachmeier, line 52: If you need to refactor off Alex's PR I believe we remove `number of traits` and this logic can be removed

cbachmeier (changes requested):
Slack comments on light mode and if swapping from items to eth should clear sweep?

package.json
JackShort: created a pr that changes the price range slider to use rc-slider

src/nft/hooks/useBag.ts
JackShort: I like that - also going to do the same for addAssetToBag

src/nft/hooks/useBag.ts
JackShort: going to change everything to use the bulk addAssetsToBag

src/nft/queries/genie/SweepFetcher.ts
JackShort: W

cbachmeier (approved):
LGTM

PR #4974, opened 2022-10-20
fix: returned sparkline call to memo

cartcrom:
The sparkline query call should be inside a useMemo rather than a useEffect so that it happens pre-render rather than post-render (we want it to run in parallel with the tokens query, rather than after). The useEffect was also triggering repeated calls


zzmp (changes requested):
This should be a `useEffect` so that the outdated subscription can be unsubscribed if the deps change. `useMemo` should only be used to return a computed value with no side effects.

PR #4975, opened 2022-10-21
feat: navigate to widget-selected token

zzmp:
Updates the Token Details page to match the token selected in the widget, if the current page's token is no longer displayed.

NB: This has a visual bug where navigating to the new token goes back to a LoadingTokenDetails state. <s>I'm working on fixing this, but it will likely be a follow-up (unless it is a very small change).</s> This is fixed in a follow-up: #4981.

See WEB-1467

### DIFF
--- a/src/components/Widget/index.tsx
+++ b/src/components/Widget/index.tsx
@@ -34,6 +34,10 @@ export const WIDGET_WIDTH = 360
 
 const WIDGET_ROUTER_URL = 'https://api.uniswap.org/v1/'
 
+function useWidgetTheme() {
+  return useIsDarkMode() ? DARK_THEME : LIGHT_THEME
+}
+
 export interface WidgetProps {
   defaultToken?: Currency
   onTokensChange?: (input: Currency | undefined, output: Currency | undefined) => void
@@ -41,10 +45,9 @@ export interface WidgetProps {
 }
 
 export default function Widget({ defaultToken, onTokensChange, onReviewSwapClick }: WidgetProps) {
-  const locale = useActiveLocale()
-  const theme = useIsDarkMode() ? DARK_THEME : LIGHT_THEME
   const { connector, provider } = useWeb3React()
-
+  const locale = useActiveLocale()
+  const theme = useWidgetTheme()
   const { inputs, tokenSelector } = useSyncWidgetInputs(defaultToken)
   const { settings } = useSyncWidgetSettings()
   const { transactions } = useSyncWidgetTransactions()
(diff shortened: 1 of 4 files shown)


src/components/Widget/inputs.tsx
```diff
@@ -34,8 +34,9 @@ export function useSyncWidgetInputs(defaultToken?: Currency) {
   useEffect(() => {
     // Avoid overwriting tokens if none are specified, so that a loading token does not cause layout flashing.
     if (!defaultToken) return
-    setTokens({
-      [Field.OUTPUT]: defaultToken,
+    setTokens((tokens) => {
+      if (Object.values(tokens).some((token) => token?.equals(defaultToken))) return tokens
+      return { [Field.OUTPUT]: defaultToken }
```
aballerr, line 39: could use a ternary here

src/pages/TokenDetails/index.tsx
```diff
@@ -91,28 +92,34 @@ export default function TokenDetails() {
...
       }
     },
     [isNative, navigate, tokenQueryData?.project?.tokens]
   )
-  useOnGlobalChainSwitch(switchChains)
+  useOnGlobalChainSwitch(navigateToTokenForChain)
+  const navigateToWidgetSelectedToken = useCallback(
+    (input: Currency | undefined, output: Currency | undefined) => {
+      const update = output || input
+      if (!token || input?.equals(token) || output?.equals(token) || !update) return
```
aballerr, line 111: ```input?.equals(token) || output?.equals(token)```

I think these might be redundant with the !update at the end

src/components/Widget/index.tsx
```diff
@@ -36,22 +36,29 @@ const WIDGET_ROUTER_URL = 'https://api.uniswap.org/v1/'
...
+export default function Widget({ defaultToken, onTokensChange, onReviewSwapClick }: WidgetProps) {
   const locale = useActiveLocale()
   const theme = useIsDarkMode() ? DARK_THEME : LIGHT_THEME
   const { connector, provider } = useWeb3React()
 
   const { inputs, tokenSelector } = useSyncWidgetInputs(defaultToken)
   const { settings } = useSyncWidgetSettings()
   const { transactions } = useSyncWidgetTransactions()
+  const onSwitchChain = useCallback(
+    // TODO: Widget should not break if this rejects - upstream the catch to ignore it.
```
aballerr, line 52: do we have a task for this, just want to make sure we don't forget to do this follow up

zzmp: https://uniswaplabs.atlassian.net/browse/WEB-1757

src/pages/TokenDetails/index.tsx
zzmp: Not redundant - but I rearranged the conditionals in case it makes it clearer.

aballerr (approved):

PR #4979, opened 2022-10-21
fix: no margin-top for LoadingRows

vm:
Since the LoadedRows do not have a margin top, there shouldn't be any on LoadingRows either. This way there isn't a jump once the data is fetched.


cartcrom (approved):
LGTM

PR #4983, opened 2022-10-23
fix: Remove CMC token list

Rachel-Eichenberger:
This token list has been unreliable, and always seems to have bad decimal places


vm:
the token lists aren't actually visible in-app now, so we'll likely need to do a code cleanup to remove the lists from state instead

PR #4984, opened 2022-10-24
fix: clean up balance summaries

zzmp:
- Avoids fetching extra data until it is used (ie balances are only fetched in the BalanceSummary components)
- Changes logic for when to display balances: always displays balance if there is one, even if it is 0


cartcrom (approved):
Looks great

PR #4985, opened 2022-10-24
fix: sync network connector to wallet

zzmp:
Syncs the network connector to any active user connector (eg wallet) while the network connector is not the priority connector. Effectively, this maintains the network connector chain when a wallet is disconnected.

### DIFF
--- a/src/components/WalletModal/index.tsx
+++ b/src/components/WalletModal/index.tsx
@@ -225,7 +225,6 @@ export default function WalletModal({
         setWalletView(WALLET_VIEWS.PENDING)
         dispatch(updateConnectionError({ connectionType, error: undefined }))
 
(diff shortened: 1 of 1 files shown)


src/components/WalletModal/index.tsx
```diff
@@ -217,6 +225,7 @@ export default function WalletModal({
         setWalletView(WALLET_VIEWS.PENDING)
         dispatch(updateConnectionError({ connectionType, error: undefined }))
 
+        // TODO(zzmp)
```
lynnshaoyu, line 228: missing TODO description?

lynnshaoyu (commented):
missing description

lynnshaoyu (approved):

PR #4988, opened 2022-10-24
fix: resolve some tokens not searchable by contract address in token selector cx bug

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-1662?atlOrigin=eyJpIjoiZWJhNGZkNDIxMDRkNDA0NmEwOWE4N2E5ZjYxM2E4YzUiLCJwIjoiaiJ9

### DIFF
--- a/src/lib/hooks/useCurrency.ts
+++ b/src/lib/hooks/useCurrency.ts
@@ -137,7 +137,6 @@ export function useTokenFromActiveNetwork(tokenAddress: string | undefined): Tok
     // If the token is on another chain, we cannot fetch it on-chain, and it is invalid.
     if (typeof tokenAddress !== 'string' || !isSupportedChain(chainId) || !formattedAddress) return undefined
     if (isLoading || !chainId) return null
-    if (parsedDecimals == null) return undefined
 
     return new Token(chainId, formattedAddress, parsedDecimals, parsedSymbol, parsedName)
   }, [chainId, tokenAddress, formattedAddress, isLoading, parsedDecimals, parsedSymbol, parsedName])
--- a/src/lib/hooks/useCurrency.ts
+++ b/src/lib/hooks/useCurrency.ts
@@ -12,6 +12,7 @@ import useNativeCurrency from 'lib/hooks/useNativeCurrency'
 import { useEffect, useMemo, useState } from 'react'
 import { WrappedTokenInfo } from 'state/lists/wrappedTokenInfo'
 
+import { DEFAULT_ERC20_DECIMALS } from '../../constants/tokens'
 import { TOKEN_SHORTHANDS } from '../../constants/tokens'
 import { getContract, isAddress } from '../../utils'
 import { supportedChainId } from '../../utils/supportedChainId'
@@ -123,7 +124,11 @@ export function useTokenFromActiveNetwork(tokenAddress: string | undefined): Tok
     () => decimals.loading || symbol.loading || tokenName.loading,
     [decimals.loading, symbol.loading, tokenName.loading]
   )
-  const parsedDecimals = useMemo(() => decimals.result?.[0], [decimals.result])
+  const parsedDecimals = useMemo(
+    () => (decimals.result ? decimals.result?.[0] : DEFAULT_ERC20_DECIMALS),
+    [decimals.result]
+  )
+
   const parsedSymbol = useMemo(
     () => parseStringOrBytes32(symbol.result?.[0], symbolBytes32.result?.[0], 'UNKNOWN'),
     [symbol.result, symbolBytes32.result]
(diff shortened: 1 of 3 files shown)


src/lib/hooks/useCurrency.ts
```diff
@@ -150,9 +154,8 @@ type TokenMap = { [address: string]: Token }
  * Returns undefined if tokenAddress is invalid or token does not exist.
  */
 export function useTokenFromMapOrNetwork(tokens: TokenMap, tokenAddress?: string | null): Token | null | undefined {
-  const address = isAddress(tokenAddress)
+  const address = isAddress(tokenAddress?.toLowerCase())
```
zzmp, line 157: does this need to be lower-cased? seems like something that should happen within `isAddress`, if it is actually necessary.

src/lib/hooks/useCurrency.ts
```diff
@@ -134,9 +139,8 @@ export function useTokenFromActiveNetwork(tokenAddress: string | undefined): Tok
   return useMemo(() => {
     // If the token is on another chain, we cannot fetch it on-chain, and it is invalid.
     if (typeof tokenAddress !== 'string' || !isSupportedChain(chainId) || !formattedAddress) return undefined
-
     if (isLoading || !chainId) return null
-    if (!parsedDecimals) return undefined
+    if (parsedDecimals == null) return undefined
```
zzmp, line 143: what if it is not null and it is 0 or undefined - in this case shouldn't it be DEFAULT_ERC20_DECIMALS?

src/lib/hooks/useCurrency.ts
```diff
@@ -121,7 +122,11 @@ export function useTokenFromActiveNetwork(tokenAddress: string | undefined): Tok
     () => decimals.loading || symbol.loading || tokenName.loading,
     [decimals.loading, symbol.loading, tokenName.loading]
   )
-  const parsedDecimals = useMemo(() => decimals.result?.[0], [decimals.result])
+  const parsedDecimals = useMemo(
+    () => (decimals.result ? decimals.result?.[0] : DEFAULT_ERC20_DECIMALS),
```
zzmp, line 126: Should this be something like decimals?.result?.[0] ?? DEFAULT_ERC20_DECIMALS, or are you trying to explicitly preserve a `null` value?

zzmp (changes requested):

src/lib/hooks/useCurrency.ts
lynnshaoyu: needs to be lowercase, added comment and moved it https://docs.ethers.io/v5/api/utils/address/

src/lib/hooks/useCurrency.ts
JFrankfurt: I don't think it would be safe to default to a potentially invalid decimals value

lynnshaoyu: no 0 is a valid decimals case (10^0 = 1, see: https://docs.openzeppelin.com/contracts/3.x/erc20#a-note-on-decimals) 

by using == and not ===, we also capture the case where parsedDecimals is undefined.

src/lib/hooks/useCurrency.ts
lynnshaoyu: yeah that makes sense actually, done.

src/lib/hooks/useCurrency.ts
lynnshaoyu: Btw Jordan on line 126 of useCurrency.ts, when decimals is not available, we are defaulting to DEFAULT_ERC20_DECIMALS which is 18 because ERC20 standard defaults to 18 when the optional extension which allows you to manually set decimals is not used. See documentation here: https://docs.openzeppelin.com/contracts/3.x/erc20#a-note-on-decimals

zzmp: You should not use `==`, because it's called out by linters, and will likely be "fixed" by someone in the future. You should explicitly check for both `null` and `undefined`, at the expense of terseness.

src/lib/hooks/useCurrency.ts
```diff
@@ -121,7 +122,8 @@ export function useTokenFromActiveNetwork(tokenAddress: string | undefined): Tok
     () => decimals.loading || symbol.loading || tokenName.loading,
     [decimals.loading, symbol.loading, tokenName.loading]
   )
-  const parsedDecimals = useMemo(() => decimals.result?.[0], [decimals.result])
+  const parsedDecimals = useMemo(() => decimals?.result?.[0] ?? DEFAULT_ERC20_DECIMALS, [decimals.result])
```
zzmp, line 125: Using `??` means that `parsedDecimals` will never be `null | undefined`, so L140 is dead code.

zzmp (changes requested):

src/lib/hooks/useCurrency.ts
lynnshaoyu: ooh my linter doesnt complain when == is used with null, but it does complain with undefined? but i see the point about someone "fixing" it. noted for the future. 

removed line 140 tho due to the other comment about dead code.

src/lib/hooks/useCurrency.ts
lynnshaoyu: tru, thanks done.

JFrankfurt (approved):

zzmp (approved):

PR #4995, opened 2022-10-25
feat: page view event on NFT Details

gbugyis:
Adds a page view event for NFT details and an entry in the PageName constant (NFT_DETAILS_PAGE)


lynnshaoyu:
Nice! In the future could you add a screenshot of the event logged to your console in the PR description? just want to see that we've tested each event being properly logged. gonna approve this one though because it's relatively straightforward, but it helps for more complicated ones (and might help us catch edge cases too).

lynnshaoyu (approved):

PR #5003, opened 2022-10-25
fix: bag footer layout

JFrankfurt:


src/nft/components/bag/BagFooter.tsx
```diff
@@ -1,16 +1,42 @@
...
+  color: ${({ theme }) => theme.textPrimary};
+  padding: 12px 16px;
+  border-bottom-left-radius: 12px;
+  border-bottom-right-radius: 12px;
+`
+
+const WarningIcon = styled(AlertTriangle)`
+  width: 14px;
+  margin-right: 4px;
+  color: ${({ theme }) => theme.deprecated_warning};
```
JackShort, line 28: I think the right color is `accentWarning`

src/nft/components/bag/BagFooter.tsx
```diff
@@ -1,16 +1,42 @@
...
+`
+const WarningText = styled(ThemedText.BodyPrimary)`
+  align-items: center;
+  color: ${({ theme }) => theme.accentWarning};
+  display: flex;
+  font-size: 14px;
```
JackShort, line 34: this doesn't seem to be setting the `font-size` to `14px` when I inspect it it is still `16px`. Also the `line-height` should be `20px`

src/nft/components/bag/BagFooter.tsx
```diff
@@ -76,6 +91,16 @@ export const BagFooter = ({
             {`${ethNumberStandardFormatter(totalUsdPrice, true)}`}
           </Row>
         </Column>
+        {showWarning && (
+          <WarningText>
+            <WarningIcon />
+            {!sufficientBalance ? (
+              `Insufficient funds (${formatWeiToDecimal(balance.toString())} ETH)`
```
JackShort, line 98: I think what is left is just the warning `Insufficient funds` is all that is needed and not the balance

src/nft/components/bag/BagFooter.tsx
```diff
@@ -1,16 +1,42 @@
...
 import { ethNumberStandardFormatter, formatWeiToDecimal } from 'nft/utils'
+import { AlertTriangle } from 'react-feather'
 import { useModalIsOpen, useToggleWalletModal } from 'state/application/hooks'
 import { ApplicationModal } from 'state/application/reducer'
+import styled from 'styled-components/macro'
 import { ThemedText } from 'theme'
 
 import * as styles from './BagFooter.css'
 
+const Footer = styled(Column)`
```
JackShort, line 17: also since we want to move off vanilla extract can we use `styled.div` instead and just add `display: flex` and `flex-direction: column`

JackShort (changes requested):

src/nft/components/bag/BagFooter.tsx
JFrankfurt: thanks, not sure why I used a deprecated color here...

src/nft/components/bag/BagFooter.tsx
JFrankfurt: Thanks, double checking this w/ design.

JackShort (approved):
lgtm

PR #5007, opened 2022-10-25
fix: incorrectly displaying empty chart

cartcrom:
WEB-1746 - Fixed issue where front end would not render missing chart properly. Also updated missing chart component to match design


zzmp (changes requested):
Breaks the TokenDetails skeleton - the skeleton should be updated to match up with any layout changes to TokenDetails or its subcomponent.s

src/components/Tokens/TokenDetails/ChartSection.tsx
```diff
@@ -17,13 +17,13 @@ import { L2NetworkLogo, LogoContainer } from '../TokenTable/TokenRow'
...
 
-export const ChartHeader = styled.div`
+export const ChartHeader = styled.div<{ gap?: boolean }>`
   width: 100%;
   display: flex;
   flex-direction: column;
   color: ${({ theme }) => theme.textPrimary};
   gap: 4px;
-  margin-bottom: 24px;
+  margin-bottom: ${({ gap }) => gap && '24px'};
```
zzmp, line 26: The vertical size of the Chart should be consistent, or the skeleton loader will not match what is actually loaded (this may be feedback for design).

zzmp (changes requested):

cartcrom:
Closing in favor of #5030

PR #5012, opened 2022-10-26
fix: modify hyperlink styling in nft collection descriptions

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-1589?atlOrigin=eyJpIjoiY2QyYzQ5NDRlNDUzNDgzN2E0Y2Y4MDExNGFmYWQ1YTgiLCJwIjoiaiJ9

### DIFF
--- a/src/nft/components/collection/CollectionStats.css.ts
+++ b/src/nft/components/collection/CollectionStats.css.ts
@@ -1,4 +1,4 @@
-import { style } from '@vanilla-extract/css'
+import { globalStyle, style } from '@vanilla-extract/css'
 import { bodySmall, buttonTextSmall, headlineSmall } from 'nft/css/common.css'
 import { loadingAsset, loadingBlock } from 'nft/css/loading.css'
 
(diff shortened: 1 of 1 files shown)


src/nft/components/collection/CollectionStats.css.ts
```diff
@@ -67,6 +67,23 @@ export const description = style([
   },
 ])
 
+globalStyle(`${description} a[href]`, {
```
cbachmeier, line 70: We should try to avoid using global styles to avoid unintended consequences. You can add the selectors to the `description` style as shown here: https://vanilla-extract.style/documentation/styling/#selectors

cbachmeier (changes requested):

src/nft/components/collection/CollectionStats.css.ts
lynnshaoyu: if you scroll down on that linked documentation page Charlie it says: 
"If you need to globally target child nodes within the current element (e.g. '& a[href]'), you should use [globalStyle](https://vanilla-extract.style/documentation/global-api/global-style) instead." 

I tried using selectors with a[href] previously and it doesn't work (no styling changes are applied). seems to be an exception case where we _need_ to use globalStyle.

cbachmeier (approved):
LGTM

PR #5013, opened 2022-10-26
fix: do not flicker when navigating tokens

zzmp:
Prevents widget from firing onTokensChange when the onTokensChange handler reference updates.

### DIFF
--- a/src/components/Widget/index.tsx
+++ b/src/components/Widget/index.tsx
@@ -20,7 +20,7 @@ import {
   getTokenAddress,
 } from 'analytics/utils'
 import { useActiveLocale } from 'hooks/useActiveLocale'
-import { useCallback, useEffect, useState } from 'react'
+import { useCallback, useEffect, useRef, useState } from 'react'
 import { useIsDarkMode } from 'state/user/hooks'
 import { DARK_THEME, LIGHT_THEME } from 'theme/widget'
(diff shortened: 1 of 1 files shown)


src/components/Widget/index.tsx
```diff
@@ -51,14 +51,23 @@ export default function Widget({ defaultToken, onTokensChange, onReviewSwapClick
...
   const { settings } = useSyncWidgetSettings()
   const { transactions } = useSyncWidgetTransactions()
+
+  const lastValue = useRef([inputs.value.INPUT, inputs.value.OUTPUT])
+  useEffect(() => {
+    const [input, output] = [inputs.value.INPUT, inputs.value.OUTPUT]
+    const [lastInput, lastOutput] = lastValue.current
+    // Avoid calling onTokensChange if only the handler has changed.
+    if (input === lastInput && output === lastOutput) return
+    if (input && lastInput && input.equals(lastInput) && output && lastOutput && lastOutput.equals(lastOutput)) return
```
lynnshaoyu, line 61: are lines 60 and 61 doing the same thing / kinda redundant? can we just use one of them?

zzmp: L60 covers the case where there are no tokens selected (ie `input = lastInput = output = lastOutput = undefined`).

lynnshaoyu (approved):

PR #5014, opened 2022-10-26
feat: add medium-button and use it nft bag for removal

JFrankfurt:
- hover states on "remove" button
- hover states on the ❌  close icon for the bag
- adds a new general use `<MediumButton />` component that matches the design components figma doc

https://uniswaplabs.atlassian.net/browse/WEB-1794

### DIFF
--- a/src/components/Button/index.tsx
+++ b/src/components/Button/index.tsx
@@ -365,3 +365,33 @@ export function ButtonRadioChecked({ active = false, children, ...rest }: { acti
     )
   }
 }
+
+export const MediumButton = styled.button`
+  align-items: center;
+  background-color: ${({ theme }) => theme.backgroundInteractive};
+  border-radius: 16px;
+  border: 0;
+  color: ${({ theme }) => theme.textPrimary};
+  cursor: pointer;
+  display: flex;
+  flex-direction: row;
+  font-weight: 600;
+  gap: 12px;
+  justify-content: center;
+  padding: 16px;
+  transition: 150ms ease background-color opacity;
+
+  :disabled {
+    background-color: ${({ theme }) => theme.backgroundInteractive};
+    opacity: 0.6;
+  }
+  :hover {
+    background-color: ${({ theme }) => theme.stateOverlayHover};
+  }
+  :active {
+    background-color: ${({ theme }) => theme.stateOverlayPressed};
+  }
+  :focus {
+    background-color: ${({ theme }) => theme.stateOverlayPressed};
+  }
+`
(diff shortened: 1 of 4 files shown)


src/components/Button/index.tsx
```diff
@@ -365,3 +365,33 @@ export function ButtonRadioChecked({ active = false, children, ...rest }: { acti
...
+  cursor: pointer;
+  display: flex;
+  flex-direction: row;
+  font-weight: 600;
+  gap: 12px;
+  justify-content: center;
+  padding: 16px;
+  transition: 150ms ease background-color opacity;
+
+  :disabled {
```
cbachmeier, line 384: Does disabled have `cursor: default`?

src/nft/components/bag/BagHeader.tsx
```diff
@@ -25,6 +24,23 @@ const ControlRow = styled.div`
...
+  display: flex;
+  padding: 2px;
+  opacity: 1;
+  transition: 125ms ease opacity;
+  :hover {
+    opacity: 0.6;
+  }
+  :active {
+    opacity: 0.4;
+  }
```
cbachmeier, line 42: Food for thought and non-blocking, but since we're using these hover and active states globally, might be worth adding them to the styled library

src/nft/components/bag/BagRow.tsx
```diff
@@ -19,9 +20,18 @@ import { GenieAsset, UpdatedGenieAsset } from 'nft/types'
...
 import { MouseEvent, useCallback, useEffect, useReducer, useState } from 'react'
 import { Link } from 'react-router-dom'
+import styled from 'styled-components/macro'
 
 import * as styles from './BagRow.css'
 
+const RemoveButton = styled(MediumButton)`
+  border-radius: 12px;
+  font-size: 14px;
+  line-height: 16px;
```
cbachmeier, line 30: Do our styled components have shared typography styles? NFT has some in `common.css.ts` which would allow for these two to be pulled from a central style for `buttonTextSmall`

cbachmeier (approved):
Works well, had some minor comments

src/nft/components/bag/BagRow.tsx
JFrankfurt: we have the core ones in `src/theme/index#ThemedText`, but I don't think it has been updated to support all our current text styles (there was a large update recently)

src/nft/components/bag/BagHeader.tsx
JFrankfurt: agree

src/components/Button/index.tsx
JFrankfurt: good catch. This should get cursor: pointer by default since it's a button, but for some reason I'm not getting the user agent styles I'm expecting when I use it, so I've set it manually. I'd need to add default for disabled here to override that.

PR #5016, opened 2022-10-26
fix: fetch decimals from token query

zzmp:
Reduces latency and on-chain calls when instantiating a token from query data.

WEB-1751

### DIFF
--- a/src/graphql/data/Token.ts
+++ b/src/graphql/data/Token.ts
@@ -1,11 +1,13 @@
 import graphql from 'babel-plugin-relay/macro'
+import { DEFAULT_ERC20_DECIMALS } from 'constants/tokens'
 import { useMemo, useState } from 'react'
 import { fetchQuery, useLazyLoadQuery } from 'react-relay'
+import { WrappedTokenInfo } from 'state/lists/wrappedTokenInfo'
 
 import { Chain, TokenPriceQuery } from './__generated__/TokenPriceQuery.graphql'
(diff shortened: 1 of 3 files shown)


src/graphql/data/Token.ts
```diff
@@ -19,6 +21,7 @@ const tokenQuery = graphql`
   query TokenQuery($contract: ContractInput!, $duration: HistoryDuration!) {
     tokens(contracts: [$contract]) {
       id @required(action: LOG)
+      decimals
```
cartcrom, line 24: Would we want to add @required(action: LOG) (basically voids a returned token if decimals aren't present, so we don't have to null check for it)? I'm not sure how important decimal accuracy is or if the default is fine

cartcrom (approved):

src/graphql/data/Token.ts
zzmp: It's important to have decimal accuracy, but decimals are not required to be explicit for ERC-20s (see https://github.com/Uniswap/interface/pull/4988), so I don't think it's appropriate to require it. Instead, we should just use DEFAULT_ERC20_DECIMALS if it is not present.

lynnshaoyu: yes we dont need decimals for ERC 20, the default is 18 decimals, unless you use ERC20Detailed extension to specify a custom decimal number ( see here for documentation on diff between ERC20 and ERC20Detailed: https://docs.openzeppelin.com/contracts/2.x/api/token/erc20)

PR #5017, opened 2022-10-26
fix: clean up widget token lock

zzmp:
- Cleans up widget token selection logic
- Puts the onTokenChange handler in a callback instead of a useEffect to prevent recursive calls


cartcrom (approved):
Looks great

PR #5020, opened 2022-10-27
fix: display updated widget token immediately

zzmp:
Tracks whether the default has been set in the widget, so that a new default can be displayed immediately (with useMemo) instead of waiting for an effect to run.

This allows navigations to show the widget without showing a flash of old state.

### DIFF
--- a/src/components/Widget/inputs.tsx
+++ b/src/components/Widget/inputs.tsx
@@ -104,7 +104,8 @@ export function useSyncWidgetInputs({
     () => ({
       type,
       amount,
-      // If the default has not yet been handled, preemptively disable the widget by passing no tokens.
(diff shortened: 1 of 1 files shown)


src/components/Widget/inputs.tsx
```diff
@@ -96,9 +104,10 @@ export function useSyncWidgetInputs({
     () => ({
       type,
       amount,
-      ...tokens,
+      // If the default has not yet been handled, preemptively disable the widget by passing no tokens.
```
zzmp, line 107: Add one more comment:

    Effectively, this resets the widget - avoiding rendering stale state - because with no tokens the skeleton will be rendered.

cartcrom (approved):

PR #5021, opened 2022-10-27
refactor: add suspense loading states for Relay on Collection Page

cbachmeier:
Relay does not support callbacks for loading state instead relying solely on Suspense. We'll need a top level page for Collection page loading state and for collection assets loading state which I created by exporting a lot of the loading elements used in individual components.


vm (approved):

src/nft/components/collection/CollectionPageLoading.tsx
```diff
@@ -0,0 +1,28 @@
+import { Box } from 'nft/components/Box'
```
zzmp, line 1: Can this be called `Skeleton.tsx`? We named the TokenDetails skeleton like this, and I'd like to be consistent across the codebase.

zzmp (changes requested):

src/nft/components/collection/CollectionPageLoading.tsx
cbachmeier: Renamed to `CollectionPageSkeleton` to match `TokenDetailsPageSkeleton`

zzmp (approved):

PR #5022, opened 2022-10-27
refactor: Add Collection Stats GraphQL query

cbachmeier:
- Adds a graphql replacement for the call to `/collections` to get the stats on a collection. 
  - Various fields are still unsupported by the Backend currently including marketplace and trait counts.
- Refactor GenieCollection type to remove unused fields
- Added several null checks to collection field
- Add sudo icon


src/nft/queries/genie/CollectionStatsFetcher.ts
```diff
@@ -55,5 +56,15 @@ export const CollectionStatsFetcher = async (addressOrName: string, recursive =
   })
 
   const data = await r.json()
-  return data?.data ? data.data[0] : {}
+  const collections = data?.data
+    ? data.data.map((collection: any) => {
+        console.log(collection.traits)
+        console.log(groupBy(collection.traits, 'trait_type'))
```
JFrankfurt, line 62: should drop these before merge

src/nft/components/profile/view/ProfilePage.tsx
```diff
@@ -152,7 +152,7 @@ export const ProfilePage = () => {
     if (ownerCollections?.length && collectionStats?.length) {
       const ownerCollectionsCopy = [...ownerCollections]
       for (const collection of ownerCollectionsCopy) {
-        const floorPrice = collectionStats.find((stat) => stat.address === collection.address)?.floorPrice
+        const floorPrice = collectionStats.find((stat) => stat.address === collection.address)?.stats?.floor_price //TODO update when changing walletStats endpoint to gql
```
JFrankfurt, line 155: is this todo still relevant?

src/nft/queries/genie/CollectionStatsFetcher.ts
```diff
@@ -55,5 +56,15 @@ export const CollectionStatsFetcher = async (addressOrName: string, recursive =
   })
 
   const data = await r.json()
-  return data?.data ? data.data[0] : {}
+  const collections = data?.data
+    ? data.data.map((collection: any) => {
```
JFrankfurt, line 60: I think you could do `data?.data.map(...) ?? {}` here

src/nft/components/collection/MarketplaceSelect.tsx
```diff
@@ -141,15 +146,17 @@ export const MarketplaceSelect = () => {
...
-            <MarketplaceItem
-              key={value}
-              title={title}
-              value={value}
-              count={marketCount?.[value] || 0}
-              {...{ addMarket, removeMarket, isMarketSelected: selectedMarkets.includes(value) }}
-            />
-          ))}
+          {Object.entries(marketPlaceItems)
+            .filter(([value]) => isNftGraphQl || !graphQlMarkets.includes(value))
```
JFrankfurt, line 150: it could be better to `useMemo` this so the filtering isn't happening every render if the conditions aren't changing.

src/nft/components/collection/CollectionStats.tsx
```diff
@@ -251,18 +251,18 @@ const StatsItem = ({ children, label, isMobile }: { children: ReactNode; label:
...
-    ? roundWholePercentage((stats.stats.num_owners / stats.stats.total_supply) * 100)
+    ? roundWholePercentage(((stats.stats.num_owners ?? 1) / (stats.stats.total_supply ?? 0)) * 100)
     : 0
-  const totalSupplyStr = stats.stats ? quantityFormatter(stats.stats.total_supply) : 0
+  const totalSupplyStr = stats.stats ? quantityFormatter(stats.stats.total_supply ?? 0) : 0
   const listedPercentageStr =
-    stats.stats && stats.stats.total_listings > 0
-      ? roundWholePercentage((stats.stats.total_listings / stats.stats.total_supply) * 100)
+    stats.stats && stats.stats.total_listings
+      ? roundWholePercentage((stats.stats.total_listings / (stats.stats.total_supply ?? 0)) * 100)
```
JFrankfurt, line 259: division by 0 fallback is spooki

src/graphql/data/nft/Collection.ts
```diff
@@ -0,0 +1,132 @@
...
+        startCursor
+      }
+    }
+  }
+`
+export function useCollectionQuery(address: string): GenieCollection | undefined {
+  const collectionQueryReference = loadQuery<CollectionQuery>(RelayEnvironment, collectionQuery, {
+    address,
+  })
+  const queryData = usePreloadedQuery<CollectionQuery>(collectionQuery, collectionQueryReference)
```
JFrankfurt, line 87: I don't think `load` and `usepreload` should be used like this. iirc `loadQuery` should be happening higher up the component tree once we know the query might need to run and `usePreloadedQuery` should be happening here on its own at the last possible time.

src/graphql/data/nft/Collection.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  const queryCollection = queryData.nftCollections?.edges[0].node
+  const traits = {} as Record<string, Trait[]>
+  if (queryCollection?.traits) {
+    for (const trait of queryCollection?.traits) {
```
JFrankfurt, line 92: nit: prefer mapping/forEach over `for ... of` because it's too close to `for in` and that's unsettling

src/graphql/data/nft/Collection.ts
```diff
@@ -0,0 +1,132 @@
...
+      : undefined,
+    traits,
+    // marketplaceCount: { marketplace: string; count: number }[], // TODO add when backend supports
+    imageUrl: queryCollection?.image?.url ?? undefined,
+    twitter: queryCollection?.twitterName ?? undefined,
+    instagram: queryCollection?.instagramName ?? undefined,
+    discordUrl: queryCollection?.discordUrl ?? undefined,
+    externalUrl: queryCollection?.homepageUrl ?? undefined,
+    rarityVerified: false, // TODO update when backend supports
+    // isFoundation: boolean, // TODO ask backend to add
```
JFrankfurt, line 130: the amount of nested conditional data mapping and fallbacks feels messy here--indicates to me that there should be a cleaner approach, or that the consumer/upstream data source should change. (i.e., mapping `queryCollection.twitterName` to `twitter` smells like a data problem that we shouldn't allow to continue.

Any thoughts?

src/nft/utils/urlParams.ts
```diff
@@ -100,9 +100,9 @@ const urlParamsUtils = {
       clonedQuery['traits'] = clonedQuery['traits'].map((queryTrait: string) => {
         const modifiedTrait = trimTraitStr(queryTrait.replace(/(")/g, ''))
         const [trait_type, trait_value] = modifiedTrait.split(',')
-        const traitInStats = collectionStats.traits.find(
-          (item) => item.trait_type === trait_type && item.trait_value === trait_value
-        )
+        const traitInStats =
+          collectionStats.traits &&
+          collectionStats.traits[trait_type].find((trait) => trait.trait_value === trait_value)
```
JFrankfurt, line 105: would `collectionStats?.traits` work here?

JFrankfurt (commented):
some suggestions/questions

src/graphql/data/nft/Collection.ts
cbachmeier: Switched to foreach

src/graphql/data/nft/Collection.ts
cbachmeier: I was following the standard set by the token queries but instead replaced `useLazyLoadedQuery` with `usePreloadedQuery` as recommended by: https://relay.dev/docs/v12.0.0/api-reference/use-lazy-load-query/

src/graphql/data/nft/Collection.ts
cbachmeier: For context, when deciding on doing this migration, we saw that there was a discrepancy between how data team wanted to store data moving forward and what was done previously/what the components have been built around. The compromise agreed upon was for fast delivery but allowing for the migration of the most data intensive calls was to create these mapping functions that map the graphql data to the types currently used in the app. This will be the interim solution for phase1 and plan to migrate to objects that more closely align with the graphql schemas in the future. Sent comments in slack for reasons behind the null check logic and how it can be remediated.

src/nft/components/collection/CollectionStats.tsx
cbachmeier: mixed up the divisor and dividend. fixed.

src/nft/components/explore/ActivityFeed.tsx
```diff
@@ -66,7 +66,7 @@ const ActivityRow = ({ event, index, current }: { event: ActivityEvent; index: n
   const navigate = useNavigate()
 
   const formattedPrice = useMemo(
-    () => (event.price ? putCommas(formatEthPrice(event.price)).toString() : null),
+    () => (event.price ? putCommas(formatEthPrice(event.price))?.toString() : null),
```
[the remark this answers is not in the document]
cbachmeier: NFT pricess have different price formatting logic than the rest of the app and different logic on different pages. Additionally this bit of code will be replaced in Alex's work. I was just making this compatible with the data source changes.

src/nft/components/profile/view/ProfilePage.tsx
cbachmeier: Yes, this is related to a different endpoint being migrated https://github.com/Uniswap/interface/pull/5053

src/nft/utils/urlParams.ts
cbachmeier: No, React won't compile without the assurance that the array you're indexing won't be undefined

src/graphql/data/nft/Collection.ts
JFrankfurt: Yeah it's def better to use `usePreloadedQuery` and I'd love for you to implement that pattern in this PR, but the way you've implemented it here is (I think!) a more verbose way to accomplish `useLazyLoadedQuery`. If you look at the query preloading docs: https://relay.dev/docs/v12.0.0/api-reference/use-preloaded-query/

the example is trying to get users to do `useQueryLoader` with a query (in this case `CollectionQuery`) to get an instance of `loadQuery<CollectionQuery>` that can be called by user action ahead of the render loop that actually needs that data. The code that does that in the linked example is: 
```
<Button
      onClick={() => loadQuery({id: '4'})}
      ...
```

So I think a better way to accomplish this would be to use `useQueryLoader` directly in the component containing a link/button/card whatever that is going to route a user to the page that needs data from `CollectionQuery`. (I don't know of our page design **actually has** such an opportunity to intelligently pre-load a query, but that's the intended usage.) If there's no such opportunity to pre-load the query ahead of its usage (instead of alongside it) then the correct implementation would (I think) be `useLazyLoadedQuery`.

cbachmeier: I don't think it really makes sense to query for all potential collection queries a user can go to. I will switch usage to useLazyLoaded.

src/graphql/data/nft/Collection.ts
```diff
@@ -0,0 +1,131 @@
...
+        endCursor
+        hasNextPage
+        hasPreviousPage
+        startCursor
+      }
+    }
+  }
+`
+export function useCollectionQuery(address: string): GenieCollection | undefined {
+  const collectionQueryReference = loadQuery<CollectionQuery>(RelayEnvironment, collectionQuery, {
```
zzmp, line 86: See https://github.com/Uniswap/interface/pull/5037#discussion_r1012093462

zzmp (changes requested):

src/graphql/data/nft/Collection.ts
cbachmeier: Updated

src/graphql/data/nft/Collection.ts
JFrankfurt: I thiiink the right solution would be to load it on click of the collection a user is trying to navigate to

zzmp (dismissed):

src/nft/pages/collection/index.tsx
```diff
@@ -31,12 +33,20 @@ const Collection = () => {
...
   const isBagExpanded = useBag((state) => state.bagExpanded)
+  const isNftGraphQl = useNftGraphQlFlag() === NftGraphQlVariant.Enabled
   const { chainId } = useWeb3React()
 
-  const { data: collectionStats, isLoading } = useQuery(['collectionStats', contractAddress], () =>
+  const { data: queryCollection, isLoading } = useQuery(['collectionStats', contractAddress], () =>
     CollectionStatsFetcher(contractAddress as string)
   )
 
+  const gqlCollection = useCollectionQuery(contractAddress as string)
```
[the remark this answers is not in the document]
cbachmeier: It is handled in this case, but in the future we will look to improve this performance.

src/graphql/data/nft/Collection.ts
```diff
@@ -0,0 +1,127 @@
...
+        hasPreviousPage
+        startCursor
+      }
+    }
+  }
+`
+export function useCollectionQuery(address: string): GenieCollection | undefined {
+  const queryData = useLazyLoadQuery<CollectionQuery>(collectionQuery, { address })
+
+  const queryCollection = queryData.nftCollections?.edges[0].node as any
```
JFrankfurt, line 87: drop any cast for descriptive type

cbachmeier: Swapping this for wrapping all child elements in null checks

src/graphql/data/nft/Collection.ts
```diff
@@ -0,0 +1,130 @@
...
+            }
+          }
+        }
+      }
+      pageInfo {
+        endCursor
+        hasNextPage
+        hasPreviousPage
+        startCursor
+      }
```
JFrankfurt, line 80: double checking cursors are supported? I think I remember reading that they are, but I know they were not when we were working on the /explore page.

src/graphql/data/nft/Collection.ts
```diff
@@ -0,0 +1,130 @@
...
+    address,
+    isVerified: queryCollection?.isVerified ?? undefined,
+    name: queryCollection?.name ?? undefined,
+    description: queryCollection?.description ?? undefined,
+    standard: queryCollection?.nftContracts ? queryCollection?.nftContracts[0]?.standard ?? undefined : undefined,
+    bannerImageUrl: queryCollection?.bannerImage?.url,
+    stats: queryCollection?.markets
+      ? {
+          num_owners: queryCollection?.markets[0]?.owners ?? undefined,
+          floor_price: queryCollection?.markets[0]?.floorPrice?.value ?? undefined,
```
JFrankfurt, line 110: nit: do the `queryCollection?.markets[0]` in one place and assign it to a shorter var like `market` so we don't have so much repetition here

src/nft/components/collection/CollectionStats.tsx
```diff
@@ -264,21 +265,26 @@ const statsLoadingSkeleton = (isMobile: boolean) =>
...
 
   // round daily volume & floorPrice to 3 decimals or less
-  const totalVolumeStr = volumeFormatter(stats.stats?.total_volume)
-  const floorPriceStr = floorFormatter(stats.floorPrice)
+  const totalVolumeStr = volumeFormatter(stats.stats?.total_volume ?? 0)
+  const floorPriceStr = floorFormatter(stats.stats?.floor_price ?? 0)
   const floorChangeStr =
-    stats.stats && stats.stats.one_day_floor_change ? Math.round(Math.abs(stats.stats.one_day_floor_change) * 100) : 0
+    stats.stats && stats.stats.one_day_floor_change
+      ? Math.round(Math.abs(stats.stats.one_day_floor_change) * (isNftGraphQl ? 1 : 100))
```
JFrankfurt, line 286: can you add a comment explaining the magic numbers here? would be good to have it tagged w/ a todo for removal if possible

src/nft/components/collection/CollectionStats.tsx
```diff
@@ -264,21 +265,26 @@ const statsLoadingSkeleton = (isMobile: boolean) =>
...
-    ? roundWholePercentage((stats.stats.num_owners / stats.stats.total_supply) * 100)
-    : 0
-  const totalSupplyStr = stats.stats ? quantityFormatter(stats.stats.total_supply) : 0
+  const isNftGraphQl = useNftGraphQlFlag() === NftGraphQlVariant.Enabled
+  const uniqueOwnersPercentage =
+    stats.stats && stats.stats.total_supply
+      ? roundWholePercentage(((stats.stats.num_owners ?? 0) / stats.stats.total_supply) * 100)
+      : 0
+  const totalSupplyStr = stats.stats ? quantityFormatter(stats.stats.total_supply ?? 0) : 0
+  console.log(stats?.stats?.total_listings, stats?.stats?.total_supply)
```
JFrankfurt, line 274: remove console.log

src/graphql/data/nft/Collection.ts
cbachmeier: You can see the pagination logic with cursors working with Assets query infinite scroll

JFrankfurt (approved):

PR #5025, opened 2022-10-27
feat: Search result events

gbugyis:
Enhancement + fixes for NavBar search logging events. Also adds types for results (trending, recent, etc.). Other properties (position and totalSuggestions) now reflect the entire dropdown results rather than just within a section (ie. position now refers to a result's position in the entire dropdown, not just within Trending Collections, Recent searches, etc.)

navbar_result_selected


src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -193,6 +201,7 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput, i
   }, [toggleOpen, hoveredIndex, totalSuggestions])
 
   useEffect(() => {
+    const eventProperties = { totalSuggestions, queryText }
```
lynnshaoyu, line 204: can we make sure we keep consistent snake casing for event property names, so total_suggestions, and query_text

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -202,6 +211,10 @@ export const SearchBarDropdown = ({ toggleOpen, tokens, collections, hasInput, i
             setHoveredIndex={setHoveredIndex}
             toggleOpen={toggleOpen}
             suggestions={tokens}
+            eventProperties={{
+              suggestionType: NavBarSearchTypes.TOKEN_SUGGESTION,
```
lynnshaoyu, line 215: should be snake case: suggestion_type

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -60,16 +62,15 @@ export const SearchBarDropdownSection = ({
...
+              index={index + startingIndex}
               traceEvent={() =>
-                sendAnalyticsEvent(EventName.NAVBAR_SEARCH_EXITED, {
-                  position: index,
-                  selected_type: 'collection',
-                  suggestion_count: suggestions.length,
+                sendAnalyticsEvent(EventName.NAVBAR_RESULT_SELECTED, {
+                  position: index + startingIndex,
                   selected_name: suggestion.name,
                   selected_address: suggestion.address,
```
lynnshaoyu, line 70: can this be more descriptive, ie: "selected_search_result_name". same with address

lynnshaoyu (changes requested):

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -60,16 +62,15 @@ export const SearchBarDropdownSection = ({
               isHovered={hoveredIndex === index + startingIndex}
               setHoveredIndex={setHoveredIndex}
               toggleOpen={toggleOpen}
+              index={index + startingIndex}
               traceEvent={() =>
```
lynnshaoyu, line 66: can you use the useTrace hook (example of how to use here:
https://github.com/Uniswap/interface/blob/449c323fbb50468f82b21422c369b3dc2ca267a0/src/components/Widget/index.tsx#L61) so we can preserve properties from the parent component's contexts? similar comment for the other use of sendAnalyticsEvent below

lynnshaoyu: yep, but you can include the useTrace properties with the other eventProperties, like this for example: 
` const trace = useTrace({ section: SectionName.WIDGET })
  const [initialQuoteDate, setInitialQuoteDate] = useState<Date>()
  const onInitialSwapQuote = useCallback(
    (trade: Trade<Currency, Currency, TradeType>) => {
      setInitialQuoteDate(new Date())
      const eventProperties = {
        // TODO(1416): Include undefined values.
        ...formatSwapQuoteReceivedEventProperties(
          trade,
          /* gasUseEstimateUSD= */ undefined,
          /* fetchingSwapQuoteStartTime= */ undefined
        ),
        ...trace,
      }
      sendAnalyticsEvent(EventName.SWAP_QUOTE_RECEIVED, eventProperties)
    },
    [trace]
  )`

from here: https://github.com/Uniswap/interface/blob/449c323fbb50468f82b21422c369b3dc2ca267a0/src/components/Widget/index.tsx#L61

lynnshaoyu (changes requested):
comment regarding useTrace

lynnshaoyu (approved):

PR #5027, opened 2022-10-27
fix: attempt to use amplitude's new webAttributionPlugin 

lynnshaoyu:
this doesn't really work but idk how to fix it, no documentation? 

https://www.docs.developers.amplitude.com/data/sdks/typescript-browser/#web-attribution-enrichment-plugin

### DIFF
--- a/src/analytics/index.ts
+++ b/src/analytics/index.ts
@@ -1,5 +1,5 @@
 import { add, Identify, identify, init, track } from '@amplitude/analytics-browser'
-import client from '@amplitude/analytics-browser'
+import * as client from '@amplitude/analytics-browser'
 import { webAttributionPlugin } from '@amplitude/plugin-web-attribution-browser'
 import { isProductionEnv } from 'utils/env'
 
--- a/src/analytics/index.ts
+++ b/src/analytics/index.ts
@@ -1,4 +1,5 @@
-import { add, AmplitudeBrowser, Identify, identify, init, track } from '@amplitude/analytics-browser'
+import { add, Identify, identify, init, track } from '@amplitude/analytics-browser'
+import client from '@amplitude/analytics-browser'
 import { webAttributionPlugin } from '@amplitude/plugin-web-attribution-browser'
 import { isProductionEnv } from 'utils/env'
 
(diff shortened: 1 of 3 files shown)


src/analytics/index.ts
```diff
@@ -1,4 +1,6 @@
-import { Identify, identify, init, track } from '@amplitude/analytics-browser'
+import { add, Identify, identify, init, track } from '@amplitude/analytics-browser'
+import client from '@amplitude/analytics-browser'
```
kevinpagtakhan, line 2: I think this should fix it

```suggestion
import * as client from '@amplitude/analytics-browser'
```

lynnshaoyu: Hi Kevin, appreciate your help. This is the error I'm getting now: 
Screen Shot 2022-10-27 at 3 28 17 PM

I think the way I'm importing webAttributionPlugin might be incorrect? 
`import { webAttributionPlugin } from '@amplitude/plugin-web-attribution-browser'
`

kevinpagtakhan: Ah, your code seems correct. I remember this issue which we have fixed in `@amplitude/analytics-browser@1.3.0`, I see your `yarn.lock` says you have `v1.1.4`. I'd suggest to bump to the latest and greatest 😄

lynnshaoyu: Hi Kevin, sorry for the late reply. Same error even when we update to the latest version 1.4.0. 

The ultimate goal is just to get our referral params again. 

Would you mind taking a look at our function "initializeAnalytics" in src/analytics/index.ts ? I believe this is where we set referral params to be included. Is our configuration missing anything that would cause referral params to be missing?

kevinpagtakhan: Unfortunately, I'm not able to reproduce the error `Type 'void' is not assignable to type 'undefined'` in the newer versions.

The configuration looks fine. It's a mystery to me why the campaign parameters are not being captured after an update. If you are testing locally, you can try to lower the configuration `sessionTimeout` (perhaps 1000s) to force the session to expire and re-capture the campaign parameters on init. Reminder that for non-plugin solution, campaign parameters are only captured on new session, and optionally on campaign parameter change (configuration `attribution.trackNewCampaigns`), not necessarily on every page load.

PR #5030, opened 2022-10-27
fix: one point price charts + added suspense

cartcrom:
Fixes [WEB-1746](https://uniswaplabs.atlassian.net/browse/WEB-1746) and refactors pricechart to use suspense


src/graphql/data/TokenPrice.ts
```diff
@@ -1,84 +1,60 @@
...
+          timestamp @required(action: LOG)
+          value @required(action: LOG)
         }
       }
     }
   }
 `
-
-export type PricePoint = { timestamp: number; value: number }
+export type PricePoint = NonNullable<
```
zzmp, line 23: I find this harder to read, and would prefer if it were explicit. If the types mismatch it should still be caught by the checker. Could you revert back to

    export type PricePoint = { timestamp: number; value: number }

src/graphql/data/TokenPrice.ts
```diff
@@ -1,84 +1,60 @@
...
+  NonNullable<NonNullable<NonNullable<TokenPriceQuery['response']['tokens']>[number]>['market']['priceHistory']>[number]
+>
 export type PriceDurations = Partial<Record<TimePeriod, PricePoint[]>>
 
-export function isPricePoint(p: { timestamp: number; value: number | null } | null): p is PricePoint {
-  return Boolean(p && p.value)
+export function isPricePoint(p: { timestamp: number; value: number } | null): p is PricePoint {
```
zzmp, line 28: The parameter type would be clearer as just `PricePoint | null`.

src/graphql/data/TokenPrice.ts
```diff
@@ -1,84 +1,60 @@
...
+    const priceHistory = queryData.tokens?.[0]?.market?.priceHistory?.filter(isPricePoint)
+    const currentPrice = queryData.tokens?.[0]?.market?.price?.value
 
-        setPrices(prices)
-      },
-    })
-    return () => {
-      setPrices(undefined)
-      subscription.unsubscribe()
+    // Append the current price to the end of the priceHistory array
```
zzmp, line 51: This comment belongs above L47. Also, "Appends", not "Append".

src/graphql/data/TokenPrice.ts
```diff
@@ -1,84 +1,60 @@
...
-            .forEach((prices) => prices?.push(latest))
-        }
+  return queryReference
+}
+
+export function usePreloadedTokenPriceQuery(queryReference: PreloadedQuery<TokenPriceQuery>): PricePoint[] | undefined {
+  const queryData = usePreloadedQuery<TokenPriceQuery>(tokenPriceQuery, queryReference)
+
+  const priceHistory = useMemo(() => {
+    const priceHistory = queryData.tokens?.[0]?.market?.priceHistory?.filter(isPricePoint)
```
zzmp, line 48: To make this clearer, pull out the shared variable. Also, avoid mutating the query data by spreading the priceHistory.

    const market = queryData.tokens?.[0]?.market
    const priceHistory = market?.filter(isPricePoint)
    const currentPrice = market?.price?.value
    if (Array.isArray(priceHistory) && currentPrice !== undefined) {
      const timestamp = Date.now() / 1000
      return [...priceHistory, { timestamp, value: currentPrice })
    }
    return priceHistory

src/graphql/data/TokenPrice.ts
```diff
@@ -1,84 +1,60 @@
...
+export function isPricePoint(p: { timestamp: number; value: number } | null): p is PricePoint {
+  return p !== null
 }
 
-export function useTokenPriceQuery(address: string, chain: Chain): PriceDurations | undefined {
+export function useLoadTokenPriceQuery(address: string, chain: Chain, timePeriod: TimePeriod) {
+  const duration = toHistoryDuration(timePeriod)
   const contract = useMemo(() => ({ address: address.toLowerCase(), chain }), [address, chain])
-  const [prices, setPrices] = useState<PriceDurations>()
+  const queryReference = loadQuery<TokenPriceQuery>(useRelayEnvironment(), tokenPriceQuery, { contract, duration })
```
zzmp, line 35: `loadQuery` is only meant to be called on load (ie from a click event), and only meant to be called outside of the render cycle. This should instead use `useLazyLoadQuery`, and just rename `useLoadTokenPriceQuery` to `useLazyLoadTokenPriceQuery`, because the intention is to load the query once you are on the page.

src/pages/TokenDetails/index.tsx
```diff
@@ -35,8 +39,14 @@ export default function TokenDetails() {
...
-  const tokenQueryData = useTokenQuery(isNative ? nativeCurrency.wrapped.address : tokenAddress ?? '', chain)
-  const prices = useTokenPriceQuery(isNative ? nativeCurrency.wrapped.address : tokenAddress ?? '', chain)
+  const timePeriod = useAtomValue(filterTimeAtom)
+  const tokenQueryReference = useLoadTokenQuery(isNative ? nativeCurrency.wrapped.address : tokenAddress ?? '', chain)
+  const priceQueryReference = useLoadTokenPriceQuery(
+    isNative ? nativeCurrency.wrapped.address : tokenAddress ?? '',
+    chain,
+    timePeriod
+  )
+  const tokenQueryData = usePreloadedQuery<TokenQuery>(tokenQuery, tokenQueryReference).tokens?.[0]
```
zzmp, line 49: Because you are using the preloaded query immediately after triggering, you should just useLazyLoadQuery.

src/pages/TokenDetails/index.tsx
```diff
@@ -94,12 +104,14 @@ export default function TokenDetails() {
             <BreadcrumbNavLink to={`/tokens/${chainName}`}>
               <ArrowLeft size={14} /> Tokens
             </BreadcrumbNavLink>
-            <ChartSection
-              token={tokenQueryData}
-              currency={token}
-              nativeCurrency={isNative ? nativeCurrency : undefined}
-              prices={prices}
-            />
+            <Suspense fallback={<LoadingChart />}>
```
zzmp, line 107: This should be a part of `ChartSection`, so that any loading ux is a part of the component itself.

src/components/Tokens/TokenDetails/PriceChart.tsx
```diff
@@ -382,7 +382,7 @@ export function PriceChart({ width, height, prices }: PriceChartProps) {
               key={DISPLAYS[time]}
               active={timePeriod === time}
               onClick={() => {
-                setTimePeriod(time)
+                startTransition(() => setTimePeriod(time))
```
zzmp, line 385: Why do you do this? Don't you _want_ Suspense to kick in?

src/components/Tokens/TokenDetails/Skeleton.tsx
```diff
@@ -128,7 +128,7 @@ function Wave() {
   )
 }
 
-function LoadingChart() {
+export function LoadingChart() {
```
zzmp, line 131: This file structure is arguably wrong (but this is tech debt).

Ideally, I'd like to see the skeletons live in a file in the same folder as the component that they reflect. Eg, for `ChartSection`, make `ChartSection` a folder and then have `ChartSection/Skeleton.tsx`.

I wouldn't change this in this PR, but it's something to keep in mind - generally, I think the Skeleton should live in the same folder as the component that it reflects.

zzmp (changes requested):

src/graphql/data/Token.ts
```diff
@@ -61,10 +61,16 @@ const tokenQuery = graphql`
 
 export type TokenQueryData = NonNullable<TokenQuery$data['tokens']>[number]
 
-export function useTokenQuery(address: string, chain: Chain): TokenQueryData | undefined {
+export function useLoadTokenQuery(address: string, chain: Chain) {
   const contract = useMemo(() => ({ address: address.toLowerCase(), chain }), [address, chain])
-  const token = useLazyLoadQuery<TokenQuery>(tokenQuery, { contract }).tokens?.[0]
-  return token
+  const environment = useRelayEnvironment()
+  const queryReference = loadQuery<TokenQuery>(environment, tokenQuery, { contract })
```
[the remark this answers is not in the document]
cartcrom: Thanks for catching, I realized wrapped the hook is unnecessary now altogether. I'm switching TokenDetails to now just directly call useLazyLoadQuery

src/graphql/data/TopTokens.ts
```diff
@@ -54,8 +54,8 @@ const tokenSparklineQuery = graphql`
       address
       market(currency: USD) {
         priceHistory(duration: $duration) {
-          timestamp
-          value
+          timestamp @required(action: LOG)
```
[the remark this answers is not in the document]
cartcrom: I created a logger in my draft PR for switching the search bar to GQL, I'll add it into this PR

src/components/Tokens/TokenDetails/PriceChart.tsx
cartcrom: We want the current chart line to remain and then re-render once the new line is ready, rather than flash a loading state. Eventually the intention is to animate the first line into the second line.

src/components/Tokens/TokenDetails/Skeleton.tsx
cartcrom: sounds good

src/pages/TokenDetails/index.tsx
cartcrom: A suspense wrapper must be a parent component, rather a wrapper within the component itself, for it to catch any suspense that is thrown

src/graphql/data/TopTokens.ts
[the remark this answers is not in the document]
zzmp: This should be it's own PR - it's a distinct functionality that can be implemented/tested separately.

src/pages/TokenDetails/index.tsx
zzmp: You could use the preloaded prices query from within the `ChartSection` so that it can still catch the thrown suspense.
This should be fine - I don't think prices are used elsewhere. It's also an idiomatic usage of using a preloaded query from children components, with the parent component doing the actual fetch.

src/components/Tokens/TokenDetails/PriceChart.tsx
zzmp: You should add a comment to make this clear, just something simple like your first sentence here.

src/graphql/data/TopTokens.ts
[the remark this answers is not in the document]
cartcrom: Do you think it makes sense to leave my minimal implementation w/ console.err in this PR for now (before doing a full implementation later)? Else I realized the app crashes if it comes across a missing field and has no logger

src/components/Tokens/TokenDetails/ChartSection.tsx
```diff
@@ -1,112 +1,33 @@
...
+  priceQueryReference?: PreloadedQuery<TokenPriceQuery> | null
 }) {
-  const chainId = CHAIN_NAME_TO_CHAIN_ID[token.chain]
-  const L2Icon = getChainInfo(chainId)?.circleLogoUrl
-  const timePeriod = useAtomValue(filterTimeAtom)
-
-  const logoSrc = useTokenLogoURI(token, nativeCurrency)
+  return (
+    <Suspense fallback={<LoadingChart />}>
+      {priceQueryReference && (
```
zzmp, line 17: Shouldn't `Chart` always be rendered?

src/components/Tokens/TokenDetails/ChartSection.tsx
```diff
@@ -1,112 +1,33 @@
...
-        </TokenActions>
-      </TokenInfoContainer>
-      <ChartContainer>
-        <ParentSize>
-          {({ width }) => <PriceChart prices={prices ? prices?.[timePeriod] : null} width={width} height={436} />}
-        </ParentSize>
-      </ChartContainer>
-    </ChartHeader>
+    <ChartContainer>
+      <ParentSize>{({ width }) => <PriceChart prices={prices ?? null} width={width} height={436} />}</ParentSize>
```
zzmp, line 30: If there is no `priceQueryReference`, then should you render `<LoadingChart />`?

src/pages/TokenDetails/index.tsx
```diff
@@ -94,12 +175,31 @@ export default function TokenDetails() {
             <BreadcrumbNavLink to={`/tokens/${chainName}`}>
               <ArrowLeft size={14} /> Tokens
             </BreadcrumbNavLink>
-            <ChartSection
-              token={tokenQueryData}
-              currency={token}
-              nativeCurrency={isNative ? nativeCurrency : undefined}
-              prices={prices}
-            />
+            <ChartHeader>
```
zzmp, line 178: This is already a large file - can ChartHeader be moved to `components/Tokens/TokenDetails/ChartHeader.tsx`? Alternatively, have this be `components/Tokens/TokenDetails/Chart.tsx` and nest `ChartSection.tsx` within this.

zzmp (changes requested):
Why is `pages/TokenDetails` split from `components/Tokens/TokenDetails`?

src/graphql/data/TokenPrice.ts
```diff
@@ -1,84 +1,23 @@
...
-      },
-    })
-    return () => {
-      setPrices(undefined)
-      subscription.unsubscribe()
-    }
-  }, [contract])
 
-  return prices
+export function isPricePoint(p: PricePoint | null): p is PricePoint {
```
zzmp, line 21: This should probably go to `./utils.ts`, as it is shared between queries.

src/pages/TokenDetails/index.tsx
```diff
@@ -1,150 +1,55 @@
...
-import { useIsUserAddedTokenOnChain } from 'hooks/Tokens'
-import { useOnGlobalChainSwitch } from 'hooks/useGlobalChainSwitch'
-import { useCallback, useMemo, useState, useTransition } from 'react'
-import { ArrowLeft } from 'react-feather'
-import { useNavigate, useParams } from 'react-router-dom'
+import { filterTimeAtom } from 'components/Tokens/state'
+import TokenDetails from 'components/Tokens/TokenDetails'
+import { TokenDetailsPageSkeleton } from 'components/Tokens/TokenDetails/Skeleton'
+import { NATIVE_CHAIN_ID, nativeOnChain } from 'constants/tokens'
+import { TokenPriceQuery } from 'graphql/data/__generated__/TokenPriceQuery.graphql'
```
zzmp, line 5: TokenPriceQuery and TokenQuery should be re-exported through `graphql/data/{Token,TokenPrice}` to avoid having to find them in a generated folder.

You _could_ alias `useQueryLoader<TokenQuery>(tokenQuery)` to `useTokenQueryLoader` (similarly with `usePreloadedTokenQuery`), but that's up to you.

src/components/Tokens/TokenDetails/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+  tokenQueryReference: PreloadedQuery<TokenQuery>
+  priceQueryReference: PreloadedQuery<TokenPriceQuery> | null | undefined
+  refetchTokenPrices: RefetchPricesFunction
+}
+export default function TokenDetails({
+  tokenQueryReference,
+  priceQueryReference,
+  refetchTokenPrices,
+}: TokenDetailsProps) {
+  const { tokenAddress, chainName } = useParams<{ tokenAddress?: string; chainName?: string }>()
```
zzmp, line 75: Page params should probably be passed from the page.

src/components/Tokens/TokenDetails/ChartSection.tsx
```diff
@@ -1,112 +1,77 @@
...
+  const prices = usePreloadedTokenPriceQuery(priceQueryReference)
+  // Initializes time period to global & maintain separate time period for subsequent changes
+  const [timePeriod, setTimePeriod] = useState(useAtomValue(filterTimeAtom))
+
+  return (
+    <ChartContainer>
+      <ParentSize>
+        {({ width }) => <PriceChart prices={prices ?? null} width={width} height={436} timePeriod={timePeriod} />}
+      </ParentSize>
+      <TimePeriodSelector
```
zzmp, line 68: You may want this to be visible while suspended.

zzmp (changes requested):

src/components/Tokens/TokenDetails/ChartSection.tsx
cartcrom: The design has it not appearing on initial load, and since we're using a startTransition between time period changes, the time selector stays.

zzmp (approved):

PR #5035, opened 2022-10-27
fix: when scrolled, remove blur on nav bar

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-1565?atlOrigin=eyJpIjoiMmRlM2MzYWM2NGMyNGIyMGE4MmYwMGZiNzA2MzA0NGUiLCJwIjoiaiJ9

### DIFF
--- a/src/components/WalletDropdown/AuthenticatedHeader.tsx
+++ b/src/components/WalletDropdown/AuthenticatedHeader.tsx
@@ -139,8 +139,8 @@ const AuthenticatedHeader = () => {
 
   return (
     <AuthenticatedHeaderWrapper>
-      <HeaderWrapper>
-        <StatusWrapper>
+      <HeaderWrapper className="header-wrapper">
+        <StatusWrapper className="something">
           <FlexContainer>
             <StatusIcon connectionType={connectionType} size={24} />
             <Text fontSize={16} fontWeight={600} marginTop="2.5px">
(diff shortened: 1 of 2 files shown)


src/components/WalletDropdown/AuthenticatedHeader.tsx
```diff
@@ -139,8 +139,8 @@ const AuthenticatedHeader = () => {
 
   return (
     <AuthenticatedHeaderWrapper>
-      <HeaderWrapper>
-        <StatusWrapper>
+      <HeaderWrapper className="header-wrapper">
+        <StatusWrapper className="something">
```
gbugyis, line 143: are these classNames necessary?

gbugyis (changes requested):

src/pages/App.tsx
```diff
@@ -65,11 +65,15 @@ const BodyWrapper = styled.div`
   `};
 `
 
-const HeaderWrapper = styled.div`
+const HeaderWrapper = styled.div<{ scrolledState?: boolean }>`
```
cbachmeier, line 68: We only want this scroll behaviour on NFT Pages not all pages

cbachmeier (changes requested):

src/pages/App.tsx
lynnshaoyu: switched to be gated by nft flag / phase 1 enabled, as discussed in slack.

src/components/WalletDropdown/AuthenticatedHeader.tsx
lynnshaoyu: oops my comment didnt get saved earlier - they totally are not necessary lol, thank you for catching

gbugyis (approved):

PR #5037, opened 2022-10-28
refactor: Asset Details GQL Query

cbachmeier:
- Adds GraphQL query for Asset Details
- Makes a lot of GenieAsset params optional and adds needed checks
- Adds support for pooled assets
  - The sell order types from the v3 endpoint and graphql so there is some necessary type checking logic added to calculate pooled asset prices. This will able to be simplified when the migration is complete and the old data type can be removed.


src/graphql/data/nft/Details.ts
```diff
@@ -0,0 +1,160 @@
...
+            value
+          }
+        }
+      }
+    }
+  }
+`
+
+export function useDetailsQuery(address: string, tokenId: string): [GenieAsset, CollectionInfoForAsset] | undefined {
+  const detailsQueryReference = loadQuery<DetailsQuery>(RelayEnvironment, detailsQuery, {
```
zzmp, line 93: You should not be using `loadQuery` directly, as you'll then need to deal with disposing of it as well.

In this case, it looks like you may want `useLazyLoadQuery` (instead of `loadQuery`/`usePreloadedQuery`) because `usePreloadedQuery` would put the component in suspense anyways.

In most cases, you probably want `useQueryLoader` at the top-level parent (probably the `pages/<PAGE_NAME>/index.tsx` file), and then `usePreloadedQuery` within the component using the data, so that only that component is suspended should the query not yet be loaded.

zzmp (dismissed):

src/nft/types/sell/sell.ts
```diff
@@ -10,7 +12,7 @@ export interface ListingWarning {
   message: string
 }
 
-export interface SellOrder {
+export interface OldSellOrder {
```
JackShort, line 15: A better name might be `Deprecated_SellOrder`?

src/nft/utils/calcPoolPrice.ts
```diff
@@ -1,12 +1,19 @@
...
+  if (!asset.sellorders) return ''
+
+  const nft =
+    (asset.sellorders[0] as OldSellOrder).ammFeePercent === undefined
+      ? (asset.sellorders[0] as SellOrder).protocolParameters
+      : (asset.sellorders[0] as OldSellOrder)
+
   const decimals = BigNumber.from(1).mul(10).pow(18)
-  const ammFee = nft.ammFeePercent ? (100 + nft.ammFeePercent) * 100 : 110 * 100
+  const ammFee = nft.ammFeePercent ? ((100 + (nft.ammFeePercent as number)) as number) * 100 : 110 * 100
```
JackShort, line 16: nit: don't think you really need the second `number` cast

src/nft/utils/calcPoolPrice.ts
```diff
@@ -71,6 +102,7 @@ export const recalculateBagUsingPooledAssets = (uncheckedItemsInBag: BagItem[])
   const possibleMarkets = itemsInBag.reduce((markets, item) => {
     const asset = item.asset
     const market = asset.marketplace
+    if (!market) return {}
```
JackShort, line 105: should actually return `markets` since that is the accumulator

src/nft/utils/calcPoolPrice.ts
```diff
@@ -85,6 +117,7 @@ export const recalculateBagUsingPooledAssets = (uncheckedItemsInBag: BagItem[])
   const updatedPriceMarkets = itemsInBag.reduce((markets, item) => {
     const asset = item.asset
     const market = asset.marketplace
+    if (!market) return {}
```
JackShort, line 120: same here

JackShort (changes requested):

PR #5038, opened 2022-10-28
docs: adding comment to env

aballerr:
https://uniswapteam.slack.com/archives/C023ZR8Q1DE/p1666967317793309

Any thoughts on the verbiage? or if I should add more explicit comments?

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,4 @@
-# We are well aware these api keys are public, please do not report them, it is intentional.
+# We are aware these api keys are public, please do not report them, it is intentional.
 REACT_APP_INFURA_KEY="4bf032f2d38a4ed6bb975b80d6340847"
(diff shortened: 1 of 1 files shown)


.env
```diff
@@ -1,3 +1,4 @@
+# We are aware these api keys are public, please do not report them, it is intentional.
```
cbachmeier, line 1: "Thank you for your concern, these api keys are intentionally public. Please do not report them." ?

.env
```diff
@@ -1,3 +1,4 @@
+# Thank you for your concern, these api keys are intentionally public. Please do not report them.
```
zzmp, line 1: Put the statement of fact in front:

    These API keys are intentionally public. Please do not report them - thank you for your concern.

vm (approved):
smart

PR #5042, opened 2022-10-30
feat: explore carousel

JackShort:
* implementing the updated explore table

PR #5043, opened 2022-10-30
style: updating explore table

JackShort:
* updating the style of the explore table to match the new figma - meant to interact with # # #5042


src/nft/components/explore/Cells/Cells.tsx
```diff
@@ -19,48 +69,85 @@ interface CellProps {
...
+  value,
+  denomination,
+  usdPrice,
+}: {
+  value?: number
+  denomination: Denomination
+  usdPrice?: number
+}) => {
+  const denominatedValue =
+    denomination === Denomination.ETH
```
vm, line 100: can you make this a function and do if statements? this is hard to follow

src/nft/components/explore/Cells/Cells.tsx
```diff
@@ -19,48 +69,85 @@ interface CellProps {
...
+      ? value
+      : usdPrice && value
+      ? usdPrice * parseFloat(formatEther(value))
+      : undefined
+  const formattedValue = denominatedValue
```
vm, line 105: should we use the formatting we have in the rest of the app?

src/nft/components/explore/Cells/Cells.tsx
```diff
@@ -19,48 +69,85 @@ interface CellProps {
...
+export const VolumeCell = ({
+  value,
+  denomination,
+  usdPrice,
+}: {
+  value?: number
+  denomination: Denomination
+  usdPrice?: number
+}) => {
+  const denominatedValue = denomination === Denomination.ETH ? value : usdPrice && value ? usdPrice * value : undefined
```
vm, line 127: i think this is the name as above

src/nft/components/explore/Cells/Cells.tsx
```diff
@@ -19,48 +69,85 @@ interface CellProps {
...
+
+  const formattedValue = denominatedValue
```
vm, line 129: also same as above. perhaps a function?

src/nft/components/explore/Cells/Cells.tsx
JackShort: no they requested special formatting for the different cases here

src/nft/components/explore/CollectionTable.tsx
```diff
@@ -1,82 +1,133 @@
...
+
+  const volumeSort = useMemo(() => {
+    return (rowA: Row<CollectionTableColumn>, rowB: Row<CollectionTableColumn>) => {
+      return Math.round(rowA.original.volume.value * 100000) >= Math.round(rowB.original.volume.value * 100000) ? 1 : -1
+    }
+  }, [])
+
+  const volumeChangeSort = useMemo(() => {
+    return (rowA: Row<CollectionTableColumn>, rowB: Row<CollectionTableColumn>) => {
+      return Math.round(rowA.original.volume.change * 100000) >= Math.round(rowB.original.volume.change * 100000)
```
vm, line 43: this should be a function above that you can call in each of these

src/nft/components/explore/CollectionTable.tsx
```diff
@@ -1,82 +1,133 @@
...
-
 const CollectionTable = ({ data }: { data: CollectionTableColumn[] }) => {
+  const floorSort = useMemo(() => {
+    return (rowA: Row<CollectionTableColumn>, rowB: Row<CollectionTableColumn>) => {
+      const aFloor = BigNumber.from(rowA.original.floor.value)
+      const bFloor = BigNumber.from(rowB.original.floor.value)
+
+      return aFloor.gte(bFloor) ? 1 : -1
+    }
+  }, [])
```
vm, line 27: does this mean you only do this calculation once and don't update on prop changes?

src/nft/components/explore/Table.tsx
```diff
@@ -50,12 +108,98 @@ export function Table<D extends Record<string, unknown>>({
...
+
+    if (width < theme.breakpoint.sm) {
+      setHiddenColumns([
+        ColumnHeaders.Items,
+        ColumnHeaders.FloorChange,
+        ColumnHeaders.Volume,
+        ColumnHeaders.VolumeChange,
+        ColumnHeaders.Owners,
+      ])
+      setMaxCols(MAX_COLUMNS - 5)
```
vm, line 121: i would do this based on the length of the hidden columns. you can make that a variable

vm: so maybe in each of these you can just do `hiddenColumns = [...]`

then outside of the if statements you can do ```ts
setHiddenColumns(hiddenColumns)
setMaxCols(MAX_COLUMNS - hiddenColumns.length)
````

Comment (vm) [reply src/nft/components/explore/Table.tsx:121]: but also i don't think you need maxCols? you can just get that information based on the hidden columns at render instead of needing new duplicated state for it

File: src/nft/components/explore/Table.tsx [hunk 8]
```diff
@@ -50,12 +108,98 @@ export function Table<D extends Record<string, unknown>>({
...
     } else {
       setHiddenColumns([])
+      setMaxCols(MAX_COLUMNS)
     }
-  }, [hiddenColumns, setHiddenColumns])
+  }, [width, setHiddenColumns, theme.breakpoint])
+
+  if (data.length === 0) {
+    return (
+      <table {...getTableProps()} className={styles.table}>
```
vm, line 141: nit: maybe make a separate component called `LoadingTable` for this

src/nft/components/explore/Table.tsx
```diff
@@ -98,29 +243,28 @@ export function Table<D extends Record<string, unknown>>({
...
           return (
-            <TraceEvent
-              events={[Event.onClick]}
-              name={EventName.NFT_TRENDING_ROW_SELECTED}
-              properties={{ collection_address: row.original.collection.address, chain_id: chainId }}
-              element={ElementName.NFT_TRENDING_ROW}
+            <StyledRow
+              {...row.getRowProps()}
               key={i}
+              onClick={() => navigate(`/nfts/collection/${row.original.collection.address}`)}
```
vm, line 249: use a Link component for this

src/nft/components/explore/Table.tsx
```diff
@@ -98,29 +243,28 @@ export function Table<D extends Record<string, unknown>>({
...
-            </TraceEvent>
+              {row.cells.map((cell, cellIndex) => {
+                return (
+                  <td className={clsx(styles.td, classNames?.td)} {...cell.getCellProps()} key={cellIndex}>
+                    {cellIndex === 0 ? (
+                      <RankCellContainer>
+                        <ThemedText.BodySecondary fontSize="14px" lineHeight="20px">
+                          {i + 1}
+                        </ThemedText.BodySecondary>
+                        {cell.render('Cell')}
```
vm, line 259: having trouble following, what is this doing?

JackShort: this renders the line number for the cell

src/nft/components/explore/CollectionTable.tsx
JackShort: yes - you technically don't need to memo it the react-docs just rec that you do and like this

src/nft/components/explore/Table.tsx
JackShort: you need maxCols because you aren't hiding all of the hiddenColumns at once - different breakpoints have a different amount of hidden columns

src/nft/pages/explore/index.tsx
```diff
@@ -1,16 +1,21 @@
...
 
 const NftExplore = () => {
   return (
-    <>
-      <Trace page={PageName.NFT_EXPLORE_PAGE} shouldLogImpression>
-        <Banner />
-        <TrendingCollections />
-      </Trace>
-    </>
+    <ExploreContainer>
```
vm, line 15: trace got removed

PR #5047, opened 2022-10-31
style: updated collection cards

JackShort:
* updated the collection cards
* ignore the functionality for subtraction of erc1155 its setup for the handlingErc1155 ticket

### DIFF
--- a/src/graphql/data/nft/Asset.ts
+++ b/src/graphql/data/nft/Asset.ts
@@ -154,7 +154,7 @@ export function useAssetsQuery(
             basePrice: ethPrice,
           }
         : undefined,
-      susFlag: asset.suspiciousFlag,
+      openseaSusFlag: asset.suspiciousFlag,
       sellorders: asset.listings?.edges,
       smallImageUrl: asset.smallImage?.url,
       tokenId: asset.tokenId,
--- a/src/graphql/data/nft/Asset.ts
+++ b/src/graphql/data/nft/Asset.ts
@@ -154,7 +154,7 @@ export function useAssetsQuery(
             basePrice: ethPrice,
           }
         : undefined,
-      openseaSusFlag: asset.suspiciousFlag,
+      susFlag: asset.suspiciousFlag,
       sellorders: asset.listings?.edges,
       smallImageUrl: asset.smallImage?.url,
       tokenId: asset.tokenId,
--- a/src/nft/components/collection/Card.tsx
+++ b/src/nft/components/collection/Card.tsx
@@ -40,6 +40,8 @@ export interface CardContextProps {
   selected: boolean
   href: string
   setHref: (href: string) => void
+  addAssetToBag: () => void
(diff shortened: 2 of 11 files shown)


src/graphql/data/nft/Asset.ts
```diff
@@ -154,7 +154,7 @@ export function useAssetsQuery(
             basePrice: ethPrice,
           }
         : undefined,
-      susFlag: asset.suspiciousFlag,
```
cbachmeier, line 157: this was an intentional refactor to not be OS specific

src/nft/components/collection/Card.tsx
```diff
@@ -82,23 +154,35 @@ const Container = ({ asset, selected, children }: CardProps) => {
...
         transition="250"
+        cursor={asset.notForSale ? 'default' : 'pointer'}
+        onClick={(e: MouseEvent) => {
+          if (!asset.notForSale) {
+            e.preventDefault()
+            !selected ? addToBag() : removeFromBag()
+            return true
+          } else {
+            return false
+          }
```
cbachmeier, line 173: why do we need to return true/false?

src/nft/components/collection/Card.tsx
```diff
@@ -425,100 +513,25 @@ const TertiaryInfo = ({ children }: { children: ReactNode }) => {
...
-                ? 'backgroundInteractive'
-                : selected
-                ? 'accentFailure'
-                : 'accentAction'
-              : asset.notForSale
-              ? 'backgroundModule'
-              : selected
-              ? 'accentFailureSoft'
-              : 'accentActionSoft'
-          }
```
cbachmeier, line 473: rip the flow diagrams I had to write for this logic

src/nft/components/collection/CollectionAsset.tsx
```diff
@@ -73,39 +73,66 @@ export const CollectionAsset = ({
...
-      )}
+    <Card.Container
+      asset={asset}
+      selected={isSelected}
+      addToBag={() => {
+        addAssetsToBag([asset])
+        !bagExpanded && !isMobile && toggleBag()
+      }}
+      removeFromBag={() => {
+        removeAssetsFromBag([asset])
```
cbachmeier, line 84: add and remove to/from bag seems to be the same for assets and can be derived from hooks. Can we remove them as input parameters and implement directly in Card.Container?

cbachmeier (changes requested):

src/nft/components/collection/Card.tsx
JackShort: F

src/nft/components/collection/Card.tsx
JackShort: this was needed for the href but no longer needed

cbachmeier (approved):
LGTM

PR #5049, opened 2022-10-31
fix: stop nft id tag from getting cut off 

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-1488?atlOrigin=eyJpIjoiMTA3ZmMxZGI3NzAzNDMwMzlmOWQ0YTFkZWUyZmFiOGIiLCJwIjoiaiJ9

### DIFF
--- a/src/nft/components/collection/Activity.css.ts
+++ b/src/nft/components/collection/Activity.css.ts
@@ -68,7 +68,7 @@ export const detailsImage = sprinkles({
 export const detailsName = style([
   body,
   sprinkles({
-    paddingBottom: '12',
+    height: 'inherit',
     fontWeight: 'normal',
     overflow: 'hidden',
     whiteSpace: 'nowrap',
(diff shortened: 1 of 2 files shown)


JackShort (changes requested):
the problem is that the rarity tooltip container is growing too large in the column. You can see this when you inspect element the container. the `marginBottom` is just used as a gap between the rarity tooltip you could replace that with `gap` in the column element. The `lineHeight` is also correct. Also the container should be centered in the column

lynnshaoyu:
> the problem is that the rarity tooltip container is growing too large in the column. You can see this when you inspect element the container. the `marginBottom` is just used as a gap between the rarity tooltip you could replace that with `gap` in the column element. The `lineHeight` is also correct. Also the container should be centered in the column

the lineHeight is already being applied by the "body" css, see line 69.

JackShort (approved):
lgtm

PR #5053, opened 2022-10-31
refactor: Wallet Assets GraphQL Query

cbachmeier:
- Adds a gql query for a user's owned assets 
- Removes some unused parameters from WalletAsset
- Waiting on BE for the following to finish:

- [x] Add support for collection filtering
- [x] Fetch schema from prod endpoint


gbugyis (changes requested):
Seeing some key prop console warnings: 
CleanShot 2022-11-08 at 02 07 26@2x

Also seeing some odd behavior on filters, and some undefined assets (sent details/screenshots via DM)

aballerr:
> Seeing some key prop console warnings: CleanShot 2022-11-08 at 02 07 26@2x
> 
> Also seeing some odd behavior on filters, and some undefined assets (sent details/screenshots via DM)

My pr might fix this, not 100% sure.  But basically we are getting repeat trait names that only differentiate by case. IE "Blue coat" and "blue coat"
https://github.com/Uniswap/interface/pull/5116/files

cbachmeier:
The index error was caused by the 25 loading cards not having a key. I've updated for both the profile page and collection page

gbugyis (approved):

PR #5056, opened 2022-11-01
feat(nft-bag): wrong-network-cta

JFrankfurt:
https://uniswaplabs.atlassian.net/browse/WEB-1730

I'd like to revisit this and consolidate the error messaging, click handler, and disabled state logical checks into one set. The current approach feels too error-prone to me.


src/nft/components/bag/Bag.tsx
```diff
@@ -54,7 +52,9 @@ const ScrollingIndicator = ({ top, show }: SeparatorProps) => (
 )
 
 const Bag = () => {
-  const { account } = useWeb3React()
+  const { account, provider } = useWeb3React()
+  const isConnected = !!provider && !!account
```
vm, line 56: i think you could check for `isActive`?

vm: ah i guess that wouldn't work since it may be the network provider. this works then

JFrankfurt: we don't do this anywhere currently, but I think we could use `Signer.isSigner(provider)`? https://docs.ethers.io/v5/api/signer/#Signer-isSigner

vm: we probably should do this lol. ill finishing reviewing now sorry got caught up

vm (approved):

src/nft/components/bag/BagFooter.tsx
```diff
@@ -74,9 +74,12 @@ const ActionButton = ({ bagStatus, onClick, walletModalIsOpen, sufficientBalance
   } else if (bagStatus === BagStatus.PROCESSING_TRANSACTION) {
     disabled = true
     buttonText = <Trans>Transaction pending</Trans>
-  } else if (sufficientBalance !== false) {
+  } else if (sufficientBalance === true) {
```
vm: can you just if (sufficientBalance)

src/nft/components/bag/BagFooter.tsx
```diff
@@ -74,9 +74,12 @@ const ActionButton = ({ bagStatus, onClick, walletModalIsOpen, sufficientBalance
...
     disabled = false
     buttonText = <Trans>Pay</Trans>
+  } else if (sufficientBalance === false) {
```
vm: `!sufficientBalance`

PR #5060, opened 2022-11-01
fix: remove blur transition as requested by bryan

lynnshaoyu:


src/pages/App.tsx
```diff
@@ -75,6 +75,10 @@ const HeaderWrapper = styled.div<{ scrolledState?: boolean; nftFlagEnabled?: boo
   width: 100%;
   justify-content: space-between;
   position: fixed;
+  transition: ${({ theme, nftFlagEnabled }) =>
+    nftFlagEnabled &&
```
vm, line 79: i don't think this is what we want? doesn't this mean when this is released this will happen for tokens too?

lynnshaoyu: yeah we want to remove it for all pages (tokens, nft, everything) in phase 1: https://uniswaplabs.atlassian.net/browse/WEB-1565

vm:
i think we should just merge this without the flag fwiw

PR #5062, opened 2022-11-02
feat: Buy Bag events

gbugyis:
Adds events for NFT Buy Bag (adding to bag, Pay, Bag Update (unavailable/price change), Bag signed, Success screen)


src/analytics/constants.ts
```diff
@@ -16,6 +16,12 @@ export enum EventName {
   NAVBAR_SEARCH_SELECTED = 'Navbar Search Selected',
   NAVBAR_SEARCH_EXITED = 'Navbar Search Exited',
   NFT_ACTIVITY_SELECTED = 'NFT Activity Selected',
+  NFT_BUY_ADDED = 'NFT Buy Bag Added',
+  NFT_BUY_BAG_CHANGED = 'NFT Buy Bag Changed',
+  NFT_BUY_BAG_PAY = 'NFT Buy Bag Pay Clicked',
+  NFT_BUY_BAG_REFUND = 'NFT Buy Bag Refund',
```
vm, line 22: Refunded

src/analytics/constants.ts
```diff
@@ -16,6 +16,12 @@ export enum EventName {
...
+  NFT_BUY_BAG_SIGNED = 'NFT Buy Bag Signed',
+  NFT_BUY_BAG_SUCCESS = 'NFT Buy Bag Success',
```
vm, line 24: Succeeded

src/nft/components/bag/BagContent.tsx
```diff
@@ -54,25 +57,47 @@ export const BagContent = () => {
...
+              usdPrice={fetchedPriceData}
+              clearUnavailableAssets={() => setItemsInBag(availableItems)}
+              didOpenUnavailableAssets={didOpenUnavailableAssets}
+              setDidOpenUnavailableAssets={setDidOpenUnavailableAssets}
+              isMobile={isMobile}
+            />
+          </Trace>
+        )}
+        {priceChangedAssets.length > 0 && (
+          <Trace
```
vm, line 80: would this be better to do with a useeffect?

src/nft/components/bag/BagFooter.tsx
```diff
@@ -48,6 +50,7 @@ interface BagFooterProps {
   bagStatus: BagStatus
   fetchAssets: () => void
   assetsAreInReview: boolean
+  eventProperties: Record<string, unknown>
```
vm, line 53: i think we would ideally type this, but if we're not doing that elsewhere then it's okay

vm (approved):
I think if we are okay with using Trace for modals then what we have works. Rest are nits.

src/nft/components/bag/BagContent.tsx
gbugyis: yes, good call

PR #5065, opened 2022-11-02
fix: show <.001 ETH in collection floors

vm:
This makes sure we show <0.001 ETH for any usage of this function.


JackShort (approved):
lgtm

PR #5070, opened 2022-11-02
fix: prices passed in scientific notation

cbachmeier:
Handle graphql prices passed in as scientific notation


lynnshaoyu (approved):
w00t

PR #5074, opened 2022-11-03
feat(bag): add check if the user has manually closed the bag

JFrankfurt:
wont reopen on asset add until bag is cleared

https://uniswaplabs.atlassian.net/browse/WEB-1489

this is my first zustand pr, so pay close attention to that and make sure I didn't screw anything up.


src/nft/hooks/useBag.ts
```diff
@@ -52,19 +55,22 @@ export const useBag = create<BagState>()(
...
           didOpenUnavailableAssets: didOpen,
         })),
-      bagExpanded: false,
       toggleBag: () =>
-        set(({ bagExpanded }) => ({
-          bagExpanded: !bagExpanded,
-        })),
+        set(({ bagExpanded }) => {
+          if (bagExpanded === false) {
+            return { bagExpanded: !bagExpanded, bagManuallyClosed: true }
```
JackShort, line 61: I don't think this is valid logic - sometimes we close the bag on purpose for instance if a user goes to another page we might toggle the bag - also if bagExpanded is false that means we are setting the bag to open now. Maybe a better way to go about this is by passing an optional param for manualClose that sets manually close if its there

JackShort, line 61: I don't think this is valid logic - sometimes we close the bag on purpose for instance if a user goes to another page we might toggle the bag - also if bagExpanded is false that means we are setting the bag to open now. Maybe a better way to go about this is by passing an optional param for manualClose that sets manually close if its there

JackShort (changes requested):

JackShort (approved):
i like this implementation - also ty for adding a setBagExpanded function

PR #5077, opened 2022-11-03
refactor: apollo integration for nft assets

cbachmeier:
- Adds Apollo integration
  - [documentation](https://www.apollographql.com/docs/react/get-started)
  - This was requested as Relay did not have any native support for custom polling for a paginated query, which is the desired behaviour for the nft assets query.
- Generates schema as part of the `yarn prepare` step
- Replace the Relay Assets Query with an Apollo query
  - Doesn't split up the query into a fragment
  - Custom poll interval
  - Required some extra work to define a fetchMore function which Relay has built in
    - That being said the schema was setup for Relay style pagination originally as opposed to Apollo standard limit & offset
    - [Apollo pagination documentation](https://www.apollographql.com/docs/react/pagination/core-api)
      - [Apollo cursor based pagination documentation](https://www.apollographql.com/docs/react/pagination/cursor-based/)


zzmp:
Can you make sure to point out exactly what apollo is required for in the PR description?

zzmp (changes requested):
When testing, I do not see any polling occuring.

cbachmeier:
@zzmp Can you try now. There's an open issue with pollInterval and ReactSctrictMode. Instead you can use the startPoll function returned from useQuery https://github.com/apollographql/apollo-client/issues/9819

zzmp:
I'm now seeing polling, but I'm only seeing the initial query being re-polled. The paged assets are not being re-polled.

cbachmeier:
Looks like Apollo out of the box only supports polling the first call. There is an [existing workaround](https://github.com/apollographql/apollo-client/issues/1121) which is now implemented.

codegen.ts
```diff
@@ -0,0 +1,17 @@
+import type { CodegenConfig } from '@graphql-codegen/cli'
+
+const config: CodegenConfig = {
+  overwrite: true,
+  schema: './src/graphql/data/schema.graphql',
+  documents: ['./src', '!**/node_modules/**', '!**/__mocks__/**', '!**/__generated__/**', '!**/thegraph/**'],
```
justindomingue, line 6: nit. i think it's cleaner to enforce queries to live in `ts` or `graphql` types explicitly (on mobile right now we enforce queries to all be in the same file, although i expect we will split it up post-migration   `documents: 'src/data/queries.graphql'`)

codegen.ts
```diff
@@ -0,0 +1,17 @@
...
+  generates: {
+    'src/graphql/data/__generated__/types-and-hooks.ts': {
```
justindomingue, line 8: should you check in this file?

src/graphql/data/nft/Asset.ts
```diff
@@ -117,22 +116,46 @@ export function useAssetsQuery(
...
-  const queryData = useLazyLoadQuery<AssetQuery>(assetQuery, {
-    address,
-    orderBy,
-    asc,
-    filter,
-    first,
-    after,
-    last,
-    before,
+  const queryResult = useQuery(assetQuery, {
```
justindomingue, line 119: in `types-and-hooks` you should find a custom hook for this query 😄 

```
useAssetQuery({
  variables: { ... },
  pollInterval: 999
})
```

justindomingue: also i would use the `pollInterval` config over `startPolling`

src/graphql/data/nft/Asset.ts
```diff
@@ -117,22 +116,46 @@ export function useAssetsQuery(
...
+      before,
+    },
   })
-  const { data, hasNext, loadNext, isLoadingNext } = usePaginationFragment<AssetPaginationQuery, any>(
-    assetPaginationQuery,
-    queryData
-  )
+  const { loading, data, fetchMore, startPolling } = queryResult
+
+  startPolling(999)
```
justindomingue, line 133: see above, i recommend `pollInterval`, but regardless, this should probably be called in a useEffect

```
useEffect(() => { startPolling(999) }, [])
```

src/graphql/data/nft/Asset.ts
```diff
@@ -117,22 +116,46 @@ export function useAssetsQuery(
...
+  const loadMore = () => {
+    fetchMore({
+      variables: {
+        after: data?.nftAssets?.pageInfo?.endCursor,
+      },
+    })
+  }
+
+  useEffect(() => {
+    const intervalId = setInterval(() => {
```
justindomingue, line 144: doesn't the interface codebase have a `useInterval` hook?

justindomingue: https://github.com/Uniswap/interface/blob/caa0a2967d169dbe75a42a9a2ef3ba8b736614ba/src/lib/hooks/useInterval.ts#L1

src/graphql/data/nft/Asset.ts
justindomingue: also nit. can use the ms macro to format intervals

```
ms`10s`
```

src/graphql/data/nft/Asset.ts
```diff
@@ -117,22 +116,46 @@ export function useAssetsQuery(
...
+      })
+    }, 999)
+
+    return () => clearInterval(intervalId)
+  }, [address, orderBy, asc, filter, first, after, last, before, queryResult.data?.nftAssets?.pageInfo?.endCursor])
+
+  const hasNext = data?.nftAssets?.pageInfo?.hasNextPage
 
-  const assets: GenieAsset[] = data.nftAssets?.edges?.map((queryAsset: { node: any }) => {
+  const assets: GenieAsset[] = data?.nftAssets?.edges?.map((queryAsset: { node: any }) => {
```
justindomingue, line 158: should probably memoize this since `map` will make it a diff object every render?

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -530,7 +526,7 @@ export const CollectionNfts = ({ contractAddress, collectionStats, rarityVerifie
         </Box>
       </AnimatedBox>
       <InfiniteScroll
-        next={() => (isNftGraphQl ? loadNext(DEFAULT_ASSET_QUERY_AMOUNT) : fetchNextPage())}
+        next={() => (isNftGraphQl ? loadMore() : fetchNextPage())}
```
justindomingue, line 529:
```suggestion
        next={isNftGraphQl ? loadMore : fetchNextPage}
```

assuming `loadMore` and `fetchNextPage` are memoized / referentially stable across renders, this would also keep the value passed to `next` referentially stable

src/graphql/data/nft/Asset.ts
justindomingue: oh just seeing this https://github.com/Uniswap/interface/pull/5077#issuecomment-1302959247

PR #5083, opened 2022-11-04
feat: nft polling

zzmp:
Implements NFT polling on a 999ms interval if the user has not yet loaded any additional pages.

- Fixes the RelayEnvironment to allow bypassing the cache (using `fetchQuery`), by removing the redundant cache around `fetchGraphQL` and relying solely on the store.
- Fixes Asset.ts to memoize returned assets, preventing redundant work in render cycles when they have not changed.
- Implements polling using the technique described in the relay docs: https://relay.dev/docs/guided-tour/refetching/refetching-queries-with-different-data/#if-you-need-to-avoid-suspense-1
  - The InfiniteScroll does not mark which asset the user has on-screen, and does not use window virtualization, so polling only occurs if the user has not yet loaded any additional pages. This is necessary to avoid loading arbitrarily large requests from the graphql endpoint.
  
  The entire page could be polled, but it would eventually take >1s to fetch. This is accomplishable by setting `first` to `data.nftAssets?.edges?.length`.
  
  Ideally, (a) the scroll would be virtualized, and (b) a cursor to the first visible asset would be passed to `useAssetsQuery`. This would allow the relevant page (after the asset's cursor) to be re-polled, without having to re-poll an arbitrarily large number of assets.

In general, polling every second seems like an anti-pattern: it uses a lot of user bandwidth, and can be achieved in other ways (eg pull down to refresh, which is supported by the InfiniteScroll component).

### DIFF
--- a/src/graphql/data/RelayEnvironment.ts
+++ b/src/graphql/data/RelayEnvironment.ts
@@ -1,38 +1,16 @@
 import ms from 'ms.macro'
-import { Variables } from 'react-relay'
-import { Environment, Network, RecordSource, RequestParameters, Store } from 'relay-runtime'
-import RelayQueryResponseCache from 'relay-runtime/lib/network/RelayQueryResponseCache'
+import { Environment, Network, RecordSource, Store } from 'relay-runtime'
 
 import fetchGraphQL from './fetchGraphQL'
-// max number of request in cache, least-recently updated entries purged first
-const size = 250
-// number in milliseconds, how long records stay valid in cache
-const ttl = ms`5m`
-export const cache = new RelayQueryResponseCache({ size, ttl })
 
-const fetchQuery = async function wrappedFetchQuery(params: RequestParameters, variables: Variables) {
-  const queryID = params.name
-  const cachedData = cache.get(queryID, variables)
+// This makes it possible (and more likely) to be able to reuse data when navigating back to a page,
+// tab or piece of content that has been visited before. These settings together configure the cache
+// to serve the last 250 records, so long as they are less than 5 minutes old:
+const gcReleaseBufferSize = 250
+const queryCacheExpirationTime = ms`5m`
 
-  if (cachedData !== null) return cachedData
-
-  return fetchGraphQL(params, variables).then((data) => {
-    if (params.operationKind !== 'mutation') {
-      cache.set(queryID, variables, data)
-    }
-    return data
-  })
-}
-// This property tells Relay to not immediately clear its cache when the user
-// navigates around the app. Relay will hold onto the specified number of
-// query results, allowing the user to return to recently visited pages
-// and reusing cached data if its available/fresh.
-const gcReleaseBufferSize = 10
-const queryCacheExpirationTime = ms`1m`
-const store = new Store(new RecordSource(), { gcReleaseBufferSize, queryCacheExpirationTime })
-const network = Network.create(fetchQuery)
-// Export a singleton instance of Relay Environment configured with our network function:
-export default new Environment({
-  network,
-  store,
+export const CachingEnvironment = new Environment({
+  network: Network.create(fetchGraphQL),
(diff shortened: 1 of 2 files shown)


src/graphql/data/RelayEnvironment.ts
```diff
@@ -1,38 +1,16 @@
...
-export const cache = new RelayQueryResponseCache({ size, ttl })
 
-const fetchQuery = async function wrappedFetchQuery(params: RequestParameters, variables: Variables) {
-  const queryID = params.name
-  const cachedData = cache.get(queryID, variables)
+// This makes it possible (and more likely) to be able to reuse data when navigating back to a page,
+// tab or piece of content that has been visited before. These settings together configure the cache
+// to serve the last 250 records, so long as they are less than 5 minutes old:
+const gcReleaseBufferSize = 250
+const queryCacheExpirationTime = ms`5m`
```
cbachmeier, line 10: any reason for changing `queryCacheExpirationTime` to 5m instead of the previous 1m?

src/graphql/data/nft/Asset.ts
```diff
@@ -117,74 +121,100 @@ export function useAssetsQuery(
...
+  // Poll for updates.
+  const POLLING_INTERVAL = ms`999ms`
+  const environment = useRelayEnvironment()
+  const refresh = useCallback(async () => {
+    const length = data.nftAssets?.edges?.length
+    // Only poll if the user has not started paging.
+    // Polling after paging requires reloading *all* pages, which is untenable.
+    // TODO(cbachmeier): Refactor CollectionNfts to pass the first visible cursor to useAssetsQuery,
+    // to allow for polling of the visible page only. This would make polling tenable after the user has paged.
+    if (length > DEFAULT_ASSET_QUERY_AMOUNT) return
```
cbachmeier, line 145: Do assets still refresh after the `queryCacheExpirationTime` expires?

src/graphql/data/nft/Asset.ts
```diff
@@ -117,74 +121,100 @@ export function useAssetsQuery(
...
+    setQueryOptions(({ fetchKey }) => ({
+      fetchKey: fetchKey + 1,
+      fetchPolicy: 'store-only',
+    }))
+  }, [data.nftAssets?.edges?.length, environment, vars])
+  // NB: This will poll every POLLING_INTERVAL, *not* every POLLING_INTERVAL from the last successful poll.
+  // TODO(WEB-2004): Update useInterval to wait for the fn to complete before rescheduling.
+  useInterval(refresh, POLLING_INTERVAL)
+
+  // It is especially important for this to be memoized to avoid re-rendering from polling if data is unchanged.
```
cbachmeier, line 159: great catch

src/graphql/data/RelayEnvironment.ts
zzmp: The effective time was previously `5m`, from the redundant response cache that was used in `fetchGraphQL`.
This just moves that expiration to the store, instead of maintaining two separate caches (with the old response cache being impossible to override).

src/graphql/data/nft/Asset.ts
zzmp: AFAICT they'll only refresh if they are re-rendered, because that will trigger a re-fetch. They do not refresh when the time expires, nor do they do that right now.

JFrankfurt (approved):
great implementation! I agree that 1s seems much too fast a polling interval.

cbachmeier:
For context on polling times, one of the biggest value props for this launch is having the most accurate listings. Many of our competitors already implement 1s polling for their listings so if we can't match, then our listings will be out of date comparatively. This is an important value add from product.

cbachmeier (changes requested):

cbachmeier:
Data team is currently running some tests on 1s polling. Everything looks good and pending their testing will approve.

PR #5084, opened 2022-11-04
fix: move /profile to /nfts/profile

vm:
In the future if we decide to use /profile we can add a redirect.


src/nft/components/bag/Bag.tsx
```diff
@@ -76,9 +76,9 @@ const Bag = () => {
   const setTotalUsdPrice = useBag((s) => s.setTotalUsdPrice)
 
   const { pathname } = useLocation()
-  const isProfilePage = pathname.startsWith('/profile')
+  const isProfilePage = pathname.startsWith('/nfts/profile')
   const isNFTPage = pathname.startsWith('/nfts')
-  const shouldShowBag = isNFTPage || isProfilePage
+  const shouldShowBag = isNFTPage
```
aballerr, line 81: this seems a bit redundant now we are only doing one boolean check.  Probably should replace all ```shouldShowBag``` with ```isNFTPage```

vm: i can do that

aballerr (approved):

PR #5086, opened 2022-11-04
fix: Remove rank on trending collection table (mobile)

gbugyis:
Removes rank on trending collection table (mobile-only) and unused NFT Explore styles

### DIFF
--- a/src/nft/components/explore/Table.tsx
+++ b/src/nft/components/explore/Table.tsx
@@ -1,6 +1,7 @@
 import clsx from 'clsx'
 import { LoadingBubble } from 'components/Tokens/loading'
 import { useWindowSize } from 'hooks/useWindowSize'
+import { useIsMobile } from 'nft/hooks'
 import { useEffect } from 'react'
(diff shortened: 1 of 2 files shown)


src/nft/components/explore/Table.tsx
```diff
@@ -186,9 +191,11 @@ export function Table<D extends Record<string, unknown>>({
                   <td className={clsx(styles.td, classNames?.td)} {...cell.getCellProps()} key={cellIndex}>
                     {cellIndex === 0 ? (
                       <RankCellContainer>
-                        <ThemedText.BodySecondary fontSize="14px" lineHeight="20px">
-                          {i + 1}
-                        </ThemedText.BodySecondary>
+                        {!isMobile && (
```
vm, line 194: nit: I think it'd be easier in the future to handle this within css. Display none if it's below a breakpoint. That way we can add further conditional breakpoints without needing to add more hooks to the component, and it separates the concerns a bit more. Not blocking though.

PR #5087, opened 2022-11-04
fix: add back floor to search results

cbachmeier:
When restructuring GenieCollection type, floor price stopped showing in universal search results. This fixes that bug.


JackShort (approved):
should we just make the slice amounts consts?

src/nft/queries/genie/SearchCollectionsFetcher.ts
```diff
@@ -39,10 +39,21 @@ export const fetchSearchCollections = async (addressOrName: string, recursive =
...
+          return {
+            ...collection,
+            stats: {
+              ...collection.stats,
+              floor_price: collection.floorPrice,
+            },
+          }
+        })
+      : []
+    return formattedData.slice(0, 6)
```
JackShort, line 54: also I thought we wanted 5 and not 6?

PR #5089, opened 2022-11-04
fix: replaceState instead of pushState for collection filters

vm:
I have a separate PR WIP to move this all to react-router, but for now replacing should solve the issue.


JackShort (approved):
lgtm

PR #5091, opened 2022-11-04
fix: marketplace icons on details page

cbachmeier:
Marketplaces returned from gql are in all caps, this has them match the format of the public icons.


aballerr:
thank you for addressing this

PR #5092, opened 2022-11-04
fix: wrong types for set min

JackShort:
* numbers were being passed to set min which were expected to be strings


vm (approved):
i don't think this is ideal fwiw, we should just be converting to string where the functions need it

khaled200286 (commented):
New

PR #5094, opened 2022-11-04
fix: price range and details screen sizing

aballerr:
1. Temporary fix for price range while we port over to graphql
2. Updated padding on details to be correct as screen shrinks


src/nft/queries/genie/CollectionStatsFetcher.ts
```diff
@@ -57,6 +57,9 @@ export const CollectionStatsFetcher = async (addressOrName: string, recursive =
 
   const data = await r.json()
   const collections = data?.data.map((collection: Record<string, unknown>) => {
+    // @ts-ignore
+    collection.stats.floor_price = collection.floorPrice
```
aballerr, line 61: I don't want to spend a ton of time trying to fix these types since we are moving away from this pretty soon

src/nft/pages/asset/Asset.tsx
```diff
@@ -16,7 +16,17 @@ const AssetContainer = styled.div`
   width: 100%;
   justify-content: center;
   gap: 60px;
-  padding: 48px 40px 0 40px;
+  padding: 48px 48px 0 48px;
+
+  @media (max-width: 960px) {
```
lynnshaoyu, line 21: can you use the predefined BREAKPOINT constants here pls for consistency: code/interface/src/components/Tokens/constants.ts 

although we should probably also move these constants to a different shared location, not just under Tokens🤔

aballerr: I checked just now and the exact sizings are different, not sure why the design doesn't match our usual padding

lynnshaoyu (approved):

PR #5096, opened 2022-11-04
fix: defer useInterval until callback resolves

zzmp:
Updates `useInterval` to defer the next interval until the callback has completed.

For synchronous callbacks, this has no effect. For callbacks returning a `Promise`, this prevents stacking invocations, should the invocation take longer than the interval.

### DIFF
--- a/src/hooks/useMachineTime.ts
+++ b/src/hooks/useMachineTime.ts
@@ -1,12 +1,15 @@
 import useInterval from 'lib/hooks/useInterval'
-import { useState } from 'react'
+import { useCallback, useState } from 'react'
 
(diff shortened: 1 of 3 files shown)


cbachmeier (commented):
I'm experiencing a bug when this and the polling branch are merged together. The paginated query continues to poll after leaving the page. Test by going to a collection, scrolling, and then going to a different page. In the network tab you should see continued queries for the old collection. This didn't happen on the polling branch without this fix merged in.

zzmp:
I think it's probably from using a ref for the timeoutId. I can update it.

cbachmeier (approved):
LGTM

PR #5097, opened 2022-11-04
chore: update design-system grays

vm:
I am unclear if this is enough to merge incrementally, or if we need further component level changes first.


callil (approved):
lgtm - double checked against the design file

PR #5110, opened 2022-11-07
fix: only poll every 12s

zzmp:
- Updates network providers to only poll for a new block every 12s, to match mainnet's block frequency.
- Exports this polling interval, and reuses it for routing staleness.

### DIFF
--- a/src/constants/providers.ts
+++ b/src/constants/providers.ts
@@ -3,9 +3,12 @@ import { deepCopy } from '@ethersproject/properties'
 // eslint-disable-next-line @typescript-eslint/no-restricted-imports
 import { StaticJsonRpcProvider } from '@ethersproject/providers'
 import { isPlain } from '@reduxjs/toolkit'
+import ms from 'ms.macro'
 
-import { SupportedChainId } from './chains'
+import { CHAIN_IDS_TO_NAMES, SupportedChainId } from './chains'
 import { RPC_URLS } from './networks'
+
+export const POLLING_INTERVAL = ms`12s`
 
 class AppJsonRpcProvider extends StaticJsonRpcProvider {
   private _blockCache = new Map<string, Promise<any>>()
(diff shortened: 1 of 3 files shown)


vm (approved):

vm:
should we be using 12 for all chains?

zzmp:
> should we be using 12 for all chains?

Considering that we were only considering routes stale after 15s, I think this is fine.
To take it further, I've updated route staleness to also use 12s.

src/constants/providers.ts
```diff
@@ -3,10 +3,13 @@ import { deepCopy } from '@ethersproject/properties'
 // eslint-disable-next-line @typescript-eslint/no-restricted-imports
 import { StaticJsonRpcProvider } from '@ethersproject/providers'
 import { isPlain } from '@reduxjs/toolkit'
+import ms from 'ms.macro'
 
-import { SupportedChainId } from './chains'
+import { CHAIN_IDS_TO_NAMES, SupportedChainId } from './chains'
 import { RPC_URLS } from './networks'
 
+export const POLLING_INTERVAL = ms`12s`
```
vm: add comment for why 12s

src/constants/providers.ts
```diff
@@ -18,8 +21,9 @@ class AppJsonRpcProvider extends StaticJsonRpcProvider {
     return this._blockCache
   }
 
-  constructor(urls: string[]) {
-    super(urls[0])
+  constructor(chainId: SupportedChainId) {
+    super(RPC_URLS[chainId][0], { chainId, name: CHAIN_IDS_TO_NAMES[chainId] })
```
vm, line 28: why did we add the new options object?

src/constants/providers.ts
```diff
@@ -3,10 +3,16 @@ import { deepCopy } from '@ethersproject/properties'
...
+import ms from 'ms.macro'
 
-import { SupportedChainId } from './chains'
+import { CHAIN_IDS_TO_NAMES, SupportedChainId } from './chains'
 import { RPC_URLS } from './networks'
 
+// NB: Third-party providers (eg MetaMask) will have their own polling intervals,
+// which should be left as-is to allow operations (eg transaction confirmation) to resolve faster.
+// Network providers (eg AppJsonRpcProvider) need to update less frequently to be considered responsive.
+export const POLLING_INTERVAL = ms`12s` // mainnet block frequency
```
vm, line 14: explain why this is also used for other chains

src/constants/providers.ts
zzmp: adding an options object for a StaticJsonRpcProvider allows ethers to skip the initial detectNetwork call to determine _which_ static network the provider fetches for. Considering that this runs during startup, it eases congestion for initial outgoing network requests.

vm (approved):

PR #5114, opened 2022-11-08
refactor: remove token promo banner

yyip-dev:
Remove token promo banner since 1) it's no longer needed and 2) we do not want it to conflict with banners of future product launches.

Verified with eng and product that currently the vast majority of users are accessing tokens page via the top tab directly, instead of via the promo banner, so no expected user/product impact here.


vm:
we may want to rename the component to "PromoBanner" since we'll end up re-using it for NFT stuff soon? rather than delete completely

yyip-dev:
@vm Good question, it looks like the component itself mostly consists of custom styles so my thinking was that their version will end up fairly different anyway. They can also refer to this previous PR if needed.

Separately, we'll likely want an actual "banner" component that is shareable/reusable as part of a component library in the future.

vm (approved):

PR #5121, opened 2022-11-08
style: Font weight on Trending Collections table

gbugyis:
Fix font weight to match Figma (Collection Names)

### DIFF
--- a/src/nft/components/explore/Cells/Cells.tsx
+++ b/src/nft/components/explore/Cells/Cells.tsx
@@ -27,10 +27,11 @@ const CollectionName = styled.div`
   text-overflow: ellipsis;
 `
 
-const TruncatedSubHeader = styled(ThemedText.SubHeader)`
+const TruncatedSubHeader = styled(ThemedText.BodyPrimary)`
   overflow: hidden;
   white-space: nowrap;
   text-overflow: ellipsis;
+  font-weight: 400;
 `
 
 const RoundedImage = styled.div<{ src?: string }>`
--- a/src/nft/components/explore/Cells/Cells.tsx
+++ b/src/nft/components/explore/Cells/Cells.tsx
@@ -12,30 +12,32 @@ import { Box } from '../../Box'
 import { Column, Row } from '../../Flex'
 import * as styles from './Cells.css'
 
-const truncatedText = `
+const TruncatedText = styled.div`
   overflow: hidden;
   white-space: nowrap;
   text-overflow: ellipsis;
 `
 
-const CollectionNameContainer = styled.div`
+const CollectionNameContainer = styled(TruncatedText)`
   display: flex;
   padding: 14px 0px 14px 8px;
   align-items: center;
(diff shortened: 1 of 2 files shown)


src/nft/components/explore/Cells/Cells.tsx
```diff
@@ -27,10 +27,11 @@ const CollectionName = styled.div`
   text-overflow: ellipsis;
 `
 
-const TruncatedSubHeader = styled(ThemedText.SubHeader)`
```
cbachmeier, line 30: Sorry it's not clear on the ticket but this is supposed to only be smaller on mobile. Also the font color should still be textPrimary

cbachmeier (changes requested):

src/nft/components/explore/Cells/Cells.tsx
gbugyis: thats my bad, it was in the mobile explore epic so I should've noticed. It should be fixed now. I also put some repeated styles in a variable. I didn't see an existing pattern for something like that (besides sprinkles) - maybe there is a better way?

gbugyis: thats my bad, it was in the mobile explore epic so I should've noticed. It should be fixed now. I also put some repeated styles in a variable. I didn't see an existing pattern for something like that (besides sprinkles) - maybe there is a better way?

gbugyis:
Sneaking in one more ticket for maxWidth on the Collection Name

src/nft/components/explore/Cells/Cells.tsx
cbachmeier: This usage is fine, you can also replaced styled.div with the common style similar to here: https://github.com/Uniswap/interface/blob/main/src/components/Card/index.tsx#L12

src/nft/components/explore/Table.tsx
```diff
@@ -188,7 +183,12 @@ export function Table<D extends Record<string, unknown>>({
             >
               {row.cells.map((cell, cellIndex) => {
                 return (
-                  <td className={clsx(styles.td, classNames?.td)} {...cell.getCellProps()} key={cellIndex}>
+                  <td
+                    className={styles.td}
+                    {...cell.getCellProps()}
+                    key={cellIndex}
+                    style={{ maxWidth: cellIndex === 0 ? (isMobile ? '300px' : '360px') : '160px' }}
```
cbachmeier, line 190: non-blocking nit. It would be nice to have constant names for these px amounts, such as MOBILE_ROW_WIDTH DESKTOP_ROW_WIDTH, etc

cbachmeier, line 190: maxWidth seems large on mobile, DMed screenshots.

cbachmeier (changes requested):
Only thing blocking is the mobile width seems large

cbachmeier (approved):
LGTM

PR #5131, opened 2022-11-08
fix: show transparent box on empty banner

vm:
Nouns for example have no banner image, so if that's the case we should just make the bg empty.

### DIFF
--- a/src/nft/pages/collection/index.tsx
+++ b/src/nft/pages/collection/index.tsx
@@ -104,7 +104,7 @@ const Collection = () => {
                     <CollectionBannerLoading />
                   ) : (
                     <Box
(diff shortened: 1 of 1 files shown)


src/nft/pages/collection/index.tsx
```diff
@@ -104,10 +104,14 @@ const Collection = () => {
                     <CollectionBannerLoading />
                   ) : (
                     <Box
-                      as="img"
+                      as={collectionStats?.bannerImageUrl ? 'img' : undefined}
```
cbachmeier, line 107: show `'div'` in the else condition

cbachmeier (commented):
looks good, minor comment and then I'm good to approve

cbachmeier (approved):
LGTM

PR #5141, opened 2022-11-09
feat: Mobile status bar

aballerr:
https://uniswaplabs.atlassian.net/browse/WEB-1974

See task for screenshots

### DIFF
--- a/src/components/Web3Status/index.tsx
+++ b/src/components/Web3Status/index.tsx
@@ -109,7 +109,7 @@ const Web3StatusConnected = styled(Web3StatusGeneric)<{ pending?: boolean; isNft
     }
   }
 
-  @media only screen and (max-width: ${({ theme }) => `${theme.breakpoint.md}px`}) {
+  @media only screen and (max-width: ${({ theme }) => `${theme.breakpoint.lg}px`}) {
     width: ${({ isNftActive, pending }) => isNftActive && !pending && '44px'};
     height: ${({ isNftActive, pending }) => isNftActive && !pending && '44px'};
     div {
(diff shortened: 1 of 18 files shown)


src/components/Web3Status/index.tsx
```diff
@@ -107,6 +108,22 @@ const Web3StatusConnected = styled(Web3StatusGeneric)<{ pending?: boolean }>`
           pending ? darken(0.1, theme.deprecated_primary1) : darken(0.1, theme.deprecated_bg2)};
     }
   }
+
+  @media only screen and (max-width: ${({ theme }) => `${theme.breakpoint.lg}px`}) {
```
vm, line 112: what should the width be if its pending? not sure why that's related

src/components/Web3Status/index.tsx
```diff
@@ -107,6 +108,22 @@ const Web3StatusConnected = styled(Web3StatusGeneric)<{ pending?: boolean }>`
...
+    width: ${({ isNftActive, pending }) => isNftActive && !pending && '44px'};
+    height: ${({ isNftActive, pending }) => isNftActive && !pending && '44px'};
+    div {
```
vm, line 115: don't love the inner div selector here, can you just make the other thing a styled component?

src/components/Web3Status/index.tsx
aballerr: pending should act as the status bar normally does which is showing the pending transactions and the loading circle.  Otherwise it should just be a circle

vm (approved):

PR #5143, opened 2022-11-09
refactor: replace sweep query with gql call

cbachmeier:
Replace sweep queries with graphql. This is the last query that'll need to be migrated.

Out of Scope:
- Apollo migration
- Adding larvalabs and sudo to Sweep (routing code currently doesn't support)

### DIFF
--- a/src/graphql/data/nft/Asset.ts
+++ b/src/graphql/data/nft/Asset.ts
@@ -2,12 +2,19 @@ import graphql from 'babel-plugin-relay/macro'
 import { parseEther } from 'ethers/lib/utils'
 import useInterval from 'lib/hooks/useInterval'
 import ms from 'ms.macro'
-import { GenieAsset } from 'nft/types'
+import { GenieAsset, Trait } from 'nft/types'
 import { useCallback, useMemo, useState } from 'react'
 import { fetchQuery, useLazyLoadQuery, usePaginationFragment, useRelayEnvironment } from 'react-relay'
 
 import { AssetPaginationQuery } from './__generated__/AssetPaginationQuery.graphql'
-import { AssetQuery, NftAssetsFilterInput, NftAssetSortableField } from './__generated__/AssetQuery.graphql'
+import {
+  AssetQuery,
+  AssetQuery$variables,
+  NftAssetsFilterInput,
+  NftAssetSortableField,
+  NftAssetTraitInput,
+  NftMarketplace,
+} from './__generated__/AssetQuery.graphql'
 import { AssetQuery_nftAssets$data } from './__generated__/AssetQuery_nftAssets.graphql'
 
 const assetPaginationQuery = graphql`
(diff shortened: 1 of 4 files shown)


src/graphql/data/nft/Asset.ts
```diff
@@ -116,6 +123,61 @@ type NftAssetsQueryAsset = NonNullable<
...
+    owner: asset.ownerAddress,
+    creator: {
+      profile_img_url: asset.collection?.creator?.profileImage?.url,
+      address: asset.collection?.creator?.address,
+    },
+    metadataUrl: asset.metadataUrl,
+  }
+}
+
 export function useAssetsQuery(
```
vm, line 181: did this change other than moving it? hard to tell

src/nft/components/collection/Sweep.tsx
```diff
@@ -267,8 +225,11 @@ export const Sweep = ({ contractAddress, collectionStats, minPrice, maxPrice, sh
     jointCollections.forEach((asset) => {
       if (!asset.susFlag) {
         const isNFTX = asset.marketplace === Markets.NFTX
-        asset.priceInfo.ETHPrice = calcPoolPrice(asset, isNFTX ? counterNFTX : counterNFT20)
-        BigNumber.from(asset.priceInfo.ETHPrice).gte(0) && (isNFTX ? counterNFTX++ : counterNFT20++)
+        const poolPrice = calcPoolPrice(asset, isNFTX ? counterNFTX : counterNFT20)
```
vm, line 228: is the gte -> gt intentional?

src/nft/utils/calcPoolPrice.ts
```diff
@@ -21,7 +21,7 @@ export const calcPoolPrice = (asset: GenieAsset, position = 0) => {
       : (asset.sellorders[0] as Deprecated_SellOrder)
 
   const decimals = BigNumber.from(1).mul(10).pow(18)
-  const ammFee = nft.ammFeePercent ? (100 + (nft.ammFeePercent as number)) * 100 : 110 * 100
+  const ammFee = nft?.ammFeePercent ? (100 + (nft.ammFeePercent as number)) * 100 : 110 * 100
```
vm, line 24: how did this previously pass typescript lint if it can be undefined?

vm: and is the cast to number necessary?

src/nft/components/collection/Sweep.tsx
```diff
@@ -211,52 +209,12 @@ export const Sweep = ({ contractAddress, collectionStats, minPrice, maxPrice, sh
...
-
-    {
-      refetchOnWindowFocus: false,
-      refetchOnMount: false,
-      refetchOnReconnect: false,
-    }
-  )
-
-  const allAssetsFetched = isCollectionAssetsFetched && isNftxCollectionAssetsFetched && isNft20CollectionAssetsFetched
+  const collectionAssets = useSweepAssetsQuery(getSweepFetcherParams('others'))
```
vm, line 212: can you name this otherCollectionAssets?. also i think Others should be in the Markets enum

src/nft/components/collection/Sweep.tsx
cbachmeier: Yes, 0 is returned for non-pooled asset so we just use the base price in this case

src/nft/utils/calcPoolPrice.ts
cbachmeier: Previously we only called calcPoolPrice on pooled assets or assets from the v3 sweep endpoint that had a placeholder ammFeePercent field. The change to gql no longer provided that field.

src/nft/components/collection/Sweep.tsx
cbachmeier: There's scoped work to include new marketplaces from gql in sweep which at that point I think it would be worth looking into a smarter handling of if assets are pooled or not. I intentionally didn't want to alter that behaviour in this PR to avoid scope creep.

src/graphql/data/nft/Asset.ts
cbachmeier: it just moved the formatting logic to its own helper function

vm (approved):

PR #5146, opened 2022-11-09
chore: add accentAction to sprinkles

vm:
It should be blue400 in light and dark mode.

### DIFF
--- a/src/nft/css/sprinkles.css.ts
+++ b/src/nft/css/sprinkles.css.ts
@@ -7,6 +7,7 @@ const themeContractValues = {
     accentFailureSoft: '',
     accentAction: '',
     accentActionSoft: '',
+    accentActive: '',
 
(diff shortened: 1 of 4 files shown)


src/nft/css/sprinkles.css.ts
```diff
@@ -7,6 +7,7 @@ const themeContractValues = {
     accentFailureSoft: '',
     accentAction: '',
     accentActionSoft: '',
+    accentActive: '',
```
cbachmeier, line 10: if it's the same in dark and light mode you can just add it to vars.color in sprinkles instead of to the themes

JFrankfurt:
defer to @cbachmeier on this one as I'm not super familiar w/ sprinkles stuff

cbachmeier (approved):
LGTM

PR #5147, opened 2022-11-09
fix: accentAction for FilterButton

vm:
.

### DIFF
--- a/src/nft/components/collection/FilterButton.css.ts
+++ b/src/nft/components/collection/FilterButton.css.ts
@@ -4,7 +4,7 @@ import { sprinkles, themeVars, vars } from 'nft/css/sprinkles.css'
 
 export const filterButton = sprinkles({
   backgroundColor: 'accentAction',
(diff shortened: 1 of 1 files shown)


src/nft/components/collection/FilterButton.css.ts
```diff
@@ -3,7 +3,7 @@ import { loadingAsset } from 'nft/css/loading.css'
 import { sprinkles, themeVars, vars } from 'nft/css/sprinkles.css'
 
 export const filterButton = sprinkles({
-  backgroundColor: 'blue400',
+  backgroundColor: 'accentAction',
   color: 'explicitWhite',
```
cbachmeier, line 7: can you also make color `textPrimary`

vm: sure

cbachmeier (approved):
LGTM. filterbutton icon when open is matching the text color

PR #5151, opened 2022-11-09
refactor: remove graphql flag and default to gql endpoints

cbachmeier:
- Remove graphql flag and related conditionals
- Defaults calls to graphql
- Remove deprecated sell order type and required logic
- Remove no longer used queries
- Remove so no longer needed todos


src/nft/utils/calcPoolPrice.ts
```diff
@@ -43,29 +31,23 @@ export const calcPoolPrice = (asset: GenieAsset, position = 0) => {
...
-        >
-      )?.poolMetadata?.tokenReserves
-    )?.toLocaleString('fullwide', { useGrouping: false }) ?? 1
+      nft as Record<
+        string,
+        {
+          tokenReserves: number
+        }
+      >
+    )?.poolMetadata?.tokenReserves?.toLocaleString('fullwide', { useGrouping: false }) ?? 1
```
vm, line 50: don't fully understanding this change

vm (approved):

src/nft/utils/calcPoolPrice.ts
cbachmeier: The sell order data returned from the v3 endpoint was structured differently than from gql. Previously this logic handled both but not only has to handle the one case. We still need to have this typecasting as the protocolParameters are passed back as JSON which we parse on the FE. Talked to BE about having it be more structured but that's in the backlog

PR #5152, opened 2022-11-09
chore: remove genie endpoint

vm:


cbachmeier:
BE still needs to add some gating logic to the temp url for nfts

cbachmeier:
I also have a branch for this change https://github.com/Uniswap/interface/tree/cab/v3_to_tempEndpoint

cbachmeier:
Was unrelated to production event

PR #5154, opened 2022-11-10
fix: don't fetch collections without flag

vm:
This will avoid failing requests in productions, since we can't avoid the rule of hooks (useQuery can't be conditionally called).


cbachmeier (approved):
Tested locally. LGTM.

PR #5155, opened 2022-11-10
fix: don't fetch tokens on empty search

vm:
When the search query is empty we don't need to make a request. This will fail from the backend.

### DIFF
--- a/src/components/NavBar/SearchBar.tsx
+++ b/src/components/NavBar/SearchBar.tsx
@@ -58,6 +58,7 @@ export const SearchBar = () => {
       refetchOnWindowFocus: false,
       refetchOnMount: false,
       refetchOnReconnect: false,
+      enabled: !!searchValue,
     }
   )
 


src/components/NavBar/SearchBar.tsx
```diff
@@ -58,6 +58,7 @@ export const SearchBar = () => {
       refetchOnWindowFocus: false,
       refetchOnMount: false,
       refetchOnReconnect: false,
+      enabled: !!searchValue,
```
cbachmeier, line 61: `debouncedSearchValue`?

src/components/NavBar/SearchBar.tsx
```diff
@@ -58,6 +58,7 @@ export const SearchBar = () => {
       refetchOnWindowFocus: false,
       refetchOnMount: false,
       refetchOnReconnect: false,
+      enabled: !!debouncedSearchValue,
```
cbachmeier, line 61: you might need to condition based off length, I'm still able to fire off a blank search. 
Steps:
- enter in a character(including space)
- delete character
- In network tab a blank search fires off

cbachmeier (approved):
LGTM

PR #5156, opened 2022-11-10
fix: resolves infinite load for WalletConnect

vm:
This was added as an updated peerDep in web3-react: https://github.com/Uniswap/web3-react/pull/702

See here for more context: https://github.com/WalletConnect/walletconnect-monorepo/pull/977

I haven't yet figured out how to deploy web3-react, feel free to do that and update this PR if you want also.

Other thoughts:
- should we be lazy loading wallet connect
- should we upgrade to v2

### DIFF
--- a/package.json
+++ b/package.json
@@ -158,7 +158,7 @@
     "@visx/react-spring": "^2.12.2",
     "@visx/responsive": "^2.10.0",
     "@visx/shape": "^2.11.1",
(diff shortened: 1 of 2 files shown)


zzmp (changes requested):
This can be further de-duplicated.
Why not use 1.8.0? It should be backwards-compatible.

zzmp (approved):

PR #5158, opened 2022-11-10
fix: give a bunch of list renders keys

JFrankfurt:


src/nft/components/explore/CarouselCard.tsx
```diff
@@ -223,13 +223,13 @@ export const CarouselCard = ({ collection, onClick }: CarouselCardProps) => {
...
             <HeaderRow>{gqlCollection.marketplaceCount?.reduce((acc, cur) => acc + cur.count, 0)} Listings</HeaderRow>
-            {MARKETS_TO_CHECK.map((market) => {
+            {MARKETS_TO_CHECK.map((market, index) => {
               const marketplace = gqlCollection.marketplaceCount?.find(
                 (marketplace) => marketplace.marketplace === market
               )
               return (
                 <MarketplaceRow
-                  key={'trendingCollection' + collection.address}
+                  key={`CarouselCard-key-${collection.address}-${index}`}
```
vm, line 232: this should be keyed on market not index? how can there be duplicates

JFrankfurt: done

vm (approved):
i do think we should filter out any undefined market places instead of needing the conditionals

JFrankfurt:
@vm yeah that makes sense. I'll do that before I merge

lynnshaoyu (approved):

PR #5161, opened 2022-11-10
fix: clear or retain amounts on token select

zzmp:
Clears the amount when the token with an amount is switched.
Swaps the active token if the token with an amount is moved to the other spot.

### DIFF
--- a/src/components/Widget/inputs.tsx
+++ b/src/components/Widget/inputs.tsx
@@ -84,11 +84,15 @@ export function useSyncWidgetInputs({
 
       setType((type) => {
         // If flipping the tokens, also flip the type/amount.
-        if (isFlip) return invertTradeType(type)
+        if (isFlip) {
+          return invertTradeType(type)
+        }
 
         // Setting a new token should clear its amount, if it is set.
         const activeField = type === TradeType.EXACT_INPUT ? Field.INPUT : Field.OUTPUT
-        if (selectingField === activeField) setAmount(() => EMPTY_AMOUNT)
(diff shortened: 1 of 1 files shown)


zzmp:
Pending input from design. Once design has cleared this, I'll mark it ready for review.
It's also pretty light, so feel free to review ahead of that.

src/components/Widget/inputs.tsx
```diff
@@ -71,20 +71,30 @@ export function useSyncWidgetInputs({
...
+
+      setType((type) => {
+        // If flipping the tokens, also flip the type/amount.
+        if (isFlip) return invertTradeType(type)
+
+        // Setting a new token should clear its amount, if it is set.
+        const activeField = type === TradeType.EXACT_INPUT ? Field.INPUT : Field.OUTPUT
+        if (selectingField === activeField) setAmount(() => EMPTY_AMOUNT)
+
+        return type
```
JFrankfurt, line 93: nit: add curlies on the return statement--took me way too long to notice the early return on inversion and thought there was a bug here

JFrankfurt (approved):
lgtm

src/components/Widget/inputs.tsx
zzmp: I think we just need to add this as a lint rule to prevent people (aka _me, the biggest offender_) from doing it.

zzmp: That'd also let us fix it programmatically, instead of wasting code review effort on it.

PR #5164, opened 2022-11-10
fix: add back the border-radius for swap tx conf

JFrankfurt:


zzmp (approved):
lgtm - but please include either a screenshot or a link to repro

JFrankfurt:
@zzmp sry, internet so bad that doing that was kind of a pain 🙈 

### before
image

### after
image

PR #5167, opened 2022-11-10
fix: analytics contention

zzmp:
- Batches analytics to avoid network contention from many requests on startup (which could potentially delay startup)
- Initializes analytics outside of React, to guarantee that it is only initialized once

### DIFF
--- a/src/analytics/index.ts
+++ b/src/analytics/index.ts
@@ -29,6 +29,10 @@ export function initializeAnalytics() {
         region: false,
         dma: false, // designated market area
       },
+      // Configure the SDK to use batched events (to reduce network contention)
+      // by grouping events from the same render cycle
+      flushIntervalMillis: 50, // assuming 20fps renders
+      useBatch: true,
     }
   )
 }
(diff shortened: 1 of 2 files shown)


src/analytics/index.ts
```diff
@@ -29,6 +29,10 @@ export function initializeAnalytics() {
         region: false,
         dma: false, // designated market area
       },
+      // Configure the SDK to use batched events (to reduce network contention)
+      // by grouping events from the same render cycle
+      flushIntervalMillis: 50, // assuming 20fps renders
+      useBatch: true,
```
lynnshaoyu, line 35: useBatch affects the attribution tracking (can disable default behavior) https://www.docs.developers.amplitude.com/data/sdks/typescript-browser/#configuration

I'm not sure this is what we want?

lynnshaoyu (commented):
.

zzmp:
This has been resolved upstream - useBatch was not enabled, but other changes (eg moving initialization outside of react) were made. See Uniswap/analytics/pull/8 and Uniswap/analytics/pull/9

PR #5173, opened 2022-11-11
fix: initialize analytics outside of react lifecycle

zzmp:
Initializes analytics outside of React. Initializing inside of React often leads to multiple calls, and initializeAnalytics is not set up to be idempotent.


dannydaniil (approved):
lgtm

PR #5178, opened 2022-11-11
fix: fixing header on light mode

aballerr:
https://uniswaplabs.atlassian.net/browse/WEB-2064

1. The box shadow from the nfts shows the edges of the navbar
2. Moving everything 14px in (left and right) moves the edges of the navbar away from the box shadow so it doesn't show the edges
3. 14px was the smallest width I could find where this works
4. See Jira for before and after


cbachmeier (approved):
LGTM

PR #5179, opened 2022-11-11
fix: fixing color of buttons

aballerr:


src/nft/components/bag/BagHeader.tsx
```diff
@@ -13,10 +13,6 @@ const ClearButton = styled(ButtonText)`
   font-size: 14px;
   line-height: 16px;
   transition: 150ms ease color;
-
-  :hover {
-    color: ${({ theme }) => theme.accentActive};
-  }
```
aballerr, line 19: we had a random blue hoverstate

src/nft/components/collection/Filters.css.ts
```diff
@@ -8,7 +8,8 @@ export const container = style([
     paddingTop: '4',
   }),
   {
-    width: '300px',
+    width: '308px',
+    paddingRight: '8px',
```
aballerr, line 12: the shadow on the right slider overlaps the adjacent div, which creates a line to the right that looks strange(light mode only).  This maintains the original dimensions while solving that issue

src/nft/css/common.css.ts
```diff
@@ -41,7 +41,6 @@ export const commonButtonStyles = style([
   sprinkles({
     borderRadius: '12',
     transition: '250',
-    boxShadow: { hover: 'elevation' },
```
aballerr, line 44: common button styles has a boxShadow.  Not a heavily used style, but we removed boxShadows from buttons

src/nft/components/collection/FilterButton.tsx
```diff
@@ -37,7 +37,7 @@ export const FilterButton = ({
       width={isMobile ? '44' : 'auto'}
       height="44"
       whiteSpace="nowrap"
-      color="textPrimary"
+      color="white"
```
unipadmini, line 40: are we sure we want to hardcode this tow hite?

unipadmini (commented):
Do we want to hardcode white vs use design system colors?

src/nft/components/collection/PriceRange.tsx
```diff
@@ -186,10 +187,11 @@ export const PriceRange = () => {
             top: '3px',
             width: '12px',
             height: '20px',
-            backgroundColor: `${theme.textPrimary}`,
+            opacity: '1',
+            backgroundColor: `white`,
             borderRadius: '4px',
             border: 'none',
-            boxShadow: `${theme.shallowShadow.slice(0, -1)}`,
+            boxShadow: colorsDark.shallowShadow.slice(0, -1),
```
unipadmini, line 194: Why aren't we pulling this from theme anymore?

unipadmini (changes requested):
A few comments. (I think it might also be easy to split out the number formatting change in a different PR since I think the colors change is a bit tricker)

src/nft/components/collection/FilterButton.tsx
vm: this is a merge conflict, i changed this upstream

src/nft/components/details/AssetPriceDetails.tsx
```diff
@@ -372,7 +372,7 @@ export const AssetPriceDetails = ({ asset, collection }: AssetPriceDetailsProps)
                     }
                   }}
                 >
-                  <SubHeader lineHeight={'20px'}>
+                  <SubHeader color="white" lineHeight={'20px'}>
```
vm, line 375: should this be primaryColor?

aballerr: primary isn't white in light theme

src/nft/components/collection/PriceRange.tsx
aballerr: theme has a different shadow color.  I couldn't find this exact shadow defined anywhere but in darktheme.

src/nft/components/collection/FilterButton.tsx
aballerr: Screen Shot 2022-11-14 at 9 40 02 AM

This is what it looks like with textPrimary

src/nft/components/details/AssetPriceDetails.tsx
unipadmini: @aballerr could you sync with yushin on this? Apparently they defined a `button-label` in the design system that reflects what the text colors should be for light and dark.

src/nft/pages/collection/index.tsx
```diff
@@ -34,8 +39,27 @@ const MobileFilterHeader = styled(Row)`
...
+  }
+
+  @media screen and (max-width: ${SMALL_MEDIA_BREAKPOINT}) {
+    padding-left: 4px;
+    padding-right: 4px;
+  }
+
+  @media screen and (max-width: ${MOBILE_MEDIA_BREAKPOINT}) {
+    padding-left: 16px;
+  }
```
aballerr, line 62: https://uniswapteam.slack.com/archives/C04822DFATE/p1668386627323939

src/nft/components/details/AssetPriceDetails.tsx
aballerr: Screen Shot 2022-11-14 at 11 03 31 AM
Screen Shot 2022-11-14 at 11 03 38 AM


This is what buy now currently looks like and this is what buttons on our site look like.  I'm not really sure what I need to sync on to be honest

vm (approved):
approving, but i feel like we should be avoiding using specific colors vs theme variables

unipadmini (approved):

PR #5181, opened 2022-11-11
fix: enforce uniform asset height

zzmp:
Sets the loading assets to the same height as loaded assets.
This also fixes an inconsistent bug where the uniformHeight would be reset due to an inconsistent aspect ratio.

### DIFF
--- a/src/nft/components/collection/Card.tsx
+++ b/src/nft/components/collection/Card.tsx
@@ -256,6 +256,7 @@ const Image = ({ uniformHeight, setUniformHeight }: ImageProps) => {
         width="full"
         style={{
           aspectRatio: uniformHeight === UniformHeights.notUniform ? '1' : 'auto',
+          minHeight: uniformHeight,
           transition: 'transform 0.4s ease 0s',
         }}
         src={asset.imageUrl || asset.smallImageUrl}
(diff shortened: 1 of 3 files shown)


src/nft/components/collection/Card.tsx
```diff
@@ -256,6 +256,7 @@ const Image = ({ uniformHeight, setUniformHeight }: ImageProps) => {
         width="full"
         style={{
           aspectRatio: uniformHeight === UniformHeights.notUniform ? '1' : 'auto',
+          minHeight: uniformHeight,
```
cbachmeier, line 259: I think this change had an unintended effect as the aspect ratio shift seems to happen more frequently. More details sent in DM

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -322,9 +322,9 @@ export const CollectionNfts = ({ contractAddress, collectionStats, rarityVerifie
     }
   }, [contractAddress])
 
-  const Nfts =
-    collectionNfts &&
-    collectionNfts.map((asset) =>
+  const assets = useMemo(() => {
+    if (!collectionNfts) return null
+    return collectionNfts.map((asset) =>
       asset ? (
```
cbachmeier, line 328: this check for `asset ?` should no longer be needed

cbachmeier (approved):
LGTM

PR #5182, opened 2022-11-11
fix(bag): review asset buttons

JFrankfurt:
https://uniswaplabs.atlassian.net/browse/WEB-2015

### DIFF
--- a/src/components/Button/index.tsx
+++ b/src/components/Button/index.tsx
@@ -473,13 +473,6 @@ const BaseThemeButton = styled.button<BaseButtonProps>`
   }
 `
 
-// enum idk {
-//   idle,
-//   hovered,
-//   pressed,
-//   disabled,
-// }
-
 interface ThemeButtonProps extends React.ComponentPropsWithoutRef<'button'>, BaseButtonProps {}
 
 export const ThemeButton = ({ ...rest }: ThemeButtonProps) => {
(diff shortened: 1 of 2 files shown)


src/components/Button/index.tsx
```diff
@@ -366,33 +366,121 @@ export function ButtonRadioChecked({ active = false, children, ...rest }: { acti
...
+  :hover {
+    ${ButtonOverlay} {
+      background-color: ${({ theme }) => theme.stateOverlayHover};
+    }
   }
 `
+
+interface ThemeButtonProps extends React.ComponentPropsWithoutRef<'button'>, BaseButtonProps {}
+
+export const ThemeButton = ({ ...rest }: ThemeButtonProps) => {
```
aballerr, line 478: Should it be this?

```{children, ...rest}```

Comment (JFrankfurt) [reply src/components/Button/index.tsx:478]: yes ty

Comment (JFrankfurt) [reply src/components/Button/index.tsx]: done

File: src/nft/components/bag/BagRow.tsx [hunk 2]
```diff
@@ -24,13 +24,20 @@ import styled from 'styled-components/macro'
 
 import * as styles from './BagRow.css'
 
-const RemoveButton = styled(MediumButton)`
+const RemoveButton = styled(ThemeButton)`
   border-radius: 12px;
```
vm, line 28: ideally shouldn't need some of this theming now?

vm (approved):
would ideally live in nft/components instead of top-level, but i'm okay with this incremental step towards a shared button

PR #5185, opened 2022-11-11
fix(bag): simplify header

JFrankfurt:
https://uniswaplabs.atlassian.net/browse/WEB-2059

### DIFF
--- a/src/nft/components/bag/BagHeader.tsx
+++ b/src/nft/components/bag/BagHeader.tsx
@@ -24,9 +24,9 @@ const IconWrapper = styled.button`
   color: ${({ theme }) => theme.textPrimary};
   cursor: pointer;
   display: flex;
-  flex: 1 1 auto;
   flex-direction: row;
   justify-content: flex-end;
+  margin-left: auto;
   padding: 2px;
   opacity: 1;
   &:hover {
--- a/src/nft/components/bag/BagHeader.tsx
+++ b/src/nft/components/bag/BagHeader.tsx
@@ -38,7 +38,7 @@ const IconWrapper = styled.button`
     },
   }) => `opacity ${duration.medium} ${timing.ease}`};
 `
-const NumberDot = styled.div`
+const CounterDot = styled.div`
   align-items: center;
   background-color: ${({ theme }) => theme.accentAction};
   border-radius: 100px;
@@ -70,7 +70,7 @@ export const BagHeader = ({ numberOfAssets, closeBag, resetFlow, isProfilePage }
       <ThemedText.HeadlineSmall>{isProfilePage ? <Trans>Sell</Trans> : <Trans>Bag</Trans>}</ThemedText.HeadlineSmall>
       {numberOfAssets > 0 && (
         <>
-          <NumberDot>{numberOfAssets}</NumberDot>
+          <CounterDot>{numberOfAssets}</CounterDot>
           <ClearButton onClick={resetFlow}>Clear all</ClearButton>
         </>
       )}
--- a/src/nft/components/bag/BagHeader.tsx
+++ b/src/nft/components/bag/BagHeader.tsx
@@ -1,10 +1,7 @@
-import { Column, Row } from 'nft/components/Flex'
+import { Trans } from '@lingui/macro'
 import { BagCloseIcon } from 'nft/components/icons'
-import { roundAndPluralize } from 'nft/utils/roundAndPluralize'
 import styled from 'styled-components/macro'
 import { ButtonText, ThemedText } from 'theme'
-
-import * as styles from './BagHeader.css'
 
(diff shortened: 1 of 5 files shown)


src/nft/components/bag/EmptyContent.tsx
```diff
@@ -6,8 +6,8 @@ import { themeVars } from 'nft/css/sprinkles.css'
 import styled from 'styled-components/macro'
 
 const StyledColumn = styled(Column)<{ isProfilePage?: boolean }>`
-  gap: ${({ isProfilePage }) => !isProfilePage && '12'};
-  margin-top: 36;
+  gap: ${({ isProfilePage }) => !isProfilePage && '12'}px;
```
aballerr, line 9: nit: maybe just ```'12px'```

aballerr (dismissed):
I posted screenshots in the jira that shows this, I think we need a bit more padding/margin below the titles

vm (approved):

src/nft/components/bag/BagHeader.tsx
```diff
@@ -47,6 +43,26 @@ const IconWrapper = styled.button`
     },
   }) => `opacity ${duration.medium} ${timing.ease}`};
 `
+const NumberDot = styled.div`
+  align-items: center;
+  background-color: ${({ theme }) => theme.accentActive};
+  border-radius: 50%;
```
lynnshaoyu, line 49: border radius should be 100px according to figma

src/nft/components/bag/BagHeader.tsx
```diff
@@ -56,19 +72,17 @@ interface BagHeaderProps {
...
-          <BagCloseIcon />
-        </IconWrapper>
-      </Row>
+    <Wrapper>
+      <ThemedText.HeadlineSmall>{isProfilePage ? <Trans>Sell</Trans> : <Trans>Bag</Trans>}</ThemedText.HeadlineSmall>
       {numberOfAssets > 0 && (
-        <ControlRow>
-          <ThemedText.BodyPrimary>{roundAndPluralize(numberOfAssets, 'NFT')}</ThemedText.BodyPrimary>
+        <>
+          <NumberDot>{numberOfAssets}</NumberDot>
```
lynnshaoyu, line 79: Can you wrap numberOfAssets in its own div, and set padding to 4px on the left and right.

Setting a fixed height and width in NumberDot leads to situations like this if the number is very long: 
Screen Shot 2022-11-14 at 12 27 06 PM

src/nft/components/bag/BagHeader.tsx
```diff
@@ -47,6 +43,26 @@ const IconWrapper = styled.button`
     },
   }) => `opacity ${duration.medium} ${timing.ease}`};
 `
+const NumberDot = styled.div`
+  align-items: center;
+  background-color: ${({ theme }) => theme.accentActive};
+  border-radius: 50%;
+  display: flex;
+  font-size: 10px;
```
lynnshaoyu, line 51: can we extract font settings to a new preset style called "utilityBadge" in src/nft/css/common.css.ts?

In Figma it seems like this is a new typography preset style: 
//styleName: Utility/Badge;
font-family: Inter;
font-size: 10px;
font-weight: 600;
line-height: 12px;
letter-spacing: 0em;
text-align: center;

https://www.figma.com/file/bCN7G6oArm7pPts03jvJ5x/Unification?node-id=17444%3A89794&t=v1DnxqDOOGLhXeum-0

lynnshaoyu: also this can be applied to the text div described in below comment

src/nft/components/bag/BagHeader.tsx
lynnshaoyu, line 49: border radius should be 100px according to Figma, it makes a visual difference and looks closer to the design.

lynnshaoyu (dismissed):

src/nft/components/bag/BagHeader.tsx
vm: i think this could wait until we handle Text styling in the component library, as I don't think this is particularly solidified in the design system yet

src/nft/components/bag/BagHeader.tsx
lynnshaoyu: Also in figma the NumberDot element is called CounterDot, might be helpful to rename for consistency with future uses of this component in Figma / designs.

JFrankfurt: Out max bag size is 100, so the most we can expect is triple digits here 
image

JFrankfurt: yeah, the pr now does that--just took this pic before I'd made that change (removing the fixed width).

PR #5187, opened 2022-11-12
fix: approve button font size

saranonearth:
### Changes
Styled the home screen "`Allow the uniswap Protocol to user your `TOKEN``" CTA. Was looking super odd with 20px

### Before
before-fix
### Now 🦄
before-fix

### DIFF
--- a/src/pages/Swap/index.tsx
+++ b/src/pages/Swap/index.tsx
@@ -677,7 +677,15 @@ export default function Swap() {
                         }
                       >
                         <AutoRow justify="space-between" style={{ flexWrap: 'nowrap' }} height="20px">
-                          <span style={{ display: 'flex', alignItems: 'center' }}>
+                          <span
+                            style={{
+                              display: 'flex',
+                              alignItems: 'center',
+                              fontSize: '16px',
+                              width: '100%',
+                              justifyContent: 'center',
+                            }}
+                          >
                             {/* we need to shorten this string on mobile */}
                             {approvalState === ApprovalState.APPROVED ||
                             signatureState === UseERC20PermitState.SIGNED ? (


src/pages/Swap/index.tsx
```diff
@@ -677,7 +677,15 @@ export default function Swap() {
...
                         <AutoRow justify="space-between" style={{ flexWrap: 'nowrap' }} height="20px">
-                          <span style={{ display: 'flex', alignItems: 'center' }}>
+                          <span
+                            style={{
+                              display: 'flex',
+                              alignItems: 'center',
+                              fontSize: '16px',
+                              width: '100%',
+                              justifyContent: 'center',
+                            }}
```
aballerr, line 687: can you use a styled-component here?

src/pages/Swap/index.tsx
```diff
@@ -677,7 +677,15 @@ export default function Swap() {
...
+                          >
```
aballerr, line 688: can you use a styled-component here?

aballerr (changes requested):

callil:
Font size looks good!

Can we `text-align: left` for mobile sizes (or just in general) so that it appears like this in the overflow state:

Screen Shot 2022-11-14 at 12 02 29 PM

saranonearth:
> Font size looks good!
> 
> Can we `text-align: left` for mobile sizes (or just in general) so that it appears like this in the overflow state:
> 
> Screen Shot 2022-11-14 at 12 02 29 PM

Left align would look weird for mobiles devices I guess. Center align looks way better is what I think, would love to know what you think @aballerr @callil  

Screenshot 2022-11-14 at 10 54 03 PM

src/pages/Swap/index.tsx
saranonearth: Sure

grabbou:
I believe @aballerr meant a Text component from `ThemedText` collection (in this case, maybe `BodyPrimary` or `BodySecondary`?)

src/pages/Swap/index.tsx
```diff
@@ -677,15 +677,15 @@ export default function Swap() {
                         }
                       >
                         <AutoRow justify="space-between" style={{ flexWrap: 'nowrap' }} height="20px">
-                          <span style={{ display: 'flex', alignItems: 'center' }}>
+                          <Text display="flex" fontSize={'16px'} justifyContent={'center'} width={'100%'}>
```
grabbou, line 680: Is there a pattern in the codebase for using plain strings vs wrapping them with curlies? I think I've seen both in the codebase. Either way, it would be great to maintain consistency going forward!

saranonearth: @aballerr, could you clarify which one would keep things consistent? Using styled-component `Text` from rebass or `ThemedText`. Would make the changes accordingly. Thanks

saranonearth: Thanks @grabbou, looks like the ThemedText pattern is prominent in the codebase, have changed to that. Could you review again? Thanks
cc: @aballerr

saranonearth: Also, ended up using `ThemedText.SubHeader` over `ThemedText.BodyPrimary` because of the fontWeight, ThemedText.BodyPrimary with fontWeight 400 looked odd in compared to the `Swap` text below. Hope this works.

### With ThemedText.BodyPrimary
Screenshot 2022-11-15 at 9 07 10 AM

### With ThemedText.SubHeader
Screenshot 2022-11-12 at 12 35 13 PM

src/pages/Swap/index.tsx
```diff
@@ -677,15 +677,15 @@ export default function Swap() {
                         }
                       >
                         <AutoRow justify="space-between" style={{ flexWrap: 'nowrap' }} height="20px">
-                          <span style={{ display: 'flex', alignItems: 'center' }}>
+                          <ThemedText.SubHeader width={'100%'} textAlign="center">
```
grabbou, line 680: I know that it's super inconsistent right now throughout the codebase, but I'll be sending a PR later today to switch to plain strings instead of `{'100%'}` (should be `"100%"`). Feel free to update if you have time. If not - my PR will most likely run an update globally on the entire code.

grabbou (approved):
🤙

src/pages/Swap/index.tsx
saranonearth: I have changed it to plain string, @grabbou. Cheers

aballerr (approved):

saranonearth:
@grabbou @aballerr can go ahead and merge this PR? Thanks

Dacaryare143:
#5187

PR #5195, opened 2022-11-14
style: Update banner text, font size, color.

unipadmini:
Updated the banner text content, size, and color. 

Fixes https://uniswaplabs.atlassian.net/browse/WEB-2047.

### DIFF
--- a/src/nft/components/explore/Banner.tsx
+++ b/src/nft/components/explore/Banner.tsx
@@ -39,11 +39,7 @@ const HeaderContainer = styled.div`
   align-items: start;
   padding-top: 40px;
   flex-shrink: 0;
-  background: linear-gradient(180deg, rgba(255, 255, 255, 0.4) 0%, rgba(255, 255, 255, 0) 100%), #fc72ff;
-  -webkit-background-clip: text;
-  -webkit-text-fill-color: ${({ theme }) => theme.textPrimary};
-  background-clip: text;
-  text-fill-color: ${({ theme }) => theme.textPrimary};
+  color: ${({ theme }) => theme.textPrimary};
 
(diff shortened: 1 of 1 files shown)


src/nft/components/explore/Banner.tsx
```diff
@@ -33,17 +33,17 @@ const HeaderContainer = styled.div`
...
+  font-size: 72px;
   line-height: 73px;
   justify-content: start;
   align-items: start;
   padding-top: 40px;
   flex-shrink: 0;
   background: linear-gradient(180deg, rgba(255, 255, 255, 0.4) 0%, rgba(255, 255, 255, 0) 100%), #fc72ff;
   -webkit-background-clip: text;
-  -webkit-text-fill-color: transparent;
+  -webkit-text-fill-color: ${({ theme }) => theme.textPrimary};
```
vm, line 44: we shouldn't need this because we have auto prefixer

src/nft/components/explore/Banner.tsx
```diff
@@ -33,17 +33,17 @@ const HeaderContainer = styled.div`
   display: flex;
   max-width: 500px;
   font-weight: 500;
-  font-size: 60px;
+  font-size: 72px;
   line-height: 73px;
```
vm, line 37: do we need this one off line height? if so does it need to change?

src/nft/components/explore/Banner.tsx
unipadmini: actually without this, it doesn't work on chrome

vm: let me see why, that's odd to me

PR #5199, opened 2022-11-14
feat: add market logos to filters and chips

cbachmeier:
Adds marketplace logos to filter and chips on collection page

https://uniswaplabs.atlassian.net/browse/WEB-2058

### DIFF
--- a/src/nft/components/collection/CollectionNfts.tsx
+++ b/src/nft/components/collection/CollectionNfts.tsx
@@ -45,7 +45,7 @@ import styled from 'styled-components/macro'
 import { ThemedText } from 'theme'
 
 import { CollectionAssetLoading } from './CollectionAssetLoading'
-import { MARKETPLACE_ITEMS } from './MarketplaceSelect'
+import { MARKETPLACE_ITEMS, MarketplaceLogo } from './MarketplaceSelect'
 import { Sweep, useSweepFetcherParams } from './Sweep'
 import { TraitChip } from './TraitChip'
 
@@ -164,6 +164,10 @@ export const LoadingButton = styled.div`
   );
   will-change: background-position;
   background-size: 400%;
+`
+
+const MarketNameWrapper = styled(Row)`
+  gap: 8px;
 `
 
 const loadingAssets = (height?: number) => (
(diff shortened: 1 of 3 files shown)


src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -459,7 +463,12 @@ export const CollectionNfts = ({ contractAddress, collectionStats, rarityVerifie
             {markets.map((market) => (
               <TraitChip
                 key={market}
-                value={MARKETPLACE_ITEMS[market as keyof typeof MARKETPLACE_ITEMS]}
+                value={
+                  <MarketNameWrapper>
+                    <MarketplaceLogo src={`/nft/svgs/marketplaces/${market.toLowerCase()}.svg`} />
+                    {MARKETPLACE_ITEMS[market as keyof typeof MARKETPLACE_ITEMS]}
```
vm, line 469: rather than casting you can check if the key exists in the map? it also should work if it's a record and you type markets with the same type

src/nft/components/collection/MarketplaceSelect.tsx
```diff
@@ -34,12 +44,19 @@ export const MARKETPLACE_ITEMS = {
...
+  return `/nft/svgs/marketplaces/${marketplaceItem}.svg`
+}
+
 export const FilterItem = ({
   title,
   element,
   onClick,
 }: {
-  title: string
+  title: string | JSX.Element
```
vm, line 59: ReactNode should work

vm (approved):

src/nft/components/collection/MarketplaceSelect.tsx
cbachmeier: Is there a reason for using one of the other?

src/nft/components/collection/CollectionNfts.tsx
cbachmeier: We'd still need the cast as Object().keys only returns string[]

src/nft/components/collection/MarketplaceSelect.tsx
```diff
@@ -34,12 +44,19 @@ export const MARKETPLACE_ITEMS = {
   sudoswap: 'SudoSwap',
 }
 
+function getMarketLogoSrc(market: string) {
+  const marketplaceItem = Object.keys(MARKETPLACE_ITEMS).find(
+    (key) => MARKETPLACE_ITEMS[key as keyof typeof MARKETPLACE_ITEMS] === market
+  )
+  return `/nft/svgs/marketplaces/${marketplaceItem}.svg`
+}
```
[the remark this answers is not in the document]
cbachmeier: We need the key not the value from the map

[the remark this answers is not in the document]
grabbou: This is not getting executed too often, so I don't think that optimisation matters that much. Running Object.keys, followed by `find` seems a bit redundant. I think either a `useMemo` or ideally, a single function to flip keys with values once for quick indexing would work a bit better!

PR #5203, opened 2022-11-14
refactor: lazy load asset details and profile page

cbachmeier:
Uses proper pattern for loading and lazy loading data for profile and asset details page.

### DIFF
--- a/src/graphql/data/nft/NftBalance.ts
+++ b/src/graphql/data/nft/NftBalance.ts
@@ -114,13 +114,23 @@ type NftBalanceQueryAsset = NonNullable<
   NonNullable<NonNullable<NftBalanceQuery_nftBalances$data['nftBalances']>['edges']>[number]
 >
 
-export function useLoadNftBalanceQuery(ownerAddress?: string | string[]): void {
+export function useLoadNftBalanceQuery(
+  ownerAddress?: string,
+  collectionAddress?: string | string[],
+  tokenId?: string
+): void {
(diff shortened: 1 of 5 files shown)


src/graphql/data/nft/NftBalance.ts
```diff
@@ -111,26 +113,54 @@ const nftBalanceQuery = graphql`
...
+): void {
+  const [, loadQuery] = useQueryLoader(nftBalanceQuery)
+  useEffect(() => {
+    if (ownerAddress) {
+      loadQuery({
+        ownerAddress,
+        filter: tokenId
+          ? { assets: [{ address: collectionAddress, tokenId }] }
+          : { addresses: Array.isArray(collectionAddress) ? collectionAddress : [collectionAddress] },
+        first: tokenId ? 1 : DEFAULT_WALLET_ASSET_QUERY_AMOUNT,
```
vm, line 130: not clear to me what the `first` means here

vm: basically, when would tokenId get passed in

vm (approved):

src/graphql/data/nft/NftBalance.ts
cbachmeier: first means GraphQL will return the first `n` assets. When tokenId is passed in we expect only 1 item to return that corresponds to the given collection and tokenId, otherwise we expect to be grabbing the standard page size, ie first 25 items the user owns

PR #5204, opened 2022-11-14
feat: Details List CTAs correctly routes users to the list page with asset selected

cbachmeier:
When on the Details page for an asset the user owns, the List CTAs direct the user to the listing page with the viewed asset selected. Depends on the changes made in #5203. Also removes Cancel CTA as per https://uniswaplabs.atlassian.net/browse/WEB-2160

Note:
- Ownership/Listing info is inconsistent and BE is currently working on improving its accuracy
- `Cancel Listing` will require custom logic but is out of the scope of this task and is tracked separately


src/nft/pages/profile/profile.tsx
```diff
@@ -35,12 +35,16 @@ const ProfileContent = () => {
   }, [removeAllMarketplaceWarnings, sellPageState, setListingStatus])
 
   const { account } = useWeb3React()
+  const accountRef = useRef(account)
```
vm, line 38: why is this needed?

cbachmeier: The desired outcome in the useEffect was so that when a user changed wallet all the profile states would reset. However just using `account` as the useEffect condition also caused the reset to happen on pageload which prevented me from navigating to the listing page with a pre-selected asset. By adding the check to make for accountRef.current, the expected behaviour now occurs.

vm (approved):

PR #5210, opened 2022-11-14
fix: err and warn if user modifies tx

zzmp:
Omit tracking modified transactions, and instead show the user an error modal warning them to immediately cancel their tx.

Some wallets allow users to modify transactions before approving them, but this (a) risks converting the tx to something else, like a `transfer()`; and (b) means that we no longer have accurate data with which to track the tx in our transactions history views.

image

### DIFF
--- a/src/lib/hooks/swap/useSendSwapTransaction.tsx
+++ b/src/lib/hooks/swap/useSendSwapTransaction.tsx
@@ -29,7 +29,7 @@ interface FailedCall extends SwapCallEstimate {
   error: Error
 }
 
-class SwapError extends Error {}
+class InvalidSwapError extends Error {}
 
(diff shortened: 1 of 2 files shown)


src/lib/hooks/swap/useSendSwapTransaction.tsx
```diff
@@ -29,6 +29,8 @@ interface FailedCall extends SwapCallEstimate {
   error: Error
 }
 
+class SwapError extends Error {}
```
vm, line 32: perhaps name this `InvalidSwapError` or something. such that it's clear it's prior to executing the swap?

vm (approved):
feel free to edit the copy. idk if we care about translating this.

zzmp:
I'd also like analytics for when/if this happens.

PR #5213, opened 2022-11-14
fix: update analytics package to use default event

dannydaniil:
# Notes

- This PR integrates the codebase with the `1.1.1` version of the analytics package.
- The main driver for the change is that the new package's initialization method accepts a `defaultEventName`, which used to be hardcoded in the *analytics* package.
- The hardcoding (as a string vs an enum) in the *anaytics* package, cause a misanming to occur. The reason it was hardcoded as a string inside the *analytics* package, is because we don't want the *infra* portion to be tied to the *analytics-events* package
- This change ensures that the consuming app defines the `defaultEventName`

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,7 +131,7 @@
     "@react-hook/window-scroll": "^1.3.0",
     "@reduxjs/toolkit": "^1.6.1",
     "@types/react-relay": "^13.0.2",
-    "@uniswap/analytics": "1.0.3",
+    "@uniswap/analytics": "1.1.1",
     "@uniswap/analytics-events": "1.1.0",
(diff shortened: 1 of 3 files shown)


adjkant (approved):

zzmp (changes requested):
Please either clear out the PR description or (preferably) update it to describe the PR

package.json
```diff
@@ -238,4 +238,4 @@
     "node": "14",
     "yarn": ">=1.22"
   }
-}
+}
```
zzmp, line 241: keep the newline at the end (otherwise its unnecessary file churn)

zzmp (approved):

PR #5214, opened 2022-11-15
fix: do not show floor/volume change on all

grabbou:
Fixes WEB-2225


vm (approved):

grabbou:
@vm had to apply small changes to satisfy Typescript and also make sure the text renders correctly

vm (approved):

PR #5215, opened 2022-11-15
fix: remove cryptopunks from trending 

grabbou:
Fixes #2140


src/nft/components/explore/Banner.tsx
```diff
@@ -77,6 +85,8 @@ const Banner = () => {
     }
   )
 
+  const collections = data?.filter((collection) => !EXCLUDED_COLLECTIONS.includes(collection.address))
```
cbachmeier, line 88: Also set a slice to prevent 6 from showing if punks isn't top 5

src/nft/components/explore/Banner.tsx
```diff
@@ -61,14 +61,22 @@ const HeaderContainer = styled.div`
 `
 const DEFAULT_TRENDING_COLLECTION_QUERY_AMOUNT = 5
 
+// Exclude collections that are not available in any of the following - OpenSea, X2Y2 and LooksRare:
+// CryptoPunkts
```
cbachmeier, line 65: don't need to name them in the comment

src/nft/components/explore/Banner.tsx
```diff
@@ -61,14 +61,22 @@ const HeaderContainer = styled.div`
...
+const EXCLUDED_COLLECTIONS = ['0xb47e3cd837ddf8e4c57f05d70ab865de6e193bbb']
```
cbachmeier, line 66: We could also make this more dynamic where instead of a blocklist we filter collections where total number of  listings across the 3 marketplaces is 0. We can grab top 25, filter, and slice for max 5

grabbou: Yes! I also thought about that, but the ticket itself treats this as a one-off edge case and asks to filter it out. We can follow-up there too.

grabbou:
I am going to blame autocorrection for the title misspellings

src/nft/components/explore/Banner.tsx
grabbou: Good catch!

src/nft/components/explore/Banner.tsx
JFrankfurt: agree with charlie--don't think we should introduce an exceptions list.

grabbou: Cool, I just let Patric know that we're changing this to be more generic. I will ping you when it's implemented.

grabbou:
Moving to drafts as we're having an ongoing discussion internally on best way forward here

src/nft/components/explore/Banner.tsx
```diff
@@ -68,7 +70,11 @@ const Banner = () => {
   const { data: collections } = useQuery(
     ['trendingCollections'],
     () => {
-      return fetchTrendingCollections({ volumeType: 'eth', timePeriod: TimePeriod.OneDay, size: 5 })
+      return fetchTrendingCollections({
+        volumeType: 'eth',
+        timePeriod: TimePeriod.OneDay,
+        size: TRENDING_COLLECTION_SIZE + LOADING_BUFFER,
```
JFrankfurt, line 76: what's the purpose of the +5 increase here? Is it just to make it more likely there's a renderable collection?

src/nft/components/explore/Banner.tsx
```diff
@@ -2,7 +2,7 @@ import { useLoadCollectionQuery } from 'graphql/data/nft/Collection'
 import { useIsMobile } from 'nft/hooks'
 import { fetchTrendingCollections } from 'nft/queries'
 import { TimePeriod } from 'nft/types'
-import { Suspense, useMemo } from 'react'
+import React, { Suspense, useMemo } from 'react'
```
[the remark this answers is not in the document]
grabbou: done. interestingly, I would expect ESLint to get rid of that

src/nft/components/explore/Banner.tsx
grabbou: left overs from the previous wip take on dynamic filtering (where @cbachmeier suggested we fetch like 20-25 items). With hardcoded addresses, we'll just extend by the length of the excluded array.

src/nft/components/explore/Banner.tsx
```diff
@@ -77,6 +84,8 @@ const Banner = () => {
     }
   )
 
+  const collections = data?.filter((collection) => !EXCLUDED_COLLECTIONS.includes(collection.address)).slice(0, 5)
```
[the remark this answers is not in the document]
grabbou: done! I think I would argue whether there's a benefit of using useMemo in this context instead of filtering on a relatively small data set. However, I think it makes sense to keep the conde consistent (especially in context of this file), subject to any changes in the future!

cbachmeier (approved):
LGTM

PR #5217, opened 2022-11-15
fix: add filter text when filter button open

lynnshaoyu:
1) https://uniswaplabs.atlassian.net/browse/WEB-1640?atlOrigin=eyJpIjoiY2NhN2Q0ZmFmZWEzNGM4YWEwYTZlYjA3N2RlZjk1ZmEiLCJwIjoiaiJ9

2) https://uniswaplabs.atlassian.net/browse/WEB-2300


aballerr (commented):
If you open the filter and then shrink the screen (for example ipad), the search is very small now.  I think on smaller screen sizes less than(1024) this current behavior makes sense.  Can you double check on this? 

Also, should this show up on mobile? it does not currently.

I attached some images to jira

src/nft/components/collection/FilterButton.tsx
```diff
@@ -36,12 +36,10 @@ export const FilterButton = ({
       <FilterIcon />
       {!isMobile ? (
         <>
```
JFrankfurt, line 38: can drop the fragment here I think

JFrankfurt (approved):

PR #5219, opened 2022-11-15
build: don't fail local build on lint

vm:
When developing it is really helpful to comment out sections of code (for example, if wanting to test a loading or failure state). However, the no-unused-imports for example will get triggered quickly when that is done.

We can ignore these lint issues on build, and devs can add the requirements to their editor, pre-commit hook, etc. The build will also fail on GitHub anyways, so it will be obvious.


cbachmeier:
This is a blessed fix

aballerr (approved):

zzmp:
How would you typecheck?

PR #5223, opened 2022-11-15
fix(token-details): remove balance summary links to current page

JFrankfurt:
Balance summary was linking to its current page (no-op) so removed it.


zzmp (approved):
excellent. might want to reach out to design to see if it was intended to link elsewhere.

JFrankfurt:
@zzmp I think this is a remnant of when we were going to do multi-network balances and link across network pages (and wrapped assets)

PR #5225, opened 2022-11-15
fix: adding amm pricing updates

JackShort:
updating amm pricing on addition to bag: https://uniswaplabs.atlassian.net/browse/WEB-1943

### DIFF
--- a/src/nft/components/collection/CollectionAsset.tsx
+++ b/src/nft/components/collection/CollectionAsset.tsx
@@ -1,3 +1,4 @@
+import { BigNumber } from '@ethersproject/bignumber'
 import { sendAnalyticsEvent, useTrace } from '@uniswap/analytics'
 import { EventName, PageName } from '@uniswap/analytics-events'
 import { useBag } from 'nft/hooks'
@@ -56,16 +57,18 @@ export const CollectionAsset = ({
   }, [asset])
 
   const handleAddAssetToBag = useCallback(() => {
-    addAssetsToBag([asset])
-    if (!bagExpanded && !isMobile && !bagManuallyClosed) {
-      setBagExpanded({ bagExpanded: true })
+    if (BigNumber.from(asset.priceInfo ? asset.priceInfo.ETHPrice : 0).gte(0)) {
+      addAssetsToBag([asset])
+      if (!bagExpanded && !isMobile && !bagManuallyClosed) {
+        setBagExpanded({ bagExpanded: true })
+      }
+      sendAnalyticsEvent(EventName.NFT_BUY_ADDED, {
+        collection_address: asset.address,
+        token_id: asset.tokenId,
+        token_type: asset.tokenType,
+        ...trace,
+      })
     }
-    sendAnalyticsEvent(EventName.NFT_BUY_ADDED, {
-      collection_address: asset.address,
-      token_id: asset.tokenId,
-      token_type: asset.tokenType,
-      ...trace,
-    })
   }, [addAssetsToBag, asset, bagExpanded, bagManuallyClosed, isMobile, setBagExpanded, trace])
 
   const handleRemoveAssetFromBag = useCallback(() => {
--- a/src/nft/components/collection/CollectionNfts.tsx
+++ b/src/nft/components/collection/CollectionNfts.tsx
@@ -1,3 +1,4 @@
+import { BigNumber } from '@ethersproject/bignumber'
 import { TraceEvent } from '@uniswap/analytics'
 import { BrowserEvent, ElementName, EventName } from '@uniswap/analytics-events'
 import { useWeb3React } from '@web3-react/core'
@@ -34,12 +35,11 @@ import {
 } from 'nft/hooks'
 import { useIsCollectionLoading } from 'nft/hooks/useIsCollectionLoading'
 import { usePriceRange } from 'nft/hooks/usePriceRange'
-import { DropDownOption, GenieCollection, Markets, TokenType } from 'nft/types'
-import { getRarityStatus } from 'nft/utils/asset'
-import { pluralize } from 'nft/utils/roundAndPluralize'
+import { DropDownOption, GenieAsset, GenieCollection, Markets, TokenType } from 'nft/types'
+import { calcPoolPrice, getRarityStatus, pluralize } from 'nft/utils'
 import { scrollToTop } from 'nft/utils/scrollToTop'
 import { applyFiltersFromURL, syncLocalFiltersWithURL } from 'nft/utils/urlParams'
-import { useEffect, useMemo, useRef, useState } from 'react'
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import InfiniteScroll from 'react-infinite-scroll-component'
 import { useLocation } from 'react-router-dom'
 import styled from 'styled-components/macro'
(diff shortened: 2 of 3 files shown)


src/nft/components/collection/CollectionAsset.tsx
```diff
@@ -56,16 +57,18 @@ export const CollectionAsset = ({
   }, [asset])
 
   const handleAddAssetToBag = useCallback(() => {
-    addAssetsToBag([asset])
-    if (!bagExpanded && !isMobile && !bagManuallyClosed) {
-      setBagExpanded({ bagExpanded: true })
+    if (BigNumber.from(asset.priceInfo ? asset.priceInfo.ETHPrice : 0).gte(0)) {
```
JFrankfurt, line 60: nit: can you do `asset.priceInfo?.ETHPrice ?? 0` here instead?

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -287,6 +288,72 @@ export const CollectionNfts = ({ contractAddress, collectionStats, rarityVerifie
...
+      (asset) =>
+        (asset.marketplace === Markets.NFTX || asset.marketplace === Markets.NFT20) &&
+        (asset.priceInfo.ETHPrice = calculatePrice(asset))
+    )
+
+    if (sortBy === SortBy.HighToLow || sortBy === SortBy.LowToHigh) {
+      collectionAssets.sort((a, b) => {
+        const bigA = BigNumber.from(a.priceInfo.ETHPrice ?? -1)
+        const bigB = BigNumber.from(b.priceInfo.ETHPrice ?? -1)
+        const diff = bigA.sub(bigB)
```
JFrankfurt, line 338: nit: define `diff` after the first return block

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -287,6 +288,72 @@ export const CollectionNfts = ({ contractAddress, collectionStats, rarityVerifie
...
+
+        if ((bigA.gte(0) || bigB.gte(0)) && (bigA.lt(0) || bigB.lt(0))) {
+          return bigA.gte(0) ? (sortBy === SortBy.LowToHigh ? -1 : 1) : sortBy === SortBy.LowToHigh ? 1 : -1
```
JFrankfurt, line 341: nit: unpack this compound ternary into nest if/else so it's a little easier to follow (I like lines 344-348 below--that could be written this way, but it's easier to follow w/ if/elseIf and return <single ternary expression>

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -287,6 +288,72 @@ export const CollectionNfts = ({ contractAddress, collectionStats, rarityVerifie
...
+    collectionAssets.forEach(
+      (asset) =>
+        (asset.marketplace === Markets.NFTX || asset.marketplace === Markets.NFT20) &&
+        (asset.priceInfo.ETHPrice = calculatePrice(asset))
+    )
+
+    if (sortBy === SortBy.HighToLow || sortBy === SortBy.LowToHigh) {
+      collectionAssets.sort((a, b) => {
+        const bigA = BigNumber.from(a.priceInfo.ETHPrice ?? -1)
+        const bigB = BigNumber.from(b.priceInfo.ETHPrice ?? -1)
```
JFrankfurt, line 337: above you access ETHPrice differently--are you sure this works and you don't need the `?` after `priceInfo`?

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -287,6 +288,72 @@ export const CollectionNfts = ({ contractAddress, collectionStats, rarityVerifie
...
+              .filter((item) => item.asset.address === asset.address && item.asset.marketplace === asset.marketplace)
+              .map((item) => item.asset.tokenId)
+              .indexOf(asset.tokenId)
+          )
+        : calcPoolPrice(
+            asset,
+            itemsInBag.filter(
+              (item) => item.asset.address === asset.address && item.asset.marketplace === asset.marketplace
+            ).length
+          )
```
JFrankfurt, line 313: should use the ternary to determine arguments and only write one function call

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -287,6 +288,72 @@ export const CollectionNfts = ({ contractAddress, collectionStats, rarityVerifie
 
   const { assets: collectionNfts, loadNext, hasNext, isLoadingNext } = useLazyLoadAssetsQuery(assetQueryParams)
 
+  const isSelected = useCallback(
```
JFrankfurt, line 291: nit: `is` prefix makes me assume this is a boolean value, not a function

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -287,6 +288,72 @@ export const CollectionNfts = ({ contractAddress, collectionStats, rarityVerifie
...
+
+  const collectionAssets = useMemo(() => {
+    if (
+      !collectionNfts ||
+      !collectionNfts.some((asset) => asset.marketplace === Markets.NFTX || asset.marketplace === Markets.NFT20)
+    ) {
+      return collectionNfts
+    }
+
+    const collectionAssets = [...collectionNfts]
```
JFrankfurt, line 326: is this line necessary?

src/nft/utils/currency.ts
```diff
@@ -54,7 +54,7 @@ export const ethNumberStandardFormatter = (
   const amountInDecimals = parseFloat(amount.toString())
   const conditionalDollarSign = includeDollarSign ? '$' : ''
 
-  if (amountInDecimals === 0) return '-'
+  if (amountInDecimals <= 0) return '-'
```
JFrankfurt, line 57: double checking this is intentional

src/nft/components/collection/CollectionNfts.tsx
JackShort: yes it does seem to make a difference

src/nft/utils/currency.ts
JackShort: yes

src/nft/components/collection/CollectionNfts.tsx
JFrankfurt: it should overwrite the line 315 value of `collectionAssets` mutate it conditionally in the `forEach`, then potentially mutate it again with .sort, then return a new value for it that `useMemo` will use. I expect there's some weird variable shadowing issue here. I think you should make this work as you need w/o overwriting collectionAssets. I'm pretty concerned about unexpected buggy behavior with this as written.

JFrankfurt (changes requested):
see https://github.com/Uniswap/interface/pull/5225/files#r1023286511

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -287,6 +288,71 @@ export const CollectionNfts = ({ contractAddress, collectionStats, rarityVerifie
...
+    (asset: GenieAsset) => {
+      return calcPoolPrice(
+        asset,
+        assetIsSelected(asset)
+          ? itemsInBag
+              .filter((item) => item.asset.address === asset.address && item.asset.marketplace === asset.marketplace)
+              .map((item) => item.asset.tokenId)
+              .indexOf(asset.tokenId)
+          : itemsInBag.filter(
+              (item) => item.asset.address === asset.address && item.asset.marketplace === asset.marketplace
```
JFrankfurt, line 308: can you define this filter callback in one spot too?

JFrankfurt (approved):

PR #5227, opened 2022-11-15
fix: limit max volume change value to 9999%

grabbou:
Fix WEB-2246, implemented a workaround as per ticket

### DIFF
--- a/src/nft/components/explore/CollectionTable.tsx
+++ b/src/nft/components/explore/CollectionTable.tsx
@@ -3,15 +3,7 @@ import { CollectionTableColumn, TimePeriod } from 'nft/types'
 import { useMemo } from 'react'
 import { CellProps, Column, Row } from 'react-table'
 
-import {
-  ChangeCell,
-  ChangeTextCell,
-  CollectionTitleCell,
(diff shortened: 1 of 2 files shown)


src/nft/components/explore/CollectionTable.tsx
```diff
@@ -113,7 +115,7 @@ const CollectionTable = ({ data, timePeriod }: { data: CollectionTableColumn[];
           return timePeriod === TimePeriod.AllTime ? (
             <TextCell value="-" />
           ) : (
-            <ChangeCell change={cell.row.original.volume.change} />
+            <ChangeCell change={Math.min(cell.row.original.volume.change, VOLUME_CHANGE_MAX_VALUE)} />
```
aballerr, line 118: this logic looks good to me, one thing I'm not sure on though

I think the task wants us to show  "> 9999%" and not just 9999

grabbou: makes sense! already updated the code. Wasn't sure whether to leave or keep a space between `>` and a number, pasted screenshots in the ticket.

aballerr (approved):

PR #5228, opened 2022-11-15
fix: remove Select All and Sell buttons in Profile

vm:
This is an incremental step to removing "sell mode" in the NFT product.


aballerr (approved):

src/nft/components/profile/view/ProfilePage.tsx
```diff
@@ -181,8 +131,8 @@ export const ProfilePage = () => {
                       ? ownerAssets.map((asset, index) => (
                           <div key={index}>
                             <ViewMyNftsAsset
+                              isSellMode={false}
```
cbachmeier, line 134: setting this to true would actually allow listing to be reachable in between yours and @lynnshaoyu 's PR

vm: genius

PR #5230, opened 2022-11-15
refactor: remove unused WalletAssetDisplay

vm:
While messing with sell mode this component was found unused. LMK if there's a reason to leave this up.


unipadmini (approved):
if unused, best to get rid of it! can always bring it back if we need it.

PR #5231, opened 2022-11-15
fix: sweep marketplaces

JackShort:
* allowing you to sweep specific marketplaces

### DIFF
--- a/src/nft/components/collection/Sweep.tsx
+++ b/src/nft/components/collection/Sweep.tsx
@@ -177,13 +177,17 @@ export const Sweep = ({ contractAddress, minPrice, maxPrice }: SweepProps) => {
   const nft20Assets = useLazyLoadSweepAssetsQuery(nft20Params)
 
   const { sortedAssets, sortedAssetsTotalEth } = useMemo(() => {
-    if (!collectionAssets || !nftxAssets || !nft20Assets)
+    if (!collectionAssets && !nftxAssets && !nft20Assets) {
       return { sortedAssets: undefined, sortedAssetsTotalEth: BigNumber.from(0) }
+    }
 
     let counterNFTX = 0
     let counterNFT20 = 0
 
-    let jointCollections = [...nftxAssets, ...nft20Assets]
+    let jointCollections: GenieAsset[] = []
+
+    if (nftxAssets) jointCollections = [...jointCollections, ...nftxAssets]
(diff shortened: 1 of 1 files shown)


src/nft/components/collection/Sweep.tsx
```diff
@@ -206,7 +210,14 @@ export const Sweep = ({ contractAddress, minPrice, maxPrice }: SweepProps) => {
       (asset) => BigNumber.from(asset.priceInfo.ETHPrice).gte(0) && !asset.susFlag
     )
 
-    validAssets = validAssets.slice(0, Math.max(collectionAssets.length, nftxAssets.length, nft20Assets.length))
+    validAssets = validAssets.slice(
+      0,
+      Math.max(
+        collectionAssets ? collectionAssets.length : 0,
```
vm, line 216: collectionAssets?.length ?? 0

src/nft/components/collection/Sweep.tsx
```diff
@@ -413,7 +424,7 @@ export function useSweepFetcherParams(
       if (market === 'others') {
         return { contractAddress, traits, markets }
       }
-      if (!markets.includes(market)) return { contractAddress: '', traits: [], markets: [] }
```
vm, line 416: why this change?

src/nft/components/collection/Sweep.tsx
```diff
@@ -177,13 +177,17 @@ export const Sweep = ({ contractAddress, minPrice, maxPrice }: SweepProps) => {
   const nft20Assets = useLazyLoadSweepAssetsQuery(nft20Params)
 
   const { sortedAssets, sortedAssetsTotalEth } = useMemo(() => {
-    if (!collectionAssets || !nftxAssets || !nft20Assets)
+    if (!collectionAssets && !nftxAssets && !nft20Assets) {
```
vm, line 180: why this change?

src/nft/components/collection/Sweep.tsx
JackShort: if its being filtered its already being passed as a parameter when market === `'others'`. This prevents it from happening twice

src/nft/components/collection/Sweep.tsx
JackShort: if any markets were selected it would return undefined because nftxAssets or nft20 assets would be undefined

vm (approved):

PR #5235, opened 2022-11-15
fix: use isNftProfilePage vs isSellMode in Bag

vm:
We no longer need an explicit sell mode for use in Bag. I will remove the `isSellMode` state from that hook soon.

### DIFF
--- a/src/components/NavBar/Bag.tsx
+++ b/src/components/NavBar/Bag.tsx
@@ -1,4 +1,5 @@
 import { NavIcon } from 'components/NavBar/NavIcon'
+import { useIsNftProfilePage } from 'hooks/useIsNftPage'
 import { BagIcon, HundredsOverflowIcon } from 'nft/components/icons'
 import { useBag, useSellAsset } from 'nft/hooks'
 import { useCallback, useEffect, useState } from 'react'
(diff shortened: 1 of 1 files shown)


src/components/NavBar/Bag.tsx
```diff
@@ -24,25 +25,24 @@ export const Bag = () => {
...
   )
   const handleIconClick = useCallback(() => {
-    if (isSellMode && bagExpanded) {
+    if (isNftProfilePage && bagExpanded) {
       resetSellAssets()
-      setIsSellMode(false)
     }
     setBagExpanded({ bagExpanded: !bagExpanded })
-  }, [bagExpanded, isSellMode, resetSellAssets, setBagExpanded, setIsSellMode])
+  }, [bagExpanded, isNftProfilePage, resetSellAssets, setBagExpanded])
```
aballerr, line 45: Hey,

I think you might not need this as I'm removing this behavior here

https://github.com/Uniswap/interface/pull/5224/files

vm: oh ok cool once you merge i will rebase this or close

vm: actually i think i can just close this entire pr

PR #5240, opened 2022-11-15
fix: don't include accentActiveSoft background on navicon active state

JFrankfurt:


cbachmeier (approved):
Looks good to me but check with design, I believe there was a desired active state background color cc: @infredible

PR #5245, opened 2022-11-16
feat: improve formatting of NFT prices with shared module

grabbou:
As a follow-up, we should consider using that module everywhere for NFT prices as it's compatible with the Notion spec. This is going to be tracked in a separate ticket.

### DIFF
--- a/src/nft/components/explore/CarouselCard.tsx
+++ b/src/nft/components/explore/CarouselCard.tsx
@@ -1,3 +1,4 @@
+import { formatNumberOrString, NumberType } from '@uniswap/conedison/format'
 import { loadingAnimation } from 'components/Loader/styled'
 import { LoadingBubble } from 'components/Tokens/loading'
 import { useCollectionQuery } from 'graphql/data/nft/Collection'
@@ -162,22 +163,6 @@ const TableElement = styled.div`
   line-height: 20px;
 `
 
-interface MarketplaceRowProps {
-  marketplace: string
-  floorInEth?: string
-  listings?: string
-}
-
-export const MarketplaceRow = ({ marketplace, floorInEth, listings }: MarketplaceRowProps) => {
-  return (
-    <>
-      <TableElement>{marketplace}</TableElement>
-      <TableElement>{floorInEth ?? '-'} ETH</TableElement>
-      <TableElement>{listings ?? '-'}</TableElement>
-    </>
-  )
-}
-
 interface CarouselCardProps {
   collection: TrendingCollection
   onClick: () => void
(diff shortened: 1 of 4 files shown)


src/nft/components/explore/CarouselCard.tsx
```diff
@@ -209,12 +194,13 @@ export const CarouselCard = ({ collection, onClick }: CarouselCardProps) => {
...
-              <MarketplaceRow
-                key={`CarouselCard-key-${collection.address}-${marketplace.marketplace}`}
-                marketplace={MARKETS_ENUM_TO_NAME[market]}
-                listings={marketplace.count.toString()}
-                floorInEth={marketplace.floorPrice.toString()}
-              />
+              <>
+                <TableElement>{MARKETS_ENUM_TO_NAME[market]}</TableElement>
+                <TableElement>
+                  {formatNumberOrString(marketplace.floorPrice, NumberType.NFTTokenFloorPrice)} ETH
```
zzmp, line 200: Is there a regression in not handling the null case for floor price or listings count?

src/nft/components/explore/CarouselCard.tsx
```diff
@@ -209,12 +194,13 @@ export const CarouselCard = ({ collection, onClick }: CarouselCardProps) => {
...
+                </TableElement>
+                <TableElement>{marketplace.count.toString()}</TableElement>
```
zzmp, line 202: Shouldn't this also be run through the formatter?

src/nft/components/explore/CarouselCard.tsx
```diff
@@ -209,12 +194,13 @@ export const CarouselCard = ({ collection, onClick }: CarouselCardProps) => {
               return null
             }
             return (
-              <MarketplaceRow
-                key={`CarouselCard-key-${collection.address}-${marketplace.marketplace}`}
-                marketplace={MARKETS_ENUM_TO_NAME[market]}
-                listings={marketplace.count.toString()}
-                floorInEth={marketplace.floorPrice.toString()}
-              />
+              <>
```
zzmp, line 197: Why does this no longer need to be keyed?

zzmp (changes requested):

src/nft/components/explore/CarouselCard.tsx
vm: i also don't think you need a toString

grabbou: Yeah, that is definitely not needed lol

grabbou:
Released a new version of `@uniswap/conedison` that has a proper formatter + special one with trailing zeros for the carousel. This is ready to ship.

unipadmini (approved):
Looks great!!

JFrankfurt (approved):

src/nft/components/explore/CarouselCard.tsx
```diff
@@ -164,15 +165,20 @@ const TableElement = styled.div`
...
+  listings?: number
 }
 
 export const MarketplaceRow = ({ marketplace, floorInEth, listings }: MarketplaceRowProps) => {
   return (
     <>
       <TableElement>{marketplace}</TableElement>
-      <TableElement>{floorInEth ?? '-'} ETH</TableElement>
+      <TableElement>
+        {typeof floorInEth === 'number'
```
zzmp, line 177: Wouldn't a simpler check just be `floorInEth !== undefined`?

zzmp (changes requested):

src/nft/components/explore/CarouselCard.tsx
grabbou: Yeah, that works too. Updated. Initially, I looked at the direct replacement of `??` but should've just looked up Typescript type to realize it can be either undefined or a number.

zzmp (approved):

PR #5248, opened 2022-11-16
feat: add glow to trending table

grabbou:
Fixes WEB-2178
Fixes WEB-2181

I will extend this code to accept the "network" parameter, but this will come under a different PR. There's a separate ticket for adding glow to Swap/Pool pages, and that's where the network comes into play. 

It's WEB-1809 if you want to check for more details.


src/nft/components/explore/Table.tsx
```diff
@@ -19,6 +19,11 @@ const CELL_WIDTH = '160px'
 const MOBILE_CELL_WIDTH = '240px'
 const DESKTOP_CELL_WIDTH = '360px'
 
+const GlowEffect = styled.div`
+  border-radius: 32px;
```
grabbou, line 23: Not sure if it's worth documenting, but generally, I applied a generous border radius to make sure it renders correctly when wrapping a component with border radius.

cbachmeier (approved):
LGTM

PR #5249, opened 2022-11-16
feat: Web 2233 erc 1155 card redesign

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-2233?atlOrigin=eyJpIjoiZWQ1OWI2ZTA3Y2RjNDcwZGFiNDIzMDZiMzU4ZjE0OTYiLCJwIjoiaiJ9


src/nft/css/sprinkles.css.ts
```diff
@@ -41,6 +41,8 @@ const themeContractValues = {
   opacity: {
     hover: '0.6',
     pressed: '0.4',
+    disabled: '0.5',
+    enabled: '1',
```
cbachmeier, line 45: Do you need to add these to the sprinkles theme? Looks like you use the shared styles from `styled.d.ts` in the component

cbachmeier (changes requested):
Experiencing a bug where hovering over the tooltip causes it to flash

src/nft/css/sprinkles.css.ts
lynnshaoyu: ur right i don't need this, removed. btw also added in the tooltip fixes

cbachmeier (commented):
disabled overlay is still flashing when hovering on tooltip vs card, video in DMs

PR #5250, opened 2022-11-16
fix: Can list and relist to all marketplaces

cbachmeier:
- Abstract away OS listing to BE endpoint
- Allow for relisting of X2Y2 assets
- Support new LR listing strategy
- Fixes bug with incorrectly formatted basis points (creator royalties)

Tested by being able to successfully list and relist to all 3 marketplaces

### DIFF
--- a/src/nft/queries/openSea/PostOpenSeaSellOrder.ts
+++ b/src/nft/queries/openSea/PostOpenSeaSellOrder.ts
@@ -1,3 +1,5 @@
+import ms from 'ms.macro'
+
 export async function PostOpenSeaSellOrder(payload?: Record<string, unknown>): Promise<boolean> {
   const body = payload ? JSON.stringify(payload) : undefined
   const url = `${process.env.REACT_APP_GENIE_V3_API_URL}/postOpenSeaSellOrderWithApiKey`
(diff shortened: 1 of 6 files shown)


src/nft/queries/openSea/PostOpenSeaSellOrder.ts
```diff
@@ -1,77 +1,24 @@
...
-  if (response.ok) {
-    return response
-  }
-
-  let result
-  let errorMessage
+    body,
+    signal: ac.signal,
+  })
+  const timeout = setTimeout(() => ac.abort(), 60 * 1000)
```
JackShort, line 13: non-blocking: we could use `ms` for clarity

src/nft/utils/listNfts.ts
```diff
@@ -238,14 +244,15 @@ export async function signListing(
...
         await signOrderData(provider, order)
         const payload: OrderPayload = {
           order: encodeOrder(order),
           isBundle: false,
           bundleName: '',
           bundleDesc: '',
-          orderIds: [],
-          changePrice: false,
+          orderIds: prevOrderId ? [prevOrderId] : [],
+          changePrice: Boolean(prevOrderId),
```
JackShort, line 255: non-blocking: instead of casting you could `!!`

JackShort (approved):
lgtm

src/nft/utils/listNfts.ts
cbachmeier: Boolean is actually faster as of 2020

PR #5253, opened 2022-11-16
fix: retain permit sig if amount decreases

zzmp:
This brings signature logic to parity with widget.

See WEB-2259


JackShort (approved):
lgtm

PR #5252, opened 2022-11-16
feat: adding sudoswap price calculation

JackShort:
* this adds the appropriate price calculations for sudoswap assets

PR #5257, opened 2022-11-16
feat: adding sudoswap to sweep

JackShort:
* depends on sudoswap pr ( #5252 ) - just adding  sudo to sweep with appropriate price calculations


src/nft/components/collection/Sweep.tsx
```diff
@@ -190,16 +192,31 @@ export const Sweep = ({ contractAddress, minPrice, maxPrice }: SweepProps) => {
...
       if (!asset.susFlag) {
-        const isNFTX = asset.marketplace === Markets.NFTX
-        const poolPrice = calcPoolPrice(asset, isNFTX ? counterNFTX : counterNFT20)
-        if (BigNumber.from(poolPrice).gt(0)) {
-          isNFTX ? counterNFTX++ : counterNFT20++
-          asset.priceInfo.ETHPrice = poolPrice
+        if (asset.marketplace === Markets.Sudoswap) {
+          const poolPrice = calcSudoSwapPrice(
+            asset,
+            sudoSwapAssetsInJointCollections
```
vm, line 208: this may be slow if the list is huge?

src/nft/components/collection/Sweep.tsx
```diff
@@ -411,7 +433,7 @@ const ALL_OTHER_MARKETS = [Markets.Opensea, Markets.X2Y2, Markets.LooksRare]
 
 export function useSweepFetcherParams(
   contractAddress: string,
-  market: Markets.NFTX | Markets.NFT20 | 'others',
+  market: Markets.Sudoswap | Markets.NFTX | Markets.NFT20 | 'others',
```
vm, line 436: can't you use the type you defined for combining these?

src/nft/components/collection/Sweep.tsx
```diff
@@ -423,12 +445,23 @@ export function useSweepFetcherParams(
   return useMemo(() => {
     if (isMarketFiltered) {
       if (market === 'others') {
-        return { contractAddress, traits, markets }
+        const marketsToQuery = markets.filter((market) => !isPooledMarket(market as Markets))
```
vm, line 448: should be `Market` not `Markets` fwiw

vm (approved):
think we could abstract the pooled marketplace stuff more but for now this is good

src/nft/components/collection/Sweep.tsx
vm: ignore, the list is 50 max

src/nft/components/collection/Sweep.tsx
vm: later problem

src/nft/components/collection/Sweep.tsx
vm: later problem

PR #5258, opened 2022-11-16
fix: max depth error

aballerr:
issue

Clicking on search causes maximum depth error.  It's because an object is being passed to the dependency array

Screen Shot 2022-11-16 at 3 56 17 PM

after

Screen Shot 2022-11-16 at 3 57 28 PM

### DIFF
--- a/src/components/NavBar/SearchBarDropdown.tsx
+++ b/src/components/NavBar/SearchBarDropdown.tsx
@@ -204,10 +204,10 @@ export const SearchBarDropdown = ({
     (isNFTPage && (hasVerifiedCollection || !hasVerifiedToken)) ||
     (!isNFTPage && !hasVerifiedToken && hasVerifiedCollection)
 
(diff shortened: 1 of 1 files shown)


src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -204,10 +204,10 @@ export const SearchBarDropdown = ({
     (isNFTPage && (hasVerifiedCollection || !hasVerifiedToken)) ||
     (!isNFTPage && !hasVerifiedToken && hasVerifiedCollection)
 
-  const trace = useTrace({ section: SectionName.NAVBAR_SEARCH })
+  const trace = JSON.stringify(useTrace({ section: SectionName.NAVBAR_SEARCH }))
```
aballerr, line 207: object was getting passed, causing the error

lynnshaoyu: omg thanks for fixing

lynnshaoyu (approved):

PR #5259, opened 2022-11-16
refactor: better theming

grabbou:
This PR refactors our theming a bit, removing the number of files/types you need to modify to add a new value. Instead of having a DefaultTheme interface where all types live, we infer its type from an actual theme. High level goal for this work was to make it easy to a) edit theme, b) reason about its contents.

To avoid circular references, I had to remove all Styled Components-related things from the theme. That's why `flexRowNoWrap` and `flexColumnNoWrap` are now located in a separate file called `styles.ts`, together with animation utilities (pretty much everything that is meant to be reusable).

You will notice that `deprecatedColors.ts` file has a different shape. I didn't refactor it, as it's likely to go away in a follow-up PR.

This work will be divided into separate PRs. Next step is to get rid of deprecated colors and remove `export * from './compoents` from index file.


grabbou:
Feel free to leave some further comments/ideas if you'd structure things in a different way!

src/theme/components/index.tsx
```diff
@@ -376,7 +378,7 @@ export const CopyHelper = forwardRef<CopyHelperRefType, CopyHelperProps>(
       iconSize = 20,
       gap = 12,
       iconPosition = 'left',
-      iconColor,
+      iconstring,
```
vm, line 381: iconString

vm: also this is actually wrong?

grabbou: yes, sorry... I accidentialy replaced `Color` here. Will double check the entire diff for any other issues like this.

grabbou:
Resolved some conflicts in the meantime.

vm (approved):

PR #5264, opened 2022-11-16
feat: simplified duration dropdown

cbachmeier:
Simplified duration dropdown on list page

Also fixes bug where bottom padding in sort dropdown was the wrong color


src/nft/components/profile/list/SetDurationModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+            mini
+            miniPrompt={displayDuration + (displayDuration === duration ? pluralize(amount) : 's')}
+            left={38}
+            top={38}
+          />
+        </DropdownWrapper>
+      </InputWrapper>
+      {errorState !== ErrorState.valid && (
+        <ErrorMessage className={caption}>
+          {' '}
```
grabbou, line 129: What's the reason for this character?

src/nft/components/profile/list/SetDurationModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+          <WarningIcon /> {errorState === ErrorState.overMax ? 'Maximum 6 months' : 'Set duration'}
```
grabbou, line 130: nit: moving to `getMessageForError` + a switch/case for future extendability? (if any)

src/nft/components/profile/list/SetDurationModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+        </ErrorMessage>
+      )}
+    </ModalWrapper>
+  )
+}
+
+const convertDurationToExpiration = (amount: number, duration: Duration) => {
+  const durationFactor =
+    duration === Duration.hour ? 1 : duration === Duration.day ? 24 : duration === Duration.week ? 24 * 7 : 24 * 30
```
grabbou, line 139: For this kind of nested tenary, I always tend to prefer a `switch(true) case "expression"`, but it's just personal preference, so leaving as a nit for you to consider

src/nft/components/profile/list/SetDurationModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+      {
+        displayText: 'Months',
+        onClick: () => selectDuration(Duration.month),
+      },
+    ],
+    []
+  )
+  useEffect(() => {
+    const expiration = convertDurationToExpiration(amount, duration)
+    if (expiration * 1000 - Date.now() < ms`60 seconds`) setErrorState(ErrorState.empty)
```
grabbou, line 95: Question: do we maintain a general guideline as far as if statements? Shall we skip `{}`? If yes, when?

grabbou (approved):

src/nft/components/profile/list/SetDurationModal.tsx
unipadmini: +1

src/nft/components/profile/list/SetDurationModal.tsx
cbachmeier: this was auto-added by lint

src/nft/components/profile/list/SetDurationModal.tsx
cbachmeier: We don't have any plans for future error messages for this dropdown

src/nft/components/profile/list/SetDurationModal.tsx
cbachmeier: No official guidelines, I generally will skip for 1 line, simple executions. If readability is compromised I'll add braces

PR #5272, opened 2022-11-17
fix: rendering of glow effect on mobile safari

grabbou:
Sometimes Safari would render the Banner as if it had overflow set to hidden. This is odd because that does not happen on Chrome. The issue happens when there's a filter (blur) applied to an element and there's a hover state triggered either on that element or any of the elements on top of it. 

Applying `translate3d` seems to do the trick. Decided to link to the StackOverflow issue because this is a strong workaround and there's not many resources on it, beyond that answer. 

While reviewing, check the details on Slack: https://uniswapteam.slack.com/archives/C04822DFATE/p1668686941493839?thread_ts=1668668097.346379&cid=C04822DFATE

My guess is that the glitch doesn't happen when GPU acceleration kicks in for that element, which we achieve by applying `translate3d`.


cbachmeier (approved):
LGTM

PR #5273, opened 2022-11-17
fix: right arrow in carousel not rotated

grabbou:
Unfortunately, Safari doesn't support `transform` of SVG2 spec. More cross-browser approach is to use CSS + styled components.


JackShort (approved):
lgtm

PR #5274, opened 2022-11-17
fix: border radius on safari for carousel

grabbou:
Fixes WEB-2179

Safari doesn't apply border-radius on an outline, resulting in a square border on a rounded element. Workaround is to use `::before` and render an element outside of the container.


JFrankfurt:
Can we just make the rule a border instead of an outline?

JFrankfurt (approved):
approving w/ possible simpler implementation

JFrankfurt:
### default:
image
### hover:
image

src/nft/components/explore/CarouselCard.tsx
```diff
@@ -13,13 +13,32 @@ const CarouselCardContainer = styled.div`
...
+    border-radius: 21px;
+    border: 1px solid ${({ theme }) => theme.backgroundOutline};
+  }
+
+  :hover::before {
+    top: -3px;
+    right: -3px;
+    bottom: -3px;
+    left: -3px;
+    border-radius: 23px;
```
JFrankfurt, line 40: is the 21 -> 23px change on hover intentional?

JFrankfurt (changes requested):
Just noticed there seems to be some kind of odd misalignment going on here that renders the background through the element on the bottom. I think the absolute positioning of the pseudo-element is the culprit

src/nft/components/explore/CarouselCard.tsx
grabbou: Yes, this is needed since the border on hover is 3px wide, which requires more radius to align with the inner element.

grabbou:
> Can we just make the rule a border instead of an outline?

Border will shift the element when it changes. So outline is our only solution, unless we want to use a pseudoelements like in this example. I haven't seen any misalignment to be honest, let me pull and double check.

JFrankfurt:
try setting a transparent border and just changing the color instead of the whole thing (can do it on the wrapper element in your solution so you get the 1px static border on the inner element, if you'd like to preserve the wrapping portion of your current solution)

grabbou:
Why not, that could actually yield similar results but with less negative margins. I don’t really have a strong opinion here, now that I believe the outstanding UI glitch’s been fixed, but I am keen to give it a go!

src/nft/components/explore/CarouselCard.tsx
```diff
@@ -12,14 +12,20 @@ const CarouselCardContainer = styled.div`
...
+  overflow: hidden;
+  height: 100%;
+`
+const CarouselCardBorder = styled.div`
   width: 100%;
+  position: relative;
+  border-radius: 22px;
   cursor: pointer;
-  overflow: hidden;
+  border: 2px solid transparent;
```
JFrankfurt, line 25: I know it didn't have this before, but we should add a transition on border (guessing theme.transition.duration.fast or .medium and theme.transition.timing.inOut will feel best)

src/nft/components/explore/CarouselCard.tsx
```diff
@@ -12,14 +12,20 @@ const CarouselCardContainer = styled.div`
...
 
   :hover {
-    outline: 3px solid ${({ theme }) => theme.backgroundOutline};
+    border: 2px solid ${({ theme }) => theme.backgroundOutline};
     box-shadow: ${({ theme }) => theme.deepShadow};
```
JFrankfurt, line 29: would also be good to add a transition on this, but see: https://tobiasahlin.com/blog/how-to-animate-box-shadow/

src/nft/components/explore/CarouselCard.tsx
grabbou: Okay, let's add fast/inOut and ship it! Not a launch blocker, I guess we can always revisit later

grabbou: Generally i agree that default transition feels a bit rough

grabbou:
Added transitions 🔥

JFrankfurt (approved):

PR #5275, opened 2022-11-17
feat: add glow on all pages

grabbou:
Fixes WEB-1809

This removes linear gradient background on all pages and brings a box-shadow (glow) effect instead. This is done with a special component (or a CSS helper). 

It takes optional `chainId` that you can use to make it react to chain changes, or don't pass it, when you want it to keep its default value.

The intent of this PR was to structure this in a way that has minimal changes (especially to the render tree).

Other notable changes;
- removed default export from AppBody, seems like it's a 1:1 copy of BodyWrapper

For testing, please carefully click around, especially on other pages (such as pool pages or proposals) that are not as visible as token/swap at a first sight.


public/index.html
```diff
@@ -1,120 +1,105 @@
...
       html {
-        font-size: 16px;
-        font-variant: none;
-        font-smooth: always;
-        -webkit-font-smoothing: antialiased;
-        -moz-osx-font-smoothing: grayscale;
-        -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
-      }
-
-      #background-radial-gradient {
```
grabbou, line 79: I removed just this part, but Prettier formatted the document automatically.

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -45,9 +46,9 @@ const NoTokenDisplay = styled.div`
   gap: 8px;
 `
 
-function NoTokensState({ message }: { message: ReactNode }) {
+function NoTokensState({ chainId, message }: { chainId: number; message: ReactNode }) {
   return (
-    <GridContainer>
+    <GridContainer chainId={chainId}>
```
grabbou, line 51: The implementation of `glowEffect` is quite nice presentation of Styled Components capabilities as far as accepting params and ensuring type safety across different components. 

I love how adding a simple `css` rule to your existing styled component automatically updates its prop types and if there's something now required to build up its appearance, you know it

src/theme/components/GlowEffect.tsx
```diff
@@ -0,0 +1,49 @@
...
+    case SupportedChainId.POLYGON:
+    case SupportedChainId.POLYGON_MUMBAI:
+      return theme.networkPolygonShadow
+    case SupportedChainId.CELO:
+    case SupportedChainId.CELO_ALFAJORES:
+      return theme.networkCeloShadow
+    default:
+      return theme.networkDefaultShadow
+  }
+}
```
grabbou, line 23: Now that I've spent some time in other places of the codebase, it could make more sense to make a map of ids to shadows and do something like `map[shadow] || defaultShadow`, which I think is similar conceptually to what we're already doing inside chain utils

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -77,14 +78,18 @@ export function LoadingTokenTable({ rowCount = PAGE_SIZE }: { rowCount?: number
...
   const { tokens, sparklines } = useTopTokens(chainName)
   setRowCount(tokens?.length ?? PAGE_SIZE)
 
+  const chainId = chainBackendNameToId(chainName)
+
   /* loading and error state */
   if (!tokens) {
     return (
       <NoTokensState
+        chainId={chainId}
```
vm, line 92: we may want the wallet's chain id not this one

src/components/Widget/index.tsx
```diff
@@ -136,7 +137,7 @@ export default function Widget({ token, onTokenChange, onReviewSwapClick }: Widg
   }
 
   return (
-    <>
+    <GlowEffect chainId={chainId}>
```
vm, line 140: GlowEffect should handle getting the chainId itself, not need it to be passed in

src/components/swap/styleds.tsx
```diff
@@ -23,7 +24,9 @@ export const PageWrapper = styled.div`
 `
 
 // Mostly copied from `AppBody` but it was getting too hard to maintain backwards compatibility.
-export const SwapWrapper = styled.main<{ margin?: string; maxWidth?: string }>`
+export const SwapWrapper = styled.main`
+  ${glowEffect}
```
vm, line 28: why do we have glowEffect and also a component GlowEffect?

vm (changes requested):
commented

src/components/Tokens/TokenTable/TokenTable.tsx
grabbou: setting this to wallet's chain id would give a wrong glow effect when you change chain from the filters above the table. This corresponds to whatever is currently being displayed by the table, making sure we don't show e.g. Polygons glow for Celo etc.

src/components/Widget/index.tsx
grabbou: Initially I did it this way, but then, I figured that making it more explicit gives us a bit more control on it, e.g:
- TokenTable needs different `chainId` than Wallet
- All NFT pages do not pass `chainId` because it has a network default glow as per design
Making it "intelligent" that is resolving `chainId` internally would mean we need some conditionals, which we avoid with this approach.

src/components/swap/styleds.tsx
grabbou: So, if you already have a Styled component or don't want to wrap your components in yet another `div`, `glowEffect` (plain style) comes into play. This makes it much easier to add a glow effect and avoid introducing any unwanted side effects as a result of wrapping a layout in a `div`. You can't also do `styled(GlowEffect)` in this particular scenario, because we're using `main` and not a `div`. 

I left GlowEffect component as an alternative for components that are either:
- not using Styled Components (see Widget), and we would have to create a component for them anyway
- it is not possible to apply a style to a component because it already utilizes `box-shadow` or some other properties (e.g. many components such as Table use a `1px` box-shadow to show a border)

src/components/Widget/index.tsx
grabbou: Also, there’s no way to automatically get chainId if we’re using style (from css), not a component. And we’re using style in 99% cases because it’s opaque and doesn’t mess with the layout

src/components/swap/styleds.tsx
grabbou: Removed component

src/pages/CreateProposal/index.tsx
```diff
@@ -241,7 +241,7 @@ ${bodyValue}
   return (
     <Trace page={PageName.VOTE_PAGE} shouldLogImpression>
       <PageWrapper>
-        <AppBody {...{ maxWidth: '800px' }}>
+        <BodyWrapper chainId={chainId} maxWidth="800px">
```
vm, line 244: feels like BodyWrapper should handle getting the chainId itself vs passing around so many levels deep. I also am not clear why BodyWrapper is adding the glow vs doing it on the component itself.

vm (approved):

PR #5277, opened 2022-11-17
fix: search bar width sizing

aballerr:
https://uniswaplabs.atlassian.net/browse/WEB-2057

Only change from task was to center

I had Yushin check and she approved of the preview


src/components/NavBar/SearchBarNft.css.ts
```diff
@@ -0,0 +1,242 @@
...
+const DESKTOP_NAVBAR_WIDTH_L = 440
+const DESKTOP_NAVBAR_WIDTH_XL = 540
+const DESKTOP_NAVBAR_WIDTH_XXL = 640
+
+const baseSearchStyle = style([
+  sprinkles({
+    paddingY: '12',
+    width: { sm: 'viewWidth' },
+    borderStyle: 'solid',
+    borderWidth: '2px',
```
aballerr, line 16: Increased border width to 2px

src/components/NavBar/SearchBarNft.css.ts
```diff
@@ -0,0 +1,242 @@
...
+    borderColor: 'backgroundOutline',
+  }),
+  {
+    '@media': {
+      [`screen and (min-width: ${breakpoints.sm}px)`]: {
```
aballerr, line 21: this media query is the same, every other media query is new.  Getting this to be compatible required a bunch of specific sizing.  Given time constraints, I don't have a lot of time to make this perfect, hence a lot of manual screen sizes

src/components/NavBar/SearchBarNft.css.ts
```diff
@@ -0,0 +1,242 @@
...
+        right: `-${DESKTOP_NAVBAR_WIDTH_XL / 2}px`,
+      },
+      [`screen and (min-width: ${1480}px)`]: {
+        right: `-${DESKTOP_NAVBAR_WIDTH_XXL / 2}px`,
+      },
+    },
+  },
+])
+
+export const searchBar = style([
```
aballerr, line 69: everything from here and below is the same.  I want to keep it simple, so I copied and pasted the file.  See above comment about removing flag, this should simplify

src/components/NavBar/SearchBarNft.css.ts
```diff
@@ -0,0 +1,242 @@
...
+  sprinkles({
+    right: '0',
+    top: '0',
+    zIndex: '3',
+    display: 'inline-block',
+  }),
+  {
+    '@media': {
+      [`screen and (min-width: ${breakpoints.lg}px)`]: {
+        right: `-${DESKTOP_NAVBAR_WIDTH / 2}px`,
```
aballerr, line 50: I removed the navbar icon from the equation.  The addition of the "nft" option to the left affected screen placement

src/components/NavBar/SearchBar.tsx
```diff
@@ -33,6 +34,9 @@ export const SearchBar = () => {
   const phase1Flag = useNftFlag()
   const isMobile = useIsMobile()
   const isTablet = useIsTablet()
+  const isPhase1 = phase1Flag === NftVariant.Enabled
+  // when removing this flag, copy content of SearchBarNft.css into SearchBar.css and use the styles import only
+  const finalStyles = isPhase1 ? nftStyles : styles
```
cbachmeier, line 39: a lot cleaner approach would be to create individual styles for the differences such as the borderWidth and the breakpoint widths and use the phase1 conditional to switch between them instead of switching the entire file and duplicating a lot of code

src/components/NavBar/SearchBarNft.css.ts
cbachmeier: I believe it's from the increased borderwidth but the searchbar is not longer centered vertically, screenshot in DMs

cbachmeier (changes requested):

src/components/NavBar/SearchBar.css.ts
```diff
@@ -40,6 +72,35 @@ export const searchBarContainer = style([
...
+    right: '0',
+    top: '0',
+    zIndex: '3',
+    display: 'inline-block',
+  }),
+  {
+    '@media': {
+      [`screen and (min-width: ${breakpoints.lg}px)`]: {
+        right: `-${DESKTOP_NAVBAR_WIDTH / 2}px`,
+        top: '-2px',
```
cbachmeier, line 86: `top: '-6px'` gets it to line up

src/components/NavBar/SearchBar.css.ts
```diff
@@ -23,6 +26,35 @@ const baseSearchStyle = style([
...
+    width: { sm: 'viewWidth' },
+    borderStyle: 'solid',
+    borderWidth: '2px',
+    borderColor: 'backgroundOutline',
+  }),
+  {
+    '@media': {
+      [`screen and (min-width: ${breakpoints.sm}px)`]: {
+        width: `${DESKTOP_NAVBAR_WIDTH}px`,
+      },
```
cbachmeier, line 41: These styles are just duplicates of `baseSearchStyle` so you can replace this section with importing that style

cbachmeier: Alternatively you could delete that section and then in `nftSearchBar` just import `baseSearchStyle` and these added breakpoints

cbachmeier (approved):
LGTM

PR #5283, opened 2022-11-17
fix: tx screen tweak

cbachmeier:
Add some horizontal padding to fix bug with shadow being cutoff
Align items center


JackShort (approved):
perfect

PR #5285, opened 2022-11-17
style: adding header to top of profile page

JackShort:
https://uniswaplabs.atlassian.net/browse/WEB-2285

* this looks like more than it is but the diff is mostly indentation lol


src/nft/components/profile/view/ProfilePage.tsx
```diff
@@ -33,6 +33,23 @@ const ProfilePageColumn = styled(Column)`
   ${ScreenBreakpointsPaddings}
 `
 
+const ProfileHeader = styled.div`
+  margin-right: 20px;
+  margin-left: 20px;
+  font-size: 28px;
+  font-weight: 500;
+  line-height: 38px;
```
cbachmeier, line 41: nit: do we have a common typography we could use here

cbachmeier (approved):
LGTM minor nit

src/nft/components/profile/view/ProfilePage.tsx
JackShort: unfortunately no

grabbou: doesn't theme/components/text work here?

PR #5287, opened 2022-11-17
fix: collection screen for mobile

JFrankfurt:
https://uniswaplabs.atlassian.net/browse/WEB-2261


src/nft/pages/collection/index.tsx
```diff
@@ -27,7 +27,8 @@ import { ThemedText } from 'theme'
 
 const FILTER_WIDTH = 332
 const BAG_WIDTH = 324
-export const COLLECTION_BANNER_HEIGHT = 288
+const BannerWrapper = styled(Box)``
```
grabbou, line 30: Why not use Box directly? Seems a bit redundant to have an empty wrapper like this?

grabbou: Also, just nitpicking, but I think it would be great to group constants together (including the one on line 31) and separate from this component

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -66,9 +66,19 @@ const rarityStatusCache = new Map<string, boolean>()
...
+  margin: 0 16px;
+  @media screen and (min-width: ${({ theme }) => theme.breakpoint.sm}px) {
+    margin: 0 20px;
+  }
+  @media screen and (min-width: ${({ theme }) => theme.breakpoint.md}px) {
+    margin: 0 26px;
+  }
+  @media screen and (min-width: ${({ theme }) => theme.breakpoint.lg}px) {
+    margin: 0 48px;
+  }
```
grabbou, line 81: nit: I think the general pattern is to go from default to smaller screens with `max-width`. Looking at the codebase, we have 2x more occurrences of `max-width`!

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -93,7 +103,7 @@ export const SortDropdownContainer = styled.div<{ isFiltersExpanded: boolean }>`
 
 const EmptyCollectionWrapper = styled.div`
   display: block;
-  textalign: center;
+  text-align: center;
```
grabbou, line 106: 👀

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -111,6 +121,19 @@ const ClearAllButton = styled.button`
...
+  padding: 0 16px;
+  @media screen and (min-width: ${({ theme }) => theme.breakpoint.sm}px) {
+    padding: 0 20px;
+  }
+  @media screen and (min-width: ${({ theme }) => theme.breakpoint.md}px) {
+    padding: 0 26px;
+  }
+  @media screen and (min-width: ${({ theme }) => theme.breakpoint.lg}px) {
+    padding: 0 48px;
+  }
```
grabbou, line 134: Can you move that out to a reusable `css` and share with `ActionsContainer`? Looks the same!

src/nft/components/collection/CollectionNfts.tsx
JFrankfurt: yeah that spooked me a little

need to get that dangling vscode plugin suggestions pr merged

src/nft/components/collection/CollectionNfts.tsx
JFrankfurt: done!

src/nft/components/collection/CollectionNfts.tsx
vm: this is not particularly consistent enough to require this yet. i think once we decide we can add a lint rule

src/nft/components/collection/CollectionNfts.tsx
vm: yikes. eslint plz

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -564,35 +586,37 @@ export const CollectionNfts = ({ contractAddress, collectionStats, rarityVerifie
...
-          </Center>
-        ) : (
-          <CollectionNftsLoading />
-        )}
-      </InfiniteScroll>
+      <InfiniteScrollWrapper>
+        <InfiniteScroll
+          next={handleNextPageLoad}
+          hasMore={hasNext}
+          loader={Boolean(hasNext && hasNfts) && <LoadingAssets />}
```
vm, line 593: how could this ever not be a boolean?

JFrankfurt: couldn't--I just hate the pattern of unwrapped compound bool statements in templates because SOMETIMES IT BITES

JFrankfurt: I don't let myself get comfortable with unwrapped ones so I don't accidentally ship a bug

PR #5288, opened 2022-11-17
style: correct collection loading

JackShort:
* fixes incorrect padding on loading animations

### DIFF
--- a/src/nft/components/collection/CollectionPageSkeleton.tsx
+++ b/src/nft/components/collection/CollectionPageSkeleton.tsx
@@ -3,10 +3,20 @@ import { Column, Row } from 'nft/components/Flex'
 import { useIsMobile } from 'nft/hooks'
 import { CollectionBannerLoading } from 'nft/pages/collection'
 import { COLLECTION_BANNER_HEIGHT } from 'nft/pages/collection'
+import * as styles from 'nft/pages/collection/index.css'
+import styled from 'styled-components/macro'
 
 import { ActivitySwitcherLoading } from './ActivitySwitcher'
 import { CollectionNftsAndMenuLoading } from './CollectionNfts'
 import { CollectionStatsLoading } from './CollectionStats'
+
+const CollectionDescriptionSection = styled(Column)`
+  ${styles.ScreenBreakpointsPaddings}
+`
+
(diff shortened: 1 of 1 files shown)


src/nft/components/collection/CollectionPageSkeleton.tsx
```diff
@@ -3,27 +3,37 @@ import { Column, Row } from 'nft/components/Flex'
...
 import { CollectionBannerLoading } from 'nft/pages/collection'
 import { COLLECTION_BANNER_HEIGHT } from 'nft/pages/collection'
+import * as styles from 'nft/pages/collection/index.css'
+import styled from 'styled-components/macro'
 
 import { ActivitySwitcherLoading } from './ActivitySwitcher'
 import { CollectionNftsAndMenuLoading } from './CollectionNfts'
 import { CollectionStatsLoading } from './CollectionStats'
 
+const CollectionDescriptionSection = styled(Column)`
```
grabbou, line 13: nit / idea: Maybe we could try using Column/Row from `src/components`? I just found them, looks like they're direct equivalents of the Column/Box from Sprinkles that we're eventually going to get rid of. Might be a good practice to start replacing them as we refactor things, just like we move to Styled Components.

src/nft/components/collection/CollectionPageSkeleton.tsx
```diff
@@ -3,27 +3,37 @@ import { Column, Row } from 'nft/components/Flex'
 import { useIsMobile } from 'nft/hooks'
 import { CollectionBannerLoading } from 'nft/pages/collection'
 import { COLLECTION_BANNER_HEIGHT } from 'nft/pages/collection'
+import * as styles from 'nft/pages/collection/index.css'
```
grabbou, line 6: nit: I think it's better to use named imports instead of `* as styles` for better tooling support and potential tree shaking in the future. Easier to move things around too.

src/nft/components/collection/CollectionPageSkeleton.tsx
JackShort: seems to work pretty well

vm (approved):

PR #5289, opened 2022-11-17
fix: adding back listing icon to profile page

JackShort:
https://uniswaplabs.atlassian.net/browse/WEB-2228


src/components/NavBar/Bag.tsx
```diff
@@ -34,14 +37,18 @@ export const Bag = () => {
...
-    setBagQuantity(itemsInBag.length)
-  }, [itemsInBag])
+    setBagQuantity(isProfilePage ? sellAssets.length : itemsInBag.length)
+  }, [isProfilePage, itemsInBag, sellAssets])
 
   const bagHasItems = bagQuantity > 0
 
   return (
     <NavIcon isActive={bagExpanded} onClick={handleIconClick}>
-      <BagIcon viewBox="0 0 24 24" width={24} height={24} strokeWidth="2px" />
```
grabbou, line 44: I think a slightly better abstraction could be to create a reusable private `Bag` component that takes `icon` prop and `quantity` (number`) and then, based on it, export two different components:
- Bag
- NftBag
(names tbd)
and render them conditionally based on the location.

grabbou (commented):
Overall looks good, left some comments! Hope that helps.

src/components/NavBar/Bag.tsx
```diff
@@ -34,14 +37,18 @@ export const Bag = () => {
   }, [bagExpanded, setBagExpanded])
 
   useEffect(() => {
-    setBagQuantity(itemsInBag.length)
-  }, [itemsInBag])
+    setBagQuantity(isProfilePage ? sellAssets.length : itemsInBag.length)
+  }, [isProfilePage, itemsInBag, sellAssets])
```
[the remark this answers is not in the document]
JackShort: based

src/components/NavBar/Bag.tsx
JackShort: def future improvement

cbachmeier (approved):
LGTM

src/components/NavBar/Bag.tsx
grabbou: Is there any way for us to track these improvements? Could we mark it as a todo in the code? Just afraid of this getting out of the radar (which is very likely, because we're moving fast)!

PR #5291, opened 2022-11-17
fix: working profile scroll on mobile

cbachmeier:
- fix bug where profile assets couldn't scroll
- removes deprecated mobile styles


JackShort (approved):
lgtm

PR #5292, opened 2022-11-17
style: moving mobile sell bag above tab bar

JackShort:
https://uniswaplabs.atlassian.net/browse/WEB-2234


src/nft/components/profile/view/ProfilePage.tsx
```diff
@@ -179,9 +181,10 @@ export const ProfilePage = () => {
             onClick={() => setSellPageState(ProfilePageStateType.LISTING)}
```
cbachmeier, line 181: this functionality should be switched to toggleBag instead of go straight to list

cbachmeier (changes requested):

cbachmeier (approved):
LGTM

PR #5299, opened 2022-11-18
fix: Can't scroll behind bag

cbachmeier:
- Fixes the bug on collection and profile page where you could scroll assets underneath
- Stops the cards from moving around underneath the bag causing them to snap in place when the bag was closed
- Fixes the laggy animation when filters were opened/closed on profile

### DIFF
--- a/src/nft/components/profile/view/FilterSidebar.tsx
+++ b/src/nft/components/profile/view/FilterSidebar.tsx
@@ -11,6 +11,7 @@ import { WalletCollection } from 'nft/types'
 import { Dispatch, FormEvent, SetStateAction, useCallback, useEffect, useReducer, useState } from 'react'
 import { easings, useSpring } from 'react-spring'
 import styled from 'styled-components/macro'
+import { TRANSITION_DURATIONS } from 'theme/styles'
 
 import * as styles from './ProfilePage.css'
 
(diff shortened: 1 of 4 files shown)


src/nft/components/profile/view/FilterSidebar.tsx
```diff
@@ -22,14 +22,18 @@ const ItemsContainer = styled.div`
...
+  const isBagExpanded = useBag((state) => state.bagExpanded)
   const walletCollections = useWalletCollections((state) => state.walletCollections)
 
   const [isFiltersExpanded, setFiltersExpanded] = useFiltersExpanded()
   const isMobile = useIsMobile()
 
   const { sidebarX } = useSpring({
     sidebarX: isFiltersExpanded ? 0 : -360,
+    config: {
+      duration: 250,
```
grabbou, line 34: does it make sense to use durations from theme?

cbachmeier: this will require some refactoring as the durations give the full ms whereas the config requires a number. I've come up with an intermediary solution that uses an enum for duration values

src/nft/components/profile/view/ProfilePage.tsx
```diff
@@ -107,6 +112,7 @@ export const ProfilePage = () => {
               <Column width="full">
                 <AnimatedBox
                   flexShrink="0"
+                  position={isMobile && (isFiltersExpanded || isBagExpanded) ? 'fixed' : 'static'}
```
JackShort, line 115: non-blocking: this does work but because of the check `!isMobile || !isFiltersExpanded` above I think you can actually just do `isMoblie && isBagExpanded` because it will never be `isMobile && isFiltersExpanded`

JackShort (approved):
noice

PR #5300, opened 2022-11-18
fix: make carousel responsive

JFrankfurt:


src/nft/components/explore/Carousel.tsx
```diff
@@ -37,36 +32,32 @@ const CarouselCardContainer = styled.div`
...
-  padding-right: 8px;
+  color: ${({ theme }) => theme.textPrimary};
   cursor: pointer;
+  display: none;
+  padding-left: 4px 8px 32px;
   user-select: none;
-
-  color: ${({ theme }) => theme.textPrimary};
-
+  @media only screen and (min-width: ${({ theme }) => `${theme.breakpoint.sm}px`}) {
```
grabbou, line 55: Any particular reason for this change? I think conceptually the original version was a bit cleaner with `display: flex` being default and `display: none` override for small devices (only).

src/nft/components/explore/Carousel.tsx
```diff
@@ -37,36 +32,32 @@ const CarouselCardContainer = styled.div`
...
-  ${carouselItemStyle}
-
-  display: flex;
   align-items: center;
-  padding-left: 8px;
-  padding-right: 8px;
+  color: ${({ theme }) => theme.textPrimary};
   cursor: pointer;
+  display: none;
+  padding-left: 4px 8px 32px;
```
grabbou, line 53: Did you mean `padding` here? Also, previously the top/bottom paddings were shared between an icon and a card, which I think is quite useful as they depend on each other.

If this is `padding`, this is pretty much equivalent to previous:
```
padding-left: 8px;
padding-right: 8px;
```
+ shared `carouselItemStyle` adding 4px/32px top/bottom paddings. 

So yeah, I think my conclusion is I'd love for these paddings to stay shared between an item and a card so it is easier to manage them in the future.

src/nft/components/explore/Carousel.tsx
```diff
@@ -37,36 +32,32 @@ const CarouselCardContainer = styled.div`
...
-  padding: 3px 32px 32px 32px;
-
+  padding: 4px 12px 32px;
   position: absolute;
   will-change: transform;
+
+  @media screen and (min-width: ${({ theme }) => theme.breakpoint.lg}px) {
+    padding: 4px 32px 32px;
+  }
 `
```
grabbou, line 45: Idea:

Bring back `carouselItemStyle`, set `padding-left` and `padding-right` to `12px` and override them in a media query (not the whole `padding`, just left/right).

src/nft/components/explore/Carousel.tsx
```diff
@@ -8,18 +8,13 @@ import styled, { css } from 'styled-components/macro'
 const MAX_CARD_WIDTH = 530
 
 const carouselHeightStyle = css`
-  height: 315px;
+  height: 296px;
 
-  @media only screen and (max-width: ${({ theme }) => `${theme.breakpoint.md}px`}) {
-    height: 296px;
+  @media only screen and (min-width: ${({ theme }) => `${theme.breakpoint.md}px`}) {
+    height: 316px;
```
grabbou, line 14: Any particular reason for this revert? I think other media queries within the application go from highest screen through `max-width`, so I'd love to keep this the same, unless there's a good reason for doing otherwise!

grabbou (changes requested):
Overall I like this PR, a few subtle changes that make the appearance much better! However, because this component is highly complex and contains a lot of interconnected CSS properties, I would love if we could reduce the diff to absolute minimum and keep as much properties shared as possible!

src/nft/components/explore/Carousel.tsx
JFrankfurt: We currently have a mix of both, but mobile defaults with increasing over-rides is the preferred method. (Hasn't been enforced v well.)

src/nft/components/explore/Carousel.tsx
JFrankfurt: carouselItemStyle was just the padding rules, and one of the components that used it overwrote them anyway. I don't think the re-use benefit is meaningful in this case

src/nft/components/explore/Carousel.tsx
JFrankfurt: yes, ty

src/nft/components/explore/Carousel.tsx
JFrankfurt: following the mobile-first media query pattern w/ min-width overrides

src/nft/components/explore/Carousel.tsx
JFrankfurt: see: https://github.com/Uniswap/interface/pull/5300#discussion_r1026588984  regarding re-use in this case

src/nft/components/explore/Carousel.tsx
vm: as stated elsewhere i think we can discuss min vs max advantages elsewhere. really depends if it matters and if we build for mobile first.

src/nft/components/explore/Carousel.tsx
vm: i don't think it matters but i would suggest not switching the pattern for existing components as it makes prs harder to read. at least until we decide on a path forwards as a team.

src/nft/components/explore/CarouselCard.tsx
```diff
@@ -45,7 +48,7 @@ const CarouselCardBorder = styled.div`
     box-shadow: ${({ theme }) => theme.deepShadow};
     transition-property: opacity;
     transition-duration: ${({ theme }) => theme.transition.duration.fast};
-    transition-timing: ${({ theme }) => theme.transition.timing.inOut};
+    transition-timing-function: ${({ theme }) => theme.transition.timing.inOut};
```
vm, line 51: this wasn't working?

vm (approved):

src/nft/components/explore/Carousel.tsx
vm: i do think abstracting this might make it harder to follow

src/nft/components/explore/CarouselCard.tsx
JFrankfurt: correct

grabbou: when fast is too fast

PR #5304, opened 2022-11-18
fix: updating to have proper width

aballerr:
search bar is supposed to have 2px width

### DIFF
--- a/src/components/NavBar/SearchBar.css.ts
+++ b/src/components/NavBar/SearchBar.css.ts
@@ -29,6 +29,7 @@ const baseSearchStyle = style([
 const baseSearchNftStyle = style([
   baseSearchStyle,
   {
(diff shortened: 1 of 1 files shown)


vm (approved):

src/components/NavBar/SearchBar.css.ts
```diff
@@ -29,6 +29,7 @@ const baseSearchStyle = style([
 const baseSearchNftStyle = style([
   baseSearchStyle,
   {
+    borderWidth: '2px',
```
vm, line 32: why is this only for the nft one?

aballerr: I don't want to update main right now.  This is all behind phase 1 flag.  They should both use the same search once we go live

PR #5305, opened 2022-11-18
fix: erc 1155 tooltip behavior on mobile

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-2328?atlOrigin=eyJpIjoiODk5YzU5ODhiNjRmNGQxYmI4MDYxMjRlYmVhMzg3ZTUiLCJwIjoiaiJ9

### DIFF
--- a/src/components/Tooltip/index.tsx
+++ b/src/components/Tooltip/index.tsx
@@ -1,5 +1,5 @@
 import { transparentize } from 'polished'
-import { ReactNode, useCallback, useState } from 'react'
+import { ReactNode, useCallback, useEffect, useState } from 'react'
 import styled from 'styled-components/macro'
 
 import Popover, { PopoverProps } from '../Popover'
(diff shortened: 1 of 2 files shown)


src/components/Tooltip/index.tsx
```diff
@@ -53,10 +54,24 @@ function TooltipContent({ content, wrap = false, ...rest }: TooltipContentProps)
...
-export function MouseoverTooltip({ text, disableHover, children, ...rest }: Omit<TooltipProps, 'show'>) {
+export function MouseoverTooltip({ text, disableHover, children, timeout, ...rest }: Omit<TooltipProps, 'show'>) {
   const [show, setShow] = useState(false)
   const open = useCallback(() => setShow(true), [setShow])
   const close = useCallback(() => setShow(false), [setShow])
+
+  useEffect(() => {
+    if (show && timeout) {
+      const tooltipTimer = setTimeout(() => {
+        setShow(false)
```
vm, line 65: don't you only want to hide if their mouse left?

lynnshaoyu: no on mobile the behavior is different, the tooltip appears when user taps, and it never disappears unless you retap. So we want it to automatically time out after 2 seconds.

vm (approved):

PR #5309, opened 2022-11-18
fix: glitch on nft collection + small bg update on dark mode

grabbou:
Fix WEB-2326

Moving to Styled components so we can actually get proper background color off the theme.

PR #5312, opened 2022-11-18
style: light mode background style for floating sell bag

JackShort:
https://uniswaplabs.atlassian.net/browse/WEB-2327

### DIFF
--- a/src/nft/components/profile/view/ProfilePage.tsx
+++ b/src/nft/components/profile/view/ProfilePage.tsx
@@ -202,10 +202,11 @@ export const ProfilePage = () => {
           borderRadius="12"
           paddingX="16"
           paddingY="12"
+          background="backgroundModule"
           borderStyle="solid"
           borderColor="backgroundOutline"
           borderWidth="1px"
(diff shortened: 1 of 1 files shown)


src/nft/components/profile/view/ProfilePage.tsx
```diff
@@ -202,10 +202,11 @@ export const ProfilePage = () => {
           borderRadius="12"
           paddingX="16"
           paddingY="12"
+          background="backgroundModule"
```
grabbou, line 205: This will be off, Styled Components use `background` (not `backgroundModule`) for setting a background in dark/light mode. Unless this is rendered in some other container, there will be a difference in shade. 

Ideally, the easiest would be to rewrite this component to Styled Components and reference proper theme variable from there! 

Check my PR for some direction on how this can be done: #5309

grabbou: ok!

Just wanted to comment that we're shipping a change to master that changes background color. On your branch, background and backgroundModule are the same (gray800). But soon, background will be gray850.

JackShort: ok cool - good to know

vm (approved):

PR #5314, opened 2022-11-18
feat: Listing marketplace dropdown

cbachmeier:
- Creates a marketplace dropdown on listing page
- Adds hover state to the duration dropdown
- Ignore the page layout, this will be fixed in a subsequent PR
- The modal color blends into the background with the latest update, this fix will be on the future background update


src/nft/components/profile/list/SelectMarketplacesDropdown.tsx
```diff
@@ -0,0 +1,172 @@
...
+  padding: 16px 0px;
+  background-color: ${({ theme }) => theme.backgroundModule};
+  display: ${({ isOpen }) => (isOpen ? 'flex' : 'none')};
+  position: absolute;
+  top: 52px;
+  width: 220px;
+  border-radius: 12px;
+  gap: 12px;
+  z-index: ${Z_INDEX.modalBackdrop};
+`
```
JFrankfurt, line 68: I'm a little confused about the intended styles here, especially considering the other two dropdowns.

JFrankfurt: whichever style we're going with, I think the styles should be pulled out so they can be shared/consistent

JFrankfurt (approved):

src/nft/components/profile/list/SelectMarketplacesDropdown.tsx
```diff
@@ -0,0 +1,172 @@
...
+  background: ${({ theme }) => theme.backgroundModule};
+  cursor: pointer;
+  &:hover {
+    background-color: ${({ theme }) => theme.backgroundInteractive};
+  }
+`
+
+const HeaderButtonContentWrapper = styled.div`
+  display: flex;
+  flex-direction: row;
```
vm, line 27: nit: this is the default

src/nft/components/profile/list/SelectMarketplacesDropdown.tsx
```diff
@@ -0,0 +1,172 @@
...
+`
+
+const MarketIcon = styled.img<{ index: number; totalSelected: number }>`
+  height: 20px;
+  width: 20px;
+  margin-right: 8px;
+  border: 1px solid;
```
vm, line 34: nit: can just put the color here

src/nft/components/profile/list/SelectMarketplacesDropdown.tsx
```diff
@@ -0,0 +1,172 @@
...
+  display: ${({ isOpen }) => (isOpen ? 'flex' : 'none')};
+  position: absolute;
+  top: 52px;
+  width: 220px;
+  border-radius: 12px;
+  gap: 12px;
+  z-index: ${Z_INDEX.modalBackdrop};
+`
+
+export const SelectMarketplacesDropdown = ({
```
vm, line 70: would be helpful to put this at the bottom of the component since it's what is exported

src/nft/components/profile/list/SelectMarketplacesDropdown.tsx
vm: which other dropdowns?

vm (approved):
I'm good w this but would check what Jordan's comment around standardizing the dropdown styles is

src/nft/components/profile/list/SelectMarketplacesDropdown.tsx
cbachmeier: the days and set price by are supposed to be similar whereas the marketplace one is unique

this is an interim as design is trying to standardize dropdowns across the app

src/nft/components/profile/list/SelectMarketplacesDropdown.tsx
cbachmeier: I personally prefer having the conditional as atomic as possible

vm: @cbachmeier there is no conditional in this case?

PR #5316, opened 2022-11-18
feat: add glow effect

grabbou:
Fixes WEB-1809

This removes linear gradient background on all pages and brings a box-shadow (glow) effect instead. This is done with a special component (or a CSS helper).

It takes optional chainId that you can use to make it react to chain changes, or don't pass it, when you want it to keep its default value.

The intent of this PR was to structure this in a way that has minimal changes (especially to the render tree).

Other notable changes;

removed default export from AppBody, seems like it's a 1:1 copy of BodyWrapper
For testing, please carefully click around, especially on other pages (such as pool pages or proposals) that are not as visible as token/swap at a first sight.


grabbou:
Unfortunately, I am going to close this one. I don't think that change is every shipping, in case it is, one can just use my work and make it happen.

PR #5317, opened 2022-11-18
feat: various improvements to banner

grabbou:
Fixes WEB-2317
Fixes WEB-2183
Fixes WEB-2258
Fixes WEB-2177
Fixes WEB-2257
Fixes broken blur on Safari (regression)
Make card wider on xl screens for better space utilization (600 instead of 530, checked with the design team)

Video on Slack presenting the overall behavior.

Please verify whether you're happy with the speed of the carousel (animations) and if not, adjust friction accordingly. There was a change due to the different positioning of off-screen card items. 

Known issues (present in main, but not in scope for this PR)
- There's an issue with a box-shadow on hover on screens between medium to large. A temporary quick fix is to add padding `32px`, long-term fix is to make the shadow smaller. Let's not consider that as a blocker, as the list of fixes is bigger. We will add that one for later.
- We need scenario for handling titles when they don't fit

^ both screenshots are in the Slack thread for reference


src/nft/components/explore/Carousel.tsx
```diff
@@ -38,6 +34,7 @@ const CarouselItemCard = styled(a.div)`
   padding: 4px 12px 32px;
   position: absolute;
   will-change: transform;
+  width: calc(100%);
```
grabbou, line 37: So, this is interesting. `calc(100%)` evaluates to a width of the nearest relative parent. Thanks to that, we don't have to set card width manually later in the `render` method. As a result, we can remove listeners to window dimensions as well as setting a ref on a container.

src/nft/components/explore/Carousel.tsx
```diff
@@ -65,53 +62,46 @@ interface CarouselProps {
...
-  const { width } = useWindowSize()
-  const carouselCardContainerRef = useRef<HTMLDivElement>(null)
-  const [cardWidth, setCardWidth] = useState(MAX_CARD_WIDTH)
-
-  useEffect(() => {
-    if (carouselCardContainerRef.current) {
-      setCardWidth(Math.min(carouselCardContainerRef.current.offsetWidth, MAX_CARD_WIDTH))
-    }
-  }, [width])
+const MAX_CARD_WIDTH = 800
```
grabbou, line 68: One comment: the difference will be transparent but meaningful from the implementation perspective. In fact, I believe this variable could be renamed (ideas anyone?)

Now, the value of this constant will be used to calculate the offset by which items in the card should be positioned on the x-axis. 

It would seem that having a correct and accurate container width is necessary for carousel to work, but in fact - it is not.

This changes the layout from:
-[item]-[item]-[item]-

to
<-800px->
-[item]----[item]----[item]

this has no impact on the way the carousel works because we're always animating the `x` to 0. The only actual difference is that the animation may feel slightly faster since there's more travel for the card to get into `x = 0` state, but we can always mimick that with slower speed.

Either way, I think this is a great improvement that greatly simplifies the logic here. All possible because we no longer need to manually set `width`, but can use `calc(100%)`.

grabbou: Either way, I think this is a great improvement that greatly simplifies the logic here. However, if the change in speed is not desired/unacceptable, we can bring back previous behavior, so I am marking this as an idea.

grabbou:
There's an issue with a box-shadow on hover on screens between medium to large, temporary quick fix is to add padding `32px`, long-term fix is to make the shadow smaller. Let's not consider that as a blocker, as the list of fixes is bigger. We will add that one for later.

unipadmini:
Fixes WEB-2317 - verified ✅
Fixes WEB-2183 - almost verified. inconsistent when I switch dev settings from desktop->mobile. not sure if that's just dev settings though.
Fixes WEB-2258 - verified ✅
Fixes WEB-2177 - verified ✅
Fixes WEB-2257 - verified ✅

Would recommend sharing this branch with design to see if this looks good to them!

src/nft/components/explore/CarouselCard.tsx
```diff
@@ -158,44 +143,42 @@ const LoadingCollectionImage = styled.div`
...
+
+const TableElement = styled.div`
+  display: flex;
+  align-items: center;
+  gap: 6px;
+`
 
-  @media only screen and (min-width: ${({ theme }) => `${theme.breakpoint.lg}px`}) {
-    row-gap: 16px;
+const FirstColumnTextWrapper = styled.div`
```
unipadmini, line 156: clever!

unipadmini (commented):
Code looks great, and styling looks great, but I wasn't able to consistently get the icons only view on mobile when I tried viewing it through chrome inspect. Not sure if you were able to test it and get consistent results on mobile.

unipadmini (approved):
Approved given you've tested the mobile case and seen the correct behavior (icons) consistently.

src/nft/components/explore/Carousel.tsx
cbachmeier: instead of calc render wouldn't just `100%` use the full width of the parent element?

src/nft/components/explore/Carousel.tsx
```diff
@@ -22,38 +11,44 @@ const CarouselContainer = styled.div`
...
 
 const CarouselItemIcon = styled.div`
-  ${carouselHeightStyle}
   align-items: center;
-  color: ${({ theme }) => theme.textPrimary};
+  color: ${({ theme }) => theme.accentAction};
   cursor: pointer;
   display: none;
   user-select: none;
+  height: calc(100%);
```
cbachmeier, line 45: do you need `calc(100%)` or can you just use `100%`?

src/nft/components/explore/Carousel.tsx
cbachmeier: I honestly felt like it was too slow with the new friction speed but if you could run it by Bryan, Cal, or Yushin on design to verify I'm fine with current behaviour

grabbou:
CC: @callil to check on the animation speed!

grabbou:
> Code looks great, and styling looks great, but I wasn't able to consistently get the icons only view on mobile when I tried viewing it through chrome inspect. Not sure if you were able to test it and get consistent results on mobile.

@unipadmini, the icons "only" view is on anything above small breakpoint and under large breakpoint. It is not specifically pinned to a mobile. Hope that helps!

grabbou:
Okay, I got confirmation from @callil via DM that he likes the animation, so I am proceeding with merging it!

PR #5319, opened 2022-11-18
fix: include user-added tokens in selector

zzmp:
- Retools the `useXTokens` hooks to only include userAdded tokens for `useAllTokens`
- Instantiates userAdded tokens as `UserAddedToken` instances
- Filters out `UserAddedToken`s from the selector if they have no balance


cartcrom (approved):
LGTM

PR #5322, opened 2022-11-18
fix: no price adjustment popups for pooled assets

JackShort:
* pooled assets always triggered a price adjustment to popup this fixes that

### DIFF
--- a/src/nft/utils/txRoute/combineItemsWithTxRoute.ts
+++ b/src/nft/utils/txRoute/combineItemsWithTxRoute.ts
@@ -31,6 +31,25 @@ const isAveragePriceOfPooledAssets = (
   return !isPriceDiff(calcAvgGroupPoolPrice(asset, numberOfAssetsInPool), expectedPrice)
 }
 
+const isAveragedPrice = (
+  item: UpdatedGenieAsset,
+  items: UpdatedGenieAsset[],
+  route: RoutingItem,
+  txRoute?: RoutingItem[]
+): boolean => {
+  if (!(route && 'priceInfo' in route.assetOut)) return false
+
+  return (
+    !!item.marketplace &&
+    isPooledMarket(item.marketplace) &&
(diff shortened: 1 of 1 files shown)


src/nft/utils/txRoute/combineItemsWithTxRoute.ts
```diff
@@ -36,14 +70,24 @@ export const combineBuyItemsWithTxRoute = (
...
+    if (route && 'priceInfo' in route.assetOut) {
+      if (isPriceDiff(newPriceInfo.basePrice, route.assetOut.priceInfo.basePrice)) {
+        if (
+          !(
+            item.marketplace &&
+            isPooledMarket(item.marketplace) &&
+            isAveragePriceOfPooledAssets(
+              item,
+              items.filter((routeItem) => itemInRouteAndSamePool(item, routeItem, txRoute)).length,
+              route.assetOut.priceInfo.basePrice
```
cbachmeier, line 82: can we turn this into a callback/memo, something like `isAveragedPrice` to help clean up this conditional?

cbachmeier (approved):
Works well from my testing, just want to clean up the massive if block to make more readable

src/nft/utils/txRoute/combineItemsWithTxRoute.ts
```diff
@@ -36,14 +70,24 @@ export const combineBuyItemsWithTxRoute = (
...
     // if the price changed
-    if (
-      route &&
-      'priceInfo' in route.assetOut &&
-      isPriceDiff(newPriceInfo.basePrice, route.assetOut.priceInfo.basePrice)
-    ) {
-      return {
-        ...item,
-        updatedPriceInfo: route.assetOut.priceInfo,
+    if (route && 'priceInfo' in route.assetOut) {
```
grabbou, line 73: for this one, I think we can do some early returns, should help reduce the indentation @cbachmeier

PR #5325, opened 2022-11-18
wip: Filter bar state fixes

lynnshaoyu:


lynnshaoyu:
this is not an FE issue, it's an OS api issue

PR #5326, opened 2022-11-18
refactor: move listinggrid to its own file

cbachmeier:


src/graphql/data/nft/NftBalance.ts
```diff
@@ -159,7 +159,7 @@ export function useNftBalanceQuery(
       last,
       before,
     },
-    { fetchPolicy: 'store-or-network' }
+    { fetchPolicy: 'network-only' }
```
vm, line 162: unrelated?

vm (approved):
idk if the fetch policy is intended but ideally can rm for this pr

src/graphql/data/nft/NftBalance.ts
cbachmeier: good catch

PR #5354, opened 2022-11-21
fix: Issue where decimal appears as 0 on listings page.

unipadmini:
Fixes WEB-2362: Bug: Period (.) ignored by price input field on the listing page.

Fixed it by leaving all string equivalents of `0` (like '0.") as is if price is undefined. Feels a tiny bit hacky, but couldn't think of a better solution that leaves a small footprint.

### DIFF
--- a/src/nft/components/profile/list/NFTListingsGrid.tsx
+++ b/src/nft/components/profile/list/NFTListingsGrid.tsx
@@ -184,6 +184,9 @@ const PriceTextInput = ({
           ref={inputRef}
           onChange={(v: FormEvent<HTMLInputElement>) => {
             const val = parseFloat(v.currentTarget.value)
+            if (!listPrice && val === 0) {
+              return
+            }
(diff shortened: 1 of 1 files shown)


src/nft/components/profile/list/NFTListingsGrid.tsx
```diff
@@ -184,6 +184,9 @@ const PriceTextInput = ({
           ref={inputRef}
           onChange={(v: FormEvent<HTMLInputElement>) => {
             const val = parseFloat(v.currentTarget.value)
```
cbachmeier, line 186: I would personally make it more explicit what edge case this is covering and do `if (v.currentTarget.value === '0.') return`

unipadmini: done

cbachmeier (approved):
LGTM

PR #5359, opened 2022-11-21
feat: approve collections 1x1

cbachmeier:
Previously we let you approve all collections all at once, this work on desktop but isn't possible on mobile as you can't queue up the requests. Now all collections must be approved one by one (unless already approved)


JackShort (approved):
lgtm

PR #5360, opened 2022-11-21
feat: fetch top tokens based on volume & update data descriptions

cartcrom:
Switched top tokens query to fetch by volume & added tooltips explaining data. Updated tooltip component to use non-deprecated color/font scheme + replaced InfoTip.tsx component w/ direct usage of tooltip component. Replaced infotips in Stats.tsx w/ tooltips w/ out Icons based on @callil's preferences


src/components/Tokens/TokenDetails/PopoverText.tsx
```diff
@@ -0,0 +1,36 @@
+import { Placement } from '@popperjs/core'
```
grabbou, line 1: Could you explain to me the usage of `popperjs` in the project? I've searched for all its occurrences and so far, I've only found it being used in 2 files as a type only import! Do we use it somewhere, e.g. to power a popover? If yes, where is that line/file located?

grabbou: Also, nit, you can use `import type` when importing types only, that way, the whole import gets stripped down while transpiling to Javascript! It is also a great way to import types from dev dependencies. But this is additional use case beyond this comment.

grabbou (dismissed):
Thanks for the PR! Overall, it feels great! Left some comments around styling generally

src/components/Tokens/TokenDetails/PopoverText.tsx
cartcrom: Thanks for the review! & for the tip abt importing types. I ended up deleting these components as I realized we could just update the existing tooltip component to reflect the latest UI theme, rather than wrapping it.

From what I know, popperjs is a library that handles creating a popover ui element that intuitively moves based on the size of the screen / positioning of other elements. we use this in src/components/Popover which is then used by src/components/Tooltip

src/pages/Tokens/index.tsx
```diff
@@ -94,9 +97,14 @@ const Tokens = () => {
...
         <TitleContainer>
-          <ThemedText.LargeHeader>
-            <Trans>Top tokens on Uniswap</Trans>
-          </ThemedText.LargeHeader>
+          <MouseoverTooltip
+            text={<Trans>This table contains the top tokens by Uniswap volume, sorted based on your input.</Trans>}
+            placement="bottom"
+          >
+            <ThemedText.LargeHeader>
+              <Trans>Top tokens on Uniswap</Trans>{' '}
```
JFrankfurt, line 105: confirming you want the added `{' '}` here

JFrankfurt (approved):

n6wbb6 (commented):
Nice job

PR #5362, opened 2022-11-21
fix: Can remove marketplace or listing from modal

cbachmeier:
Previously removing a marketplace or listing from the modal would disable the whole flow, this allows you to continue after removing either.


lynnshaoyu (approved):
would like @JackShort to take a look too, not familiar with the Listing's page and expected behavior

JackShort (approved):
lgtm

PR #5363, opened 2022-11-21
fix: include user-added tokens in selector

zzmp:


src/components/SearchModal/CurrencySearch.tsx
```diff
@@ -90,18 +90,17 @@ export function CurrencySearch({
   const sortedTokens: Token[] = useMemo(
     () =>
       !balancesAreLoading
-        ? [...filteredTokens]
+        ? filteredTokens
             .filter((token) => {
```
zzmp, line 94: `filter` returns a shallow copy of the array, so it is not necessary to first make our own shallow copy (eg `[...filteredTokens]`)

cartcrom (approved):
LGTM

PR #5365, opened 2022-11-21
feat: multiline ellipsis when text is too large

grabbou:
Fixes WEB-2376

When collection name is too large, we want to split it into two lines and eventually, truncate when the name is too long. We also want to center things vertically while keeping the overall header height. 

This is done with a `line-ellipsis` CSS feature that has relatively good support across the landscape, despite being under a prefix (Edge & Firefox support it under this prefix too). In case this is not supported, I added a maximum height for the title component that corresponds to two lines and added an `overflow: hidden`. This will hide overflowing text without putting an ellipsis, but I think it's a fair trade-off give how easy this is to support with pure CSS.

Since this is hard to reproduce without manually repeating the collection name programatically, I left some comments on Slack showing the behaviour: https://uniswapteam.slack.com/archives/C04AYQQLVQU/p1669054341919679?thread_ts=1668816740.434209&cid=C04AYQQLVQU

### DIFF
--- a/src/nft/components/explore/CarouselCard.tsx
+++ b/src/nft/components/explore/CarouselCard.tsx
@@ -81,16 +81,15 @@ const CardHeaderColumn = styled.div`
   align-items: center;
   flex-direction: column;
   gap: 8px;
-  padding: 40px;
+  height: 202px;
+  justify-content: center;
+  padding: 0 40px;
   z-index: 1;
 `
 
 const CardNameRow = styled.div`
   display: flex;
   gap: 2px;
-  overflow: hidden;
-  white-space: nowrap;
-  text-overflow: ellipsis;
 `
 const IconContainer = styled.div`
   display: flex;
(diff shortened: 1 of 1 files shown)


src/nft/components/explore/CarouselCard.tsx
```diff
@@ -81,16 +81,15 @@ const CardHeaderColumn = styled.div`
   align-items: center;
   flex-direction: column;
   gap: 8px;
-  padding: 40px;
+  height: 202px;
+  justify-content: center;
+  padding: 0 40px;
```
cbachmeier, line 86: Unrelated but can we reduce the 40px padding on mobile screens? Names are truncating seemingly early from the lost space

cbachmeier (approved):

src/nft/components/explore/CarouselCard.tsx
grabbou: Yes, I think we could reduce it to 20px on smaller devices?

grabbou: I'll do that in a follow-up when I am working on responsive tweaks

PR #5369, opened 2022-11-21
feat: nft explore promo banner

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-449?atlOrigin=eyJpIjoiODczOTZlMDZlOWYyNGM5ZDgzZDM0OGI5YTVmM2EzMDgiLCJwIjoiaiJ9


src/nft/components/nftExploreBanner/NftExploreBanner.tsx
```diff
@@ -0,0 +1,129 @@
...
+          style={{
+            width: '25%',
+            cursor: 'pointer',
+            aspectRatio: '1',
+            transition: 'transform 0.25s ease 0s',
+          }}
+          src={randomizedNftImage}
+          objectFit="contain"
+          draggable={false}
+        />
```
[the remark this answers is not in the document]
cbachmeier: img also needs rounded corners

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
```diff
@@ -0,0 +1,129 @@
...
+  z-index: ${Z_INDEX.sticky};
+  transition: ${({
+    theme: {
+      transition: { duration, timing },
+    },
+  }) => `${duration.slow} opacity ${timing.in}`};
+  :hover {
+    border: double 1px transparent;
+    border-radius: 12px;
+    background-image: ${({ theme }) =>
```
cbachmeier, line 44: it might be the screenshots in Jira are out of date but the gradient looks reversed here. In the pictures it goes light -> dark but in the build I'm getting dark -> light

cbachmeier (changes requested):

src/components/TopLevelModals/index.tsx
```diff
@@ -17,13 +19,17 @@ export default function TopLevelModals() {
...
 
   useAccountRiskCheck(account)
   const open = Boolean(blockedAccountModalOpen && account)
   return (
     <>
       <AddressClaimModal isOpen={addressClaimOpen} onDismiss={addressClaimToggle} />
       <ConnectedAccountBlocked account={account} isOpen={open} />
+      {location.pathname.includes('/swap') && <NftExploreBanner />}
+      {location.pathname.includes('/tokens') && <NftExploreBanner />}
+      {location.pathname.includes('/pool') && <NftExploreBanner />}
```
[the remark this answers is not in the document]
grabbou: Also, just a thought for future compatibility, maybe `startsWith` is better suited to avoid false positives when we nest URLs like `/swap/tokens`

src/components/TopLevelModals/index.tsx
```diff
@@ -17,13 +19,17 @@ export default function TopLevelModals() {
...
       <Bag />
       {useNftFlag() === NftVariant.Enabled && <TransactionCompleteModal />}
       {useNftFlag() === NftVariant.Enabled && <AirdropModal />}
```
grabbou, line 35: BTW, I know this is completely unrelated to this PR, but can we rewrite this too? Why not a single:
```
const isNftEnabled = useNftFlag()

{isNftEnabled && (
  <>
    <TransactionCompleteModal />
    <AirdropModal />
  </>
)}
```

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
```diff
@@ -0,0 +1,129 @@
...
+  cursor: pointer;
+  color: ${({ theme }) => theme.textPrimary};
+  height: 100px;
+  display: ${({ show }) => (show ? 'flex' : 'none')};
+  flex-direction: column;
+  bottom: 48px;
+  position: fixed;
+  right: clamp(0px, 1vw, 16px);
+  text-decoration: none;
+  width: ${Math.min(391, window.innerWidth - 5)}px;
```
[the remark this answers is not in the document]
grabbou: I think we could improve this a bit. Since we have CSS units such as vw, we should be able to refer to innerWidth of window without JavaScript. Also, we have calc and other functions to achieve what we want.

This will also get rid of the issue that @cbachmeier describes, because `window` is resolved once at the time of evaluating the CSS and is not updated dynamically. A good real life use case is changing rotation on a mobile.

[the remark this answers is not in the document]
grabbou: Also, I think it would be cleaner to use `@media` queries here.

For smaller devices, make it `width: 100vw` (or similar, depends on your padding/margin), otherwise, `391px`. Also, please make sure to follow `mobile-first` approach when writing media queries. This means the base CSS should define layout for smallest breakpoint and media queries should be starting with `min-width`.

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
```diff
@@ -0,0 +1,129 @@
...
+  font-size: 14px;
+  line-height: 20px;
+  margin: 0;
+`
+
+const TextContainer = styled.div`
+  display: flex;
+  cursor: pointer;
+  flex-direction: column;
+  flex: 0.93;
```
grabbou, line 71: I would recommend putting `padding-right` on the `InnerContainer` instead. It is the element that has `position: relative` and is responsible for laying out the `x` button that we're trying to offset here. 

I believe `flex: 1` will be much cleaner and easier to reason about.

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
```diff
@@ -0,0 +1,129 @@
...
+  gap: 8px;
+  height: 100%;
+  padding: 8px;
+`
+
+const BannerBodyText = styled.p`
+  font-size: 14px;
+  line-height: 20px;
+  margin: 0;
+`
```
grabbou, line 65: Can we use `ThemedText.BodySmall` instead from `components/text`? Looks like it has appropriate font size and line height settings out of the box. Depending on font weight, you may consider using a different one.

Either way, thanks to that, we keep text in a single unified place and don't have to re-define font sizes and line heights here! Since ThemedText is not a `p`, we don't need to override margins.

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
grabbou: Suggestion: since we're offsetting `x` anyway, maybe it's not necessary to position it absolutely in the top right corner, but use a single flex layout and display <image> <text> <x> with vertical alignement to flex-start?

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+const TextContainer = styled.div`
+  display: flex;
+  cursor: pointer;
+  flex-direction: column;
+  flex: 0.93;
+  align-items: flex-start;
+  justify-content: center;
+  z-index: 4;
```
grabbou, line 74: Consider using `zIndex` from theme

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
```diff
@@ -0,0 +1,129 @@
...
+const randomizedNftImage = getRandom([nftPromoImage1, nftPromoImage2, nftPromoImage3])
+
+const PopupContainer = styled.div<{ show: boolean }>`
+  background-color: ${({ theme }) => theme.backgroundSurface};
+  box-shadow: ${({ theme }) => theme.deepShadow};
+  border: 1px solid ${({ theme }) => theme.backgroundOutline};
+  border-radius: 12px;
+  cursor: pointer;
+  color: ${({ theme }) => theme.textPrimary};
+  height: 100px;
```
grabbou, line 27: Do we need a hardcoded height here? My approach is usually this:
- when defining height, I do not set paddings and align items vertically. This is to make sure content is always of the set height
- when defining paddings, I let the height be result of text + padding. 

I think for our banner, it doesn't have to be pixel perfect, so we can safely opt-out from height and let flex layout control its appearance!

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
```diff
@@ -0,0 +1,129 @@
...
+    background-clip: padding-box, border-box;
+  }
+`
+
+const InnerContainer = styled.div`
+  overflow: hidden;
+  display: flex;
+  position: relative;
+  gap: 8px;
+  height: 100%;
```
grabbou, line 57: Is `height: 100%` needed here? I toggled it off in the Chrome debugger and had no negative effect on the layout! I am often finding myself placing too many of these, so always good to reduce!

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
```diff
@@ -0,0 +1,129 @@
...
+            aspectRatio: '1',
+            transition: 'transform 0.25s ease 0s',
+          }}
+          src={randomizedNftImage}
+          objectFit="contain"
+          draggable={false}
+        />
+        <TextContainer>
+          {/* <HeaderText> */}
+          <div className={subhead}>
```
grabbou, line 104: Consider replacing this with `ThemedText.Subhead` or anything that works for you from `components/text.tsx`. This is pretty much equivalent to Sprinkles:
> export const subhead = sprinkles({ fontWeight: 'medium', fontSize: '16', lineHeight: '24' })

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
```diff
@@ -0,0 +1,129 @@
...
+            <Trans>Introducing Uniswap NFT</Trans>
+          </div>
+          {/* </HeaderText> */}
+
+          {/* <Description> */}
+          <BannerBodyText>
+            <Trans>Buy and sell NFTs across more listings at better prices.</Trans>{' '}
+            <StyledInternalLink to="/nfts">Explore NFTs</StyledInternalLink>{' '}
```
grabbou, line 112: General question: do we have any pattern on applying translations? Looks like the paragraph is translated, but link is not!

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
```diff
@@ -0,0 +1,129 @@
...
+        <X
+          size={20}
+          color={theme.textSecondary}
+          onClick={(e) => {
+            e.preventDefault()
+            e.stopPropagation()
+            toggleShowNftPromoBanner()
+          }}
+          style={{ cursor: 'pointer', position: 'absolute', right: 10, zIndex: 5 }}
+        />
```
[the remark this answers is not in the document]
grabbou: I +1 moving to Styled components

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
```diff
@@ -0,0 +1,129 @@
...
+  background-color: ${({ theme }) => theme.backgroundSurface};
+  box-shadow: ${({ theme }) => theme.deepShadow};
+  border: 1px solid ${({ theme }) => theme.backgroundOutline};
+  border-radius: 12px;
+  cursor: pointer;
+  color: ${({ theme }) => theme.textPrimary};
+  height: 100px;
+  display: ${({ show }) => (show ? 'flex' : 'none')};
+  flex-direction: column;
+  bottom: 48px;
```
grabbou, line 30: This should have a different unit and/or be under media query. On mobile, the banner "covers" part of the menu. I assume this was to escape the bottom right status that is displayed on desktop. On `mobile`, we need a different number.

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
```diff
@@ -0,0 +1,129 @@
...
+  position: fixed;
+  right: clamp(0px, 1vw, 16px);
```
grabbou, line 32: Do we need `clamp` here? Can we just use `adding: 1rem;`, like the bottom right status section? Would not only simplify this code, but also make sure it's always perfectly right aligned with it, no matter the window size.

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
```diff
@@ -0,0 +1,129 @@
...
+  text-decoration: none;
```
grabbou, line 33: Isn't this `none` by default on a `div`?

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
grabbou: BTW, do we really need `z-index`, `position: relative` and `align-items: flex-start` on this component? I think we don't!

grabbou: I also think we can safely drop `cursor: pointer` since we have it on the container?

grabbou (changes requested):
Thank you for sending over this PR! Overall, I played around with it and the NFT banner looks great! I just left some comments on the styling aspect, I think we could use some improvements there!

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
cbachmeier: A lot of the nft components were transferred over from Genie where we didn't have translations in place. The general practice is to always add but for existing NFT work pre launch we've agreed to go without and add later.

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
cbachmeier: Links have underlines by default

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
```diff
@@ -0,0 +1,129 @@
...
+  position: relative;
+`
+
+export default function NftExploreBanner() {
+  const theme = useTheme()
+  const [showNftPromoBanner, toggleShowNftPromoBanner] = useShowNftPromoBanner()
+  const navigate = useNavigate()
+
+  const navigateToNfts = () => {
+    navigate('/nfts')
```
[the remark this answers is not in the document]
lynnshaoyu: yes that makes sense thanks done

src/pages/App.tsx
```diff
@@ -97,15 +97,23 @@ const Marginer = styled.div`
...
+    case locationPathname.startsWith('/tokens'):
       return PageName.TOKENS_PAGE
+    case locationPathname.startsWith('/nfts/profile'):
+      return PageName.NFT_PROFILE_PAGE
+    case locationPathname.startsWith('/nfts/asset'):
+      return PageName.NFT_DETAILS_PAGE
+    case locationPathname.startsWith('/nfts/collection'):
+      return PageName.NFT_COLLECTION_PAGE
+    case locationPathname.startsWith('/nfts'):
+      return PageName.NFT_EXPLORE_PAGE
```
lynnshaoyu, line 116: btw i'm adding NFT pages here for analytics handling

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
```diff
@@ -54,51 +59,48 @@ const InnerContainer = styled.div`
...
+  z-index: 5;
+  &:hover {
+    opacity: ${({ theme }) => theme.opacity.hover};
+  }
+  &:active {
+    opacity: ${({ theme }) => theme.opacity.click};
+  }
+`
+
+const StyledImageContainer = styled(Box)`
```
cbachmeier, line 83: nit: could just be a styled `div` with display `flex`

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
```diff
@@ -0,0 +1,131 @@
...
+const TextContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  flex: 1;
+  justify-content: center;
+`
+
+const StyledXButton = styled(X)`
+  color: ${({ theme }) => theme.textSecondary};
+  z-index: 5;
```
cbachmeier, line 74: nit: still using a custom z index here

cbachmeier (approved):
LGTM only minor non-blocking nits

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
lynnshaoyu: true we don't actually need this z index removed.

PR #5371, opened 2022-11-21
style: updating details

JackShort:
* updating details to new design
* moving asset name and collection name to right side

### DIFF
--- a/src/nft/components/details/AssetPriceDetails.tsx
+++ b/src/nft/components/details/AssetPriceDetails.tsx
@@ -1,8 +1,7 @@
 import { useWeb3React } from '@web3-react/core'
 import { OpacityHoverState } from 'components/Common'
 import { useNftBalanceQuery } from 'graphql/data/nft/NftBalance'
-import useCopyClipboard from 'hooks/useCopyClipboard'
-import { CancelListingIcon } from 'nft/components/icons'
+import { CancelListingIcon, VerifiedIcon } from 'nft/components/icons'
 import { useBag, useProfilePageState, useSellAsset } from 'nft/hooks'
 import { CollectionInfoForAsset, GenieAsset, ProfilePageStateType, WalletAsset } from 'nft/types'
 import { ethNumberStandardFormatter, formatEthPrice, getMarketplaceIcon, timeLeft, useUsdPrice } from 'nft/utils'
@@ -12,6 +11,9 @@ import { Upload } from 'react-feather'
 import { Link, useNavigate } from 'react-router-dom'
 import styled, { css, useTheme } from 'styled-components/macro'
 import { ThemedText } from 'theme'
+
+const TWITTER_WIDTH = 560
+const TWITTER_HEIGHT = 480
 
 interface AssetPriceDetailsProps {
   asset: GenieAsset
@@ -45,9 +47,12 @@ const hoverState = css`
 `
 
 const Container = styled.div`
+  display: flex;
+  flex-direction: column;
   width: 100%;
+  gap: 24px;
 
-  @media (min-width: 960px) {
+  @media (min-width: 961px) {
     position: fixed;
     width: 360px;
     margin-top: -6px;
@@ -129,7 +134,8 @@ const DiscoveryContainer = styled.div`
 `
 
 const OwnerText = styled.a`
-  font-size: 14px;
+  font-size: 16px;
+  font-weight: 600;
   line-height: 20px;
   color: ${({ theme }) => theme.textSecondary};
   text-decoration: none;
@@ -140,9 +146,49 @@ const OwnerText = styled.a`
 const OwnerInformationContainer = styled.div`
   color: ${({ theme }) => theme.textSecondary};
   display: flex;
+  flex-direction: column;
   justify-content: space-between;
   padding: 0 8px;
   margin-bottom: 20px;
+`
+
+const AssetInfoContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: 8px;
+`
+
+const AssetHeader = styled.div`
+  display: -webkit-box;
+  align-items: center;
+  font-size: 28px;
+  font-weight: 500;
+  line-height: 36px;
+  -webkit-line-clamp: 2;
+  -webkit-box-orient: vertical;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  color: ${({ theme }) => theme.textPrimary};
+`
+
+const CollectionNameContainer = styled.div`
+  display: flex;
+  justify-content: space-between;
+`
+
+const CollectionHeader = styled.span`
+  display: flex;
+  align-items: center;
+  font-size: 16px;
+  font-weight: 400;
+  line-height: 24px;
+  color: ${({ theme }) => theme.textPrimary};
+  text-decoration: none;
+  ${OpacityHoverState};
+`
+
+const DefaultLink = styled(Link)`
+  text-decoration: none;
 `
 
 export const OwnerContainer = ({ asset }: { asset: GenieAsset }) => {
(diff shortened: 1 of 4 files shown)


src/nft/components/details/AssetPriceDetails.tsx
```diff
@@ -268,40 +311,43 @@ export const AssetPriceDetails = ({ asset, collection }: AssetPriceDetailsProps)
     }
   }, [asset, itemsInBag])
 
-  const isOwner = asset.owner ? account?.toLowerCase() === asset.owner?.address?.toLowerCase() : false
-
-  if (isOwner) {
-    return <OwnerContainer asset={asset} />
+  const shareTweet = () => {
+    window.open(
+      `https://twitter.com/intent/tweet?text=Check%20out%20${
```
cbachmeier, line 316: can we create a helper fn to get tweet url from a given asset?

src/nft/components/details/AssetPriceDetails.tsx
```diff
@@ -268,40 +311,43 @@ export const AssetPriceDetails = ({ asset, collection }: AssetPriceDetailsProps)
...
+        asset.name ?? `${asset.collectionName} #${asset.tokenId}`
+      }%20(${asset.collectionName})%20https://app.uniswap.org/%23/nfts/asset/${asset.address}/${
```
unipadmini, line 318: I'm not seeing the URL in the tweet

src/nft/components/details/AssetPriceDetails.tsx
```diff
@@ -342,6 +388,22 @@ export const AssetPriceDetails = ({ asset, collection }: AssetPriceDetailsProps)
       ) : (
         <NotForSale collectionName={collection.collectionName ?? 'this collection'} collectionUrl={asset.address} />
       )}
+      <OwnerInformationContainer>
+        <ThemedText.BodySmall color="textSecondary" lineHeight="20px">
+          Seller:
```
cbachmeier, line 393: Says `Seller` when not for sale

src/nft/components/details/AssetPriceDetails.tsx
```diff
@@ -268,40 +311,43 @@ export const AssetPriceDetails = ({ asset, collection }: AssetPriceDetailsProps)
...
-
-      {cheapestOrder && asset.priceInfo ? (
+      <AssetInfoContainer>
+        <CollectionNameContainer>
+          <DefaultLink to={`/nfts/collection/${asset.address}`}>
+            <CollectionHeader>
+              {collection.collectionName} {collection.isVerified && <VerifiedIcon />}
+            </CollectionHeader>
+          </DefaultLink>
+          <UploadLink onClick={shareTweet} target="_blank">
```
cbachmeier, line 339: want to check to see if they want to update the icon? Feels weird to use an upload icon to go to a different page, let alone twitter

src/nft/components/details/AssetPriceDetails.tsx
```diff
@@ -342,6 +388,22 @@ export const AssetPriceDetails = ({ asset, collection }: AssetPriceDetailsProps)
       ) : (
         <NotForSale collectionName={collection.collectionName ?? 'this collection'} collectionUrl={asset.address} />
       )}
+      <OwnerInformationContainer>
```
cbachmeier, line 391: there's a lot more bottom padding under Seller on mobile compared to the other gaps

src/nft/components/details/AssetPriceDetails.tsx
```diff
@@ -140,11 +146,51 @@ const OwnerText = styled.a`
...
+  flex-direction: column;
+  gap: 8px;
+`
+
+const AssetHeader = styled.div`
+  display: -webkit-box;
+  align-items: center;
+  font-size: 28px;
+  font-weight: 500;
+  line-height: 36px;
```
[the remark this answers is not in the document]
JackShort: no to the above - we need to add a lot to themedText

src/nft/components/details/TraitsContainer.tsx
```diff
@@ -11,7 +11,7 @@ const Grid = styled.div`
   gap: 16px;
   max-width: 780px;
 
-  @media (max-width: 960px) {
+  @media (max-width: 1080px) {
```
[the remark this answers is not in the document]
JackShort: `1080` works a lot better than `1024`

src/nft/components/details/AssetPriceDetails.tsx
JackShort: its weird and they do

unipadmini (approved):
LGTM but charlie should probably approve. Twitter link works though.

src/nft/components/details/AssetPriceDetails.tsx
```diff
@@ -52,7 +59,7 @@ const Container = styled.div`
   width: 100%;
   gap: 24px;
 
-  @media (min-width: 961px) {
+  @media (min-width: calc(960px + 1px)) {
```
cbachmeier, line 62: could you use the `MEDIA_WIDTHS.deprecated_upToMedium + 1` here?

cbachmeier (approved):
LGTM

src/nft/components/details/AssetPriceDetails.tsx
grabbou: Shouldn't we use a non deprecated value? We do have breakpoints on theme that are widely used!

BTW, why do you had to do `+1` here? I've run into this issue recently while working on a different component and noticed we had some conditionals in React code regarding breakpoints. The issue was the programatic breakpoint used `<` instead of `<=` for min-width behaviour, that made it work unlike CSS, causing this `+1px` glitch.

I would really love if we could resolve that.

src/nft/components/details/AssetPriceDetails.tsx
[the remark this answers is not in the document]
grabbou: Mind elaborating a bit on that? While it clearly lacks some features at the moment, I think it's a great addition that is being rolled out gradually in a lot of incoming PRs. 

If you feel like it's lacking some features, I think it would be great to define them and resolve (maybe even before PRs like this are merged), so that we don't make ourselves more additional work in the future.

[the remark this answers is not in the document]
grabbou: In this particular example, I am not sure what's stopping us from wrapping a ThemedText in a `styled.div`. You can always override `line-height`, if that's the intent.

src/nft/components/details/TraitsContainer.tsx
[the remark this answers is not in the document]
grabbou: doesn't it introduce a bit of inconsistency tho? I think it would be great to use same breakpoints throughout the code, even if they are not perfect in some places.

src/nft/components/details/AssetPriceDetails.tsx
```diff
@@ -342,6 +391,24 @@ export const AssetPriceDetails = ({ asset, collection }: AssetPriceDetailsProps)
...
+        <OwnerInformationContainer>
+          <ThemedText.BodySmall color="textSecondary" lineHeight="20px">
+            Seller:
+          </ThemedText.BodySmall>
+          <OwnerText
+            target="_blank"
+            href={`https://etherscan.io/address/${asset.owner.address}`}
+            rel="noopener noreferrer"
+          >
+            {asset.tokenType === 'ERC1155' ? (
```
grabbou, line 404: nit: what about flipping conditional so that we can do `{bool && ()}` instead?

src/nft/components/details/AssetPriceDetails.tsx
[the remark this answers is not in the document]
grabbou: Looks like `CollectionHeader` could also wrap a `ThemedText`, font weight 400, font-size 16 and textPrimary is exactly what BodyPrimary defines.

grabbou (dismissed):
Thanks for the PR! Overall, I don't have any comments on the functionality itself, but I do have some questions regarding styling aspect that I would like to address/discuss before this PR gets merged, hence request for a review. Thanks!

PR #5372, opened 2022-11-21
fix: after retry, put listing button in a ready state

cbachmeier:
Previously if a user had failed or rejected txs and they successfully retried via in line button press, the listing button would still be inactive. This fix checks for that case and put button back in active state.

Also fixes a color discrepancy.
https://uniswaplabs.atlassian.net/browse/WEB-2337
https://uniswaplabs.atlassian.net/browse/WEB-2373


JackShort (approved):
niceeee this works well

PR #5373, opened 2022-11-21
fix: tokenlists not being updated

cartcrom:
in progress: need to fix infinite loop caused by useInterval


src/state/lists/updater.ts
```diff
@@ -45,7 +45,7 @@ export default function Updater(): null {
     UNSUPPORTED_LIST_URLS.forEach((listUrl) => {
       const list = lists[listUrl]
       if (!list || (!list.current && !list.loadingRequestId && !list.error)) {
-        fetchList(listUrl, undefined, true).catch((error) => console.debug('list added fetching error', error))
+        fetchList(listUrl, true).catch((error) => console.debug('list added fetching error', error))
```
zzmp, line 48: `fetchList(listUrl, /* isUnsupportedList= */ true).`

zzmp (dismissed):
lgtm, I'd just like the boolean param to have an inline comment with its name for maintainability

PR #5377, opened 2022-11-22
chore: various responsive tweaks to carousel

grabbou:
Fixes WEB-2399

Fixes:
- display badge next to collection title when collection name is broken down into two lines
- text-align columns to the left (as well as hide marketplace name) when the banner width is more or less under 400px*
- slightly faster animation on carousel (follow-up from previous PR)
- smaller paddings on mobile for banner to fit more text (follow-up from previous PR)
- updated arrow colours to black/white on light/dark mode

The reason I said `more or less`:

Initially, our intention was to use CSS container queries to write a rule in a similar fashion to media queries, but relative to container width. This would let us automatically hide/show certain UI elements based on the parent dimensions.

Here's an example:

```
@supports (container-type: inline-size) {
  @container (max-width: 396px) {
    .tableElement { flex-align: left; }
  }
}

@supports not (container-type: inline-size) {
 /* backwards compatible workaround - currently implemented in this PR */
}
```

Unfortunately, it turns out this is not supported in Styled Components. The parser doesn't support `@container` queries as of 5.x version. This was just added to the parser and is being shipped with Styled Components `6.0.0-beta.4`. It might be possible to use it in the future, as soon as the final release is out and we do the upgrade!

For now, I decided to workaround it with media queries. While the upper breakpoint of 880px is what I came up with during testing to be most good looking, I am not particularly fixed on that number.

 I am pasting a reference video / pictures for this and other tweaks mentioned in this PR in a Slack thread. Feel free to check them out!

### DIFF
--- a/src/nft/components/explore/Carousel.tsx
+++ b/src/nft/components/explore/Carousel.tsx
@@ -84,7 +84,7 @@ export const Carousel = ({ children, activeIndex, toggleNextSlide }: CarouselPro
         return {
           x: (-y % (MAX_CARD_WIDTH * children.length)) + MAX_CARD_WIDTH * rank,
           immediate: vy < 0 ? prevPosition > position : prevPosition < position,
-          config: { tension: 250, friction: 35 },
+          config: { tension: 250, friction: 30 },
         }
       })
       prev.current = [firstVis, firstVisIdx]
--- a/src/nft/components/explore/Carousel.tsx
+++ b/src/nft/components/explore/Carousel.tsx
@@ -38,7 +38,7 @@ const CarouselItemCard = styled(a.div)`
 
 const CarouselItemIcon = styled.div`
   align-items: center;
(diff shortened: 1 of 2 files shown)


src/nft/components/explore/Carousel.tsx
```diff
@@ -84,7 +84,7 @@ export const Carousel = ({ children, activeIndex, toggleNextSlide }: CarouselPro
         return {
           x: (-y % (MAX_CARD_WIDTH * children.length)) + MAX_CARD_WIDTH * rank,
           immediate: vy < 0 ? prevPosition > position : prevPosition < position,
-          config: { tension: 250, friction: 35 },
+          config: { tension: 250, friction: 30 },
```
grabbou, line 87: cc: @callil

lynnshaoyu (approved):

PR #5379, opened 2022-11-22
feat: welcome nft banner

grabbou:
Fixes WEB-595

This is my first take on providing more device optimised assets by using `srcSet`.


src/nft/components/explore/WelcomeModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+import { X } from 'react-feather'
+import styled, { useTheme } from 'styled-components/macro'
+import { ThemedText } from 'theme'
+
+const Container = styled.div`
+  position: relative;
+  display: flex;
+  padding: 30% 24px 24px;
+  overflow: hidden;
+  height: fit-content;
```
grabbou, line 14: I am not particularly happy to set height as well as paddings in this case, but `fit-content` together with `overflow: hidden` was required to hide an overflow caused by an absolute background image. 

Sure, we could've just used a CSS background, but an equivalent of `src-set` for CSS backgrounds is still not widely supported.

src/nft/components/explore/WelcomeModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+`
+
+const Link = styled.a`
+  outline: none;
+  text-decoration: none;
+
+  &:hover {
+    text-decoration: underline;
+  }
+`
```
grabbou, line 32: General question: is there a better way to handle links within the app? This looks like something you'd typically want to do for every link!

src/state/user/hooks.tsx
```diff
@@ -117,6 +118,15 @@ export function useShowSurveyPopup(): [boolean | undefined, (showPopup: boolean)
...
 }
 
+export function useHideNFTWelcomeModal(): [boolean | undefined, () => void] {
+  const dispatch = useAppDispatch()
+  const hideNFTWelcomeModal = useAppSelector((state) => state.user.hideNFTWelcomeModal)
+  const hideModal = useCallback(() => {
+    dispatch(updateHideNFTWelcomeModal({ hideNFTWelcomeModal: true }))
+  }, [dispatch])
+  return [hideNFTWelcomeModal, hideModal]
+}
```
grabbou, line 128: Open to feedback if there's better way to do it! Based on other PRs, but it looks a bit verbose for just a single boolean!

lynnshaoyu: i think this is the standard way we do it, so lgtm

grabbou: Thanks! I was referring to your PR actually, but just couldn't bring the number up while writing this comment!

src/nft/components/explore/WelcomeModal.tsx
lynnshaoyu: Yes pls look at StyledInternalLink or StyledLink from src/theme/components/index.tsx

src/nft/pages/explore/index.tsx
```diff
@@ -24,6 +25,7 @@ const ExploreContainer = styled.div`
 
 const NftExplore = () => {
   const setBagExpanded = useBag((state) => state.setBagExpanded)
+  const [isModalHidden, hideModal] = useHideNFTWelcomeModal()
```
lynnshaoyu, line 28: smart🤓🧠

lynnshaoyu (approved):
pls just change the link to use one of the predefined Link Components but other than that LGTM

grabbou:
@lynnshaoyu I changed it as per your suggestion, thanks for pointing this out for me! 

Important, there's an ongoing discussion on how this should look like on mobile which has to do with the way Modal component is rendered on mobile viewports. In case we decide to apply margins, I will send a follow up PR.

PR #5380, opened 2022-11-22
fix: searchbar fetching promise resolve fallback

cartcrom:
Added check for whether API_URL's exist from process.env


cbachmeier (approved):
Tested with and w/o nft feature flag. LGTM

PR #5381, opened 2022-11-22
fix: throw on missing env var url

zzmp:
Throws on missing env var URLs. This prevents red herrings / false alarms from network fetches that are not actually intended.

### DIFF
--- a/src/graphql/data/fetchGraphQL.ts
+++ b/src/graphql/data/fetchGraphQL.ts
@@ -2,17 +2,13 @@ import { Variables } from 'react-relay'
 import { GraphQLResponse, RequestParameters } from 'relay-runtime'
 
 const TOKEN_URL = process.env.REACT_APP_AWS_API_ENDPOINT
-const NFT_URL = process.env.REACT_APP_NFT_AWS_API_ENDPOINT ?? ''
+const NFT_URL = process.env.REACT_APP_NFT_AWS_API_ENDPOINT
+if (!TOKEN_URL) throw new Error('Token URL missing from environment')
 
-if (!TOKEN_URL) {
-  throw new Error('AWS URL MISSING FROM ENVIRONMENT')
-}
-
-const baseHeaders = {
-  'Content-Type': 'application/json',
-}
+const baseHeaders = { 'Content-Type': 'application/json' }
+const tokenHeaders = { ...baseHeaders }
 const nftHeaders = {
-  'Content-Type': 'application/json',
+  ...baseHeaders,
   'from-x-api-key': process.env.REACT_APP_NFT_FROM_AWS_X_API_KEY ?? '',
   'x-api-key': process.env.REACT_APP_NFT_AWS_X_API_KEY ?? '',
 }
@@ -29,16 +25,25 @@ const NFT_QUERIES = [
   'NftBalancePaginationQuery',
 ]
 
+const fetchToken = (body: string): Promise<Response> => {
+  return fetch(TOKEN_URL, { method: 'POST', body, headers: tokenHeaders })
+}
+
+const fetchNFT = (body: string): Promise<Response> => {
+  if (!NFT_URL) {
+    throw new Error('NFT URL missing from environment')
+  }
+  return fetch(NFT_URL, { method: 'POST', body, headers: nftHeaders })
+}
+
 const fetchQuery = (params: RequestParameters, variables: Variables): Promise<GraphQLResponse> => {
-  const isNFT = NFT_QUERIES.includes(params.name)
   const body = JSON.stringify({
     query: params.text, // GraphQL text from input
     variables,
   })
-  const url = isNFT ? NFT_URL : TOKEN_URL
-  const headers = isNFT ? nftHeaders : baseHeaders
-
-  return fetch(url, { method: 'POST', body, headers })
+  const isNFT = NFT_QUERIES.includes(params.name)
+  const response = isNFT ? fetchNFT(body) : fetchToken(body)
+  return response
     .then((res) => res.json())
     .catch((e) => {
       console.error(e)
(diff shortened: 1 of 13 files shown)


src/graphql/data/fetchGraphQL.ts
```diff
@@ -2,17 +2,13 @@ import { Variables } from 'react-relay'
 import { GraphQLResponse, RequestParameters } from 'relay-runtime'
 
 const TOKEN_URL = process.env.REACT_APP_AWS_API_ENDPOINT
-const NFT_URL = process.env.REACT_APP_NFT_AWS_API_ENDPOINT ?? ''
+const NFT_URL = process.env.REACT_APP_NFT_AWS_API_ENDPOINT
+if (!TOKEN_URL) throw new Error('Token URL missing from environment')
```
cbachmeier, line 6: should this be moved to `fetchToken` for consistency?

src/nft/queries/genie/SearchTokensFetcher.ts
```diff
@@ -1,7 +1,9 @@
 import { FungibleToken } from '../../types'
+import { getTokenUrl } from '../url'
 
 export const fetchSearchTokens = async (tokenQuery: string): Promise<FungibleToken[]> => {
-  const url = `${process.env.REACT_APP_TEMP_API_URL}/tokens/search?tokenQuery=${tokenQuery}`
+  // TODO
```
cbachmeier, line 5: todo?

src/nft/queries/genie/TrendingTokensFetcher.ts
```diff
@@ -1,9 +1,11 @@
 import { unwrapToken } from 'graphql/data/util'
 
 import { FungibleToken } from '../../types'
+import { getTokenUrl } from '../url'
 
 export const fetchTrendingTokens = async (numTokens?: number): Promise<FungibleToken[]> => {
-  const url = `${process.env.REACT_APP_TEMP_API_URL}/tokens/trending${numTokens ? `?numTokens=${numTokens}` : ''}`
+  // TODO
```
cbachmeier, line 7: todo?

src/nft/queries/genie/ActivityFetcher.ts
```diff
@@ -13,9 +14,9 @@ export const ActivityFetcher = async (
 
   const tokenId = filters?.token_id ? `&token_id=${filters?.token_id}` : ''
 
-  const url = `${process.env.REACT_APP_GENIE_V3_API_URL}/collections/${contractAddress}/activity?limit=${
-    limit ? limit : '25'
-  }${filterParam}${cursor ? `&cursor=${cursor}` : ''}${tokenId}`
+  const url = `${getNftUrl()}}/collections/${contractAddress}/activity?limit=${limit ? limit : '25'}${filterParam}${
```
cbachmeier, line 17: extra curly brace after `getNftUrl()`, currently breaks activity fetch

cbachmeier (changes requested):
1 minor bug, rest looks good

src/graphql/data/fetchGraphQL.ts
```diff
@@ -2,17 +2,13 @@ import { Variables } from 'react-relay'
...
 
-if (!TOKEN_URL) {
-  throw new Error('AWS URL MISSING FROM ENVIRONMENT')
-}
-
-const baseHeaders = {
-  'Content-Type': 'application/json',
-}
+const baseHeaders = { 'Content-Type': 'application/json' }
+const tokenHeaders = { ...baseHeaders }
```
grabbou, line 9: Mind providing a bit more context into why we're copying headers here?

src/graphql/data/fetchGraphQL.ts
grabbou: Actually I came in here to leave the same question, but the other way around! If the intent is to `throw on missing env var URLs.`, shouldn't we throw on a top-level to break early?

If I am not mistaken, this code is bundled and then, run client side, so the only time where this can happen is while development and/or bundling.

grabbou: One additional suggestion, we could move all environmental variables into a single `env` file. Not only would it be easier to handle missing variables, but also to provide proper Typescript support

src/graphql/data/fetchGraphQL.ts
```diff
@@ -29,16 +25,25 @@ const NFT_QUERIES = [
...
-  const isNFT = NFT_QUERIES.includes(params.name)
   const body = JSON.stringify({
     query: params.text, // GraphQL text from input
     variables,
   })
-  const url = isNFT ? NFT_URL : TOKEN_URL
-  const headers = isNFT ? nftHeaders : baseHeaders
-
-  return fetch(url, { method: 'POST', body, headers })
+  const isNFT = NFT_QUERIES.includes(params.name)
```
grabbou, line 44: If we have `fetchNFT` and `fetchToken`, do we need a generic `fetchQuery` anymore? There's something about `isNFT` that I don't like... 🫣

cbachmeier: this bool will be removed once we've launched, are not hidden behind a feature flag, and get the data from the same endpoint

src/graphql/data/fetchGraphQL.ts
cbachmeier: the NFT queries are hidden behind the feature flag so we wouldn't want missing that endpoint to break the default site

src/graphql/data/fetchGraphQL.ts
grabbou: thanks for explaining that!

cbachmeier:
Is this something we still want to do @zzmp or should this PR be closed?

zzmp:
I'd like to clean this up so it's handled more cleanly, so don't close it quite yet.

cbachmeier:
@zzmp I think this can be closed with the Apollo migration

PR #5383, opened 2022-11-22
fix: hide collection search for > 300 collections due to lack of search param input for OS api

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-2379?atlOrigin=eyJpIjoiNTQxYWU2NzRkODc2NGExMTg1YjY1NzU5MGI0MTEwNzIiLCJwIjoiaiJ9


src/nft/components/profile/view/FilterSidebar.tsx
```diff
@@ -81,6 +92,12 @@ export const FilterSidebar = ({
       easing: easings.easeOutSine,
     },
   })
+
+  const hideSearch = useMemo(
+    () => walletCollections && walletCollections?.length >= WALLET_COLLECTIONS_PAGINATION_LIMIT,
+    [walletCollections]
```
cbachmeier, line 98: this won't take effect unless a user scroll so that they have > 300 collections loaded, I'd suggest passing in `hasNextPage` from the infinite query and then checking if that's true and they have 300 collections loaded

cbachmeier (changes requested):

src/nft/components/profile/view/FilterSidebar.tsx
lynnshaoyu: it will, because its greater than or EQUAL to. I tested with pranksy's wallet in simulator. also hasNextPage will become false at some point at the bottom of the scroll, which leads to the search bar appearing again (undesirable effect, can confuse users)

src/nft/components/profile/view/FilterSidebar.tsx
```diff
@@ -200,10 +220,12 @@ const CollectionSelect = ({
       </Box>
       <Box paddingBottom="12" borderRadius="8">
         <Column as="ul" paddingLeft="0" gap="10" style={{ maxHeight: '80vh' }}>
-          <CollectionFilterSearch
-            collectionSearchText={collectionSearchText}
-            setCollectionSearchText={setCollectionSearchText}
-          />
+          {!hideSearch && (
```
cbachmeier, line 223: we should also hide when loading so users don't see the search bar while collections are loading in and then it gets removed

cbachmeier (changes requested):

src/nft/components/profile/view/FilterSidebar.tsx
lynnshaoyu: great catch thanks done.

cbachmeier (approved):
LGTM

PR #5386, opened 2022-11-23
fix: a few bugs on details

JackShort:
* hover listing state
* tweet url

### DIFF
--- a/src/nft/components/details/AssetPriceDetails.tsx
+++ b/src/nft/components/details/AssetPriceDetails.tsx
@@ -394,7 +394,7 @@ export const AssetPriceDetails = ({ asset, collection }: AssetPriceDetailsProps)
       )}
       {isForSale && (
         <OwnerInformationContainer>
-          {asset.owner.address && (
+          {asset.tokenType !== 'ERC1155' && asset.owner.address && (
             <ThemedText.BodySmall color="textSecondary" lineHeight="20px">
               Seller:
             </ThemedText.BodySmall>
--- a/src/nft/components/details/AssetPriceDetails.tsx
+++ b/src/nft/components/details/AssetPriceDetails.tsx
@@ -394,9 +394,11 @@ export const AssetPriceDetails = ({ asset, collection }: AssetPriceDetailsProps)
       )}
       {isForSale && (
         <OwnerInformationContainer>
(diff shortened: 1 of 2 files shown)


src/nft/components/details/AssetPriceDetails.tsx
```diff
@@ -96,6 +96,7 @@ const MarketplaceIcon = styled.img`
 `
```
cbachmeier, line 96: While you're in the area, could you hide `Seller` if owner data is undefined?

src/nft/components/details/AssetPriceDetails.tsx
```diff
@@ -96,6 +96,7 @@ const MarketplaceIcon = styled.img`
...
 
 const BuyNowButton = styled.div<{ assetInBag: boolean; margin: boolean; useAccentColor: boolean }>`
+  position: relative;
```
cbachmeier, line 99: not getting the new hover state

JackShort: I am getting it on all browsers - it should be really subtle

src/nft/components/details/AssetPriceDetails.tsx
```diff
@@ -393,9 +394,11 @@ export const AssetPriceDetails = ({ asset, collection }: AssetPriceDetailsProps)
       )}
       {isForSale && (
         <OwnerInformationContainer>
-          <ThemedText.BodySmall color="textSecondary" lineHeight="20px">
-            Seller:
-          </ThemedText.BodySmall>
+          {asset.owner.address && (
```
cbachmeier, line 397: might need to cast as a bool, still seeing it here `nfts/asset/0xded56b58644373020393ad35b9eb5841296b886b/1`

src/nft/components/details/AssetPriceDetails.tsx
cbachmeier: can see it in light mode but personally can't tell on dark, that's on design though

cbachmeier (approved):
LGTM minor bug with Seller hiding

src/nft/components/details/AssetPriceDetails.tsx
JackShort: ohh this was because its an erc1155

PR #5387, opened 2022-11-23
feat: differentiate modal from bottomsheet

grabbou:
Modal is a really useful component, but in reality, it's has a behaviour closer to a bottom sheet on mobile devices. This results in a welcome modal being rendered full-screen on mobile.

After discussing with product and @ianlapham, decided to differentiate that behaviour behind a `isBottomSheet` prop.

Also did some smaller refactoring of a styled component, mainly:
- removed `attrs` and just passed a regular `aria-label` (HTML output is the same)
- removed props destructuring. If we really don't want to pass props down to an underlying component, we can prefix with a $ sign
- added types to make sure we don't misuse some of the props
- removed extraneous "as" from StyledDialogOverlay - it is a StyledDialogOverlay anyway, so that line was redundant 
- added `onSwipe` to be used instead of `onDismiss` when all you want is just a swipe to close functionality

Now, if you want to render a nice modal within the app that is not meant to look like a BottomSheet, use `isBottomSheet={false}` and you're all set!

### DIFF
--- a/src/components/Modal/index.tsx
+++ b/src/components/Modal/index.tsx
@@ -24,50 +24,51 @@ const StyledDialogOverlay = styled(AnimatedDialogOverlay)<{ scrollOverlay?: bool
   }
 `
 
+type StyledDialogProps = {
+  $minHeight?: number | false
+  $maxHeight?: number
+  $isBottomSheet?: boolean
+  $scrollOverlay?: boolean
+  $hideBorder?: boolean
+  $maxWidth: number
+}
+
 const AnimatedDialogContent = animated(DialogContent)
-// destructure to not pass custom props to Dialog DOM element
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const StyledDialogContent = styled(
-  ({ hideBorder, maxWidth, minHeight, maxHeight, mobile, isOpen, scrollOverlay, ...rest }) => (
-    <AnimatedDialogContent {...rest} />
-  )
-).attrs({
-  'aria-label': 'dialog',
-})`
+const StyledDialogContent = styled(AnimatedDialogContent)<StyledDialogProps>`
   overflow-y: auto;
 
   &[data-reach-dialog-content] {
     margin: auto;
     background-color: ${({ theme }) => theme.deprecated_bg0};
-    border: ${({ theme, hideBorder }) => !hideBorder && `1px solid ${theme.deprecated_bg1}`};
+    border: ${({ theme, $hideBorder }) => !$hideBorder && `1px solid ${theme.deprecated_bg1}`};
     box-shadow: ${({ theme }) => theme.deepShadow};
     padding: 0px;
     width: 50vw;
     overflow-y: auto;
     overflow-x: hidden;
 
-    align-self: ${({ mobile }) => mobile && 'flex-end'};
-    max-width: ${({ maxWidth }) => maxWidth}px;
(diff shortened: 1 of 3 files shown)


src/components/Modal/index.tsx
```diff
@@ -141,13 +143,13 @@ export default function Modal({
...
-                scrollOverlay={scrollOverlay}
-                hideBorder={hideBorder}
-                maxWidth={maxWidth}
+                aria-label="dialog"
+                $minHeight={minHeight}
+                $maxHeight={maxHeight}
+                $isBottomSheet={isBottomSheet}
+                $scrollOverlay={scrollOverlay}
+                $hideBorder={hideBorder}
+                $maxWidth={maxWidth}
```
grabbou, line 152: While I wasn't necessarily happy with the prefix when I first saw it, I think it's a really good differentiator between props for the styled component vs actual props for the component (they can have impact on styling too, but have actual impact on the behaviour)

src/components/Modal/index.tsx
```diff
@@ -116,7 +121,7 @@ export default function Modal({
         y: state.down ? state.movement[1] : 0,
       })
       if (state.movement[1] > 300 || (state.velocity > 3 && state.direction[1] > 0)) {
-        onDismiss()
+        onSwipe?.()
```
grabbou, line 124: Right now, I left the ability to control both dismissing (which in this context means pressing an ESC key or tapping an overlay outside of the modal) and swiping (which means dismissing a modal with a gesture). 

While this works, I wonder whether we should automatically disable dismissing when `isBottomSheet` is configured to false. In such scenario, it behaves more like a standard modal and these interactions are typical for an actual bottom sheet.

Thoughts?

src/components/Modal/index.tsx
ianlapham: yeah i think this is fine

src/components/Modal/index.tsx
```diff
@@ -99,7 +102,9 @@ export default function Modal({
   maxWidth = 420,
   initialFocusRef,
   children,
+  onSwipe = onDismiss,
```
ianlapham, line 105: so is the idea if we dont want a modal to be dismissable by a swipe, we pass undefined for onSwipe and a valid onDismiss function?

src/components/Modal/index.tsx
```diff
@@ -99,7 +102,9 @@ export default function Modal({
...
   scrollOverlay,
+  isBottomSheet = isMobile,
```
ianlapham, line 107: ahh this line is the key

ianlapham (approved):
Looks good to me! Good idea on this change, i think eventually we'll want to start creating different modal components, but not something we want to do right now.

src/components/Modal/index.tsx
grabbou: Yeah, you could pass `null` explicitly, but I didn't really think about that use case.

The primary use case was to disable dismissing with overlay/ESC key (onDismiss from Overlay component) but allow gesture. Theoretically, what you are describing is possible too!

grabbou:
Yes! I think at least right now, I see opportunity for a BottomSheet and a PopUp, but I will definitely have a better understanding when I build a few more features!

JackShort (changes requested):
lgtm

JackShort (approved):
i meant to approve

PR #5388, opened 2022-11-23
chore: fix invalid css color prop on text components

grabbou:
Before:
Screenshot 2022-11-23 at 16 32 22

After:
Screenshot 2022-11-23 at 16 32 32

Eventually, I think we're going to migrate away from this Rebass Text component, so I consider this a temporary solution. Ideally, we shouldn't be passing any of the props down the tree, or even configure it this way.

For example, this:
```js
export const ThemedText = {
  HeadlineSmall(props: TextProps) {
    return <TextWrapper fontWeight={600} fontSize={20} lineHeight="28px" color="textPrimary" {...props} />
  },
};
```
could be as simple as
```js
export const HeadlineSmall = styled(BaseText)`
   fontWeight: 600;
   fontSize: 20px;
   lineHeight: 28px;
   color: ${({theme}) => theme.textPrimary}
`
```

Like I said, something to keep in mind for the future

### DIFF
--- a/src/components/SearchModal/CurrencyList/__snapshots__/index.test.tsx.snap
+++ b/src/components/SearchModal/CurrencyList/__snapshots__/index.test.tsx.snap
@@ -171,7 +171,7 @@ exports[`renders currency rows correctly when currencies list is non-empty 1`] =
             </div>
           </div>
           <div
-            class="c10 css-1j6a53a"
+            class="c10 css-yfjwjl"
           >
             DAI
           </div>
@@ -246,7 +246,7 @@ exports[`renders currency rows correctly when currencies list is non-empty 1`] =
             </div>
           </div>
           <div
-            class="c10 css-1j6a53a"
+            class="c10 css-yfjwjl"
           >
             USDC
           </div>
@@ -321,7 +321,7 @@ exports[`renders currency rows correctly when currencies list is non-empty 1`] =
             </div>
           </div>
           <div
-            class="c10 css-1j6a53a"
+            class="c10 css-yfjwjl"
           >
             WBTC
           </div>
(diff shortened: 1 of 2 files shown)


src/theme/components/text.tsx
```diff
@@ -5,7 +5,9 @@
 import { Text, TextProps as TextPropsOriginal } from 'rebass'
 import styled from 'styled-components/macro'
 
-const TextWrapper = styled(Text)<{ color: keyof string }>`
+const TextWrapper = styled(Text).withConfig({
+  shouldForwardProp: (prop) => prop !== 'color',
```
zzmp, line 9: nit: I think this becomes more maintainable using the Array.prototype.includes pattern showcased in the [docs](https://styled-components.com/docs/api#shouldforwardprop):

    shouldForwardProp: (prop) => !['color'].includes(prop)

zzmp (approved):
lgtm, but I have a nit.

I'd prefer you make the suggested change, but it is a nit, so if you disagree with it you're also entitled to merge as-is. Either way, you can merge without requesting a re-review.

(In general with my reviews, anything prefixed with "nit: " carries this caveat, in the interest of developer velocity: I'd prefer it be changed, but if you disagree then it can be left as-is.)

grabbou:
Thanks @zzmp for a prompt review and additional context on the nit picks! That's totally fine by me! Actually, what's even better is that, while the nit may look really irrelevant/small at a first sight, it opens up an interesting discussion on interacting with Styled Components API, so I think it's more than welcome!

src/components/SearchModal/CurrencyList/__snapshots__/index.test.tsx.snap
```diff
@@ -171,7 +171,7 @@ exports[`renders currency rows correctly when currencies list is non-empty 1`] =
             </div>
           </div>
           <div
-            class="c10 css-1j6a53a"
+            class="c10 css-yfjwjl"
```
zzmp, line 174: This is a job for another PR, but I bet we could configure styled-components to make these names more deterministic. See this section of the docs:

- https://styled-components.com/docs/tooling#better-debugging

PR #5389, opened 2022-11-23
fix: always sort not for sale assets towards the end when sorting by price

cbachmeier:


JackShort (approved):
lgtm - can you just make the undefined check set it to 0 instead of -1 for consistency now?

PR #5390, opened 2022-11-23
test: install fixed version of yarn-deduplicate

zzmp:


.github/workflows/test.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup
-      - run: npx yarn-deduplicate --strategy=highest --list --fail
+      - run: yarn yarn-deduplicate --strategy=highest --list --fail
```
grabbou, line 24: nit: `yarn deduplicate --list --fail` instead to share the strategy with `deduplicate` script?

grabbou (approved):
LGTM! Left one nit. I follow similar approach to you when it comes to nitpicks, so feel free to merge without any further action.

.github/workflows/test.yml
vm: the list and fail is for the build to error

PR #5391, opened 2022-11-23
fix: Stop traits from url re-appearing

cbachmeier:
Previously we had a bug where when you loaded a page with a url that included filters, when you tried to remove a filter it would re-appear. This was caused by the useeffect that applied filters to the url and vice versa firing when it shouldn't have. This removes the unneeded triggering condition. You should still be able to apply and remove filters as usual and when you land on a page from a url with filters and remove it, the url should be updated and the filter should not re-appear.


JackShort (approved):
lgtm

PR #5392, opened 2022-11-23
fix: incorrect looks rare creator royalties

cbachmeier:
LooksRare has a flat 0.5% creator royalty fee as opposed to using the creator set basis points. This information is now correctly displayed to users. The listing functionality already took that into account and did not need to be updated.

### DIFF
--- a/src/nft/components/bag/profile/utils.ts
+++ b/src/nft/components/bag/profile/utils.ts
@@ -2,7 +2,7 @@ import type { JsonRpcSigner, Web3Provider } from '@ethersproject/providers'
 import { LOOKSRARE_MARKETPLACE_CONTRACT, X2Y2_TRANSFER_CONTRACT } from 'nft/queries'
 import { OPENSEA_CROSS_CHAIN_CONDUIT } from 'nft/queries/openSea'
 import { AssetRow, CollectionRow, ListingMarket, ListingRow, ListingStatus, WalletAsset } from 'nft/types'
-import { approveCollection, signListing } from 'nft/utils/listNfts'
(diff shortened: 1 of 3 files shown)


src/nft/components/bag/profile/utils.ts
```diff
@@ -124,11 +124,10 @@ export const getTotalEthValue = (sellAssets: WalletAsset[]) => {
...
     if (asset.newListings?.length) {
       const maxListing = asset.newListings.reduce((a, b) => ((a.price ?? 0) > (b.price ?? 0) ? a : b))
-      return (
-        total +
-        (maxListing.price ?? 0) -
-        (maxListing.price ?? 0) * ((maxListing.marketplace.fee + asset.basisPoints / 100) / 100)
-      )
+      // LooksRare is a unique case where creator royalties are a flat 0.5% or 50 basis points
+      const maxFee =
+        maxListing.marketplace.fee + (maxListing.marketplace.name === 'LooksRare' ? 50 : asset.basisPoints) / 100
```
unipadmini, line 129: could we extract `50` into a constant so it's super clear that this represents their flat fee?

src/nft/components/profile/list/NFTListingsGrid.tsx
```diff
@@ -370,6 +370,12 @@ const MarketplaceRow = ({
     }
   }
 
+  // LooksRare is a unique case where royalties for creators are a flat 0.5%
+  const royalties = useMemo(
+    () => (selectedMarkets.length === 1 && selectedMarkets[0].name === 'LooksRare' ? 0.5 : asset.basisPoints * 0.01),
```
unipadmini, line 375: same, could we extract 0.5 into a const?

JackShort (dismissed):
lgtm

unipadmini (approved):
🚀

PR #5399, opened 2022-11-23
fix: reduce success image size

cbachmeier:
Reduces the size of a single purchases asset in success screen. Also aligns the unicorn icon with the Header text.

### DIFF
--- a/src/nft/utils/transactionResponse.ts
+++ b/src/nft/utils/transactionResponse.ts
@@ -54,7 +54,7 @@ export const parseTransactionResponse = (transactionResponse: TxResponse | undef
 export const getSuccessfulImageSize = (numSuccessful: number, isMobile: boolean) => {
   const sizeModifier = isMobile ? 2 : 1
   if (numSuccessful === 1) {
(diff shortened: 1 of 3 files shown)


src/nft/utils/transactionResponse.ts
```diff
@@ -54,7 +54,7 @@ export const parseTransactionResponse = (transactionResponse: TxResponse | undef
 export const getSuccessfulImageSize = (numSuccessful: number, isMobile: boolean) => {
   const sizeModifier = isMobile ? 2 : 1
   if (numSuccessful === 1) {
-    return 574 / sizeModifier
+    return 474 / sizeModifier
```
JackShort, line 57: I think we should make each of these breakpoints into a enum for this function

JackShort (approved):
lgtm

src/nft/utils/transactionResponse.ts
cbachmeier: This will all be re-written and standardized when we make the tx complete modal its own page

PR #5402, opened 2022-11-26
feat: Match tokens table & price designs with nft

yyip-dev:
This addresses WEB-2044, WEB-2021, and WEB-2250 to align token section designs more closely with nft's:
- Update price change text styling
- Reduce price change arrow icon size
- Increase tokens table + header max-width

Before vs. After:

Table
Screen Shot 2022-11-26 at 1 59 34 PM
Screen Shot 2022-11-26 at 2 01 13 PM

Detail
Screen Shot 2022-11-26 at 2 01 48 PM
Screen Shot 2022-11-26 at 2 01 55 PM

Search
Screen Shot 2022-11-26 at 2 02 02 PM
Screen Shot 2022-11-26 at 2 02 09 PM


src/components/Tokens/TokenDetails/PriceChart.tsx
```diff
@@ -42,28 +42,30 @@ export function calculateDelta(start: number, current: number) {
   return (current / start - 1) * 100
 }
 
-export function getDeltaArrow(delta: number | null | undefined) {
+export function getDeltaArrow(delta: number | null | undefined, iconSize: number) {
```
vm, line 45: should we default to 20?

src/components/NavBar/SuggestionRow.tsx
```diff
@@ -22,11 +22,27 @@ import { useCallback, useEffect, useState } from 'react'
...
 `
+const PriceChangeContainer = styled.div`
+  display: flex;
+  align-items: center;
+`
+
+const PriceChangeText = styled.span<{ delta: number }>`
+  font-size: 14px;
+  line-height: 20px;
+  color: ${({ theme, delta }) => (delta >= 0 ? theme.accentSuccess : theme.accentFailure)};
```
vm, line 39: nit: would suggest passing in `isNegative` or `isPositive` instead of delta

vm (approved):

src/components/Tokens/TokenDetails/PriceChart.tsx
yyip-dev: Great idea! Nice to get that by default and less redundancy across the few places we're using it

src/components/NavBar/SuggestionRow.tsx
yyip-dev: Updated this too

PR #5406, opened 2022-11-28
fix(banner): use a high-contrast text color against average banner image color

JFrankfurt:


src/hooks/useColor.ts
```diff
@@ -63,7 +63,7 @@ async function getColorFromUriPath(uri: string): Promise<string | null> {
     detectedHex = shade(0.005, detectedHex)
     AAscore = hex(detectedHex, '#FFF')
   }
-
```
cbachmeier, line 66: left in logging

src/nft/components/explore/CarouselCard.tsx
```diff
@@ -295,13 +306,16 @@ const CollectionName = styled(ThemedText.MediumHeader)`
...
+  useEffect(() => {
+    getColorFromUriPath(collection.bannerImageUrl).then((c) => c && setAverageBanerImageColor(c))
+  }, [collection.bannerImageUrl])
   return (
     <CardHeaderContainer src={collection.bannerImageUrl}>
       <CardHeaderColumn>
         <CollectionImage src={collection.imageUrl} />
         <CollectionNameContainer>
-          <CollectionName color={theme.accentTextLightPrimary} fontWeight="500">
+          <CollectionName style={{ color: readableColor(averageBanerImageColor) }} fontWeight="500">
```
cbachmeier, line 318: could you add this style to the styled component `CollectionName` and pass in the color?

cbachmeier: this also not effective for BAYC or MAYC, the text is dark on a dark background

src/nft/components/explore/CarouselCard.tsx
```diff
@@ -113,6 +108,22 @@ const CollectionImage = styled.img`
   border-radius: 100px;
 `
 
+const LoadingCardHeaderContainer = styled.div`
```
cbachmeier, line 111: Could you add some context for what this loading change is for?

src/nft/components/explore/Banner.tsx
```diff
@@ -165,7 +165,7 @@ const Banner = () => {
                 <CarouselCard
                   key={collection.address}
                   collection={collection}
-                  onClick={() => navigate(`/nfts/collection/${collection.address}`)}
+                  href={`/nfts/collection/${collection.address}`}
```
cbachmeier, line 168: need to remove `navigate` to resolve linter

cbachmeier (changes requested):

src/nft/components/explore/CarouselCard.tsx
JFrankfurt: yeah, this isn't a perfect solution for i.e., collections w/ a light enough average color that `readableColor` selects black, even though the actual text render area on the banner image is v dark :/

src/nft/components/explore/CarouselCard.tsx
JFrankfurt: the old loading component used the same wrapper as the non-loading component, so if we use `link` (which we should) it demands a `to` prop which we don't want it to have. I also re-organized the styled-components definitions to group the loading<name> components.

src/nft/components/explore/Banner.tsx
JFrankfurt: thx for the catch

src/nft/components/explore/CarouselCard.tsx
```diff
@@ -228,12 +239,12 @@ const MARKETS_ENUM_TO_NAME = {
...
 }
 
-export const CarouselCard = ({ collection, onClick }: CarouselCardProps) => {
+export const CarouselCard = ({ collection, href }: CarouselCardProps) => {
   const gqlCollection = useCollectionQuery(collection.address)
 
   return (
-    <CarouselCardBorder>
-      <CarouselCardContainer onClick={onClick}>
+    <CarouselCardBorder to={href}>
```
grabbou, line 246: nit: I feel like this component could now carry a better name (definitely with a `Link` at the end) to work better with a `to` prop.

grabbou: `CarouselLink`?

src/nft/components/explore/CarouselCard.tsx
```diff
@@ -113,6 +108,22 @@ const CollectionImage = styled.img`
...
+  background: linear-gradient(
+    to left,
+    ${({ theme }) => theme.backgroundInteractive} 25%,
+    ${({ theme }) => theme.backgroundOutline} 50%,
+    ${({ theme }) => theme.backgroundInteractive} 75%
+  );
+  will-change: background-position;
+  background-size: 400%;
+`
+const LoadingCarouselCardBorder = styled.div`
```
grabbou, line 124: tip: try https://styled-components.com/docs/api#as-polymorphic-prop instead of creating a new component

src/nft/components/explore/CarouselCard.tsx
```diff
@@ -228,12 +239,12 @@ const MARKETS_ENUM_TO_NAME = {
   [Markets.LooksRare]: 'LooksRare',
 }
 
-export const CarouselCard = ({ collection, onClick }: CarouselCardProps) => {
+export const CarouselCard = ({ collection, href }: CarouselCardProps) => {
```
grabbou, line 242: Is this change documented in the PR desc/title? I think I initially got confused what that had to do with the high-contrast feature!

src/nft/components/explore/CarouselCard.tsx
grabbou: +1 on hiding this behind a styled components interface! Also, consider prefixing a custom property with a `$` so that it doesn't propagate to children (generally good practice for props that are Styled Components exclusive and used for styling)

grabbou (changes requested):
Thanks for the PR! This is definitely a great UX additional to the already beautiful banner! I am going to request changes on this PR to move it off the list and also to further follow your work in that area!

src/nft/components/explore/CarouselCard.tsx
cbachmeier: Honestly I think the Cal's solution of making the color explicitly white for both modes would look better for the majority of cases. https://uniswapteam.slack.com/archives/C04BVC4J5HU/p1669653436830549

JFrankfurt:
closing because white works best for current banner collections and the readable text solution isn't great

PR #5407, opened 2022-11-28
fix: Minor polish fixes

aballerr:
https://uniswaplabs.atlassian.net/browse/WEB-2441
https://uniswaplabs.atlassian.net/browse/WEB-2436
https://uniswaplabs.atlassian.net/browse/WEB-2377

I also fixed an issue where the loading state wasn't aligned properly on the collection page.  I didn't realize there was a task for it

### DIFF
--- a/src/nft/components/collection/CollectionNfts.tsx
+++ b/src/nft/components/collection/CollectionNfts.tsx
@@ -185,7 +185,7 @@ export const CollectionNftsLoading = () => (
 export const CollectionNftsAndMenuLoading = () => (
   <InfiniteScrollWrapper>
     <Column alignItems="flex-start" position="relative" width="full">
-      <Row marginY="12" gap="12">
+      <Row marginY="12" gap="12" marginBottom="40">
         <Box className={loadingAsset} borderRadius="12" width={{ sm: '44', md: '100' }} height="44" />
         <Box
           className={loadingAsset}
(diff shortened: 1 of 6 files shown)


src/pages/App.tsx
```diff
@@ -252,7 +252,14 @@ export default function App() {
 
                   {nftFlag === NftVariant.Enabled && (
                     <>
-                      <Route path="/nfts" element={<NftExplore />} />
+                      <Route
+                        path="/nfts"
+                        element={
+                          <Suspense fallback={null}>
```
aballerr, line 258: I blue circle shows up by default and this removes that blue circle

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -185,7 +185,7 @@ export const CollectionNftsLoading = () => (
 export const CollectionNftsAndMenuLoading = () => (
   <InfiniteScrollWrapper>
     <Column alignItems="flex-start" position="relative" width="full">
-      <Row marginY="12" gap="12">
+      <Row marginY="12" gap="12" marginBottom="40">
```
aballerr, line 188: the space between the nfts and the filter button/ search and sort is a bit off for the loading state.  This makes the transition look more natural

src/pages/App.tsx
cbachmeier: this isn't the correct solution, we should create a suspense fallback for the explore page but this would just be replaced by the Apollo migration and would be throwaway work. Can you add the comment : `// TODO: replace loading state during Apollo migration` so that is tracked

cbachmeier (approved):
LGTM just add that additional context to help with later Apollo migration

PR #5408, opened 2022-11-28
chore: various banner responsive tweaks

grabbou:
Fixes WEB-2303
Fixes WEB-2304
Align banner and trending section with each other (see Slack)

Related:
Added a new `xs` breakpoint since it started appearing in Figma already

### DIFF
--- a/src/nft/components/explore/Banner.tsx
+++ b/src/nft/components/explore/Banner.tsx
@@ -18,9 +18,10 @@ const BannerContainer = styled.div`
   width: 100%;
   padding-top: 32px;
   position: relative;
+  max-width: 1200px;
 
   @media only screen and (min-width: ${({ theme }) => `${theme.breakpoint.sm}px`}) {
-    padding: 32px 16px 0 16px;
+    padding: 32px 16px;
   }
 `
 
--- a/src/nft/components/explore/Banner.tsx
+++ b/src/nft/components/explore/Banner.tsx
@@ -117,7 +117,6 @@ const TRENDING_COLLECTION_SIZE = 5
 
 const Banner = () => {
   const navigate = useNavigate()
-  const isMobile = useIsMobile()
 
   const { data } = useQuery(
     ['trendingCollections'],
--- a/src/nft/components/explore/Banner.tsx
+++ b/src/nft/components/explore/Banner.tsx
@@ -48,10 +48,6 @@ const PlainBackground = styled(AbsoluteFill)`
   background: ${({ theme }) => `linear-gradient(${opacify(10, theme.userThemeColor)}, transparent)`};
 `
 
-const Line = styled.span`
-  word-wrap: keep-all;
-`
-
 const BannerMainArea = styled.div`
   display: flex;
   flex-direction: row;
--- a/src/nft/components/explore/Banner.tsx
+++ b/src/nft/components/explore/Banner.tsx
@@ -1,4 +1,5 @@
 import { useLoadCollectionQuery } from 'graphql/data/nft/Collection'
+import { useIsMobile } from 'nft/hooks'
 import { fetchTrendingCollections } from 'nft/queries'
 import { TimePeriod } from 'nft/types'
 import { calculateCardIndex } from 'nft/utils'
@@ -47,6 +48,10 @@ const PlainBackground = styled(AbsoluteFill)`
   background: ${({ theme }) => `linear-gradient(${opacify(10, theme.userThemeColor)}, transparent)`};
 `
 
+const Line = styled.span`
+  word-wrap: keep-all;
+`
+
 const BannerMainArea = styled.div`
   display: flex;
   flex-direction: row;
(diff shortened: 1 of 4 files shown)


src/nft/components/explore/Banner.tsx
```diff
@@ -155,8 +166,17 @@ const Banner = () => {
...
-          More listings.
+          {isMobile ? (
+            <>
+              Better prices. <br />
+              Every listing.
+            </>
+          ) : (
+            <>
+              Better prices. <br />
+              More listings.
```
cbachmeier, line 177: this was unclear from the ticket, the text should stay the same but on mobile it should be on 2 lines

src/nft/components/explore/Banner.tsx
```diff
@@ -91,12 +90,24 @@ const HeaderContainer = styled.div`
...
+    text-align: center;
+    padding-bottom: 16px;
+
+    br {
+      display: none;
+    }
+  }
+
+  /* Custom breakpoint to split into two lines on smaller screens */
+  @media only screen and (max-width: 550px) {
```
cbachmeier, line 103: can we use a themed breakpoint here

src/nft/components/explore/CarouselCard.tsx
```diff
@@ -295,13 +296,12 @@ const CollectionName = styled(ThemedText.MediumHeader)`
...
 
 const CarouselCardHeader = ({ collection }: { collection: TrendingCollection }) => {
-  const theme = useTheme()
   return (
     <CardHeaderContainer src={collection.bannerImageUrl}>
       <CardHeaderColumn>
         <CollectionImage src={collection.imageUrl} />
         <CollectionNameContainer>
-          <CollectionName color={theme.accentTextLightPrimary} fontWeight="500">
+          <CollectionName color="accentTextLightPrimary" fontWeight="500">
```
cbachmeier, line 304: could you just add these styles to the styled component?

cbachmeier (changes requested):

src/nft/components/explore/Banner.tsx
grabbou: Unfortunately, we don't have a breakpoint for this. It's a one-off specifically calculated at a approximate width where the text would start breaking into two lines

src/nft/components/explore/Banner.tsx
grabbou: Might be worth checking with the product, because Figma does have two different copies on mobile/desktop.

grabbou: I'll follow-up and let you know

src/nft/components/explore/CarouselCard.tsx
grabbou: It's a `ThemedText.MediumHeader`, so I wanted to stick to the way it does colors/fonts until we come up with better abstraction (if any - just have some general thoughts on this component, but it's not related to this PR)

src/nft/components/explore/Banner.tsx
grabbou: For the record, I added `xs` breakpoint as that is actually the one that exists in Figma and is being referenced across different layouts!

src/nft/components/explore/Banner.tsx
JFrankfurt: sounds like a mistake in Figma

JFrankfurt (approved):
approve, but 90% sure you'll need to change that copy so it's not diff on mobile/desktop

src/nft/components/explore/CarouselCard.tsx
```diff
@@ -171,7 +171,8 @@ const CarouselCardContainer = styled.div`
     justify-self: right;
   }
 
-  @media (max-width: 396px) or ((min-width: ${({ theme }) => theme.breakpoint.sm}px) and (max-width: 880px)) {
+  @media (max-width: ${({ theme }) => theme.breakpoint.xs}px) or ((min-width: ${({ theme }) =>
+      theme.breakpoint.sm}px) and (max-width: 880px)) {
```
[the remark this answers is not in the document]
grabbou: it's a backwards-compatible poor man's implementation of container media queries! Basically, we want to hide marketplace name and leave icon only when it gets too tight and `880px` is an arbitrary breakpoint where we can start displaying the name. 

Once container queries are widely supported, we can simply do a query "if container width < 100px, hide the name".

In this PR, I just updated the 396px to an actual `xs` breakpoint as it started appearing more often!

src/nft/components/explore/Banner.tsx
```diff
@@ -117,7 +117,6 @@ const TRENDING_COLLECTION_SIZE = 5
 
 const Banner = () => {
   const navigate = useNavigate()
-  const isMobile = useIsMobile()
```
cbachmeier, line 109: need to remove the import as well

grabbou: rule number one: never do inline Github tweaks lol

grabbou:
@cbachmeier please go ahead and merge this PR once all tests are satisfied, I might be offline by the time that happens!

cbachmeier (approved):
LGTM

PR #5410, opened 2022-11-28
fix: nft explore banner polish

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-2419?atlOrigin=eyJpIjoiZjZjYmIzYzAyMmI0NGNhZDk1MGViYWQ4Y2RmOTQzZjQiLCJwIjoiaiJ9

### DIFF
--- a/src/nft/components/nftExploreBanner/NftExploreBanner.tsx
+++ b/src/nft/components/nftExploreBanner/NftExploreBanner.tsx
@@ -59,14 +59,14 @@ const InnerContainer = styled.div`
   display: flex;
   position: relative;
   gap: 8px;
-  padding: 8px;
+  padding: 12px;
 `
 
 const TextContainer = styled.div`
   display: flex;
   flex-direction: column;
   flex: 1;
-  justify-content: center;
+  justify-content: flex-start;
 `
 
 const StyledXButton = styled(X)`
@@ -80,7 +80,7 @@ const StyledXButton = styled(X)`
 `
 
 const StyledImageContainer = styled(Box)`
-  width: 23%;
+  width: 20%;
   cursor: pointer;
   aspectratio: 1;
   transition: transform 0.25s ease 0s;
@@ -110,7 +110,7 @@ export default function NftExploreBanner() {
           {/* <Description> */}
           <div className={bodySmall}>
             <Trans>Buy and sell NFTs across more listings at better prices.</Trans>{' '}
-            <StyledInternalLink to="/nfts">
+            <StyledInternalLink $uniformColorInLightAndDarkMode to="/nfts">
               <Trans>Explore NFTs</Trans>
             </StyledInternalLink>{' '}
           </div>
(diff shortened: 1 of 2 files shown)


src/theme/components/index.tsx
```diff
@@ -99,9 +99,11 @@ export const ClickableStyle = css`
...
 `
 
-export const LinkStyle = css`
-  color: ${({ theme }) => theme.accentAction};
-  stroke: ${({ theme }) => theme.accentAction};
+export const LinkStyle = css<{ $uniformColorInLightAndDarkMode?: boolean }>`
+  color: ${({ theme, $uniformColorInLightAndDarkMode }) =>
+    $uniformColorInLightAndDarkMode ? theme.accentActive : theme.accentAction};
+  stroke: ${({ theme, $uniformColorInLightAndDarkMode }) =>
+    $uniformColorInLightAndDarkMode ? theme.accentActive : theme.accentAction};
```
grabbou, line 106: Not sure about this one, to be honest! I am quite confused if this isn't just an NFT specific exception. Anyway, for my PR, I did this: https://github.com/Uniswap/interface/pull/5417

grabbou: I think going forward we need to develop much better abstraction for this, which is similar to what we have to do for Text components as well! 

Meanwhile, I think this is better while handled on a component-level, so that we don't introduce a new API that can dangerously lead to our links design getting inconsistent!

grabbou (changes requested):
Thanks for the PR! I posted some questions to clarify the intents of the design team behind links in the app in general, which can influence further direction for this PR! Meanwhile, my recommendation is to keep things simple and try styling components locally (aka overriding ExternalLink style with a Styled component) to avoid increasing the API layer.

src/theme/components/index.tsx
lynnshaoyu: done! thanks!

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
```diff
@@ -80,13 +81,23 @@ const StyledXButton = styled(X)`
...
   transition: transform 0.25s ease 0s;
   object-fit: contain;
 `
 
+const LinkStyle = css`
+  color: ${({ theme }) => theme.accentActive};
+  stroke: ${({ theme }) => theme.accentActive};
+`
+
+const StyledLink = styled(Link)`
```
grabbou, line 96: nit: you could just use a `styled(StyledLink)` and simply define a custom color/stroke for it!

grabbou (approved):
LGTM! 🔥 Just left a small nit, but you can consider it non-blocking and totally up to you! It's a styling preference that I have applied in my other PR, but both solutions do the trick!

src/nft/components/nftExploreBanner/NftExploreBanner.tsx
lynnshaoyu: not sure what "StyledLink" you are referring to? Also, we need ClickableStyle as well, not just the custom colors.

PR #5412, opened 2022-11-28
feat: increase hit-target for remove button

grabbou:
Fixes WEB-2421

Screenshot of a hit target in ^ Jira ticket

### DIFF
--- a/src/nft/components/bag/BagRow.tsx
+++ b/src/nft/components/bag/BagRow.tsx
@@ -39,6 +39,26 @@ const ReviewButton = styled(ThemeButton)`
   padding: 8px;
   width: 50%;
 `
+const RemoveAssetOverlay = styled.div`
+  position: absolute;
+  display: block;
+  right: 0px;
+  top: 0px;
+  transform: translate(30%, -30%);
+  z-index: 1;
+  transition: 250ms;
+  width: calc(100%);
+  height: calc(100%);
+  display: flex;
+  align-items: center;
+  justify-content: center;
+`
+
+export const RemoveAssetButton = ({ onClick }: { onClick: (e: MouseEvent<HTMLDivElement>) => void }) => (
+  <RemoveAssetOverlay onClick={onClick}>
+    <CircularCloseIcon />
+  </RemoveAssetOverlay>
+)
 
 const NoContentContainer = () => (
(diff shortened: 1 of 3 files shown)


grabbou:
I am going to update this PR from a hardcoded value bound to the NFT image, to a more generic solution with a 44px hit target unified across the

src/nft/components/bag/BagRow.tsx
```diff
@@ -39,6 +39,25 @@ const ReviewButton = styled(ThemeButton)`
   padding: 8px;
   width: 50%;
 `
+const RemoveAssetOverlay = styled.div`
```
cbachmeier, line 42: Can you give a demo build of this to product/requester, seems a little to large and currently I can click outside of the card to remove it which doesn't seem to be expected

grabbou: Interesting - what do you mean by "outside of the card"? 

As far as the preview, I've pasted a screenshot of a clickable area on Jira (feel free to check the ticket for reference). I'll ask product to double-check on it too.  

I think the size requirement is largely influenced by Apple:
> Apple’s [iPhone Human Interface Guidelines](https://developer.apple.com/iphone/library/documentation/UserExperience/Conceptual/MobileHIG/DesigningNativeApp/DesigningNativeApp.html#//apple_ref/doc/uid/TP40006556-CH4-SW1) recommends a minimum target size of 44 pixels wide 44 pixels tall. Microsoft’s Windows Phone UI Design and Interaction Guide suggests a touch target size of 34px with a minimum touch target size of 26px. Nokia’s developer guidelines suggest that the target size should be no smaller than 1cm x 1cm square or 28 x 28 pixels.

cbachmeier: I attached a screenshot in the thread but basically part of the click area is outside of the image

cbachmeier (approved):
LGTM if product approves

src/nft/components/bag/BagRow.tsx
grabbou: Yeah, let's follow up internally and see where we land.

grabbou:
I am going to go ahead and merge it, Scott confirm this is working as expected. Feel free to revert in case decision changes over night.

PR #5414, opened 2022-11-28
fix(banner): fix 1px non-overlay gap

JFrankfurt:
now (seeing on firefox/windows): 
image

prior:
image

### DIFF
--- a/src/nft/components/explore/CarouselCard.tsx
+++ b/src/nft/components/explore/CarouselCard.tsx
@@ -104,6 +104,7 @@ const HeaderOverlay = styled.div`
   top: 0px;
   z-index: 0;
 
+  // overlay workaround for firefox only
   @supports (-moz-appearance: none) {
     bottom: -1px;
   }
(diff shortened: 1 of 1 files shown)


src/nft/components/explore/CarouselCard.tsx
```diff
@@ -97,7 +97,7 @@ const LoadingCollectionNameContainer = styled(LoadingBubble)`
 
 const HeaderOverlay = styled.div`
   position: absolute;
-  bottom: 0px;
+  bottom: -1px;
```
grabbou, line 100: interesting 🫣

grabbou:
AFAIK this is Firefox specific rendering glitch, right? In that case, would it make sense to use something like
```css
@supports (-moz-appearance:none){
 
}
```
to make this hack run on Firefox only?

grabbou (approved):
LGTM! It's a nit, so feel free to ignore and just merge! I think it would be great to document this by adding a comment - this is a workaround for Firefox only, just to avoid someone getting confused in the future!

PR #5415, opened 2022-11-28
fix: add outline to bottom navigation

grabbou:
Fixes WEB-2420

### DIFF
--- a/src/components/NavBar/index.tsx
+++ b/src/components/NavBar/index.tsx
@@ -9,12 +9,30 @@ import { Row } from 'nft/components/Flex'
 import { UniIcon } from 'nft/components/icons'
 import { ReactNode } from 'react'
 import { NavLink, NavLinkProps, useLocation } from 'react-router-dom'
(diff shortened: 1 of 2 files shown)


src/components/NavBar/index.tsx
```diff
@@ -9,13 +9,31 @@ import { Row } from 'nft/components/Flex'
...
 import * as styles from './style.css'
 
+const MobileBottomBar = styled.div`
+  position: fixed;
+  display: flex;
+  bottom: 0;
+  right: 0;
+  left: 0;
+  justify-content: space-between;
+  padding: 4px 8px 0;
```
cbachmeier, line 27: missing bottom padding, I believe should be just `4px 8px`

cbachmeier (approved):
LGTM, minor nit

unipadmini (approved):

src/components/NavBar/index.tsx
grabbou: good catch!

PR #5420, opened 2022-11-28
fix: use accentAction for VerifiedIcon

vm:
.


src/nft/components/icons.tsx
```diff
@@ -667,7 +670,7 @@ export const ActivityTransferIcon = (props: SVGProps) => (
   </svg>
 )
 
-export const ActivityExternalLinkIcon = (props: SVGProps) => (
+export const ActivityExternalLinkIcon = (_props: SVGProps) => (
```
grabbou, line 673: nit: can we just remove it?

src/nft/components/icons.tsx
```diff
@@ -59,18 +59,21 @@ export const WarningIcon = (props: SVGProps) => (
...
+        fill={theme.accentAction}
```
grabbou, line 68: Are we okay with this change? 

Looking at the codebase, `<VerifiedIcon />` is being used in ~9 places, only two of them used to define a `fill` property. 

After this change, all call sites will get a Verified Icon filled with `accentAction`, so it is going to have much wider impact than just these two components.

Should that be fine, you can consider this PR approved! Otherwise, I suggest keeping `fill` as a boolean and when present, setting it to `theme.accentAction`.

grabbou (approved):
Approved tentatively! Details in comments, feel free to resolve on your own and merge!

src/nft/components/icons.tsx
vm: I did check all of them, so no problem there! I think we should never have a verified icon that is not based on the accentAction color but if that changes we should make an update.

src/nft/components/icons.tsx
vm: i just wanted to keep the pattern so someone knows they _could_ use the props in the future

PR #5422, opened 2022-11-28
refactor: connect to prod gql and api endpoint

cbachmeier:
Use prod endpoints for all NFT queries

### DIFF
--- a/src/graphql/data/fetchGraphQL.ts
+++ b/src/graphql/data/fetchGraphQL.ts
@@ -33,11 +33,10 @@ const fetchQuery = (params: RequestParameters, variables: Variables): Promise<Gr
     query: params.text, // GraphQL text from input
     variables,
   })
-  const url = GRAPHQL_URL
   // TODO: When gating is removed from gql endpoint, we can remove the isNFT check and just use base headers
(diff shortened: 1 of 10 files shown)


vm (approved):

src/graphql/data/fetchGraphQL.ts
```diff
@@ -35,10 +33,10 @@ const fetchQuery = (params: RequestParameters, variables: Variables): Promise<Gr
     query: params.text, // GraphQL text from input
     variables,
   })
-  const url = isNFT ? NFT_URL : TOKEN_URL
+  // TODO: When gating is removed from gql endpoint, we can remove the isNFT check and just use base headers
```
grabbou, line 36: Awesome! I think we can also simplify our env management/guarding logic, unless that PR wasn't merged yet (tbh, I don't remember 😅)

grabbou (approved):
Changes look legit, but I have some questions as far as API layer in general. Definitely non blocking 👋

src/graphql/data/fetchGraphQL.ts
cbachmeier: That PR hasn't merged yet. That'll be the last one before launch

PR #5423, opened 2022-11-28
fix(cypress): patch to add another parallelization and make timeout longer than block time

JFrankfurt:
running into lots of timeouts with tests that do testnet interactions

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -3,7 +3,7 @@ import { defineConfig } from 'cypress'
 export default defineConfig({
   projectId: 'yp82ef',
   videoUploadOnPasses: false,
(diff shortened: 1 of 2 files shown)


cypress.config.ts
```diff
@@ -3,7 +3,7 @@ import { defineConfig } from 'cypress'
 export default defineConfig({
   projectId: 'yp82ef',
   videoUploadOnPasses: false,
-  defaultCommandTimeout: 10000,
+  defaultCommandTimeout: 24000, // 2x average block time
```
zzmp, line 6: ty for noting _why_ you've chosen this timeout time

zzmp (approved):

PR #5424, opened 2022-11-28
fix: sweep adding the correct assets from amms

JackShort:
https://uniswaplabs.atlassian.net/browse/WEB-2438

### DIFF
--- a/src/nft/components/collection/Sweep.tsx
+++ b/src/nft/components/collection/Sweep.tsx
@@ -197,7 +197,7 @@ export const Sweep = ({ contractAddress, minPrice, maxPrice }: SweepProps) => {
     if (nft20Assets) jointCollections = [...jointCollections, ...nft20Assets]
 
     const sudoSwapAssetsInJointCollections = jointCollections.filter(
-      (sweepAsset) => sweepAsset.marketplace === Markets.Sudoswap
+      (sweepAsset) => sweepAsset.marketplace === Markets.Sudoswap && !sweepAsset.susFlag
     )
 
     jointCollections.forEach((asset) => {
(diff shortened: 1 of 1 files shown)


src/nft/components/collection/Sweep.tsx
```diff
@@ -197,7 +197,7 @@ export const Sweep = ({ contractAddress, minPrice, maxPrice }: SweepProps) => {
     if (nft20Assets) jointCollections = [...jointCollections, ...nft20Assets]
 
     const sudoSwapAssetsInJointCollections = jointCollections.filter(
-      (sweepAsset) => sweepAsset.marketplace === Markets.Sudoswap
+      (sweepAsset) => sweepAsset.marketplace === Markets.Sudoswap && !sweepAsset.susFlag
```
grabbou, line 200: newbie question: what is a `susFlag`?

src/nft/components/collection/Sweep.tsx
```diff
@@ -215,8 +215,8 @@ export const Sweep = ({ contractAddress, minPrice, maxPrice }: SweepProps) => {
           const poolPrice = calcPoolPrice(asset, isNFTX ? counterNFTX : counterNFT20)
           if (BigNumber.from(poolPrice).gt(0)) {
             isNFTX ? counterNFTX++ : counterNFT20++
-            asset.priceInfo.ETHPrice = poolPrice
           }
+          asset.priceInfo.ETHPrice = poolPrice
```
grabbou, line 219: so, previously, the price would be set to pool price only if its value was greater than zero, right? Now, it will be set regardless of the value

JackShort: yes

JackShort: no because the ones that are not sus are the only ones that are selectable and thus should be counted

PR #5425, opened 2022-11-28
fix: verified icon sizes and color are messed up in view my nfts

lynnshaoyu:


src/nft/components/collection/Card.tsx
```diff
@@ -517,7 +517,7 @@ const ProfileNftDetails = ({ asset, hideDetails }: ProfileNftDetailsProps) => {
           <TruncatedTextRow className={bodySmall} style={{ color: themeVars.colors.textSecondary }}>
             {!!asset.asset_contract.name && <span>{asset.asset_contract.name}</span>}
           </TruncatedTextRow>
-          {asset.collectionIsVerified && <VerifiedIcon height="16px" width="16px" fill={colors.magentaVibrant} />}
+          {asset.collectionIsVerified && <VerifiedIcon height="18px" width="18px" />}
```
grabbou, line 520: So, increasing width/height but setting `flex` to 98% seems to do the trick! Generally, I had bad experience styling SVGs in our codebase, often source of misalignments. I wonder if there's a less magical way 🤔

lynnshaoyu: im confused by this comment, i think i'm doing exactly what you're describing already?😅 if it looks good can you approve? :P

aballerr (approved):

PR #5426, opened 2022-11-29
feat: make bag more distinctive from the nav bar 

grabbou:
Fixes WEB-2458

Also, fixed one more thing not specced out in the ticket, which is bag on the explore page not playing really well with a transparent navbar.


src/nft/components/bag/Bag.tsx
```diff
@@ -336,4 +338,25 @@ const Bag = () => {
...
+  top: 0px;
+  width: 100%;
+  height: 100%;
+  right: 0px;
+  background: ${({ theme }) => theme.backgroundSurface};
+  border-left-style: solid;
+  border-color: ${({ theme }) => theme.backgroundOutline};
+  border-width: 1px;
+  color: ${({ theme }) => theme.textPrimary};
+  z-index: ${Z_INDEX.dropdown};
```
grabbou, line 354: Need extra attention on this one! Previously, with Sprinkles, this Z-INDEX was set to `modalOverTooltip`. That means we would have to create a new `Z_INDEX` value of `1090` to render over a tooltip.

However, I don't think that's expected - it's a bag and I think it should render more or less on the same level as the nav bar. When there's a popup or a tooltip (especially rendered in the bottom right corner), I don't think it should render below the bag. It's potentially dangerous (see screenshots in Jira).

I used a `dropdown` value instead. Please note that this is lower value than z-index of the navbar. Thanks to that, we can actually provide a background for a transparent navbar on the NFT Explore page (see screenshots in Jira for more context).

src/pages/App.tsx
```diff
@@ -179,16 +173,28 @@ export default function App() {
   }, [isExpertMode])
 
   useEffect(() => {
+    const scrollListener = (e: Event) => {
+      if (window.scrollY > 0) {
+        setScrolledState(true)
+      } else {
+        setScrolledState(false)
```
aballerr, line 180: this can just be ```setScrolledState(window.scrollY > 0)``` I think

aballerr (approved):

PR #5427, opened 2022-11-29
fix: style on transaction modal

grabbou:
Fixes WEB-2434

If anyone wants to help with testing and perform a real transaction, that would be much better! For testing, I mocked few booleans to render a preview, but was missing some content.

### DIFF
--- a/src/nft/themes/darkTheme.ts
+++ b/src/nft/themes/darkTheme.ts
@@ -29,6 +29,8 @@ export const darkTheme: Theme = {
     textPrimary: '#FFFFFF',
     textSecondary: vars.color.gray300,
     textTertiary: vars.color.gray500,
+
+    dropShadow: `0px 4px 16px rgba(76, 130, 251, 0.4)`,
   },
   shadows: {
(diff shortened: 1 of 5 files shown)


src/nft/themes/darkTheme.ts
```diff
@@ -29,6 +29,8 @@ export const darkTheme: Theme = {
     textPrimary: '#FFFFFF',
     textSecondary: vars.color.gray300,
     textTertiary: vars.color.gray500,
+
+    dropShadow: `0px 4px 16px rgba(76, 130, 251, 0.4)`,
```
grabbou, line 33: Purposefully sticked to Sprinkles in this component to reduce number of components affected by this PR. This is definitely not the easiest part of the code to test.

Also, Sprinkles doesn't make it easier by passing around `vars(--name)` instead of an actual HEX color. Since it is not possible yet to add opacity to a color that comes from a variable (yet, spec is already submitted), I decided to copy/paste accentAction with appropriate opacity for time being, until we migrate off Sprinkles.

PR #5428, opened 2022-11-29
fix: use same loader in view my nfts as collection

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-2383?atlOrigin=eyJpIjoiM2E0ZDAyNDUwZDcyNGU0ZGI2NGM3YjI0YTliZjNmMDMiLCJwIjoiaiJ9

### DIFF
--- a/src/nft/components/collection/CollectionNfts.tsx
+++ b/src/nft/components/collection/CollectionNfts.tsx
@@ -168,17 +168,17 @@ const MarketNameWrapper = styled(Row)`
   gap: 8px;
 `
 
-const LoadingAssets = () => (
+const LoadingAssets = ({ count }: { count?: number }) => (
   <>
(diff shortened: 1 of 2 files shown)


aballerr (changes requested):
If you look at the loading state on a collection, you will notice that the loading state fills the remaining empty spaces.  Currently the loading state on view leaves empty spaces that we need to fill as well.  Also, the loading state needs some margin top.  Added a screen shot to the jira task that shows what I'm talking about

lynnshaoyu:
> If you look at the loading state on a collection, you will notice that the loading state fills the remaining empty spaces. Currently the loading state on view leaves empty spaces that we need to fill as well. Also, the loading state needs some margin top. Added a screen shot to the jira task that shows what I'm talking about

thank you for catching that. fixed in the latest commit.

aballerr (approved):

PR #5429, opened 2022-11-29
fix: keep bag closed after manual close

vm:
This makes it so the bagManuallyClosed state doesn't reset when it goes to 0 items in the bag.


JackShort (approved):
lgtm

PR #5436, opened 2022-11-29
feat: upgrade widget to 2.20.0

cartcrom:


package.json
```diff
@@ -151,7 +151,7 @@
     "@uniswap/v3-core": "1.0.0",
     "@uniswap/v3-periphery": "^1.1.1",
     "@uniswap/v3-sdk": "^3.9.0",
-    "@uniswap/widgets": "^2.19.2",
+    "@uniswap/widgets": "2.20.0",
```
zzmp, line 154: NB: you've transitioned this from a `^` to a pinned version, which is probably better, tbh. We treat all of these like pinned versions because of the `yarn.lock`. I think this is a good change - eg mobile explicitly pins all of their deps for security purposes.

zzmp (approved):

PR #5439, opened 2022-11-29
fix: reverting to previous styles

aballerr:
this commit added some regressions: https://github.com/Uniswap/interface/commit/af2725ec25ad8c4660b00e721ef7da192801baac

The new BagContainer causes a couple of issues

1. adding a bunch of items causes the "pay" button go off screen
2. The overlay when the bag opens does not always function properly


cbachmeier (approved):
LGTM

PR #5442, opened 2022-11-29
chore: remove /swap/:outputCurrency redirect

vm:
I'd like to start simplifying the swap/home redirect logic as we start working on a landing page, and this is the first step. I think this is reasonable enough to remove.


JFrankfurt:
It looks like @moodysalem added this to solve an invalid output problem that it looks like is no longer happening.

JFrankfurt:
see https://github.com/Uniswap/interface/commit/1aea452f3b9f99d8bc0f3f95fdbbef3964adfefe

JFrankfurt (approved):

PR #5443, opened 2022-11-29
chore: remove react-confetti

vm:
See how this looks because react-confetti is a large dependency that we want to remove.


grabbou (approved):
Will be missing confetti tho 😅

cartcrom (approved):

PR #5445, opened 2022-11-29
chore: remove /#/claim

vm:
By this point we can assume that there aren't many active links to the /claim url. Instead, users could just claim from the menu.

### DIFF
--- a/src/pages/Swap/redirects.tsx
+++ b/src/pages/Swap/redirects.tsx
@@ -1,19 +1,7 @@
-import { useEffect } from 'react'
 import { Navigate, useLocation } from 'react-router-dom'
-import { useAppDispatch } from 'state/hooks'
-
-import { ApplicationModal, setOpenModal } from '../../state/application/reducer'
 
 // Redirects to swap but only replace the pathname
 export function RedirectPathToSwapOnly() {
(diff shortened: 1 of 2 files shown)


src/pages/Swap/redirects.tsx
```diff
@@ -30,11 +26,3 @@ export function RedirectToSwap() {
     />
   )
 }
-
-export function OpenClaimAddressModalAndRedirectToSwap() {
-  const dispatch = useAppDispatch()
-  useEffect(() => {
-    dispatch(setOpenModal(ApplicationModal.ADDRESS_CLAIM))
```
zzmp, line 37: Can't we remove a lot of other code that this calls into? Does this PR render that dead code?

zzmp (changes requested):

src/pages/Swap/redirects.tsx
vm: no i was gonna leave all the other places that can open the claim modal for now until product decides we can remove

zzmp (approved):
Should be merged after a 404 page exists, so that users navigating to /claim aren't redirected back to swap with no acknowledgment.

PR #5447, opened 2022-11-29
feat: remove gated key for graphql query

cbachmeier:
Api key gate is no longer needed for nft graphql queries. Approved by product. Tested and got data locally.


JackShort (approved):
lets gooooooo

mz- (approved):

PR #5450, opened 2022-11-29
fix: collection names truncating early

cbachmeier:
Collection names on profile cards were truncating early


lynnshaoyu:
pls don't approve this (anyone who's looking) i need to submit a fix to account for view my nfts side effects

lynnshaoyu (approved):

PR #5464, opened 2022-11-30
fix: lisiting modal

cbachmeier:
Listing modal needs the bag styles to properly size

This is a quick fix for expediency


JackShort (approved):
lgtm

PR #5484, opened 2022-11-30
chore: remove nft feature flag

vm:
Now that it's ungated we can remove the feature flag fully.

### DIFF
--- a/src/components/NavBar/SearchBar.tsx
+++ b/src/components/NavBar/SearchBar.tsx
@@ -3,7 +3,6 @@ import { t } from '@lingui/macro'
 import { sendAnalyticsEvent, Trace, TraceEvent, useTrace } from '@uniswap/analytics'
 import { BrowserEvent, ElementName, EventName, SectionName } from '@uniswap/analytics-events'
 import clsx from 'clsx'
-import { NftVariant, useNftFlag } from 'featureFlags/flags/nft'
 import useDebounce from 'hooks/useDebounce'
 import { useIsNftPage } from 'hooks/useIsNftPage'
 import { useOnClickOutside } from 'hooks/useOnClickOutside'
@@ -47,10 +46,8 @@ export const SearchBar = () => {
   const searchRef = useRef<HTMLDivElement>(null)
   const inputRef = useRef<HTMLInputElement>(null)
   const { pathname } = useLocation()
-  const phase1Flag = useNftFlag()
   const isMobile = useIsMobile()
   const isTablet = useIsTablet()
-  const isPhase1 = phase1Flag === NftVariant.Enabled
 
   useOnClickOutside(searchRef, () => {
     isOpen && toggleOpen()
@@ -63,7 +60,7 @@ export const SearchBar = () => {
       refetchOnWindowFocus: false,
       refetchOnMount: false,
       refetchOnReconnect: false,
-      enabled: !!debouncedSearchValue.length && isPhase1,
+      enabled: !!debouncedSearchValue.length,
     }
   )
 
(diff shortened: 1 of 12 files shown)


src/components/WalletModal/index.tsx
```diff
@@ -314,31 +309,6 @@ export default function WalletModal({
...
-          <ExternalLink href="https://uniswap.org/terms-of-service/">Terms of Service</ExternalLink> and consent to its{' '}
-          <ExternalLink href="https://uniswap.org/privacy-policy">Privacy Policy</ExternalLink>.
-        </Trans>
-      )
-      return nftFlagEnabled ? (
-        <AutoRow style={{ flexWrap: 'nowrap', padding: '4px 16px' }}>
-          <ThemedText.BodySecondary fontSize={16} lineHeight="24px">
-            {content}
-          </ThemedText.BodySecondary>
-        </AutoRow>
```
cbachmeier, line 332: I would appreciate @aballerr input here, but it seems like this section is still to be included

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -343,7 +337,7 @@ export const SearchBarDropdown = ({
   ])
 
   return (
-    <Box className={isPhase1 ? styles.searchBarDropdownNft : styles.searchBarDropdown}>
```
cbachmeier, line 346: can you remove this no longer used style?

src/components/NavBar/SearchBar.tsx
```diff
@@ -199,15 +190,13 @@ export const SearchBar = () => {
                 setSearchValue(event.target.value)
               }}
               onBlur={() => sendAnalyticsEvent(EventName.NAVBAR_SEARCH_EXITED, navbarSearchEventProperties)}
-              className={`${styles.searchBarInput} ${
-                showCenteredSearchContent ? styles.searchContentCentered : styles.searchContentLeftAlign
```
cbachmeier, line 203: can you remove this style if it is no longer used

src/components/NavBar/SearchBar.tsx
```diff
@@ -160,12 +151,12 @@ export const SearchBar = () => {
...
         width={{ sm: isOpen ? 'viewWidth' : 'auto', md: 'auto' }}
         ref={searchRef}
-        className={isPhase1 ? styles.searchBarContainerNft : styles.searchBarContainer}
+        className={styles.searchBarContainerNft}
         display={{ sm: isOpen ? 'inline-block' : 'none', xl: 'inline-block' }}
       >
         <Row
           className={clsx(
-            ` ${isPhase1 ? styles.nftSearchBar : styles.searchBar} ${!isOpen && !isMobile && magicalGradientOnHover} ${
+            ` ${styles.nftSearchBar} ${!isOpen && !isMobile && magicalGradientOnHover} ${
```
cbachmeier, line 159: can you remove these no longer used styles?

cbachmeier (commented):
I think there's one section of Web3Status that isn't supposed to be removed but other than that only minor non-blocking nits

src/components/WalletModal/index.tsx
aballerr: I tracked down the original pr, it looks like Charlie is right and this should stay.  

https://github.com/Uniswap/interface/pull/4790

vm: will add back

src/components/NavBar/SearchBarDropdown.tsx
vm: yes

cbachmeier (approved):
LGTM

PR #5485, opened 2022-11-30
fix: add retries to relay network layer

cartcrom:


src/graphql/data/RelayEnvironment.ts
```diff
@@ -1,16 +1,38 @@
...
+}
+
+const network = new RelayNetworkLayer([
+  urlMiddleware({
+    url: GRAPHQL_URL,
+    headers: {
+      'Content-Type': 'application/json',
+    },
+  }),
+  cacheMiddleware({
```
zzmp, line 23: I don't think we want this additional cache over the store

zzmp (changes requested):

src/graphql/data/RelayEnvironment.ts
```diff
@@ -1,16 +1,38 @@
...
 const gcReleaseBufferSize = 250
 const queryCacheExpirationTime = ms`5m`
 
+const GRAPHQL_URL = process.env.REACT_APP_AWS_API_ENDPOINT
+if (!GRAPHQL_URL) {
+  throw new Error('AWS URL MISSING FROM ENVIRONMENT')
+}
+
+const MAX_RETRIES = 10
+const network = new RelayNetworkLayer([
```
zzmp, line 17: Can you add a note that the network layer must not cache, or it will break cache-evicting network policies (like network-only)?

Also check that this new class does not cache by default.

zzmp (changes requested):

src/graphql/data/RelayEnvironment.ts
```diff
@@ -1,16 +1,38 @@
 import ms from 'ms.macro'
-import { Environment, Network, RecordSource, Store } from 'relay-runtime'
-
-import fetchGraphQL from './fetchGraphQL'
```
zzmp, line 4: Is this file still used? I thought it had other semantics that aren't duplicated here.

src/graphql/data/RelayEnvironment.ts
```diff
@@ -1,16 +1,39 @@
...
+      'Content-Type': 'application/json',
+    },
+  }),
+  retryMiddleware({
+    fetchTimeout: ms`15s`,
+    retryDelays: (attempt) => Math.pow(2, attempt + 4) * 100, // or simple array [3200, 6400, 12800],
+    statusCodes: [500, 503, 504],
+    beforeRetry: ({ abort, attempt }) => {
+      if (attempt > MAX_RETRIES) abort()
+      console.log(attempt)
```
vm, line 30: rm

src/graphql/data/RelayEnvironment.ts
```diff
@@ -1,16 +1,39 @@
...
+const network = new RelayNetworkLayer([
+  urlMiddleware({
+    url: GRAPHQL_URL,
+    headers: {
+      'Content-Type': 'application/json',
+    },
+  }),
+  retryMiddleware({
+    fetchTimeout: ms`15s`,
+    retryDelays: (attempt) => Math.pow(2, attempt + 4) * 100, // or simple array [3200, 6400, 12800],
```
vm, line 26: make a function called `getDelayWithBackoff`

src/graphql/data/RelayEnvironment.ts
```diff
@@ -1,16 +1,39 @@
...
+    statusCodes: [500, 503, 504],
```
vm, line 27: can we do any 5xx error or need to list them out manually?

src/graphql/data/RelayEnvironment.ts
cartcrom: the new library does all of this for us. This is what fetchGraphql was:

```
const fetchQuery = (params: RequestParameters, variables: Variables): Promise<GraphQLResponse> => {
  const body = JSON.stringify({
    variables,
  })

  return fetch(GRAPHQL_URL, { method: 'POST', body, headers })
    .then((res) => res.json())
    .catch((e) => {
      console.error(e)
      return { data: [] }
    })
}
```

src/graphql/data/RelayEnvironment.ts
```diff
@@ -1,16 +1,44 @@
...
+    url: GRAPHQL_URL,
+    headers: {
+      'Content-Type': 'application/json',
+    },
+  }),
+  retryMiddleware({
+    fetchTimeout: ms`30s`,
+    retryDelays: getDelayWithBackoff,
+    statusCodes: [500, 503, 504],
+    beforeRetry: ({ abort, attempt }) => {
```
zzmp, line 34: Either this should be its own fn, or getBackoff should be inlined. They should be consistent.

src/graphql/data/RelayEnvironment.ts
vm: big

src/graphql/data/RelayEnvironment.ts
cartcrom: probably has to be manual

src/graphql/data/RelayEnvironment.ts
zzmp: Can you delete the old file?

src/graphql/data/RelayEnvironment.ts
```diff
@@ -1,16 +1,44 @@
...
+// This network layer must not cache, or it will break cache-evicting network policies
+const network = new RelayNetworkLayer([
+  urlMiddleware({
+    url: GRAPHQL_URL,
+    headers: {
+      'Content-Type': 'application/json',
+    },
+  }),
+  retryMiddleware({
+    fetchTimeout: ms`30s`,
```
vm, line 31: comment here to say this mirrors the backend's timeout, incase that fails.

vm (dismissed):

src/graphql/data/RelayEnvironment.ts
cartcrom: It already was. Might've accidentally added it back in between pushes maybe

yarn.lock
```diff
@@ -7140,6 +7140,11 @@ core-js@^2.4.0:
   resolved "https://registry.npmjs.org/core-js/-/core-js-2.6.12.tgz"
   integrity sha512-Kb2wC0fvsWfQrgk8HU5lW6U/Lcs8+9aaYcy4ZFc6DDlo4nZ7n70dEgE5rtR0oG6ufKDUnrwfWL1mXR5ljDatrQ==
 
+core-js@^3.0.0:
```
zzmp: This should coalesce with 3.6.5

zzmp (changes requested):

zzmp (dismissed):

src/graphql/data/fetchGraphQL.ts
```diff
@@ -1,28 +0,0 @@
...
-  const body = JSON.stringify({
-    query: params.text, // GraphQL text from input
-    variables,
-  })
-
-  return fetch(GRAPHQL_URL, { method: 'POST', body, headers })
-    .then((res) => res.json())
-    .catch((e) => {
-      console.error(e)
-      return { data: [] }
```
zzmp, line 24: Are you sure this "catch and return `{data:[]}`" behavior is done by the new network class?

cartcrom: I don't think it is, looking into how to recreate

zzmp (commented):
unapproving pending handling errors

src/graphql/data/RelayEnvironment.ts
```diff
@@ -25,6 +32,14 @@ function beforeRetry({ abort, attempt }: { abort: AbortFn; attempt: number }) {
   if (attempt > MAX_RETRIES) abort()
 }
 
+function customErrorRemovalMiddleware(next: MiddlewareNextFn) {
+  return async (req: RelayRequestAny) => {
+    const res = await next(req)
+    res.errors = undefined
+    return res
```
zzmp, line 39: will this return `{data: []}`

src/graphql/data/RelayEnvironment.ts
```diff
@@ -35,7 +35,7 @@ function beforeRetry({ abort, attempt }: { abort: AbortFn; attempt: number }) {
 function customErrorRemovalMiddleware(next: MiddlewareNextFn) {
   return async (req: RelayRequestAny) => {
     const res = await next(req)
-    res.errors = undefined
+    res.errors = { data: [] }
```
zzmp, line 38: errors should not be data, the returned result should be data.

zzmp (changes requested):

src/graphql/data/RelayEnvironment.ts
```diff
@@ -1,16 +1,67 @@
...
+  return statusCode >= 500 && statusCode < 600
+}
+
+function beforeRetry({ abort, attempt }: { abort: AbortFn; attempt: number }) {
+  if (attempt > MAX_RETRIES) abort()
+}
+
+function customErrorRemovalMiddleware(next: MiddlewareNextFn) {
+  return async (req: RelayRequestAny) => {
+    const res = await next(req)
```
zzmp, line 37: Shouldn't this only occur if there _is_ an error? Should the whole thing be in a try-catch? Or should we check for res.errors?

zzmp (approved):

PR #5487, opened 2022-11-30
fix: allow 0 after decimal

cbachmeier:
When a user typed `.` followed by `0` into the listing price input, the decimal would be cleared, this fixes that bug


JackShort (approved):
lgtm

PR #5498, opened 2022-12-01
feat: show 1155s in collection filter

cbachmeier:
Removes the filter for 1155 collections in the filter menu on profile. Allows them to be searchable as well. Can test by impersonating my wallet.

Before:
Screen Shot 2022-12-01 at 09 31 53

After:
Screen Shot 2022-12-01 at 09 31 46
Screen Shot 2022-12-01 at 09 32 04

### DIFF
--- a/src/nft/queries/openSea/OSCollectionsFetcher.ts
+++ b/src/nft/queries/openSea/OSCollectionsFetcher.ts
@@ -13,12 +13,14 @@ export const OSCollectionsFetcher = async ({ params }: any): Promise<WalletColle
   const r = await fetch(`https://api.opensea.io/api/v1/collections?${new URLSearchParams(params).toString()}`)
   const walletCollections = await r.json()
   if (walletCollections) {
-    return walletCollections.map((collection: any) => ({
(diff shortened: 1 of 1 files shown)


src/nft/queries/openSea/OSCollectionsFetcher.ts
```diff
@@ -13,17 +13,12 @@ export const OSCollectionsFetcher = async ({ params }: any): Promise<WalletColle
   const r = await fetch(`https://api.opensea.io/api/v1/collections?${new URLSearchParams(params).toString()}`)
   const walletCollections = await r.json()
   if (walletCollections) {
-    return walletCollections
-      .filter(
-        (collection: any) =>
-          collection.primary_asset_contracts.length && collection.primary_asset_contracts[0].schema_name === 'ERC721'
```
lynnshaoyu, line 19: should we keep the filter on collection.primary_asset_contracts.length so we make sure its not undefined?

cbachmeier: Great catch, we do still want this

PR #5499, opened 2022-12-01
feat: add mobile profile menu filter header

cbachmeier:
Add's Mobile Filter Menu Header with the text `Filter` and the `X` icon moved to the top right. Previously the X Mark icon did not properly set the fill color, this fixes that bug.

https://uniswaplabs.atlassian.net/browse/WEB-2400

Before:
Screen Shot 2022-12-01 at 09 59 38
After:
Screen Shot 2022-12-01 at 09 59 17


JackShort (approved):
lgtm - do we not want to add a click state to change the opacity? Unsure of what design expects for mobile buttons

PR #5501, opened 2022-12-01
fix: replace many blue instances with accentAction

cbachmeier:
A lot of places used blue400/genieBlue where instead it should be using accentAction. This has been updated and no longer used styles have been removed.


bryankuuu:
LGTM! thanks charlie

JackShort (approved):

PR #5507, opened 2022-12-01
fix: disabled treatment for unlisted collection page card + tooltip styling fixes

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-2386
https://uniswaplabs.atlassian.net/browse/WEB-2387


src/nft/components/collection/CollectionAsset.tsx
```diff
@@ -145,13 +146,28 @@ export const CollectionAsset = ({
...
           />
         )}
-        {assetMediaType === AssetMediaType.Image ? (
-          <Card.Image />
-        ) : assetMediaType === AssetMediaType.Video ? (
-          <Card.Video shouldPlay={mediaShouldBePlaying} setCurrentTokenPlayingMedia={setCurrentTokenPlayingMedia} />
-        ) : (
-          <Card.Audio shouldPlay={mediaShouldBePlaying} setCurrentTokenPlayingMedia={setCurrentTokenPlayingMedia} />
-        )}
+        <MouseoverTooltip
```
JackShort, line 149: I don't think this is the correct way to implement this. I think it should be implemented in a similar fashion as the added and removed to bag popover. If you look this has some weird implications  - it causes the rarity tooltips to become slightly misaligned
Screen Shot 2022-12-01 at 3 35 22 PM

lynnshaoyu: hmm i don't see whats misaligned about it? 

it can't be implemented the same way as added /removed because it needs to wrap the component (image) where the mouseover triggers the tooltip.

src/nft/components/profile/view/ViewMyNftsAsset.tsx
```diff
@@ -126,7 +126,8 @@ export const ViewMyNftsAsset = ({
             text={getUnsupportedNftTextComponent(asset)}
             placement="bottom"
             offsetX={0}
-            offsetY={-100}
+            offsetY={-60}
```
JackShort, line 129: why is this -60 when the other one is -50?

JackShort (approved):

src/nft/components/profile/view/ViewMyNftsAsset.tsx
lynnshaoyu: cuz it looks weird when its 50 😅 too close to the bottom

PR #5509, opened 2022-12-01
feat: add Sentry / improve error handling

grabbou:
Fixes WEB-2239

### Highlights:

- removed "open Github issue" + associated code
- removed Google Analytics "sendEvent('exception')" from all the places. By default, a "fatal" error is reported (there's no recovery rather than a reload). Optionally, we can log other errors/messages with e.g. "captureMessage" and "log" or "debug"
- basic Sentry integration inside an existing ErrorBoundary. We're tracking releases by commits, that is aligned with other analytics
- created Sentry project under the name of this Github repository. We may need to update other Uniswap projects on Sentry to follow this pattern too 

Sample report: https://sentry.io/organizations/uniswap-labs/issues/3781734233/ to check for tracked data. Right now, it's a user agent information (same as before) and additional dump of a console. 

We have some first-class integrations with React Router and Redux available too as well as an additional package for measuring performance that we may consider using in the future too. 

### How to test the PR:

First of all, comment out `isProduction` conditional in `index.tsx` so that you enable Sentry logging.

**Via ErrorBoundary (render errors)** 

Place:
```ts
 useEffect(() => {
   throw new Error('test-inline')
  })
```
somewhere in the render method of any component visible on the page you want to test. 

Remember about the following:

> In development mode, React will rethrow errors caught within an error boundary. This will result in errors being reported twice to Sentry with the above setup, but this won’t occur in your production build.

and that React renders its own error overlay in development mode that you need to dismiss with ESC or "x" to see a fallback UI.

**All other exceptions**

Just place a `setTimeout(() => { throw new Error('inline-timeout') }, 5000)` anywhere in the code on a top-level or in an event handler. React will not catch it, but Sentry will report. 

### Todo (next PR):
- source maps / integration with release pipeline
- redesign error page once designs are done
- drop Google Analytics completely (different task, but somewhat related to removal of exception tracking in GA)
- [investigate service workers setup](https://github.com/Uniswap/interface/pull/5509#issuecomment-1335154249)


package.json
```diff
@@ -243,4 +244,4 @@
     "node": "14",
     "yarn": ">=1.22"
   }
-}
+}
```
zzmp, line 247: This should be unchanged.

src/components/LiquidityChartRangeInput/index.tsx
```diff
@@ -157,7 +157,7 @@ export default function LiquidityChartRangeInput({
   )
 
   if (error) {
-    sendEvent('exception', { description: error.toString(), fatal: false })
+    Sentry.captureMessage(error.toString(), 'log')
```
zzmp, line 160: Is there a reason this (and other usages) can't import only the used bit, eg `import { captureMessage } from '@sentry/react`?

src/index.tsx
```diff
@@ -47,6 +49,13 @@ function Updaters() {
   )
 }
 
+if (isProductionEnv()) {
```
zzmp, line 52: What happens in non-production environments?
Will Sentry still log, or will `captureMessage` calls in other files be nops?

zzmp, line 52: I would prefer this be above `Updaters`, next to other initialization code, so that all the jsx can remain adjacent.

src/components/LiquidityChartRangeInput/index.tsx
zzmp, line 160: Don't we also want to capture any stack in the `error`?

grabbou: No, I actually thought about it, but somewhat found `captureMessage` alone a bit unintuitive. Not a strong preference tbh.

src/index.tsx
grabbou: A noop. I had it turned on by default initially in all environments with an appropriate tag, but that produces an unnecessary noise. Do we want to track in staging tho?

src/components/ErrorBoundary/index.tsx
```diff
@@ -33,179 +31,78 @@ const CodeBlockWrapper = styled.div`
...
+              </ExternalLink>
+            </Padding>
+          </AutoRow>
+        </AutoColumn>
+      </BodyWrapper>
+    </FallbackWrapper>
+  )
 }
 
-const IS_UNISWAP = window.location.hostname === 'app.uniswap.org'
```
zzmp, line 49: This is important - we don't want to bother tracking forks, or usages through IPFS.

src/components/ErrorBoundary/index.tsx
```diff
@@ -33,179 +31,78 @@ const CodeBlockWrapper = styled.div`
...
-}
-`
+export default function ErrorBoundary({ children }: PropsWithChildren): JSX.Element {
+  return (
+    <Sentry.ErrorBoundary
+      fallback={Fallback}
+      beforeCapture={(scope) => {
+        scope.setLevel('fatal')
+      }}
+      onError={reloadIfUpdateAvailable}
```
zzmp, line 103: Does this mean we will reload immediately? This could be bad UX - a user may think its even more broken if the site just reloads instead of displays breakage information. I would rather reload only occurs on user interaction (eg clicking a "reload" button).

src/components/ErrorBoundary/index.tsx
```diff
@@ -33,179 +31,78 @@ const CodeBlockWrapper = styled.div`
...
   padding: 6px 24px;
 `
 
-const SomethingWentWrongWrapper = styled.div`
-  padding: 6px 24px;
-`
-
-type ErrorBoundaryState = {
-  error: Error | null
+const Fallback = ({ error }: { error: Error }) => {
```
zzmp, line 38: I think this component should include a reload button, using `reloadIfUpdateAvailable`-like code - but I think it should reload even if there isn't an update available.

zzmp (changes requested):

src/components/LiquidityChartRangeInput/index.tsx
grabbou: I am actually not sure whether we want to track this and the other error outside of the ErrorBoundary. This is inconsistent in a sense that we're not tracking other errors throughout the app in a similar way. 

Answering your question, I take this error as a backend error and that's probably why we never tracked its stack trace.

Unless we find that information necessary, I would remove it.

src/components/ErrorBoundary/index.tsx
grabbou: Yes, this is what we do right now anyway, so I just renamed the function to be more explicit for time being on what it actually does!

I am also in favour of doing a button, I was planning to implement this in a follow-up once design arrives (together with toggling stack trace and other UI changes).

src/components/ErrorBoundary/index.tsx
grabbou: Yes, users will trigger reload either manually by pressing a button or CMD+R (or other combination). However, we will still run part of the code from that function in the background to actually update service workers in the background ahead of users reloading. 

The `update` function is async and actually downloads the actual code, so I expect this to take some time depending on the network connection.

Currently, reload happens after update is finished, which means that on a fast connection users don't see error page, on slower connection they have random reload after some time. 

Unfortunately, this means that it will be technically possible for users to press reload while service worker update is in progress, as there's no way for us to prevent a reload from happening. FWIW, this was happening already, so it's not a new behaviour, just something to keep in mind.

src/components/ErrorBoundary/index.tsx
grabbou: Correct, however, we don't do that while initialising Uniswap Analytics in `App.tsx`. Does that service have special handling for hostname already included?

grabbou: I think we can safely ignore this on a code-level (since forks can remove that code too) and instead, configure Sentry to only accept requests coming from `app.uniswap.org`. We would have to do it anyway, since the URL for sending events is public.

This can be done under Project Settings -> Client Security.

grabbou: Screenshot 2022-12-02 at 02 39 15

Here's a screenshot. Purposefully set to `*` in case we have staging or other environments, but we can narrow it down to just `app` too.

grabbou:
Update: ErrorBoundary only works/catches error during the render, the fallback UI will not work if we have an async error during setTimeout, event handler etc. 

I don't consider this a blocker for this PR since Sentry actually tracks and catches all errors (including the ones coming from async handlers that React doesn't), but it would be great to improve this going forward.

package.json
grabbou: I don't know why `yarn` keeps changing this locally, I'll look into it. Maybe we need some updates to editor config

zzmp:
There are some patterns around to overcome this by setting/throwing async errors in state. It'd be great to find a more sustainable pattern for that.

package.json
grabbou: Solved: had to disable "default" JSON formatter.

src/index.tsx
grabbou: Re-grouped

src/components/LiquidityChartRangeInput/index.tsx
grabbou: Following up on this, we have a choice between:
```
import * as Sentry from 'sentry'
```
vs
```
import {init as initSentry} from 'sentry'
```
while I am a huge proponent of latter approach in code in general (over defaults and import all), I think former makes more sense in this context.

zzmp (approved):
lgtm because it's just a cleanup, but I'd like to see a few changes in a follow-up PR, to "fix" existing UX:
- The page should not reload automatically. This is disruptive to UX. It _should_ still update the ServiceWorker automatically.
- The page should include a "reload" button for a user to click to self-resolve. The reload button should be disabled until the attempt to reload the ServiceWorker is complete.

zzmp:
Can you move your ServiceWorker thoughts to a separate document so the conversation isn't stuck in a (soon-to-be-merged) PR?

grabbou:
Yes @zzmp, I move this to a Notion document and share internally so we can follow-up. Agree on your other comment, this will be addressed in my next PR with new designs.

PR #5510, opened 2022-12-01
chore: remove all remaining unused exports

vm:
Removes all remaining dead exports! Now we can add this as a condition on our build.


src/components/FeatureFlagModal/FeatureFlagModal.tsx
```diff
@@ -209,6 +210,14 @@ export default function FeatureFlagModal() {
           label="Enables JSON-RPC tracing"
         />
       </FeatureFlagGroup>
+      <FeatureFlagGroup name="Tokens">
```
zzmp, line 213: Should this be in a separate PR?

zzmp (changes requested):

src/components/FeatureFlagModal/FeatureFlagModal.tsx
vm: i can do in sep pr yes, but i did it so we had no more unused things

vm: https://github.com/Uniswap/interface/pull/5511

zzmp (approved):
just merge it after #5511

PR #5512, opened 2022-12-01
fix: add back previously deleted trending row selected event

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-2487


cbachmeier (approved):
LGTM

PR #5513, opened 2022-12-01
build: lint for unused exports

vm:
This makes sure that the lint fails if there are any exports from modules that are unused. We can always ignore a file or line if it's needed while something is in-progress.


src/nft/types/discover/react-table-config.d.ts
```diff
@@ -2,19 +2,23 @@ import { UseSortByColumnOptions, UseSortByColumnProps, UseSortByOptions, UseSort
 
 /* https://github.com/TanStack/table/issues/2970 */
 declare module 'react-table' {
+  // eslint-disable-next-line import/no-unused-modules
```
zzmp, line 5: You should instead be disabling this rule for .d.ts files. Is that possible?

zzmp (changes requested):

src/nft/types/discover/react-table-config.d.ts
vm: yes

yyip-dev (approved):
Lgtm

zzmp (approved):

PR #5514, opened 2022-12-01
chore: remove some unused exports

vm:
removing more exports that aren't needed. when we export something that isn't used outside of the file this will fail the new lint rule.


src/state/user/reducer.ts
```diff
@@ -140,14 +140,6 @@ const userSlice = createSlice({
       state.tokens[serializedToken.chainId][serializedToken.address] = serializedToken
       state.timestamp = currentTimestamp()
     },
-    removeSerializedToken(state, { payload: { address, chainId } }) {
```
grabbou, line 143: Looks like we have "addSerializedToken" too. Looks like we needed to remove them at some point. Since I am not sure what's a serialised token in this case, hard for me to judge whether this is/may be needed in the future.

src/theme/styled.d.ts
```diff
@@ -3,7 +3,7 @@ import { getTheme } from './index'
 type InferredTheme = ReturnType<typeof getTheme>
 
 declare module 'styled-components/macro' {
-  export interface DefaultTheme extends InferredTheme {
+  interface DefaultTheme extends InferredTheme {
```
grabbou, line 6: While this makes sense, I wonder why official documentation (and every single example out there) go with "export" https://styled-components.com/docs/api#create-a-declarations-file

One thing that comes to my mind is being able to actually import a "DefaultTheme" type from "styled-components/macro" directly (not from this file), which could confuse the linter. 

Can you check `DefaultTheme` on this line:
https://github.com/Uniswap/interface/blob/91157b7a43f0c7c75238ac716298bbc72b30810a/src/components/Badge/index.tsx#L3
has `any` or an actual type? I am afraid it may be `any`.

grabbou (changes requested):
LGTM, except for one concern regarding Styled Components types declaration file. The other comment is just a comment, but I did it all in one pass.

src/state/user/reducer.ts
vm: unclear to me too. but i will make a task to come back to look at this... we probably don't need addSerializedToken

src/theme/styled.d.ts
vm: if it's in the docs i can just ignore this line

yyip-dev (approved):
Lgtm, let's make sure we have a task for what Mike pointed out

PR #5515, opened 2022-12-01
chore: remove more unused exports

vm:
Removing even more unused exports


src/components/Button/index.tsx
```diff
@@ -7,7 +7,7 @@ import { RowBetween } from '../Row'
 
 type ButtonProps = Omit<ButtonPropsOriginal, 'css'>
 
-export const BaseButton = styled(RebassButton)<
+const BaseButton = styled(RebassButton)<
```
grabbou, line 10: this is awesome, impossible to spot without a linter, can be confusing to see exported members while refactoring

src/components/Menu/index.tsx
```diff
@@ -146,179 +90,6 @@ const InternalMenuItem = styled(Link)`
...
-                )
-            }
-          })()}
-      </StyledMenu>
-      <PrivacyPolicyModal />
-      <FeatureFlagModal />
-    </>
-  )
-}
-
```
grabbou, line 321: Looks like this file exports another member `NewMenu` (can't see it in the diff, have to check the file). Let's rename that to `export default` now that we got rid of the `OldMenu`

src/components/Popups/SurveyPopup.tsx
```diff
@@ -51,7 +51,7 @@ const WrappedCloseIcon = styled(X)`
 
 const END_TIMESTAMP = 1642272346 // Jan 15th
 
-export default function SurveyPopup() {
+function SurveyPopup() {
```
vm, line 54: this entire file is unused it seems

src/components/Popups/SurveyPopup.tsx
```diff
@@ -51,7 +51,7 @@ const WrappedCloseIcon = styled(X)`
...
   const theme = useTheme()
```
grabbou, line 55: How did that pass your local linter 🫣

grabbou: This is really interesting, we now have a file that has no exported members... I would assume it should be automatically removed!

grabbou (changes requested):
LGTM, except for one lint issue and one nit 🤞

src/components/Popups/SurveyPopup.tsx
vm: it's because the other thing is just a warning, not an error!

yyip-dev (approved):
Lgtm

PR #5526, opened 2022-12-02
fix: only show owner container if we have NftBalance data

cbachmeier:
Previously there could be a BE discrepancy where an asset could be shown as owned by a user but when calling their NftBalance, the asset would not be present. This would cause a crash if they tried to list said asset. We now hide the owner container in Details unless the NftBalanceQuery for that asset returns data.

### DIFF
--- a/src/nft/components/details/AssetPriceDetails.tsx
+++ b/src/nft/components/details/AssetPriceDetails.tsx
@@ -314,9 +314,12 @@ export const AssetPriceDetails = ({ asset, collection }: AssetPriceDetailsProps)
 
   const USDPrice = useUsdPrice(asset)
 
-  const assetsFilter = [{ address: asset.address, tokenId: asset.tokenId }]
+  const assetsFilter = [{ address: asset.address, tokenId: asset.tokenId + 'adskhdashk' }]
   const { walletAssets: ownerAssets } = useNftBalanceQuery(account ?? '', [], assetsFilter, 1)
-  const walletAsset: WalletAsset = useMemo(() => ownerAssets[0], [ownerAssets])
+  const walletAsset: WalletAsset | undefined = useMemo(
+    () => (ownerAssets?.length > 0 ? ownerAssets[0] : undefined),
+    [ownerAssets]
+  )
 
   const { assetInBag } = useMemo(() => {
     return {
--- a/src/nft/components/details/AssetPriceDetails.tsx
+++ b/src/nft/components/details/AssetPriceDetails.tsx
@@ -314,7 +314,7 @@ export const AssetPriceDetails = ({ asset, collection }: AssetPriceDetailsProps)
 
   const USDPrice = useUsdPrice(asset)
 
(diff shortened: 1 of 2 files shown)


src/nft/components/details/AssetPriceDetails.tsx
```diff
@@ -330,7 +336,7 @@ export const AssetPriceDetails = ({ asset, collection }: AssetPriceDetailsProps)
     )
   }
 
-  const isOwner = asset.owner ? account?.toLowerCase() === asset.owner?.address?.toLowerCase() : false
+  const isOwner = asset.owner && !!walletAsset ? account?.toLowerCase() === asset.owner?.address?.toLowerCase() : false
```
lynnshaoyu, line 339: nit: could just do 
` const isOwner = asset.owner && !!walletAsset && account?.toLowerCase() === asset.owner?.address?.toLowerCase() `

src/nft/components/details/AssetPriceDetails.tsx
```diff
@@ -312,6 +314,10 @@ export const AssetPriceDetails = ({ asset, collection }: AssetPriceDetailsProps)
 
   const USDPrice = useUsdPrice(asset)
 
+  const assetsFilter = [{ address: asset.address, tokenId: asset.tokenId }]
+  const { walletAssets: ownerAssets } = useNftBalanceQuery(account ?? '', [], assetsFilter, 1)
+  const walletAsset: WalletAsset = useMemo(() => ownerAssets[0], [ownerAssets])
```
lynnshaoyu, line 319: do we need a check to see if ownerAssets.length > 0?

lynnshaoyu: maybe walletAsset can be undefined

lynnshaoyu (approved):
few comments but trust you will sort them out (if they are even real issues in the first place?) before merging

src/nft/components/details/AssetPriceDetails.tsx
```diff
@@ -349,7 +355,7 @@ export const AssetPriceDetails = ({ asset, collection }: AssetPriceDetailsProps)
         <AssetHeader>{asset.name ?? `${asset.collectionName} #${asset.tokenId}`}</AssetHeader>
       </AssetInfoContainer>
       {isOwner ? (
```
lynnshaoyu, line 357: if walletAsset can be undefined should check here (like: isOwner && walletAsset ? )

src/nft/components/details/AssetPriceDetails.tsx
cbachmeier: It can be undefined which is what is checked for in the isOwner bool

src/nft/components/details/AssetPriceDetails.tsx
cbachmeier: walletAsset being undefined is part of the isOwner check

PR #5528, opened 2022-12-02
fix: add back “NFT Sell item added” event

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-2486


src/nft/hooks/useSellAsset.ts
```diff
@@ -23,6 +25,12 @@ export const useSellAsset = create<SellAssetState>()(
       sellAssets: [],
       selectSellAsset: (asset) =>
         set(({ sellAssets }) => {
+          const trace = useTrace()
```
cbachmeier, line 28: I am currently unable to select items to add to bag in the preview

cbachmeier: Elsewhere I believe we have the event triggering on the button press instead of the hook, we may want to move this event there to maintain consistency

lynnshaoyu: you're right thanks for catching yikes ... shouldn't be using a hook here for trace. I moved it to the button components.

src/nft/components/details/AssetPriceDetails.tsx
```diff
@@ -214,11 +217,17 @@ const OwnerContainer = ({ asset }: { asset: WalletAsset }) => {
     () => (USDValue ? USDValue * asset.floor_sell_order_price : undefined),
     [USDValue, asset.floor_sell_order_price]
   )
+  const trace = useTrace()
 
   const goToListPage = () => {
     resetSellAssets()
     navigate('/nfts/profile')
     selectSellAsset(asset)
+    sendAnalyticsEvent(EventName.NFT_SELL_ITEM_ADDED, {
```
cbachmeier, line 226: great job catching this case

cbachmeier (approved):
LGTM

PR #5529, opened 2022-12-02
chore: remove more unused exports

vm:
almost the last pr!


src/constants/tokens.ts
```diff
@@ -124,21 +117,6 @@ export const DAI_OPTIMISM = new Token(
   'DAI',
   'Dai stable coin'
 )
-export const USDC: { [chainId in SupportedChainId]: Token } = {
```
yyip-dev, line 127: Curious how come we don't use this anywhere, isn't it a token we list?

yyip-dev (approved):
Lgtm, one quick question about USDC just to make sure

PR #5530, opened 2022-12-02
fix: add missing amplitude sell events

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-2521


src/nft/components/bag/profile/ListingModal.tsx
```diff
@@ -174,10 +168,12 @@ const ListingModal = () => {
     } else if (!paused) {
       setListingStatus(ListingStatus.FAILED)
     }
-    sendAnalyticsEvent(EventName.NFT_LISTING_COMPLETED, {
+    sendAnalyticsEvent(EventName.NFT_LISTING_SIGNED, {
       signatures_requested: listings.length,
       signatures_approved: listings.filter((asset) => asset.status === ListingStatus.APPROVED),
-      ...approvalEventProperties,
+      list_quantity: listings.length,
+      usd_value: ethPriceInUSD * totalEthListingValue,
```
cbachmeier, line 175: If this event is supposed to be for individual listings it should be part of the function `signListingRow` right now this only triggers once after all the signature requests have been complete (approved, rejected, failed, etc)

Otherwise what is list_quantity supposed to represent here and how is it different than `signatures_requested` and `signatures_approved`?

cbachmeier: you actually should probably move this to `signListing` fn in `src/nft/utils/listNfts.ts` which is per listing and will give the tx hash context requested in the ticket

cbachmeier (changes requested):

cbachmeier (approved):
LGTM, merged in latest. Punting tx hashes for now

PR #5531, opened 2022-12-02
feat: show Browser Wallet instead of Injected

vm:
Injected confuses people, hopefully this is clearer!


yyip-dev (approved):
Great. Discussed with and approved by Product.

PR #5534, opened 2022-12-05
style: prevent content of remove liquidity page from displaying itself outside of the main page

nikli2009:
A Smol UI Fix 
- fix: prevent page content from displaying itself outside of RemoveLiquidity page wrapper  

### Before:
Screenshot 2022-12-05 at 14 04 48

### After:
Screenshot 2022-12-05 at 14 05 48


nikli2009:
Seems some just fixed that using another approach few hours ago, closing the PR

PR #5536, opened 2022-12-05
feat: better error page

grabbou:
Implements a new design of the error page as per Figma, aligns with new components/sizing where possible.

To reproduce, uncomment `isProduction` check in `index.tsx` to init Sentry and then, add the following:
```ts
useEffect(() => {
    throw new Error('test')
  })
```
somewhere in the code, preferably in a top-level page component.

---

Screenshot 2022-12-05 at 21 44 56
Screenshot 2022-12-05 at 21 45 02
Screenshot 2022-12-05 at 21 45 16

---

Todo:
- [x] check translations
- [x] vertical scroll when text overflows on smaller screens / large error stack trace
- [x] error id from Sentry instead of a placeholder

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,7 +132,7 @@
     "@reach/portal": "^0.10.3",
     "@react-hook/window-scroll": "^1.3.0",
     "@reduxjs/toolkit": "^1.6.1",
-    "@sentry/react": "^7.23.0",
+    "@sentry/react": "7.20.1",
     "@types/react-relay": "^13.0.2",
     "@types/react-window-infinite-loader": "^1.0.6",
     "@uniswap/analytics": "1.2.0",
--- a/src/components/Column/index.tsx
+++ b/src/components/Column/index.tsx
@@ -1,9 +1,6 @@
-import styled from 'styled-components/macro'
+import styled, { DefaultTheme } from 'styled-components/macro'
 
-type Gap = 'sm' | 'md' | 'lg' | 'xl' | string
-
-const getGapValue = (gap: Gap) =>
-  (gap === 'sm' && '8px') || (gap === 'md' && '12px') || (gap === 'lg' && '24px') || (gap === 'xl' && '32px') || gap
+type Gap = keyof DefaultTheme['grids']
 
 export const Column = styled.div<{
   gap?: Gap
@@ -11,7 +8,7 @@ export const Column = styled.div<{
   display: flex;
   flex-direction: column;
   justify-content: flex-start;
-  gap: ${({ gap }) => gap && getGapValue(gap)};
+  gap: ${({ gap, theme }) => gap && theme.grids[gap]};
 `
 export const ColumnCenter = styled(Column)`
   width: 100%;
@@ -19,12 +16,12 @@ export const ColumnCenter = styled(Column)`
 `
 
 export const AutoColumn = styled.div<{
-  gap?: Gap
+  gap?: Gap | string
   justify?: 'stretch' | 'center' | 'start' | 'end' | 'flex-start' | 'flex-end' | 'space-between'
 }>`
   display: grid;
   grid-auto-rows: auto;
-  grid-row-gap: ${({ gap }) => gap && getGapValue(gap)};
+  grid-row-gap: ${({ gap, theme }) => (gap && theme.grids[gap as Gap]) || gap};
   justify-items: ${({ justify }) => justify && justify};
 `
 
--- a/src/components/Column/index.tsx
+++ b/src/components/Column/index.tsx
@@ -11,12 +11,13 @@ export const ColumnCenter = styled(Column)`
 `
 
 export const AutoColumn = styled.div<{
-  gap?: 'sm' | 'md' | 'lg' | string
+  gap?: 'sm' | 'md' | 'lg' | 'xl' | string
   justify?: 'stretch' | 'center' | 'start' | 'end' | 'flex-start' | 'flex-end' | 'space-between'
 }>`
   display: grid;
   grid-auto-rows: auto;
-  grid-row-gap: ${({ gap }) => (gap === 'sm' && '8px') || (gap === 'md' && '12px') || (gap === 'lg' && '24px') || gap};
+  grid-row-gap: ${({ gap }) =>
+    (gap === 'sm' && '8px') || (gap === 'md' && '12px') || (gap === 'lg' && '24px') || (gap === 'xl' && '32px') || gap};
   justify-items: ${({ justify }) => justify && justify};
 `
 
--- a/src/components/Column/index.tsx
+++ b/src/components/Column/index.tsx
@@ -1,9 +1,17 @@
 import styled from 'styled-components/macro'
 
-const Column = styled.div`
+type Gap = 'sm' | 'md' | 'lg' | 'xl' | string
+
+const getGapValue = (gap: Gap) =>
+  (gap === 'sm' && '8px') || (gap === 'md' && '12px') || (gap === 'lg' && '24px') || (gap === 'xl' && '32px') || gap
+
+export const Column = styled.div<{
+  gap?: Gap
+}>`
   display: flex;
   flex-direction: column;
   justify-content: flex-start;
+  gap: ${({ gap }) => gap && getGapValue(gap)};
 `
 export const ColumnCenter = styled(Column)`
   width: 100%;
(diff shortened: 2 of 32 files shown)


src/components/ErrorBoundary/index.tsx
```diff
@@ -174,7 +174,7 @@ const updateServiceWorkerInBackground = async () => {
 export default function ErrorBoundary({ children }: PropsWithChildren): JSX.Element {
   return (
     <Sentry.ErrorBoundary
-      fallback={Fallback}
+      fallback={({ error, eventId }) => <Fallback error={error} eventId={eventId} />}
```
grabbou, line 177: this is required to allow using hooks in a Fallback component (ErrorBoundary is a class component)

package.json
```diff
@@ -132,7 +132,7 @@
     "@reach/portal": "^0.10.3",
     "@react-hook/window-scroll": "^1.3.0",
     "@reduxjs/toolkit": "^1.6.1",
-    "@sentry/react": "^7.23.0",
+    "@sentry/react": "7.20.1",
```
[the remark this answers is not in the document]
vm: please make a task for this

src/components/earn/UnstakingModal.tsx
```diff
@@ -90,19 +90,19 @@ export default function UnstakingModal({ isOpen, onDismiss, stakingInfo }: Staki
           </RowBetween>
           {stakingInfo?.stakedAmount && (
             <AutoColumn justify="center" gap="md">
-              <ThemedText.DeprecatedBody fontWeight={600} fontSize={36}>
+              <ThemedText.HeadlineLarge>
```
grabbou, line 93: After adding `HeadlineLarge` from Figma, I did a quick search to replace places where `DeprecatedBody` was used to achieve same settings

src/components/Column/index.tsx
```diff
@@ -1,22 +1,30 @@
 import styled from 'styled-components/macro'
 
-const Column = styled.div`
+type Gap = 'sm' | 'md' | 'lg' | 'xl' | string
+
+const getGapValue = (gap: Gap) =>
```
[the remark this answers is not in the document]
vm: this function is chaos, but i guess it was already implemented like this

src/components/ErrorBoundary/index.tsx
```diff
@@ -1,66 +1,157 @@
...
-  flex-direction: column;
-  width: 100%;
-  align-items: center;
-  z-index: 1;
+  width: 100vw;
+  height: 100vh;
 `
 
 const BodyWrapper = styled.div<{ margin?: string }>`
+  width: min(100%, 500px);
```
vm, line 20: why not just min-width

src/components/ErrorBoundary/index.tsx
```diff
@@ -1,66 +1,157 @@
...
+  font-size: 16px;
+  padding: 10px 16px;
+  border-radius: 12px;
+`
+
+const StretchedRow = styled.div`
+  display: flex;
+  gap: 24px;
+
+  > * {
```
vm, line 42: i am not a fan of child selectors in styled components. better to create a shared style and extend the components for the children imo

src/components/ErrorBoundary/index.tsx
```diff
@@ -1,66 +1,157 @@
...
+`
+
+const Code = styled.code`
+  font-weight: 300;
+  font-size: 12px;
+  line-height: 16px;
+  word-wrap: break-word;
   width: 100%;
+  color: ${({ theme }) => theme.textPrimary};
+  font-family: courier, courier new, serif;
```
vm, line 55: feels like this should live as a variable somewhere

src/components/ErrorBoundary/index.tsx
```diff
@@ -1,66 +1,157 @@
...
+const CodeTitle = styled.div`
+  display: flex;
+  gap: 14px;
+  align-items: center;
+  word-break: break-word;
+`
+
+const Fallback = ({ error, eventId }: { error: Error; eventId: string | null }) => {
+  const [isExpanded, setExpanded] = useState(false)
+  const isMobile = useIsMobile()
```
vm, line 99: not a huge fan of this hook. can you just use css for this?

src/components/earn/ClaimRewardModal.tsx
```diff
@@ -88,9 +88,7 @@ export default function ClaimRewardModal({ isOpen, onDismiss, stakingInfo }: Sta
           </RowBetween>
           {stakingInfo?.earnedAmount && (
             <AutoColumn justify="center" gap="md">
-              <ThemedText.DeprecatedBody fontWeight={600} fontSize={36}>
-                {stakingInfo?.earnedAmount?.toSignificant(6)}
-              </ThemedText.DeprecatedBody>
+              <ThemedText.HeadlineLarge>{stakingInfo?.earnedAmount?.toSignificant(6)}</ThemedText.HeadlineLarge>
```
vm, line 91: conedison?

src/components/ErrorBoundary/index.tsx
grabbou: Why not, I can put it under theme as `codeFont`, albeit we're only using it in this component.

src/components/ErrorBoundary/index.tsx
grabbou: Actually, I think this could be `max-width: 500px` and `width: 100%`! `min-width` would make it render full screen on all sizes

src/components/ErrorBoundary/index.tsx
grabbou: Theoretically yes, I am not happy of it either as per my comment few lines later. However, I don't think there's a reasonable and clean way to do it right now while referring to values and names used in Figma. Figma specifies HeadlineSmall/BodySmall and if I were to use media queries, this would require hardcoding values for smaller viewports. 

This approach is generally preferred and considered good practice, it's just that it doesn't play well with the way we do things atm. 

Imagine, if text wasn't a component, but a `css` from Styled Components, we could just include different styles based on media queries, such as:
```
styled.div`
${BodySmall}

@media (min-width: 500px) {
  ${BodyLarge}
 }
 `
 ```
 and problem would be solved!

src/components/ErrorBoundary/index.tsx
vm: right

src/components/ErrorBoundary/index.tsx
vm: i think we would potentially use it elsewhere

src/components/ErrorBoundary/index.tsx
grabbou: Not a fan either, but I somewhat disliked the alternative approach of wrapping all children in a StyledComponent to give it `flex: 1, display: flex`. Don't really have a strong opinion here

src/components/ErrorBoundary/index.tsx
vm: i think we would just display: none if on mobile vs not

src/components/earn/ClaimRewardModal.tsx
vm: np

package.json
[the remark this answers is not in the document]
grabbou: WEB-2550 assigned to me

src/components/ErrorBoundary/index.tsx
vm: it's not a huge deal for now, figured i'd mention

src/components/ErrorBoundary/index.tsx
grabbou: Actually FWIW Figma references a different font, but I didn't justify pulling it for a single component at this point.

vm (approved):

src/components/ErrorBoundary/index.tsx
grabbou: Design team is aware and ok

src/components/ErrorBoundary/index.tsx
grabbou: Oh yes, that works too! Still, I think I'd do it behind a switch to potentially avoid rendering text twice if SEO ever mattered for this page haha

grabbou:
Will wait for another pass, since I ended up doing few more changes!

src/components/Column/index.tsx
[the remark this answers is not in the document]
grabbou: meanwhile, I removed the hardcoded values in favour of theme-based settings

grabbou:
Errors just got better 🎉 Now it's time to make sure this page appears less often!

PR #5540, opened 2022-12-05
build: send source maps to Sentry on every release

grabbou:
Adds a Github action to upload source maps for a given release (by commit sha) to Sentry


grabbou:
This will require few env variables to be configured on this repository. I will follow-up via DM

.github/workflows/release.yaml
```diff
@@ -110,3 +110,14 @@ jobs:
           projectName: ${{ secrets.CLOUDFLARE_PROJECT_NAME }}
           directory: build
           githubToken: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Upload source maps to Sentry
+        uses: getsentry/action-release@v1
```
zzmp, line 115: this should be versioned using a commit sha, not semver

zzmp (changes requested):

.github/workflows/release.yaml
grabbou: Do you mean a Github Action itself or a Sentry release? If former, any particular reason behind? Other Github actions are referenced by semver too.

If latter, Sentry uses commit sha by default, so no need to set a special release name (https://github.com/marketplace/actions/sentry-release#parameters - see "version")

zzmp: former, for security purposes. v1 can change what it does at the publisher's whim. using a sha is more stable.

other github actions should also be referenced by sha...eventually.

grabbou: Okay, makes sense. Overall I agree, I will create a ticket for updating this and other versions too at some point. Meanwhile, will apply that to this particular action.

grabbou:
Uploading locally works so we agreed with @zzmp to just go ahead and ship it. We will see tomorrow after schedule release is run whether this works w/o issues. If not, I'll send a follow-up.

zzmp (approved):

PR #5545, opened 2022-12-06
feat: permit 2 feature flag

zzmp:
Adds a Permit 2 / Universal Router feature flag.
Cleans up old feature flags, and the feature flags modal.

### DIFF
--- a/src/components/FeatureFlagModal/FeatureFlagModal.tsx
+++ b/src/components/FeatureFlagModal/FeatureFlagModal.tsx
@@ -1,5 +1,6 @@
 import { BaseVariant, FeatureFlag, featureFlagSettings, useUpdateFlag } from 'featureFlags'
 import { LandingPageVariant, useLandingPageFlag } from 'featureFlags/flags/landingPage'
+import { Permit2Variant, usePermit2Flag } from 'featureFlags/flags/permit2'
 import { TraceJsonRpcVariant, useTraceJsonRpcFlag } from 'featureFlags/flags/traceJsonRpc'
 import { useAtomValue, useUpdateAtom } from 'jotai/utils'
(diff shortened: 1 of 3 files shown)


src/featureFlags/flags/featureFlags.ts
```diff
@@ -1,6 +1,5 @@
 export enum FeatureFlag {
-  favoriteTokens = 'favoriteTokens',
```
vm, line 2: is there a lint rule for unused enum variants lol

src/featureFlags/flags/permit2.ts
```diff
@@ -0,0 +1,7 @@
+import { BaseVariant, FeatureFlag, useBaseFlag } from '../index'
```
vm, line 1: i want to refactor this system to just have a single hook "useFeatureFlags" instead of having these files eventually

src/components/FeatureFlagModal/FeatureFlagModal.tsx
```diff
@@ -202,6 +203,18 @@ export default function FeatureFlagModal() {
           <X size={24} />
         </CloseButton>
       </Header>
+      <FeatureFlagOption
```
vm, line 206: maybe a group called "features"

vm (approved):

src/featureFlags/flags/featureFlags.ts
zzmp: there should be, but I can't find one

src/components/FeatureFlagModal/FeatureFlagModal.tsx
zzmp: I don't think we should group them if they are unrelated

PR #5546, opened 2022-12-06
build: install permit2 and universal-router sdks

zzmp:


yyip-dev (approved):
🎉 also want to call out we have `universal-router-sdk` @ v1.2.0 in widgets according to https://github.com/Uniswap/widgets/pull/322/files, looks like protocol team just bumped version earlier today in case we want to update there also

zzmp:
> 🎉 also want to call out we have `universal-router-sdk` @ v1.2.0 in widgets according to https://github.com/Uniswap/widgets/pull/322/files, looks like protocol team just bumped version earlier today in case we want to update there also

widgets will be updated as well in Uniswap/widgets/pull/321 - v1.2.0 is missing the const exports for some reason.

PR #5549, opened 2022-12-06
chore: update deps

grabbou:
Fixes WEB-2552, as per https://github.com/Uniswap/interface/pull/5540

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -2,15 +2,15 @@ name: Setup
 runs:
   using: composite
   steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
 
-    - uses: actions/setup-node@v3
+    - uses: actions/setup-node@8c91899e586c5b171469028077307d293428b516
       with:
         node-version: 14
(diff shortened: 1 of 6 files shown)


grabbou:
My only concern with this PR is reduced visibility on what version we're running. I do wonder if it would make sense to annotate each commit sha with a version it refers to?

zzmp:
> My only concern with this PR is reduced visibility on what version we're running. I do wonder if it would make sense to annotate each commit sha with a version it refers to?

I think this makes sense.

I also think if we're planning on moving to a `.github` repo for shared actions, we can deprioritize this so that _that_ can be used as a source of truth.

zzmp (commented):
> My only concern with this PR is reduced visibility on what version we're running. I do wonder if it would make sense to annotate each commit sha with a version it refers to?

I think this makes sense.

I also think if we're planning on moving to a `.github` repo for shared actions, we can deprioritize this so that _that_ can be used as a source of truth.

amrsag:
Un subscribed pls

grabbou:
Let's revisit in the future once we wrap up shared GitHub actions template. While I generally agree in principle that pinning down action versions is a good action, I have mixed feelings about readability and maintenance of it in the form presented in this PR.

PR #5553, opened 2022-12-06
fix: hover button states

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-852?atlOrigin=eyJpIjoiNjgxZmQ1OWIwZWY1NDQyNTk5MjkxNzkzNjZiYTg1ZWEiLCJwIjoiaiJ9


src/components/Button/index.tsx
```diff
@@ -516,3 +529,12 @@ export const ThemeButton = ({ children, ...rest }: ThemeButtonProps) => {
...
 }
+
+export const ButtonLight = ({ children, ...rest }: BaseButtonProps) => {
+  return (
+    <BaseButtonLight {...rest}>
+      <ButtonOverlay />
+      {children}
+    </BaseButtonLight>
+  )
+}
```
JFrankfurt, line 540: I'm a little hesitant about this. It def works, but I think the ideal refactor would be to switch the consumers of the existing `ButtonLight` over to using `ThemeButton` with the appropriate configuration. If that seems like too much of a lift, then I think this should be exported as `DeprecatedButtonLight` or at least commented to indicate that the "correct" approach is `<ThemeButton size={ButtonSize[something]} emphasis={ButtonEmphasis.highSoft} />` w/ a `styled(ThemeButton)` wrapper for style exceptions.

JFrankfurt (changes requested):

src/components/Button/index.tsx
lynnshaoyu: I dont think that makes sense because ThemeButton doesn't share any base component with ButtonLight. ButtonLight's base component is BaseButton, which ThemeButton doesn't use at all. It feels a little "forced" to share code which introduces additional complexity? 

Also, Vig is going to refactor all the shared components so I think all of this is going to undergo a significant refactor soon anyways.

JFrankfurt: I mean that ThemeButton correctly implements the current design spec for all buttons, so if we're creating a one-off button to accomplish a standard button variant, then that's probably a mistake because ThemeButton already provides the implementation (with associated hover states!)

JFrankfurt: Just replacing the existing `ButtonLight` usage with `<ThemeButton size={ButtonSize[something]} emphasis={ButtonEmphasis.highSoft}><Content /></ThemeButton>` should work.

lynnshaoyu: Ok I created a separate draft PR with your suggested implementation, but I see some issues with the hover state style treatment getting messed up after one click (see recorded loom video) : https://github.com/Uniswap/interface/pull/5719 

Based on your feedback I can merge either this one or the other one, and close the remaining PR.

PR #5554, opened 2022-12-07
feat: adds permit2 and universal router integration

zzmp:
Adds Permit2 and Universal Router UI/state behind the permit2 flag.

Most of the hooks are copied verbatim from widgets, then modified to accommodate interface's fs/utils/state.


yyip-dev:
@zzmp I pushed a couple lint fixes to get the build to pass. When I tested it locally, I noticed two issues:
1. When I reject an approval from my wallet, it returns to the app with a full page error (which I think differs from the behavior in our current swap)
Screen Shot 2022-12-06 at 11 48 22 PM

2. The Swap button is disabled even after I approve some token - is there logic we need to update [on this line](https://github.com/Uniswap/interface/blob/9ef6a2141e42f10b56429cc169ce46e98d410cd9/src/pages/Swap/index.tsx#L774) in particular regarding `signatureState`?

src/hooks/useUniversalRouter.ts
```diff
@@ -0,0 +1,80 @@
+import { BigNumber } from '@ethersproject/bignumber'
+// eslint-disable-next-line @typescript-eslint/no-restricted-imports
```
zzmp, line 2: This should not be disabled. Instead, it can import the type or import through the abstract package.

zzmp:
> @zzmp I pushed a couple lint fixes to get the build to pass. When I tested it locally, I noticed two issues:
> 
> 1. When I reject an approval from my wallet, it returns to the app with a full page error (which I think differs from the behavior in our current swap)
> 
> Screen Shot 2022-12-06 at 11 48 22 PM
> 
> 2. The Swap button is disabled even after I approve some token - is there logic we need to update [on this line](https://github.com/Uniswap/interface/blob/9ef6a2141e42f10b56429cc169ce46e98d410cd9/src/pages/Swap/index.tsx#L774) in particular regarding `signatureState`?

Bot should now be resolved.

src/hooks/usePermit2.ts
```diff
@@ -0,0 +1,84 @@
...
+  } | void>
+}
+
+export default function usePermit(amount?: CurrencyAmount<Token>, spender?: string): Permit {
+  const { account } = useWeb3React()
+  const tokenAllowance = useTokenAllowance(amount?.currency, account, PERMIT2_ADDRESS)
+  const updateTokenAllowance = useUpdateTokenAllowance(amount, PERMIT2_ADDRESS)
+
+  const allowanceData = usePermitAllowance(amount?.currency, spender)
+  const [permitAllowance, setPermitAllowance] = useState(allowanceData?.amount)
```
JFrankfurt, line 34: is this state redundant? It looks like it's just being kept in sync with `allowanceData.amount`

src/hooks/usePermit2.ts
```diff
@@ -0,0 +1,84 @@
...
+    () => {
+      const now = Date.now() / 1000 - 12 // ensure it can still go into this block (assuming 12s block time)
+      if (signature && signature.sigDeadline < now) {
+        setSignature(undefined)
+      }
+      if (allowanceData && allowanceData.expiration < now) {
+        setPermitAllowance(undefined)
+      }
+    },
+    ms`12s`,
```
JFrankfurt, line 57: might be worth moving this out into a constant `AVERAGE_L1_BLOCK_TIME`

src/pages/Swap/index.tsx
```diff
@@ -752,6 +798,54 @@ export default function Swap() {
                       </ButtonError>
                     </AutoColumn>
                   </AutoRow>
+                ) : permit.state === PermitState.PERMIT_NEEDED ? (
+                  <ButtonYellow
```
JFrankfurt, line 802: nit: option to use `ThemeButton`

src/pages/Swap/index.tsx
```diff
@@ -768,8 +862,14 @@ export default function Swap() {
...
                     }}
                     id="swap-button"
-                    disabled={!isValid || routeIsSyncing || routeIsLoading || priceImpactTooHigh || !!swapCallbackError}
-                    error={isValid && priceImpactSeverity > 2 && !swapCallbackError}
+                    disabled={
+                      !isValid ||
+                      routeIsSyncing ||
+                      routeIsLoading ||
+                      priceImpactTooHigh ||
+                      Boolean(!permit2Enabled && swapCallbackError)
```
JFrankfurt, line 870: I think this is complicated enough to warrant moving it up out of the template

src/pages/Swap/index.tsx
```diff
@@ -768,8 +862,14 @@ export default function Swap() {
...
+                    }
+                    error={isValid && priceImpactSeverity > 2 && (permit2Enabled || !swapCallbackError)}
```
JFrankfurt, line 872: same here

JFrankfurt (approved):
some nits/questions, but lgtm

src/hooks/usePermit2.ts
zzmp: It is, but it's cleared should the expiration pass. It is used to prevent re-running the memoization at the frequency of the interval timer.

src/hooks/usePermit2.ts
zzmp: will do - moving to this instead of `POLLING_INTERVAL` from the providers file.

src/pages/Swap/index.tsx
```diff
@@ -752,6 +798,54 @@ export default function Swap() {
                       </ButtonError>
                     </AutoColumn>
                   </AutoRow>
+                ) : permit.state === PermitState.PERMIT_NEEDED ? (
+                  <ButtonYellow
+                    onClick={updatePermit}
+                    id="permit-button"
```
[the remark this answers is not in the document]
zzmp: cargo culting - I'll remove this.

src/pages/Swap/index.tsx
zzmp: I think moving it out to a new variable will be equally as confusing.

PR #5556, opened 2022-12-07
chore: use @uniswap eslint preset

grabbou:
Replaces internal ESLint config with a shared Uniswap one.

Note that there might be temporary warnings about unmeet peer dependencies, because React Scripts is using ESLint 7.x and we're using ESLint 8.x in our config, but I am going to work on upgrading React Scripts to latest version in a follow-up PR, so that is going to go away.

It has no effect on the application itself, but figured I make it explicit.


package.json
```diff
@@ -90,23 +91,13 @@
     "@types/ua-parser-js": "^0.7.35",
     "@types/uuid": "^8.3.4",
     "@types/wcag-contrast": "^3.0.0",
-    "@typescript-eslint/eslint-plugin": "^4",
-    "@typescript-eslint/parser": "^4",
+    "@uniswap/eslint-config": "git+ssh://git@github.com:uniswap/eslint-config.git#da8c311",
```
grabbou, line 94: This is not going to work anywhere but locally, waiting for the initial release.

.eslintrc.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * This is a workaround to allow ESLint to resolve plugins that were installed
+ * by an external config: https://github.com/eslint/eslint/issues/3458
+ */
+require('@rushstack/eslint-patch/modern-module-resolution')
```
zzmp, line 5: Why did you put this here instead of in @uniswap/eslint-config?
[The docs say either will work.](https://www.npmjs.com/package/@rushstack/eslint-patch#:~:text=For%20an%20even%20leaner%20setup%2C%20%40your%2Dcompany/eslint%2Dconfig%20can%20provide%20the%20patch%20as%20its%20own%20dependency.%20See%20%40rushstack/eslint%2Dconfig%20for%20a%20real%20world%20example%20and%20recommended%20approach.)

grabbou: > For an even leaner setup, @your-company/eslint-config can provide the patch as its own dependency. 

Good call! I guess I didn't see it initially. There might be some additional changes anyway, so I will release a new version and update this PR accordingly.

grabbou:
Some additional changes were required (in my opinion - really positive) since we didn't previously have `eslint/recommended` preset enabled. We only had `@typescript-eslint/recommended` turned on. That preset, unlike its name, is supposed to turn off some of the conflicting ESLint recommended rules off to make it work better with Typescript. 

Right now, we have ESLint recommended rules turned on by default, which I believe was desired from the very beginning.

Please note that without such preset, the number of rules we had turned on was very minimal.

.vscode/settings.json
```diff
@@ -5,15 +5,12 @@
   "editor.formatOnSaveMode": "file",
   "editor.tabCompletion": "on",
   "editor.tabSize": 2,
-  "editor.formatOnSave": true,
+  "editor.formatOnSave": false,
```
grabbou, line 8: note: this is now deprecated and `editor.codeActionsOnSave` is recommended for ESLint plugin

src/nft/hooks/useNFTSelect.ts
```diff
@@ -25,7 +25,7 @@ export const useNFTSelect = create<SelectNFTState>()(
       selectNFT: (nft) =>
         set(({ selectedNFTs }) => {
           if (selectedNFTs.length === 0) return { selectedNFTs: [nft] }
-          else if (!!selectedNFTs.find((x) => x.id === nft.id))
+          else if (selectedNFTs.find((x) => x.id === nft.id))
```
zzmp, line 28: `find` is incorrect - use [`some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)

src/nft/hooks/useNFTSelect.ts
```diff
@@ -25,7 +25,7 @@ export const useNFTSelect = create<SelectNFTState>()(
       selectNFT: (nft) =>
         set(({ selectedNFTs }) => {
           if (selectedNFTs.length === 0) return { selectedNFTs: [nft] }
```
zzmp, line 27: This bracket pattern should be linted against. If there is an `else`, it should use multi-line brackets. Only a one-liner guard should be allowed to omit the brackets (eg `if (condition) return earlyReturnValue`).

src/nft/hooks/useWalletCollections.ts
```diff
@@ -38,7 +38,7 @@ export const useWalletCollections = create<WalletCollectionState>()(
       setCollectionFilters: (address) =>
         set(({ collectionFilters }) => {
           if (collectionFilters.length === 0) return { collectionFilters: [address] }
-          else if (!!collectionFilters.find((x) => x === address))
+          else if (collectionFilters.find((x) => x === address))
```
zzmp, line 41: `find` is incorrect - use [`some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)

src/serviceWorker/document.test.ts
```diff
@@ -106,15 +106,15 @@ describe('document', () => {
       let fetched: Response
       const FETCHED_ETAGS = 'fetched'
 
+      const assertFetch = () => {
```
zzmp, line 109: This should be more verbose; and it is an expectation, not an assertion: `expectFetchToHaveBeenCalledWithRequestUrl`

src/theme/index.tsx
```diff
@@ -17,6 +17,22 @@ export const MEDIA_WIDTHS = {
...
+const makeMediaQuery =
+  (size: number) =>
+  (...args: Parameters<typeof css>) =>
+    css`
+      @media (max-width: ${size}px) {
+        ${css(...args)}
+      }
+    `
+
+const deprecated_mediaWidthTemplates = {
```
zzmp, line 29: This should be explicitly typed still to stay in sync with `MEDIA_WIDTHS`: `: { [width in keyof typeof MEDIA_WIDTHS]: typeof css }`

src/state/lists/reducer.test.ts
```diff
@@ -386,18 +386,19 @@ describe('list reducer', () => {
...
-        // this is ok.
-        Object.keys(byUrl).forEach((url) => {
-          if (url !== 'https://unpkg.com/@uniswap/default-token-list@latest/uniswap-default.tokenlist.json') {
-            expect(byUrl[url]).toEqual({
+        Object.entries(byUrl)
+          // We don't expect the Uniswap default list to be prepopulated
+          .filter(
+            ([key]) => key !== 'https://unpkg.com/@uniswap/default-token-list@latest/uniswap-default.tokenlist.json'
+          )
+          .forEach(([, state]) => {
```
zzmp, line 394: `value`, not `state`, because you're using `entries()`

package.json
```diff
@@ -134,12 +124,13 @@
     "@reach/portal": "^0.10.3",
     "@react-hook/window-scroll": "^1.3.0",
     "@reduxjs/toolkit": "^1.6.1",
+    "@rushstack/eslint-patch": "^1.2.0",
```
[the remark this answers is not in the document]
grabbou: Yes, this was resolved just now as I waited to do the upgrade to 1.1.1 eslint-config!

cypress/release.ts
```diff
@@ -9,6 +11,7 @@ describe(
   () => {
     it('loads swap page', () => {
       // We *must* wait in order to space out the retry attempts.
+      // @todo find a better way to do this
```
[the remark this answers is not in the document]
grabbou: Will address in a separate PR, we have some extra occurrences of `@todo` in the codebase already, will serve as a great pattern to test this against.

src/serviceWorker/document.test.ts
grabbou: Always to the rescue with proper name!

src/state/lists/reducer.test.ts
grabbou: I guess it's just a personal preference - I often rename `key` and `value` to give it more meaningful names, such as `url` and `state` in this case.

It looks like both patterns are currently present in the codebase, e.g.:
`Object.entries(MARKETPLACE_ITEMS).map(([value, title])`
or
`Object.entries(traitsByGroup).map(([type, traits]))`

src/theme/index.tsx
grabbou: Good catch. We can use it to enforce any potential additions in the future.

src/nft/hooks/useNFTSelect.ts
grabbou: I am going to write it down as a follow-up to this PR. This will more than likely affect other parts of the codebase too. Overall I am in favor.

src/theme/index.tsx
grabbou: Decided to revert to original code with minimum changes, this is going to go away anyway in the nearest PRs

cypress/release.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint cypress/no-unnecessary-waiting: 1 */
```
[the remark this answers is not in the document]
grabbou: Wish we actually had this as a rule, I'd rather have a warning (which acts as a todo itself, unless explicitly muted) or an ESLint rule that automatically surpasses any errors/warning on the line following TODO and makes it a warning. Otherwise, most TODOs get lost.

src/state/lists/reducer.test.ts
zzmp: > I often rename key and value to give it more meaningful names

This is better - I agree - but in this case you're being inconsistent. You should either rename `key` to `url`, or `state` to `value`.

.vscode/extensions.json
```diff
@@ -0,0 +1,6 @@
+{
+  "recommendations": [
+    "dbaeumer.vscode-eslint"
+  ],
+  "unwantedRecommendations": []
+}
```
zzmp, line 6: Add a newline

src/state/lists/reducer.test.ts
grabbou: Right 👀 I will update `filter`, don't know why I skipped it while glancing the code.

.vscode/extensions.json
grabbou: Done

zzmp (approved):

PR #5563, opened 2022-12-07
chore: replace deprecated colors with correct theme values [part 1]

grabbou:
Transparent PR, simple find&replace. It is still possible to get rid of the remaining deprecated colors (and of the entire file), but this requires a sync with the design team to make sure we're okay unifying e.g. three shades of yellow into one.

There will be a follow-up, most likely next week when I get to work together with the design team from NYC.


vm (approved):
LGTM if design approved

PR #5565, opened 2022-12-07
feat: initial apollo configutation

cbachmeier:
Initial Apollo setup

https://uniswaplabs.atlassian.net/browse/WEB-2582


codegen.ts
```diff
@@ -0,0 +1,17 @@
+import type { CodegenConfig } from '@graphql-codegen/cli'
+
+const config: CodegenConfig = {
+  overwrite: true,
+  schema: './src/graphql/data/schema.graphql',
+  documents: ['./src', '!**/node_modules/**', '!**/__mocks__/**', '!**/__generated__/**', '!**/thegraph/**'],
+  generates: {
+    'src/graphql/data/__generated__/types-and-hooks.ts': {
```
justindomingue, line 8: they recommend checking in this generated file

cbachmeier: Currently we don't check in any of the generated files, what are your thoughts @zzmp ?

cbachmeier: In latest it is checked in

cbachmeier: One thing to note is that the checked in schema file would need to be ignored by our linting

zzmp: in favor of checking it in.

codegen.ts
```diff
@@ -0,0 +1,19 @@
+import type { CodegenConfig } from '@graphql-codegen/cli'
```
zzmp, line 1: Can this file be renamed to `apollo-codegen.ts`?

codegen.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const config: CodegenConfig = {
+  overwrite: true,
+  schema: './src/graphql/data/schema.graphql',
+  documents: ['./src', '!**/node_modules/**', '!**/__mocks__/**', '!**/__generated__/**', '!**/thegraph/**'],
```
zzmp, line 6: can this be more specific?

codegen.ts
```diff
@@ -0,0 +1,19 @@
+import type { CodegenConfig } from '@graphql-codegen/cli'
+
+const config: CodegenConfig = {
```
zzmp, line 3: Please link to the docs explaining what this file does.

codegen.ts
cbachmeier: sure, but this was a direct port of what was in `relay.config.js`

PR #5570, opened 2022-12-07
fix: erc1155s handling on collection page

JackShort:
* Allows users to add multiple of the same erc1155 token to their bag

DO NOT APPROVE WITHOUT HEAVY TESTING (updates core price recalculation logic)

Screen Shot 2022-12-08 at 12 35 50 PM


src/graphql/data/nft/Asset.ts
```diff
@@ -128,7 +128,7 @@ function formatAssetQueryData(queryAsset: NftAssetsQueryAsset, totalCount?: numb
   const asset = queryAsset.node
   const ethPrice = parseEther(wrapScientificNotation(asset.listings?.edges[0]?.node.price.value ?? 0)).toString()
   return {
-    id: asset.id,
+    id: undefined,
```
JackShort, line 131: the id being passed here is the one assigned by the backend but we need to assign one ourselves when we add it to bag

src/nft/utils/txRoute/combineItemsWithTxRoute.ts
```diff
@@ -31,14 +41,44 @@ const isAveragePriceOfPooledAssets = (
...
   item: UpdatedGenieAsset,
   items: UpdatedGenieAsset[],
   route: RoutingItem,
   txRoute?: RoutingItem[]
 ): boolean => {
   if (!(route && 'priceInfo' in route.assetOut)) return false
 
+  if (item.tokenType === TokenType.ERC1155) {
+    return isErc1155TokenCorrectSum(item, items, txRoute)
+  }
```
cbachmeier, line 80: I'm currently unable to buy an 1155 that is part of a liquidity pool and always get the error: `Error: cannot estimate gas; transaction may fail or may require manual gas limit`

src/graphql/data/nft/Asset.ts
JackShort: honestly I don't really know what the purpose of that would be if we assign a new one when it gets added to the bag. Previously we only gave an asset an id when it was added to the bag

src/nft/components/collection/CollectionAsset.tsx
```diff
@@ -81,7 +85,21 @@ export const CollectionAsset = ({
 
   const handleAddAssetToBag = useCallback(() => {
     if (BigNumber.from(asset.priceInfo?.ETHPrice ?? 0).gte(0)) {
-      addAssetsToBag([asset])
+      const shouldUseSellOrders = isErc1155 && isSelected && sellOrders && sellOrders?.length > quantitySelected
```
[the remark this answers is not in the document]
JackShort: no because if sellOrders.length is 1 and the quantitySelected is 1 then it will access sellorders[1] which will be undefined

src/nft/components/collection/CollectionAsset.tsx
```diff
@@ -81,7 +85,21 @@ export const CollectionAsset = ({
...
+      const newSellOrderPrice = shouldUseSellOrders
+        ? parseEther(sellOrders[quantitySelected].price.value.toString()).toString()
```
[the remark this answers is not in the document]
JackShort: no because this is before quantitySelected increases because it hasn't been added to bag so it has already been offset

PR #5572, opened 2022-12-07
feat: add learn more button

vm:
Screen Shot 2022-12-07 at 6 06 00 PM

### DIFF
--- a/src/pages/Landing.tsx
+++ b/src/pages/Landing.tsx
@@ -71,7 +71,6 @@ const SubText = styled.h3`
 `
 
 const CTAButton = styled(BaseButton)`
-  width: 288px;
   padding: 16px;
   border-radius: 24px;
   color: ${({ theme }) => theme.white};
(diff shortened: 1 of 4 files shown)


src/pages/Landing.tsx
```diff
@@ -70,6 +70,27 @@ const SubText = styled.h3`
...
+  }
+`
+
+const ButtonCTA = styled(CTAButton)`
+  background: linear-gradient(10deg, rgba(255, 0, 199, 1) 0%, rgba(255, 159, 251, 1) 100%);
+  border: none;
+`
+
+const ButtonCTASecondary = styled(CTAButton)`
+  background: none;
```
grabbou, line 90: I am pretty sure I've seen some styles like SecondaryButton or Outline in the code that have no background, so you could potentially use that one instead?

grabbou: *ButtonOutlined

src/pages/Landing.tsx
```diff
@@ -70,6 +70,27 @@ const SubText = styled.h3`
...
+  border-radius: 24px;
+  color: ${({ theme }) => theme.white};
+
+  &:hover {
+    opacity: 50%;
+  }
+`
+
+const ButtonCTA = styled(CTAButton)`
+  background: linear-gradient(10deg, rgba(255, 0, 199, 1) 0%, rgba(255, 159, 251, 1) 100%);
```
grabbou, line 85: Does it make sense to reference these values from theme, instead of hardcoding here? Sounds like we have all shades of pink available already

src/pages/Landing.tsx
vm: probably true but i feel like that'll be more work that isn't worth it right now, until we have solidified buttons

src/pages/Landing.tsx
vm: i can look into changing the colors here

vm:
the button is not simple to change, and i have an open question to design about the color. will change in diff pr

PR #5573, opened 2022-12-07
fix: use number formatting lib for tick bounds on LP positions

JFrankfurt:
uses `@uniswap/conedison/format` instead of current util, which this PR removes. (s/o to @vm for the no-unused-imports rule that flagged it for removal!)

https://uniswaplabs.atlassian.net/browse/WEB-2549

## After
image

## Before
image


yyip-dev (approved):
Woot! 🌮 this looks SO much better, very happy it's getting fixed.

yyip-dev:
@JFrankfurt there is a failing unit test, let's fix before merging, though i'm not sure if it's related or just being flaky?

```
console.error
Warning: React does not recognize the `scrollOverlay` prop on a DOM element.
If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `scrolloverlay` instead.
If you accidentally passed it from a parent component, remove it from the DOM element.
```

PR #5574, opened 2022-12-08
fix: stable trade currency

zzmp:
Uses a stable value for the trade's currencies, so that hooks depending on the currency are not rerendered. Practically, this fixes an issue where changing the input amount might reset approval state.


yyip-dev (approved):
Lgtm

PR #5576, opened 2022-12-08
fix: handle user rejection

zzmp:
Displays a more user-friendly error - "Transaction rejected" - when user rejects a UR swap. This is the same behavior as without UR.


yyip-dev (approved):
Neat, tested too and looks good

PR #5584, opened 2022-12-08
feat: add selected state for about steps

vm:
Screen Shot 2022-12-08 at 1 19 00 PM
Screen Shot 2022-12-08 at 1 18 56 PM

will add updated images in a future pr

### DIFF
--- a/src/pages/About/Step.tsx
+++ b/src/pages/About/Step.tsx
@@ -1,6 +1,7 @@
 import styled from 'styled-components/macro'
 
 const StyledStep = styled.div<{ isLast?: boolean }>`
+  cursor: pointer;
   display: flex;
   gap: 36px;
   padding: 24px 0;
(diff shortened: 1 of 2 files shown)


src/pages/About/index.tsx
```diff
@@ -149,10 +149,47 @@ const GetStarted = styled.div`
   margin-bottom: 100px;
 `
 
+const CARDS = [
+  {
+    to: '/swap',
```
zzmp, line 154: Should these be hashed routes (eg `#/swap`)?

src/pages/About/Step.tsx
```diff
@@ -31,18 +32,22 @@ const Step = ({
...
   index: number
   title: string
   description: string
   isLast?: boolean
+  onClick: () => void
+  expanded?: boolean
 }) => {
   return (
-    <StyledStep isLast={isLast}>
+    <StyledStep isLast={isLast} onClick={onClick}>
```
zzmp, line 46: Can `isLast` be moved to a css selector?

zzmp (approved):

src/pages/About/Step.tsx
vm: good point yes

src/pages/About/index.tsx
vm: gonna fix all the routing in another pr

PR #5588, opened 2022-12-08
fix: Revert "fix: textPrimary for landing page copy"

vm:
Reverts Uniswap/interface#5580

Design said it should always be white.


vm:
replaced by https://github.com/Uniswap/interface/pull/5589

PR #5594, opened 2022-12-08
feat: add page view traces for / and /about

vm:


yyip-dev (commented):
What do you mean by constants from analytics-events? Also, after adding tracing, how do we validate it is working?

vm:
I actually updated it, and I just tested it locally

yyip-dev:
@vm Nice! And out of curiosity how do we test analytics changes?

src/pages/Landing/index.tsx
```diff
@@ -152,7 +153,7 @@ export default function Landing() {
   if (landingPageFlag === LandingPageVariant.Control || !isOpen) return null
 
   return (
-    <>
+    <Trace page="landing-page" shouldLogImpression>
```
yyip-dev, line 156: Should this be referred to by a constant too, i.e. `PageName.LANDING_PAGE`?

yyip-dev (approved):

src/pages/Landing/index.tsx
vm: yes

vm:
I just console log it manually. AFAIK we don't have a test environment atm...

PR #5597, opened 2022-12-09
feat: dynamically add padding-top on /about

vm:
image
image

### DIFF
--- a/src/pages/About/index.tsx
+++ b/src/pages/About/index.tsx
@@ -1,5 +1,5 @@
 import { ButtonOutlined } from 'components/Button'
-import { useState } from 'react'
+import { useLayoutEffect, useRef, useState } from 'react'
 import { useIsDarkMode } from 'state/user/hooks'
 import styled from 'styled-components/macro'
 import { BREAKPOINTS } from 'theme'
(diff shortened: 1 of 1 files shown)


src/pages/About/index.tsx
```diff
@@ -121,12 +122,22 @@ const PoweredBySection = styled(Panels)`
 export default function About() {
   const isDarkMode = useIsDarkMode()
 
+  const titleRef = useRef<HTMLDivElement>(null)
+  const [titleHeight, setTitleHeight] = useState(0)
+  useLayoutEffect(() => {
+    if (titleRef.current) {
+      setTitleHeight(titleRef.current.clientHeight)
```
JFrankfurt, line 129: might consider scrollHeight here instead? not sure

JFrankfurt (approved):

src/pages/About/index.tsx
vm: i can't set scroll height unless there is enough padding to be able to scroll

vm: i think that would accomplish the opposite effect. it would scroll further down the page

PR #5605, opened 2022-12-09
fix: light mode cards for /about

vm:
Screen Shot 2022-12-08 at 9 51 50 PM
Screen Shot 2022-12-08 at 9 51 57 PM

### DIFF
--- a/src/pages/About/Card.tsx
+++ b/src/pages/About/Card.tsx
@@ -1,9 +1,11 @@
 import { Link } from 'react-router-dom'
+import { useIsDarkMode } from 'state/user/hooks'
 import styled from 'styled-components/macro'
 
-const StyledCard = styled.div`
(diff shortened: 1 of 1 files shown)


src/pages/About/Card.tsx
```diff
@@ -41,13 +44,15 @@ const Card = ({
   to: string
   external?: boolean
 }) => {
+  const isDarkMode = useIsDarkMode()
```
JFrankfurt, line 47: this makes me feel like we should have a theme.isDarkMode so that styled components can access it directly

JFrankfurt (approved):

src/pages/About/Card.tsx
vm: hmm yea... i have been thinking this is mostly an anti-pattern. i'm doing it right now because i don't want to push back on design so late but we shouldn't need to special case like this mostly?

PR #5606, opened 2022-12-09
feat: amplitude analytics for landing and about

vm:


JFrankfurt (approved):

src/pages/About/index.tsx
```diff
@@ -237,22 +237,37 @@ export default function About() {
...
               <ActionsContainer>
-                <InfoButton as="a" rel="noopener noreferrer" href="https://uniswap.org" target="_blank">
-                  Learn more<WrappedExternalArrow> ↗</WrappedExternalArrow>
-                </InfoButton>
-                <InfoButton as="a" rel="noopener noreferrer" href="https://docs.uniswap.org" target="_blank">
-                  Read docs<WrappedExternalArrow> ↗</WrappedExternalArrow>
-                </InfoButton>
+                <TraceEvent
+                  events={[BrowserEvent.onClick]}
+                  name={EventName.ELEMENT_CLICKED}
```
lynnshaoyu, line 242: The event name 'Element Clicked' seems way too generic to me? could we do something like 'Landing Page Element Clicked' so at least all these relevant landing page elements are grouped together when PMs are looking through amplitude

src/pages/About/index.tsx
```diff
@@ -237,22 +237,37 @@ export default function About() {
...
+                  </InfoButton>
+                </TraceEvent>
               </ActionsContainer>
             </Intro>
           </Panels>
           <CardGrid>
-            {CARDS.map(({ darkBackgroundImgSrc, lightBackgroundImgSrc, ...card }) => (
-              <Card
+            {CARDS.map(({ darkBackgroundImgSrc, lightBackgroundImgSrc, elementName, ...card }) => (
+              <TraceEvent
```
lynnshaoyu, line 263: These events are not logging because the browser event is happening inside the <Card /> component, pls move TraceEvent to inside <Card />

lynnshaoyu (changes requested):
.

src/pages/About/index.tsx
vm: will do

src/pages/About/index.tsx
vm: I talked with them and they want these to be extremely generic, since page is on there

src/pages/About/constants.ts
```diff
@@ -15,13 +17,15 @@ export const CARDS = [
...
     lightBackgroundImgSrc: lightSwapCardImgSrc,
+    elementName: ElementName.ABOUT_PAGE_NFTS_CARD,
   },
   {
     to: '/nfts',
     title: 'Trade NFTs',
     description: 'Buy and sell NFTs across marketplaces to find more listings at better prices.',
     darkBackgroundImgSrc: darkNftCardImgSrc,
     lightBackgroundImgSrc: lightNftCardImgSrc,
+    elementName: ElementName.ABOUT_PAGE_SWAP_CARD,
```
lynnshaoyu, line 28: you need to switch these... NFTs is using elementName for swap right now, and vice versa

lynnshaoyu (approved):
after switch event name looks good

PR #5608, opened 2022-12-09
fix: landing page font-size changes

vm:
Screen Shot 2022-12-09 at 10 14 18 AM

### DIFF
--- a/src/pages/Landing/index.tsx
+++ b/src/pages/Landing/index.tsx
@@ -41,9 +41,10 @@ const PageWrapper = styled.div<{ isDarkMode: boolean }>`
 const TitleText = styled.h1<{ isDarkMode: boolean }>`
   color: transparent;
   font-size: 36px;
-  font-weight: 600;
+  line-height: 44px;
+  font-weight: 500;
   text-align: center;
-  margin-bottom: 0px;
+  margin: 0 0 24px;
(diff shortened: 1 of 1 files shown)


src/pages/Landing/index.tsx
```diff
@@ -54,24 +55,27 @@ const TitleText = styled.h1<{ isDarkMode: boolean }>`
...
+    line-height: 72px;
   }
 `
 
 const SubText = styled.h3`
   color: ${({ theme }) => theme.textSecondary};
   font-size: 16px;
   line-height: 24px;
-  font-weight: 400;
+  font-weight: 500;
```
lynnshaoyu, line 71: i took a peek at the figma and I couldn't find where font-weight was specified to be 500, seems to be 400 in most places? 

If so, this Subtext has the same text specs as ThemedText.BodySecondary so maybe we could use that component instead and apply the extra styles to that?

lynnshaoyu (approved):
left one comment

src/pages/Landing/index.tsx
vm: the figma keeps changing sadly, so you can ignore that.

and it's not the same as bodysecondary because there are weird colors and stuff. it also kept changing so wanted to avoid using the style. after launch i'm gonna come back to refactor this usage, agreed

PR #5621, opened 2022-12-10
fix: hexlify tx value from SwapRouter

zzmp:
Hexlifies the value returned from universal-router-sdk, which is required for ethers to accept it as input. Fixes an error thrown by ethers (RE: un-hexlified data).
See Uniswap/universal-router-sdk#92 (fixed in Uniswap/universal-router-sdk#93 but not yet released).

See Uniswap/widgets#332.

### DIFF
--- a/src/hooks/useUniversalRouter.ts
+++ b/src/hooks/useUniversalRouter.ts
@@ -3,7 +3,7 @@ import { BigNumber } from '@ethersproject/bignumber'
 import { Trade } from '@uniswap/router-sdk'
 import { Currency, Percent, TradeType } from '@uniswap/sdk-core'
 import { SwapRouter, UNIVERSAL_ROUTER_ADDRESS } from '@uniswap/universal-router-sdk'
-import { FeeOptions } from '@uniswap/v3-sdk'
+import { FeeOptions, toHex } from '@uniswap/v3-sdk'
 import { useWeb3React } from '@web3-react/core'
 import { useCallback } from 'react'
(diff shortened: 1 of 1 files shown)


src/hooks/useUniversalRouter.ts
```diff
@@ -38,19 +38,13 @@ export function useUniversalRouterSwapCallback(
...
-              from: account,
-              to: UNIVERSAL_ROUTER_ADDRESS(chainId),
-              data,
-            }
+      const tx = {
+        from: account,
+        to: UNIVERSAL_ROUTER_ADDRESS(chainId),
+        data,
+        // TODO: universal-router-sdk returns a non-hexlified value.
+        ...(value && !isZero(value) ? { value: toHex(value) } : {}),
```
vm, line 46:
```ts
value: value && !isZero(value) ? toHex(value) : undefined
```

vm (approved):

src/hooks/useUniversalRouter.ts
zzmp: I don't think we want the property value at all. Using this syntax would have the property defined as undefined, instead of not defined at all.

vm: it's the same thing afaik?

zzmp: Not the same, because you'll have 'value' in tx be true (ie Object.getOwnPropertyDescriptor).

vm: agreed not to do this for now if its diff

PR #5622, opened 2022-12-10
fix: update unsupported list

ianlapham:
Add GOLD1 to unsupported list

### DIFF
--- a/src/constants/tokenLists/unsupported.tokenlist.json
+++ b/src/constants/tokenLists/unsupported.tokenlist.json
@@ -1164,6 +1164,13 @@
       "name": "Synth sXRP",
       "symbol": "sXRP",
       "decimals": 18
+    },
+    {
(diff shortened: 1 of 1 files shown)


vm (approved):

s4gdev:
@ianlapham @vm GOLD1 community has been keen to understand the reason for this action. If you have an official report/flagging requests for a potential Trademark/Copyrughts violation, it is only fair to share that with us so we could directly clarify this with the reporter, otherwise we think this is a unilateral decision by Uniswap in order to silence decentralized community projects purely for personal reasons. Hope to hear back from you.

s4gdev:
@vbuterin @karlfloersch please take a look at @haydenadams (and co @ianlapham @vm) blackball GOLD1 on a THE-Centralized Uniswap. @aGoldenBull

PR #5635, opened 2022-12-11
fix: make background header a background-image

vm:
Screen Shot 2022-12-11 at 4 32 22 PM


grabbou (approved):
Looks good! However, as discussed, let's do a follow-up next week to support srcSet as in this PR https://github.com/Uniswap/interface/pull/5379/files#diff-570ce6523637fa8997b8d27838679b5d80832917a501983a0f0033ce19ae0efcR31.

PR #5639, opened 2022-12-12
feat: close landing page when wallet connected

vm:
I would rather do this in some async fashion, but we don't have that set up in the app right now. I also think it can be confusing if there are random listeners all over, so maybe this is actually good to be explicit?


grabbou:
I think that's fine. I can't think of a better place to do it atm.

grabbou (approved):

PR #5642, opened 2022-12-12
fix: no pointer-events on Swap children on landing page

vm:
This makes sure that cursor pointer is always active while on the landing page, and doesn't allow hover states on any elements. In practice, the landing page's Swap is really just a button so this show's that change.


vm:
look at https://github.com/Uniswap/interface/pull/5654 instead

PR #5644, opened 2022-12-12
feat: styled footer links

vm:
Screen Shot 2022-12-11 at 11 25 18 PM
Screen Shot 2022-12-11 at 11 25 23 PM
Screen Shot 2022-12-11 at 11 25 41 PM


src/pages/About/index.tsx
```diff
@@ -27,7 +28,7 @@ const PageBackground = styled.div<{ isDarkMode: boolean }>`
   position: absolute;
   width: 100%;
   height: 100vh;
-  top: 0;
+  top: -72px;
```
grabbou, line 31: I think you can replace this with a navHeight from the theme?

src/pages/About/index.tsx
```diff
@@ -134,12 +135,42 @@ const Thumbnail = styled.img`
...
+  }
+`
+
+const FooterLink = styled.a`
+  display: flex;
+  align-items: center;
+  gap: 16px;
+  padding: 16px;
+  color: ${({ theme }) => theme.textPrimary};
+  border: 1px solid ${({ theme }) => theme.backgroundOutline};
```
grabbou, line 158: Have you tried using a ButtonOutline or similar style from the existing collection of components?

grabbou (approved):
Overall looks good, left two nits

src/pages/About/index.tsx
vm: yes

PR #5647, opened 2022-12-12
chore: migrate collection query to apollo

cbachmeier:
Migrate NFT Collection query to apollo. 
Replace Suspense where loaded with a conditional on `loading`


src/graphql/data/nft/Collection.ts
```diff
@@ -1,10 +1,9 @@
 import graphql from 'babel-plugin-relay/macro'
 import { GenieCollection, Trait } from 'nft/types'
-import { useEffect } from 'react'
-import { useLazyLoadQuery, useQueryLoader } from 'react-relay'
 
-import { CollectionQuery } from './__generated__/CollectionQuery.graphql'
+import { useCollectionQueryQuery } from '../__generated__/types-and-hooks'
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const collectionQuery = graphql`
```
cbachmeier, line 7: By using the generated Apollo hook for 1useCollectionQueryQuery1, I don't need to directly refer to the gql query as compared to doing `useQuery<CollectionQuery>` which causes a linting error. I'm taking any suggestions for a cleaner solution here.

justindomingue: one option is to could move to a separate `gql` file

src/graphql/data/nft/Collection.ts
```diff
@@ -1,10 +1,9 @@
...
   query CollectionQuery($addresses: [String!]!) {
```
justindomingue, line 8: probably more idiomatic to not add `Query` in the document node(, or remove the `Query` suffix from generated hooks so it doesn't read QueryQuery)

src/nft/components/explore/CarouselCard.tsx
```diff
@@ -235,9 +235,11 @@ const MARKETS_ENUM_TO_NAME = {
 }
 
 export const CarouselCard = ({ collection, onClick }: CarouselCardProps) => {
-  const gqlCollection = useCollectionQuery(collection.address)
+  const { data: gqlCollection, loading } = useCollectionQuery(collection.address)
 
-  return (
+  return loading ? (
```
tinaszheng, line 240: prefer guard-like returns like:

```
if (loading) return <LoadingCarouselCard />

return ...
```

src/graphql/data/nft/Collection.ts
```diff
@@ -87,26 +86,14 @@ const collectionQuery = graphql`
...
-export function useLoadCollectionQuery(address?: string | string[]): void {
-  const [, loadQuery] = useQueryLoader(collectionQuery)
-  useEffect(() => {
-    if (address) {
-      loadQuery({ addresses: Array.isArray(address) ? address : [address] })
-    }
-  }, [address, loadQuery])
-}
+export function useCollectionQuery(address: string): { data: GenieCollection; loading: boolean } {
+  const { data: queryData, loading } = useCollectionQueryQuery({
```
tinaszheng, line 90: `useCollectionQueryQuery` is kind of odd. id probably rename to `useCollectionQuery` and then the outer hook seems more semantic to name `useCollection` since that's what you're returning here

src/graphql/data/nft/Collection.ts
```diff
@@ -87,26 +86,14 @@ const collectionQuery = graphql`
...
 
-export function useLoadCollectionQuery(address?: string | string[]): void {
-  const [, loadQuery] = useQueryLoader(collectionQuery)
-  useEffect(() => {
-    if (address) {
-      loadQuery({ addresses: Array.isArray(address) ? address : [address] })
-    }
-  }, [address, loadQuery])
-}
+export function useCollectionQuery(address: string): { data: GenieCollection; loading: boolean } {
```
tinaszheng, line 89: is there a helper type for gql-like return types like these? in mobile have a generic defined like `https://github.com/Uniswap/mobile/blob/main/src/features/dataApi/types.ts#L23` that defines data, loading, and error

src/graphql/data/nft/Collection.ts
```diff
@@ -123,41 +110,44 @@ export function useCollectionQuery(address: string): GenieCollection {
...
-          return {
-            marketplace: market.marketplace?.toLowerCase() ?? '',
-            count: market.listings ?? 0,
-            floorPrice: market.floorPrice ?? 0,
+    data: {
+      address,
+      isVerified: queryCollection?.isVerified ?? undefined,
+      name: queryCollection?.name ?? undefined,
+      description: queryCollection?.description ?? undefined,
+      standard: queryCollection?.nftContracts ? queryCollection?.nftContracts[0]?.standard ?? undefined : undefined,
```
justindomingue, line 118:
```suggestion
      standard: queryCollection?.nftContracts?.[0]?.standard ?? undefined,
```

src/nft/pages/collection/index.tsx
```diff
@@ -121,90 +121,94 @@ const Collection = () => {
...
+                        ? `${collectionStats.bannerImageUrl}?w=${window.innerWidth}`
+                        : undefined
+                    }
+                    className={styles.bannerImage}
+                    background="none"
+                  />
+                </BannerWrapper>
+                <CollectionDescriptionSection>
+                  {collectionStats && (
+                    <CollectionStats stats={collectionStats || ({} as GenieCollection)} isMobile={isMobile} />
```
tinaszheng, line 144: do you need to do the || {} condition here? looks like if `collectionStats` is defined it should be `GenieCollection`

src/nft/pages/collection/index.tsx
```diff
@@ -121,90 +121,94 @@ const Collection = () => {
...
+                      <>
+                        {isMobile && (
+                          <MobileFilterHeader>
+                            <ThemedText.HeadlineSmall>Filter</ThemedText.HeadlineSmall>
+                            <IconWrapper onClick={() => setFiltersExpanded(false)}>
+                              <BagCloseIcon />
+                            </IconWrapper>
+                          </MobileFilterHeader>
+                        )}
+                        <Filters traitsByGroup={collectionStats?.traits ?? {}} />
```
tinaszheng, line 176: minor perf nit: use a constant like like `const EMPTY_OBJ = {}` here so it's not a new prop on ever render

src/graphql/data/nft/Collection.ts
```diff
@@ -87,26 +86,14 @@ const collectionQuery = graphql`
...
-export function useCollectionQuery(address: string): GenieCollection {
-  const queryData = useLazyLoadQuery<CollectionQuery>( // this will suspend if not yet loaded
-    collectionQuery,
-    { addresses: [address] },
-    { fetchPolicy: 'store-or-network' }
-  )
-
-  const queryCollection = queryData.nftCollections?.edges[0]?.node
+  const queryCollection = queryData?.nftCollections?.edges[0]?.node
   const market = queryCollection?.markets && queryCollection?.markets[0]
```
justindomingue, line 97:
```suggestion
  const market = && queryCollection?.markets?.[0]
```

src/graphql/data/nft/Collection.ts
```diff
@@ -123,41 +110,44 @@ export function useCollectionQuery(address: string): GenieCollection {
...
-    imageUrl: queryCollection?.image?.url ?? '',
-    twitterUrl: queryCollection?.twitterName ?? '',
-    instagram: queryCollection?.instagramName ?? undefined,
-    discordUrl: queryCollection?.discordUrl ?? undefined,
-    externalUrl: queryCollection?.homepageUrl ?? undefined,
-    rarityVerified: false, // TODO update when backend supports
-    // isFoundation: boolean, // TODO ask backend to add
+        : {},
+      traits,
+      marketplaceCount: queryCollection?.markets
```
justindomingue, line 134:
```suggestion
      marketplaceCount: market
```

should be enough?

src/graphql/data/nft/Collection.ts
```diff
@@ -123,41 +110,44 @@ export function useCollectionQuery(address: string): GenieCollection {
     })
   }
   return {
```
justindomingue, line 112: may be worth memoizing

src/nft/pages/collection/index.tsx
```diff
@@ -216,12 +220,12 @@ const Collection = () => {
   )
 }
 
+// TODO: Remove when assets query is migrated to Apollo
 // The page is responsible for any queries that must be run on initial load.
 // Triggering query load from the page prevents waterfalled requests, as lazy-loading them in components would prevent
 // any children from rendering.
 const CollectionPage = () => {
   const { contractAddress } = useParams()
-  useLoadCollectionQuery(contractAddress)
```
justindomingue, line 224: do you not want to preload this anymore?

src/graphql/data/nft/Collection.ts
cbachmeier: One issue is that Relay requires the query to include `Query` as the suffix. The solution during this migration is to exclude this file from Relay's gql generation.

src/graphql/data/nft/Collection.ts
cbachmeier: See comment here https://github.com/Uniswap/interface/pull/5647#discussion_r1048904357

src/nft/pages/collection/index.tsx
cbachmeier: This wasn't really preloading as it's run directly before the pageload so there wasn't any time saved.

src/graphql/data/nft/Collection.ts
cbachmeier: thoughts @zzmp ? I think I prefer colocating the query in the respective file

src/graphql/data/nft/Collection.ts
tinaszheng: well, it prevents you from having to type out the whole object every time. you can just pass in `GenieCollection` as the type argument

src/graphql/data/nft/Collection.ts
```diff
@@ -1,12 +1,12 @@
-import graphql from 'babel-plugin-relay/macro'
+import { gql } from '@apollo/client'
```
justindomingue, line 1: we haven't done it yet but fyi, to reach parity with relay i think you'll have to use the macro https://www.apollographql.com/docs/react/performance/babel/

src/nft/pages/collection/index.tsx
```diff
@@ -106,12 +105,16 @@ const Collection = () => {
...
 
+  if (loading) return <CollectionPageSkeleton />
+
   const toggleActivity = () => {
     isActivityToggled
       ? navigate(`/nfts/collection/${contractAddress}`)
       : navigate(`/nfts/collection/${contractAddress}/activity`)
   }
 
+  const EMPTY_TRAIT_OBJ = {}
```
justindomingue, line 116: probably better to define outside of react components so it's not a new object every render

justindomingue (approved):

src/graphql/data/nft/Collection.ts
cbachmeier: Thank you for this, significantly improved load time. Will make the change with the migration.

PR #5648, opened 2022-12-12
chore: migrate NFT details query to apollo

cbachmeier:
Migrate NFT Details Query to Apollo. Update Loading state and remove imports from Relay generated types.


aballerr (approved):

aballerr:
tested details with various collections, everything worked

src/graphql/data/nft/Details.ts
```diff
@@ -1,14 +1,12 @@
...
+import { CollectionInfoForAsset, GenieAsset, Markets, SellOrder } from 'nft/types'
 
-import { NftStandard } from '../__generated__/types-and-hooks'
-import { DetailsQuery } from './__generated__/DetailsQuery.graphql'
+import { NftAsset, useDetailsQuery } from '../__generated__/types-and-hooks'
 
-const detailsQuery = graphql`
-  query DetailsQuery($address: String!, $tokenId: String!) {
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const detailsQuery = gql`
```
justindomingue, line 8: i saw in the other pr you got rid of the variable assignment? can you do it here?

src/graphql/data/nft/Details.ts
```diff
@@ -93,92 +91,85 @@ const detailsQuery = graphql`
...
+  tokenId: string
+): { data: [GenieAsset, CollectionInfoForAsset]; loading: boolean } {
+  const { data: queryData, loading } = useDetailsQuery({
+    variables: {
       address,
       tokenId,
     },
-    { fetchPolicy: 'store-or-network' }
-  )
+    fetchPolicy: 'cache-and-network',
```
justindomingue, line 103: this should be `cache-first` if you want the same behavior i think

src/graphql/data/nft/Details.ts
```diff
@@ -93,92 +91,85 @@ const detailsQuery = graphql`
...
+  })
 
-  const asset = queryData.nftAssets?.edges[0]?.node
+  const asset = queryData?.nftAssets?.edges[0]?.node as NonNullable<NftAsset>
```
justindomingue, line 106:
```suggestion
  const asset = queryData?.nftAssets?.edges[0]?.node as NonNullable<NftAsset> | undefined
```

isn't that the actual type?

src/graphql/data/nft/Details.ts
```diff
@@ -93,92 +91,85 @@ const detailsQuery = graphql`
...
-        providers: asset?.rarities
-          ? asset?.rarities?.map((rarity) => {
-              return {
-                rank: rarity.rank ?? undefined,
-                score: rarity.score ?? undefined,
-                provider: 'Rarity Sniper',
-              }
-            })
-          : undefined,
+  return {
```
justindomingue, line 111: have you had issues with this hook running more than required? it will re-render consumers because it's a new object every time

src/graphql/data/nft/Details.ts
```diff
@@ -93,92 +91,85 @@ const detailsQuery = graphql`
...
+    data: [
+      {
+        id: asset?.id,
+        address,
+        notForSale: asset?.listings === null,
+        collectionName: asset?.collection?.name,
+        collectionSymbol: asset?.collection?.image?.url,
+        imageUrl: asset?.image?.url,
+        animationUrl: asset?.animationUrl,
+        marketplace: listing?.marketplace.toLowerCase() as unknown as Markets,
```
justindomingue, line 121: do you mean to call lowercase on a `Markets` object?

src/nft/pages/asset/Asset.tsx
```diff
@@ -38,9 +35,9 @@ const AssetPriceDetailsContainer = styled.div`
   }
 `
 
-const Asset = () => {
+const AssetPage = () => {
   const { tokenId = '', contractAddress = '' } = useParams()
-  const data = useDetailsQuery(contractAddress, tokenId)
+  const { data, loading } = useNftAssetDetails(contractAddress, tokenId)
 
   const [asset, collection] = useMemo(() => data ?? [], [data])
```
justindomingue, line 42: i'm not sure this memo actually does anything, + it returns an empty array which will be a diff reference anyway

src/graphql/data/nft/Details.ts
cbachmeier: cache-and-network is the desired behaviour here

src/nft/pages/asset/Asset.tsx
```diff
@@ -52,34 +49,20 @@ const Asset = () => {
         shouldLogImpression
       >
         {asset && collection ? (
```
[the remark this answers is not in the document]
cbachmeier: Asset and collection would be empty `[]` because of the useMemo which return truthy, a better solution would be to memo in the graphql hook and return the loading state earlier

src/nft/pages/asset/Asset.tsx
```diff
@@ -51,35 +49,17 @@ const Asset = () => {
         properties={{ collection_address: contractAddress, token_id: tokenId }}
         shouldLogImpression
       >
-        {asset && collection ? (
+        {!!asset && !!collection && (
```
justindomingue, line 52: hm i /think/ this may be problematic, if asset===true and collection===false, wont' this expression resolves to `false`, in some cases writing the string `false` to the dom?

cbachmeier: casting these params to Bools prevents the React compiler from writing them to the DOM. This can be tested on the preview by editting the collection address or tokenId to invalid values

PR #5653, opened 2022-12-12
chore: migrate NftBalanceQuery

cbachmeier:
Migrated NftBalance Query to Apollo. This required refactoring the WalletAsset type to have all params potentially be unidentified. I also made a change to the Codegen config so that egenerated types aren't wrapped in `Maybe` as opposed to just having value vs being undefined.


apollo-codegen.ts
```diff
@@ -11,6 +11,8 @@ const config: CodegenConfig = {
       plugins: ['typescript', 'typescript-operations', 'typescript-react-apollo'],
       config: {
         withHooks: true,
+        // This avoid all generated schemas being wrapped in Maybe https://the-guild.dev/graphql/codegen/plugins/typescript/typescript#maybevalue-string-default-value-t--null
```
justindomingue, line 14: is it safe though?

src/nft/components/profile/list/NFTListingsGrid.tsx
```diff
@@ -214,10 +214,14 @@ const PriceTextInput = ({
       >
         {focused ? (
           <>
-            <Row display={asset.lastPrice ? 'flex' : 'none'} marginRight="8">
-              LAST: {formatEth(asset.lastPrice)} ETH
-            </Row>
-            <Row display={asset.floorPrice ? 'flex' : 'none'}>FLOOR: {formatEth(asset.floorPrice)} ETH</Row>
+            {!!asset.lastPrice && (
+              <Row display={asset.lastPrice ? 'flex' : 'none'} marginRight="8">
+                LAST: {formatEth(asset.lastPrice)} ETH
```
justindomingue, line 219: should this be in `Trans`?

src/nft/components/profile/list/NFTListingsGrid.tsx
```diff
@@ -214,10 +214,14 @@ const PriceTextInput = ({
...
+              </Row>
+            )}
+            {!!asset.floorPrice && (
```
justindomingue, line 222: this wlil not render when `floorPrice===0`, is that intentional?

src/nft/utils/listNfts.ts
```diff
@@ -62,7 +62,7 @@ const getConsiderationItems = (
   creatorFee?: ConsiderationInputItem
 } => {
   const openSeaBasisPoints = OPENSEA_DEFAULT_FEE * INVERSE_BASIS_POINTS
-  const creatorFeeBasisPoints = asset.basisPoints
+  const creatorFeeBasisPoints = asset?.basisPoints ?? 0
```
justindomingue, line 65: i feel like we non-null cast a lot. Should we let those fields be null until we /need/ to display them?

src/graphql/data/apollo.ts
```diff
@@ -11,5 +12,13 @@ export const apolloClient = new ApolloClient({
     'Content-Type': 'application/json',
     Origin: 'https://app.uniswap.org',
   },
-  cache: new InMemoryCache(),
+  cache: new InMemoryCache({
+    typePolicies: {
+      Query: {
+        fields: {
+          nftBalances: relayStylePagination(),
```
thomasthachil, line 19: do you need to pass in any `keyArgs` here for caching? Depends on if/ how you do caching, but on mobile i had issues where the entries between wallets were referencing the same query cache unless i added `ownerAddress`.

thomasthachil (commented):
added context from change on mobile

apollo-codegen.ts
cbachmeier: Yes, all values can still be undefined and will require that check. This prevent redundant checking for both null and undefined

src/nft/components/profile/list/NFTListingsGrid.tsx
cbachmeier: Internationalization isn't part of this migration and is handled in its own epic

src/nft/components/profile/list/NFTListingsGrid.tsx
cbachmeier: yes this is intentional

src/nft/utils/listNfts.ts
cbachmeier: This is where it's needed to calculate creator royalties, in this case if the value is undefined we can assume that no royalty amount is set

src/nft/components/bag/profile/utils.ts
```diff
@@ -142,7 +142,7 @@ export const getListings = (sellAssets: WalletAsset[]): [CollectionRow[], Listin
   sellAssets.forEach((asset) => {
     asset.marketplaces?.forEach((marketplace: ListingMarket) => {
       const newListing = {
-        images: [asset.smallImageUrl, marketplace.icon],
+        images: [asset.smallImageUrl ?? '', marketplace.icon],
```
tinaszheng, line 145: can you have images accept `(string | undefined)[]` instead?

src/nft/utils/listNfts.ts
```diff
@@ -239,14 +241,14 @@ export async function signListing(
         price: parseEther(listingPrice.toString()),
         tokens: [
           {
-            token: asset.asset_contract.address,
+            token: asset.asset_contract.address ?? '',
```
tinaszheng, line 244: feels odd to have an empty string here be a token address 🤔 what's the expected behavior?

src/graphql/data/apollo.ts
cbachmeier: Try now, I've updated the fetch policy and able to switch wallets with no issues

src/nft/utils/listNfts.ts
cbachmeier: I basically just wanted the listing to break as we shouldn't be able to list without a tokenId or collectionAddress. I've moved the check for this earlier in the function

src/graphql/data/nft/NftBalance.ts
```diff
@@ -161,14 +127,19 @@ export function useNftBalanceQuery(
...
-    queryData
-  )
-  const walletAssets: WalletAsset[] = data.nftBalances?.edges?.map((queryAsset: NftBalanceQueryAsset) => {
-    const asset = queryAsset.node.ownedAsset
+    fetchPolicy: 'cache-and-network',
+  })
+
+  const hasNext = data?.nftBalances?.pageInfo?.hasNextPage
+  const loadMore = () =>
+    fetchMore({
```
thomasthachil, line 135: does this need an argument on how many to `fetch`? Believe backend has a default but might be good to explicity define the number of items in subsequent requests

cbachmeier: `fetchMore` will automatically use the previous params unless overwritten https://www.apollographql.com/docs/react/pagination/core-api/

justindomingue (approved):
migration looks okay to me

PR #5662, opened 2022-12-12
fix: show error when present (not approval)

zzmp:
If a `swapInputError` is present, shows that text instead of the approval button.
This codepath can be triggered by trying to swap an unapproved token for which you have insufficient balance to trade:
- _before_: shows approval flow
- _after_: shows "insufficient <TOKEN> balance" error

### DIFF
--- a/src/pages/Swap/index.tsx
+++ b/src/pages/Swap/index.tsx
@@ -796,7 +796,7 @@ export default function Swap({ className }: { className?: string }) {
                       </ButtonError>
                     </AutoColumn>
                   </AutoRow>
-                ) : permit.state === PermitState.PERMIT_NEEDED ? (
+                ) : isValid && permit.state === PermitState.PERMIT_NEEDED ? (
                   <ButtonYellow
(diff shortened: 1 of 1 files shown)


src/pages/Swap/index.tsx
```diff
@@ -796,7 +796,7 @@ export default function Swap({ className }: { className?: string }) {
                       </ButtonError>
                     </AutoColumn>
                   </AutoRow>
-                ) : permit.state === PermitState.PERMIT_NEEDED ? (
+                ) : isValid && permit.state === PermitState.PERMIT_NEEDED ? (
```
JFrankfurt, line 799: nit: had to look up what `isValid` is referencing to understand this line

JFrankfurt (approved):

src/pages/Swap/index.tsx
zzmp: it's not ideal but this file is a behemoth. i'm hoping to replace with widgets soon enough.

PR #5665, opened 2022-12-12
chore: migrate Assets Query to Apollo

cbachmeier:
Migrate Assets Query to Apollo. This supports pagination but does not setup polling which will be done seperately. This required making changes to the GenieAsset type to accept more optional params which involved making some changes to several parts  of the app. I've tested by checking out multiple collections (images, videos, music), completing a tenderly buy, and sweep.

### DIFF
--- a/src/graphql/data/nft/Asset.ts
+++ b/src/graphql/data/nft/Asset.ts
@@ -1,6 +1,5 @@
-/* eslint-disable import/no-unused-modules */
-import { gql } from '@apollo/client'
 import { parseEther } from 'ethers/lib/utils'
+import gql from 'graphql-tag'
 import { GenieAsset, Markets, Trait } from 'nft/types'
 import { wrapScientificNotation } from 'nft/utils'
 import { useCallback, useMemo } from 'react'
--- a/src/graphql/data/nft/Asset.ts
+++ b/src/graphql/data/nft/Asset.ts
@@ -131,19 +131,12 @@ function formatAssetQueryData(queryAsset: NftAssetEdge, totalCount?: number) {
     animationUrl: asset.animationUrl,
     marketplace: asset.listings?.edges[0]?.node?.marketplace?.toLowerCase() as unknown as Markets,
     name: asset.name,
-    priceInfo: asset.listings
-      ? {
-          ETHPrice: ethPrice,
-          baseAsset: 'ETH',
-          baseDecimals: '18',
-          basePrice: ethPrice,
-        }
-      : {
-          ETHPrice: '0',
-          baseAsset: 'ETH',
-          baseDecimals: '18',
-          basePrice: '0',
-        },
+    priceInfo: {
+      ETHPrice: ethPrice,
+      baseAsset: 'ETH',
+      baseDecimals: '18',
+      basePrice: ethPrice,
+    },
     susFlag: asset.suspiciousFlag,
     sellorders: asset.listings?.edges.map((listingNode) => {
       return {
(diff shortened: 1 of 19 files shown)


src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -592,35 +577,37 @@ export const CollectionNfts = ({ contractAddress, collectionStats, rarityVerifie
...
+          <CollectionNftsLoading />
+        ) : (
+          <InfiniteScroll
+            next={loadMore}
+            hasMore={hasNext ?? false}
+            loader={Boolean(hasNext && hasNfts) && <LoadingAssets />}
+            dataLength={collectionAssets?.length ?? 0}
+            style={{ overflow: 'unset' }}
+            className={hasNfts || loading ? styles.assetList : undefined}
+          >
```
JackShort, line 590: should the loading state be put inside of the infinite scroll? the classname is based on `hasNfts || loading` where loading will never be true there because `CollectionNftsLoading` is rendered instead

src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -592,35 +577,37 @@ export const CollectionNfts = ({ contractAddress, collectionStats, rarityVerifie
...
+            {collectionAssets?.length === 0 ? (
```
JackShort, line 591: could be `!hasNfts`

src/nft/queries/genie/RouteFetcher.ts
```diff
@@ -1,4 +1,6 @@
-import { GenieAsset, RouteResponse, TokenType } from '../../types'
+import { NftStandard } from 'graphql/data/__generated__/types-and-hooks'
+
+import { GenieAsset, RouteResponse } from '../../types'
```
JackShort, line 3: nit: would prefer `nft/types`

src/nft/types/common/common.ts
```diff
@@ -81,7 +82,7 @@ export interface Trait {
 export interface GenieAsset {
   id?: string // This would be a random id created and assigned by front end
   address: string
-  notForSale: boolean
+  notForSale?: boolean
```
JackShort, line 85: should we just cast it from gql as false if it is not provided?

src/nft/types/common/common.ts
```diff
@@ -93,17 +94,17 @@ export interface GenieAsset {
   sellorders?: SellOrder[]
   smallImageUrl?: string
   tokenId: string
-  tokenType: TokenType
+  tokenType?: NftStandard
```
JackShort, line 97: do you think using erc721 as the default token type would be a bad idea?

src/graphql/data/nft/Asset.ts
```diff
@@ -1,25 +1,32 @@
...
 import { wrapScientificNotation } from 'nft/utils'
-import { useCallback, useEffect, useMemo, useState } from 'react'
-import { fetchQuery, useLazyLoadQuery, usePaginationFragment, useQueryLoader, useRelayEnvironment } from 'react-relay'
+import { useMemo } from 'react'
 
-import { AssetPaginationQuery } from './__generated__/AssetPaginationQuery.graphql'
 import {
-  AssetQuery,
-  AssetQuery$variables,
+  AssetQueryQueryVariables,
```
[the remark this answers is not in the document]
cbachmeier: Requires some working around Relay during the migration see https://github.com/Uniswap/interface/pull/5647#discussion_r1048904357

src/nft/components/collection/CollectionNfts.tsx
cbachmeier: `CollectionNftsLoading` needs to be outside of the infinite scroll but we can remove `loading` from the conditional

src/nft/types/common/common.ts
cbachmeier: I want to avoid as many possible incorrect instantiations as possible. I think breaking is a better use experience than showing incorrect information

src/nft/types/common/common.ts
cbachmeier: Yeah I think it's better to be undefined in this case as per https://github.com/Uniswap/interface/pull/5665#discussion_r1048978571

JackShort (approved):
everything looks good on my end

PR #5668, opened 2022-12-12
fix: jumpy nav on safari mweb

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-2407?atlOrigin=eyJpIjoiNGFiZDRlOWMxNzRlNDVmYzlmMmQwYTllY2JmMmFjMTIiLCJwIjoiaiJ9

note: I opened the vercel build link on my mobile device and tested in safari. you can't see the bug in the safari desktop mobile developer testing env.

### DIFF
--- a/src/pages/Pool/index.tsx
+++ b/src/pages/Pool/index.tsx
@@ -165,7 +165,7 @@ function WrongNetworkCard() {
   const theme = useTheme()
 
   return (
-    <>
+    <div style={{ height: '100vh' }}>
       <PageWrapper>
         <AutoColumn gap="lg" justify="center">
           <AutoColumn gap="lg" style={{ width: '100%' }}>
@@ -189,7 +189,7 @@ function WrongNetworkCard() {
         </AutoColumn>
       </PageWrapper>
       <SwitchLocaleLink />
-    </>
+    </div>
   )
 }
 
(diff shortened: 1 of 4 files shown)


JFrankfurt:
on mobile safari the nav bar is now stuck in the middle of the screen for me when I scroll down 😬 

image

JFrankfurt (changes requested):
see comment

src/pages/Pool/index.tsx
```diff
@@ -165,7 +165,7 @@ function WrongNetworkCard() {
   const theme = useTheme()
 
   return (
-    <>
+    <div style={{ height: '100vh' }}>
       <PageWrapper>
```
JFrankfurt, line 169: it looks like `PageWrapper` already has the 100vh style applied--why is the wrapper necessary here?

src/pages/Pool/index.tsx
```diff
@@ -263,7 +263,7 @@ export default function Pool() {
 
   return (
     <Trace page={PageName.POOL_PAGE} shouldLogImpression>
-      <>
+      <div style={{ height: '100vh' }}>
```
JFrankfurt, line 266: same q re: PageWrapper having 100vh applied already

src/pages/Pool/index.tsx
lynnshaoyu: the PageWrapper is not the the outermost div, this extra div is necessary to include SwitchLocaleLink. Putting switchLocaleLink inside PageWrapper has other undesired effects on the Pool component styling

lynnshaoyu:
> on mobile safari the nav bar now sticks in place for me when I scroll down 😬
> 
> image

oof thanks for catching that last commit did NOT do what I thought it was gonna😞

JFrankfurt (approved):

PR #5682, opened 2022-12-13
chore: migrate token queries to Apollo

cbachmeier:
Migrated all token queries to Apollo. They turned out to be fairly interconnected so it didn't make sense to do them piecemeal. I've tested this by going to the top token and token details pages for several tokens, on different chains, with changing durations and compared them to the data/experience on the main app. Loading states work as expected and load times showed no noticeable difference. So logic needed to be modified so would appreciate folks with more familiarity with the pages affected to take a look.


src/graphql/data/Token.ts
```diff
@@ -1,8 +1,8 @@
-import graphql from 'babel-plugin-relay/macro'
 import { DEFAULT_ERC20_DECIMALS } from 'constants/tokens'
+import gql from 'graphql-tag'
```
justindomingue, line 2: should this be `@apollo/client` or did you set up precompiled queries?

src/graphql/data/Token.ts
```diff
@@ -48,23 +48,24 @@ export const tokenQuery = graphql`
         twitterName
         logoUrl
         tokens {
```
justindomingue, line 50: one diff with relay is apollo won't add `id` automatically, which makes it impossible to properly merge cache entries with same id

src/graphql/data/TopTokens.ts
```diff
@@ -112,45 +114,42 @@ function useFilteredTokens(tokens: NonNullable<TopTokens100Query['response']['to
...
-              current?.address && (map[current.address] = current?.market?.priceHistory?.filter(isPricePoint))
-          )
-          setSparklines(map)
-        },
-      })
-    return () => subscription.unsubscribe()
-  }, [chain, chainId, duration, environment])
+  const { data: sparklineData } = useTopTokensSparklineQuery({
+    variables: { duration, chain },
+    fetchPolicy: 'cache-and-network',
```
justindomingue, line 132: probably good to add comments explaining why for one-ofs

src/graphql/data/TopTokens.ts
```diff
@@ -112,45 +114,42 @@ function useFilteredTokens(tokens: NonNullable<TopTokens100Query['response']['to
...
+  })
 
-  useEffect(() => {
-    setSparklines({})
-  }, [duration])
+  const sparklines = useMemo(() => {
+    const unwrappedTokens = sparklineData?.topTokens?.map((topToken) => unwrapToken(chainId, topToken))
```
justindomingue, line 136: would it be cleaner to have chain be from `topToken` and then `sparklineData?.topTokens?.map(unwrapToken)`

src/graphql/data/TopTokens.ts
```diff
@@ -112,45 +114,42 @@ function useFilteredTokens(tokens: NonNullable<TopTokens100Query['response']['to
...
+    const map: SparklineMap = {}
+    unwrappedTokens?.forEach(
```
justindomingue, line 138: `reduce` instead?

src/graphql/data/util.tsx
```diff
@@ -34,16 +34,16 @@ export function isPricePoint(p: PricePoint | null): p is PricePoint {
...
+  [SupportedChainId.MAINNET]: Chain.Ethereum,
+  [SupportedChainId.GOERLI]: Chain.EthereumGoerli,
+  [SupportedChainId.POLYGON]: Chain.Polygon,
+  [SupportedChainId.POLYGON_MUMBAI]: Chain.Polygon,
+  [SupportedChainId.CELO]: Chain.Celo,
+  [SupportedChainId.CELO_ALFAJORES]: Chain.Celo,
+  [SupportedChainId.ARBITRUM_ONE]: Chain.Arbitrum,
+  [SupportedChainId.ARBITRUM_RINKEBY]: Chain.Arbitrum,
+  [SupportedChainId.OPTIMISM]: Chain.Optimism,
+  [SupportedChainId.OPTIMISM_GOERLI]: Chain.Optimism,
```
justindomingue, line 46: type may be easier to work with (narrower) with `as const`

```
{
  [SupportedChainId.MAINNET]: Chain.Ethereum,
  ...
} as const
```

justindomingue (approved):
overall lgtm in terms of migration

would recommend adding `id` to each entity ([graphql-eslint](https://github.com/B2o5T/graphql-eslint) can help enforce it (needs some extra parser options though)

src/graphql/data/Token.ts
cbachmeier: I set up the precompiled queries. Great runtime performance bump

src/graphql/data/TopTokens.ts
cbachmeier: I don't want to change existing Token logic here, to avoid scope creep this PR should be judged on Relay -> Apollo changes

src/graphql/data/TopTokens.ts
cbachmeier: See comment on scope creep.

src/graphql/data/util.tsx
cbachmeier: See comment on scope creep.

cbachmeier:
Discussed id offline and will add run linter after all queries have been migrated to knock them all out at once

src/graphql/data/TopTokens.ts
cbachmeier: I actually want to update this to be the default and will do this separately

src/components/Tokens/TokenTable/TokenTable.tsx
```diff
@@ -100,7 +100,7 @@ export default function TokenTable({ setRowCount }: { setRowCount: (c: number) =
       <GridContainer>
         <HeaderRow />
         <TokenDataContainer>
-          {tokens.map(
+          {tokens?.map(
```
zzmp, line 103: This change should should not be necessary. It should be typed as non-nullable after L85.

src/graphql/data/Token.ts
```diff
@@ -48,23 +48,24 @@ export const tokenQuery = graphql`
...
-          address @required(action: LOG)
+          chain
+          address
         }
       }
     }
   }
 `
-export type { Chain, TokenQuery } from './__generated__/TokenQuery.graphql'
+// eslint-disable-next-line import/no-unused-modules
```
zzmp, line 58: This should not be necessary.

src/graphql/data/Token.ts
```diff
@@ -13,14 +13,14 @@ The difference between Token and TokenProject:
     TokenMarket is per-chain market data for contracts pulled from the graph.
     TokenProjectMarket is aggregated market data (aggregated over multiple dexes and centralized exchanges) that we get from coingecko.
 */
-export const tokenQuery = graphql`
-  query TokenQuery($contract: ContractInput!) {
+gql`
+  query Token($contract: ContractInput!) {
     tokens(contracts: [$contract]) {
-      id @required(action: LOG)
+      id
```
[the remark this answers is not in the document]
cbachmeier: Apollo doesn't support Relay `@required` directive, we would need to create our own custom directive https://www.apollographql.com/docs/apollo-server/schema/directives/

src/graphql/data/Token.ts
```diff
@@ -48,23 +48,24 @@ export const tokenQuery = graphql`
...
-export type TokenQueryData = NonNullable<TokenQuery$data['tokens']>[number]
+export type TokenQueryData = NonNullable<TokenQuery['tokens']>[number]
 
 // TODO: Return a QueryToken from useTokenQuery instead of TokenQueryData to make it more usable in Currency-centric interfaces.
 export class QueryToken extends WrappedTokenInfo {
   constructor(data: NonNullable<TokenQueryData>, logoSrc?: string) {
     super({
       chainId: CHAIN_NAME_TO_CHAIN_ID[data.chain],
-      address: data.address,
+      address: data.address ?? '',
```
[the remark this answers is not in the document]
cbachmeier: The real solution here would be to have data team update the schema to make this field non-nullable instead of enforcing non-null values on the FE.

src/graphql/data/TopTokens.ts
```diff
@@ -112,45 +114,40 @@ function useFilteredTokens(tokens: NonNullable<TopTokens100Query['response']['to
 
 // Number of items to render in each fetch in infinite scroll.
 export const PAGE_SIZE = 20
-
-export type TopToken = NonNullable<NonNullable<TopTokens100Query['response']>['topTokens']>[number]
 export type SparklineMap = { [key: string]: PricePoint[] | undefined }
+export type TopToken = NonNullable<NonNullable<TopTokens100Query>['topTokens']>[number]
+
 interface UseTopTokensReturnValue {
```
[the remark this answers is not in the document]
cbachmeier: this type is used in several other files

src/pages/TokenDetails/index.tsx
```diff
@@ -26,35 +24,35 @@ export default function TokenDetailsPage() {
...
       setTimePeriod(t)
     },
-    [contract, loadPriceQuery, setTimePeriod]
+    [setTimePeriod]
   )
 
-  if (!tokenQueryReference) {
-    return <TokenDetailsPageSkeleton />
-  }
+  if (!tokenData || tokenDataLoading) return <TokenDetailsPageSkeleton />
```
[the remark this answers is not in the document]
cbachmeier: This will allow us to eventually remove the suspense state as loading state can be managed by the returned state bool

src/pages/TokenDetails/index.tsx
```diff
@@ -24,35 +24,28 @@ export default function TokenDetailsPage() {
     [chain, isNative, pageChainId, timePeriod, tokenAddress]
   )
 
-  const { data: tokenData, loading: tokenDataLoading } = useTokenQuery({
+  const { data: tokenQuery, loading: tokenDataLoading } = useTokenQuery({
```
zzmp, line 27: `tokenQueryLoading`, not `tokenDataLoading`

src/components/Tokens/TokenDetails/ChartSection.tsx
```diff
@@ -1,62 +1,59 @@
...
     <Suspense fallback={<LoadingChart />}>
       <ChartContainer>
-        <Chart priceQueryReference={priceQueryReference} refetchTokenPrices={refetchTokenPrices} />
+        <Chart tokenPriceQuery={tokenPriceQuery} onChangeTimePeriod={onChangeTimePeriod} />
       </ChartContainer>
     </Suspense>
   )
 }
 
 export type RefetchPricesFunction = (t: TimePeriod) => void
```
zzmp, line 48: Rename this to `OnChangeTimePeriod`, or remove it entirely and define it inline here and in TokenDetails/index.ts.

src/graphql/data/Token.ts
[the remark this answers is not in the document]
zzmp: Still, I'm afraid this might cause a regression, because `address` is assumed to be valid here.
At the callsite in TokenDetails L77, address is checked to be defined. Could you just pass that address directly into the constructor (add it as a first parameter)?

src/pages/TokenDetails/index.tsx
```diff
@@ -26,35 +24,28 @@ export default function TokenDetailsPage() {
...
+      duration,
     },
-    [contract, loadPriceQuery, setTimePeriod]
-  )
+  })
 
-  if (!tokenQueryReference) {
-    return <TokenDetailsPageSkeleton />
-  }
+  if (!tokenQuery || tokenDataLoading) return <TokenDetailsPageSkeleton />
```
zzmp, line 40: In what case would `tokenQuery && tokenDataLoading`? I'm afraid this case might an the invalid case, where you would not want to show a skeleton because you have data but it is stale.

zzmp (changes requested):

src/pages/TokenDetails/index.tsx
cbachmeier: When the time period is changed would we want to show the stale data in this case or would we want to show the skeleton to inform the user that we are updating?

src/graphql/data/Token.ts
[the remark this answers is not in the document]
cbachmeier: Okay I'm passing in the address and use it as the fallback in case data.address is undefined

src/graphql/data/Token.ts
```diff
@@ -62,10 +62,10 @@ export type TokenQueryData = NonNullable<TokenQuery['tokens']>[number]
 
 // TODO: Return a QueryToken from useTokenQuery instead of TokenQueryData to make it more usable in Currency-centric interfaces.
 export class QueryToken extends WrappedTokenInfo {
-  constructor(data: NonNullable<TokenQueryData>, logoSrc?: string) {
+  constructor(address: string, data: NonNullable<TokenQueryData>, logoSrc?: string) {
     super({
       chainId: CHAIN_NAME_TO_CHAIN_ID[data.chain],
-      address: data.address ?? '',
+      address: data.address ?? address,
```
zzmp, line 68: This should just be `address`

src/pages/TokenDetails/index.tsx
zzmp: skeleton. what about if the data is refreshed - will this ever happen?

zzmp (approved):
down to nits. lgtm!

src/pages/TokenDetails/index.tsx
cbachmeier: The way it is currently setup, no. If we setup polling or a manual refresh function then it could happen.

cbachmeier: In that case though I think there is still value is showing a skeleton state to show to the user that the data is being refreshed

PR #5686, opened 2022-12-14
test: adding in some basic token tests

aballerr:
1. adding in token list test and going to token details
2. reverting timeout
3. moving walletdrop down back to primary page

### DIFF
--- a/cypress/e2e/token.test.ts
+++ b/cypress/e2e/token.test.ts
@@ -1,0 +1,20 @@
+import { getTestSelector, getTestSelectorStarsWith } from '../utils'
+
+describe('Testing nfts', () => {
+  before(() => {
+    cy.visit('/')
+  })
+
+  it('should load nft leaderboard', () => {
+    cy.visit('/tokens/ethereum')
+    cy.get(getTestSelectorStarsWith('token-table')).its('length').should('be.gte', 25)
+  })
+
+  it('should load go to ethereum token and return to token list page', () => {
+    cy.visit('/tokens/ethereum')
+    cy.get(getTestSelector('token-table-row-Ether')).first().click()
+    cy.get(getTestSelector('token-details-stats')).should('exist')
+    cy.get(getTestSelector('token-details-return-button')).click()
+    cy.get(getTestSelectorStarsWith('token-table')).its('length').should('be.gte', 25)
+  })
+})
--- a/cypress/e2e/token.test.ts
+++ b/cypress/e2e/token.test.ts
@@ -1,20 +1,20 @@
-import { getTestSelector, getTestSelectorStarsWith } from '../utils'
+import { getTestSelector, getTestSelectorStartsWith } from '../utils'
 
-describe('Testing nfts', () => {
+describe('Testing tokens on uniswap page', () => {
   before(() => {
     cy.visit('/')
   })
 
-  it('should load nft leaderboard', () => {
+  it('should load token leaderboard', () => {
     cy.visit('/tokens/ethereum')
-    cy.get(getTestSelectorStarsWith('token-table')).its('length').should('be.gte', 25)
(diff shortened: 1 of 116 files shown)


cypress/e2e/wallet-dropdown.test.ts
```diff
@@ -2,7 +2,7 @@ import { getTestSelector } from '../utils'
 
 describe('Wallet Dropdown', () => {
   beforeEach(() => {
-    cy.visit('/pool')
+    cy.visit('/')
```
JFrankfurt, line 5: @vm this is undoing the recent change to run this on /pool because the prior occlusion happening due to landing page updates -- are we sure this is the right change to make?

vm: yea we should revert this change

cypress/e2e/token.test.ts
```diff
@@ -0,0 +1,20 @@
+import { getTestSelector, getTestSelectorStarsWith } from '../utils'
+
+describe('Testing nfts', () => {
```
vm, line 3: change name

cypress/utils/index.ts
```diff
@@ -1 +1,3 @@
 export const getTestSelector = (selectorId: string) => `[data-testid=${selectorId}]`
+
+export const getTestSelectorStarsWith = (selectorId: string) => `[data-testid^=${selectorId}]`
```
vm, line 3: typo

cypress/e2e/token.test.ts
```diff
@@ -0,0 +1,20 @@
...
+  })
+
+  it('should load nft leaderboard', () => {
+    cy.visit('/tokens/ethereum')
+    cy.get(getTestSelectorStarsWith('token-table')).its('length').should('be.gte', 25)
+  })
+
+  it('should load go to ethereum token and return to token list page', () => {
+    cy.visit('/tokens/ethereum')
+    cy.get(getTestSelector('token-table-row-Ether')).first().click()
```
vm, line 15: you can do this just based on the text name within table instead of needing this test id

cypress/e2e/token.test.ts
```diff
@@ -0,0 +1,20 @@
...
+    cy.get(getTestSelector('token-details-stats')).should('exist')
```
vm, line 16: same here, should check that certain text exists vs checking for this element

cypress/e2e/token.test.ts
vm, line 15: why do you need to select first?

vm (approved):
after addressing comments

cypress/e2e/wallet-dropdown.test.ts
```diff
@@ -2,7 +2,7 @@ import { getTestSelector } from '../utils'
 
 describe.skip('Wallet Dropdown', () => {
   beforeEach(() => {
-    cy.visit('/pool')
+    cy.visit('/swap')
```
vm, line 5: don't need this change here anymore

cypress/e2e/token.test.ts
```diff
@@ -0,0 +1,20 @@
+import { getTestSelector, getTestSelectorStartsWith } from '../utils'
+
+describe('Testing tokens on uniswap page', () => {
+  before(() => {
+    cy.visit('/')
+  })
+
+  it('should load token leaderboard', () => {
+    cy.visit('/tokens/ethereum')
+    cy.get(getTestSelectorStartsWith('token-table')).its('length').should('be.gte', 25)
```
vm, line 10: why gte?

aballerr: I don't want to block the frontend if backend is failing to return 100 assets

vm: hm yea i see

vm (approved):

PR #5689, opened 2022-12-15
fix: await syncing allowance to update permit state

zzmp:
Add a `pending` state to `permit` which, once an approval is submitted, will flip to `true` and not flip back to `false` until the allowance has re-synced. This ensures that the UI stays in a pending state until an approval is processed, without re-prompting for approval.

### DIFF
--- a/src/hooks/usePermit2.ts
+++ b/src/hooks/usePermit2.ts
@@ -26,7 +26,7 @@ export enum PermitState {
 
 export interface Permit {
   state: PermitState
-  pending?: boolean
+  isSyncing?: boolean
   signature?: PermitSignature
   callback?: () => Promise<{
     response: ContractTransaction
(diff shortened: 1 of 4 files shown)


src/hooks/usePermit2.ts
```diff
@@ -19,22 +26,41 @@ export enum PermitState {
...
+  callback?: () => Promise<{
     response: ContractTransaction
     info: ApproveTransactionInfo
   } | void>
 }
 
 export default function usePermit(amount?: CurrencyAmount<Token>, spender?: string): Permit {
   const { account } = useWeb3React()
-  const tokenAllowance = useTokenAllowance(amount?.currency, account, PERMIT2_ADDRESS)
+  const { amount: tokenAllowance, syncing } = useTokenAllowance(amount?.currency, account, PERMIT2_ADDRESS)
```
vm, line 39: `isSyncing`

src/hooks/usePermit2.ts
```diff
@@ -97,6 +120,6 @@ export default function usePermit(amount?: CurrencyAmount<Token>, spender?: stri
         return { state: PermitState.PERMITTED, signature }
       }
     }
-    return { state: PermitState.PERMIT_NEEDED, callback }
-  }, [amount, callback, isAllowed, isPermitted, isSigned, permitAllowance, signature, tokenAllowance])
+    return { state: PermitState.PERMIT_NEEDED, pending: syncState !== SyncState.SYNCED, callback }
```
vm, line 123: isPending?

vm: shouldn't this be `SyncState.PENDING`

src/hooks/useTokenAllowance.ts
```diff
@@ -8,16 +8,23 @@ import { calculateGasMargin } from 'utils/calculateGasMargin'
...
-  const allowance = useSingleCallResult(contract, 'allowance', inputs).result
+  const { result, syncing } = useSingleCallResult(contract, 'allowance', inputs)
 
-  return useMemo(
-    () => (token && allowance ? CurrencyAmount.fromRawAmount(token, allowance.toString()) : undefined),
-    [token, allowance]
-  )
+  return useMemo(() => {
+    const amount = token && result && CurrencyAmount.fromRawAmount(token, result.toString())
+    return { amount, syncing }
```
vm, line 26: `tokenAllowance` instead of amount?

src/hooks/usePermit2.ts
zzmp: I'd like to leave as `syncing` to match the structure of `CallState` (from the multihook library), which is the most relevant prior work with syncable results.

src/hooks/usePermit2.ts
zzmp: Left an inline comment: we want the UI to show pending unless we are synced.

src/hooks/useTokenAllowance.ts
```diff
@@ -8,16 +8,23 @@ import { calculateGasMargin } from 'utils/calculateGasMargin'
...
+  spender?: string
+): {
+  tokenAllowance: CurrencyAmount<Token> | undefined
+  isSyncing: boolean
+} {
   const contract = useTokenContract(token?.address, false)
 
   const inputs = useMemo(() => [owner, spender], [owner, spender])
-  const allowance = useSingleCallResult(contract, 'allowance', inputs).result
+  const { result, syncing: isSyncing } = useSingleCallResult(contract, 'allowance', inputs)
```
vm, line 22: `isSyncing` here too

vm: unclear what result means here?

vm (approved):
LGTM

src/hooks/useTokenAllowance.ts
zzmp: `useSingleCallResult` is a wrapper around the multicall library, which returns a `CallState` object that includes `syncing`, not `isSyncing`. I do not want to touch it, as it is not modified from the library.

NB: I originally used `syncing` and `pending` instead of `isSyncing` because it matched the syntax exported by `CallState`.

PR #5692, opened 2022-12-15
fix: allow landing page scroll

vm:
Avoids the state for now where on mobile the button isn't clickable


src/pages/Landing/index.tsx
```diff
@@ -172,18 +171,6 @@ export default function Landing() {
 
   const landingPageFlag = useLandingPageFlag()
```
just-toby, line 172: should we remove this now?

yyip-dev (approved):

just-toby (approved):
🛳️

src/pages/Landing/index.tsx
vm: i have a pr for it

vm: this default to enabled for now, and i have a diff pr to remove completely

PR #5698, opened 2022-12-15
chore: remove top-level Trace component

grabbou:
Each top-level page already has a Trace component. Top-level code was not only reporting duplicate entries for most pages, but also returned `undefined` for some of the recently created pages (such as `Landing`).

Going forward, let's keep it at a page level.

### DIFF
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -1,5 +1,5 @@
-import { initializeAnalytics, OriginApplication, sendAnalyticsEvent, Trace, user } from '@uniswap/analytics'
-import { CustomUserProperties, EventName, getBrowser, PageName } from '@uniswap/analytics-events'
+import { initializeAnalytics, OriginApplication, sendAnalyticsEvent, user } from '@uniswap/analytics'
+import { CustomUserProperties, EventName, getBrowser } from '@uniswap/analytics-events'
 import Loader from 'components/Loader'
 import TopLevelModals from 'components/TopLevelModals'
 import { useFeatureFlagsIsLoaded } from 'featureFlags'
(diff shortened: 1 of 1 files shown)


src/pages/App.tsx
```diff
@@ -119,29 +119,6 @@ const Marginer = styled.div`
...
-function getCurrentPageFromLocation(locationPathname: string): PageName | undefined {
-  switch (true) {
-    case locationPathname.startsWith('/swap'):
-      return PageName.SWAP_PAGE
-    case locationPathname.startsWith('/vote'):
-      return PageName.VOTE_PAGE
-    case locationPathname.startsWith('/pool'):
-      return PageName.POOL_PAGE
-    case locationPathname.startsWith('/tokens'):
-      return PageName.TOKENS_PAGE
```
zzmp, line 131: This must be added as a Trace to TokenDetails.

src/pages/App.tsx
```diff
@@ -119,29 +119,6 @@ const Marginer = styled.div`
...
-    case locationPathname.startsWith('/nfts/profile'):
-      return PageName.NFT_PROFILE_PAGE
-    case locationPathname.startsWith('/nfts/asset'):
-      return PageName.NFT_DETAILS_PAGE
-    case locationPathname.startsWith('/nfts/collection'):
-      return PageName.NFT_COLLECTION_PAGE
```
zzmp, line 137: This is added in the page, and added _again_ in its components. It should be removed from the components (explicitly adding the `trace` to the `eventProperties` is redundant, and should be done automatically by the analytics library).

src/pages/App.tsx
```diff
@@ -119,29 +119,6 @@ const Marginer = styled.div`
   margin-top: 5rem;
 `
 
-function getCurrentPageFromLocation(locationPathname: string): PageName | undefined {
```
zzmp, line 122: A lot of these pages have different variants, which either are not covered, or should probably be covered under a different name.

As part of removing the trace in App, could you audit these components, and determine if they should have unique PageName's for analytics?
- `/vote`: there are separate `Landing`, `VotePage`, and `CreateProposal` pages, but they all use `PageName.VOTE_PAGE`.
- `/pool`: there are separate `PoolFinder`, `PoolV2`, `Pool`, and `PositionPage` pages, but they all use `PageName.POOL_PAGE`.
- `/tokens`: there are separate `Tokens` and `TokenDetails` pages, but they both use `PageName.TOKENS_PAGE`.
- `PageName.NFT_COLLECTION_PAGE`: this is included through `Collection`, but also added through `ActivityCells.tsx` and `CollectionAsset.tsx`. The latter two should be removed.

zzmp (changes requested):

PR #5706, opened 2022-12-15
fix: mark permit not syncing if not permitted

zzmp:


src/hooks/usePermit2.ts
```diff
@@ -81,9 +81,8 @@ export default function usePermit(amount?: CurrencyAmount<Token>, spender?: stri
 
   // Permit2 should be marked syncing from the time approval is submitted (pending) until it is
   // synced in tokenAllowance, to avoid re-prompting the user for an already-submitted approval.
-  // It should *not* be marked syncing if not permitted, because the user must still take action.
   const [syncState, setSyncState] = useState(SyncState.SYNCED)
-  const isSyncing = isPermitted || isSigned ? false : syncState !== SyncState.SYNCED
+  const isSyncing = syncState !== SyncState.SYNCED
```
vm, line 85: `isSynced`

zzmp: `!==` - this is for `isSyncing`, because all of the downstream checks check for `isApprovalPending`

src/hooks/usePermit2.ts
```diff
@@ -117,13 +116,12 @@ export default function usePermit(amount?: CurrencyAmount<Token>, spender?: stri
...
-        return { state: PermitState.PERMITTED }
-      } else if (isSigned) {
-        return { state: PermitState.PERMITTED, signature }
-      }
+    } else if (!(isPermitted || isSigned)) {
+      return { state: PermitState.PERMIT_NEEDED, callback }
+    } else if (!isAllowed) {
+      return { state: PermitState.PERMIT_NEEDED, isSyncing, callback }
+    } else {
+      return { state: PermitState.PERMITTED, signature: isPermitted ? undefined : signature }
```
vm, line 124: how could the PermitState be PERMITTED and then isPermitted false?

src/hooks/usePermit2.ts
```diff
@@ -117,13 +116,12 @@ export default function usePermit(amount?: CurrencyAmount<Token>, spender?: stri
...
-    } else if (isAllowed) {
-      if (isPermitted) {
-        return { state: PermitState.PERMITTED }
-      } else if (isSigned) {
-        return { state: PermitState.PERMITTED, signature }
-      }
+    } else if (!(isPermitted || isSigned)) {
+      return { state: PermitState.PERMIT_NEEDED, callback }
+    } else if (!isAllowed) {
+      return { state: PermitState.PERMIT_NEEDED, isSyncing, callback }
```
vm, line 122: why does pass in `isSyncing` while the above case does not?

vm (approved):

PR #5707, opened 2022-12-15
fix: layout padding/margin & overflow

grabbou:
While working on a different PR, I noticed we have some issues with paddings/margins and height. TL;DR what happens is that sometimes our container would overflow. This is especially noticeable on full screen layouts.

---

This PR automatically reverts https://github.com/Uniswap/interface/commit/082591d5dd644b1a627372ab43465bd38834c410 and applies a different fix (`position: sticky` should be replaced with `position: fixed` for the bottom navigation bar).

> An element with position: sticky; is positioned based on the user's scroll position. A sticky element toggles between relative and fixed, depending on the scroll position. It is positioned relative until a given offset position is met in the viewport - then it "sticks" in place (like position:fixed).

By having a bottom navigation bar always `fixed`, we don't have to set height of the container to `100vh` or `100%` anywhere in the codebase, it will always render at the bottom. That also makes sure it never "jumps" in all browsers, including mobile Safari.


src/components/Popups/index.tsx
```diff
@@ -74,16 +72,18 @@ export default function Popups() {
...
-        <MobilePopupInner>
-          {activePopups // reverse so new items up front
-            .slice(0)
-            .reverse()
-            .map((item) => (
-              <PopupItem key={item.key} content={item.content} popKey={item.key} removeAfterMs={item.removeAfterMs} />
-            ))}
-        </MobilePopupInner>
-      </MobilePopupWrapper>
+      {activePopups?.length > 0 && (
```
grabbou, line 75: This change fixes artificial gap caused by `padding-top` when there are no popups:

Screenshot 2022-12-15 at 14 25 10

No changes when actual popup is displayed (note styling is off, but this is not caused by this PR):

Screenshot 2022-12-15 at 14 34 32

src/pages/App.tsx
```diff
@@ -65,29 +65,18 @@ initializeAnalytics(ANALYTICS_DUMMY_KEY, OriginApplication.INTERFACE, {
   isProductionEnv: isProductionEnv(),
 })
 
-const AppWrapper = styled.div`
```
grabbou, line 68: We don't need to set a `height: 100%` on a top-level container at all. This was also causing a bottom overflow caused by a `position: sticky` of an inner `MobileBottomBar` component + `BodyWrapper` having a `height: 100%` either.

src/pages/App.tsx
```diff
@@ -65,29 +65,18 @@ initializeAnalytics(ANALYTICS_DUMMY_KEY, OriginApplication.INTERFACE, {
...
   flex-direction: column;
   width: 100%;
-  height: 100%;
-  padding: 72px 0px 0px 0px;
+  padding: ${({ theme }) => theme.navHeight}px 0px 5rem 0px;
   align-items: center;
   flex: 1;
-  ${({ theme }) => theme.deprecated_mediaWidth.deprecated_upToSmall`
-    padding: 52px 0px 16px 0px;
-  `};
```
grabbou, line 85: We do not need a different padding on mobile/desktop, because in all cases, the top navigation bar is the same (72px). 

Smaller padding was probably set by a mistake as a quick fix to overcome issues caused by a `MobilePopUp` (see PR description). Adding `20px` to `52px` makes up `72px`, so all makes sense!

This change makes it very explicit about paddings. 

Also, note that `5rem` was added to remove artificial `Marginer` component.

src/pages/Pool/index.tsx
```diff
@@ -165,7 +165,7 @@ function WrongNetworkCard() {
   const theme = useTheme()
 
   return (
-    <div style={{ height: '100vh' }}>
+    <>
```
grabbou, line 168: Empty `div` was also causing issues with paddings - now, `Pool` renders similar regardless whether it's invalid network or not.

Screenshot 2022-12-15 at 14 22 44

grabbou: Previously, padding was a bit bigger on the sides, but this keeps it aligned with other pages:

Screenshot 2022-12-15 at 14 05 06

Other pages:

Screenshot 2022-12-15 at 14 49 30

I still think we should unify horizontal paddings because they're slightly off between Pool/Swap/Tokens, but I would leave it for another PR.

lynnshaoyu (approved):
css wizard! ty for fixing

PR #5710, opened 2022-12-15
fix: use render function for SearchBar placeholder translation

just-toby:
this placeholder text wasn't updating properly when switching locales.

i think this is because it's a text attribute for a Pure Component, like [the example from the docs](https://lingui.js.org/guides/optimized-components.html)

we could probably also solve by extracting this component and wrapping with `withI18n`, as suggested there. but I didn't want to add another `Input` component to the codebase when this simple fix also works. We should standardize to one component as part of the [Web Component Library](https://www.notion.so/uniswaplabs/Web-Component-Library-198412dd549d43429382e687d7020c44) work.

before: 

https://user-images.githubusercontent.com/66155195/207982226-b9029d81-5f2f-4bfb-8a09-266a2429d3c1.mov

after:

https://user-images.githubusercontent.com/66155195/207982048-70ad075c-83c5-4f43-b975-03da9743329d.mov

### DIFF
--- a/src/components/NavBar/SearchBar.tsx
+++ b/src/components/NavBar/SearchBar.tsx
@@ -1,5 +1,5 @@
 // eslint-disable-next-line no-restricted-imports
-import { t } from '@lingui/macro'
+import { t, Trans } from '@lingui/macro'
 import { sendAnalyticsEvent, Trace, TraceEvent, useTrace } from '@uniswap/analytics'
 import { BrowserEvent, ElementName, EventName, SectionName } from '@uniswap/analytics-events'
 import clsx from 'clsx'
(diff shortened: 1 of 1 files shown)


src/components/NavBar/SearchBar.tsx
```diff
@@ -156,9 +156,10 @@ export const SearchBar = () => {
       >
         <Row
           className={clsx(
-            ` ${styles.nftSearchBar} ${!isOpen && !isMobile && magicalGradientOnHover} ${
-              isMobileOrTablet && (isOpen ? styles.visible : styles.hidden)
-            } `
+            ` ${styles.nftSearchBar} 
+              ${!isOpen && !isMobile && magicalGradientOnHover} 
+              ${isMobileOrTablet && (isOpen ? styles.visible : styles.hidden)} 
+              `
```
just-toby, line 162: eslint was yelling at me ...

yyip-dev (approved):
Lgtm, also appreciate the detailed context + video

src/components/NavBar/SearchBar.tsx
vm: if you use clsx you can just pass in arguments for each of these. don't need this string interpolation.

PR #5712, opened 2022-12-16
fix: adding fixed header

aballerr:
https://uniswaplabs.atlassian.net/browse/WEB-1893

Made header on list sticky

Screen Shot 2022-12-16 at 11 30 12 AM

### DIFF
--- a/src/nft/components/profile/list/NFTListingsGrid.tsx
+++ b/src/nft/components/profile/list/NFTListingsGrid.tsx
@@ -43,7 +43,7 @@ export const NFTListingsGrid = ({ selectedMarkets }: { selectedMarkets: ListingM
   )
 
   return (
-    <Column marginTop="20">
+    <Column>
       <Row position="sticky" top="72" paddingY="24" zIndex="1" backgroundColor="backgroundBackdrop">
         <Column
           marginLeft={selectedMarkets.length > 1 ? '36' : '0'}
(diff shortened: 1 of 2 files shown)


src/nft/components/profile/list/NFTListingsGrid.tsx
```diff
@@ -44,7 +44,7 @@ export const NFTListingsGrid = ({ selectedMarkets }: { selectedMarkets: ListingM
 
   return (
     <Column>
-      <Row marginTop="20">
+      <Row position="sticky" top="72" paddingY="24" zIndex="1" backgroundColor="backgroundBackdrop">
```
grabbou, line 47: I know this is still a draft, but shouldn't we migrate to Styled Components as we tweak different components?

aballerr: if I was adding a new component i would, but realistically this entire file should be rewritten.  Rewriting one line seems little value add here

cbachmeier: I agree with Mike here since we're already here and this will save time later on

cbachmeier: nit: Can you use one of the shared style zIndices?

cbachmeier (approved):
LGTM minor nits

PR #5714, opened 2022-12-16
fix: correct font size for the trade rate to on safari

just-toby:
https://uniswaplabs.atlassian.net/browse/WEB-1505?atlOrigin=eyJpIjoiNjMxY2Q2YmVlYzgyNDMyZjk5Y2JlNjJiYzQxNzY3MjQiLCJwIjoiaiJ9

Changes the font size for this text to 14px, which matches the gas price on the right.

I also noticed that the `showInverted` state doesn't need to be held at the top level of `Swap` and get passed down via props. It can just live in the component that is using it, `TradePrice`.

I can split the changes into 2 PRs if needed.

### DIFF
--- a/src/components/swap/TradePrice.tsx
+++ b/src/components/swap/TradePrice.tsx
@@ -1,7 +1,7 @@
 import { Trans } from '@lingui/macro'
 import { Currency, Price } from '@uniswap/sdk-core'
 import useStablecoinPrice from 'hooks/useStablecoinPrice'
-import { useCallback } from 'react'
+import { useCallback, useState } from 'react'
 import { Text } from 'rebass'
 import styled, { useTheme } from 'styled-components/macro'
 import { ThemedText } from 'theme'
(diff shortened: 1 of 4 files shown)


src/components/swap/TradePrice.tsx
```diff
@@ -58,7 +58,7 @@ export default function TradePrice({ price, showInverted, setShowInverted }: Tra
       }}
       title={text}
     >
-      <Text fontWeight={500} color={theme.textPrimary}>
+      <Text fontWeight={500} fontSize={14} color={theme.textPrimary}>
```
lynnshaoyu, line 61: I'm wondering if ThemedText.BodySmall can be used here, although ThemedText.BodySmall does have a fontWeight of 400... is there a jira or link to the original figma? if 500 is needed this looks fine to me :P

lynnshaoyu: Oooh i just checked the ticket and checked on my local version of the app, the text that Cal wants you to match does have fontWeight 400 so I think this would be a good place to use ThemedText.BodySmall!

lynnshaoyu (approved):
Nice catch on the showInverted props issue! could you fix the ThemedText thing before merging pls ty ty

src/components/swap/TradePrice.tsx
just-toby: thanks!

PR #5715, opened 2022-12-16
fix:  failing cypress test

aballerr:
this test was not updated along with a recent code change.   This will cause a recurring failing cypress test for everyone.  This change fixes that failing test.  Tested and verified locally

Screen Shot 2022-12-16 at 4 29 07 PM


just-toby (approved):

grabbou:
good call, thanks!

PR #5722, opened 2022-12-19
fix: display warning / reload prompt modal when metamask erroneously disconnects

lynnshaoyu:
Screen Shot 2022-12-19 at 10 58 04 AM


src/connection/index.ts
```diff
@@ -25,10 +25,21 @@ export interface Connection {
   type: ConnectionType
 }
 
+let metamaskErrorHandler: (error: Error) => void | undefined
+
+export function setErrorHandler(errorHandler: (error: Error) => void) {
```
zzmp, line 30: This should be named more descriptively: `setMetaMaskErrorHandler`. Also, note that `MetaMask` has two capital `M`s for naming (eg on L28).

src/components/Web3Status/index.tsx
```diff
@@ -206,10 +208,12 @@ function Web3StatusInner() {
   const theme = useTheme()
   const toggleWalletDropdown = useToggleWalletDropdown()
   const toggleWalletModal = useToggleWalletModal()
+  const toggleMetamaskConnectionErrorModal = useToggleMetamaskConnectionErrorModal()
   const walletIsOpen = useModalIsOpen(ApplicationModal.WALLET_DROPDOWN)
   const isClaimAvailable = useIsNftClaimAvailable((state) => state.isClaimAvailable)
 
   const error = useAppSelector((state) => state.connection.errorByConnectionType[getConnection(connector).type])
+  if (getIsMetaMask() && error) toggleMetamaskConnectionErrorModal()
```
zzmp, line 216: This will toggle, so this code may flicker the modal. You should be putting this in a `useEffect`, and only invoking it when there is an error.

src/components/Web3Provider/index.tsx
```diff
@@ -1,15 +1,27 @@
...
+import { updateConnectionError } from 'state/connection/reducer'
+import { useAppDispatch } from 'state/hooks'
 
 export default function Web3Provider({ children }: { children: ReactNode }) {
+  const dispatch = useAppDispatch()
+
+  // Set metamask error handler for metamask disconnection warning modal.
+  useEffect(() => {
+    setErrorHandler((error: Error) =>
+      dispatch(updateConnectionError({ connectionType: ConnectionType.INJECTED, error: error.message }))
```
zzmp, line 21: Once this error is set, will it ever be unset?

zzmp (changes requested):

src/connection/index.ts
lynnshaoyu: sg done

src/components/Web3Provider/index.tsx
lynnshaoyu: it gets unset on line 230 of src/components/WalletModal/index.tsx, but also the user is being prompted to reload the app with the metamask error, so it will get unset after the reload as well.

src/components/Web3Status/index.tsx
lynnshaoyu: not sure what u mean by the toggle may flicker the modal? I wrapped it in a useEffect regardless.

src/components/Web3Status/index.tsx
```diff
@@ -211,10 +214,14 @@ function Web3StatusInner() {
     toggleWalletDropdown()
   }, [toggleWalletDropdown])
   const toggleWalletModal = useToggleWalletModal()
+  const toggleMetamaskConnectionErrorModal = useToggleMetamaskConnectionErrorModal()
   const walletIsOpen = useModalIsOpen(ApplicationModal.WALLET_DROPDOWN)
   const isClaimAvailable = useIsNftClaimAvailable((state) => state.isClaimAvailable)
 
   const error = useAppSelector((state) => state.connection.errorByConnectionType[getConnection(connector).type])
+  useEffect(() => {
+    if (getIsMetaMask() && error) toggleMetamaskConnectionErrorModal()
```
zzmp, line 223: use brackets, as its not a guard statement:

    if (...) {
      ...
    }

zzmp (approved):

src/components/Web3Status/index.tsx
lynnshaoyu: sg done

PR #5724, opened 2022-12-19
chore: add landing page redirect flag

vm:
Will be used for redirecting the user when their wallet is already connected.

### DIFF
--- a/src/featureFlags/flags/landingRedirect.ts
+++ b/src/featureFlags/flags/landingRedirect.ts
@@ -1,24 +1,7 @@
-import { UNIVERSAL_ROUTER_ADDRESS } from '@uniswap/universal-router-sdk'
-import { useWeb3React } from '@web3-react/core'
-
 import { BaseVariant, FeatureFlag, useBaseFlag } from '../index'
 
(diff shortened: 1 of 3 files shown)


src/featureFlags/flags/landingRedirect.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export function useLandingRedirectEnabled(): boolean {
+  const flagEnabled = useLandingRedirectFlag() === BaseVariant.Enabled
+  const { chainId } = useWeb3React()
+  try {
+    // Detect if the Universal Router is not yet deployed to chainId.
+    // This is necessary so that we can fallback correctly on chains without a Universal Router deployment.
+    // It will be removed once Universal Router is deployed on all supported chains.
+    chainId && UNIVERSAL_ROUTER_ADDRESS(chainId)
```
just-toby, line 17: is this important for the landing redirect?

just-toby (approved):
approve to unblock

src/featureFlags/flags/landingRedirect.ts
vm: i copied and pasted, just removed this

PR #5727, opened 2022-12-19
chore: Migrate The Graph queries

cbachmeier:
Migrate Pooled data queries from Relay to Apollo. Similar to how we previously had two different RelayEnvironments, I set up two Apollo clients. I also setup some middleware to dynamically update the uri for the requests. This was tested locally by creating pool positions and switching chains.


src/graphql/thegraph/AllV3TicksQuery.ts
```diff
@@ -21,33 +16,30 @@ const query = graphql`
...
+    data,
+    loading: isLoading,
+    error,
+  } = useQuery(query, {
+    variables: {
+      poolAddress: poolAddress?.toLowerCase(),
+      skip,
+    },
+    pollInterval: interval,
+    client: apolloClient,
```
justindomingue, line 33: longer term, is there a way to share the client?

cbachmeier: I think there is a way but it seems like the long term solution will be using our own endpoint which handles different chains. So I don't want to over invest in a more elegant solution if it would be thrown away https://uniswapteam.slack.com/archives/C02T729PMQE/p1671469080701009

src/graphql/thegraph/apollo.ts
```diff
@@ -0,0 +1,40 @@
...
+// For more information: https://www.apollographql.com/docs/react/networking/advanced-http-networking/
+const authMiddleware = new ApolloLink((operation, forward) => {
+  // add the authorization to the headers
+  const chainId = (store.getState() as AppState).application.chainId
+
+  operation.setContext(() => ({
+    uri:
+      chainId && CHAIN_SUBGRAPH_URL[chainId]
+        ? CHAIN_SUBGRAPH_URL[chainId]
+        : CHAIN_SUBGRAPH_URL[SupportedChainId.MAINNET],
```
justindomingue, line 31: would this behave incorrectly if subgraph isn't defined for chain? should we throw?

alternatively you may be able to narrow the type of `CHAIN_SUBGRAPH_URL` with ts, `as const`

```
const CHAIN_SUBGRAPH_URL = {
  [SupportedChainId.MAINNET]: 'https://api.thegraph.com/subgraphs/name/uniswap/uniswap-v3',
  [SupportedChainId.RINKEBY]: 'https://api.thegraph.com/subgraphs/name/uniswap/uniswap-v3',
...
} as const
```

i *think* `CHAIN_SUBGRAPH_URL[chainId]` will never be undefined then

src/graphql/thegraph/FeeTierDistributionQuery.ts
```diff
@@ -42,28 +37,27 @@ export default function useFeeTierDistributionQuery(
...
-      })
-    }
-  }, [token0, token1, chainId])
-
-  // Trigger fetch on first load
-  useEffect(refreshData, [refreshData, token0, token1])
-
-  useInterval(refreshData, interval, true)
-  return { error, isLoading, data }
+  return useMemo(
```
justindomingue, line 55: nit. don't think you need to memo this one since `data`, `loading` and `error` should be stable from apollo

justindomingue (approved):

src/graphql/thegraph/apollo.ts
cbachmeier: This logic always falls back to Mainnet's chain. I also don't want to invest into changes to this logic if the long term plan is to rely on are in-house graphql data endpoint

src/graphql/thegraph/FeeTierDistributionQuery.ts
cbachmeier: It has come about to make the standard practice for all our graphql queries to memoize the return data so while it might be stable I do like keeping with the standard personally

package.json
```diff
@@ -10,9 +10,11 @@
     "contracts:compile": "yarn contracts:compile:abi && yarn contracts:compile:v3",
     "relay": "relay-compiler relay.config.js",
     "relay-thegraph": "relay-compiler relay_thegraph.config.js",
+    "apollo": "graphql-codegen --config apollo-codegen.ts",
```
zzmp, line 13: These should be named `apollo:generate:data` and `apollo:generate:thegraph`. I would also put them directly above `apollo:generate`.

zzmp (changes requested):

.eslintignore
```diff
@@ -1,3 +1,4 @@
 *.config.ts
 *.d.ts
 /src/graphql/data/__generated__/types-and-hooks.ts
```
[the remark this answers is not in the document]
cbachmeier: `uniswap`? Since it uses the Uniswap endpoint? Also I would prefer to refactor this after migrating to best keep changes localized to Relay -> Apollo

zzmp (approved):

PR #5729, opened 2022-12-19
fix: wait to render Landing content until checking wallet cache

vm:
This prevents the flash of content when the landing page loads before there's a redirect to /swap.


just-toby (approved):

amrsag:
Stop to send fucken messsage

PR #5734, opened 2022-12-19
feat: move about content into landing page

just-toby:
move the About content into the Landing page.

1. remove the `/about` route and the about page - we'll reuse some about components though, so I moved this directory from `/pages` to `/components`
2. remove the `Polling` component from the Landing page, so it doesn't appear when the user scrolls
3. implement the new design [here](https://www.figma.com/file/TBW3lYklHnv7f7Kg9ajlHb/.org-v2?node-id=2466%3A50972&t=S5aEVSKdZOKqSpWW-0), including making the landing page scrollable and adding the new tiles below the title.

https://uniswaplabs.atlassian.net/browse/WEB-2652?atlOrigin=eyJpIjoiMzU3MTYyNjZmZDQ3NDcxNGEyMGU1NzRkMWFkZjkyODkiLCJwIjoiaiJ9


src/components/About/ProtocolBanner.tsx
```diff
@@ -0,0 +1,106 @@
...
+  line-height: 36px;
+
+  @media screen and (min-width: ${BREAKPOINTS.xl}px) {
+    font-size: 28px;
+    line-height: 36px;
+  }
+`
+
+const DescriptionText = styled.div`
+  margin: 10px 10px 0 0;
```
vm, line 55: is the 10px from the design?

just-toby: the top margin is. the right margin is mine, trying to keep the text from getting too close to the button near a certain breakpoint

src/pages/Landing/index.tsx
```diff
@@ -13,30 +18,57 @@ import styled from 'styled-components/macro'
...
   bottom: 0;
   width: 100%;
+  min-height: 550px;
   background: ${({ isDarkMode }) =>
     isDarkMode
       ? 'linear-gradient(rgba(8, 10, 24, 0) 0%, rgb(8 10 24 / 100%) 45%)'
       : 'linear-gradient(rgba(255, 255, 255, 0) 0%, rgb(255 255 255 /100%) 45%)'};
   z-index: ${Z_INDEX.dropdown};
   pointer-events: none;
+  height: ${({ theme }) => `calc(100vh - ${theme.mobileBottomBarHeight}px)`};
```
vm, line 53: try flex: 1?

src/pages/Landing/index.tsx
```diff
@@ -13,30 +18,57 @@ import styled from 'styled-components/macro'
...
+  @media screen and (min-width: ${({ theme }) => theme.breakpoint.md}px) {
```
vm, line 54: same here prob dont need

src/pages/Landing/index.tsx
```diff
@@ -48,34 +80,30 @@ const Glow = styled.div`
...
+  padding: 0 0 40px;
   max-width: min(720px, 90%);
-  position: sticky;
-  bottom: 0;
+  min-height: 500px;
   z-index: ${Z_INDEX.dropdown};
-  padding: 32px 0 80px;
   transition: ${({ theme }) => `${theme.transition.duration.medium} ${theme.transition.timing.ease} opacity`};
-
-  * {
```
vm, line 65: i think you need this

src/pages/Landing/index.tsx
```diff
@@ -48,34 +80,30 @@ const Glow = styled.div`
...
-    pointer-events: auto;
-  }
-
-  @media screen and (min-width: ${BREAKPOINTS.md}px) {
-    padding: 64px 0;
-  }
+  height: ${({ theme }) => `calc(100vh - ${theme.navHeight + theme.mobileBottomBarHeight}px)`};
+  pointer-events: none;
```
vm, line 99: probably don't need this?

src/pages/Landing/index.tsx
just-toby: i didn't understand what this was doing until i read your comment. agreed that this is a better way than putting the same attribute on each of the children.

src/pages/Landing/index.tsx
just-toby: this one i do need, because ContentContainer is now layered over the LandingSwap component, so we need to allow hover events to pass through.

src/pages/Landing/index.tsx
just-toby: this doesn't work because we want the gradient to be the height of the window, not the parent. the parent is taller now since it has content below the fold and is scrollable.

src/pages/Landing/index.tsx
just-toby: see other comment

vm (approved):

PR #5735, opened 2022-12-19
fix: Remove top level Suspense

cbachmeier:
Before v0's style update we had 1 top level Suspense state with a blue loading circle. This is outdated as every page has their own loading skeleton and there is no case where we'd want to show a blank page with just the spinner. The need for this is apparent during the apollo migration where page level suspense states can be removed but this circle would still flash on initial pageload. Tested loading every page and all skeletons show as expected.

Tip: hide whitespace changes when reviewing

### DIFF
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -200,111 +200,109 @@ export default function App() {
           <Popups />
           <Polling />
           <TopLevelModals />
-          <Suspense fallback={<Loader />}>
-            {isLoaded ? (
-              <Routes>
-                <Route path="/" element={<Landing />} />
+          {isLoaded ? (
+            <Routes>
+              <Route path="/" element={<Landing />} />
 
-                <Route path="tokens" element={<Tokens />}>
-                  <Route path=":chainName" />
-                </Route>
-                <Route path="tokens/:chainName/:tokenAddress" element={<TokenDetails />} />
-                <Route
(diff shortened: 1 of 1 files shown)


vm (dismissed):

zzmp (changes requested):
This is going to cause a regression unless all paths have a suspense fallback specified.

Currently, some are set to `null` and others are implicitly using this global fallback (see the lazy-loaded components specified in L50). If this is causing a flash before rendering, considering bundling a parent component in without lazy-loading it (as this is what you really want to do) as an alternative to removing this fallback, which wouldn't fix the underlying problem: that the component is lazy-loaded and there is a flash of rendered state before it loads.

cbachmeier:
No page level paths currently use this App level loading state and it currently only causes UX issues when it appears. ie when it flashes for pages that have non-suspense loading states. This can be shown by removing the `null` Suspense state around the NFT Explore page.

zzmp (changes requested):
I still think this has some issues:
- If I navigate to TokenDetails (Tokens -> select ETH), and it does not load immediately, I get a blank page until it loads. I think there should be a loading indicator until the skeletons have loaded. This holds true for any lazy-loaded path, and will occur with slower internet connections.
- If I navigate to TokenDetails (Tokens -> select ETH) using _this PR_, I get an error.

PR #5737, opened 2022-12-20
test: add tests for landing page

vm:


cypress/support/e2e.ts
```diff
@@ -20,6 +20,7 @@ declare global {
     interface VisitOptions {
       serviceWorker?: true
       featureFlags?: Array<FeatureFlag>
+      selectedWallet?: string | null
```
just-toby, line 23: if you want this to default to `INJECTED`, maybe we should add a second VisitOption
```suggestion
      selectedWallet?: string | null // INJECTED when null
      noSelectedWallet: boolean   // no wallet when true
```
or i guess we could rely on the fact that the option is optional, but it may be harder to understand
```suggestion
      selectedWallet?: string | null // INJECTED when omitted, no wallet when null
```

PR #5738, opened 2022-12-20
chore: remove unused assets

vm:
I wrote a very rudimentary shell script to figure out which files were unused 😂


zzmp (changes requested):
You should be including the shell script, which should also be purging these files. Could you make a new `scripts/` folder? If you do it this way you can also run it as an action to verify that there are no unused files added in the future.

vm:
the shell script is legitimately terrible tho

zzmp (approved):

PR #5748, opened 2022-12-20
fix: add two more analytics events for m parity

JFrankfurt:


src/components/FiatOnrampModal/index.tsx
```diff
@@ -10,7 +11,7 @@ import Circle from '../../assets/images/blue-loader.svg'
 import Modal from '../Modal'
 
 const Wrapper = styled.div`
-  background-color: ${({ theme }) => theme.backgroundSurface};
+  background-color: ${({ theme }) => theme.white};
```
JFrankfurt, line 14: super tiny design nit unrelated to analytics (obv)

JFrankfurt: moonpay iframe content is always white, so in dark mode the wrapper was creating a thick `backgroundSurface` border around the iframe content

sijiawen (approved):

src/components/FiatOnrampAnnouncement/index.tsx
```diff
@@ -113,6 +114,7 @@ export function FiatOnrampAnnouncement() {
 
   const toggleWalletDropdown = useToggleWalletDropdown()
   const handleClick = useCallback(() => {
+    sendAnalyticsEvent('FOR Banner Click')
```
zzmp, line 117: These analytics events should be encoded as an enum through https://github.com/Uniswap/analytics-events I think

zzmp (changes requested):

src/components/FiatOnrampAnnouncement/index.tsx
JFrankfurt: Yeah, now that the feature is launched I'm planning on doing that. (We couldn't before because it was a public repo.) I'd like to get this in as is though, and do all the onramp events as one PR into the analytics repo.

src/components/FiatOnrampModal/index.tsx
```diff
@@ -74,6 +75,10 @@ export default function FiatOnrampModal() {
   const [error, setError] = useState<string | null>(null)
   const [loading, setLoading] = useState(false)
 
+  useEffect(() => {
+    if (fiatOnrampModalOpen) sendAnalyticsEvent('Fiat OnRamp Widget Opened')
```
zzmp, line 79: This should be triggered when it is opened, not inserted from the opened state.

zzmp (changes requested):

zzmp (approved):

PR #5751, opened 2022-12-20
test: disable fiat announcement on cypress tests

vm:
These skipped tests don't need to test the behavior of the announcement.


just-toby (approved):
lgtm

PR #5756, opened 2022-12-21
fix: include descriptions and links for native tokens

just-toby:
Shows the `AboutSection` for native tokens. To get the correct data from the TokenQuery for the native token, we need to use a different contract address than `nativeOnChain(pageChainId).wrapped.address` in some cases. This is because ETH is in the database with contractAddress="ETH". So, for some networks (celo, polygon) we can continue to query the wrapped native token but for ETH metadata we need to use this DB-specific key.

https://uniswaplabs.atlassian.net/browse/WEB-1982?atlOrigin=eyJpIjoiYTU0NjgwMTI4OGQyNGI5Njg5YzU4MWYxMmIyNGU3OGUiLCJwIjoiaiJ9


yyip-dev (approved):
lgtm!

PR #5757, opened 2022-12-21
feat: build correct footer into About page

just-toby:
build the new footer into the Landing/About page.

https://www.figma.com/file/TBW3lYklHnv7f7Kg9ajlHb/.org-v2?node-id=2166%3A14402&t=r0v3eB6pHbfVbuQh-0

https://uniswaplabs.atlassian.net/browse/WEB-2675?atlOrigin=eyJpIjoiMjhhNDhjYjk5MTM3NDY1M2E5YTZhZjQ3MTFhODFhOTIiLCJwIjoiaiJ9


JFrankfurt:
Can you add a link to the design for this?

just-toby:
> Can you add a link to the design for this? (surprised it's not on the jira ticket)

https://www.figma.com/file/TBW3lYklHnv7f7Kg9ajlHb/.org-v2?node-id=2166%3A14402&t=r0v3eB6pHbfVbuQh-0

JFrankfurt:
image

It looks like the top/bottom margin here should be 96px and 144px respectively

JFrankfurt:
image

I'm a little confused about the `about` link here. There's not an active state, and it links to a different page than the one you land on by default, but it doesn't trigger refresh, nav, or active state once it's clicked on and the page goes to its url. It feels broken, but I get why it does this. Not sure if it should get a different design treatment or behavior. cc @infredible what do you think?

JFrankfurt (approved):

src/components/About/Icons.tsx
```diff
@@ -0,0 +1,26 @@
+import styled from 'styled-components/macro'
+
+import { ReactComponent as DiscordI } from './images/discord.svg'
+import { ReactComponent as GithubI } from './images/github.svg'
+import { ReactComponent as TwitterI } from './images/twitter-safe.svg'
```
JFrankfurt, line 5: double check we don't already have these icons

JFrankfurt (dismissed):
image
The unicorn and footer nav shouldn't be here on mobile

just-toby:
gonna try to start a discussion ƒor the "about" link in the figma

JFrankfurt (approved):

PR #5759, opened 2022-12-21
fix: refetch and caching policy for search bar (prices are too stale)

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-2494?atlOrigin=eyJpIjoiMTE2N2Q0ZDcyMzZkNDk2NWFhMTRkOGM2NjkzM2NlN2UiLCJwIjoiaiJ9

### DIFF
--- a/src/components/NavBar/SearchBarDropdown.tsx
+++ b/src/components/NavBar/SearchBarDropdown.tsx
@@ -112,26 +112,8 @@ export const SearchBarDropdown = ({
   isLoading,
 }: SearchBarDropdownProps) => {
   const [hoveredIndex, setHoveredIndex] = useState<number | undefined>(0)
-  const { history } = useSearchHistory()
-  const shortenedHistory = history.slice(0,2)
-  const historyLatestPricesQueries = useQueries(
-    shortenedHistory.map((tokenOrCollection) => ({
-        queryKey: ['address', tokenOrCollection.address],
-        queryFn: () =>
-        tokenOrCollection.type === SearchHistoryType.GenieCollection
-            ? fetchSearchCollections(tokenOrCollection.address)
-            : fetchSearchTokens(tokenOrCollection.name ?? ''),
-        enabled: tokenOrCollection.type === SearchHistoryType.GenieCollection ? !!tokenOrCollection.address : !!tokenOrCollection.name,
-        refetchOnMount: false,
-        refetchOnReconnect: false,
-        refetchOnWindowFocus: false,
-      }))
-    )
-  const shortenedHistoryLatestPricesIsLoading = historyLatestPricesQueries.some(
-    (historyLatestPricesQuery) => historyLatestPricesQuery.isLoading
-  )
-  const shortenedHistoryLatestPrices = !shortenedHistoryLatestPricesIsLoading ? 
-  historyLatestPricesQueries.flatMap((elem) => (elem.data !== undefined ? [elem.data[0]] : [])) : []
+  const { history: searchHistory } = useSearchHistory()
+  const shortenedHistory = useMemo(() => searchHistory.slice(0, 2), [searchHistory])
   
   const { pathname } = useLocation()
   const isNFTPage = useIsNftPage()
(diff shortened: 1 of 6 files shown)


src/nft/components/explore/Banner.tsx
```diff
@@ -115,21 +115,13 @@ const TRENDING_COLLECTION_SIZE = 5
...
-    ['trendingCollections'],
-    () => {
-      return fetchTrendingCollections({
-        volumeType: 'eth',
-        timePeriod: TimePeriod.OneDay,
-        size: TRENDING_COLLECTION_SIZE + EXCLUDED_COLLECTIONS.length,
-      })
-    },
-    {
-      refetchOnReconnect: false,
```
lynnshaoyu, line 128: btw I'm not sure if I should be doing this, but I just noticed nft collection page and banner was also not refetching prices - was that deliberate or nah?

cbachmeier (commented):
Discussed on slack, this does not fix the issue with search history data being stale but removing refetchOnMount flag for fetchTrendingTokens would improve data accuracy for that query

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -111,15 +113,46 @@ export const SearchBarDropdown = ({
...
-  const shortenedHistory = useMemo(() => searchHistory.slice(0, 2), [searchHistory])
+  const shortenedHistoryStalePrices = useMemo(() => searchHistory.slice(0, 2), [searchHistory])
+
+  // TODO(lynnshaoyu): possible to fetchSearchTokens by address instead of just by name?
+  const shortenedHistoryUpdatedPrices = useQueries(
+    shortenedHistoryStalePrices.map((tokenOrCollection) => ({
+      queryKey: ['tokenOrCollectionAddress', tokenOrCollection.address],
+      queryFn: () =>
+        isCollection(tokenOrCollection)
+          ? fetchSearchCollections(tokenOrCollection.address)
```
zzmp, line 124: These are not actually searches, as you know the specific address. You should ask the BE if there is a more appropriate endpoint, or if it is correct usage to re-use the existing endpoints with a fully specified address.

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -111,15 +113,46 @@ export const SearchBarDropdown = ({
 }: SearchBarDropdownProps) => {
   const [hoveredIndex, setHoveredIndex] = useState<number | undefined>(0)
   const { history: searchHistory, updateItem: updateSearchHistory } = useSearchHistory()
-  const shortenedHistory = useMemo(() => searchHistory.slice(0, 2), [searchHistory])
+  const shortenedHistoryStalePrices = useMemo(() => searchHistory.slice(0, 2), [searchHistory])
+
+  // TODO(lynnshaoyu): possible to fetchSearchTokens by address instead of just by name?
+  const shortenedHistoryUpdatedPrices = useQueries(
+    shortenedHistoryStalePrices.map((tokenOrCollection) => ({
+      queryKey: ['tokenOrCollectionAddress', tokenOrCollection.address],
```
zzmp, line 121: this could really just be `['address', tokenOrCollection.address]`

zzmp (changes requested):
By refetching any entry in searchHistory, you essentially deprecate all but the address field of searchHistory. Could you either simplify it, so it only stores the address and type; or expand it, so that the refetch logic is contained within useSearchHistory?

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -110,16 +112,39 @@ export const SearchBarDropdown = ({
...
+  const shortenedHistory = history.slice(0,2)
+  const historyLatestPricesQueries = useQueries(
+    shortenedHistory.map((tokenOrCollection) => ({
+        queryKey: ['address', tokenOrCollection.address],
+        queryFn: () =>
+        tokenOrCollection.type === SearchHistoryType.GenieCollection
+            ? fetchSearchCollections(tokenOrCollection.address)
+            : fetchSearchTokens(tokenOrCollection.name ?? ''),
+        enabled: tokenOrCollection.type === SearchHistoryType.GenieCollection ? !!tokenOrCollection.address : !!tokenOrCollection.name,
+        refetchOnMount: false,
```
cbachmeier, line 125: Do we want to set this flag to false?

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -110,16 +112,39 @@ export const SearchBarDropdown = ({
...
+        refetchOnReconnect: false,
+        refetchOnWindowFocus: false,
+      }))
+    )
+  const shortenedHistoryLatestPricesIsLoading = historyLatestPricesQueries.some(
+    (historyLatestPricesQuery) => historyLatestPricesQuery.isLoading
+  )
+  const shortenedHistoryLatestPrices = !shortenedHistoryLatestPricesIsLoading ? 
+  historyLatestPricesQueries.flatMap((elem) => (elem.data !== undefined ? [elem.data[0]] : [])) : []
```
cbachmeier, line 134: if the call fails do we want to fallback to the past price?

lynnshaoyu:
carter's gql migration makes this unnecessary

PR #5763, opened 2022-12-22
draft: remove hash routing

grabbou:


grabbou:
Closing as this is currently not possible with Create React App. In order to have a proper and first-class routing on IPFS, we need SSG, which can be achieved with e.g. Next.js.

PR #5768, opened 2022-12-23
fix: connect button word break

### LINKED ISSUE #5480: nft pages connect button word break in mobile
**Bug Description**
nft pages connect button word break in mobile

**Steps to Reproduce**
open https://app.uniswap.org/#/nfts

**Expected Behavior**
Text should not break.

**Additional Context**

CombatCode:
Resolves #5480 

Add `word-break: keep-all;` to Connect button styles. This is the only 
East asian languages has specific rules that are followed when breaking text ([e.g. "kinsoku shori" standard for Japanese](https://en.wikipedia.org/wiki/Line_breaking_rules_in_East_Asian_languages)). In opposition to English, words are not separated with with white spaces. 

This PR provide simplest and most efficient way to fix it. Another ways are:
- recognise special characters using JavaScript,
- wrap Japanese words to special CSS classes,
- depend used styles on the currently selected language.

### DIFF
--- a/src/components/Web3Status/index.tsx
+++ b/src/components/Web3Status/index.tsx
@@ -176,6 +176,7 @@ const StyledConnectButton = styled.button`
   font-weight: 600;
   font-size: 16px;
   padding: 10px 8px 10px 12px;
+  word-break: keep-all;
 
   transition: ${({
     theme: {


tinaszheng:
hi @CombatCode - thanks so much for your contribution here! im very sorry we're late to review but we no longer have the same component for connecting wallets so this seems to no longer be an issue :( however i see the same issue with the nav tabs up top so happy to accept similar solution for those links!

PR #5780, opened 2023-01-04
fix: add missing trace events and fix logo

just-toby:
Adds the TraceEvents that I accidentally removed to the Footer Link. I will follow up with logging for other footer links that we might want to add in a separate PR.

also, The footer is using the wrong asset for the logo. this PR fixes that


vm (approved):
Perhaps, worth making a component for the links so less duplication.

PR #5782, opened 2023-01-04
chore: Migrate NFTListingGrid from vanilla-extract to styled-components

cbachmeier:
No functional difference, should only deprecate use of VE in this file and delete unused style.

### DIFF
--- a/src/nft/components/profile/list/NFTListingsGrid.tsx
+++ b/src/nft/components/profile/list/NFTListingsGrid.tsx
@@ -1,8 +1,9 @@
 import { Trans } from '@lingui/macro'
 // eslint-disable-next-line no-restricted-imports
 import { t } from '@lingui/macro'
+import { Box } from 'nft/components/Box'
 import { SortDropdown } from 'nft/components/common/SortDropdown'
-import { Column, Row } from 'nft/components/Flex'
+import { Row } from 'nft/components/Flex'
 import { useSellAsset } from 'nft/hooks'
 import { DropDownOption, ListingMarket } from 'nft/types'
 import { useMemo, useState } from 'react'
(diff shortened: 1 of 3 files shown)


src/nft/components/profile/list/NFTListingsGrid.tsx
```diff
@@ -2,16 +2,19 @@ import { Trans } from '@lingui/macro'
...
 import { DropDownOption, ListingMarket } from 'nft/types'
 import { useMemo, useState } from 'react'
-import styled from 'styled-components/macro'
+import styled, { css } from 'styled-components/macro'
+import { BREAKPOINTS } from 'theme'
 
-import * as styles from './ListPage.css'
 import { NFTListRow } from './NFTListRow'
 
+const GridWrapper = styled.div`
```
JackShort, line 13: fyi: there is a `Column` component in `components/Column`

src/nft/components/profile/list/NFTListingsGrid.tsx
```diff
@@ -53,59 +118,30 @@ export const NFTListingsGrid = ({ selectedMarkets }: { selectedMarkets: ListingM
...
-
-          <Column
-            className={bodySmall}
-            color="textSecondary"
-            flex="1"
-            display={{ sm: 'none', lg: 'flex' }}
-            textAlign="right"
-          >
+          </DropdownWrapper>
+
```
JackShort, line 137: are these whitespaces necessary?

JackShort (approved):
lgtm

src/nft/components/profile/list/NFTListingsGrid.tsx
cbachmeier: I wanted to separate the functional dropdown vs the static text headers to increase readability

Issue #1985, opened 2021-07-07
send transactions in integration tests


moodysalem:
currently we do not send any transactions in our integration tests

it would be great if we tested common workflows e2e, like doing a swap, adding liquidity, etc.

this would probably require using a local fork of a network + some additional setup to avoid introducing flakiness (e.g. set the balance of some tokens, ether via custom rpc calls before each test)

just-toby:
this is done now

PR #5786, opened 2023-01-04
test: add swap transaction e2e test

alimahdiyar:
Related to https://github.com/Uniswap/interface/issues/1985.

I got inspiration from this file: https://github.com/Uniswap/interface/blob/73e420249785fd2e087f8a20e2b56c8708d17055/cypress/support/ethereum.ts#L19 and wrote a library for mocking Metamask's functionality and contracts:
https://github.com/alimahdiyar/metamocks

You can see how this library works by looking at the tests I wrote for it:
https://github.com/alimahdiyar/metamocks/blob/master/tests/metamocks.test.ts

Then I wrote a [mock contract for SwapRouter02](https://github.com/alimahdiyar/uniswap-interface/blob/swap-e2e/cypress/utils/mock-contracts/SwapRouter02.ts) and [registered it to the instance of my library](https://github.com/alimahdiyar/uniswap-interface/blob/3f4e1d0df9e1b970c5d032412955135bc04b11a2/cypress/e2e/swap.test.ts#L49) and spied on its functions that are called for the swap transaction.

This way, we make sure that the contract is called with true parameters without the need to submit a real transaction on a blockchain.

### DIFF
--- a/cypress/e2e/swap.test.ts
+++ b/cypress/e2e/swap.test.ts
@@ -1,3 +1,11 @@
+import { UniswapInterfaceMulticallMockContract } from 'metamocks'
+
+import { MULTICALL_ADDRESS, SWAP_ROUTER_ADDRESSES } from '../../src/constants/addresses'
+import { SupportedChainId } from '../../src/constants/chains'
+import { SwapRouter02, UniswapInterfaceMulticall } from '../../src/types/v3'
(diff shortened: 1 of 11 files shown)


just-toby:
closing since this hasn't received a review in a significant amount of time and we've since added a full e2e test suite

PR #5788, opened 2023-01-05
fix: theme-aware root element color

just-toby:
changes the default background colors for our root elements in the site.

the bug here is that we were always showing the dark linear-gradient when the page loaded. the theme-aware defaults in the CSS below were never visible. now, when the page loads, we use these theme-aware backgrounds. later, when the app runs, we still set the background of `#background-radial-gradient` as usual. ([see here](https://github.com/Uniswap/interface/blob/1802f50163bf8092dac6916d64b9e08ac2ae0a74/src/theme/components/RadialGradientByChainUpdater.ts#L30)).

now, when the page loads or reloads we will show the background corresponding to the user's preference in their OS/browser. note that this may not correspond to their most recent theme in our app, which we rely on javascript to store and propagate throughout the app at runtime.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -77,7 +77,6 @@
       }
 
       #background-radial-gradient {
-        background: linear-gradient(180deg, #202738 0%, #070816 100%);
         position: fixed;
         top: 0;
         left: 0;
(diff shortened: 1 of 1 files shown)


public/index.html
```diff
@@ -77,7 +77,6 @@
       }
 
       #background-radial-gradient {
```
JFrankfurt, line 79: can you add a comment explaining usage? the id name reads like a misnomer now that there's no background info at all in the css and it's all applied once js runs

JFrankfurt (approved):

PR #5791, opened 2023-01-05
chore: migrating bag footer to styled components

JackShort:
* no functional difference only changes the components to use styled components instead of VE


cbachmeier (approved):
LGTM

PR #5792, opened 2023-01-05
chore: Migrate List PriceInput from vanilla-extract to styled-components

cbachmeier:
- no functional difference only changes the components to use styled components instead of VE
  - the only exception is the location of the top right icon when setting a global price which is now better aligned

### DIFF
--- a/src/nft/components/profile/list/PriceTextInput.tsx
+++ b/src/nft/components/profile/list/PriceTextInput.tsx
@@ -7,7 +7,7 @@ import { body } from 'nft/css/common.css'
 import { useSellAsset } from 'nft/hooks'
 import { ListingWarning, WalletAsset } from 'nft/types'
 import { formatEth } from 'nft/utils/currency'
-import { Dispatch, FormEvent, useEffect, useMemo, useRef, useState } from 'react'
+import { Dispatch, FormEvent, useEffect, useRef, useState } from 'react'
 import styled, { useTheme } from 'styled-components/macro'
 import { BREAKPOINTS } from 'theme'
 import { colors } from 'theme/colors'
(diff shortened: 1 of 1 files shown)


src/nft/components/profile/list/PriceTextInput.tsx
```diff
@@ -66,27 +121,21 @@ export const PriceTextInput = ({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [listPrice])
 
+  const borderColor = useMemo(
```
JackShort, line 124: `useMemo` is not necessary here and probably adds extra overhead

src/nft/components/profile/list/PriceTextInput.tsx
```diff
@@ -1,13 +1,67 @@
...
+  border: 2px solid;
+  border-radius: 8px;
+  border-color: ${({ borderColor }) => borderColor};
+  margin-right: auto;
+  box-sizing: border-box;
+`
+
+const CurrencyWrapper = styled.div<{ listPrice: number | undefined }>`
+  margin-right: 16px;
+  color: ${({ listPrice, theme }) => (listPrice && listPrice >= 0 ? theme.textPrimary : theme.textSecondary)};
```
JackShort, line 33: this will always be `theme.textSecondary` when `listPrice` is `0` because it gets evaluated to `false`- if this is intentional then can you remove the `>=` and make it `>` else check for `undefined` explicitly.

JackShort (approved):
few comments

src/nft/components/profile/list/PriceTextInput.tsx
cbachmeier: I've simplified to just check listPrice

PR #5793, opened 2023-01-05
fix: do not show assets in sudoswap xyk pools

JackShort:
* assets in sudoswap xyk pools were crashing the app this prevents them from being shown

### DIFF
--- a/src/nft/utils/asset.ts
+++ b/src/nft/utils/asset.ts
@@ -47,7 +47,8 @@ export const generateTweetForPurchase = (assets: UpdatedGenieAsset[], txHashUrl:
   return `https://twitter.com/intent/tweet?text=${encodeURIComponent(tweetText)}`
 }
 
-export const isValidAsset = (asset: GenieAsset): boolean => {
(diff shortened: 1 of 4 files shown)


src/nft/utils/asset.ts
```diff
@@ -44,3 +46,10 @@ export const generateTweetForPurchase = (assets: UpdatedGenieAsset[], txHashUrl:
   } with @Uniswap 🦄\n\nhttps://app.uniswap.org/#/nfts/collection/0x60bb1e2aa1c9acafb4d34f71585d7e959f387769\n${txHashUrl}`
   return `https://twitter.com/intent/tweet?text=${encodeURIComponent(tweetText)}`
 }
+
+export const isValidAsset = (asset: GenieAsset): boolean => {
```
cbachmeier, line 50: can we rename to isNotXYKPool just to keep specification very atomic?

cbachmeier: Also add the comment `// TODO: remove when BE supports sudo XYK pool"

cbachmeier (approved):
LGTM

PR #5794, opened 2023-01-06
chore: update Sentry version

grabbou:
Fixes WEB-2550

TL;DR we pinned it down due to a bug, it's fixed now, so we're good to go.


cbachmeier (approved):
LGTM

PR #5802, opened 2023-01-09
feat: migrate search tokens to gql

cartcrom:
Combination of #5805 #5894 and #5896

Replaces usage of the temp api token search url, replaced by GQL queries for search tokens, trending tokens, and recently searched tokens. Removes the Fungible Token type, replaced by apollo-generated token types.

These PR's fix:
Overall result quality and quantity
Cross-chain token result mapping
Price discrepancy between search and Token Details
Price discrepancy between search and recently searched
Floor price discrepancy between recently searched NFT Collections and NFT pages


src/graphql/data/SearchTokens.ts
```diff
@@ -0,0 +1,97 @@
+import { WRAPPED_NATIVE_CURRENCY } from 'constants/tokens'
+import gql from 'graphql-tag'
+import { useMemo } from 'react'
+
+import { Chain, SearchTokensQuery, useSearchTokensQuery } from './__generated__/types-and-hooks'
+import { chainIdToBackendName } from './util'
+
+gql`
+  query SearchTokens($searchQuery: String!) {
+    searchTokens(searchQuery: $searchQuery) {
```
cbachmeier, line 10: Is there a way we can modify the ordering? I believe the old api was based off volume but for example searching `eth` steth didn't show up for me which I would have expected.
new results:
Screen Shot 2023-01-10 at 09 08 29
old resultss:
Screen Shot 2023-01-10 at 09 08 36

src/graphql/data/SearchTokens.ts
```diff
@@ -0,0 +1,97 @@
...
+export type SearchToken = NonNullable<NonNullable<SearchTokensQuery['searchTokens']>[number]>
+
+function isMoreRevelantToken(current: SearchToken, existing: SearchToken | undefined, searchChain: Chain) {
+  if (!existing) return true
+
+  // Always priotize natives, and if both tokens are native, prefer native on current chain (i.e. Matic on Polygon over Matic on Mainnet )
+  if (current.standard === 'NATIVE' && (existing.standard !== 'NATIVE' || current.chain === searchChain)) return true
+
+  // Prefer tokens on the searched chain, otherwise prefer mainnet tokens
+  return current.chain === searchChain || (existing.chain !== searchChain && current.chain === Chain.Ethereum)
```
cbachmeier, line 48: this logic seems to be a little faulty. Searching for `eth` on arbitrum, I only get native wrapped in the current chain. Compared to current behaviour where most tokens are on the current chain
new:
Screen Shot 2023-01-10 at 09 11 47
old:
Screen Shot 2023-01-10 at 09 11 53

src/graphql/data/SearchTokens.ts
cbachmeier: This might actually be a data issue, for some reason stEth is only returning as being on Polygon

src/graphql/data/SearchTokens.ts
```diff
@@ -0,0 +1,97 @@
...
+    // Stores results, allowing overwriting cross-chain tokens w/ more 'relevant token'
+    const selectionMap: { [projectId: string]: SearchToken } = {}
+    data?.searchTokens?.forEach((token) => {
+      if (token.project?.id) {
+        const existing = selectionMap[token.project.id]
+        if (isMoreRevelantToken(token, existing, searchChain)) selectionMap[token.project.id] = token
+      }
+    })
+    return Object.values(selectionMap).sort(
+      searchTokenSortFunction.bind(null, searchChain, WRAPPED_NATIVE_CURRENCY[chainId]?.address)
```
cbachmeier, line 88: is binding with this additional null input the only way to pass in the additional data used in the compare function?

src/nft/types/navbar/navbar.ts
```diff
@@ -1,28 +1,48 @@
...
-  extensions?: FungibleTokenExtensions
-  marketCap: number
+  logoURI?: string | null
+  priceUsd?: number | null
+  price24hChange?: number | null
+  volume24h?: number | null
+  isVerified?: boolean
+}
+
+export function parseFungibleTokens(tokens: SearchToken[]) {
```
cbachmeier, line 25: this function should probably be part of `SearchToken.ts` and the useSearchTokens returns FungibleToken[].

src/nft/types/navbar/navbar.ts
```diff
@@ -1,28 +1,48 @@
...
 }
-interface BridgeInfoEntry {
-  tokenAddress?: string
-}
-
-interface FungibleTokenExtensions {
-  bridgeInfo?: { [chain: number]: BridgeInfoEntry }
-}
 
 export interface FungibleToken {
```
cbachmeier, line 12: This type is used outside of nfts it should probably be moved outside of the nft folder. Or possibly deprecate this type and use the generated type with the additional of the isVerified logic

src/graphql/data/SearchTokens.ts
cartcrom: I actually noticed this same issue! Matteen is looking into it, seems like a data issue. They are sorted by volume

src/graphql/data/SearchTokens.ts
cartcrom: the first argument in bind provides the sorting function with a value for 'this', and then the following arguments are param1, param2, etc

src/graphql/data/SearchTokens.ts
cartcrom: Thanks for catching, looks like its a data issue. Searching "wrapped" does return arbitrum weth... pinging the data team

src/nft/types/navbar/navbar.ts
cartcrom: Yes would love to get rid of this type, but I need to migrate trending tokens and recently searched tokens first too (those PR's are next).

src/nft/types/navbar/navbar.ts
cartcrom: Both this function and the FungibleToken type will be deleted soon, see comment below

src/graphql/data/SearchTokens.ts
cartcrom: this is fixed now

src/graphql/data/SearchTokens.ts
cartcrom: It seems like this is somewhat of a one-off. Arb WETH's on-chain name is "Wrapped Ethereum", while the others are "Wrapped Ether", so it has a weaker match. I don't think this is hugely important, because if a user is looking for WETH they're probably more likely to type "weth" or "wrapped"

src/nft/types/navbar/navbar.ts
cartcrom: FungibleToken is removed in #5896

src/graphql/data/SearchTokens.ts
cbachmeier: Ordering still seems funky
Screen Shot 2023-02-05 at 20 31 26

src/graphql/data/SearchTokens.ts
cbachmeier: For Arbitrum searching rEth shows the chain specific version below the mainnet
Screen Shot 2023-02-05 at 20 33 39

src/components/NavBar/SearchBar.tsx
```diff
@@ -64,16 +66,10 @@ export const SearchBar = () => {
...
-    () => fetchSearchTokens(debouncedSearchValue),
-    {
-      refetchOnWindowFocus: false,
-      refetchOnMount: false,
-      refetchOnReconnect: false,
-      enabled: !!debouncedSearchValue.length,
-    }
-  )
+  const { chainId } = useWeb3React()
+  const { data: searchQueryTokens, loading: tokensAreLoading } = useSearchTokens(debouncedSearchValue, chainId ?? 1)
```
cbachmeier, line 70: non-blocking nit but I would recommend setting a const DEFAULT_CHAIN_ID = 1

src/graphql/data/SearchTokens.ts
cbachmeier: Another example is Uni on Optimism
Screen Shot 2023-02-05 at 20 40 58

cbachmeier (approved):
LGTM

PR #5804, opened 2023-01-09
feat: adding price scaling for sudoswap xyk pool

JackShort:
* This scales the prices on the frontend to handle sudoswap xyk pool
* good collection to test: White Hearts 1.0

### DIFF
--- a/src/nft/utils/pooledAssets.ts
+++ b/src/nft/utils/pooledAssets.ts
@@ -33,6 +33,50 @@ const calculateScaledPrice = (currentPrice: BigNumber, poolFee: BigNumber): BigN
   return currentPrice.add(protocolFee).add(tradeFee)
 }
 
+const calcSudoSwapLinearBondingCurve = (currentPrice: BigNumber, delta: BigNumber, position = 0): BigNumber => {
+  for (let i = 0; i <= position; i++) {
+    currentPrice = currentPrice.add(delta)
+  }
+
+  return currentPrice
+}
+
+const calcSudoSwapExponentialBondingCurve = (currentPrice: BigNumber, delta: BigNumber, position = 0): BigNumber => {
+  for (let i = 0; i <= position; i++) {
+    currentPrice = currentPrice.mul(delta).div(BigNumber.from(PRECISION))
+  }
+
+  return currentPrice
+}
+
+const calcSudoSwapXykBondingCurve = (
+  currentPrice: BigNumber,
+  sudoSwapPool: Pool,
+  position = 0
+): BigNumber | undefined => {
+  let virtualTokenBalance = BigNumber.from(sudoSwapPool.spotPrice)
+  let virtualNFTBalance = BigNumber.from(sudoSwapPool.delta)
+
+  if (virtualNFTBalance.sub(BigNumber.from(1)).gt(BigNumber.from(0))) {
+    currentPrice = virtualTokenBalance.div(virtualNFTBalance.sub(BigNumber.from(1)))
+  } else {
+    return undefined
+  }
+
+  for (let i = 1; i <= position; i++) {
+    virtualTokenBalance = virtualTokenBalance.add(currentPrice)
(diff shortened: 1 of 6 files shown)


src/nft/utils/pooledAssets.ts
```diff
@@ -44,12 +45,33 @@ export const calcSudoSwapPrice = (asset: GenieAsset, position = 0): string | und
   let currentPrice = BigNumber.from(sudoSwapPool.spotPrice)
   const delta = BigNumber.from(sudoSwapPool.delta)
   const poolFee = BigNumber.from(sudoSwapPool.fee)
+  let virtualTokenBalance = BigNumber.from(sudoSwapPool.spotPrice)
+  let virtualNFTBalance = BigNumber.from(sudoSwapPool.delta)
 
-  for (let i = 0; i <= position; i++) {
+  if (sudoSwapPool.bondingCurve === BondingCurve.Xyk) {
```
cbachmeier, line 51: this function has gotten rather large, do we want to split it into smaller functions for each curve?

cbachmeier (changes requested):
Functionality works well. We should still delete the function `isNotXYKPool` in `nft/utils`. Also gave a organization suggestion. 

One other non blocking idea where some additional logic could be simplified if `calcPoolPrice` checked if iasset was in a sudoswap pool and routed to the appropriate helper function. Basically allowing that function to handle all routing.

src/nft/utils/pooledAssets.ts
```diff
@@ -32,6 +33,50 @@ const calculateScaledPrice = (currentPrice: BigNumber, poolFee: BigNumber): BigN
...
+  sudoSwapPool: Pool,
+  position = 0
+): BigNumber | undefined => {
+  let virtualTokenBalance = BigNumber.from(sudoSwapPool.spotPrice)
+  let virtualNFTBalance = BigNumber.from(sudoSwapPool.delta)
+
+  if (virtualNFTBalance.sub(BigNumber.from(1)).gt(BigNumber.from(0))) {
+    currentPrice = virtualTokenBalance.div(virtualNFTBalance.sub(BigNumber.from(1)))
+  } else {
+    return undefined
```
cbachmeier, line 63: I think this undefined might be causing a crash when trying to add past the available amount. Should we return 0 instead?
xyk

src/nft/utils/pooledAssets.ts
```diff
@@ -120,6 +171,13 @@ export const calcPoolPrice = (asset: GenieAsset, position = 0) => {
   return price.toString()
 }
 
+// TODO: a lot of the below typecasting logic can be simplified when GraphQL migration is complete
```
cbachmeier, line 174: my bad, this comment can be removed

src/nft/utils/pooledAssets.ts
```diff
@@ -45,19 +90,25 @@ export const calcSudoSwapPrice = (asset: GenieAsset, position = 0): string | und
   const delta = BigNumber.from(sudoSwapPool.delta)
   const poolFee = BigNumber.from(sudoSwapPool.fee)
 
-  for (let i = 0; i <= position; i++) {
-    if (sudoSwapPool.bondingCurve === BondingCurve.Linear) {
-      currentPrice = currentPrice.add(delta)
-    } else if (sudoSwapPool.bondingCurve === BondingCurve.Exponential) {
-      currentPrice = currentPrice.mul(delta).div(BigNumber.from(PRECISION))
+  if (sudoSwapPool.bondingCurve === BondingCurve.Linear) {
```
cbachmeier, line 93: v clean

cbachmeier (changes requested):

src/nft/utils/pooledAssets.ts
JackShort: its actually returning `0` after the refactor

cbachmeier (approved):
LGTM

PR #5806, opened 2023-01-09
chore: bump sdk-core

zzmp:
NB: This does not bump the widget, which continues to use an older version.

DO NOT MERGE - this will break client-side SOR.


vm (approved):
What do we need to test manually?

zzmp:
> What do we need to test manually?

Swaps, as the router has been bumped by a major version.

willpote (changes requested):
So V3 of the SOR uses Universal Router + Permit2, which has a different signing/transaction submission flow. 

Merging this as is will break the client side auto router

PR #5815, opened 2023-01-11
fix: collectionNft called before intialized

cbachmeier:
Locally there was a bug if you went directly to profile before opening a collection page as some of the styles/shared components were imported from CollectionNfts before that file was intialized. This does not happen in production. This is fixed by moving the loading styles to the appropriate loading file and the Clear button styles to a shared styles file. No functionality was changed, code was only moved.


JackShort (approved):
ty for fixing this

PR #5820, opened 2023-01-11
chore: add token explore integration test

lynnshaoyu:
following test cases spec'd out here: https://www.notion.so/uniswaplabs/Web-QA-testing-1624a246ddd6430fa69382bf2165b936

https://uniswaplabs.atlassian.net/browse/WEB-2528?atlOrigin=eyJpIjoiODJkMGE1OWNhMzk2NDU3YmJlNTA0M2UyNTExOGI5N2QiLCJwIjoiaiJ9


src/components/Tokens/TokenTable/TimeSelector.tsx
```diff
@@ -111,7 +111,7 @@ export default function TimeSelector() {
 
   return (
     <StyledMenu ref={node}>
-      <FilterOption onClick={toggleMenu} aria-label="timeSelector" active={open}>
+      <FilterOption onClick={toggleMenu} aria-label="timeSelector" active={open} data-testid="time-selector">
```
zzmp, line 114: Use the aria-label instead of a testid?

src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -469,7 +477,7 @@ export const LoadedRow = forwardRef((props: LoadedRowProps, ref: ForwardedRef<HT
 
   // TODO: currency logo sizing mobile (32px) vs. desktop (24px)
   return (
-    <div ref={ref} data-testid={`token-table-row-${tokenName}`}>
+    <div ref={ref} data-testid={`token-table-row-${tokenName?.split(' ').join('-')}`}>
```
zzmp, line 480: instead of all this logic, can you instead use `token-table-row-${tokenSymbol}`.

zzmp: similarly, while you're in this file, `tokenName`, `tokenSymbol`, and `tokenAddress` should all be removed in favor of direct access to `token`.

cypress/e2e/token-explore.test.ts
```diff
@@ -0,0 +1,74 @@
+import { getTestSelector, getTestSelectorStartsWith } from '../utils'
+
+describe('Testing tokens on uniswap page', () => {
```
zzmp, line 3: Name after the file. Don't' use the word "testing", because it's a test file.

eg: `describe('token explore', () => {`

cypress/e2e/token-explore.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    cy.contains('Ethereum is a smart contract platform that enables developers to build tokens').should('exist')
+    cy.contains('Etherscan').should('exist')
+  })
+
+  it('should update when global network changed', () => {
+    cy.visit('/tokens/ethereum')
+    cy.get(getTestSelector('tokens-network-filter-selected')).should('contain', 'Ethereum')
+    cy.get(getTestSelector('token-table-row-Ether')).should('exist')
+
+    // note: cannot switch global chain via UI because we cannot approve the network switch
```
zzmp, line 58: there's no metamask here, I thought. I thought we used a custom provider. What happens if you _try_ to switch via UI? Can you extend the custom provider to support this test case?

lynnshaoyu: the request fails with a network failed toast in the top right, and it says you need to approve in mm wallet

src/components/Tokens/TokenTable/TimeSelector.tsx
lynnshaoyu: https://docs.cypress.io/guides/references/best-practices#Selecting-Elements 
^ suggests using testid to avoid changes (eg: aria label could change)

src/components/Tokens/TokenTable/TokenRow.tsx
lynnshaoyu: makes sense done

JFrankfurt (commented):
These tests seem to be functioning more like snapshots than e2e behavior checks. 

IMO the a better approach here is to use cypress as an interaction flow check.

i.e., 
1. setup initial state
2. take an action `cy.click`
3. asset on expected result state

So instead of testing "<ui element should load>", maybe test "<ui element should behave in x manner>". i.e., "token table should be sortable" or "token table should render results from api"

JFrankfurt (approved):

PR #5821, opened 2023-01-12
fix: update permit2 to match widget implementation

zzmp:
- Updates transactions updater to fast-forward to future receipts' blocks
- Moves usePermit2 to usePermit2Allowance
- Ports useTokenAllowance, usePermitAllowance, and usePermit2Allowance from Uniswap/widgets


src/pages/Swap/index.tsx
```diff
@@ -302,36 +302,31 @@ export default function Swap({ className }: { className?: string }) {
...
     permit2Enabled ? maximumAmountIn : undefined,
     permit2Enabled && chainId ? UNIVERSAL_ROUTER_ADDRESS(chainId) : undefined
   )
-  const isApprovalLoading = permit.state === PermitState.APPROVAL_LOADING
-  const [isPermitPending, setIsPermitPending] = useState(false)
-  const [isPermitFailed, setIsPermitFailed] = useState(false)
-  const addTransaction = useTransactionAdder()
-  const updatePermit = useCallback(async () => {
-    setIsPermitPending(true)
+  const isApprovalLoading = allowance.state === AllowanceState.REQUIRED && allowance.isApprovalLoading
```
vm, line 309: bit confused here. shouldn't one of these checks be enough? what happens if allowance.isApprovalLoading is true but allowance.state !== AllowanceState.REQUIRED, or vice versa?

vm (approved):

src/pages/Swap/index.tsx
zzmp: `Allowance` is typed such that `.isApprovalLoading` only exists if `.state === AllowanceState.REQUIRED`, so the first half of this compound boolean is really just typechecking.

If `.state !== AllowanceState.REQUIRED`, then we don't care if approval is loading, it may as well be `false`.
If `.isApprovalLoading === false` while `.state === AllowanceState.REQUIRED`, then we need to keep it false so that we show the correct (non-disabled) UI.
Only if `isApprovalLoading` do we want to show a disabled UI showing that approval is loading.

PR #5824, opened 2023-01-12
fix: nft add to bag test

JackShort:
* the testing suite is failing because the pudgy penguin that was being tested is no longer for sale


cbachmeier (approved):
good catch

PR #5834, opened 2023-01-13
build: skip Monday releases

zzmp:
Skips Monday releases. This can be reverted _after_ next Monday, which should be skipped due to holiday.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,7 +1,7 @@
 name: Release
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


just-toby (approved):
is this supposed to be permanent or temporary just for this upcoming monday?

zzmp:
> is this supposed to be permanent or temporary just for this upcoming monday?

This meant to be for this upcoming Monday, but we don't have a clean way to specify our holiday calendar.
Instead, I plan to merge this and revert it after Monday.

PR #5836, opened 2023-01-16
feat: detect Brave Wallet 

grabbou:
Fixes WEB-2729

Brave Wallet sets "isMetaMask" to true for compatibility purposes (it uses the same API). This PR updates the detection mechanism to ensure that we show "Connect with MetaMask" only when connecting with MetaMask. 

Breakdown of all possible combinations of Brave & MetaMask together with UI behavior is broken down in the Jira ticket.

### DIFF
--- a/src/connection/utils.ts
+++ b/src/connection/utils.ts
@@ -12,16 +12,16 @@ export function getIsInjected(): boolean {
   return Boolean(window.ethereum)
 }
 
+export function getIsBraveWallet(): boolean {
+  return window.ethereum?.isBraveWallet ?? false
+}
+
 export function getHasMetaMaskExtensionInstalled(): boolean {
-  return window.ethereum?.isMetaMask ?? false
+  return (window.ethereum?.isMetaMask ?? false) && !getIsBraveWallet()
 }
 
 export function getHasCoinbaseExtensionInstalled(): boolean {
   return window.ethereum?.isCoinbaseWallet ?? false
-}
-
-export function getIsMetaMask(connectionType: ConnectionType): boolean {
-  return connectionType === ConnectionType.INJECTED && getHasMetaMaskExtensionInstalled()
 }
 
 const CONNECTIONS = [
--- a/src/connection/utils.ts
+++ b/src/connection/utils.ts
@@ -19,6 +19,7 @@ export function getIsBraveWallet(): boolean {
 export function getIsMetaMaskWallet(): boolean {
   // When using Brave browser, `isMetaMask` is set to true when using the built-in wallet
   // This function should return true only when using the MetaMask extension
+  // https://wallet-docs.brave.com/ethereum/wallet-detection#compatability-with-metamask
   return (window.ethereum?.isMetaMask ?? false) && !getIsBraveWallet()
 }
 
--- a/src/connection/utils.ts
+++ b/src/connection/utils.ts
@@ -16,11 +16,11 @@ export function getIsBraveWallet(): boolean {
   return window.ethereum?.isBraveWallet ?? false
 }
 
-export function getHasMetaMaskExtensionInstalled(): boolean {
+export function getIsMetaMaskWallet(): boolean {
   return (window.ethereum?.isMetaMask ?? false) && !getIsBraveWallet()
 }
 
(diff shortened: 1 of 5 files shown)


src/connection/utils.ts
```diff
@@ -12,18 +12,18 @@ export function getIsInjected(): boolean {
   return Boolean(window.ethereum)
 }
 
+export function getIsBraveWallet(): boolean {
+  return window.ethereum?.isBraveWallet ?? false
+}
+
 export function getHasMetaMaskExtensionInstalled(): boolean {
-  return window.ethereum?.isMetaMask ?? false
+  return (window.ethereum?.isMetaMask ?? false) && !getIsBraveWallet()
```
dannydaniil, line 20: This logic breaks down when a Brave wallet has Metamask installed. In that case, it just picks up the MetaMask wallet 🤔 

We need well-defined requirements from product as to how to handle that case. Most likely, we need to support extra wallet options in our [wallet options](https://github.com/Uniswap/interface/blob/main/src/components/WalletModal/index.tsx) component. It seems to be in this ticket's scope to also handle this case.

dannydaniil (changes requested):

src/connection/utils.ts
grabbou: When Brave wallet has MetaMask installed and enabled (as a preferred extension), `isBraveWallet` will be false. `isBraveWallet` is only true when users explicitly decided to use Brave as their primary extension.

grabbou: Here's a breakdown from my manual testing of this:
- **MetaMask installed, Brave Wallet disabled ([instructions](https://metamask.zendesk.com/hc/en-us/articles/360038596792-Using-MetaMask-wallet-in-Brave-browser))** - connects with MetaMask and displays MetaMask 
- **MetaMask not installed, Brave wallet disabled** - Install MetaMask
- **MetaMask not installed, Brave wallet enabled** - Connect with Browser Wallet
- **MetaMask installed, Brave Wallet enabled** - Depending on user settings - either MetaMask (if MetaMask is selected as “default”, even though Brave Wallet is enabled) or Browser Wallet (in case MetaMask is installed, but not selected as default).

src/react-app-env.d.ts
```diff
@@ -8,8 +8,11 @@ interface Window {
   // walletLinkExtension is injected by the Coinbase Wallet extension
   walletLinkExtension?: any
   ethereum?: {
-    // value that is populated and returns true by the Coinbase Wallet mobile dapp browser
+    // set by the Coinbase Wallet mobile dapp browser
     isCoinbaseWallet?: true
+    // set by the Brave browser when using built-in wallet
+    isBraveWallet?: true
+    // set by the MetaMask browser extension (in Brave browser when using built-in wallet too)
```
dannydaniil, line 15: Wondering if we should change the comment to:
```
// set by the MetaMask browser extension (also set by Brave browser when using built-in wallet)
```
The comment in its current form seems somewhat confusing

src/connection/utils.ts
dannydaniil, line 20: We could link the documentation here, as it's hard to figure out what's going on. Without it, it feels like we are prioritizing the Brave wallet inside the Brave browser.

Furthermore, `getHasMetaMaskExtensionInstalled`, does not seem to be semantically correct anymore. The user could have MetaMask installed as an extension, but prioritized the Brave browser in their settings. I was wondering if `isMetaMaskWallet` or something along those lines makes more sense now 🤔

dannydaniil: You're right, all those cases are correctly covered! It just seemed super confusing for me from a user's standpoint. Feels like we should be able to support both, but it does not seem so from Brave's documentation 🤷

dannydaniil:
As a follow up, these is great logic to abstract inside web3-react 🙏

src/connection/utils.ts
grabbou: You are right, that name could be somewhat less confusing!

dannydaniil (approved):
LGTM! Thanks for addressing my comments!

grabbou:
My pleasure @dannydaniil, thanks for spending time on the review!

PR #5839, opened 2023-01-17
chore: universal search e2e test

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-2526?atlOrigin=eyJpIjoiNjk3ZmQ0M2RjYWNmNDUyMThhODIwN2IzNWQxNjU2ODMiLCJwIjoiaiJ9


cypress/e2e/universal-search.test.ts
```diff
@@ -0,0 +1,75 @@
...
+      .and('contain.text', 'No NFT collections found.')
+  })
+
+  it('should yield no results found when contract address is search term', () => {
+    // Search for uni token contract address.
+    cy.get('[data-cy="search-bar-input"]').last().type('0x1f9840a85d5aF5bf1D1762F925BDADdC4201F984')
+    cy.get('[data-cy="search-bar"]')
+      .should('contain.text', 'No tokens found.')
+      .and('contain.text', 'No NFT collections found.')
+  })
```
cbachmeier, line 27: looks like a duplication of the block above

cypress/e2e/universal-search.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    cy.get('[data-cy="searchbar-token-row-UNI"]').click()
+
+    cy.get('div').contains('Uniswap').should('exist')
+    // Stats should have: TVL, 24H Volume, 52W low, 52W high.
+    cy.get(getTestSelector('token-details-stats')).should('exist')
+    cy.get(getTestSelector('token-details-stats')).within(() => {
+      cy.get('[data-cy="tvl"]').should('include.text', '$')
+      cy.get('[data-cy="volume-24h"]').should('include.text', '$')
+      cy.get('[data-cy="52w-low"]').should('include.text', '$')
+      cy.get('[data-cy="52w-high"]').should('include.text', '$')
```
cbachmeier, line 41: nit: these tests seem to be more for the details page than universal search. Do we want to add checks that the suggestion row has the expected data?
Screen Shot 2023-01-19 at 09 49 24

cypress/e2e/universal-search.test.ts
```diff
@@ -0,0 +1,75 @@
...
+      .should('exist')
+
+    // Most popular 3 tokens should be shown.
+    cy.get('[data-cy="searchbar-dropdown"]')
+      .contains('[data-cy="searchbar-dropdown"]', 'Popular tokens')
+      .find('[data-cy^="searchbar-token-row"]')
+      .its('length')
+      .should('be.eq', 3)
+  })
+
```
cbachmeier, line 69: Do we also want to add a test when on a different network. From the [QA Doc](https://www.notion.so/uniswaplabs/Web-QA-testing-1624a246ddd6430fa69382bf2165b936):

> - If global network isn’t mainnet:
     - Recent searches should stay the same, regardless of what network they were on
     - Popular tokens list should be the top 4 tokens of that network

cypress/e2e/universal-search.test.ts
lynnshaoyu: oof thanks for the catch, removed

cypress/e2e/universal-search.test.ts
lynnshaoyu: we can't change the global network currently due to cypress limitations (can't hit approve network change in metamask). frontend infra team will investigate adding e2e testing capabilities w metamask.

cypress/e2e/universal-search.test.ts
lynnshaoyu: good point, added checks for the row of suggested token itself

cbachmeier (approved):
LGTM just seems to need some prettier auto formatting

PR #5847, opened 2023-01-18
fix: fallback to eth_sign for permit approval

zzmp:
Falls back to eth_sign if eth_signTypedData_v4 (ie EIP-712) is not implemented, eg for zerion.
This has been tested with zerion; other known wallets implement EIP-712.

Implementing this as a polyfill overriding JsonRpcSigner.prototype._signTypedData allows the fix to propagate to widgets as well. This is preferable to a utility method (eg `function signTypedData(...)`), as that would require additional changes/releases to widgets.

This will be ported to conedison, as well as proposed as an upstream change to ethers.

WEB-2792

### DIFF
--- a/src/utils/signTypedData.ts
+++ b/src/utils/signTypedData.ts
@@ -12,17 +12,19 @@ JsonRpcSigner.prototype._signTypedData = async function signTypedDataWithFallbac
     return this.provider.resolveName(name) as Promise<string>
   })
 
-  const address = (await this.getAddress()).toLowerCase()
-  const payload = JSON.stringify(_TypedDataEncoder.getPayload(populated.domain, types, populated.value))
+  const address = await this.getAddress()
 
-  return this.provider.send('eth_signTypedData_v4', [address, payload]).catch((e) => {
-    if (!('message' in e && e.message.match(/not found/i))) throw e
-
-    console.warn('eth_signTypedData_v4 failed, falling back to eth_sign:', e)
-    const hash = _TypedDataEncoder.hash(populated.domain, types, populated.value)
-    return this.provider.send('eth_sign', [address, hash]).catch((e) => {
-      console.warn('eth_sign failed:', e)
-      throw e
-    })
-  })
+  try {
+    return await this.provider.send('eth_signTypedData_v4', [
+      address.toLowerCase(),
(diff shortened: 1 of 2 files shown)


tinaszheng (approved):

vm (approved):

src/utils/signTypedData.ts
```diff
@@ -0,0 +1,30 @@
...
+  })
+
+  const address = await this.getAddress()
+
+  try {
+    return await this.provider.send('eth_signTypedData_v4', [
+      address.toLowerCase(),
+      JSON.stringify(_TypedDataEncoder.getPayload(populated.domain, types, populated.value)),
+    ])
+  } catch (error) {
```
tinaszheng, line 22: idk if really a priority, but you could technically try `eth_signTypedData` or eth_signTypedData_v3` before falling back to `eth_sign` as a general case

zzmp: we could, but in the practical case that's been tested (eg zerion) we need to resort to `eth_sign`.

tinaszheng: there may be other wallets that only implement the general `eth_signTypedData` case, and it's always better to use that over `eth_sign` for user safety

zzmp: Certain data may not be signable using V1/V3, but will always be signable using `eth_sign`.
Rather than adding complexity to support all versions, I think it's best for _us_ to reduce complexity/be less error prone by only supported the best case (V4) and the most general case (`eth_sign`).

marktoda: messages signed with eth_signTypedData_v3 will fail in permit2 as the contract itself verifies signatures using the v4 schema

marktoda: the docs on the differences are not great tbh lol but it seems that v3 doesn't support arrays which we need for permit2 batch permits? https://github.com/MetaMask/eth-sig-util/blob/554420a9bef69fd4307ad7a762fe81f875e57c76/src/sign-typed-data.ts#L193-L198

@snreynolds do you perchance know the diff b/t v3/v4?

tinaszheng: oh yeah if you need arrays then v3 won't cut it, its just that some wallets actually do v4 signing but the method name is just `eth_signTypedData` (this is an alias for general typed data signing not just v1)

PR #5849, opened 2023-01-18
chore: convert top level profile page to styled components

cbachmeier:
- Converts top level page file from Vanilla Extract to Styled Components
- Also moves shared const to a shared file to prevent local error where const isn't defined before used
- Generated files will be resolved by #5846 

No Wallet Connected
Screen Shot 2023-01-18 at 11 59 52
Profile (same with and without flag)
Screen Shot 2023-01-18 at 12 02 35
List with V2 Flag
Screen Shot 2023-01-18 at 12 02 45
List w/o V2 flag
Screen Shot 2023-01-18 at 12 02 55


src/nft/pages/profile/profile.tsx
```diff
@@ -47,42 +79,38 @@ const ProfileContent = () => {
   const cartExpanded = useBag((state) => state.bagExpanded)
   const isNftListV2 = useNftListV2Flag() === NftListV2Variant.Enabled
 
+  const pageWidthAdjustment =
+    cartExpanded && (!isNftListV2 || sellPageState === ProfilePageStateType.VIEWING)
+      ? SHOPPING_BAG_WIDTH
+      : isNftListV2 && sellPageState !== ProfilePageStateType.VIEWING
```
JackShort, line 85: I think this should be stored in a const like `var check = isNftListV2 && sellPageState !== ProfilePageStateType.VIEWING` then it could also be used above on line 83 as `!check`

src/nft/pages/profile/profile.tsx
```diff
@@ -47,42 +79,38 @@ const ProfileContent = () => {
...
+      ? LIST_PAGE_MARGIN * 2
+      : 0
+
+  const pageMargin = isNftListV2 && sellPageState !== ProfilePageStateType.VIEWING ? LIST_PAGE_MARGIN : 0
```
JackShort, line 89: it could also be used here

src/nft/pages/profile/profile.tsx
```diff
@@ -47,42 +79,38 @@ const ProfileContent = () => {
...
+
   return (
     <Trace page={InterfacePageName.NFT_PROFILE_PAGE} shouldLogImpression>
-      <Box className={styles.profileWrapper}>
+      <ProfilePageWrapper>
         {/* <Head> TODO: figure out metadata tagging
```
JackShort, line 94: we could probably remove this now

src/nft/pages/profile/profile.tsx
```diff
@@ -47,42 +79,38 @@ const ProfileContent = () => {
...
-                  ? SHOPPING_BAG_WIDTH
-                  : isNftListV2
-                  ? LIST_PAGE_MARGIN * 2
-                  : 0
-              }px)`,
-              margin: isNftListV2 ? `0px ${LIST_PAGE_MARGIN}px` : 'unset',
-            }}
-          >
+          <LoadedAccountPage pageWidthAdjustment={pageWidthAdjustment} pageMargin={pageMargin}>
             {sellPageState === ProfilePageStateType.VIEWING ? <ProfilePage /> : <ListPage />}
```
JackShort, line 99: honestly maybe this check is being so much it could also be put in a const

JackShort (approved):
a few comments

PR #5862, opened 2023-01-19
fix: allow signing from safepal

zzmp:
Uses an unversioned call for signTypedData so that it is accepted by all wallets. SafePal hangs on _v4.
Practically, this uses eth_signTypedData for all wallets _except_ for MetaMask, which rejects it and falls back to eth_signTypedData_v4.

Tested with:
- [x] SafePal
- [x] Coinbase Wallet
- [x] Trust
- [x] MetaMask
- [x] Zerion
- [x] Rainbow
- [x] Other

### DIFF
--- a/src/utils/signTypedData.ts
+++ b/src/utils/signTypedData.ts
@@ -23,7 +23,7 @@ JsonRpcSigner.prototype._signTypedData = async function signTypedDataWithFallbac
       ])
     } catch (error) {
       // MetaMask complains that the unversioned eth_signTypedData is formatted incorrectly (32602) - it prefers _v4.
-      if (typeof error.code === 'number' && error.code === -32602) {
+      if (error.code === -32602) {
         console.warn('eth_signTypedData failed, falling back to eth_signTypedData_v4:', error)
         return await this.provider.send('eth_signTypedData_v4', [
           address.toLowerCase(),
--- a/src/utils/signTypedData.ts
+++ b/src/utils/signTypedData.ts
@@ -24,6 +24,7 @@ JsonRpcSigner.prototype._signTypedData = async function signTypedDataWithFallbac
     } catch (error) {
       // MetaMask complains that the unversioned eth_signTypedData is formatted incorrectly (32602) - it prefers _v4.
       if (typeof error.code === 'number' && error.code === -32602) {
+        console.warn('eth_signTypedData failed, falling back to eth_signTypedData_v4:', error)
         return await this.provider.send('eth_signTypedData_v4', [
           address.toLowerCase(),
           JSON.stringify(_TypedDataEncoder.getPayload(populated.domain, types, populated.value)),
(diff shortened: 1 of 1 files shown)


src/utils/signTypedData.ts
```diff
@@ -15,11 +15,21 @@ JsonRpcSigner.prototype._signTypedData = async function signTypedDataWithFallbac
...
-      address.toLowerCase(),
-      JSON.stringify(_TypedDataEncoder.getPayload(populated.domain, types, populated.value)),
-    ])
+    try {
+      return await this.provider.send('eth_signTypedData', [
+        address.toLowerCase(),
+        JSON.stringify(_TypedDataEncoder.getPayload(populated.domain, types, populated.value)),
+      ])
+    } catch (error) {
+      if (typeof error.code === 'number' && error.code === -32602) {
```
tinaszheng, line 24: can you leave a comment describing the error code above this line? the `typeof` check also feels unnecessary

src/utils/signTypedData.ts
```diff
@@ -15,11 +15,21 @@ JsonRpcSigner.prototype._signTypedData = async function signTypedDataWithFallbac
...
+        return await this.provider.send('eth_signTypedData_v4', [
+          address.toLowerCase(),
+          JSON.stringify(_TypedDataEncoder.getPayload(populated.domain, types, populated.value)),
+        ])
+      }
+    }
   } catch (error) {
+    console.log(error)
```
tinaszheng, line 32: remove console

tinaszheng (changes requested):
hmm i would also test this out with other wallets, my hunch is that `eth_signTypedData_v4` should be tried before the generic `eth_signTypedData`

src/utils/signTypedData.ts
zzmp: typeof check should be necessary because error is unknown.

zzmp:
> hmm i would also test this out with other wallets, my hunch is that `eth_signTypedData_v4` should be tried before the generic `eth_signTypedData`

the unversioned must be tried first, because SafePal will hang on v4, so there's no way to tell it isn't working.

src/utils/signTypedData.ts
tinaszheng: i dont get it, `error.code === -32602` implicitly implies it's a number already

tinaszheng:
> the unversioned must be tried first, because SafePal will hang on v4, so there's no way to tell it isn't working.

wouldn't this be true the other way around if other wallets hang on unversioned?

src/utils/signTypedData.ts
tinaszheng: if `error` is unknown then you should be type guarding on the `error` object, not the internal `.code` property. for example if `error` is a primitive like `1` then this would fail - not a huge deal since we expect error formats to be relatively consistent, but at runtime this check isn't really doing anything

tinaszheng (approved):

PR #5863, opened 2023-01-19
fix: display specific gas estimation error

zzmp:
Display a specific gas estimation error, so it is not the more verbose (less specific) general error message.
CX-140


src/hooks/useUniversalRouter.ts
```diff
@@ -50,9 +50,8 @@ export function useUniversalRouterSwapCallback(
       try {
         gasEstimate = await provider.estimateGas(tx)
       } catch (gasError) {
-        await provider.call(tx) // this should throw the actual error
-        // If the actual error is not thrown, just try again:
-        gasEstimate = await provider.estimateGas(tx)
+        console.warn(gasError)
```
just-toby, line 53: is this the preexisting behavior or are we adding a new console log ? if new, do you think it's necessary?

just-toby (approved):

tinaszheng:
questionnn: why are we estimating gas at this step? seems extraneous to have to do it here vs just letting the wallet handle it

PR #5870, opened 2023-01-20
feat: [ListV2] Setup styles for Listing Modal

cbachmeier:
- Added styling to the V2 ListModal aligned with Figma
- Currently not functional ie you're not able to start the listing flow nor do the rows update to show the status of the listing flow
- Created a re-useable ListingModalSection which is used for both Approving Collections and Signing Listings
- Added an isVerified flag for Collection rows
- Moved the Modal files to their own folder

Screen Shot 2023-01-20 at 00 48 30
Screen Shot 2023-01-20 at 00 48 40


src/nft/components/profile/list/Modal/ListModalSection.tsx
```diff
@@ -0,0 +1,186 @@
...
+                text={<Trans>Listing an NFT requires a one-time marketplace approval for each NFT collection.</Trans>}
+              >
+                <StyledInfoIcon />
+              </MouseoverTooltip>
+            </Row>
+          )}
+          <ContentRowContainer>
+            {content.map((row, index) => {
+              return (
+                <ContentRow key={index}>
```
JackShort, line 168: nit: keys really shouldn't be indexes if it can be avoided

JackShort (approved):

PR #5874, opened 2023-01-20
feat: caching and polling on apollo token queries

cartcrom:
Removes loading state when navigating to already-viewed Token Details/Explore pages, also adds polling to Token Explore

### DIFF
--- a/src/graphql/data/TopTokens.ts
+++ b/src/graphql/data/TopTokens.ts
@@ -15,7 +15,15 @@ import {
   useTopTokens100Query,
   useTopTokensSparklineQuery,
 } from './__generated__/types-and-hooks'
-import { CHAIN_NAME_TO_CHAIN_ID, isPricePoint, PricePoint, toHistoryDuration, unwrapToken } from './util'
+import {
+  CHAIN_NAME_TO_CHAIN_ID,
+  isPricePoint,
+  PollingInterval,
+  PricePoint,
+  toHistoryDuration,
+  unwrapToken,
+  usePollQueryWhileMounted,
+} from './util'
 
 gql`
   query TopTokens100($duration: HistoryDuration!, $chain: Chain!) {
@@ -26,24 +34,30 @@ gql`
       address
       symbol
       market(currency: USD) {
+        id
         totalValueLocked {
+          id
           value
           currency
         }
         price {
+          id
           value
           currency
         }
         pricePercentChange(duration: $duration) {
+          id
           currency
           value
         }
         volume(duration: $duration) {
+          id
           value
           currency
         }
       }
       project {
+        id
         logoUrl
       }
     }
@@ -53,9 +67,13 @@ gql`
 gql`
   query TopTokensSparkline($duration: HistoryDuration!, $chain: Chain!) {
     topTokens(pageSize: 100, page: 1, chain: $chain) {
+      id
       address
+      chain
       market(currency: USD) {
+        id
         priceHistory(duration: $duration) {
+          id
           timestamp
           value
         }
@@ -64,11 +82,12 @@ gql`
   }
 `
 
-function useSortedTokens(tokens: NonNullable<TopTokens100Query['topTokens']>) {
+function useSortedTokens(tokens: TopTokens100Query['topTokens']) {
   const sortMethod = useAtomValue(sortMethodAtom)
   const sortAscending = useAtomValue(sortAscendingAtom)
 
   return useMemo(() => {
+    if (!tokens) return undefined
     let tokenArray = Array.from(tokens)
     switch (sortMethod) {
       case TokenSortMethod.PRICE:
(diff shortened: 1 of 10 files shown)


src/graphql/data/util.tsx
```diff
@@ -1,9 +1,31 @@
...
 
+export enum PollingInterval {
+  Slow = ms`5m`,
+  Normal = ms`1m`,
+  Fast = ms`12s`, // 12 seconds, block times for mainnet
+  LightningMcQueen = ms`3s`, // 3 seconds, approx block times for polygon
+}
+
+// Polls a query only when the current component is mounted, as useQuery's pollInterval prop will continue to poll after unmount
+export function usePollQueryWhileMounted<T, K>(queryResult: QueryResult<T, K>, interval: PollingInterval) {
```
cbachmeier, line 18: Do we need to create a new wrapper function or could we just leverage useInterval in the apollo fetching functions.

src/pages/TokenDetails/index.tsx
```diff
@@ -14,38 +14,41 @@ export const pageTimePeriodAtom = atomWithStorage<TimePeriod>('tokenDetailsTimeP
...
+
+  if (!tokenQuery) return <TokenDetailsPageSkeleton />
 
   return (
     <TokenDetails
       urlAddress={tokenAddress}
       chain={chain}
       tokenQuery={tokenQuery}
-      tokenPriceQuery={tokenPriceQuery}
+      tokenPriceQuery={currentPriceData}
```
cbachmeier, line 51: we should rename this param to match

src/pages/TokenDetails/index.tsx
```diff
@@ -14,38 +14,41 @@ export const pageTimePeriodAtom = atomWithStorage<TimePeriod>('tokenDetailsTimeP
...
+      chain,
       duration,
     },
   })
 
-  if (!tokenQuery || tokenQueryLoading) return <TokenDetailsPageSkeleton />
+  const [currentPriceData, setCurrentPriceData] = useState(tokenPriceQuery)
+  useEffect(() => {
+    if (tokenPriceQuery) setCurrentPriceData(tokenPriceQuery)
+  }, [setCurrentPriceData, tokenPriceQuery])
```
cbachmeier, line 42: is this needed if the fetch function is memoized?

src/graphql/data/TopTokens.ts
```diff
@@ -26,24 +34,30 @@ gql`
       address
       symbol
       market(currency: USD) {
+        id
```
cbachmeier, line 37: did you use a lib to add the `id` field or was this done manually?

cartcrom: id's are needed for caching/normalization to work, had to do this manually.

src/graphql/data/util.tsx
cartcrom: I think it makes more sense to use Apollo's polling functions over our own interval. This doesn't need to be a wrapper function, but I think it will reduce duplicate code whenever this functionality is needed.

src/pages/TokenDetails/index.tsx
cartcrom: When the timePeriod changes, tokenPriceQuery becomes null; I added this extra currentPriceData variable so that in the chart, we display existing data until new data  is available, rather than repeatedly showing a loading state

src/pages/TokenDetails/index.tsx
```diff
@@ -14,38 +14,41 @@ export const pageTimePeriodAtom = atomWithStorage<TimePeriod>('tokenDetailsTimeP
...
   const isNative = tokenAddress === NATIVE_CHAIN_ID
   const [timePeriod, setTimePeriod] = useAtom(pageTimePeriodAtom)
-  const [contract, duration] = useMemo(
-    () => [
-      { address: isNative ? nativeOnChain(pageChainId).wrapped.address : tokenAddress ?? '', chain },
-      toHistoryDuration(timePeriod),
-    ],
-    [chain, isNative, pageChainId, timePeriod, tokenAddress]
+  const [address, duration] = useMemo(
+    () => [isNative ? getNativeTokenDBAddress(chain) : tokenAddress ?? '', toHistoryDuration(timePeriod)],
```
tinaszheng, line 20: if you set up the `tokenAddress === NATIVE_CHAIN_ID` check as a typescript guard function like:

```isNativeAddress(tokenAddress: string): tokenAddress is string { return  tokenAddress === NATIVE_CHAIN_ID}
```

then you'd be able to skip this `tokenAddress ?? ''` coalescing bcuz typescript will know it's always defined :P

src/pages/TokenDetails/index.tsx
```diff
@@ -14,38 +14,41 @@ export const pageTimePeriodAtom = atomWithStorage<TimePeriod>('tokenDetailsTimeP
...
-      contract,
+      address,
+      chain,
       duration,
     },
   })
 
-  if (!tokenQuery || tokenQueryLoading) return <TokenDetailsPageSkeleton />
+  const [currentPriceQuery, setCurrentPriceQuery] = useState(tokenPriceQuery)
+  useEffect(() => {
```
tinaszheng, line 40: can you write a comment above here that explains why this state + useEffect is necessary? took me a bit of parsing to understand, but is it basically like `tokenPriceQuery` can sometimes be null, so use state to make sure it's always defined?

src/pages/TokenDetails/index.tsx
cartcrom: but then a non-native address would be incorrectly seen by typescript as NOT a string?

PR #5887, opened 2023-01-28
fix: nav text is wrap

chuanyu0201:
Chinese translator, pool navigator can wrap lines
image


cbachmeier (approved):
Thanks for the submission

just-toby (approved):

cbachmeier:
This change can cause other elements to overlap when the wrap doesn't occur. We're looking to significantly improve our translations holistically, thank you for calling it out.

PR #5889, opened 2023-01-30
feat: create Save-for-Later bag

gnewfield:
This PR adds a Save-for-Later (SFL) tab to the existing Bag feature, allowing customers to keep track of items which they may want to purchase. Previously, the bag interface only had an option for customers to remove items. Now, they may move items between their main and SFL bags.

This feature is implemented in an extensible manner, in order to reuse functionality across the existing bag types (Main, Sell, Save-for-Later) and make it easier to add new bag types in the future (e.g. Wishlist, Category-specific bags, etc). 

Check out a recorded demo here: https://youtu.be/vx7YDi9MHTs

### DIFF
--- a/src/nft/components/bag/SaveForLaterContent.tsx
+++ b/src/nft/components/bag/SaveForLaterContent.tsx
@@ -1,0 +1,47 @@
+import { Column } from 'nft/components/Flex'
+import { useBag, useIsMobile } from 'nft/hooks'
+import { BagItemStatus } from 'nft/types'
+import { fetchPrice, recalculateBagUsingPooledAssets } from 'nft/utils'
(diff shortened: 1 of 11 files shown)


src/nft/components/bag/SaveForLaterContent.tsx
```diff
@@ -0,0 +1,47 @@
...
+    return { saveForLaterAssets }
+  }, [itemsInBag])
+
+  const isMobile = useIsMobile()
+
+  return (
+    <Column>
+      {saveForLaterAssets
+        .slice(0)
+        .reverse()
```
justindomingue, line 33: why not do array manipulatin operations in the memo above?

PR #5892, opened 2023-01-30
fix: center playback switch on video asset cards

gnewfield:
On collection pages, hovering over an asset with video content is supposed to display a playback switch button on the center of the card. However, the switch is currently appearing in the bottom right corner. This PR applies new styling to properly position the switch. Attaching screenshots for reference.
before
after


cbachmeier:
Thanks for the PR however we did intentionally want the playback button on the bottom right as opposed to the center so as to be less distracting from the image themselves

PR #5894, opened 2023-01-30
feat: refetch recently searched

cartcrom:
Fixes misaligned search results across the search bar's search results, "recently fetched", and "popular" sections using a "RecentlySearched" tokens/nfts query and the apollo cache. Also ensures that token details pages will have the same prices search bar prices.


src/components/NavBar/RecentlySearchedAssets.ts
```diff
@@ -0,0 +1,81 @@
...
+  const history = useAtomValue(recentlySearchedAssetsAtom)
+  const shortenedHistory = useMemo(() => history.slice(0, 4), [history])
+
+  const { data: queryData, loading } = useRecentlySearchedAssetsQuery({
+    variables: {
+      collectionAddresses: shortenedHistory.filter((asset) => asset.isNft).map((asset) => asset.address),
+      contracts: shortenedHistory
+        .filter((asset) => !asset.isNft)
+        .map((token) => ({
+          address: token.address === 'NATIVE' ? (null as unknown as string) : token.address,
```
cartcrom, line 39: `(null as unknown as string)` is a temporary measure until the backend accepts "NATIVE" as input. We currently need to pass null for caching to work properly, but our apollo setup doesn't accept null.

cbachmeier: I'm not exactly sure what's causing it but I'm seeing WETH show up for Popular Tokens instead of ETH
Screen Shot 2023-01-31 at 21 35 13

cbachmeier: Same thing happens with MATIC on polygon
Screen Shot 2023-01-31 at 21 47 17

cbachmeier (approved):
Other than the issue with wrapped natives showing, this LGTM. I believe the wrapped natives isn't related to this change but it somewhere else on this branch?

src/components/NavBar/RecentlySearchedAssets.ts
cartcrom: great callout, meant to unwrap those in the [migrate_trending](https://github.com/Uniswap/interface/tree/migrate_trending) pr, just fixed on there 👍

PR #5897, opened 2023-01-30
feat: quote token price for nfts

JackShort:
* displays price of new token
* does not implement loading indicator

Screen Shot 2023-01-30 at 5 31 48 PM

### DIFF
--- a/src/nft/components/bag/BagFooter.tsx
+++ b/src/nft/components/bag/BagFooter.tsx
@@ -1,9 +1,10 @@
 import { BigNumber } from '@ethersproject/bignumber'
+import { formatEther } from '@ethersproject/units'
 import { parseEther } from '@ethersproject/units'
 import { Trans } from '@lingui/macro'
 import { TraceEvent } from '@uniswap/analytics'
 import { BrowserEvent, InterfaceElementName, NFTEventName } from '@uniswap/analytics-events'
-import { Currency } from '@uniswap/sdk-core'
+import { Currency, TradeType } from '@uniswap/sdk-core'
 import { useWeb3React } from '@web3-react/core'
 import Column from 'components/Column'
 import Loader from 'components/Loader'
@@ -12,6 +13,8 @@ import Row from 'components/Row'
 import { SupportedChainId } from 'constants/chains'
 import { PayWithAnyTokenVariant, usePayWithAnyTokenFlag } from 'featureFlags/flags/payWithAnyToken'
 import { useCurrency } from 'hooks/Tokens'
+import { useBestTrade } from 'hooks/useBestTrade'
+import tryParseCurrencyAmount from 'lib/utils/tryParseCurrencyAmount'
 import { useBag } from 'nft/hooks/useBag'
 import { useTokenInput } from 'nft/hooks/useTokenInput'
 import { useWalletBalance } from 'nft/hooks/useWalletBalance'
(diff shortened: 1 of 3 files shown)


src/nft/components/bag/BagFooter.tsx
```diff
@@ -221,12 +236,16 @@ export const BagFooter = ({
...
+                    : ethNumberStandardFormatter(swapTrade?.inputAmount.toExact())
+                  : formatWeiToDecimal(totalEthPrice.toString())}
+                &nbsp;{activeCurrency?.symbol ?? 'ETH'}
               </ThemedText.HeadlineSmall>
-              <ThemedText.BodySmall color="textSecondary" lineHeight="20px">{`${ethNumberStandardFormatter(
-                totalUsdPrice,
-                true
-              )}`}</ThemedText.BodySmall>
+              <ThemedText.BodySmall color="textSecondary" lineHeight="20px">
+                {`${ethNumberStandardFormatter(totalUsdPrice, true)}`}
```
cbachmeier, line 247: Looks like there's discrepancy between stable swap inputAmount and totalUsdPrice. We may want to change our usd price fetcher to use the same swap logic for a stable so estimates match

src/nft/components/bag/BagFooter.tsx
```diff
@@ -221,12 +236,16 @@ export const BagFooter = ({
                 <Trans>Total</Trans>
               </ThemedText.SubHeaderSmall>
               <ThemedText.HeadlineSmall>
-                {formatWeiToDecimal(totalEthPrice.toString())}&nbsp;{activeCurrency?.symbol ?? 'ETH'}
+                {inputCurrency
+                  ? swapState !== TradeState.VALID
+                    ? '-'
```
cbachmeier, line 241: should we disable buy button when swap state is invalid?

src/nft/components/bag/BagFooter.tsx
cbachmeier: Looks like there is a useStablecoinPrice hook

src/nft/components/bag/BagFooter.tsx
```diff
@@ -199,6 +203,17 @@ export const BagFooter = ({
   const isPending = PENDING_BAG_STATUSES.includes(bagStatus)
   const activeCurrency = inputCurrency ?? defaultCurrency
 
+  const parsedAmount = useMemo(() => {
+    if (!inputCurrency) return undefined
+    return tryParseCurrencyAmount(formatEther(totalEthPrice.toString()), defaultCurrency ?? undefined)
```
cbachmeier, line 208: should the second input for `tryParseCurrencyAmount` be inputCurrency? I'm a little confused why it's not used by this function

JackShort: this is just to convert the total amount of eth to a CurrencyAmount not the inputCurrency

src/nft/components/bag/BagFooter.tsx
JackShort: not right now because I am going to setup some disabling logic when I integrate with permit2 which is in the next pr

JackShort: also this isn't even the right state checking logic - I need to talk to design about some stuff so I am waiting on implementing it yet also

src/nft/components/bag/BagFooter.tsx
JackShort: wondering if we should change the eth pricing to use the same hook or keep it with the way we have been previously doing it

src/nft/components/bag/BagFooter.tsx
JackShort: basically this function takes in a string and the currency you want. so like '78.2' and 'ETH' and it tries to create a `CurrencyAmount` from that input

src/nft/components/bag/BagFooter.tsx
cbachmeier: I think the best course would be to replace the 3rd party price fetching with the useStable hook to help consolidate. Added a tech debt ticket https://uniswaplabs.atlassian.net/browse/NFT-1062

cbachmeier (approved):
LGTM

PR #5901, opened 2023-01-31
refactor: L2 icons in component

cartcrom:
Updates L2 icon to be part of AssetLogo component. Pre-req for moving to widget repo


src/components/Tokens/TokenDetails/BalanceSummary.tsx
```diff
@@ -81,7 +81,7 @@ export default function BalanceSummary({ token }: { token: Currency }) {
           <Trans>Your balance on {label}</Trans>
         </ThemedText.SubHeaderSmall>
         <BalanceRow>
-          <CurrencyLogo currency={token} size="2rem" />
+          <CurrencyLogo currency={token} size="2rem" hideL2Icon={false} />
```
just-toby, line 84: we should make this prop false by default, so you can omit it here

just-toby (approved):
LGTM. does it make sense to add a test for the new AssetLogo functionality ?

PR #5903, opened 2023-01-31
chore: removeing fetch usd price for bag footer

JackShort:
* Removes the call to fetch the usdPrice and gets it from the `useStablecoinValue` hook for the bag footer
* NOTE: when testing with pay with any token it breaks when going back from WETH -> ETH. This is actually a bug in `useStablecoinPrice` and happens on swap I am looking into it #5904


cbachmeier (approved):
LGTM

PR #5906, opened 2023-02-01
feat: [ListV2] Add mobile styles

cbachmeier:
Adds mobile support for Listing Page V2 and the Listing Modal
Screen Shot 2023-02-01 at 09 42 34
Screen Shot 2023-02-01 at 09 47 58

### DIFF
--- a/src/nft/components/profile/list/ListPage.tsx
+++ b/src/nft/components/profile/list/ListPage.tsx
@@ -9,7 +9,7 @@ import { BackArrowIcon } from 'nft/components/icons'
 import { headlineLarge, headlineSmall } from 'nft/css/common.css'
 import { themeVars } from 'nft/css/sprinkles.css'
 import { useBag, useIsMobile, useNFTList, useProfilePageState, useSellAsset } from 'nft/hooks'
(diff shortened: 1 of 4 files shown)


src/nft/components/profile/list/ListPage.tsx
```diff
@@ -98,6 +98,12 @@ const FloatingConfirmationBar = styled(Row)`
   transform: translateX(-50%);
   max-width: 1200px;
   z-index: ${Z_INDEX.under_dropdown};
+
+  @media screen and (max-width: ${BREAKPOINTS.sm}px) {
+    width: calc(100% - ${LIST_PAGE_MARGIN_MOBILE * 2}px);
```
JackShort, line 103: slightly confused if this needs the isOnV2Page check or not?

JackShort (approved):

src/nft/components/profile/list/ListPage.tsx
cbachmeier: The `FloatingConfirmationBar` component is only shown when on V2

PR #5909, opened 2023-02-01
feat: create feature flag for swap widget

just-toby:
adds a new feature flag to the interface for toggling the Widget in the main swap page.

also made a small change to the Widget code so we can preset the INPUT currency instead of the OUTPUT currency


src/pages/Swap/index.tsx
```diff
@@ -559,319 +562,329 @@ export default function Swap({ className }: { className?: string }) {
...
-                </TraceEvent>
-              </ArrowWrapper>
-            </div>
-            <AutoColumn gap="md">
-              <div>
-                <OutputSwapSection showDetailsDropdown={showDetailsDropdown}>
-                  <Trace section={InterfaceSectionName.CURRENCY_OUTPUT_PANEL}>
+          {swapWidgetEnabled ? (
+            <Widget token={loadedInputCurrency ?? undefined} width="100%" defaultField={Field.INPUT} />
+          ) : (
```
just-toby, line 567: this is the only change in this file - it just looks big  because the formatting changed

cartcrom (approved):

PR #5912, opened 2023-02-02
fix: adding usd_value back to trace event

JackShort:
* accidentally removed `usd_value` adding it back

### DIFF
--- a/src/nft/components/bag/BagFooter.tsx
+++ b/src/nft/components/bag/BagFooter.tsx
@@ -207,6 +207,10 @@ export const BagFooter = ({ totalEthPrice, bagStatus, fetchAssets, eventProperti
   )
 
   const usdcValue = useStablecoinValue(inputCurrency ? swapTrade?.inputAmount : parsedAmount)
+  const traceEventProperties = {
+    usd_value: usdcValue?.toExact(),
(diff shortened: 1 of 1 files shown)


src/nft/components/bag/BagFooter.tsx
```diff
@@ -207,6 +207,10 @@ export const BagFooter = ({ totalEthPrice, bagStatus, fetchAssets, eventProperti
   )
 
   const usdcValue = useStablecoinValue(inputCurrency ? swapTrade?.inputAmount : parsedAmount)
+  const traceEventProperties = {
+    usd_value: usdcValue?.toExact(),
+    ...eventProperties,
```
cbachmeier, line 212: The past in eventsProperty is instantiated in `Bag` but only ever passed down. To reduce `BagFooter` args you could instantiate here and include the usd_value

cbachmeier (approved):
LGTM non-blocking suggestion

src/nft/components/bag/BagFooter.tsx
JackShort: actually it still needs to be passed down because the other event properties come from the items in the bag that aren't being passed to the footer

PR #5916, opened 2023-02-02
feat: widget on tx success callback

just-toby:
currently we're not logging this event for widget transactions, because the [current logging logic](https://github.com/Uniswap/interface/blob/9719af66e5a74a5a3c58ed6333cba1340a838145/src/state/transactions/updater.tsx#L84) relies on interface-level state to exist (for a widget transaction, that state is managed in the widget instead, so the event is not logged)

note: this should be applied to the widget in token details too, so i'm not gating it behind the feature flag

image


just-toby:
this needs to be rebased onto #5915 for the test to pass

lynnshaoyu (approved):

PR #5917, opened 2023-02-02
fix: landing page styling of widget

just-toby:
fixes the landing page styling for the widget-enabled variant (see preview link for the difference between test and control)

### DIFF
--- a/src/pages/Landing/index.tsx
+++ b/src/pages/Landing/index.tsx
@@ -6,6 +6,7 @@ import Card, { CardType } from 'components/About/Card'
 import { MAIN_CARDS, MORE_CARDS } from 'components/About/constants'
 import ProtocolBanner from 'components/About/ProtocolBanner'
 import { BaseButton } from 'components/Button'
+import { useSwapWidgetEnabled } from 'featureFlags/flags/swapWidget'
 import Swap from 'pages/Swap'
 import { parse } from 'qs'
 import { useEffect, useRef, useState } from 'react'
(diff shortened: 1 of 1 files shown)


src/graphql/thegraph/__generated__/types-and-hooks.ts
```diff
@@ -37,6 +37,7 @@ export type Bundle = {
 export type Bundle_Filter = {
```
zzmp, line 37: This should be changed in a separate PR, right?

src/pages/Landing/index.tsx
```diff
@@ -306,9 +326,15 @@ export default function Landing() {
               name={SharedEventName.ELEMENT_CLICKED}
               element={InterfaceElementName.LANDING_PAGE_SWAP_ELEMENT}
             >
-              <Link to="/swap">
-                <LandingSwap />
-              </Link>
+              {swapWidgetEnabled ? (
```
zzmp, line 329: nit: You can inline this: `{useSwapWidgetEnabled() ? (`.

zzmp (changes requested):

src/pages/Landing/index.tsx
just-toby: i tried that but we can't because it would be conditionally called

src/graphql/thegraph/__generated__/types-and-hooks.ts
just-toby: i think we're supposed to check these changes in, according to @cbachmeier

zzmp: but nothing in this PR should have generated this. It's not related to this PR's changes, and should be checked in through a separate PR.

zzmp (changes requested):
types-and-hooks.ts should still be reverted

src/graphql/thegraph/__generated__/types-and-hooks.ts
just-toby: ok, removed

zzmp (approved):

PR #5924, opened 2023-02-04
feat: [ListV2] Success modal + Twitter Share

cbachmeier:
Creates a Success Screen when all listings have been successful. Shows the successfull listings, a return to profile button, and a twitter share button.

Screenshots:
Modal with 1 listing
Screen Shot 2023-02-03 at 17 24 20
Multiple Listings
Screen Shot 2023-02-03 at 17 26 58
Mobile
Screen Shot 2023-02-03 at 17 27 06

Tweets
Listing Multiple Items
Screen Shot 2023-02-03 at 17 22 54
Listing Single Item
Screen Shot 2023-02-03 at 17 24 30


src/nft/components/profile/list/Modal/SuccessScreen.tsx
```diff
@@ -0,0 +1,128 @@
...
+import { useMemo } from 'react'
+import { Twitter, X } from 'react-feather'
+import styled, { css, useTheme } from 'styled-components/macro'
+import { BREAKPOINTS, ThemedText } from 'theme'
+import { formatDollar, priceToPreciseFloat } from 'utils/formatNumbers'
+
+import { TitleRow } from '../shared'
+
+const SuccessImage = styled.img<{ numImages: number }>`
+  width: calc(${({ numImages }) => (numImages > 1 ? (numImages > 2 ? '33%' : '50%') : '100%')} - 16px);
```
JackShort, line 18: what does the `- 16px` do here and can it not be changed to padding instead?

src/nft/components/profile/list/Modal/SuccessScreen.tsx
```diff
@@ -0,0 +1,128 @@
...
+      <Row justify="space-between" align="flex-start" marginBottom="16px">
+        <ThemedText.SubHeader lineHeight="24px">
+          <Trans>Proceeds if sold</Trans>
+        </ThemedText.SubHeader>
+        <ProceedsColumn>
+          <ThemedText.SubHeader lineHeight="24px">{formatEth(totalEthListingValue)} ETH</ThemedText.SubHeader>
+          {nativeCurrencyPrice && (
+            <ThemedText.BodySmall lineHeight="20px" color="textSecondary">
+              {formatDollar({
+                num: priceToPreciseFloat(nativeCurrencyPrice) ?? 0 * totalEthListingValue,
```
JackShort, line 108: instead of doing this you could parse `totalEthListingValue` into a `CurrencyAmount` and just use the `useStablecoinValue` hook on that to get the price

src/nft/utils/asset.ts
```diff
@@ -44,3 +45,36 @@ export const generateTweetForPurchase = (assets: UpdatedGenieAsset[], txHashUrl:
   } with @Uniswap 🦄\n\nhttps://app.uniswap.org/#/nfts/collection/0x60bb1e2aa1c9acafb4d34f71585d7e959f387769\n${txHashUrl}`
   return `https://twitter.com/intent/tweet?text=${encodeURIComponent(tweetText)}`
 }
+
+function getMinListingPrice(listings: Listing[]): number {
+  return min(listings.map((listing) => listing.price ?? 0)) ?? 0
```
JackShort, line 50: you don't need to import from `d3` you can just do `return Math.min(...listings.map((listing) => listing.price ?? 0)) ?? 0`

src/nft/utils/asset.ts
```diff
@@ -44,3 +45,36 @@ export const generateTweetForPurchase = (assets: UpdatedGenieAsset[], txHashUrl:
...
+        .filter((asset) => asset.collection?.twitterUrl === collection || asset.collection?.name === collection)
+        .map((asset) => asset.name ?? ''),
+    }
+  })
+}
+
+export const generateTweetForList = (assets: WalletAsset[]): string => {
+  const tweetText =
+    assets.length == 1
+      ? `I just listed ${assets[0].collection?.twitterUrl ? `${assets[0].collection?.twitterUrl} ` : ''}${
```
JackShort, line 69: you should check if the twitterUrl is defined or else it returns null for instance LilNouns:
Screenshot 2023-02-04 at 2 33 20 PM

JackShort (approved):
looks good and works well - some small htings

src/nft/utils/asset.ts
cbachmeier: Great catch, this was actually an issue not with the tweet generator but how we parsed in the twitterName from gql. Fixed now.

src/nft/components/profile/list/Modal/SuccessScreen.tsx
cbachmeier: It accounts for the gap from the spacing between images. Attempting to use padding instead causes issues with the border radius

PR #5936, opened 2023-02-06
feat: [ListV2] Changes based off Bug Bash

cbachmeier:
A couple of changes requested from the debugging session:
- Improved grid responsiveness at tablet and mobile breakpoints
- Dynamic button width on mobile
- Duration warning text no longer wraps on large inputs
- Don't allow listings at the same price as current
- Don't show loading spinner on listing button in V2


JackShort (approved):
lgtm

PR #5937, opened 2023-02-07
fix: default to null when query address is undefined

cartcrom:
fixes issue where non-addressed native tokens would not normalize in cache correctly.

moves interface away from using `"ETH"` convention for natives, matches mobile's usage of `null`


cbachmeier (approved):
LGTM

src/utils/nativeTokens.ts
```diff
@@ -12,6 +12,6 @@ export function getNativeTokenDBAddress(chain: Chain): string {
     case Chain.Arbitrum:
     case Chain.EthereumGoerli:
     case Chain.Optimism:
-      return 'ETH'
+      return undefined
```
tinaszheng, line 15: i would just return `null` here so you don't need to override in the actual query

PR #5938, opened 2023-02-07
fix: [ListV2] Better Collection and Listing Retry and Removal Logic

cbachmeier:
- Previously removing a collection after rejecting/failing would get the user stuck where all approval states would reset and there would be no way to restart the flow. This maintains approved, rejected, and failed statuses to allow the user to continue on. 
- The old method for updating a rows status and callback would did not account for row removal, I've updated the logic to be dynamic and tied to the zustand hook
- Also fixes a bug where removing a collection would create duplicate approval rows.

To test you should try listing and approving multiple collections and markets on V1 and V2 and be able to reject, remove, and retry as expected

### DIFF
--- a/src/nft/components/bag/profile/ListingButton.tsx
+++ b/src/nft/components/bag/profile/ListingButton.tsx
@@ -211,7 +211,7 @@ export const ListingButton = ({ onClick, buttonText, showWarningOverride = false
 
   const warningWrappedClick = () => {
     if ((!disableListButton && canContinue) || showWarningOverride) {
-      if (issues && isNftListV2 && !showResolveIssues) toggleShowResolveIssues()
+      if (issues && isNftListV2) !showResolveIssues && toggleShowResolveIssues()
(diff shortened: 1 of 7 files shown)


src/nft/components/bag/profile/ListingButton.tsx
```diff
@@ -211,7 +211,7 @@ export const ListingButton = ({ onClick, buttonText, showWarningOverride = false
 
   const warningWrappedClick = () => {
     if ((!disableListButton && canContinue) || showWarningOverride) {
-      if (issues && isNftListV2 && !showResolveIssues) toggleShowResolveIssues()
+      if (issues && isNftListV2) !showResolveIssues && toggleShowResolveIssues()
```
cbachmeier, line 214: This was actually correct. I basically want this click to do nothing if there is issues and we're already showing the 'Resolve Issues' text. Including`showResolveIssues` in the check triggered the onClick function when we wanted it to do nothing

JackShort (approved):

PR #5941, opened 2023-02-07
fix: handle wallets that don't send default 4001 error code for user rejected txns

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-2809?atlOrigin=eyJpIjoiYTdkNmFjMDk2YzU4NDAyY2JkOWJkNDZjNWQ2OWMxNmIiLCJwIjoiaiJ9

### DIFF
--- a/src/utils/swapErrorToUserReadableMessage.tsx
+++ b/src/utils/swapErrorToUserReadableMessage.tsx
@@ -22,6 +22,16 @@ export function swapErrorToUserReadableMessage(error: any): string {
     error = error.error ?? error.data?.originalError
   }
 
+  // The 4001 error code doesn't capture the case where users reject a transaction for all wallets.
+  if (
+    // For rainbow wallet:
+    (reason?.toLowerCase().includes('request') && reason?.toLowerCase().includes('reject')) ||
+    // For frame wallet: https://frame.sh/
+    reason?.toLowerCase().includes('declined')
+  ) {
+    return t`Transaction rejected`
+  }
+
   if (reason?.indexOf('execution reverted: ') === 0) reason = reason.substr('execution reverted: '.length)
 
   switch (reason) {
--- a/src/utils/swapErrorToUserReadableMessage.tsx
+++ b/src/utils/swapErrorToUserReadableMessage.tsx
@@ -26,9 +26,11 @@ export function swapErrorToUserReadableMessage(error: any): string {
   // so we need to parse the reason for these special cases:
   if (
     // For rainbow wallet:
-    (reason?.toLowerCase().includes('request') && reason?.toLowerCase().includes('reject')) ||
+    (reason?.match(/request/i) && reason.match(/reject/i)) ||
     // For frame wallet: https://frame.sh/
-    reason?.toLowerCase().includes('declined')
+    reason?.match(/declined/i) ||
+    // For safepal:
+    reason?.toLowerCase().includes('cancelled by user')
   ) {
     return t`Transaction rejected`
   }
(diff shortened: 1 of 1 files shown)


src/utils/swapErrorToUserReadableMessage.tsx
```diff
@@ -22,6 +22,17 @@ export function swapErrorToUserReadableMessage(error: any): string {
     error = error.error ?? error.data?.originalError
   }
 
+  // The 4001 error code doesn't capture the case where users reject a transaction for all wallets,
+  // so we need to parse the reason for these special cases:
+  if (
+    // For rainbow wallet:
+    (reason?.toLowerCase().includes('request') && reason?.toLowerCase().includes('reject')) ||
```
zzmp, line 29: Could you simplify this by using regex? Do you know the ordering of `request` and `reject`?

    if (
      reason?.match(/request/i) && reason.match(/reject/i) ||
      reason.match(/declined/i)
    ) {

lynnshaoyu: we want to handle the case where request appears before reject and vice versa as both occur (see here: https://uniswapteam.slack.com/archives/C047U65H422/p1675798801630839). i don't think regex is necessary for what we're trying to do here for the sake of readability / simplicity ?

zzmp: I'd prefer regex - I think it is simpler because it omits toLowerCase and is still very clear in what it does (`/i` means case-insensitive) - but this is not a blocker.

src/utils/swapErrorToUserReadableMessage.tsx
```diff
@@ -22,6 +22,17 @@ export function swapErrorToUserReadableMessage(error: any): string {
...
+    // For frame wallet: https://frame.sh/
+    reason?.toLowerCase().includes('declined')
+  ) {
```
zzmp, line 32: SafePal says "Cancelled by User" 🤦

src/utils/swapErrorToUserReadableMessage.tsx
lynnshaoyu: sg that makes sense done!

src/utils/swapErrorToUserReadableMessage.tsx
lynnshaoyu: oof ty ty added

src/utils/swapErrorToUserReadableMessage.tsx
```diff
@@ -22,6 +22,19 @@ export function swapErrorToUserReadableMessage(error: any): string {
     error = error.error ?? error.data?.originalError
   }
 
+  // The 4001 error code doesn't capture the case where users reject a transaction for all wallets,
+  // so we need to parse the reason for these special cases:
+  if (
+    // For rainbow wallet:
+    (reason?.match(/request/i) && reason?.match(/reject/i)) ||
+    // For frame wallet: https://frame.sh/
```
zzmp, line 30: nit: no need to include the url, you don't do so for any other wallet
nit: capitalize the wallet names (SafePal has two caps)
nit: omit or include "wallet" for every named wallet

none of these are blocking at all

src/utils/swapErrorToUserReadableMessage.tsx
```diff
@@ -22,6 +22,19 @@ export function swapErrorToUserReadableMessage(error: any): string {
...
+    reason?.match(/declined/i) ||
+    // For safepal:
+    reason?.match('cancelled by user')
```
zzmp, line 33: use `/cancelled by user/i` instead of a string so it's clear that it is case-insensitive.

zzmp (approved):
approved pending the change to `/cancelled by user/i`

PR #5943, opened 2023-02-07
fix: URL params for widget

just-toby:
enables the callsites of our Widget integration to pass either/both of the input and output currencies. 

this is needed to handle URL params for the /swap page. for example, see this link: 

https://app.uniswap.org/#/swap?inputCurrency=0x6B175474E89094C44Da98b954EedeAC495271d0F&outputCurrency=0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48

we still keep support for specifying a "default" token, which is currently just used to do automatic navigations on the token details page.

### DIFF
--- a/src/components/Tokens/TokenDetails/index.tsx
+++ b/src/components/Tokens/TokenDetails/index.tsx
@@ -217,7 +217,6 @@ export default function TokenDetails({
 
         <RightPanel>
           <Widget
-            outputToken={token ?? undefined}
             defaultToken={token ?? undefined}
             onDefaultTokenChange={navigateToWidgetSelectedToken}
             onReviewSwapClick={onReviewSwapClick}
--- a/src/components/Tokens/TokenDetails/index.tsx
+++ b/src/components/Tokens/TokenDetails/index.tsx
@@ -217,7 +217,9 @@ export default function TokenDetails({
 
         <RightPanel>
           <Widget
-            defaultToken={token ?? undefined}
+            defaultTokens={{
+              default: token ?? undefined,
+            }}
             onDefaultTokenChange={navigateToWidgetSelectedToken}
             onReviewSwapClick={onReviewSwapClick}
           />
--- a/src/components/Tokens/TokenDetails/index.tsx
+++ b/src/components/Tokens/TokenDetails/index.tsx
@@ -2,7 +2,6 @@ import { Trans } from '@lingui/macro'
 import { Trace } from '@uniswap/analytics'
 import { InterfacePageName } from '@uniswap/analytics-events'
 import { Currency } from '@uniswap/sdk-core'
-import { Field } from '@uniswap/widgets'
 import { useWeb3React } from '@web3-react/core'
 import CurrencyLogo from 'components/Logo/CurrencyLogo'
 import { AboutSection } from 'components/Tokens/TokenDetails/About'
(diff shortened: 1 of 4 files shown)


src/components/Tokens/TokenDetails/index.tsx
```diff
@@ -218,9 +217,9 @@ export default function TokenDetails({
 
         <RightPanel>
           <Widget
-            token={token ?? undefined}
-            defaultField={Field.OUTPUT}
-            onTokenChange={navigateToWidgetSelectedToken}
+            outputToken={token ?? undefined}
```
zzmp, line 220: Why do you have `outputToken` _and_ `defaultToken` here? Will the token details "page token" now always match the output token?

zzmp (changes requested):

src/components/Tokens/TokenDetails/index.tsx
just-toby: no, the page token can be switched to input. we only navigate (to output) when the page token is replaced.

i set both props because I thought it was clearer, instead of assuming that the default token will be assigned to the Field you expect. 

e.g. this API would support a situation where the Input is the "default" (i.e. required) token, if you specify all 3 props like this:

```
inputToken={USDC}
outputToken={DAI}
defaultToken={USDC}
```

but since we do automatically assign default -> output, we can remove one here like you suggest

src/components/Widget/inputs.tsx
```diff
@@ -20,30 +20,39 @@ function includesDefaultToken(tokens: SwapTokens) {
...
+        const update = {
+          ...tokens,
+          [Field.INPUT]: inputToken ?? tokens[Field.INPUT],
+          [Field.OUTPUT]: outputToken ?? tokens[Field.OUTPUT] ?? defaultToken,
+          default: defaultToken,
+        }
+        return update
+      })
+    }
+  }, [defaultToken, inputToken, outputToken, tokens])
```
zzmp, line 55: Note that this introduces a delay (because `useEffect` will run after a render with the deps-array values), so it won't look good for controlled values (like `inputToken`/`outputToken`).

Are these actually controlled? Or are they just here for initialization?

I _think_ they're just here for initialization, in which case could they share a prop with `defaultToken`? eg

    defaultTokens: Currency | { input?: Currency, output?: Currency }

zzmp (changes requested):
I'm confused as to whether `inputToken`/`outputToken` are meant to be controlled in `Swap/index.tsx` - I _think_ they're just for initialization, is that right?

src/components/Widget/inputs.tsx
just-toby: updated to be one single prop

src/components/Widget/index.tsx
```diff
@@ -46,24 +45,26 @@ function useWidgetTheme() {
 }
 
 interface WidgetProps {
-  token?: Currency
+  inputToken?: Currency
```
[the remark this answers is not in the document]
just-toby: renamed

src/components/Widget/inputs.tsx
```diff
@@ -20,30 +21,31 @@ function includesDefaultToken(tokens: SwapTokens) {
...
-  defaultField,
-  onTokenChange,
+  defaultTokens,
+  onDefaultTokenChange,
 }: {
-  token?: Currency
-  defaultField: Field
-
-  onTokenChange?: (token: Currency) => void
+  defaultTokens: DefaultSwapTokens
```
zzmp, line 27: These should match, if possible. Could you just name the type `DefaultTokens`?

zzmp (approved):

PR #5948, opened 2023-02-08
fix: Web 1992 token details update blocked token detail pages

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-1992?atlOrigin=eyJpIjoiM2Q5MGFjOWIwY2UzNGQ3M2IyMTE2NDFjNzE0YjAyNGMiLCJwIjoiaiJ9

### DIFF
--- a/src/components/Tokens/TokenDetails/index.tsx
+++ b/src/components/Tokens/TokenDetails/index.tsx
@@ -153,6 +153,8 @@ export default function TokenDetails({
 
   const [continueSwap, setContinueSwap] = useState<{ resolve: (value: boolean | PromiseLike<boolean>) => void }>()
 
+  const [openTokenSafetyModal, setOpenTokenSafetyModal] = useState(false)
+
   // Show token safety modal if Swap-reviewing a warning token, at all times if the current token is blocked
   const shouldShowSpeedbump = !useIsUserAddedTokenOnChain(address, pageChainId) && tokenWarning !== null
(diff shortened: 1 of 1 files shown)


src/components/Tokens/TokenDetails/index.tsx
```diff
@@ -216,23 +218,27 @@ export default function TokenDetails({
...
         )}
 
-        <RightPanel>
-          <Widget
-            token={token ?? undefined}
-            defaultField={Field.OUTPUT}
-            onTokenChange={navigateToWidgetSelectedToken}
-            onReviewSwapClick={onReviewSwapClick}
-          />
+        <RightPanel onClick={() => isBlockedToken && setOpenTokenSafetyModal(true)}>
```
just-toby, line 221:
```suggestion
        <RightPanel onClick={isBlockedToken ? () => setOpenTokenSafetyModal(true) : null}>
```

? so we don't have an extra callback running unless needed?

just-toby (approved):
lgtm, one suggestion though. also, good candidate for a test ?

src/components/Tokens/TokenDetails/index.tsx
lynnshaoyu: niceeee catch ty

lynnshaoyu: oops seems to lead to error actually / won't build

PR #5955, opened 2023-02-09
fix: use general sort rank instead of only volume rank for row numbers

JFrankfurt:
https://uniswaplabs.atlassian.net/browse/WEB-2855


just-toby:
sorting logic would be a good candidate for a unit test

JFrankfurt:
@just-toby I agree it would be an easy test to write, but nothing on this page is core to a value-transfer ux so I didn't feel it was worth the time. Prioritizing other bug fixes instead.

PR #5958, opened 2023-02-09
build: upgrade ethers@5.7.2, smart-order-router@3.6.0, widgets@2.43.2

zzmp:
Upgrades ethers to the latest v5 minor version. Removes `safeNamehash`, as the issue it addressed was fixed upstream.

This upgrade is necessary before bringing in hardhat as another dependency, to avoid multiple ethers instances in the deps list.


JFrankfurt (approved):

tinaszheng (approved):

zzmp:
Wallet rejections are handled differently in 5.7. Requires testing.

src/components/FeeSelector/shared.tsx
```diff
@@ -1,6 +1,7 @@
 import { Trans } from '@lingui/macro'
 import { FeeAmount } from '@uniswap/v3-sdk'
-import { ALL_SUPPORTED_CHAIN_IDS, SupportedChainId } from 'constants/chains'
+import { SupportedChainId } from 'constants/chains'
+import { ALL_SUPPORTED_CHAIN_IDS } from 'constants/chains'
```
just-toby, line 4: these should be combined right?

src/components/Tokens/TokenDetails/About.tsx
```diff
@@ -1,6 +1,6 @@
 import { Trans } from '@lingui/macro'
-import { SupportedChainId } from '@uniswap/sdk-core'
 import { getChainInfo } from 'constants/chainInfo'
+import { SupportedChainId } from 'constants/chains'
```
just-toby, line 3: so we're converging on the `interface` version of this enum? 

i thought using the updated SDK version of the enum was a goal of the upgrade? 

just checking

src/hooks/useENSAddress.ts
```diff
@@ -11,10 +11,7 @@ import useDebounce from './useDebounce'
  */
 export default function useENSAddress(ensName?: string | null): { loading: boolean; address: string | null } {
   const debouncedName = useDebounce(ensName, 200)
-  const ensNodeArgument = useMemo(
-    () => [debouncedName === null ? undefined : safeNamehash(debouncedName)],
-    [debouncedName]
-  )
+  const ensNodeArgument = useMemo(() => [debouncedName ? namehash(debouncedName) : undefined], [debouncedName])
```
just-toby, line 14: are you intentionally changing the biz logic here? `ensName` can be undefined, and this will behave differently in that case with your change.

src/hooks/useENSContentHash.ts
```diff
@@ -9,7 +9,7 @@ import { useENSRegistrarContract, useENSResolverContract } from './useContract'
  * Does a lookup for an ENS name to find its contenthash.
  */
 export default function useENSContentHash(ensName?: string | null): { loading: boolean; contenthash: string | null } {
-  const ensNodeArgument = useMemo(() => [ensName === null ? undefined : safeNamehash(ensName)], [ensName])
+  const ensNodeArgument = useMemo(() => [ensName ? namehash(ensName) : undefined], [ensName])
```
just-toby, line 12: same here

just-toby (changes requested):
see `useENSAddress`

src/components/Tokens/TokenDetails/About.tsx
vm: we might do that but right now we don't support arbitrum goerli yet and we should have a team discussion on that. the goal is to just remove chains that aren't supported by permit2

src/hooks/useENSAddress.ts
vm: yes this is intentional. the case this was handling before was just related to emojis. really, we shouldn't need to check for both null and undefined but i didn't wanna deal with changing atm

src/hooks/useENSContentHash.ts
vm: same comment

just-toby (approved):

src/components/Tokens/TokenDetails/About.tsx
vm: for now we can't use the sdk-core version bc that would require us to support BNB and aribitrum goerli

PR #5960, opened 2023-02-10
feat: upgrade interface to wc2

grabbou:
Draft PR. This is not going to work, since WC2 package is not yet released (I work with local symlinks).

Early demonstration of what the upgrade looks like.


grabbou:
Closing and will send a follow-up with updated dependencies.

PR #5961, opened 2023-02-10
fix: don't render LP position when token name or symbol includes a url

JFrankfurt:
https://uniswaplabs.atlassian.net/browse/WEB-2802


just-toby (approved):
nice

PR #5965, opened 2023-02-13
feat: widget page-centered dialogs

just-toby:
starts using the `dialog` option of the API to provide a component into which the widget will mount its dialogs. we manage this component, so we can control its position.

some implementation notes:
- we only want to use page-centered modals in Token Detail Page when the widget test is enabled
- we use a custom modal component instead of the existing `Modal` because the Widget API requires a `ref` to an HTML element (into which it will mount the dialog content). this element needs to always exist in the page, we can't mount/unmount it like we do with others.
  - instead, we observe our Dialog element for changes made by the widget and enable/disable it


src/components/Widget/WidgetDialogContainer.tsx
```diff
@@ -0,0 +1,46 @@
...
+  transition: visibility 0.25s linear, opacity 0.25s;
+  visibility: ${({ visible }) => (visible ? 'visible' : 'hidden')};
+
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  width: 100%;
+  align-items: center;
+  justify-content: center;
+  position: absolute;
```
cartcrom, line 18: might need to be `fixed`
image

src/components/Widget/index.tsx
```diff
@@ -169,6 +172,10 @@ export default function Widget({
         locale={locale}
         theme={theme}
         width={width}
+        dialog={dialog}
+        dialogOptions={{
+          animationType: dialog ? DialogAnimationType.FADE : DialogAnimationType.SLIDE,
+        }}
```
cartcrom, line 178: I'm seeing some animation issues here w/ the corners

https://user-images.githubusercontent.com/39385577/219112929-21c909e2-feff-4afc-a3b3-0d058f7a6119.mov

cartcrom (commented):
The code looks good I'm just not sure about having the centered modals in the interface, versus using it as default in the widgets repo. & there could be an input prop for whether to use in-widget vs page-centered dialogs.

I might be missing updated context but I think I remember design noting they'd want these page centered modals to be the norm for implementors of the widget too

cartcrom:
Also sep issue but related to the new UX:

we should really get rid of whatever code causes dialogs to close when the user navigates tabs

https://user-images.githubusercontent.com/39385577/219113901-bdfa1eec-b1f6-4f31-94d4-5eb80eae144f.mov

just-toby:
will probably abandon this in favor of this implementation: https://github.com/Uniswap/widgets/pull/487

PR #5970, opened 2023-02-14
fix: allow signing with smart contract wallets

vm:
Resolves an issue with permit signatures not working with smart contract wallets overloading the signature field.

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,7 +152,7 @@
     "@uniswap/v3-core": "1.0.0",
     "@uniswap/v3-periphery": "^1.1.1",
     "@uniswap/v3-sdk": "^3.9.0",
-    "@uniswap/widgets": "^2.27.0",
+    "@uniswap/widgets": "^2.29.3",
     "@vanilla-extract/css": "^1.7.2",
     "@vanilla-extract/css-utils": "^0.1.2",
     "@vanilla-extract/dynamic": "^2.0.2",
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -5090,7 +5090,7 @@
   resolved "https://registry.yarnpkg.com/@uniswap/token-lists/-/token-lists-1.0.0-beta.30.tgz#2103ca23b8007c59ec71718d34cdc97861c409e5"
   integrity sha512-HwY2VvkQ8lNR6ks5NqQfAtg+4IZqz3KV1T8d2DlI8emIn9uMmaoFbIOg0nzjqAVKKnZSbMTRRtUoAh6mmjRvog==
 
-"@uniswap/universal-router-sdk@1.3.4", "@uniswap/universal-router-sdk@^1.3.4":
+"@uniswap/universal-router-sdk@^1.3.4":
```
vm, line 5093: this seems wrong

package.json
```diff
@@ -145,7 +145,7 @@
     "@uniswap/sdk-core": "^3.0.1",
     "@uniswap/smart-order-router": "^2.10.0",
     "@uniswap/token-lists": "^1.0.0-beta.30",
-    "@uniswap/universal-router-sdk": "1.3.4",
+    "@uniswap/universal-router-sdk": "^1.3.6",
```
zzmp, line 148: Why `^`?

zzmp (approved):
Make sure to also update `widgets`.

package.json
vm: every other package basically uses ^ and i don't see a downside, especially since we have a yarn.lock

PR #5972, opened 2023-02-14
feat: [ListV2] Updated page responsiveness and breakpoints

cbachmeier:
- Updated page responsiveness for showing certain grid columns at certain width, updated margins
- Restyled Back button and added label to match other pages

Responsiveness demo:
pageResponsiveness


src/nft/components/profile/list/MarketplaceRow.tsx
```diff
@@ -267,6 +279,7 @@ export const MarketplaceRow = ({
             globalOverride={globalOverride}
             warning={warning}
             asset={asset}
+            shrink={isMobile}
```
JackShort, line 282: nit: considering this is the only time this component is being used in the codebase why not dynamically change the size of the wrapper in CSS in the component?

JackShort (approved):

src/nft/components/profile/list/MarketplaceRow.tsx
cbachmeier: good call out, updated

PR #5973, opened 2023-02-14
fix: remove gap on tdp bottom bar

just-toby:
https://uniswaplabs.atlassian.net/browse/WEB-2849?atlOrigin=eyJpIjoiNmI1YjIxOGJjMzBkNDkyYmJhNDYwYzYxNzFkZDI2OTAiLCJwIjoiaiJ9

Screenshot 2023-02-14 at 12 07 08 PM


src/components/Tokens/TokenDetails/MobileBalanceSummaryFooter.tsx
```diff
@@ -15,7 +15,7 @@ const Wrapper = styled.div`
   border: 1px solid ${({ theme }) => theme.backgroundOutline};
   background-color: ${({ theme }) => theme.backgroundSurface};
   border-radius: 20px 20px 0px 0px;
-  bottom: 56px;
+  bottom: 50px;
```
JFrankfurt, line 18: I think this value should be 52px. (I'm looking at the bottom element height value.)

As a general rule we try to make sure all px values are divisible by 4.

JFrankfurt (changes requested):
see comment

src/components/Tokens/TokenDetails/MobileBalanceSummaryFooter.tsx
just-toby: good catch, i think the border messed this up. just fixed it

JFrankfurt (approved):

PR #5974, opened 2023-02-14
feat: introduce statsig and expose dummy feature gate

just-toby:
installs the Statsig React SDK and logs an exposure on the Token Detail page for a dummy feature flag - `web_first_feature_flag`

update:
we're now using the Amplitude User/Device ID (exposed by our analytics API) in our statsig initialization. this is to link the statsig exposure logs with the amplitude dataset


src/index.tsx
```diff
@@ -40,6 +43,17 @@ if (isSentryEnabled()) {
...
+const ANALYTICS_DUMMY_KEY = '00000000000000000000000000000000'
+const ANALYTICS_PROXY_URL = process.env.REACT_APP_AMPLITUDE_PROXY_URL
+const COMMIT_HASH = process.env.REACT_APP_GIT_COMMIT_HASH
+initializeAnalytics(ANALYTICS_DUMMY_KEY, OriginApplication.INTERFACE, {
+  proxyUrl: ANALYTICS_PROXY_URL,
+  defaultEventName: SharedEventName.PAGE_VIEWED,
+  commitHash: COMMIT_HASH,
+  isProductionEnv: isProductionEnv(),
+})
+
```
just-toby, line 56: I moved this initialization from `App.tsx` so that the values are availbe to the `StatsigProvider` in this file. We could put `StatsigProvider` at the lower level too, but for some reason that caused issues with e2e tests

src/index.tsx
```diff
@@ -62,23 +76,36 @@ createRoot(container).render(
...
-            </LanguageProvider>
-          </HashRouter>
-        </QueryClientProvider>
+        <StatsigProvider
+          user={{
+            userID: getDeviceId(),
+            customIDs: { session_id: String(getSessionId() ?? -1) },
+          }}
+          // TODO: replace with proxy and cycle key
+          sdkKey="client-1rY92WZGidd2hgW4x1lsZ7afqm1Qfr3sJfH3A5b8eJa"
```
thomasthachil, line 85: this should be stored in an `.env` variable, no?

src/index.tsx
```diff
@@ -62,23 +76,36 @@ createRoot(container).render(
...
-                  </BlockNumberProvider>
-                </ApolloProvider>
-              </Web3Provider>
-            </LanguageProvider>
-          </HashRouter>
-        </QueryClientProvider>
+        <StatsigProvider
+          user={{
+            userID: getDeviceId(),
+            customIDs: { session_id: String(getSessionId() ?? -1) },
```
thomasthachil, line 82: what do `customIDs` represent here?

src/components/Tokens/TokenDetails/MobileBalanceSummaryFooter.tsx
```diff
@@ -86,6 +87,7 @@ export default function MobileBalanceSummaryFooter({ token }: { token: Currency
   const formattedBalance = formatCurrencyAmount(balance, NumberType.TokenNonTx)
   const formattedUsdValue = formatCurrencyAmount(useStablecoinValue(balance), NumberType.FiatTokenStats)
   const chain = CHAIN_ID_TO_BACKEND_NAME[token.chainId].toLowerCase()
+  const { value: enabled } = useGate('web_dummy_gate_amplitude_id')
```
thomasthachil, line 90: I would recommend adding a feature flag constants file where the gate names are added, and they can be referenced from. 

Would provide a good example for future feature flags + see them all in one place. 

something like `featureFlags.ts`:

```
export enum FEATURE_FLAGS {
  DummyFlag = 'web_dummy_gate_amplitude_id',
  DummyFlag2 = `web_dummy_gate_2_amplitude_id`
}
```

thomasthachil, line 90: nit:
instead of `enabled` I prefer more explicitly naming like `isDummyGateFlagEnabled`

src/index.tsx
thomasthachil: I think this moreso depends on what types of flags are being used. If it's anything related to startup / loading, I could see the flag provider being at the outermost level being important. If just for features within, I don't think it's a huge concern

lynnshaoyu: I don't think you're using any of these values in StatsigProvider rn, do we anticipate using them there in future PRs?

thomasthachil (commented):
overall pretty good, just a few comments and questions

src/index.tsx
just-toby: i think they allow you to have a multi-ID for ID-based experiments. i was using this when i thought we wanted both User and Device IDs ... but i think we're only going to be using Device ID. i will remove

src/index.tsx
just-toby: i'm not sure... i figured it's fine for our testrun experiment because we're already planning to cycle the API key when we start proxying. plus it's easier to setup.

src/pages/App.tsx
```diff
@@ -184,124 +192,141 @@ export default function App() {
...
-                <Route path="/" element={<Landing />} />
+        <StatsigProvider
+          user={statsigUser}
+          // TODO: replace with proxy and cycle key
+          sdkKey="client-1rY92WZGidd2hgW4x1lsZ7afqm1Qfr3sJfH3A5b8eJa"
+          waitForInitialization={true}
+          options={{
+            environment: { tier: getEnvName() },
+          }}
+        >
```
just-toby, line 215: ended up moving the provider here because the Amplitude Device ID isn't available at first render

cypress/support/ethereum.ts
```diff
@@ -14,7 +14,7 @@ const TEST_PRIVATE_KEY = '0xe580410d7c37d26c6ad1a837bbae46bc27f9066a466fb3a66e77
 // address of the above key
 const TEST_ADDRESS_NEVER_USE = new Wallet(TEST_PRIVATE_KEY).address
 
-const provider = new JsonRpcProvider('https://goerli.infura.io/v3/4bf032f2d38a4ed6bb975b80d6340847', 4)
+const provider = new JsonRpcProvider('https://goerli.infura.io/v3/4bf032f2d38a4ed6bb975b80d6340847', 5)
```
just-toby, line 17: no idea why this bug only became a problem with my change ... but this should be 5 for goerli!

src/components/Tokens/TokenDetails/MobileBalanceSummaryFooter.tsx
just-toby: suggestion taken

just-toby: suggestion taken

just-toby:
image


^^^ proof that the userId is working as expected now in the exposure logs

src/index.tsx
just-toby: @lynnshaoyu I am, via the `getDeviceId` function of the analytics package. this requires amplitude to be initialized

thomasthachil (approved):

src/pages/App.tsx
```diff
@@ -184,124 +193,143 @@ export default function App() {
...
-          <Polling />
-          <TopLevelModals />
-          <Suspense fallback={<Loader />}>
-            {isLoaded ? (
-              <Routes>
-                <Route path="/" element={<Landing />} />
+        <StatsigProvider
+          user={statsigUser}
+          // TODO: replace with proxy and cycle key
+          sdkKey="client-1rY92WZGidd2hgW4x1lsZ7afqm1Qfr3sJfH3A5b8eJa"
```
thomasthachil, line 213: i would recommend moving this into `process.env` still?

lynnshaoyu (approved):

PR #5975, opened 2023-02-14
fix: use latest version of tokenlists over cache

cartcrom:
fixes bug where new versions of tokenlists weren't being checked into cache

before (Gold1 is supposed to be blocked): 
image

after:
Screen Shot 2023-02-14 at 3 05 34 PM

### DIFF
--- a/src/hooks/useFetchListCallback.ts
+++ b/src/hooks/useFetchListCallback.ts
@@ -9,30 +9,25 @@ import { useAppDispatch } from 'state/hooks'
 
 import { fetchTokenList } from '../state/lists/actions'
 
-export function useFetchListCallback(): (
-  listUrl: string,
-  sendDispatch?: boolean,
-  skipValidation?: boolean
-) => Promise<TokenList> {
+export function useFetchListCallback(): (listUrl: string, skipValidation?: boolean) => Promise<TokenList> {
   const dispatch = useAppDispatch()
 
-  // note: prevent dispatch if using for list search or unsupported list
   return useCallback(
-    async (listUrl: string, sendDispatch = true, skipValidation?: boolean) => {
+    async (listUrl: string, skipValidation?: boolean) => {
       const requestId = nanoid()
-      sendDispatch && dispatch(fetchTokenList.pending({ requestId, url: listUrl }))
+      dispatch(fetchTokenList.pending({ requestId, url: listUrl }))
       return getTokenList(
         listUrl,
         (ensName: string) => resolveENSContentHash(ensName, RPC_PROVIDERS[SupportedChainId.MAINNET]),
         skipValidation
       )
         .then((tokenList) => {
-          sendDispatch && dispatch(fetchTokenList.fulfilled({ url: listUrl, tokenList, requestId }))
+          dispatch(fetchTokenList.fulfilled({ url: listUrl, tokenList, requestId }))
           return tokenList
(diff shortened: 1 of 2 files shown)


src/hooks/useFetchListCallback.ts
```diff
@@ -9,30 +9,25 @@ import { useAppDispatch } from 'state/hooks'
...
 
-export function useFetchListCallback(): (
-  listUrl: string,
-  sendDispatch?: boolean,
-  skipValidation?: boolean
-) => Promise<TokenList> {
+export function useFetchListCallback(): (listUrl: string, skipValidation?: boolean) => Promise<TokenList> {
   const dispatch = useAppDispatch()
 
-  // note: prevent dispatch if using for list search or unsupported list
```
cartcrom, line 19: we now always dispatch, since list search was removed

src/state/lists/updater.ts
```diff
@@ -20,15 +21,15 @@ export default function Updater(): null {
   const fetchList = useFetchListCallback()
   const fetchAllListsCallback = useCallback(() => {
     if (!isWindowVisible) return
-    Object.keys(lists).forEach((url) => {
+    DEFAULT_LIST_OF_LISTS.forEach((url) => {
       // Skip validation on unsupported lists
       const isUnsupportedList = UNSUPPORTED_LIST_URLS.includes(url)
-      fetchList(url, false, isUnsupportedList).catch((error) => console.debug('interval list fetching error', error))
```
zzmp, line 26: This is a change of behavior. Why are you now dispatching these changes?

zzmp (approved):

src/state/lists/updater.ts
```diff
@@ -20,15 +21,15 @@ export default function Updater(): null {
...
+      fetchList(url, isUnsupportedList).catch((error) => console.debug('interval list fetching error', error))
     })
-  }, [fetchList, isWindowVisible, lists])
+  }, [fetchList, isWindowVisible])
 
   // fetch all lists every 10 minutes, but only after we initialize provider
-  useInterval(fetchAllListsCallback, provider ? 1000 * 60 * 10 : null)
+  useInterval(fetchAllListsCallback, provider ? ms`10m` : null)
```
cartcrom, line 32: @zzmp I noticed useInterval will trigger fetchAllListsCallback again if fetchAllListsCallback updates. Is this behavior
 expected? It's new since [WEB-2004](https://github.com/Uniswap/interface/tree/WEB-2004) and was causing a inf looping condition until I refactored to use `DEFAULT_LIST_OF_LISTS` instead of `lists`

It's not an issue here now, was just curious if that's expected or if its a bug in useInterval

zzmp: This _could_ be considered a bug, but not necessarily.

interface's useInterval will retrigger if the callback changes - the callback is expected to be memoized - see https://github.com/Uniswap/interface/blob/main/src/lib/hooks/useInterval.ts
widget's useInterval will never retrigger - see https://github.com/Uniswap/widgets/blob/main/src/hooks/useInterval.ts

I think both can be considered valid, and have different pros/cons. eg widget's will not auto-update if something changes, you'd need to wait the full 10 minutes.

cartcrom: Makes sense, maybe we'll consolidate them eventually

PR #5982, opened 2023-02-15
fix: update subgraph query to actually fetch all tick data

JFrankfurt:
https://uniswaplabs.atlassian.net/browse/WEB-2816

image


just-toby (approved):
maybe leave a comment indicating that we're intentionally fetching all the ticks?

just-toby:
also, it's harder to follow this kind of logic when implemented with hooks. are you doing this on purpose so we display each page of data as it comes in? if so, we should leave as is. if not, maybe consider just making it a single async function that fetches all the pages before returning the result?

PR #5983, opened 2023-02-15
fix: wait for page to load in e2e tests

just-toby:
testing some changes to reduce flakiness in e2e tests

### DIFF
--- a/cypress/e2e/nfts.test.ts
+++ b/cypress/e2e/nfts.test.ts
@@ -6,7 +6,11 @@ const BONSAI_COLLECTION_ADDRESS = '0xec9c519d49856fd2f8133a0741b4dbe002ce211b'
 describe('Testing nfts', () => {
   beforeEach(() => {
     cy.visit('/')
-    cy.get(getTestSelector('FiatOnrampAnnouncement-close')).first().click()
+    cy.find(getTestSelector('FiatOnrampAnnouncement-close')).then((result) => {
+      if (result.length) {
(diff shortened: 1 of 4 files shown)


cypress/e2e/nfts.test.ts
```diff
@@ -5,8 +5,9 @@ const BONSAI_COLLECTION_ADDRESS = '0xec9c519d49856fd2f8133a0741b4dbe002ce211b'
 
 describe('Testing nfts', () => {
   beforeEach(() => {
-    cy.visit('/')
-    cy.get(getTestSelector('FiatOnrampAnnouncement-close')).first().click()
+    cy.visit('/').then(() => {
+      cy.get(getTestSelector('FiatOnrampAnnouncement-close')).first().click()
+    })
```
JFrankfurt, line 10: gah, surprised this is necessary 🙈

src/utils/nativeTokens.ts
```diff
@@ -12,6 +12,7 @@ export function getNativeTokenDBAddress(chain: Chain): string | undefined {
     case Chain.Arbitrum:
     case Chain.EthereumGoerli:
     case Chain.Optimism:
+    default:
```
JFrankfurt, line 15: this fix is on main rn

cypress/e2e/nfts.test.ts
just-toby: i'm not sure if it's necessary in this case but i think it's good practice? especially in the other test, where we want to make sure the page loads before looking for specific elements

JFrankfurt (approved):

PR #5987, opened 2023-02-15
fix: fixing pwat routing edge case

JackShort:


src/nft/utils/txRoute/combineItemsWithTxRoute.ts
```diff
@@ -19,16 +19,19 @@ const isTheSame = (item: GenieAsset, routeAsset: BuyItem | PriceInfo) => {
   }
 }
 
-const isPriceDiff = (oldPrice: string, newPrice: string) => {
-  return formatWeiToDecimal(oldPrice) !== formatWeiToDecimal(newPrice)
+const getPriceDiff = (oldPrice: string, newPrice: string): { hasPriceDiff: boolean; hasVisiblePriceDiff: boolean } => {
```
cbachmeier, line 22: nit: a potentially cleaner alternative is to remove this fn and instead have a visible threshold (ie .0001 or so) and determine behaviour if the absolute price difference is greater or less than the threshold

cbachmeier (approved):
Successfully purchased nftx with shib

src/nft/utils/txRoute/combineItemsWithTxRoute.ts
```diff
@@ -74,27 +77,33 @@ const itemInRouteAndSamePool = (
...
 
-    // if the price changed
     if (route && 'priceInfo' in route.assetOut) {
-      if (isPriceDiff(newPriceInfo.basePrice, route.assetOut.priceInfo.basePrice)) {
+      const { hasPriceDiff, hasVisiblePriceDiff } = getPriceDiff(
+        newPriceInfo.basePrice,
+        route.assetOut.priceInfo.basePrice
+      )
+
+      newPriceInfo = route.assetOut.priceInfo
```
cbachmeier, line 100: nit: don't think it makes a difference but shouldn't this only be reset if `hasPriceDiff`

src/nft/utils/txRoute/combineItemsWithTxRoute.ts
```diff
@@ -74,27 +77,33 @@ const itemInRouteAndSamePool = (
...
-    // if the item is not found in txRoute, it means it's no longer for sale
     if (txRoute && !route) {
       return {
         ...item,
         isUnavailable: true,
       }
     }
 
-    const newPriceInfo = item.updatedPriceInfo ? item.updatedPriceInfo : item.priceInfo
+    let newPriceInfo = item.updatedPriceInfo ? item.updatedPriceInfo : item.priceInfo
```
cbachmeier, line 92: it almost seems like if `route.assetOut.priceInfo` exists, newPriceInfo should be set to it no matter what, otherwise do `item.updatedPriceInfo ? item.updatedPriceInfo : item.priceInfo`

src/nft/components/bag/Bag.tsx
```diff
@@ -66,7 +66,7 @@ const BagContainer = styled.div<{ raiseZIndex: boolean; isProfilePage: boolean }
   border-radius: 16px;
   box-shadow: ${({ theme }) => theme.shallowShadow};
   z-index: ${({ raiseZIndex, isProfilePage }) =>
-    raiseZIndex ? (isProfilePage ? Z_INDEX.modalOverTooltip : Z_INDEX.modalBackdrop) : 3};
+    raiseZIndex ? (isProfilePage ? Z_INDEX.modalOverTooltip : Z_INDEX.modalBackdrop + 2) : 3};
```
cbachmeier, line 69: q: Do we need the `raiseZIndex` check? It seems like we always want the zindex either `modalOverTooltip` or `modalBackdrop`

JackShort: yeah we actually need the `raiseZIndex` check because otherwise the wallet dropdown will go behind the bag

src/nft/utils/txRoute/combineItemsWithTxRoute.ts
JackShort: well the reason this exists is because product thought it was confusing when users were getting displayed a price change but visually it was the same prices because of our rounding style. So the visible threshold is actually dependent on how we render the number.

src/nft/utils/txRoute/combineItemsWithTxRoute.ts
JackShort: yes

src/nft/utils/txRoute/combineItemsWithTxRoute.ts
JackShort: that is actually what is happening and regarding the above comment that is why I didn't do it based on if hasPrice diff because if `route.assetOut.priceinfo` exists `newPriceInfo` will be set to it

src/nft/utils/txRoute/combineItemsWithTxRoute.ts
JackShort: see below basically I think `newPriceInfo` should always be set to `route.assetOut.priceInfo` if it exists as you said. We probably don't need to set `newPriceInfo` initially however just in case some data falls through or something it can't really hurt in my eyes

PR #5989, opened 2023-02-16
fix: Market icons displaying over dropdown

cbachmeier:
The table headers Z-Index was below the multiple marketplace icons layering which causes them to display over the set price by dropdown. This fixes that bug. There was also a bug where the remove marketplace on an expanded row just removed the first marketplace and not the one clicked.

After
Screen Shot 2023-02-16 at 11 57 13
Before
Screen Shot 2023-02-16 at 10 53 00


JackShort (approved):
lgtm

PR #5991, opened 2023-02-16
feat: pwat amp events

JackShort:


src/nft/hooks/useSendTransaction.ts
```diff
@@ -63,13 +64,15 @@ export const useSendTransaction = create<TxState>()(
...
               txReceipt,
+              purchasedWithErc20,
             }
           } else {
             set({ state: TxStateType.Failed })
             return {
               nftsPurchased: [],
               nftsNotPurchased: selectedAssets,
               txReceipt,
+              purchasedWithErc20,
```
cbachmeier, line 75: It doesn't seem like this property is ever actually ingested in an event. Additionally is there a way to parse that a tx used erc20 from the route response or receipt so this boolean doesn't have to be passed through 3 functions? Alternatively could we set `purchasedWithErc20` or some equivalent field in the `useSendTransaction` hook?

src/nft/components/bag/BagFooter.tsx
```diff
@@ -531,7 +549,14 @@ export const BagFooter = ({ totalEthPrice, fetchAssets, eventProperties }: BagFo
                 <ThemedText.SubHeaderSmall>
                   <Trans>Pay with</Trans>
                 </ThemedText.SubHeaderSmall>
-                <CurrencyInput onClick={() => (bagIsLocked ? undefined : setTokenSelectorOpen(true))}>
+                <CurrencyInput
+                  onClick={() => {
+                    if (bagIsLocked) return
```
cbachmeier, line 554: nit: could you do `if(!bagIsLocked) { doThings()... }` with no return?

src/nft/components/bag/BagFooter.tsx
```diff
@@ -518,9 +518,27 @@ export const BagFooter = ({ totalEthPrice, fetchAssets, eventProperties }: BagFo
...
+      return undefined
+    }
+
+    return {
+      token_address: inputCurrency.address,
+      token_symbol: inputCurrency.symbol,
+    }
+  }, [inputCurrency])
+
+  useEffect(() => {
```
cbachmeier, line 536: Instead of this useEffect could we trigger this event manually via the `onCurrencySelect` function in the `CurrencySearchModal`?

src/nft/components/bag/BagFooter.tsx
JackShort: ok but hear me out 0 levels of indentation vs 1

cbachmeier: only if we change from spaces to tabs

cbachmeier (approved):
LGTM

PR #5992, opened 2023-02-17
fix: fix token link

chenxsan:
Say I'm on this USDT token page https://app.uniswap.org/#/tokens/arbitrum/0xfd086bc7cd5c481dcc9c85ebe478a1c0b69fcbb9

`More analytics` button links to `https://info.uniswap.org/#/tokens/0xFd086bC7CD5C481DCC9C85ebE478A1C0b69FCbb9` which is wrong since it doesn't exist under mainnet.

CleanShot 2023-02-17 at 10 46 30@2x

This pull request would include the network label in url which would update the `More analytics` link to `https://info.uniswap.org/#/arbitrum/tokens/0xFd086bC7CD5C481DCC9C85ebE478A1C0b69FCbb9`.

### DIFF
--- a/src/components/Tokens/TokenDetails/About.tsx
+++ b/src/components/Tokens/TokenDetails/About.tsx
@@ -82,7 +82,7 @@ export function AboutSection({ address, chainId, description, homepageUrl, twitt
 
   const baseExplorerUrl = getChainInfo(chainId).explorer
 
-  const { label } = getChainInfo(isSupportedChain(chainId) ? chainId : SupportedChainId.MAINNET)
+  const { infoLink } = getChainInfo(isSupportedChain(chainId) ? chainId : SupportedChainId.MAINNET)
 
   return (
     <AboutContainer data-testid="token-details-about-section">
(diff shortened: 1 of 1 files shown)


src/components/Tokens/TokenDetails/About.tsx
```diff
@@ -108,7 +111,12 @@ export function AboutSection({ address, chainId, description, homepageUrl, twitt
           name={chainId === SupportedChainId.MAINNET ? 'Etherscan' : 'Block Explorer'}
           link={`${baseExplorerUrl}${address === 'NATIVE' ? '' : 'address/' + address}`}
         />
-        <Resource name="More analytics" link={`https://info.uniswap.org/#/tokens/${address}`} />
+        <Resource
+          name="More analytics"
+          link={`https://info.uniswap.org/#/${
```
tinaszheng, line 116: this isn't going to work since some `labels` are not valid strings to be used for the URL here, like `Celo Alfajores` or `BNB Chain` 

`getChainInfo` returns `infoLink` for the prefix you want - can you just use that here instead?

tinaszheng (changes requested):

chenxsan:
Seems it's already fixed in https://github.com/Uniswap/interface/pull/6216/files#diff-bf9dfbe910f6b75e43cb3d910ea280ab8a48508f074690b188d4e76a3ab9cb57R111

PR #5993, opened 2023-02-17
fix: darkmode FoR iframe wrapper background color

JFrankfurt:
image

### DIFF
--- a/src/components/FiatOnrampModal/index.tsx
+++ b/src/components/FiatOnrampModal/index.tsx
@@ -10,8 +10,9 @@ import { CustomLightSpinner, ThemedText } from 'theme'
 import Circle from '../../assets/images/blue-loader.svg'
 import Modal from '../Modal'
 
-const Wrapper = styled.div`
-  background-color: ${({ theme }) => theme.white};
+const Wrapper = styled.div<{ isDarkMode: boolean }>`
+  // #1c1c1e is the background color for the darkmode moonpay iframe as of 2/16/2023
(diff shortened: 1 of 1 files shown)


src/components/FiatOnrampModal/index.tsx
```diff
@@ -10,8 +10,9 @@ import { CustomLightSpinner, ThemedText } from 'theme'
 import Circle from '../../assets/images/blue-loader.svg'
 import Modal from '../Modal'
 
-const Wrapper = styled.div`
-  background-color: ${({ theme }) => theme.white};
+const Wrapper = styled.div<{ isDarkMode: boolean }>`
+  // #1c1c1e is the background color for the darkmode moonpay iframe as of 2/16/2023
+  background-color: ${({ isDarkMode, theme }) => (isDarkMode ? '#1c1c1e' : theme.white)};
```
cbachmeier, line 15: we should avoid using the hardcoded hexes when possible. Should we instead add `#1c1c1e` and white as a theme color such as theme.backgroundMoonpay?

JFrankfurt: ehhh, not sure. I don't think this belongs on the theme, since it definitely shouldn't be used anywhere other than this iframe wrapper.

vm: can you make the hex at least a var at the top of the file?

vm (approved):

PR #5996, opened 2023-02-17
fix: update user rejected txn errors to stay up to date with widget

lynnshaoyu:
keep up to date with widget

context: ethers upgrade to 5.7.0 changed error code when user rejects txn, so we need to handle the extra edge case.

### DIFF
--- a/src/utils/swapErrorToUserReadableMessage.tsx
+++ b/src/utils/swapErrorToUserReadableMessage.tsx
@@ -25,6 +25,8 @@ export function swapErrorToUserReadableMessage(error: any): string {
   // The 4001 error code doesn't capture the case where users reject a transaction for all wallets,
   // so we need to parse the reason for these special cases:
   if (
+    // ethers v5.7.0 wrapped error
+    error?.code === 'ACTION_REJECTED' ||
     // For Rainbow :
     (reason?.match(/request/i) && reason?.match(/reject/i)) ||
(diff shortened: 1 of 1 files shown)


src/utils/swapErrorToUserReadableMessage.tsx
```diff
@@ -25,14 +25,18 @@ export function swapErrorToUserReadableMessage(error: any): string {
   // The 4001 error code doesn't capture the case where users reject a transaction for all wallets,
   // so we need to parse the reason for these special cases:
   if (
+    // ethers v5.7.0 wrapped error
+    error?.code === 'ACTION_REJECTED' ||
```
tinaszheng, line 29: you could import this from ethers directly :D its `Logger.errors.ACTION_REJECTED` sorry i didnt have it on my branch!

zzmp (approved):

src/utils/swapErrorToUserReadableMessage.tsx
lynnshaoyu: i think we need the ethers upgrade to 5.7.0 first, not seeing it ACTION_REJECTED in Logger.errors right now :O

tinaszheng: very true

tinaszheng (approved):

PR #5997, opened 2023-02-17
fix: not using stablecoin to calc price impact pwat

JackShort:
`usePriceImpact` was taken from the widgets codebase to create parity between pwat and the widget


src/nft/hooks/usePriceImpact.ts
```diff
@@ -0,0 +1,46 @@
+import { Currency, Percent, TradeType } from '@uniswap/sdk-core'
+import { useMemo } from 'react'
+import { InterfaceTrade } from 'state/routing/types'
+import { useTheme } from 'styled-components/macro'
+import { computeRealizedPriceImpact, getPriceImpactWarning } from 'utils/prices'
+
+export interface PriceImpact {
+  percent: Percent
+  warning?: 'warning' | 'error'
```
cbachmeier, line 9: nit: would rename `severity` or `warningType`

src/nft/hooks/usePriceImpact.ts
```diff
@@ -0,0 +1,46 @@
+import { Currency, Percent, TradeType } from '@uniswap/sdk-core'
+import { useMemo } from 'react'
+import { InterfaceTrade } from 'state/routing/types'
+import { useTheme } from 'styled-components/macro'
+import { computeRealizedPriceImpact, getPriceImpactWarning } from 'utils/prices'
+
+export interface PriceImpact {
+  percent: Percent
```
cbachmeier, line 8: I don't believe this percent field is ever used

src/nft/hooks/usePriceImpact.ts
```diff
@@ -0,0 +1,46 @@
...
+  warning?: 'warning' | 'error'
+  toString(): string
```
cbachmeier, line 10: instead of overwriting the built in toString I would rename this `displayPercentage`. When I first saw `toString` I assumed it would also display all params such as the warning level and color

src/nft/hooks/usePriceImpact.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  return useMemo(() => {
+    const marketPriceImpact = trade ? computeRealizedPriceImpact(trade) : undefined
+    const priceImpactWarning = marketPriceImpact ? getPriceImpactWarning(marketPriceImpact) : undefined
+    const warningColor =
+      priceImpactWarning === 'error'
+        ? theme.accentCritical
+        : priceImpactWarning === 'warning'
+        ? theme.accentWarning
+        : undefined
```
cbachmeier, line 25: nit: the warning and warning color are inherently tied, you could define a PriceImpactWarning interface with fields: `type` and `color` which is returned from your `getPriceImpactWarning` helper function.

src/nft/hooks/usePriceImpact.ts
```diff
@@ -0,0 +1,46 @@
...
+          percent: marketPriceImpact,
+          warning: priceImpactWarning,
+          toString: () => toHumanReadablePercent(marketPriceImpact),
+          warningColor,
+        }
+      : undefined
+  }, [theme.accentCritical, theme.accentWarning, trade])
+}
+
+function toHumanReadablePercent(priceImpact: Percent): string {
```
cbachmeier, line 38: nit: should move this to src/utils/price with the priceImpactWarning helper fn

cbachmeier (approved):
Works well. Minor nits and suggestions

PR #6006, opened 2023-02-21
fix: use ipfs token list temporarily

just-toby:
changes our default token list to be the IPFS hosted one, while we investigate a cloudflare issue


vm (approved):
I would rename the description to not say fallback as it's just temporarily switching the URL

lynnshaoyu (approved):

PR #6009, opened 2023-02-21
chore: remove listv2 feature flag

cbachmeier:
Removes NFTListV2 Feature flag. Any dependent logic has been removed as well as the tooltip warnings which are not used in V2. Added a couple translations where previously missing. Saving refactoring listing state logic for subsequent PR(s).


src/nft/components/bag/profile/ListingButton.tsx
```diff
@@ -229,49 +216,6 @@ export const ListingButton = ({ onClick, buttonText, showWarningOverride = false
   return (
     <>
       <Box position="relative">
-        {!showWarningOverride && showWarning && warningMessage.length > 0 && (
```
cbachmeier, line 232: not used in V2

PR #6011, opened 2023-02-22
feat: add statsig feature gate for widget

just-toby:
enables the swap widget replacement when the corresponding statsig flag is `true`. with this approach, we'll be keeping the legacy feature flag (so we can force-enable the feature in dev builds), but we'll use the statsig gate to roll the feature out on prod

### DIFF
--- a/src/featureFlags/flags/dummyFeatureGate.ts
+++ b/src/featureFlags/flags/dummyFeatureGate.ts
@@ -1,7 +1,7 @@
 import { BaseVariant, FeatureFlag, useBaseFlag } from '../index'
 
 function useDummyGateFlag(): BaseVariant {
-  return useBaseFlag(FeatureFlag.statsigDummy, BaseVariant.Control)
+  return useBaseFlag(FeatureFlag.statsigDummy)
 }
 
 export function useDummyGateEnabled(): boolean {
--- a/src/featureFlags/flags/dummyFeatureGate.ts
+++ b/src/featureFlags/flags/dummyFeatureGate.ts
@@ -1,0 +1,9 @@
+import { BaseVariant, FeatureFlag, useBaseFlag } from '../index'
+
+function useDummyGateFlag(): BaseVariant {
+  return useBaseFlag(FeatureFlag.statsigDummy, BaseVariant.Control)
+}
+
+export function useDummyGateEnabled(): boolean {
+  return useDummyGateFlag() === BaseVariant.Enabled
+}
--- a/src/featureFlags/flags/featureFlags.ts
+++ b/src/featureFlags/flags/featureFlags.ts
@@ -1,13 +1,12 @@
+/**
+ * The value here should match the value in the statsig dashboard, if you plan to use statsig.
+ */
 export enum FeatureFlag {
   traceJsonRpc = 'traceJsonRpc',
   permit2 = 'permit2',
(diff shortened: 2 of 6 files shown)


src/pages/Swap/index.tsx
```diff
@@ -159,7 +161,8 @@ export default function Swap({ className }: { className?: string }) {
   const loadedUrlParams = useDefaultsFromURLSearch()
   const [newSwapQuoteNeedsLogging, setNewSwapQuoteNeedsLogging] = useState(true)
   const [fetchingSwapQuoteStartTime, setFetchingSwapQuoteStartTime] = useState<Date | undefined>()
-  const swapWidgetEnabled = useSwapWidgetEnabled()
+  const swapWidgetManualOverrideEnabled = useSwapWidgetEnabled()
```
zzmp, line 164: I would prefer to see statsig integrated into the featureFlags/ logic, so that manual overrides don't need special-cased logic going forward.

The base hook (for featureFlag) could delegate to statSig, and then _this_ file would just call to `useSwapWidgetEnabled`.

zzmp (changes requested):

src/pages/Swap/index.tsx
just-toby: thoughts on the update?

src/featureFlags/flags/swapWidget.ts
```diff
@@ -1,11 +1,14 @@
...
+
 import { BaseVariant, FeatureFlag, useBaseFlag } from '../index'
 
 export function useSwapWidgetFlag(): BaseVariant {
   return useBaseFlag(FeatureFlag.swapWidget, BaseVariant.Control)
 }
 
 export function useSwapWidgetEnabled(): boolean {
-  return useSwapWidgetFlag() === BaseVariant.Enabled
+  const { value: statsigValue } = useGate(FeatureFlag.swapWidget)
```
zzmp, line 10: Can this be moved into `useBaseFlag`, which `useBaseFlag` taking an optional parameter with the gate name?

zzmp (changes requested):

src/featureFlags/flags/dummyFeatureGate.ts
```diff
@@ -0,0 +1,9 @@
+import { BaseVariant, FeatureFlag, useBaseFlag } from '../index'
+
+function useDummyGateFlag(): BaseVariant {
+  return useBaseFlag(FeatureFlag.statsigDummy, BaseVariant.Control)
```
zzmp, line 4: nit: you do not need to pass BaseVariant.Control as a default, as this _is_ the default.

```suggestion
  return useBaseFlag(FeatureFlag.statsigDummy)
```

src/featureFlags/flags/featureFlags.ts
```diff
@@ -1,12 +1,12 @@
+/**
+ * The value here should match the value in the statsig dashboard, if you plan to use statsig.
```
zzmp, line 2: nit: s/should/must

```suggestion
 * The value here must match the value in the statsig dashboard, if you plan to use statsig.
```

zzmp (approved):
I love that this adds statsig for widget without even touching the widget code - just by improving infra for the future ⏳ 🥇

PR #6015, opened 2023-02-22
feat: implement Sentry for state tracking

grabbou:
As per design doc, this PR implements Sentry Redux enhancer with an explicit list of whitelisted keys to track the data we need and to avoid any payload changes in case the underlying state shape changes in the future.

Here's a preview of what it looks like, when enabled:

Screenshot 2023-03-08 at 01 04 34

### DIFF
--- a/src/state/hooks.ts
+++ b/src/state/hooks.ts
@@ -1,7 +1,6 @@
 import { TypedUseSelectorHook, useDispatch, useSelector } from 'react-redux'
 
 import store from './index'
-import { AppState } from './types'
 
 export const useAppDispatch = () => useDispatch<typeof store.dispatch>()
-export const useAppSelector: TypedUseSelectorHook<AppState> = useSelector
+export const useAppSelector: TypedUseSelectorHook<ReturnType<typeof store.getState>> = useSelector
--- a/src/state/hooks.ts
+++ b/src/state/hooks.ts
@@ -1,5 +1,7 @@
 import { TypedUseSelectorHook, useDispatch, useSelector } from 'react-redux'
-import { AppDispatch, AppState } from 'state'
 
-export const useAppDispatch = () => useDispatch<AppDispatch>()
+import store from './index'
+import { AppState } from './types'
+
+export const useAppDispatch = () => useDispatch<typeof store.dispatch>()
 export const useAppSelector: TypedUseSelectorHook<AppState> = useSelector
--- a/src/state/index.ts
+++ b/src/state/index.ts
@@ -1,61 +1,14 @@
-import { configureStore, Reducer } from '@reduxjs/toolkit'
+import { configureStore } from '@reduxjs/toolkit'
 import { setupListeners } from '@reduxjs/toolkit/query/react'
-import * as Sentry from '@sentry/react'
-import multicall from 'lib/state/multicall'
 import { load, save } from 'redux-localstorage-simple'
 import { isTestEnv } from 'utils/env'
 
-import application from './application/reducer'
-import burn from './burn/reducer'
-import burnV3 from './burn/v3/reducer'
-import connection from './connection/reducer'
 import { updateVersion } from './global/actions'
-import lists from './lists/reducer'
-import logs from './logs/slice'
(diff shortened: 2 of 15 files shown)


src/state/index.ts
```diff
@@ -21,27 +22,87 @@ import wallets from './wallets/reducer'
...
+type GetState<T> = T extends Reducer<infer State> ? State : never
+
+/* Utility type to mark all properties of a type as optional */
+type DeepPartial<T> = T extends object
+  ? {
+      [P in keyof T]?: DeepPartial<T[P]>
+    }
+  : T
+
+/* eslint-disable-next-line import/no-unused-modules -- reports false positive error, we use that type in other modules */
```
zzmp, line 52: This was probably a local-only error - can you remove the eslint-disable's unless the PR actually fails?

src/state/index.ts
```diff
@@ -21,27 +22,87 @@ import wallets from './wallets/reducer'
...
   middleware: (getDefaultMiddleware) =>
     getDefaultMiddleware({ thunk: true })
       .concat(routingApi.middleware)
-      .concat(save({ states: PERSISTED_KEYS, debounce: 1000 })),
+      .concat(save({ states: PERSISTED_KEYS, debounce: 1000 }))
+      .concat(
+        Sentry.createReduxEnhancer({
+          actionTransformer: () => null,
+          /**
+           * This function runs on every state update, so keeping it as fast as possible by avoiding any function
```
zzmp, line 70: This comment should also explain _why_ this is being done.

grabbou: Added some more comments

src/state/hooks.ts
```diff
@@ -1,5 +1,7 @@
 import { TypedUseSelectorHook, useDispatch, useSelector } from 'react-redux'
-import { AppDispatch, AppState } from 'state'
 
-export const useAppDispatch = () => useDispatch<AppDispatch>()
+import store from './index'
+import { AppState } from './types'
+
+export const useAppDispatch = () => useDispatch<typeof store.dispatch>()
 export const useAppSelector: TypedUseSelectorHook<AppState> = useSelector
```
zzmp, line 7: nit: this could be `TypedUseSelectorHook<ReturnType<store.getState>>` to avoid bringing in `./types`.

src/state/logging.ts
```diff
@@ -0,0 +1,57 @@
...
+
+import { AppState } from './types'
+
+/* Utility type to mark all properties of a type as optional */
+type Partial<T> = T extends object
+  ? {
+      [P in keyof T]?: Partial<T[P]>
+    }
+  : T
+
```
zzmp, line 11: Call this `DeepPartial`, as `Partial` is already a utility type.
```suggestion
type DeepPartial<T> = T extends object
  ? {
      [P in keyof T]?: DeepPartial<T[P]>
    }
  : T

```

src/state/logging.ts
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * We don't want to store actions as breadcrumbs in Sentry, so we return null to disable the default behavior.
+   */
+  actionTransformer: () => null,
+  /**
+   * We only want to store a subset of the state in Sentry, containing only the relevant parts for debugging.
+   * Note: This function runs on every state update, so we're keeping it as fast as possible by avoiding any function
+   * calls and deep object traversals.
+   */
+  stateTransformer: (state: AppState): Partial<AppState> => {
```
zzmp, line 26:
```suggestion
  stateTransformer: (state: AppState): DeepPartial<AppState> => {
```

src/state/logging.ts
```diff
@@ -0,0 +1,57 @@
...
+    return {
```
zzmp, line 27: nit: I think this may be more readable if you destructure it first, because it'll keep things on a single line.

zzmp (approved):

src/state/logging.ts
grabbou: Haha, had it like this before actually!

zzmp (approved):

PR #6017, opened 2023-02-22
fix: token selector correct zIndex on mobile for pwat

JackShort:
the z-index needs to decrease and not increase to show correctly on mobile


cbachmeier (approved):
LGTM

PR #6019, opened 2023-02-22
fix: token list load issue WEB-2083

JFrankfurt:
https://uniswaplabs.atlassian.net/browse/WEB-2877?atlOrigin=eyJpIjoiYjFkNDYxNWIyMWJhNDAwMzkxMWFhOWMzNTY5MGIyZWEiLCJwIjoiaiJ9

### DIFF
--- a/src/utils/nativeTokens.ts
+++ b/src/utils/nativeTokens.ts
@@ -4,6 +4,9 @@ import { CHAIN_NAME_TO_CHAIN_ID } from 'graphql/data/util'
 
 export function getNativeTokenDBAddress(chain: Chain): string | undefined {
   const pageChainId = CHAIN_NAME_TO_CHAIN_ID[chain]
+  if (pageChainId === undefined) {
+    return undefined
+  }
   switch (chain) {
     case Chain.Celo:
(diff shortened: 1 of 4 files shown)


src/utils/nativeTokens.ts
```diff
@@ -4,6 +4,9 @@ import { CHAIN_NAME_TO_CHAIN_ID } from 'graphql/data/util'
 
 export function getNativeTokenDBAddress(chain: Chain): string | undefined {
   const pageChainId = CHAIN_NAME_TO_CHAIN_ID[chain]
+  if (pageChainId === undefined) {
+    return undefined
+  }
```
just-toby, line 9: why is this not caught by the default case below? is this related to the token lists changes?

just-toby (approved):
list logic changes LGTM

src/utils/nativeTokens.ts
JFrankfurt: this actually is, but I was working w/ the tsconfig suggestion @tinaszheng made in the linked slack thread and it flagged this case too, so why not?

PR #6026, opened 2023-02-24
fix: reset tokens, amounts when switching chains

vm:
This makes sure that you can't get to a state where the selected tokens are disabled.

### DIFF
--- a/src/components/Widget/inputs.tsx
+++ b/src/components/Widget/inputs.tsx
@@ -1,7 +1,9 @@
 import { sendAnalyticsEvent, useTrace } from '@uniswap/analytics'
 import { InterfaceSectionName, SwapEventName } from '@uniswap/analytics-events'
 import { Currency, Field, SwapController, SwapEventHandlers, TradeType } from '@uniswap/widgets'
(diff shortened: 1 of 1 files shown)


src/components/Widget/inputs.tsx
```diff
@@ -29,24 +31,35 @@ export function useSyncWidgetInputs({
...
+  useEffect(() => {
+    if (chainId !== previousChainId && !!previousChainId) {
+      setTokens({
+        ...tokens,
+        [Field.INPUT]: undefined,
+        [Field.OUTPUT]: undefined,
+      })
+      setAmount(EMPTY_AMOUNT)
+    }
+  }, [chainId, previousChainId, tokens])
```
just-toby, line 61: should we reset to the interface-defined default instead? what if the `inputCurrency` URL param is set?

vm: my issue with this is then we need to handle resetting the state differently per route. which doesn't align with how the sync hook works right now. i think we can ignore the url param in this case since it will be invalid in the new network that's switched to

just-toby (approved):

PR #6027, opened 2023-02-24
chore: refactoring bag component to sub components

JackShort:


cbachmeier:
Encountered a few bugs:
When  there's an update that all items are unavailable, instead of showing the update in the bag all items get removed and the bag closes.
payCloseAndRemove

When I reject the tx in my wallet, all items get removed from my bag and it closes.
rejectRemoves

src/nft/components/bag/Bag.tsx
```diff
@@ -1,9 +1,5 @@
-import { BigNumber } from '@ethersproject/bignumber'
 import { sendAnalyticsEvent } from '@uniswap/analytics'
```
cbachmeier, line 1: Great job, this file is looking very clean

src/nft/hooks/useFetchAssets.ts
```diff
@@ -0,0 +1,238 @@
...
+        provider?.getSigner(),
+        itemsInBag.filter((item) => item.status !== BagItemStatus.UNAVAILABLE).map((item) => item.asset),
+        routingData,
+        purchasingWithErc20
+      )
+
+      setBagLocked(false)
+      setTransactionResponse(purchaseResponse)
+      setBagExpanded({ bagExpanded: false })
+      resetBag()
```
cbachmeier, line 83: there was an if statement around this logic previously and I wonder if it's related to the two bugs I posted:
```
    if (
      purchaseResponse &&
      (transactionStateRef.current === TxStateType.Success || transactionStateRef.current === TxStateType.Failed)
    ) {

File: src/nft/hooks/useSubscribeTransactionState.ts [hunk 3]
```diff
@@ -0,0 +1,36 @@
...
+      setLocked,
+    }),
+    shallow
+  )
+
+  useEffect(() => {
+    useSendTransaction.subscribe((state) => (transactionStateRef.current = state.state))
+  }, [])
+
+  useEffect(() => {
```
cbachmeier, line 25: Not for now but when the major refactoring goes down, it may be worth exploring if the useEffects can be removed and bagStatus is directly updated only from the hooks in useFetchAssets or from direct user interaction

JackShort: yeah I was thinking about that

src/nft/hooks/useFetchAssets.ts
JackShort: yeah it was

src/nft/hooks/useFetchAssets.ts
```diff
@@ -0,0 +1,238 @@
...
+    },
+    [itemsInBag, provider, resetBag, sendTransaction, setBagExpanded, setBagLocked, setTransactionResponse]
+  )
+
+  return useCallback(async () => {
```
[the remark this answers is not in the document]
JackShort: that was el plan

PR #6028, opened 2023-02-24
fix: start with an empty user, then update statsig user when users connect a wallet

JFrankfurt:


just-toby (changes requested):
we're discussing this in chat, but just RC'ing to remove from my queue in the meantime. this breaks the user ID in the statsig exposure logs

PR #6033, opened 2023-02-27
chore: schema changed gen types

JFrankfurt:
looks like the schema changed and so the generated types need to change to match


cbachmeier (approved):
LGTM

PR #6036, opened 2023-02-27
feat: Add BNB Network support

JFrankfurt:
Shout out to @ilyamk for doing most of the work here.


src/constants/networks.ts
```diff
@@ -15,7 +15,7 @@ if (typeof INFURA_KEY === 'undefined') {
  *
  * These "Safe" URLs are listed first, followed by other fallback URLs, which are taken from chainlist.org.
  */
-export const FALLBACK_URLS: { [key in SupportedChainId]: string[] } = {
+export const FALLBACK_URLS = {
```
just-toby, line 18: why not keep the typing here?

just-toby (approved):
the `tokens` file is getting pretty big. i wonder if we should split it up into one file per-chain... (as a backlog item)

src/constants/networks.ts
JFrankfurt: it keeps the same resulting type via inference, but it doesn't require exhaustive enumeration of the enum in the declaration

src/components/AccountDetailsV2/index.tsx
```diff
@@ -1,6 +1,6 @@
+import { SupportedChainId } from '@uniswap/sdk-core'
```
tinaszheng, line 1: hmm i think i would still prefer us to define our own `SupportedChainId` on the interface rather than strictly taking from sdk-core

i can see a scenario in which sdk-core starts supporting a chain id that we don't - for example, adding a testnet, and then that breaks some of our typescript handling 

but not too picky about it

src/constants/addresses.ts
```diff
@@ -22,6 +22,15 @@ const CELO_QUOTER_ADDRESSES = '0x82825d0554fA07f7FC52Ab63c961F330fdEFa8E8'
 const CELO_NONFUNGIBLE_POSITION_MANAGER_ADDRESSES = '0x3d79EdAaBC0EaB6F08ED885C05Fc0B014290D95A'
 const CELO_TICK_LENS_ADDRESSES = '0x5f115D9113F88e0a0Db1b5033D90D4a9690AcD3D'
 
+// bsc v3 addresses
+const BSC_V3_CORE_FACTORY_ADDRESSES = '0x128Ce3A3D48f27CE35A3F810cF2cddD2f6879b13'
+const BSC_ROUTER_ADDRESS = '0xa64Dd3A63fEed608660226112E748F93B87E650F'
```
tinaszheng, line 27: is this SwapRouter02? do we need this since we're on Universal Router? or just putting here cuz we haven't deprecated non-permit2 flow?

src/constants/networks.ts
tinaszheng: 🤔 dont understand dis

src/graphql/thegraph/apollo.ts
```diff
@@ -14,6 +14,8 @@ const CHAIN_SUBGRAPH_URL: Record<number, string> = {
   [SupportedChainId.POLYGON]: 'https://api.thegraph.com/subgraphs/name/ianlapham/uniswap-v3-polygon',
 
   [SupportedChainId.CELO]: 'https://api.thegraph.com/subgraphs/name/jesse-sawa/uniswap-celo',
+
+  [SupportedChainId.BSC]: 'https://api.thegraph.com/hosted-service/subgraph/ilyamk/uniswap-v3---bnb-chain',
```
tinaszheng, line 18: this link probably subject to change 👀 will let u kno

src/utils/getExplorerLink.ts
```diff
@@ -12,6 +12,7 @@ const BLOCK_EXPLORER_PREFIXES: { [chainId: number]: string } = {
   [SupportedChainId.POLYGON_MUMBAI]: 'https://mumbai.polygonscan.com',
   [SupportedChainId.CELO]: 'https://celoscan.io',
   [SupportedChainId.CELO_ALFAJORES]: 'https://alfajores-blockscout.celo-testnet.org',
+  [SupportedChainId.BSC]: 'https://bscscan.com/',
```
tinaszheng, line 15: nit 
```suggestion
  [SupportedChainId.BSC]: 'https://bscscan.com',
```

for the consistency lulz

src/components/AccountDetailsV2/index.tsx
tinaszheng: for example, i think in this scenario you wouldn't have had to add arbitrum goerli to also meet ts lint checking, but not a big deal

tinaszheng (approved):

src/constants/addresses.ts
JFrankfurt: @ilyamk

src/components/AccountDetailsV2/index.tsx
cartcrom: I'm also curious abt why we're switching away from defining our own SupportedChainId, as it changes the meaning of something like `isSupportedChain` from _interface supports this chain_ => _the sdk supports this chain_

JFrankfurt: I think about it as *Uniswap Labs* supports X chain, which is why I think a canonical sdk chain definition makes sense.

src/theme/components/RadialGradientByChainUpdater.ts
```diff
@@ -87,6 +87,15 @@ export default function RadialGradientByChainUpdater(): null {
         backgroundRadialGradientElement.style.background = darkMode ? celoDarkGradient : celoLightGradient
         break
       }
+      case SupportedChainId.BSC: {
+        setBackground(backgroundResetStyles)
+        const bscLightGradient =
+          'radial-gradient(100% 100% at 50% 0%, rgba(240, 185, 11, 0.7) 0%, rgba(255, 219, 103, 0.6536) 49.48%, rgba(255, 255, 255, 0) 100%), #FFFFFF'
```
lynnshaoyu, line 93: colors r being set here :D @infredible

src/graphql/thegraph/apollo.ts
tinaszheng: link to use is https://api.thegraph.com/subgraphs/name/ianlapham/uniswap-v3-bsc

src/components/AccountDetailsV2/index.tsx
tinaszheng: That may not be the case across all platforms though. Sdk-core is meant to be shared across apps and [redacted] may have different chain support than interface in the future.

src/constants/networks.ts
JFrankfurt: updated phrasing, sorry!

tinaszheng:
https://github.com/Uniswap/interface/blob/main/src/utils/switchChain.ts#L13 you probably also want to add bsc here?

src/constants/networks.ts
just-toby: in the widget we use this type: `Record<SupportedChainId, string[]>`

.env
```diff
@@ -11,3 +11,4 @@ REACT_APP_INFURA_KEY="4bf032f2d38a4ed6bb975b80d6340847"
 REACT_APP_MOONPAY_API="https://api.moonpay.com"
 REACT_APP_MOONPAY_LINK="https://us-central1-uniswap-mobile.cloudfunctions.net/signMoonpayLinkStaging?platform=web"
 REACT_APP_MOONPAY_PUBLISHABLE_KEY="pk_test_DycfESRid31UaSxhI5yWKe1r5E5kKSz"
+REACT_APP_QUICKNODE_RPC_URL="https://rough-sleek-hill.bsc.quiknode.pro/413cc98cbc776cda8fdf1d0f47003583ff73d9bf"
```
zzmp, line 14: This should be called the `REACT_APP_BNB_RPC_URL`, as quicknode includes many chains.

src/components/swap/styleds.tsx
```diff
@@ -1,3 +1,4 @@
+import { ChainId } from '@uniswap/smart-order-router'
```
zzmp, line 1: Shouldn't this be `SupportedChainId` from `constants/chains`?

src/constants/addresses.ts
```diff
@@ -22,6 +22,15 @@ const CELO_QUOTER_ADDRESSES = '0x82825d0554fA07f7FC52Ab63c961F330fdEFa8E8'
 const CELO_NONFUNGIBLE_POSITION_MANAGER_ADDRESSES = '0x3d79EdAaBC0EaB6F08ED885C05Fc0B014290D95A'
 const CELO_TICK_LENS_ADDRESSES = '0x5f115D9113F88e0a0Db1b5033D90D4a9690AcD3D'
 
+// bsc v3 addresses
```
zzmp, line 25: Confusing that we use BNB and BSC interchangeably.

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -1,5 +1,5 @@
 import { Trans } from '@lingui/macro'
-import { Currency } from '@uniswap/sdk-core'
+import { Currency, SupportedChainId } from '@uniswap/sdk-core'
```
zzmp, line 2: Should `SupportedChainId` come from `constants/chains`?

src/utils/getExplorerLink.test.ts
```diff
@@ -1,3 +1,5 @@
+import { SupportedChainId } from '@uniswap/sdk-core'
```
zzmp, line 1: Should this be from `constants/chains`?

yarn.lock
```diff
@@ -5174,6 +5174,15 @@
     "@uniswap/v2-core" "1.0.1"
     "@uniswap/v3-core" "1.0.0"
 
+"@uniswap/universal-router@^1.0.1":
```
zzmp, line 5177: This should be de-duped - just run `yarn deduplicate`. I'd expect it to fail deps-test as-is.

zzmp (commented):
All of the assets look the same - can you just use a single SVG?

src/constants/networks.ts
tinaszheng: wait so you should really have a separate `SupportedChainId` just locally to interface and then this map should exhaustively enumerate through that

src/constants/addresses.ts
JFrankfurt: the pr was initially made w/ both, and I've tried to transition to just using BNB. Thanks for catching this.

src/components/TransactionConfirmationModal/index.tsx
grabbou: Could we have a follow-up PR where we migrate to that SDK constants? I had a brief look and I am not sure whether it has 100% overlap with our constants, but this could help.

yarn.lock
JFrankfurt: `yarn deduplicate` did nothing 🤔

PR #6037, opened 2023-02-27
chore: update coinbase-wallet sdk

vm:
This means that cancelled user requests are sent with a 4001 error code, vs being an undefined error.


just-toby (approved):
nice!

PR #6042, opened 2023-02-28
chore: Cleanup and refactor significant portion of listing logic

cbachmeier:
- Removes logic that is no longer used in the latest version of list
- Moves shared functions and hooks to util or shared file
- Moves complex logic that dynamically updates shown data to shared hooks
- Use stablecoin value over 3rd party eth price fetch
- Try to remove as many useEffects as possible


src/nft/components/profile/list/ListPage.tsx
```diff
@@ -317,9 +294,7 @@ export const ListPage = () => {
             <EthValueWrapper totalEthListingValue={!!totalEthListingValue}>
               {totalEthListingValue > 0 ? formatEth(totalEthListingValue) : '-'} ETH
             </EthValueWrapper>
-            {!!totalEthListingValue && !!ethPriceInUSD && (
-              <UsdValue>{formatUsdPrice(totalEthListingValue * ethPriceInUSD)}</UsdValue>
-            )}
+            {!!totalEthListingValue && !!usdcValue && <UsdValue>{usdcAmount}</UsdValue>}
```
JackShort, line 297: you don't need to check for `totalEthListingValue` here if its undefined `usdcValue` will be too

src/nft/components/profile/list/utils.ts
```diff
@@ -123,69 +108,89 @@ export const getListings = (sellAssets: WalletAsset[]): [CollectionRow[], Listin
...
+  useEffect(() => {
+    let price: number | undefined
+    if (globalOverride) {
+      if (!listPrice) setListPrice(globalPrice)
+      price = globalPrice
+    } else {
+      price = listPrice
+    }
+    setPrice(price)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
JackShort, line 147: do we have to disable eslint here? I don't see anything we can't add to the deps

src/nft/components/profile/list/utils.ts
```diff
@@ -123,69 +108,89 @@ export const getListings = (sellAssets: WalletAsset[]): [CollectionRow[], Listin
...
+      setGlobalPrice(asset.floorPrice)
+    } else if (globalPriceMethod === SetPriceMethod.LAST_PRICE) {
+      setListPrice(asset.lastPrice)
+      setGlobalPrice(asset.lastPrice)
+    } else if (globalPriceMethod === SetPriceMethod.SAME_PRICE)
+      listPrice && !globalPrice ? setGlobalPrice(listPrice) : setListPrice(globalPrice)
+
+    setGlobalOverride(false)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [globalPriceMethod])
```
JackShort, line 172: same question here do we not want this to update when listPrice changes?

src/nft/components/profile/list/utils.ts
```diff
@@ -123,69 +108,89 @@ export const getListings = (sellAssets: WalletAsset[]): [CollectionRow[], Listin
...
+  listPrice?: number
+) {
+  useEffect(() => {
+    setWarningType(WarningType.NONE)
+    const price = listPrice ?? 0
+    inputRef.current.value = `${price}`
+    if (price < (asset?.floorPrice ?? 0) && price > 0) setWarningType(WarningType.BELOW_FLOOR)
+    else if (asset.floor_sell_order_price && price >= asset.floor_sell_order_price)
+      setWarningType(WarningType.ALREADY_LISTED)
+  }, [asset?.floorPrice, asset.floor_sell_order_price, inputRef, listPrice, setWarningType])
```
JackShort, line 188: does `inputRef` need to be a dep here?

JackShort (approved):
lgtm - works well - small questions

src/nft/components/profile/list/utils.ts
cbachmeier: Yeah linter required it.

src/nft/components/profile/list/utils.ts
cbachmeier: Yeah we want this atomic to only when the globalPriceMethod changes, otherwise it can fire prematurely and overwrite user input

src/nft/components/profile/list/utils.ts
cbachmeier: Same as below, we want this effect to be atomic to only when a user toggles the button to link/unlink an input row from the Same Price method, otherwise it unintentionally will write into the inputs

PR #6043, opened 2023-02-28
chore: refactor useFetchAssets to make it more readable

JackShort:
* note didn't refactor the old fetch code since it will be removed

### DIFF
--- a/src/nft/utils/bag.ts
+++ b/src/nft/utils/bag.ts
@@ -1,0 +1,54 @@
+import { BagItem, BagItemStatus, BagStatus, UpdatedGenieAsset } from 'nft/types'
+
+// eslint-disable-next-line import/no-unused-modules
+export function getPurchasableAssets(itemsInBag: BagItem[]): UpdatedGenieAsset[] {
+  return itemsInBag.filter((item) => item.status !== BagItemStatus.UNAVAILABLE).map((item) => item.asset)
+}
+
+// eslint-disable-next-line import/no-unused-modules
+export function createBagFromUpdatedAssets(
+  unavailable: UpdatedGenieAsset[],
+  priceChanged: UpdatedGenieAsset[],
+  unchanged: UpdatedGenieAsset[]
+): BagItem[] {
+  return [
+    ...unavailable.map((unavailableAsset) => ({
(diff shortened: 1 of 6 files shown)


src/nft/utils/bag.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    return { nextBagStatus: BagStatus.CONFIRM_REVIEW, lockBag: false }
+  }
+
+  if (shouldRefetchCalldata) {
+    return { nextBagStatus: BagStatus.CONFIRM_QUOTE, lockBag: false }
+  }
+
+  return { nextBagStatus: BagStatus.CONFIRMING_IN_WALLET, lockBag: true }
```
cbachmeier, line 53: Are there cases where the BagStatuses and lockBag state differ from what's here? You could create an enum/map mapping status to lockBag state so you don't have to track them separately.

src/nft/utils/bag.ts
```diff
@@ -0,0 +1,81 @@
...
+  const { hasPriceAdjustment, updatedAssets } = compareAssetsWithTransactionRoute(wishAssetsToBuy, route)
+  const shouldRefetchCalldata = hasPriceAdjustment && purchasingWithErc20
+
+  const { unchanged, priceChanged, unavailable } = filterUpdatedAssetsByState(updatedAssets)
+
+  const hasReviewedAssets = unchanged.length > 0
+  const hasAssetsInReview = priceChanged.length > 0
+  const hasUnavailableAssets = unavailable.length > 0
+
+  const hasAssets = hasReviewedAssets || hasAssetsInReview || hasUnavailableAssets
```
cbachmeier, line 70: very nit: could also just be !!updatedAssets.length

cbachmeier (approved):
A lot cleaner and works well.

src/nft/utils/bag.ts
JackShort: its crazy how true that is

src/nft/utils/bag.ts
JackShort: crazier how none of this is actually necessary when one state only returns false lmao

PR #6046, opened 2023-02-28
fix: hide missing % changes/prices

cartcrom:
new
image

old
image

### DIFF
--- a/src/components/NavBar/SuggestionRow.tsx
+++ b/src/components/NavBar/SuggestionRow.tsx
@@ -18,20 +18,15 @@ import { putCommas } from 'nft/utils/putCommas'
 import { useCallback, useEffect, useState } from 'react'
 import { Link, useNavigate } from 'react-router-dom'
 import styled from 'styled-components/macro'
+import { ThemedText } from 'theme'
 
-import { getDeltaArrow } from '../Tokens/TokenDetails/PriceChart'
+import { DeltaText, getDeltaArrow } from '../Tokens/TokenDetails/PriceChart'
 import { useAddRecentlySearchedAsset } from './RecentlySearchedAssets'
 import * as styles from './SearchBar.css'
 
 const PriceChangeContainer = styled.div`
   display: flex;
   align-items: center;
-`
-
-const PriceChangeText = styled.span<{ isNegative: boolean }>`
-  font-size: 14px;
-  line-height: 20px;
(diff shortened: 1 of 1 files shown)


just-toby (approved):

src/components/NavBar/SuggestionRow.tsx
```diff
@@ -191,18 +186,20 @@ export const TokenRow = ({ token, isHovered, setHoveredIndex, toggleOpen, index,
...
-          </Row>
-        )}
-        {token.market?.pricePercentChange?.value && (
-          <PriceChangeContainer>
-            <ArrowCell>{arrow}</ArrowCell>
-            <PriceChangeText isNegative={token.market.pricePercentChange.value < 0}>
-              {Math.abs(token.market.pricePercentChange.value).toFixed(2)}%
-            </PriceChangeText>
-          </PriceChangeContainer>
+        {!!token.market?.price?.value && (
```
just-toby, line 189: `!!` seems unclear to me. are we doing it because the value can be `0` which is falsy? is there a clearer way to express that, or should I just get with the program?

src/components/NavBar/SuggestionRow.tsx
```diff
@@ -191,18 +186,20 @@ export const TokenRow = ({ token, isHovered, setHoveredIndex, toggleOpen, index,
...
+          <>
+            <Row gap="4">
+              <Box className={styles.primaryText}>{formatUSDPrice(token.market.price.value)}</Box>
+            </Row>
+            <PriceChangeContainer>
+              <ArrowCell>{arrow}</ArrowCell>
+              <ThemedText.BodySmall>
+                <DeltaText delta={token.market?.pricePercentChange?.value}>
+                  {Math.abs(token.market?.pricePercentChange?.value ?? 0).toFixed(2)}%
```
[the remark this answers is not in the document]
cartcrom: Our designs currently always show 0% in green with an upward arrow. Design seems to prefer that over adding a third color or not showing at all 🤷

src/components/NavBar/SuggestionRow.tsx
cartcrom: Yeah, had to add it for the 0 issue. Boolean(token.market?.price?.value) doesn't appease the typechecker 😢 and creating a sep var for this seems like too much

PR #6053, opened 2023-03-02
build: upgrade widgets version and smart-order-router

vm:
Latest, which adds current ethers version in widget and also the smart-order-router upgrade in both

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5027,10 +5027,10 @@
     "@uniswap/v2-sdk" "^3.0.1"
     "@uniswap/v3-sdk" "^3.8.3"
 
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -5114,6 +5141,15 @@
     "@uniswap/v2-core" "1.0.1"
     "@uniswap/v3-core" "1.0.0"
 
+"@uniswap/universal-router@^1.0.1":
```
zzmp: This should be deduped with the above - how come there are two versions?

yarn.lock
```diff
@@ -5063,6 +5063,33 @@
     node-cache "^5.1.2"
     stats-lite "^2.2.0"
 
+"@uniswap/smart-order-router@^3.6.0":
```
zzmp, line 5042: How come this adds a new version of SOR? What still depends on ^2?

zzmp (approved):

PR #6056, opened 2023-03-02
fix: getQuote return value

JFrankfurt:


zzmp (commented):
In what case would this be returning null data? Should it instead return an error for no route? 

Does widget handle this differently (I know I've added add'l types to deal with non-error non-quote states.

src/state/routing/slice.ts
```diff
@@ -115,7 +115,7 @@ export const routingApi = createApi({
             )
           }
 
-          return { data: result.data as GetQuoteResult }
+          return { data: (result.data || {}) as GetQuoteResult }
```
justindomingue, line 118: should you return undefined instead?

probably not type safe to cast `{} as GetQuoteResult` given `GetQuoteResult` has required props?

JFrankfurt: It's currently returning undefined, and that's causing a bunch of console errors on networks where this query fails. (BNB) Though, no observable ux issues afaict.

JFrankfurt: The way I'm reading the error, I think returning `undefined` **and** an `error` value would work too, but I wasn't sure about that and this seemed like the smallest change.

JFrankfurt: image

JFrankfurt: what do you think of this? 

```suggestion
          if (result.data) {
            return { data: result.data as GetQuoteResult }
          }
          return { data: undefined, error: result.error as FetchBaseQueryError }
```

PR #6057, opened 2023-03-02
fix: show error message when token is not found

just-toby:
fixes a live bug in prod where we show an empty modal for tokens that don't match the current chain

i think we should use this as a short term fix, so we show _something_ in the case that we can't find the token. 

in the future, we may want to try auto-switching chains, or showing a message that says "try a different chain"

to test, append `?outputCurrency=0x4200000000000000000000000000000000000006` to the preview link (after `/swap`)

# before
https://user-images.githubusercontent.com/66155195/223544058-aabd47c6-3114-4033-9369-7de492bc7e7a.mov

# after 

https://user-images.githubusercontent.com/66155195/223544401-d4011331-f019-4259-a539-1d86693dc7d7.mov

### DIFF
--- a/src/components/TokenSafety/index.tsx
+++ b/src/components/TokenSafety/index.tsx
@@ -77,7 +77,7 @@ const Buttons = ({
   showCancel,
 }: {
   warning: Warning
-  onContinue: () => void
+  onContinue?: () => void
   onCancel: () => void
   onBlocked?: () => void
   showCancel?: boolean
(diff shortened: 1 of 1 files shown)


src/components/TokenSafety/index.tsx
```diff
@@ -251,32 +251,49 @@ export default function TokenSafety({
...
+      <Container>
+        <ShortColumn>
+          <SafetyLabel warning={tokenNotFoundWarning} />
+        </ShortColumn>
+        <ShortColumn>
+          <InfoText>
+            {heading} {description} {learnMoreUrl}
+          </InfoText>
+        </ShortColumn>
+        <Buttons warning={tokenNotFoundWarning} onCancel={onCancel} showCancel={true} onContinue={() => undefined} />
```
vm, line 295: can you make onContinue optional or something?

vm (approved):
check with Annabel and LGTM

PR #6058, opened 2023-03-03
build: fix vanilla-extract for M1

zzmp:
Works around a bug where vanilla-extract compilation hangs on M1 macs, causing `yarn start` to never resolve.
See https://github.com/vanilla-extract-css/vanilla-extract/issues/771#issuecomment-1249524366.

This configures vanilla-extract class names to use 7+ character hashes, which will make them more difficult to debug. However, I don't know that the debug names are being used, and they are making it very difficult to develop with an M1 macbook, so this seems worthwhile.

### DIFF
--- a/craco.config.cjs
+++ b/craco.config.cjs
@@ -22,7 +22,7 @@ module.exports = {
   },
   webpack: {
     plugins: [
-      new VanillaExtractPlugin(),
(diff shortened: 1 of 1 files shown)


cbachmeier (approved):
LGTM. I can't speak on the hanging issue but I was able to build locally with this branch and all styles appeared correct.

PR #6059, opened 2023-03-03
fix: don't show a price impact warning if the user is (un)wrapping

JFrankfurt:


src/hooks/useClientSideV3Trade.ts
```diff
@@ -57,16 +57,18 @@ export function useClientSideV3Trade<TTradeType extends TradeType>(
     gasRequired: chainId ? QUOTE_GAS_OVERRIDES[chainId] ?? DEFAULT_GAS_QUOTE : undefined,
   })
 
+  const currenciesAreTheSame = useMemo(
+    () => currencyIn && currencyOut && (currencyIn.equals(currencyOut) || currencyIn.wrapped.equals(currencyOut)),
```
just-toby, line 61: does `equals` always do the check we want here? will it checksum both sides?

just-toby (approved):

src/hooks/useClientSideV3Trade.ts
JFrankfurt: the currency entity checksums both sides when the asset is created. I think this might still let through a quote attempt if the same asset is provided as input and output, but only one is checksummed.

tinaszheng: note there is an optional argument on `Currency` creation that bypasses the checksum, but i think sdk-core should fix the address comparison to be case insensitive 🤔

tinaszheng:
image
we should probably hide the usd change % for wraps/unwraps too

src/hooks/useClientSideV3Trade.ts
tinaszheng: 🤔 feels like we should also make it more explicit that this is for wraps/unwraps by making the var name `isWrap`. the semantics of calling it the same currency isnt totally accurate imo

and i know this was existing code but also doesnt the store already prevent the case where `currencyIn.equals(currencyOut)`?

PR #6061, opened 2023-03-03
feat: Unicons ✨

cartcrom:
image


tinaszheng (approved):

katspaugh:
These avatars look really dope! Can we use it in the Safe interface?

PR #6062, opened 2023-03-03
chore: update types to match schema changes

JFrankfurt:


cbachmeier (approved):
LGTM. Is there a GH action we could setup that runs `yarn prepare` and checks in any changes daily? This way we don't have to do it manually whenever infra makes a shema change?

PR #6066, opened 2023-03-03
refactor: variable for goerli in test

vm:
This will avoid bugs in the future where people don't know that the hex return is tied to the expected testnet id.


just-toby (approved):
looks good

PR #6068, opened 2023-03-03
refactor: always enable permit2

vm:
Now that only permit2-deployed chains are supported in the interface, we can start to simplify this logic.


vm:
Just wanted to add this to make sure there wasn't confusion in the interim, but eventually rest of the code will get deleted with the move to Widget anyways

just-toby (dismissed):
why not remove the dead code in the same PR?

just-toby (approved):
LGTM

PR #6070, opened 2023-03-03
test: disable /send redirect test temporarily

vm:
Unclear what caused this to start being flake-y but it does work on localhost for me. I will renable this once I can look at it later today, but just don't want this to block deploys on Monday.


just-toby (approved):
we probably just need to set up a valid goerli ens address

PR #6073, opened 2023-03-03
fix: deep link by swapping sooner after user interaction

zzmp:
Uses `@uniswap/conedison/provider` implementation of `sendTransaction`, which includes a gas margin computation.

It also omits the gas margin computation for Uniswap Wallet, and uses sendUncheckedTransaction under the hood. This allows the app to be linked to within 300ms, which allows iOS to deep-link through WalletConnect to the mobile app instead of navigating the to app's webpage.

To test (this may not work on cloudflare/vercel deploys - I'm not sure why, but it may be because they add instrumentation):
- Using your iOS device, connect to the App using Uniswap Wallet via WalletConnect
- Initiate a swap/wrap/etc on the app
- The application should deep-link directly to the wallet

Requires https://github.com/Uniswap/conedison/pull/13 to poll before merging.


src/hooks/useUniversalRouter.ts
```diff
@@ -3,14 +3,14 @@ import { BigNumber } from '@ethersproject/bignumber'
 import { t } from '@lingui/macro'
 import { sendAnalyticsEvent } from '@uniswap/analytics'
 import { SwapEventName } from '@uniswap/analytics-events'
+import { sendTransaction } from '@uniswap/conedison/provider/index'
```
tinaszheng, line 6: should this be configured at the webpack level to resolve as an alias?

src/hooks/useUniversalRouter.ts
```diff
@@ -52,30 +52,19 @@ export function useUniversalRouterSwapCallback(
...
-      }
-      const gasLimit = calculateGasMargin(gasEstimate)
-      const response = await provider
-        .getSigner()
-        .sendTransaction({ ...tx, gasLimit })
-        .then((response) => {
-          sendAnalyticsEvent(
-            SwapEventName.SWAP_SIGNED,
-            formatSwapSignedAnalyticsEventProperties({ trade, txHash: response.hash })
+      const response = await sendTransaction(provider, tx, 0.2).then((response) => {
```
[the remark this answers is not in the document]
tinaszheng: looks like it's being used in the Add liquidity code too - maybe a global const `TX_GAS_LIMIT_PADDING_FACTOR`?

tinaszheng (approved):
🙌

src/hooks/useUniversalRouter.ts
zzmp: I couldn't figure it out and didn't want to sink too much time into it.

src/nft/hooks/useSendTransaction.ts
```diff
@@ -36,21 +37,19 @@ export const useSendTransaction = create<TxState>()(
...
       setState: (newState) => set(() => ({ state: newState })),
-      sendTransaction: async (signer, selectedAssets, transactionData, purchasedWithErc20) => {
-        const address = await signer.getAddress()
+      sendTransaction: async (provider, selectedAssets, transactionData, purchasedWithErc20) => {
+        const address = await provider.getSigner().getAddress()
         try {
-          const txNoGasLimit = {
+          set({ state: TxStateType.Signing })
+
+          // TODO(signer should be provider)
```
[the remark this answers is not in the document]
zzmp: This was a note to myself that I'd resolved and forgotten to remove: literally to replace the signer on L39/40 with a provider, because the new method uses a provider.

src/hooks/useUniversalRouter.ts
tinaszheng: i meant inside the webpack config for interface, same way we've alias'd the import for `conedison` to the dist

PR #6077, opened 2023-03-05
chore: upgrade web3-react, widgets

vm:
- Upgrades web3-react gnosis safe connector to use the latest safe packages. This has EIP712 support, which is required for permit2.
- Resolves an issue with walletconnect connector where the wrong chainId was set for initial requests


yarn.lock
```diff
@@ -5804,12 +5804,12 @@
   dependencies:
     "@walletconnect/window-getters" "^1.0.0"
 
-"@web3-react/coinbase-wallet@8.0.35-beta.0":
-  version "8.0.35-beta.0"
-  resolved "https://registry.yarnpkg.com/@web3-react/coinbase-wallet/-/coinbase-wallet-8.0.35-beta.0.tgz#903df113a0987f0b28c07941a82c05f227cbccf8"
-  integrity sha512-wSITb75xIfERw1rZA8t35mJ3Lq459Emt2ybraWvT0TAIVTzHGCKvEm3WglEVWj5CMy1X1lrJvWJ0ZfWUj1/0Lg==
+"@web3-react/coinbase-wallet@8.1.0-beta.0":
```
zzmp, line 5807: You have both versions, likely because of widgets. I don't think your fix will affect widgets.
Would you consider first fixing widgets and then updating interface?

src/connection/index.ts
```diff
@@ -66,7 +66,6 @@ const [web3WalletConnect, web3WalletConnectHooks] = initializeConnector<WalletCo
     actions,
     options: {
       rpc: RPC_URLS_WITHOUT_FALLBACKS,
-      qrcode: true,
```
zzmp, line 69: Why are you removing this?

vm:
This PR is incorrect, I will update once Mike releases new versions on 8.0.x.beta.0

yarn.lock
vm: yes

vm:
dependent on https://github.com/Uniswap/widgets/pull/537

yarn.lock
```diff
@@ -20260,9 +20334,16 @@ zen-observable@0.8.15:
...
   integrity sha512-PQ2PC7R9rslx84ndNBZB/Dkv8V8fZEpk83RLgXtYd0fwUgEjseMn1Dgajh2x6S8QbZAFa9p2qVCEuYZNgve0dQ==
 
+zustand@4.3.3:
+  version "4.3.3"
+  resolved "https://registry.yarnpkg.com/zustand/-/zustand-4.3.3.tgz#c9113499074dde2d6d99c1b5f591e9329572c224"
+  integrity sha512-x2jXq8S0kfLGNwGh87nhRfEc2eZy37tSatpSoSIN+O6HIaBhgQHSONV/F9VNrNcBcKQu/E80K1DeHDYQC/zCrQ==
+  dependencies:
+    use-sync-external-store "1.2.0"
+
 zustand@^4.0.0-rc.0, zustand@^4.0.0-rc.1:
```
grabbou, line 20344: Looks like these packages are resolved to `4.3.5` and there's another `4.3.3`. Can you try running dedupe to see if the consensus is `4.3.3` here?

grabbou: `4.3.4` is broken FYI, so `4.3.5` is "safe". I will send follow-ups to widget/interface myself later.

grabbou (approved):
LGTM, but let's see if we can do sth about `zustand` versions (see my comment)

yarn.lock
vm: does not work

PR #6080, opened 2023-03-06
feat: export window.GIT_COMMIT_HASH

zzmp:
Export the `GIT_COMMIT_HASH` onto the `window` object so that we can verify which version is deployed.

### DIFF
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -63,6 +63,10 @@ const AMPLITUDE_PROXY_URL = process.env.REACT_APP_AMPLITUDE_PROXY_URL
 const STATSIG_DUMMY_KEY = 'client-0000000000000000000000000000000000000000000'
 const STATSIG_PROXY_URL = process.env.REACT_APP_STATSIG_PROXY_URL
 const COMMIT_HASH = process.env.REACT_APP_GIT_COMMIT_HASH
+
+// Dump some metadata into the window to allow client verification.
+window.GIT_COMMIT_HASH = COMMIT_HASH
+
 initializeAnalytics(AMPLITUDE_DUMMY_KEY, OriginApplication.INTERFACE, {
   proxyUrl: AMPLITUDE_PROXY_URL,
   defaultEventName: SharedEventName.PAGE_VIEWED,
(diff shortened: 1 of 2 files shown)


src/pages/App.tsx
```diff
@@ -63,6 +63,10 @@ const AMPLITUDE_PROXY_URL = process.env.REACT_APP_AMPLITUDE_PROXY_URL
 const STATSIG_DUMMY_KEY = 'client-0000000000000000000000000000000000000000000'
 const STATSIG_PROXY_URL = process.env.REACT_APP_STATSIG_PROXY_URL
 const COMMIT_HASH = process.env.REACT_APP_GIT_COMMIT_HASH
+
+// Dump some metadata into the window to allow client verification.
+window.GIT_COMMIT_HASH = COMMIT_HASH
```
grabbou, line 68: nit: not that have any particular preference, but just wanted to point out that we seem to be touching global namespace / interacting with top-level code in a few places already, particularly in `index.tsx` (where I configure Sentry or we set `window.ethereum.autoRefreshOnNetworkChange`) or `polyfills.ts`.  I think it would make sense to move it one level up, to be side by side `react-app-env.d.ts`.

src/react-app-env.d.ts
```diff
@@ -5,6 +5,8 @@ declare module '@metamask/jazzicon' {
 }
 
 interface Window {
+  GIT_COMMIT_HASH?: string
```
grabbou, line 8: Isn't this always defined? If we put it in `index.tsx`, at the top, where polyfills are, I think this would be a safe assumption.

Maybe we could have e `window.ts` top-level file next to `polyfills` where we have all the window touching things + export the type.

grabbou (approved):
Overall, it's all good and tiny PR! Just left a high-level comment on the place/type, but it's good to go overall.

src/react-app-env.d.ts
zzmp: the env variable is typed as `string | undefined`, so I made it optional to match that.
I think this is the idiomatic file to extend Window in, from a CRA perspective.

src/pages/App.tsx
zzmp: I don't disagree - I was going to move around the 3P initializations in a follow-up, and I could include this as well.

PR #6081, opened 2023-03-06
feat: NFT Collection Search via GraphQL

cbachmeier:
- Adds hook to use GQL query for searching collections
- Converts some functionality of Collection query to shared function
- Conditionally call search based off feature flag

### DIFF
--- a/src/graphql/data/nft/CollectionSearch.ts
+++ b/src/graphql/data/nft/CollectionSearch.ts
@@ -16,68 +16,23 @@ gql`
       edges {
         cursor
         node {
-          bannerImage {
-            url
-          }
-          collectionId
-          description
-          discordUrl
-          homepageUrl
           image {
             url
           }
-          instagramName
           isVerified
           name
           numAssets
-          twitterName
           nftContracts {
             address
             chain
             name
-            standard
             symbol
             totalSupply
-          }
-          traits {
-            name
-            values
-            stats {
-              name
-              value
-              assets
-              listings
-            }
           }
           markets(currencies: ETH) {
             floorPrice {
               currency
               value
-            }
-            owners
-            totalVolume {
-              value
-              currency
-            }
-            listings {
-              value
-            }
-            volume(duration: DAY) {
-              value
-              currency
-            }
-            volumePercentChange(duration: DAY) {
-              value
-              currency
-            }
-            floorPricePercentChange(duration: DAY) {
-              value
-              currency
-            }
-            marketplaces {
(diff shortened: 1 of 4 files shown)


src/graphql/data/nft/CollectionSearch.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }, [loading, queryData])
+}
+
+export function useCollectionSearch(queryOrAddress: string): useCollectionSearchReturnProps {
+  const isNftGraphqlEnabled = useNftGraphqlEnabled()
+  const isName = !isAddress(queryOrAddress.toLowerCase())
+  const queryResult = useCollectionQuerySearch(queryOrAddress, isNftGraphqlEnabled ? !isName : true)
+  const addressResult = useCollection(queryOrAddress, isNftGraphqlEnabled ? isName : true)
+  return isName ? queryResult : { data: [addressResult.data], loading: addressResult.loading }
```
JackShort, line 87: since addressResult almost returns the correct type why not change the return type so it wraps it in an array?

src/graphql/data/nft/CollectionSearch.ts
```diff
@@ -0,0 +1,88 @@
...
+import { useNftGraphqlEnabled } from 'featureFlags/flags/nftlGraphql'
+import gql from 'graphql-tag'
+import { GenieCollection } from 'nft/types'
+import { blocklistedCollections } from 'nft/utils'
+import { useMemo } from 'react'
+
+import { NftCollection, useCollectionSearchQuery } from '../__generated__/types-and-hooks'
+import { formatCollectionQueryData, useCollection } from './Collection'
+
+const MAX_SEARCH_RESULTS = 6
```
JackShort, line 11: are the max search results not 5? `slice(0, 5)` will return 5 items - or did we change the spec to 6?

JackShort (approved):
lgtm

src/graphql/data/nft/CollectionSearch.ts
cbachmeier: useCollection is our hook for getting collection data on the Collection page. I've repurposed it here so refactoring it for just this purpose would also require a refactor to the whole collection page.

src/graphql/data/nft/CollectionSearch.ts
cbachmeier: Max is 6, you can be shown up to 6 if there are less than 3 token results

PR #6084, opened 2023-03-06
feat: add Peer Wallet Name to WALLET_CONNECT_TXN_COMPLETED event

just-toby:
https://uniswaplabs.atlassian.net/browse/WEB-2879?atlOrigin=eyJpIjoiZGQ1NTVmNDU2MzY1NGRmMjgwMzU4ZTM2MGYzYmEwMTMiLCJwIjoiaiJ9

log the "Peer wallet name" to the `WALLET_CONNECT_TXN_COMPLETED` event - this is the WalletConnect-ed wallet (i.e. if you use wallet connect to connect to Rainbow, this field will be Rainbow)
Screenshot 2023-03-06 at 2 00 35 PM

update: the actual value will be the Peer Wallet "Agent" which includes some more technical details about the connected wallet.

### DIFF
--- a/src/components/WalletModal/index.tsx
+++ b/src/components/WalletModal/index.tsx
@@ -1,6 +1,7 @@
 import { Trans } from '@lingui/macro'
 import { sendAnalyticsEvent, user } from '@uniswap/analytics'
 import { CustomUserProperties, InterfaceEventName, WalletConnectionResult } from '@uniswap/analytics-events'
+import { getWalletMeta } from '@uniswap/conedison/provider/meta'
 import { useWeb3React } from '@web3-react/core'
 import { Connector } from '@web3-react/types'
 import { WalletConnect } from '@web3-react/walletconnect'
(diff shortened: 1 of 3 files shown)


src/components/WalletModal/index.tsx
```diff
@@ -211,13 +215,14 @@ export default function WalletModal({
   useEffect(() => {
     if (account && account !== lastActiveWalletAddress) {
       const walletType = getConnectionName(getConnection(connector).type)
+      const peerWalletName = provider ? getWalletMeta(provider)?.name : undefined
```
zzmp, line 218: I think you should use the agent instead of the name, as it will disambiguate how the wallet is being used (eg it will include things like if a user connected via qr, or connected via WC or Injected). This was the intention of making the more verbose `agent` available ([src](https://github.com/Uniswap/conedison/blob/main/src/provider/meta.ts#L32)) along with the `name`, which is intended for UI.

zzmp (changes requested):

src/components/WalletModal/index.tsx
```diff
@@ -129,16 +130,19 @@ const sendAnalyticsEventAndUserInfo = (
...
-  isReconnect: boolean
+  isReconnect: boolean,
+  peerWalletAgent: string | undefined
 ) => {
   sendAnalyticsEvent(InterfaceEventName.WALLET_CONNECT_TXN_COMPLETED, {
     result: WalletConnectionResult.SUCCEEDED,
     wallet_address: account,
     wallet_type: walletType,
     is_reconnect: isReconnect,
+    peer_wallet_name: peerWalletAgent,
```
zzmp, line 141: Should this now read `peer_wallet_agent`?
```suggestion
    peer_wallet_agent: peerWalletAgent,
```

src/components/WalletModal/index.tsx
```diff
@@ -129,16 +130,19 @@ const sendAnalyticsEventAndUserInfo = (
...
   })
   user.set(CustomUserProperties.WALLET_ADDRESS, account)
   user.set(CustomUserProperties.WALLET_TYPE, walletType)
+  user.set(CustomUserProperties.PEER_WALLET_NAME, peerWalletAgent ?? '')
```
zzmp, line 145: Should this now read `PEER_WALLET_AGENT`? This would need to be upstreamed.

zzmp (approved):
lgtm, I'm just wondering if analytics fields should be called "agent" instead of "user". Either is fine, I'm just concerned "user" may cause confusion in the future.

src/components/WalletModal/index.tsx
zzmp: Could you update the PR description to match this?

PR #6086, opened 2023-03-07
fix: react router updated appropriate fix

JackShort:
.

### DIFF
--- a/src/nft/components/collection/CollectionNfts.tsx
+++ b/src/nft/components/collection/CollectionNfts.tsx
@@ -428,7 +428,9 @@ export const CollectionNfts = ({ contractAddress, collectionStats, rarityVerifie
       const modifiedQuery = applyFiltersFromURL(location, collectionStats)
 
       requestAnimationFrame(() => {
-        useCollectionFilters.setState(modifiedQuery as any)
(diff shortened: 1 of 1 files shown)


src/nft/components/collection/CollectionNfts.tsx
```diff
@@ -428,7 +428,9 @@ export const CollectionNfts = ({ contractAddress, collectionStats, rarityVerifie
       const modifiedQuery = applyFiltersFromURL(location, collectionStats)
 
       requestAnimationFrame(() => {
-        useCollectionFilters.setState(modifiedQuery as any)
+        if (modifiedQuery) {
+          useCollectionFilters.setState(modifiedQuery as any)
```
cbachmeier, line 432: an alternative: `useCollectionFilters.setState(modifiedQuery || initialCollectionFilterState)`

JackShort: Actually resetting the state completely isn't what we want it will wipe out the market count. Setting the state to undefined actually does not mutate the state at all so it's like no change happened

cbachmeier (approved):
LGTM

PR #6090, opened 2023-03-07
chore: upgrade web3-react, zustand

vm:
Reverts Uniswap/interface#6085

This time I will upgrade web3-react with a version that has zustand unpinned.


package.json
```diff
@@ -165,16 +165,16 @@
...
+    "@web3-react/coinbase-wallet": "8.1.1-beta.0",
+    "@web3-react/core": "8.1.1-beta.0",
+    "@web3-react/eip1193": "8.1.1-beta.0",
+    "@web3-react/empty": "8.1.1-beta.0",
+    "@web3-react/gnosis-safe": "8.1.1-beta.0",
+    "@web3-react/metamask": "8.1.1-beta.0",
+    "@web3-react/network": "8.1.1-beta.0",
+    "@web3-react/types": "8.1.1-beta.0",
+    "@web3-react/url": "8.1.1-beta.0",
+    "@web3-react/walletconnect": "8.1.1-beta.0",
```
zzmp, line 177: Can we just remove these and use the dependencies brought in by @uniswap/widgets? That way they will stay in sync.

vm: that seems a bit too implicit for me... what if we used resolutions?

grabbou: Agreed. We're using this directly in this project, so makes sense to have an explicit dependency.

yarn.lock
```diff
@@ -20260,9 +20334,9 @@ zen-observable@0.8.15:
   resolved "https://registry.yarnpkg.com/zen-observable/-/zen-observable-0.8.15.tgz#96415c512d8e3ffd920afd3889604e30b9eaac15"
   integrity sha512-PQ2PC7R9rslx84ndNBZB/Dkv8V8fZEpk83RLgXtYd0fwUgEjseMn1Dgajh2x6S8QbZAFa9p2qVCEuYZNgve0dQ==
 
-zustand@^4.0.0-rc.0, zustand@^4.0.0-rc.1:
-  version "4.0.0"
-  resolved "https://registry.yarnpkg.com/zustand/-/zustand-4.0.0.tgz#739cba69209ffe67b31e7d6741c25b51496114a7"
-  integrity sha512-OrsfQTnRXF1LZ9/vR/IqN9ws5EXUhb149xmPjErZnUrkgxS/gAHGy2dPNIVkVvoxrVe1sIydn4JjF0dYHmGeeQ==
+zustand@^4.0.0-rc.0, zustand@^4.3.5, zustand@^4.3.6:
```
grabbou, line 20337: Do we know where's `zustand@^4.0.0-rc.0` coming from? I don't see it being mentioned anywhere in the code. I wonder if removing it and running `yarn` again would update the lock file. Looks like a leftover from previous `package.json`. Did you re-run `yarn` after updating it?

vm: yes i did, but yarn.locks don't get previous requirements removed afaik

vm: because -rc isn't shared with the stable release

grabbou: Sounds like a bug to me, but this is V1, so maybe it got fixed over time.

grabbou: I would be tempted to tweak this manually, but the file is explicitly annotated as "do not modify" so I guess we're good.

grabbou (approved):
LGTM

PR #6091, opened 2023-03-07
chore: derive SupportedChainId from sdk-core

vm:
Right now, we have SupportedChainId in sdk-core exported, but in practice every client creates their own SupportedChainId enum.

examples:
https://github.com/Uniswap/widgets/blob/main/src/constants/chains.ts#L4
https://github.com/Uniswap/smart-order-router/blob/main/src/util/chains.ts#L4
https://github.com/Uniswap/interface/blob/main/src/constants/chains.ts#L5

I was initially thinking of making interface and widgets use the sdk-core enum, but that has some downsides...
- if we use the enum directly, then we need to add full support for all new chains if we want to upgrade sdk-core (say, for a bug fix)
- if we use Partial on top of the enum, we lose type safety that makes sure we cover all chains
- It's not possible fr the supported chain id's value to come from the sdk's enum. This is because the values are not literals, and this breaks typescript's enum rules.

Instead, we just have a test to make sure that all the values in interface's supported chain id also exist in sdk-core's enum.

### DIFF
--- a/src/components/NetworkAlert/NetworkAlert.tsx
+++ b/src/components/NetworkAlert/NetworkAlert.tsx
@@ -151,9 +151,7 @@ export function NetworkAlert() {
   }
 
   const chainInfo = getChainInfo(chainId)
-  if (!chainInfo) {
-    return null
-  }
+  if (!chainInfo) return null
 
   const { label, logoUrl, bridge } = chainInfo
   const textColor = TEXT_COLORS[chainId]
--- a/src/components/NetworkAlert/NetworkAlert.tsx
+++ b/src/components/NetworkAlert/NetworkAlert.tsx
@@ -150,7 +150,12 @@ export function NetworkAlert() {
     return null
   }
 
-  const { label, logoUrl, bridge } = getChainInfo(chainId)
+  const chainInfo = getChainInfo(chainId)
+  if (!chainInfo) {
+    return null
+  }
+
+  const { label, logoUrl, bridge } = chainInfo
   const textColor = TEXT_COLORS[chainId]
 
   return bridge ? (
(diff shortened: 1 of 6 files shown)


tinaszheng (dismissed):

vm:
Turns out this is maybe not valid syntax

src/components/NetworkAlert/NetworkAlert.tsx
```diff
@@ -150,7 +150,12 @@ export function NetworkAlert() {
     return null
   }
 
-  const { label, logoUrl, bridge } = getChainInfo(chainId)
+  const chainInfo = getChainInfo(chainId)
+  if (!chainInfo) {
+    return null
+  }
```
zzmp, line 156: nit: I find guards more readable as one-liners. [This was Google's policy](https://google.github.io/styleguide/tsguide.html#control-flow-statements-blocks), but is not ours.

```suggestion
  if (!chainInfo) return null
```

src/constants/chains.ts
```diff
@@ -1,22 +1,20 @@
-/**
- * List of all the networks supported by the Uniswap Interface
- * TODO(INFRA-90): Eventually this may be derived from sdk-core.
- */
+import { SupportedChainId as SdkSupportedChainId } from '@uniswap/sdk-core'
```
zzmp, line 1: This should be accompanied by a lint rule that prevents importing SupportedChainId elsewhere to prevent regressions, which would be appropriate to add as an optional rule in our shared uniswap eslint config.

zzmp (changes requested):
This should be accompanied by a new eslint rule to prevent regressions, otherwise its not enforced.

zzmp (approved):
lgtm but could you update the PR description?

Issue #6089, opened 2023-03-07
RPC error when creating or adding a position in v4.200.2


kallydev:
**Bug Description**

Creating or adding liquidity to a position is not available in v4.200.2 but works fine in v4.200.1.

**Steps to Reproduce**

### v4.200.2

https://bafybeigecqkvhdfhplp74xvgtst7firstgkpmyxl6ys3fvntw5xvk6gk74.ipfs.cf-ipfs.com/

### v4.200.1

https://bafybeicnf6njex3i2u5adegdzb2exxor63hygebd4s2hkz5y2xhp46un2i.ipfs.cf-ipfs.com/

1. Go to the `Pool`.
2. Click on `New Position`.
3. Select any token pair and add liquidity.

![image](https://user-images.githubusercontent.com/36319157/223452837-57266d08-eeb8-48c2-be93-0cf5d4ca2e07.png)

<img width="1223" alt="image" src="https://user-images.githubusercontent.com/36319157/223450175-e41e2919-8fb7-48d9-978a-6c0a09e1e9dc.png">

**Expected Behavior**
Rollback to v4.200.1, or fix it.

**Additional Context**
The error occurs on desktop and mobile and can be reproduced on any currently supported network.

JFrankfurt:
Thanks for reporting. We're working on this now.

PR #6092, opened 2023-03-07
fix: Revert "fix: deep link by swapping sooner after user interaction"

vm:
Reverts Uniswap/interface#6073

This is leading to LP add liquidity to fail.

### DIFF
--- a/src/utils/calculateGasMargin.ts
+++ b/src/utils/calculateGasMargin.ts
@@ -1,10 +1,9 @@
 import { BigNumber } from '@ethersproject/bignumber'
-import { TX_GAS_MARGIN } from 'constants/misc'
 
 /**
  * Returns the gas value plus a margin for unexpected or variable gas costs
  * @param value the gas value to pad
  */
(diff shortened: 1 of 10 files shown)


src/utils/calculateGasMargin.ts
```diff
@@ -1,10 +1,9 @@
 import { BigNumber } from '@ethersproject/bignumber'
-import { TX_GAS_MARGIN } from 'constants/misc'
 
 /**
  * Returns the gas value plus a margin for unexpected or variable gas costs
  * @param value the gas value to pad
  */
 export function calculateGasMargin(value: BigNumber): BigNumber {
-  return value.add(value.mul(Math.floor(TX_GAS_MARGIN * 100)).div(100))
+  return value.mul(120).div(100)
```
vm, line 8: pretty sure this is the root issue. we can just change this back if that is better

JFrankfurt (approved):

JFrankfurt:
This should resolve #6089

PR #6094, opened 2023-03-07
feat: Support listing to Seaport v1.4

cbachmeier:
OpenSea will no longer accept listings to previous versions of Seaport starting [March 9th](https://twitter.com/apiopensea/status/1631416147865382913?s=20). This updates our listing logic to use v1.4. Also updates UX to reflect OS's 0% fees and fixes a bug where long creator royalties weren't rounded.

Screenshots:
Screen Shot 2023-03-07 at 10 29 49
Screen Shot 2023-03-07 at 10 29 55


src/nft/utils/listNfts.ts
```diff
@@ -147,21 +145,20 @@ export async function signListing(
...
                 token: asset.asset_contract.address,
                 identifier: asset.tokenId,
-                amount: '1',
               },
             ],
             consideration: considerationItems,
             endTime: asset.expirationTime.toString(),
-            zone: OPENSEA_DEFAULT_ZONE,
-            restrictedByZone: true,
+            zone: ZERO_ADDRESS,
```
cbachmeier, line 152: The tweet recommends doing this however OS's web app still uses the default zone. Currently asking their team for guidance.

src/nft/utils/listNfts.ts
```diff
@@ -40,7 +41,7 @@ export const ListingMarkets: ListingMarket[] = [
   },
   {
     name: 'OpenSea',
-    fee: 2.5,
+    fee: 0,
```
JackShort, line 44: maybe make this a const since they are probably going to change it lmao

src/nft/utils/listNfts.ts
```diff
@@ -147,21 +145,20 @@ export async function signListing(
...
-            zone: OPENSEA_DEFAULT_ZONE,
-            restrictedByZone: true,
+            zone: ZERO_ADDRESS,
             allowPartialFills: true,
           },
           signerAddress
         )
 
         const order = await executeAllActions()
+        const seaportV14Order = { ...order, protocol_address: OPENSEA_SEAPORT_V1_4_CONTRACT }
```
JackShort, line 159: why not just change the order response to include this since this is going to be the new contract going forward?

yarn.lock
```diff
@@ -13887,10 +13887,10 @@ merge@2.1.0:
   resolved "https://registry.npmjs.org/merge/-/merge-2.1.0.tgz"
   integrity sha512-TcuhVDV+e6X457MQAm7xIb19rWhZuEDEho7RrwxMpQ/3GhD5sDlnP188gjQQuweXHy9igdke5oUtVOXX1X8Sxg==
 
-merkletreejs@^0.2.31:
```
JackShort, line 13890: wondering if there was any reason this was pinned to 0.2?

JackShort (approved):
lgtm

yarn.lock
cbachmeier: Might have been a rec for a previous version?

src/nft/utils/listNfts.ts
cbachmeier: executeAllActions is from seaport.createOrder so I can't change the response type, I just need to include this `protocol_address` param in the body of the api request. I tried adding it here and modifying the payload in `PostOpenSeaSellOrder` and this was cleaner imo

src/nft/utils/listNfts.ts
cbachmeier: This is basically the const for the OS ListingMarket object, so changing it here propagated throughout the UX, the only exception is not creating the OS fee consideration item.

PR #6097, opened 2023-03-07
chore: group third-party initializations

zzmp:
Groups 3P initializations (GA, Amplitude, and Sentry).
Also updates the Sentry initialization to include environment.


src/integrations.ts
```diff
@@ -0,0 +1,28 @@
+import 'components/analytics'
```
grabbou, line 1: nit: this is probably out of scope for this PR, however, a nice follow-up would be to move `components/analytics` file out of there, since it's not a component, but a utility.

grabbou (approved):

PR #6102, opened 2023-03-08
fix: dont fetch a route for wrap transactions

JFrankfurt:


zzmp (commented):
Does this gain us anything in the UI, or is it just an optimization to reduce backend calls? I'm not sure if it's worthwhile to add this complexity given that the widget is slated to be swapped in soon.

JFrankfurt:
This fixes a UI bug: https://www.notion.so/uniswaplabs/Token-price-quotes-impact-shouldn-t-render-for-wrapping-57e46736ab9d44a981a7d1a0332123f8?pvs=4#ce847178e1754f76a85fe73b54c45aaa

src/hooks/useBestTrade.ts
```diff
@@ -32,16 +36,28 @@ export function useBestTrade(
     200
   )
 
+  const wethContract = useWETHContract()
+  const isAWrapTransaction = useMemo(() => {
+    if (!wethContract || !chainId || !amountSpecified || !debouncedOtherCurrency) return undefined
```
zzmp, line 41: This should only return a boolean, and then you shouldn't need to cast `shouldGetTrade` to a boolean.

```suggestion
    if (!wethContract || !chainId || !amountSpecified || !debouncedOtherCurrency) return false
```

zzmp, line 41: Is it necessary to check if a `wethContract` exists, or is it enough to just check that the currency pair is native/weth?

src/hooks/useBestTrade.ts
```diff
@@ -32,16 +36,28 @@ export function useBestTrade(
...
+    const weth = chainId ? WRAPPED_NATIVE_CURRENCY[chainId] : undefined
+    return (
+      (amountSpecified.currency.isNative && weth?.equals(debouncedOtherCurrency)) ||
+      (debouncedOtherCurrency.isNative && weth?.equals(amountSpecified.currency))
+    )
+  }, [amountSpecified, chainId, debouncedOtherCurrency, wethContract])
+
+  const shouldGetTrade = Boolean(!isAWrapTransaction && isWindowVisible)
```
zzmp, line 49:
```suggestion
  const shouldGetTrade = !isAWrapTransaction && isWindowVisible
```

tinaszheng (approved):

PR #6103, opened 2023-03-08
feat: Migrate NFT Activity to GraphQL

cbachmeier:
Adds activity graphql endpoint, used on the Collection page activity tab and the asset Details page.


src/nft/components/collection/Activity.tsx
```diff
@@ -102,11 +106,47 @@ export const Activity = ({ contractAddress, rarityVerified, collectionName, chai
...
+      events: isNftGraphqlEnabled
+        ? gqlEventsData
+        : isSuccess
+        ? eventsData?.pages.map((page) => page.events).flat()
+        : undefined,
+      gatedHasNext: isNftGraphqlEnabled ? hasNext : hasNextPage,
+      gatedLoadMore: isNftGraphqlEnabled ? loadMore : fetchNextPage,
+      gatedLoading: isNftGraphqlEnabled ? loading : isLoading,
+      gatedIsLoadingMore: isNftGraphqlEnabled ? hasNext && gqlEventsData?.length : isFetchingNextPage,
+    }
```
JackShort, line 135: these boolean checks really don't need to be in a memo

JackShort: just the mapping

src/nft/types/collection/collection.ts
```diff
@@ -83,31 +85,29 @@ export interface TokenRarity {
...
-  fromAddress: string
+  eventType?: NftActivityType
+  marketplace?: Markets | string
+  fromAddress?: string
   toAddress?: string
   transactionHash?: string
-  orderHash?: string
-  orderStatus?: OrderStatus
-  price?: string
+  orderStatus?: QueryOrderStatus | OrderStatus
```
JackShort, line 107: will QueryOrderStatus not map directly to OrderStatus this whole double checking might not be necessary

src/nft/components/collection/ActivityCells.tsx
```diff
@@ -57,16 +59,18 @@ const AddressLink = styled(ExternalLink)`
   }
 `
 
-const formatListingStatus = (status: OrderStatus): string => {
+const formatListingStatus = (status: QueryOrderStatus | OrderStatus): string => {
   switch (status) {
-    case OrderStatus.EXECUTED:
+    case QueryOrderStatus.EXECUTED || OrderStatus.Executed:
```
JackShort, line 64: I think you can make it just take in a string and match it against `OrderStatus` and it will evaluate correctly

src/nft/components/collection/ActivityCells.tsx
```diff
@@ -112,19 +117,19 @@ export const BuyCell = ({
 
   return (
     <Column display={{ sm: 'none', lg: 'flex' }} height="full" justifyContent="center" marginX="auto">
-      {event.eventType === ActivityEventType.Listing && event.orderStatus ? (
+      {event.eventType === NftActivityType.Listing && event.orderStatus ? (
         <Box
           as="button"
-          className={event.orderStatus === OrderStatus.VALID && isSelected ? styles.removeCell : styles.buyCell}
+          className={event.orderStatus === QueryOrderStatus.VALID && isSelected ? styles.removeCell : styles.buyCell}
```
JackShort, line 123: why change it from `OrderStatus` to `QueryOrderStatus`?

src/nft/components/collection/ActivityCells.tsx
```diff
@@ -203,23 +217,23 @@ export const PriceCell = ({ marketplace, price }: { marketplace?: Markets; price
 }
 
 interface EventCellProps {
-  eventType: ActivityEventType
+  eventType: NftActivityType
   eventTimestamp?: number
   eventTransactionHash?: string
   eventOnly?: boolean
-  price?: string
+  price?: string | number
```
JackShort, line 224: wondering if we can cast this to a string in the hook above just to make it easier to deal with. Since these values are somewhat large I find dealing with strings always better

src/nft/components/details/AssetDetails.tsx
```diff
@@ -365,13 +385,48 @@ export const AssetDetails = ({ asset, collection }: AssetDetailsProps) => {
...
+      activityTypes: Object.keys(activeFilters)
+        .map((key) => key as NftActivityType)
+        .filter((key) => activeFilters[key]),
+      address: contractAddress,
+      tokenId: token_id,
+    },
+    25
+  )
+
+  const { events, gatedHasNext, gatedLoadMore, gatedLoading, gatedSuccess } = useMemo(() => {
```
JackShort, line 405: I feel like since this hook is used across two pages it could actually be pulled out to its own

JackShort (approved):
works well a few questions

src/nft/types/collection/collection.ts
cbachmeier: No it doesn't map directly

src/nft/components/collection/ActivityCells.tsx
cbachmeier: QueryOrderStatus is only needed while we still have the old query method, once the flag is merged in this whole type can be deleted

src/nft/components/details/AssetDetails.tsx
cbachmeier: This is only needed during this interim state while we have both the old query and the new graphql query, once this flag is merged in we can deprecate this and only use the graphql data.

PR #6107, opened 2023-03-08
fix: update formatting of position prices

vm:
image

- Updates to use the transaction formatting for position information
- Formats the current price using conedison

### DIFF
--- a/src/pages/Pool/PositionPage.tsx
+++ b/src/pages/Pool/PositionPage.tsx
@@ -179,7 +179,7 @@ function CurrentPriceCard({
           <Trans>Current price</Trans>
         </ExtentsText>
         <ThemedText.DeprecatedMediumHeader textAlign="center">
-          {formatPrice(inverted ? pool.token1Price : pool.token0Price, NumberType.TokenNonTx)}{' '}
+          {formatPrice(inverted ? pool.token1Price : pool.token0Price, NumberType.TokenTx)}{' '}
         </ThemedText.DeprecatedMediumHeader>
         <ExtentsText>
           <Trans>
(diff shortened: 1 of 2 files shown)


JFrankfurt:
Is the trailing `xxx.00` intended in this case?

image

vm:
yea i think the rule says always have the trailing 2 decimals for tx formatting. we can discuss if that's bad, but seems reasonable in most cases to me

JFrankfurt:
This makes the mobile LP list pretty cramped with all the extra decimals. from the new `formatTickPrice` implementation. I'm pretty sure the correct formatting here is `TokenNonTx`. (see [Token quantities outside transactions](https://www.notion.so/uniswaplabs/Number-standards-fbb9f533f10e4e22820722c2f66d23c0?pvs=4#33d214dbe8b047fa9da0a7c12016bd54))

**current**
image

**proposed**
image

src/pages/Pool/PositionPage.tsx
```diff
@@ -178,7 +179,7 @@ function CurrentPriceCard({
           <Trans>Current price</Trans>
         </ExtentsText>
         <ThemedText.DeprecatedMediumHeader textAlign="center">
-          {(inverted ? pool.token1Price : pool.token0Price).toSignificant(6)}{' '}
+          {formatPrice(inverted ? pool.token1Price : pool.token0Price, NumberType.TokenTx)}{' '}
```
JFrankfurt, line 182: I bet using `formatTickPrice` would be a better experience here. The current implementation doesn't use `,` to separate thousands.

vm:
We want to use a tooltip to show the updated formatting instead.

PR #6114, opened 2023-03-09
fix: use `chain` query param properly when activating

vm:
Previously, if the connector was still activating while a chain param came in, we'd ignore the update. Now, we cache the value until the connecter has activated, then update the chain.

### DIFF
--- a/src/hooks/useSyncChainQuery.ts
+++ b/src/hooks/useSyncChainQuery.ts
@@ -27,12 +27,13 @@ export default function useSyncChainQuery() {
 
   const urlChainId = getParsedChainId(parsedQs)
   const previousUrlChainId = usePrevious(urlChainId)
-  const [nextChainId, setNextChainId] = useState<number | undefined>(undefined)
 
   const selectChain = useSelectChain()
 
   // Can't use `usePrevious` because `chainId` can be undefined while activating.
   const [previousChainId, setPreviousChainId] = useState<number | undefined>(undefined)
+  const [nextChainId, setNextChainId] = useState<number | undefined>(undefined)
+
   useEffect(() => {
     if (chainId && chainId !== previousChainId) {
       setPreviousChainId(chainId)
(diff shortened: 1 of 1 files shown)


src/hooks/useSyncChainQuery.ts
```diff
@@ -27,6 +27,7 @@ export default function useSyncChainQuery() {
 
   const urlChainId = getParsedChainId(parsedQs)
   const previousUrlChainId = usePrevious(urlChainId)
+  const [nextChainId, setNextChainId] = useState<number | undefined>(undefined)
```
JFrankfurt, line 30: nit: `nextUrlChainId` to be clear about the relationship

JFrankfurt, line 30: why do we have a current, next, and previous?

vm: sorta confusing but open to better variable names. `urlChainId` is what's in the query at the time. `previous` is what allows us to check if it changed. `next` is what we're saying it should update the actual chainId to when ready.

vm: it's really not the `nextUrlChainId` though, it's just updating the chain id to what was in the url param

JFrankfurt: maybe move it down to be defined alongside `previousChainId`? putting it here makes it look like it might have been misnamed.

vm: good point

src/hooks/useSyncChainQuery.ts
```diff
@@ -40,14 +41,20 @@ export default function useSyncChainQuery() {
...
 
   return useEffect(() => {
-    if (chainQueryManuallyUpdated) {
+    if (nextChainId && isActive) {
       // If the query param changed, and the chain didn't change, then activate the new chain
-      selectChain(urlChainId)
+      selectChain(nextChainId)
       searchParams.delete('chain')
       setSearchParams(searchParams)
+      setNextChainId(undefined)
```
JFrankfurt, line 58: wondering if this would work better, since the chain switch might be rejected or unsupported and we wouldn't want to update the url in that case
```suggestion
      selectChain(nextChainId).then(() => {
        searchParams.delete('chain')
        setSearchParams(searchParams)
        setNextChainId(undefined)
      })
```

JFrankfurt (approved):

PR #6117, opened 2023-03-09
build: ignore types and hooks and update test-size

zzmp:


scripts/test-size.js
```diff
@@ -16,10 +16,10 @@ try {
 }
 
 // The last recorded size for these assets, as reported by `yarn build`.
-const MAX_SIZE_MAIN_KB = 361.36
+const LAST_SIZE_MAIN_KB = 364
```
vm, line 19: do this in a diff pr. tbh whenever i see "and" in a pr title i suggest splitting

vm (approved):

PR #6123, opened 2023-03-10
fix: Revert "fix: Revert "fix: use `chain` query param properly when activating""

vm:
Reverts Uniswap/interface#6122


vm:
Seems this option causes a race condition, and the test failure was legitimate. I will make a JIRA task and we can consider coming back to it.

PR #6126, opened 2023-03-10
fix: check if supported chainId before calling universal-router-sdk

vm:

- feat: A new feature
- fix: A bug fix

Example commit messages:

- feat: adds support for gnosis safe wallet
- fix: removes a polling memory leak
- chore: bumps redux version

Other things to note:

### DIFF
--- a/src/pages/Swap/index.tsx
+++ b/src/pages/Swap/index.tsx
@@ -307,7 +307,7 @@ export default function Swap({ className }: { className?: string }) {
       (parsedAmounts[Field.INPUT]?.currency.isToken
         ? (parsedAmounts[Field.INPUT] as CurrencyAmount<Token>)
         : undefined),
-    chainId ? UNIVERSAL_ROUTER_ADDRESS(chainId) : undefined
+    isSupportedChain(chainId) ? UNIVERSAL_ROUTER_ADDRESS(chainId) : undefined
   )
   const isApprovalLoading = allowance.state === AllowanceState.REQUIRED && allowance.isApprovalLoading
   const [isAllowancePending, setIsAllowancePending] = useState(false)


tinaszheng (approved):
this change lgtm, but this kind of begs a bigger question of how our interface handles unsupported chain ids in general. i think doing the `isSupportedChainId` check in individual places like this PR seems quite fickle. ideally, we should be:

- using `SupportedChainId` everywhere as our chain id to catch these issues at compile type, where mainnet is a fallback if the user isn't connected
- ideally ^ is something web3-react can handle for us - since we already pass in a list of supported chain ids, they should ideally be able to separate the concepts of wallet chain id vs network chain id on the page
- change universal router sdk to only accept sdk-core/SupportedChainId for universal router address lookup. this would've prevented this bug

tinaszheng:
@vm yes, my comment is a suggestion to how to "do something about that"

vm:
1) Unclear if we could catch this at compile time? We can't guarantee that only a SupportedChainId is set from the wallet side AFAIK. I think the check would still have to happen somewhere closer to the top level, and have conditional logic in the component layer for empty state stuff was mostly my point.
2) If MetaMask says the network is Sepolia, would that be the wallet chain id? What would the network chain id be in that case? What happens in a multichain world?
3) Only concern with this one is we'd need to then convert from interface's SupportedChainId to sdk-core's SupportedChainId in a few places. Doable and I do think that'd at least force the user to consider unsupported chains earlier.

tinaszheng:
I'm not saying that you're guaranteeing that a SupportedChainId is set from the wallet. I'm saying that the interface or web3-react should handle catching the case where you're setting an unsupported chain at a higher level, and then fallback to mainnet. This was the previous behavior from v6. If you're only passing around a `SupportedChainId`, then you wouldn't need to do the `isSupportedChainId` check in isolated areas of the codebase.

We could absolutely have caught the compile time check if UNIVERSAL_ROUTER_ADDRESS only accepted a `SupportedChainId` as its chain input. Then the ts compiler would complain that you're passing in a generic `number`, and would have required us to think more carefully about that case.

> If MetaMask says the network is Sepolia, would that be the wallet chain id? What would the network chain id be in that case? What happens in a multichain world?

"If Metamask says the network is Sepolia" that pretty much implies the wallet's network is currently on sepolia? The network provider chain id could be mainnet if sepolia is unsupported. This was the previous behavior. Not sure what changes theoretically here in a multichain world, that feels like tangential to the discussion.

vm:
https://github.com/Uniswap/web3-react/pull/567 here's the pr that changed the error logic. essentially, wanted to be able to catch errors inline through the promise rejection rather than listening async for an error value in the hook. that means we can do things like check if the chain switch failed without needing complicated logic.

however, it felt weird to then also have errors internal to web3-react, and also to enforce how clients would handle unsupported chains. instead, now clients have to implement that themselves. i think that's relatively fair since for example different connectors could support different chains as well.

PR #6131, opened 2023-03-11
fix: allow defaultTokens to reset state in Widget

just-toby:
fixes the "back" bug with TDP widget inputs and cleans up the logic in this hook a little bit

### test plan

main swap

https://user-images.githubusercontent.com/66155195/224459487-1a0bd30b-c270-4154-9c58-cacb02875f5f.mov

TDP and "back" bug fixed

https://user-images.githubusercontent.com/66155195/224459534-19df1d0f-4463-4cbd-8af6-fb60878800cb.mov

### DIFF
--- a/src/components/Widget/inputs.tsx
+++ b/src/components/Widget/inputs.tsx
@@ -18,6 +18,25 @@ function missingDefaultToken(tokens: SwapTokens) {
   return !tokens[Field.INPUT]?.equals(tokens.default) && !tokens[Field.OUTPUT]?.equals(tokens.default)
 }
 
+function currenciesEqual(a: Currency | undefined, b: Currency | undefined) {
+  if (a && b) {
+    return a.equals(b)
+  } else {
+    return !a && !b
+  }
+}
+
+function tokensEqual(a: SwapTokens | undefined, b: SwapTokens | undefined) {
+  if (!a || !b) {
+    return !a && !b
+  }
+  return (
+    currenciesEqual(a[Field.INPUT], b[Field.INPUT]) &&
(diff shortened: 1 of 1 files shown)


src/components/Widget/inputs.tsx
```diff
@@ -37,18 +56,21 @@ export function useSyncWidgetInputs({
 
   const [type, setType] = useState<SwapValue['type']>(TradeType.EXACT_INPUT)
   const [amount, setAmount] = useState<SwapValue['amount']>(EMPTY_AMOUNT)
-  const [tokens, setTokens] = useState<SwapTokens>(defaultTokens)
+  const [tokens, setTokens] = useState<SwapTokens>({
+    ...defaultTokens,
+    [Field.OUTPUT]: defaultTokens[Field.OUTPUT] ?? defaultTokens.default,
+  })
```
just-toby, line 62: set the initial value here the classic way, rather than relying on the hook below

src/components/Widget/inputs.tsx
```diff
@@ -37,18 +56,21 @@ export function useSyncWidgetInputs({
...
 
+  // The most recent set of defaults, which can be used to check when the defaults are actually changing.
+  const baseTokens = usePrevious(defaultTokens)
   useEffect(() => {
-    if (!tokens[Field.INPUT] && !tokens[Field.OUTPUT]) {
+    if (!tokensEqual(baseTokens, defaultTokens)) {
       setTokens({
```
just-toby, line 68: if the defaults have changed, then we want to reset the `tokens` value

vm (approved):

PR #6132, opened 2023-03-11
feat: Use ETH based pricing instead of USDC based pricing for quote USD values

tinaszheng:
[will add more in the morning]

We were previously assuming USDC = $1 and routing all pricing through a token <-> USDC swap. That is sadgely no longer the case. This new algo does a route through ETH:

Token -> ETH -> USD where the ETH -> USD quote uses our gql endpoint to fetch ETH price.

image

### DIFF
--- a/src/components/swap/TradePrice.tsx
+++ b/src/components/swap/TradePrice.tsx
@@ -1,8 +1,8 @@
 import { Trans } from '@lingui/macro'
 import { formatNumber, NumberType } from '@uniswap/conedison/format'
-import { Currency, CurrencyAmount, Price } from '@uniswap/sdk-core'
-import { parseUnits } from 'ethers/lib/utils'
+import { Currency, Price } from '@uniswap/sdk-core'
 import { useUSDPrice } from 'hooks/useUSDPrice'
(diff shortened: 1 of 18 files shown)


JFrankfurt:
image

src/components/swap/TradePrice.tsx
```diff
@@ -27,10 +29,18 @@ const StyledPriceContainer = styled.button`
   user-select: text;
 `
 
+function getOneCurrencyRawAmount(currency: Currency): string {
+  return parseUnits('1', currency.decimals).toString()
+}
```
JFrankfurt, line 34: this might be useful to you here: https://github.com/Uniswap/interface/blob/main/src/lib/utils/tryParseCurrencyAmount.ts

JFrankfurt (approved):
lgtm overall

1. I'd like to refactor some of the `CurrencyAmount` -> `number` areas to make it extra clear that the vars are for presentation only, not downstream future calculation. (especially concerned about future usage in `useSwapCallback`) Maybe we pass in an optional `analyticsCallback` prop instead?
2. We should discuss adding a `FiatCurrency` entity to the sdk or conedison

PR #6134, opened 2023-03-11
fix: Use non-subgraph fields for calculating USD prices on explore and token details pages

tinaszheng:
Update our gql queries to call `token.project.markets.x` instead of `token.market.x` because the pricing is more accurate there. `token.market.x` refers to our subgraph which hardcodes usdc as the USD indicator


mzywang (approved):

ianlapham (approved):

matteenm (approved):

cartcrom (changes requested):
I think we're missing just one price update, in RecentlySearched.ts

cartcrom (approved):

PR #6138, opened 2023-03-13
fix: fix overflowing lp ui

lynnshaoyu:
https://uniswaplabs.atlassian.net/browse/WEB-2138?atlOrigin=eyJpIjoiOWVkMDU1NTFhM2QwNDg3MDg3NzZhN2NiN2ZiNzNmYmUiLCJwIjoiaiJ9

before:
Screen Shot 2023-03-12 at 11 34 50 PM

after:
Screen Shot 2023-03-12 at 11 31 30 PM

### DIFF
--- a/src/components/PositionListItem/index.tsx
+++ b/src/components/PositionListItem/index.tsx
@@ -77,8 +77,7 @@ const DoubleArrow = styled.span`
   color: ${({ theme }) => theme.textTertiary};
 `
 
-const RangeText = styled.span`
-  display: block;
+const RangeText = styled.div`
   word-break: break-word;
   padding: 0.25rem 0.25rem;
   border-radius: 8px;
--- a/src/components/PositionListItem/index.tsx
+++ b/src/components/PositionListItem/index.tsx
@@ -78,6 +78,8 @@ const DoubleArrow = styled.span`
 `
 
 const RangeText = styled.span`
+  display: block;
+  word-break: break-word;
   padding: 0.25rem 0.25rem;
   border-radius: 8px;
 `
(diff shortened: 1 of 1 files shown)


src/components/PositionListItem/index.tsx
```diff
@@ -78,6 +78,8 @@ const DoubleArrow = styled.span`
 `
 
 const RangeText = styled.span`
+  display: block;
```
tinaszheng, line 81: you should probably use <div> as the container element if you want block behavior, since spans are by default meant to not impact positioning 🤔

if you use div, you get the display: block for free!

src/components/PositionListItem/index.tsx
```diff
@@ -78,6 +78,8 @@ const DoubleArrow = styled.span`
...
+  word-break: break-word;
   padding: 0.25rem 0.25rem;
   border-radius: 8px;
 `
```
JFrankfurt, line 85:
```suggestion
const RangeText = styled.div`
  word-break: break-word;
  padding: 0.25rem 0.25rem;
  border-radius: 8px;
`
```

src/components/PositionListItem/index.tsx
lynnshaoyu: sounds like a good idea to me! done

src/components/PositionListItem/index.tsx
lynnshaoyu: done ty!

JFrankfurt (approved):

tinaszheng (approved):

PR #6145, opened 2023-03-13
test: widget integration tests

just-toby:
adds e2e tests for basic inputs on the Swap Widget

the tests have some limitations because they use goerli tokens. most notable, the TDP test can't actually change the inputs.
we also can't switch chains in these

depends on https://github.com/Uniswap/interface/pull/6176

### DIFF
--- a/cypress/e2e/swap-widget.cy.ts
+++ b/cypress/e2e/swap-widget.cy.ts
@@ -38,15 +38,15 @@ describe('swap widget integration tests', () => {
 
   describe('widget on swap page', () => {
     it('should have the correct default input/output and token selection should work', () => {
-      cy.visit('/swap', { featureFlags: [FeatureFlag.swapWidget] })
+      cy.visit('/swap', { featureFlags: [FeatureFlag.swapWidget] }).then(() => {
+        verifyInputToken('ETH')
+        verifyOutputToken('Select token')
 
-      verifyInputToken('ETH')
-      verifyOutputToken('Select token')
+        selectOutputAndSwitch('UNI')
 
-      selectOutputAndSwitch('UNI')
-
-      verifyInputToken('UNI')
-      verifyOutputToken('ETH')
+        verifyInputToken('UNI')
+        verifyOutputToken('ETH')
+      })
     })
 
     it('should have the correct default input from URL params ', () => {
--- a/cypress/e2e/swap-widget.cy.ts
+++ b/cypress/e2e/swap-widget.cy.ts
@@ -1,14 +1,15 @@
 import { FeatureFlag } from '../../src/featureFlags/flags/featureFlags'
+import { getClassContainsSelector, getTestSelector } from '../utils'
 
 const UNI_GOERLI = '0x1f9840a85d5aF5bf1D1762F925BDADdC4201F984'
 
 describe('swap widget integration tests', () => {
   const verifyInputToken = (inputText: string) => {
-    cy.get('*[class*="TokenButtonRow"]').first().contains(inputText)
+    cy.get(getClassContainsSelector('TokenButtonRow')).first().contains(inputText)
   }
 
   const verifyOutputToken = (outputText: string) => {
-    cy.get('*[class*="TokenButtonRow"]').last().contains(outputText)
+    cy.get(getClassContainsSelector('TokenButtonRow')).last().contains(outputText)
   }
 
   const selectOutputAndSwitch = (outputText = 'UNI') => {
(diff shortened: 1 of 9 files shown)


cypress/e2e/swap-widget.cy.ts
```diff
@@ -0,0 +1,97 @@
...
+    // select token...
+    cy.contains(outputText).click()
+
+    cy.get('body')
+      .then(($body) => {
+        if ($body.find(getTestSelector('TokenSafetyWrapper')).length) {
+          return 'I understand'
+        }
+
+        return 'You pay' // Just click on a random element as a no-op
```
vm, line 27: you can just return undefined here imo and have it check for undefined in the next statement

cypress/e2e/swap-widget.cy.ts
```diff
@@ -0,0 +1,97 @@
...
+      })
+      .then((selector) => {
+        cy.contains(selector).click()
+      })
+
+    // token selector should close...
+    cy.contains('Search name or paste address').should('not.exist')
+
+    cy.get(getClassContainsSelector('ReverseButton')).first().click()
```
vm, line 36: why do you click this? should there be tests based on hitting it?

cypress/e2e/swap-widget.cy.ts
```diff
@@ -0,0 +1,97 @@
...
+describe('swap widget integration tests', () => {
+  const verifyInputToken = (inputText: string) => {
+    cy.get(getClassContainsSelector('TokenButtonRow')).first().contains(inputText)
+  }
+
+  const verifyOutputToken = (outputText: string) => {
+    cy.get(getClassContainsSelector('TokenButtonRow')).last().contains(outputText)
+  }
+
+  const selectOutputAndSwitch = (outputText = 'UNI') => {
```
vm, line 15: i don't think it's worth defaulting this param

cypress/e2e/swap-widget.cy.ts
just-toby: yes, all the tests that call this function expect it to switch the inputs

cypress/utils/index.ts
```diff
@@ -1,3 +1,5 @@
 export const getTestSelector = (selectorId: string) => `[data-testid=${selectorId}]`
 
 export const getTestSelectorStartsWith = (selectorId: string) => `[data-testid^=${selectorId}]`
+
+export const getClassContainsSelector = (selectorId: string) => `[class*=${selectorId}]`
```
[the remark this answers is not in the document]
just-toby: cypress actually discourages using classes as test selectors, so i don't think it does (maybe i just couldn't find it though)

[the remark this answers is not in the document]
vm: oh interesting. should we consider not using the classes then?

cypress/e2e/swap-widget.cy.ts
```diff
@@ -0,0 +1,99 @@
...
+    // token selector should close...
+    cy.contains('Search name or paste address').should('not.exist')
+
+    cy.get(getClassContainsSelector('ReverseButton')).first().click()
+  }
+
+  describe('widget on swap page', () => {
+    it('should have the correct default input/output and token selection should work', () => {
+      cy.visit('/swap', { featureFlags: [FeatureFlag.swapWidget] }).then(() => {
+        cy.wait('@eth_blockNumber')
```
vm, line 44: why is this only for the first test? should this be a precondition?

cypress/e2e/swap-widget.cy.ts
```diff
@@ -0,0 +1,99 @@
...
+
+        selectOutputAndSwitch('UNI')
+
+        verifyInputToken('UNI')
+        verifyOutputToken('ETH')
+      })
+    })
+
+    it('should have the correct default input from URL params ', () => {
+      cy.visit(`/swap?inputCurrency=${UNI_GOERLI}`, {
```
vm, line 56: can you put this in beforeEach?

cypress/e2e/swap-widget.cy.ts
vm: i might suggest adding an 'exists' case to make sure it actually reversed

cypress/e2e/swap-widget.cy.ts
just-toby: yeah i think it's fine to use this in every test - it's just a proxy for "page loaded" which seemed to matter for the first test (only in the GH action) --- although maybe this was just an issue with the flakiness yesterday

cypress/e2e/swap-widget.cy.ts
just-toby: i didn't use `beforeEach` for this because it's different for each test!

vm (approved):
i still think there's a way to abstract the feature flag part of this but don't feel strongly

cypress/e2e/swap-widget.cy.ts
just-toby: these are checked at the callsites

PR #6147, opened 2023-03-14
feat: trace swap.send

zzmp:
Adds a performance trace for `swap.send`
Also adds the necessary abstraction for collecting that trace, in `src/tracing`.

### DIFF
--- a/src/hooks/useUniversalRouter.ts
+++ b/src/hooks/useUniversalRouter.ts
@@ -10,13 +10,15 @@ import { FeeOptions, toHex } from '@uniswap/v3-sdk'
 import { useWeb3React } from '@web3-react/core'
 import { formatSwapSignedAnalyticsEventProperties } from 'lib/utils/analytics'
 import { useCallback } from 'react'
+import { trace } from 'tracing'
 import { calculateGasMargin } from 'utils/calculateGasMargin'
 import isZero from 'utils/isZero'
-import { swapErrorToUserReadableMessage } from 'utils/swapErrorToUserReadableMessage'
+import { didUserReject, swapErrorToUserReadableMessage } from 'utils/swapErrorToUserReadableMessage'
 
 import { PermitSignature } from './usePermitAllowance'
 
-class InvalidSwapError extends Error {}
+class ModifiedSwapError extends Error {}
+class GasEstimationError extends Error {}
 
 interface SwapOptions {
   slippageTolerance: Percent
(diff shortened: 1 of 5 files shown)


src/hooks/useUniversalRouter.ts
```diff
@@ -33,55 +35,63 @@ export function useUniversalRouterSwapCallback(
...
+              throw new ModifiedSwapError(
+                t`Your swap was modified through your wallet. If this was a mistake, please cancel immediately or risk losing your funds.`
+              )
+            }
+            return response
+          })
+        return response
+      } catch (swapError: unknown) {
+        if (swapError instanceof ModifiedSwapError) throw swapError
+        if (didUserReject(swapError)) setTraceStatus('cancelled')
```
zzmp, line 90: Cancellations are not failures, and must be treated differently.

src/utils/swapErrorToUserReadableMessage.tsx
```diff
@@ -1,51 +1,54 @@
...
     error?.code === 'ACTION_REJECTED' ||
     // For Rainbow :
     (reason?.match(/request/i) && reason?.match(/reject/i)) ||
     // For Frame:
     reason?.match(/declined/i) ||
     // For SafePal:
-    reason?.match(/cancelled by user/i) ||
+    reason?.match(/cancell?ed by user/i) ||
+    // For Trust:
+    reason?.match(/user cancell?ed/i) ||
```
lynnshaoyu, line 25: i think this is a typo, `cancell?ed ` should be `cancelled`

zzmp: This covers both spellings of `cancelled`, as Trust spells it with one `l` and I wanted the regex to be flexible.

src/tracing/trace.ts
```diff
@@ -0,0 +1,59 @@
...
+interface TraceCallbackOptions {
+  traceChild<T>(name: string, callback: TraceCallback<T>, metadata?: TraceMetadata): Promise<T>
+  setTraceData(key: string, value: unknown): void
+  setTraceTag(key: string, value: string | number | boolean): void
+  setTraceStatus(status: number | SpanStatusType): void
+  setTraceError(error: unknown): void
+}
+type TraceCallback<T> = (options: TraceCallbackOptions) => Promise<T>
+
+function traceTransaction(transaction?: Span) {
```
lynnshaoyu, line 22: can we add more documentation on this file since it's introducing important new methods that will likely be used in a lot of places? 

Including js docs on traceTransaction method, and each input param would be helpful for clarity.

src/utils/swapErrorToUserReadableMessage.tsx
```diff
@@ -1,51 +1,54 @@
...
   ) {
+    return true
+  }
+  return false
+}
+
+/**
+ * This is hacking out the revert reason from the ethers provider thrown error however it can.
+ * This object seems to be undocumented by ethers.
+ * @param error an error from the ethers provider
```
lynnshaoyu, line 39: can we change this to `@param error  - An error from the ethers provider.`

src/hooks/useUniversalRouter.ts
```diff
@@ -10,13 +10,15 @@ import { FeeOptions, toHex } from '@uniswap/v3-sdk'
...
+import { trace } from 'tracing'
 import { calculateGasMargin } from 'utils/calculateGasMargin'
 import isZero from 'utils/isZero'
-import { swapErrorToUserReadableMessage } from 'utils/swapErrorToUserReadableMessage'
+import { didUserReject, swapErrorToUserReadableMessage } from 'utils/swapErrorToUserReadableMessage'
 
 import { PermitSignature } from './usePermitAllowance'
 
-class InvalidSwapError extends Error {}
+class ModifiedSwapError extends Error {}
```
lynnshaoyu, line 20: can we add developer docs on what ModifiedSwapError is, and when does it occur

src/hooks/useUniversalRouter.ts
```diff
@@ -33,55 +35,72 @@ export function useUniversalRouterSwapCallback(
...
+            slippageTolerance: options.slippageTolerance,
+            deadlineOrPreviousBlockhash: options.deadline?.toString(),
+            inputTokenPermit: options.permit,
+            fee: options.feeOptions,
+          })
+          const tx = {
+            from: account,
+            to: UNIVERSAL_ROUTER_ADDRESS(chainId),
+            data,
+            // TODO: universal-router-sdk returns a non-hexlified value.
```
lynnshaoyu, line 58: what exactly is the TODO part of this comment?😅 like make universal-router-sdk return a hexlified value or something else?

src/tracing/trace.ts
```diff
@@ -0,0 +1,59 @@
+import * as Sentry from '@sentry/react'
+import { Span, SpanStatusType } from '@sentry/tracing'
+
+interface TraceMetadata {
+  /** Arbitrary data stored on a trace. */
+  data?: Record<string, unknown>
+  /** Indexed (ie searchable) tags associated with a trace. */
+  tags?: Record<string, string | number | boolean>
```
just-toby, line 8: if we're allowing these to be anything, can that cause difficulties on the backend later when we have a lot of unmanaged values? i know it will "work", but we chose to type / constrain enum values in `analytics` for a reason - assuming it makes the data set easier to work with on the backend?

src/tracing/trace.ts
```diff
@@ -0,0 +1,59 @@
...
+  setTraceError(error: unknown): void
+}
+type TraceCallback<T> = (options: TraceCallbackOptions) => Promise<T>
+
+function traceTransaction(transaction?: Span) {
+  const traceChild = <T>(name: string, callback: TraceCallback<T>, metadata?: TraceMetadata) => {
+    const child = transaction?.startChild({ ...metadata, op: name })
+    return traceTransaction(child)(name, callback)
+  }
+  const setTraceData = (key: string, value: unknown) => void transaction?.setData(key, value)
```
just-toby, line 27: subjective nit: braces make these easier to read

src/hooks/useUniversalRouter.ts
just-toby: oh this isn't part of this PR, it's just moved around

just-toby: this is why TODOs should have usernames attached 😆

just-toby (approved):

src/utils/swapErrorToUserReadableMessage.tsx
zzmp: yes, but note that both are valid (see https://jsdoc.app/tags-param.html) and will be parsed the same:

image

src/hooks/useUniversalRouter.ts
zzmp: yes, I'd call this a bug in universal-router-sdk

src/tracing/trace.ts
zzmp: I think this is a good point, but can be added as a constraint later.

src/tracing/trace.ts
zzmp: I have a strong preference for one-liners (probably from C++), so I often can't tell. Adding brackets :)

src/tracing/trace.ts
zzmp: constrained through a local typing. good idea!

src/tracing/trace.ts
```diff
@@ -0,0 +1,89 @@
+import * as Sentry from '@sentry/react'
+import { Span, SpanStatusType } from '@sentry/tracing'
+
+type TraceTags = {
+  widget: boolean
```
just-toby, line 5: i thought i saw `isWidget` in another PR

src/hooks/useUniversalRouter.ts
```diff
@@ -10,13 +10,31 @@ import { FeeOptions, toHex } from '@uniswap/v3-sdk'
...
-import { swapErrorToUserReadableMessage } from 'utils/swapErrorToUserReadableMessage'
+import { didUserReject, swapErrorToUserReadableMessage } from 'utils/swapErrorToUserReadableMessage'
 
 import { PermitSignature } from './usePermitAllowance'
 
-class InvalidSwapError extends Error {}
+/** Thrown when gas estimation fails. This class of error usually requires an emulator to determine the root cause. */
+class GasEstimationError extends Error {
+  constructor() {
+    super(t`Your swap is expected to fail.`)
```
just-toby, line 23: this is probably just copied from somewhere, but why don't we mention Gas as the reason for this error?

zzmp: AFAIK it's just a design choice, to hide the abstraction from the user.
We did switch it at one point (during Permit2), and it was requested that we switch it back.

just-toby (approved):

PR #6152, opened 2023-03-14
fix: don't fire off a graphql query on an empty search

cbachmeier:
In an effort to cut down on extraneous graphql queries, previously every pageload would do an empty query to SearchTokens with blank input. This skips all queries with empty input. We currently have collection search via graphql behind a feature flag so I've added this fix there as well.


cartcrom (approved):

cartcrom:
was just thinking abt this, ty for doing

PR #6154, opened 2023-03-14
fix: reset to defaultTokens when chain changes

just-toby:
these undefined inputs are not getting reset when we change chains, because of some recent changes to the token state management in the hook above.

https://user-images.githubusercontent.com/66155195/225092236-8f34043e-b3b2-4e33-92c6-53c2db51a8a0.mov


vm (approved):
starting to think `tokens` is a confusing name for the state here

PR #6159, opened 2023-03-15
feat: trace jsonrpc

zzmp:
Conditionally adds a performance trace for `jsonrpc.send`, if it occurs during another active trace.
Also adds the necessary abstraction for conditionally collecting trace, in `src/tracing` (`maybeTrace`).

### DIFF
--- a/src/tracing/jsonrpc.ts
+++ b/src/tracing/jsonrpc.ts
@@ -1,0 +1,25 @@
+/**
+ * Traces all JsonRpc requests as spans, reporting them if there is an active transaction.
+ * This is analagous to import("@sentry/tracing").BrowserTracingOptions.shouldCreateSpanForRequest.
+ *
+ * This is able to collect all requests because there is only one copy of @ethersproject/providers, and web3-react wraps
+ * any external (EIP-1193) provider in that prototype - overriding the prototype will override send for all instances.
+ */
+import { JsonRpcProvider } from '@ethersproject/providers'
+
+import { maybeTrace } from './trace'
+
+const JsonRpcProviderSend = JsonRpcProvider.prototype.send
+JsonRpcProvider.prototype.send = async function LoggingAwareSend(this, method, params) {
+  maybeTrace('jsonRpc.send', async ({ setTraceData }) => {
+    setTraceData('method', method)
+    if (method === 'eth_call') {
+      // Trim the calldata to the method hash to avoid recording large payloads and sensitive information.
+      const methodHash = (params[0].data as string).substring(0, 10)
+      setTraceData('params', { ...params, [0]: { ...params[0], data: methodHash } })
+    } else {
+      setTraceData('params', params)
+    }
+    return JsonRpcProviderSend.call(this, method, params)
+  })
+}
--- a/src/tracing/jsonrpc.ts
+++ b/src/tracing/jsonrpc.ts
@@ -9,17 +9,18 @@ import { JsonRpcProvider } from '@ethersproject/providers'
 
 import { maybeTrace } from './trace'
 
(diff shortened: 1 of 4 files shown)


src/tracing/jsonrpc.ts
```diff
@@ -0,0 +1,25 @@
...
+ * This is analagous to import("@sentry/tracing").BrowserTracingOptions.shouldCreateSpanForRequest.
+ *
+ * This is able to collect all requests because there is only one copy of @ethersproject/providers, and web3-react wraps
+ * any external (EIP-1193) provider in that prototype - overriding the prototype will override send for all instances.
+ */
+import { JsonRpcProvider } from '@ethersproject/providers'
+
+import { maybeTrace } from './trace'
+
+const JsonRpcProviderSend = JsonRpcProvider.prototype.send
```
vm, line 12: lowercase here since it's not a class, it's just a method

src/tracing/jsonrpc.ts
```diff
@@ -0,0 +1,25 @@
...
+JsonRpcProvider.prototype.send = async function LoggingAwareSend(this, method, params) {
+  maybeTrace('jsonRpc.send', async ({ setTraceData }) => {
```
vm, line 14: capitalize Json here? not sure the naming pattern

src/tracing/jsonrpc.ts
```diff
@@ -0,0 +1,25 @@
...
+    setTraceData('method', method)
+    if (method === 'eth_call') {
+      // Trim the calldata to the method hash to avoid recording large payloads and sensitive information.
+      const methodHash = (params[0].data as string).substring(0, 10)
```
vm, line 18: why 10 specifically?

src/tracing/jsonrpc.ts
```diff
@@ -0,0 +1,25 @@
...
+      setTraceData('params', { ...params, [0]: { ...params[0], data: methodHash } })
```
vm, line 19: slightly confused what the `[0]` key thing is?

src/tracing/jsonrpc.ts
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * This is able to collect all requests because there is only one copy of @ethersproject/providers, and web3-react wraps
+ * any external (EIP-1193) provider in that prototype - overriding the prototype will override send for all instances.
+ */
+import { JsonRpcProvider } from '@ethersproject/providers'
+
+import { maybeTrace } from './trace'
+
+const JsonRpcProviderSend = JsonRpcProvider.prototype.send
+JsonRpcProvider.prototype.send = async function LoggingAwareSend(this, method, params) {
```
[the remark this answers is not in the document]
zzmp: This is meant to override `send`, so that any invocations that occur during an active trace are sent, so it can't be named something else.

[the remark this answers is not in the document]
vm: i meant to rename `LoggingAwareSend` or just avoid naming that function maybe

src/tracing/jsonrpc.ts
vm: okay that's good for me

src/tracing/jsonrpc.ts
zzmp: The method hash will be 10 chars. I've added that to the comment above.

src/tracing/jsonrpc.ts
zzmp: the calldata is included under the first param.

vm (approved):
IDK if this is really a feature fwiw. probably more like build?

naming stuff is all nuts

PR #6162, opened 2023-03-15
feat: link to token details from /pool/{}

vm:
This is a cherry-pick commit from the work related to https://uniswaplabs.atlassian.net/browse/WEB-3011

We want to link to a TDP link instead of an external explorer when possible. However, for chains that don't have support in the Uniswap interface we'll continue to link externally.

This also will open in a new tab regardless. That's to prevent the user from losing their existing page state.

https://user-images.githubusercontent.com/4899429/225652763-5a17e129-cdfb-40f5-b45c-8782bbd6fa98.mov

Test:
- Go to a position page on Mainnet, Arbitrum
- Click on one of the token links --  should be a TDP link
- Go to a position page on Goerli
- Click on one of the token links -- should be a scanner link

### DIFF
--- a/src/pages/Pool/PositionPage.tsx
+++ b/src/pages/Pool/PositionPage.tsx
@@ -20,6 +20,7 @@ import { RowBetween, RowFixed } from 'components/Row'
 import { Dots } from 'components/swap/styleds'
 import Toggle from 'components/Toggle'
 import TransactionConfirmationModal, { ConfirmationModalContent } from 'components/TransactionConfirmationModal'
+import { CHAIN_ID_TO_BACKEND_NAME } from 'graphql/data/util'
 import { useToken } from 'hooks/Tokens'
 import { useV3NFTPositionManagerContract } from 'hooks/useContract'
 import useIsTickAtLimit from 'hooks/useIsTickAtLimit'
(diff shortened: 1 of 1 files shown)


src/pages/Pool/PositionPage.tsx
```diff
@@ -49,6 +50,22 @@ import { calculateGasMargin } from '../../utils/calculateGasMargin'
...
+const getTokenLink = (chainId: SupportedChainId, address: string) => {
+  const chainName = CHAIN_ID_TO_NAME[chainId]
+  if (chainName) {
+    return `${window.location.hostname}/${chainName}/${address}`
+  } else {
+    return getExplorerLink(chainId, address, ExplorerDataType.TOKEN)
+  }
+}
+
+const CHAIN_ID_TO_NAME: Partial<Record<SupportedChainId, string>> = {
```
vm, line 62: @cartcrom I notice we have something similar in graphql data utils. Do you think it makes sense to use that, or create a separate mapping for just the frontend somewhere?

src/pages/Pool/PositionPage.tsx
```diff
@@ -49,6 +50,22 @@ import { calculateGasMargin } from '../../utils/calculateGasMargin'
 import { ExplorerDataType, getExplorerLink } from '../../utils/getExplorerLink'
 import { LoadingRows } from './styleds'
 
+const getTokenLink = (chainId: SupportedChainId, address: string) => {
```
vm, line 53: wondering if we should put this util somewhere that's shared

src/pages/Pool/PositionPage.tsx
just-toby: i've used existing maps without needing to create a new one - if the existing ones work for your use case i'd definitely suggest re-using

just-toby (approved):
accept to unblock, but you should probably dedup the chain name map

src/pages/Pool/PositionPage.tsx
vm: okay i'll switch to using that

PR #6172, opened 2023-03-16
fix: skip when address undefined

cartcrom:


src/graphql/thegraph/AllV3TicksQuery.ts
```diff
@@ -1,12 +1,9 @@
-import { useQuery } from '@apollo/client'
 import gql from 'graphql-tag'
-import { useMemo } from 'react'
 
 import { AllV3TicksQuery } from './__generated__/types-and-hooks'
-import { apolloClient } from './apollo'
 
 const query = gql`
-  query AllV3Ticks($poolAddress: String!, $skip: Int!) {
+  query AllV3Ticks($poolAddress: String, $skip: Int!) {
```
zzmp, line 6: If we're skipping when we don't have a poolAddress, do we need to change this to an optional parameter?

zzmp (approved):

src/graphql/thegraph/AllV3TicksQuery.ts
cartcrom: yes the input will still be validated

PR #6173, opened 2023-03-16
feat: support both /pool and /pools

vm:
We want to have the nav say Pools instead of Pool, and be /pools instead of /pool. This is so we are consistent in the Nav (ex: Tokens, NFTs are pluralized).

Plan:
- Support both /pool and /pools for a bit. This is so external links don't break.
- Change all internal links to use /pools. Can happen in tandem with ^.
- Add a server redirect in CloudFlare to switch external links from /pool to /pools.

After this we should check that /pool isn't being logged in Amplitude.

Finally:
- Remove /pool Routes in codebase (since they will not be hit anymore).

This PR handles the first 2 steps of this plan.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ For steps on local deployment, development, and code contribution, please see [C
 The Uniswap Interface supports swapping, adding liquidity, removing liquidity and migrating liquidity for Uniswap protocol V2.
 
 - Swap on Uniswap V2: <https://app.uniswap.org/#/swap?use=v2>
-- View V2 liquidity: <https://app.uniswap.org/#/pool/v2>
+- View V2 liquidity: <https://app.uniswap.org/#/pools/v2>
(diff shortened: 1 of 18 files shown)


zzmp (changes requested):
How is `/pool` still being supported?

Unless you already have the redirect set up in CF, you should not remove it from the interface. Also, should we move off of cloudflare (eg to IPFS), this will break. Could you include a new file of legacy redirects and add `/pool -> /pools` so it's a permanent redirect within the codebase as well?

zzmp (approved):
https://github.com/Uniswap/interface/pull/6173/files#diff-11b6c68fcd0989e7541fed72fed10b836738171fe1b5f1bb01721f51b3e5e622R238

vm:
We also agreed that using a CF redirect is better in this case because CloudFlare will be hit first regardless of which hosting solution we use.

PR #6175, opened 2023-03-16
fix: only report errors to sentry through ErrorBoundary

zzmp:
Omits the TryCatch and GlobalHandlers sentry integrations, which had been sending uncaught/ignored errors to sentry.
Some errors are intentionally ignored and/or caused by 3P wallet integrations, so this effectively spammed our sentry issues.
Instead, we should only be reporting errors which crash the app, which should be reported through the Sentry.ErrorBoundary.

### DIFF
--- a/src/components/ErrorBoundary/index.tsx
+++ b/src/components/ErrorBoundary/index.tsx
@@ -1,5 +1,4 @@
 import { Trans } from '@lingui/macro'
-import * as Sentry from '@sentry/react'
 import { sendAnalyticsEvent } from '@uniswap/analytics'
 import { SwapEventName } from '@uniswap/analytics-events'
 import { ButtonLight, SmallButtonPrimary } from 'components/Button'
@@ -9,6 +8,7 @@ import React, { PropsWithChildren, useState } from 'react'
 import { Copy } from 'react-feather'
 import { useLocation } from 'react-router-dom'
 import styled from 'styled-components/macro'
+import { TracingErrorBoundary } from 'tracing'
 import { isSentryEnabled } from 'utils/env'
 
 import { CopyToClipboard, ExternalLink, ThemedText } from '../../theme'
@@ -222,7 +222,7 @@ const updateServiceWorkerInBackground = async () => {
 export default function ErrorBoundary({ children }: PropsWithChildren): JSX.Element {
   const { pathname } = useLocation()
   return (
-    <Sentry.ErrorBoundary
+    <TracingErrorBoundary
       fallback={({ error, eventId }) => <Fallback error={error} eventId={eventId} />}
       beforeCapture={(scope) => {
         scope.setLevel('fatal')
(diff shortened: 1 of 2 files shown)


grabbou:
Isn't every uncaught exception going to crash the app or malformed the state?

> Instead, we should only be reporting errors which crash the app, which should be reported through the Sentry.ErrorBoundary.

I think of it as errors that happened inside ErrorBoundary. So, if there is an exception that happens outside of it (e.g. in the Root component), it will not get reported.

 Also, keep in mind that:
 
> Error boundaries do not catch errors for:
> - Event handlers ([learn more](https://reactjs.org/docs/error-boundaries.html#how-about-event-handlers))
> - Asynchronous code (e.g. setTimeout or requestAnimationFrame callbacks)
> - Server side rendering
> - Errors thrown in the error boundary itself (rather than its children)

That's why, specifically with 1 & 2 points in mind, we need to leave the default configuration as is. 

I think a better long-term and cleaner solution would be to handle the exceptions that we want to handle/do not consider critical where they happen, by doing try/catch + adjusting the UI/code if necessary.

zzmp:
> to handle the exceptions that we want to handle/do not consider critical where they happen

I think this is too big of an ask for the web team. There are too many async calls being made, and not all of them are even under our control (eg some of them go through 3P applications). It's a paradigm change to ask that every call to eg `fetch` or `send` now be wrapped in a try-catch, particularly for event handlers and asynchronous code (points 1 & 2).

I think if we add more comprehensive error catching, we'll see a lot of mundane errors triggering the ErrorBoundary, which is what I want to avoid with this PR.

grabbou:
> I think this is too big of an ask for the web team. There are too many async calls being made, and not all of them are even under our control (eg some of them go through 3P applications). It's a paradigm change to ask that every call to eg fetch, or send now be wrapped in a try-catch, particularly for event handlers and asynchronous code (points 1 & 2).

Yes, it is a change. Thanks for pointing it out, I wasn't aware of that principle and clearly assumed my personal approach with always explicitly handling all the errors. If the error happens while performing an asynchronous action, I expect the application to handle it properly.

My problem with this approach is a lack of clear understanding of what happens due to such an error. Is the UI going to be responsive? What if the Redux store enters a corrupted state that causes some follow-up incorrect actions?

Example: I unset the loading spinner after the fetch completes. Fetch fails. Spinner continues loading, and the button is "disabled".

It is good to be aware of this and I am going to follow-up on this in another discussion.

grabbou:
With that in mind, instead of globally disabling it, I would try to - worst case scenario - "ignore" the most popular ones and annotate with a comment. It is similar to React Native's `YellowBox.ignoreLogs([])`, which takes an array of warnings/exceptions that should not be reported by the UI. 

That way, we could be intentional about this and easily discover new dangerous patterns when they start eating up our Sentry quota again.

Also, if we're able to identify the most popular source of warnings, maybe we can try to provide proper error handling when it makes sense (e.g. in case the UI was corrupted otherwise). In case we don't have time, we can always leave it as a TODO, but at least be aware of certain suboptimal UI behavior.

PR #6190, opened 2023-03-20
fix: Revert "feat: trace jsonrpc (#6159)"

tinaszheng:
This reverts commit 6fee37c4b880db06cb998e71a450ed21a325c763.

Fixes an issue where users aren't able to import their own tokens because the multicall jsonrpc calls to get token details aren't being sent


src/tracing/jsonrpc.ts
```diff
@@ -1,26 +0,0 @@
...
- *
- * This is able to collect all requests because there is only one copy of @ethersproject/providers, and web3-react wraps
- * any external (EIP-1193) provider in that prototype - overriding the prototype will override send for all instances.
- */
-import { JsonRpcProvider } from '@ethersproject/providers'
-
-import { maybeTrace } from './trace'
-
-const jsonRpcProviderSend = JsonRpcProvider.prototype.send
-JsonRpcProvider.prototype.send = async function LoggingAwareSend(this, method, params) {
```
grabbou, line 13: Just 0.5 cents, isn't this line the reason for regression?

It's been a while since I worked with prototypes, but I would assume that this could should be:
```ts
onRpcProvider.prototype.send = async function LoggingAwareSend(method, params) {
```
in other words, no `this`. I am surprised ESLint didn't warn against using that name nor it was a reserved keyword.

Anyway, the result here is that we're then most likely calling `jsonRpcProviderSend` on an instance of whatever was the first argument, which is most likely `method` (string)?
```ts
 jsonRpcProviderSend.call(this, method, params)
```

PR #6191, opened 2023-03-20
fix: improve fetch-schema logging and platform support

JFrankfurt:
- uses npx instead of relying on the package already being present
- uses node's `fs` package instead of `tee` to write to disk
- logs error if there is one


just-toby (approved):
nice

PR #6193, opened 2023-03-20
feat: list 1155s

cbachmeier:
- Allow listing 1155s.
  - Disable the warning for 1155s where you can't list an item for higher if there's an existing sell order
- Handles case where creator royalties are not set
  - Updates tooltip and grid UX to match. Dropdown will show base as it's global

Screenshots:
Screen Shot 2023-03-17 at 23 58 50


src/nft/components/profile/list/ListingButton.tsx
```diff
@@ -71,7 +72,11 @@ export const ListingButton = ({ onClick }: { onClick: () => void }) => {
             !listing.overrideFloorPrice
           )
             listingsBelowFloor.push([asset, listing])
-          else if (asset.floor_sell_order_price && listing.price >= asset.floor_sell_order_price)
+          else if (
```
JackShort, line 75: i feel like this whole code block this is in should be moved to a function because its starting to get cluttered

src/nft/components/profile/list/utils.ts
```diff
@@ -183,14 +198,35 @@ export function useUpdateInputAndWarnings(
...
+      ? asset.basisPoints
+        ? LOOKS_RARE_CREATOR_BASIS_POINTS
+        : 0
+      : asset.basisPoints ?? 0
 
   return baseFee * 0.01
 }
+
+// OpenSea has a 0.5% fee for all assets that do not have a royalty set
+export const getMarketplaceFee = (listingMarket: ListingMarket, asset: WalletAsset) => {
```
JackShort, line 230: it might be better to just have an object that stores both the royalties and the basis points for each marketplace that way in a situation where its not just opensea or not just looksrare this doesn't get hectic

JackShort (approved):

src/nft/components/profile/list/ListingButton.tsx
cbachmeier: good call out, I moved the issue finding logic to its own helper fn

src/nft/components/profile/list/utils.ts
cbachmeier: OS is unique in that they will only charge a market fee if no royalty is set "to fight spam", so there would still need to be unique logic to handle this distinction. 

Current royalty wars are also making so this logic is constantly in flux so I think having these helper fns that we can update quickly as opposed to trying to make long term structural decisions is the best current route until the dust settles.

PR #6195, opened 2023-03-20
fix: link to lowercase chain name from position page

vm:
Previously, the backend chain name was used for linking purposes. However, this is not ideal because those values are uppercased. Similarly, they won't support any chains that aren't supported by the backend (ex: Optimism Goerli).

Instead, we use the existing CHAIN_IDS_TO_NAME mapping.

Test:
- Go to a position page on a chain which the graphql API supports (ex: mainnet) and click on a token (should take to token details with a lowercase chain in url.
- Go to a position page on a chain which the graphql API does not support (ex: goerli) and click on a token (should take to etherscan).

### DIFF
--- a/src/graphql/data/util.tsx
+++ b/src/graphql/data/util.tsx
@@ -93,20 +93,6 @@ const URL_CHAIN_PARAM_TO_BACKEND: { [key: string]: Chain } = {
   optimism: Chain.Optimism,
 }
 
-export const CHAIN_ID_TO_URL_NAME: Record<SupportedChainId, string> = {
-  [SupportedChainId.ARBITRUM_ONE]: 'arbitrum',
-  [SupportedChainId.ARBITRUM_GOERLI]: 'arbitrum_goerli',
-  [SupportedChainId.BNB]: 'bnb',
-  [SupportedChainId.CELO]: 'celo',
-  [SupportedChainId.CELO_ALFAJORES]: 'celo_alfajores',
-  [SupportedChainId.MAINNET]: 'ethereum',
-  [SupportedChainId.GOERLI]: 'goerli',
-  [SupportedChainId.OPTIMISM]: 'optimism',
-  [SupportedChainId.OPTIMISM_GOERLI]: 'optimism_goerli',
-  [SupportedChainId.POLYGON]: 'polygon',
-  [SupportedChainId.POLYGON_MUMBAI]: 'polygon_mumbai',
-}
-
 export function validateUrlChainParam(chainName: string | undefined) {
   return chainName && URL_CHAIN_PARAM_TO_BACKEND[chainName] ? URL_CHAIN_PARAM_TO_BACKEND[chainName] : Chain.Ethereum
 }
--- a/src/graphql/data/util.tsx
+++ b/src/graphql/data/util.tsx
@@ -93,6 +93,20 @@ const URL_CHAIN_PARAM_TO_BACKEND: { [key: string]: Chain } = {
   optimism: Chain.Optimism,
 }
 
+export const CHAIN_ID_TO_URL_NAME: Record<SupportedChainId, string> = {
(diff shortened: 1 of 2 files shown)


src/graphql/data/util.tsx
```diff
@@ -93,6 +93,20 @@ const URL_CHAIN_PARAM_TO_BACKEND: { [key: string]: Chain } = {
   optimism: Chain.Optimism,
 }
 
+export const CHAIN_ID_TO_URL_NAME: Record<SupportedChainId, string> = {
```
vm, line 96: this maybe should live in a different file also. it is starting to not deal with graphql specific things

vm: i will deal with moving this later once i migrate the token details page to also use this constant. lmk if you have a better idea for the naming!

lynnshaoyu (approved):

src/graphql/data/util.tsx
```diff
@@ -93,6 +93,20 @@ const URL_CHAIN_PARAM_TO_BACKEND: { [key: string]: Chain } = {
   optimism: Chain.Optimism,
 }
```
lynnshaoyu, line 95: could be helpful to add inline developer documentation here about how this is distinct from CHAIN_ID_TO_BACKEND_NAME (ie: has mappings for all chains that FE supports, which is a superset of BE chains)

lynnshaoyu (approved):
approved with suggestion for documentation

src/graphql/data/util.tsx
```diff
@@ -93,6 +93,20 @@ const URL_CHAIN_PARAM_TO_BACKEND: { [key: string]: Chain } = {
...
+export const CHAIN_ID_TO_URL_NAME: Record<SupportedChainId, string> = {
+  [SupportedChainId.ARBITRUM_ONE]: 'arbitrum',
+  [SupportedChainId.ARBITRUM_GOERLI]: 'arbitrum_goerli',
+  [SupportedChainId.BNB]: 'bnb',
+  [SupportedChainId.CELO]: 'celo',
+  [SupportedChainId.CELO_ALFAJORES]: 'celo_alfajores',
+  [SupportedChainId.MAINNET]: 'ethereum',
+  [SupportedChainId.GOERLI]: 'goerli',
+  [SupportedChainId.OPTIMISM]: 'optimism',
+  [SupportedChainId.OPTIMISM_GOERLI]: 'optimism_goerli',
```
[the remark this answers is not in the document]
grabbou: This is what it looks like right now:
```js
export const CHAIN_IDS_TO_NAMES = {
  [SupportedChainId.MAINNET]: 'mainnet',
  [SupportedChainId.GOERLI]: 'goerli',
  [SupportedChainId.POLYGON]: 'polygon',
  [SupportedChainId.POLYGON_MUMBAI]: 'polygon_mumbai',
  [SupportedChainId.CELO]: 'celo',
  [SupportedChainId.CELO_ALFAJORES]: 'celo_alfajores',
  [SupportedChainId.ARBITRUM_ONE]: 'arbitrum',
  [SupportedChainId.ARBITRUM_GOERLI]: 'arbitrum_goerli',
  [SupportedChainId.OPTIMISM]: 'optimism',
  [SupportedChainId.OPTIMISM_GOERLI]: 'optimism_goerli',
  [SupportedChainId.BNB]: 'bnb',
}
```

grabbou (changes requested):
Overall, the PR looks good, but I am requesting changes to clarify whether we need a separate mapping vs reuse existing one.

src/graphql/data/util.tsx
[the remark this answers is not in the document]
vm: great point, i did not notice we had this constant! i'll update the PR.

grabbou (approved):
LGTM

PR #6197, opened 2023-03-20
feat: don't cache recent search history

cbachmeier:
We updated our recent search history to re-query to update prices however we still allowed caching which could lead to outdated price history. This sets the query fetch-policy to 'no-cache' to always take the latest price info.

### DIFF
--- a/src/components/NavBar/RecentlySearchedAssets.ts
+++ b/src/components/NavBar/RecentlySearchedAssets.ts
@@ -53,6 +53,7 @@ export function useRecentlySearchedAssets() {
           chain: token.chain,
         })),
     },
+    fetchPolicy: 'no-cache',
   })
 
   const data = useMemo(() => {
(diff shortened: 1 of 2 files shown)


JackShort (approved):
lgtm

cartcrom (changes requested):
I'm not sure how this small fetchPolicy change would break this, but "recently searched" but recently search don't seem to be appearing anymore for me? see video below.

Also, a UX preference:
Upon reopening search with recently fetched tokens in local storage, display the tokens with a loading state for the price component, rather than showing the entire row as a loading state. I think this could be done pretty easily using the default `cache-and-network` w/ the `loading` var.

https://user-images.githubusercontent.com/39385577/227286252-209cc19c-2a6d-433d-841f-1f3d76367d4a.mov

just-toby:

PR #6199, opened 2023-03-20
fix: NPE when connector is undefined

just-toby:
this fixes a page crash that happens when `getConnection` returns undefined. this is happening to me because I selected `UNIWALLET` on a different build, but it's not available in the current branch/build.

### DIFF
--- a/src/hooks/useOrderedConnections.ts
+++ b/src/hooks/useOrderedConnections.ts
@@ -22,6 +22,6 @@ export default function useOrderedConnections() {
     // Add network connection last as it should be the fallback.
     orderedConnectionTypes.push(ConnectionType.NETWORK)
 
-    return orderedConnectionTypes.map(getConnection)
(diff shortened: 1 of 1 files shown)


vm (changes requested):
I don't follow. How could this ever happen on production? Or how can I reproduce this?

just-toby:
@vm to reproduce, i think you'd need to run the mgtm build locally, have "uniwallet" saved to the redux store as your wallet selection, then switch back to the current build and try to run it. i'm not sure if this can happen in production, since you shouldn't be able to select a wallet option if it's not available. i mostly made this PR in case there was an edge case i couldn't imagine - maybe related to un-shipping a wallet option someday in the future?

at the very least, this improves the dev workflow for those of us working in both repos

vm:
If an invalid wallet is selected in the redux store, I think we may be better off clearing out that state? Rather than adding the defensive logic here

PR #6205, opened 2023-03-21
Enhancing with Deadline Checks and Improved Transaction Status

boivlad:
We are excited to share our latest contribution, aiming to enhance the user experience and provide a clearer understanding of transaction statuses. With this update, we have incorporated deadline checks for five key transaction types: SWAP, ADD_LIQUIDITY_V3_POOL, ADD_LIQUIDITY_V2_POOL, MIGRATE_LIQUIDITY_V3, and REMOVE_LIQUIDITY_V3.

## Here's a summary of the changes we made:

1. Deadline Check: We have added a new feature that checks the deadline value for the aforementioned transaction types. Users will now be able to see whether their transaction deadline has expired while it is still pending. This will give them more clarity on the state of their transactions and help them make informed decisions.

2. Transaction Status: If the deadline has expired and the transaction is still not mined, the status will now be displayed as "Failed." This change will make it easier for users to quickly identify transactions that have not been executed within the specified deadline.

Our team believes that these enhancements will provide a more seamless and transparent experience for Uniswap users. We look forward to receiving your valuable feedback and continuing to collaborate on improving the Uniswap platform.


cmcewen:
@boivlad this is a great feature idea, thanks! could you add some tests to this PR validating the behavior you expect? i'll also ask for some design advice from our team since i think maybe adding a different state rather than failed might be more clear to the user

boivlad:
> @boivlad this is a great feature idea, thanks! could you add some tests to this PR validating the behavior you expect? i'll also ask for some design advice from our team since i think maybe adding a different state rather than failed might be more clear to the user

Hello, we have written a test and are waiting for design suggestions from your side.

boivlad:
@cmcewen Hi, do you have any updates on the design so that we can successfully implement improvements?

src/lib/hooks/transactions/updater.tsx
```diff
@@ -49,15 +52,27 @@ export default function Updater({ pendingTransactions, onCheck, onReceipt }: Upd
...
         () =>
           provider.getTransactionReceipt(hash).then((receipt) => {
-            if (receipt === null) {
+            if (receipt === null && blockTimestamp) {
+              if (blockTimestamp.gt(deadline)) {
+                onReceipt({
+                  chainId,
+                  hash,
+                  receipt: {
+                    ...(receipt as TransactionReceipt),
```
JFrankfurt, line 71: this is effectively `...(null as TransactionReceipt)`, right?

I think a better approach would be updating `getReceipt` to explicitly remove the transaction from the app transactions store instead of forcing a faked transaction to "confirm"

just-toby:
hi @boivlad , sorry for the delay in getting reviews here. if you still want to merge this PR, can you please do the following 
1. rebase or merge the PR onto the latest `main`
2. take a screen recording of the UI changes and add it to the PR description.

just-toby:
feel free to re-open the PR if still relevant, once updated

PR #6209, opened 2023-03-21
feat: preserve input currency on TDP when navigating

just-toby:
this preserves the input/output currency on the TDP when you replace the "default" (detailed) token.

behavior:
1. when navigating to TDP, the detailed/default token is the output and input is empty
2. you can reverse input/output, and select a token to trade in either direction
3. when you replace the detailed token, we automatically navigate to the TDP for the output in the current selections (this could be the newly selected if you replaced the output, or the other token if you replaced the input)
4. <This PR>  we add a query param to the destination URL allowing us to preserve the input token when navigating

this seemed like a simple solution, but there may be alternatives - if we used a global state variable (jotai atom) we could preserve the input without a URL change. the downside there would be the complexity of determining when to clear the value (or conversely, knowing when to use it i.e. after a navigation caused by the flow above).

### DIFF
--- a/src/graphql/data/util.tsx
+++ b/src/graphql/data/util.tsx
@@ -108,8 +108,18 @@ export const CHAIN_NAME_TO_CHAIN_ID: { [key: string]: SupportedChainId } = {
 
 export const BACKEND_CHAIN_NAMES: Chain[] = [Chain.Ethereum, Chain.Polygon, Chain.Optimism, Chain.Arbitrum, Chain.Celo]
 
-export function getTokenDetailsURL({ address, chain }: { address?: string | null; chain: Chain }) {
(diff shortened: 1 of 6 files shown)


src/graphql/data/util.tsx
```diff
@@ -108,8 +108,18 @@ export const CHAIN_NAME_TO_CHAIN_ID: { [key: string]: SupportedChainId } = {
...
+  chain,
+  inputAddress,
+}: {
+  address?: string | null
+  chain: Chain
+  inputAddress?: string | null
+}) {
+  return `/tokens/${chain.toLowerCase()}/${address ?? NATIVE_CHAIN_ID}${
+    inputAddress ? `?inputCurrency=${inputAddress}` : ''
+  }`
```
JFrankfurt, line 122: nit, move the logic out of the template string

JFrankfurt (approved):

PR #6212, opened 2023-03-21
feat: cypress test coverage reporting

just-toby:
adds instrumentation to the app for test runs, and enables coverage reporting in cypress.

pros:
- provides an ongoing guidance about where to focus our attention for new e2e tests (which are be more powerful than unit tests)
- more accurate reporting of exactly how much code is actually executing automatically in the test runs
- number go up

cons:
- e2e tests might execute a line of code, but they won't actually test functionality in specific ways like unit tests would. the only test is the high-level user flow. we might not want to consider these lines "covered" 
- performance of e2e test runs could be slightly worse (i haven't tested this yet) since the build has a bunch of instrumentation code now.


package.json
```diff
@@ -178,6 +179,7 @@
     "@web3-react/walletconnect": "8.1.2-beta.0",
     "array.prototype.flat": "^1.2.4",
     "array.prototype.flatmap": "^1.2.4",
+    "babel-plugin-istanbul": "^6.1.1",
```
JFrankfurt, line 182: what's up with this babel plugin?

just-toby: this is what actually does the instrumentation: https://docs.cypress.io/guides/tooling/code-coverage#Using-code-transpilation-pipeline

cmcewen (approved):
have you tested to see if this has any speed impact due to collecting code coverage? not sure how it works in e2es tbh

just-toby:
> have you tested to see if this has any speed impact due to collecting code coverage? not sure how it works in e2es tbh

the Github actions are not taking significantly longer, so the speed impact seems low

PR #6201, opened 2023-03-20
feat: add chain to position url

vm:
wip

PR #6214, opened 2023-03-22
feat: show error state when on invalid position

vm:
This cherry-picks one part of https://github.com/Uniswap/interface/pull/6201, which is a more complete solution for handling position pages across chains. However, we want to wait on that until we make sure that it fits with our multichain plans.

For now, this makes sure that we show an invalid state when there is no position information to render, instead of showing an infinite loading state.

To test this...
- Go to pools/364815 on mainnet to see the position render
- Switch chain to Polygon to see the error state

On main...:
- Go to pools/364815 on mainnet to see the position render
- Switch chain to Polygon to see an infinite loading state

Resolves https://uniswaplabs.atlassian.net/browse/WEB-3006 and resolves https://uniswaplabs.atlassian.net/browse/WEB-3010

### DIFF
--- a/src/pages/Pool/PositionPage.tsx
+++ b/src/pages/Pool/PositionPage.tsx
@@ -61,6 +61,14 @@ const getTokenLink = (chainId: SupportedChainId, address: string) => {
   }
 }
 
+const PositionPageButtonPrimary = styled(ButtonPrimary)`
+  width: 228px;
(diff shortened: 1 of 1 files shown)


JFrankfurt (approved):
nit: I think the title copy could be improved.

"Position <xxxxx> does not exist on <network_name>."

PR #6219, opened 2023-03-22
feat: warn when adding liquidity to an unowned LP position

JFrankfurt:


src/pages/AddLiquidity/index.tsx
```diff
@@ -548,6 +550,11 @@ export default function AddLiquidity() {
     }),
     [usdcValueCurrencyB]
   )
+
+  const owner = useSingleCallResult(tokenId ? positionManager : null, 'ownerOf', [tokenId]).result?.[0]
+  const ownsNFT = owner === account || existingPositionDetails?.operator === account
```
tinaszheng, line 555: worth maybe doing lowercase comparison here in case one of these doesn't come back as checksum format? do we have a util for comparing addresses maybe :3

tinaszheng (approved):

PR #6220, opened 2023-03-22
fix: update url regex to correctly match URLs

JFrankfurt:


src/utils/urlChecks.ts
```diff
@@ -1,10 +1,9 @@
 export function hasURL(str: string): boolean {
   const pattern = new RegExp(
-    '(http|https):\\/\\/' + // Match either "http" or "https" for the protocol
```
JFrankfurt, line 3: this was not optional and only specified http(s)

just-toby (approved):

PR #6216, opened 2023-03-22
feat: ✨🤑 Portfolios 🖼✨

cartcrom:
Replaces the wallet dropdown/modal with a drawer that contains a new way to view portfolio balances for tokens, NFTs, and LP positions.

PR #6228, opened 2023-03-22
feat: improved injected connection display logic

cartcrom:
returns checks for `isInjected` / `isMetaMask` & other utilities to being function calls, rather than checking static vars initialized on page load, in order to fix issue where wallets that don't immediately inject display as 'MetaMask' rather than 'Browser Wallet'

also tweaks display logic to ensure the follow the following:
* on a _known_ wallet browser (MM, CB, TrustWallet, etc) , display only the injected connector; while on an _unknown_ wallet browser, display injected options alongside wallet connect & coinbase
* display "Install MetaMask" rather than just "MetaMask" while on a non-injected desktop browser

### DIFF
--- a/src/components/WalletDropdown/UniwalletModal.tsx
+++ b/src/components/WalletDropdown/UniwalletModal.tsx
@@ -5,7 +5,7 @@ import { WalletConnect } from '@web3-react/walletconnect'
 import Column, { AutoColumn } from 'components/Column'
 import Modal from 'components/Modal'
 import { RowBetween } from 'components/Row'
-import { uniwalletConnectConnection } from 'connection'
+import { getUniwalletConnectConnection } from 'connection'
 import { UniwalletConnect } from 'connection/WalletConnect'
 import { QRCodeSVG } from 'qrcode.react'
 import { useCallback, useEffect, useState } from 'react'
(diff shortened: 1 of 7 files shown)


grabbou (changes requested):
I don't think assumptions in this PR are correct. 

TL;DR every connector is now a function and is evaluated each time `getConnections()` is called. 

While this may work in test environments (where we specifically re-execute that function), we cannot guarantee this will happen in production. The reason for this to work is probably the fact that React re-renders our tree quite often and this function is re-evaluated. 

However, adding a simple `useMemo` to this function itself (in the future, when someone wants to "optimise our performance") or accidentally executing this function inside a memoized component will yield incorrect results.

A better and more future-proof solution would be for `getConnections` to remain a static object. There is no need for it to be a function. All Web3React initialisations are lazy anyway and subject to bundle splitting.

Then, we should have a special hook, such as `useInjectedWallets` that uses something like `useEffect` internally to "watch" for any injections that can happen at a later stage and in case that happens - update its state, causing React to re-render all affected parties. This is also going to be important for future emigrations to Next.js, where we can't refer to `window` outside of `useEffect`, otherwise it is not possible to server-side render affected components.

I will be looking at the previous PRs related to this file and leaving some additional comments too (will cross-reference for context).

grabbou:
Also, note that this has additional performance implications every time `connector` is referenced in a hook or passed as a prop to a memoized component. The default behavior is to check for reference equality. Because we create the `connection` object each time (because it's a function), that check will return false and tell React to re-execute the hook (or re-render).

This is already caused to a certain degree by the previous PR (#6216) that converted static named exports to a function, but this increases that risk even more by adding nested functions for each connector. 

Unless we implement deep comparison (which is not the default and is easy to overlook), that can lead to unwanted and hard to debug side effects where hooks keep running every re-render.

tinaszheng:
@cartcrom i believe the prev PR may have caused the regression issue here: https://github.com/Uniswap/interface/issues/6242 😢

grabbou:
I am going to convert this PR to a draft for clarity. I believe it requires a fundamental rewrite based on previous comments, happy to follow up via DM on how to tackle this best. However, we may need to look into the underlying PR too to understand whether this issue has existed before or was only introduced after previous PR was merged.

PR #6232, opened 2023-03-22
feat: implement new designs for tx notifs

just-toby:
- update the "Failed to Switch Network" notification component to match the new design
- update the general transaction notification component to re-use Mini Portfolio code so they are consistent with design.
 

### test plan

Screenshot 2023-03-22 at 2 27 21 PM

image


src/components/WalletDropdown/MiniPortfolio/PortfolioRow.tsx
```diff
@@ -38,29 +35,21 @@ export default function PortfolioRow({
   setIsHover?: (b: boolean) => void
   onClick?: () => void
 }) {
-  const onHover = useMemo(
-    () =>
```
just-toby, line 42: this seems to be unused so i removed it

grabbou (commented):
Overall, the changes look like step in the right direction. My main concern is that - at least at this point - I have mixed feelings about using `MiniPortfolio` component in a `Popup`. 

My general rule of thumb is that as long as component starts depending on another components' internal components (like `TransactionPopup` using `ActivityTabStyledDescriptor`), they should either be in the single folder (as related to each other) or that `ActivityTabStyledDescriptor` be converted to a shared primitive that works for both.

I know that our codebase is far from following this approach, but just wanted to point that out to see if there are any quick wins that could bring us closer to that scenario.

Maybe this is the best we can do - which is fine.

src/components/Popups/TransactionPopup.tsx
```diff
@@ -1,46 +1,55 @@
...
   const tx = useTransaction(hash)
-  const theme = useContext(ThemeContext)
+  const tokens = useCombinedActiveList()
+  const success = Boolean(tx?.receipt && tx?.receipt.status === 1)
 
-  if (!tx) return null
-  const success = Boolean(tx.receipt && tx.receipt.status === 1)
+  const activity = tx ? parseLocalActivity(tx, chainId ?? 0, tokens) : null
+  const { ENSName } = useENSName(activity?.otherAccount)
+  const explorerUrl = getExplorerLink(chainId ?? 0, hash, ExplorerDataType.TRANSACTION)
```
tinaszheng, line 24: here too

src/components/Popups/TransactionPopup.tsx
```diff
@@ -1,46 +1,55 @@
...
+      }
+      title={<ThemedText.SubHeader fontWeight={500}>{activity.title}</ThemedText.SubHeader>}
+      descriptor={
+        <ActivityRowStyledDescriptor color="textSecondary">
+          {activity.descriptor}
+          {ENSName ?? activity.otherAccount}
+        </ActivityRowStyledDescriptor>
+      }
+      onClick={() => window.open(explorerUrl, '_blank')}
+      right={undefined}
```
tinaszheng, line 52:
```suggestion

```

src/components/Popups/TransactionPopup.tsx
```diff
@@ -1,46 +1,55 @@
...
+        )
+      }
+      title={<ThemedText.SubHeader fontWeight={500}>{activity.title}</ThemedText.SubHeader>}
+      descriptor={
+        <ActivityRowStyledDescriptor color="textSecondary">
+          {activity.descriptor}
+          {ENSName ?? activity.otherAccount}
+        </ActivityRowStyledDescriptor>
+      }
+      onClick={() => window.open(explorerUrl, '_blank')}
```
tinaszheng, line 51: ooof to `onClick` handlers being on divs - this isn't going to work for accessibility / keyboard tabbing unless you also manually define  tab indices and focus states

tinaszheng: you also aren't going to get link previews

src/components/WalletDropdown/MiniPortfolio/Activity/index.tsx
```diff
@@ -101,7 +101,7 @@ function combineActivities(localMap: ActivityMap = {}, remoteMap: ActivityMap =
 
 const lastFetchedAtom = atom<number | undefined>(0)
 
-export default function ActivityTab({ account }: { account: string }) {
+export function ActivityTab({ account }: { account: string }) {
```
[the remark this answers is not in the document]
tinaszheng: yeah agreed pls move this component into its own file!

src/hooks/useSelectChain.ts
```diff
@@ -25,7 +25,12 @@ export default function useSelectChain() {
         console.error('Failed to switch networks', error)
 
         dispatch(updateConnectionError({ connectionType, error: error.message }))
-        dispatch(addPopup({ content: { failedSwitchNetwork: targetChain }, key: `failed-network-switch` }))
+        dispatch(
+          addPopup({
+            content: { failedSwitchNetwork: targetChain },
+            key: `failed-network-switch`,
```
tinaszheng, line 31: are these just raw strings and not enums? 
```suggestion
            key: "failed-network-switch",
```

tinaszheng (changes requested):

src/components/Popups/TransactionPopup.tsx
just-toby: i think this is a problem with the MP row that needs to be addressed outside of this PR - also the tabbing matters more in the sidebar than in the toast probably.

src/components/WalletDropdown/MiniPortfolio/Activity/index.tsx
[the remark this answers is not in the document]
just-toby: i separated these into two files

just-toby (commented):
@grabbou re: your comment about relying on components that should be internal to MP - this was totally unnecessary and i updated to stop doing this

src/components/Popups/FailedNetworkSwitchPopup.tsx
```diff
@@ -12,23 +12,26 @@ const RowNoFlex = styled(AutoRow)`
...
-        <RowNoFlex style={{ alignItems: 'center' }}>
-          <div style={{ paddingRight: 13 }}>
-            <AlertTriangle color={theme.accentWarning} size={24} display="flex" />
-          </div>
-          <ThemedText.SubHeader>
-            <Trans>Failed to switch networks</Trans>
-          </ThemedText.SubHeader>
-        </RowNoFlex>
+      <PopupAlertTriangle />
+      <AutoColumn gap="sm" margin="0 12px 0">
```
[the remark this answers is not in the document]
just-toby: removed this

src/components/Popups/FailedNetworkSwitchPopup.tsx
```diff
@@ -12,26 +12,33 @@ const RowNoFlex = styled(AutoRow)`
   flex-wrap: nowrap;
 `
 
+const ColumnContainer = styled(AutoColumn)`
+  margin: 0 12px 0;
+`
+
+export const PopupAlertTriangle = styled(AlertTriangleFilled)`
+  flex-shrink: 0;
+  color: ${({ theme }) => (theme.darkMode ? theme.backgroundSurface : 'white')};
```
JFrankfurt, line 21: this is a pretty uncommon situation--should double check w/ design that there's a theme exception here

src/components/WalletDropdown/MiniPortfolio/index.tsx
```diff
@@ -20,6 +21,12 @@ const Wrapper = styled(Column)`
   flex-direction: column;
   height: 100%;
   gap: 12px;
+
+  ${PortfolioRowWrapper} {
+    &:hover {
+      background: ${({ theme }) => theme.hoverDefault};
```
JFrankfurt, line 27: confirm the `background` value change will trigger the `background-color` transition on `PortfolioRowWrapper`

just-toby: yep it still works

src/components/Popups/FailedNetworkSwitchPopup.tsx
just-toby: you made me realize i didn't need this line anymore thanks

src/components/Popups/FailedNetworkSwitchPopup.tsx
```diff
@@ -12,26 +12,32 @@ const RowNoFlex = styled(AutoRow)`
   flex-wrap: nowrap;
 `
 
+const ColumnContainer = styled(AutoColumn)`
+  margin: 0 12px 0;
```
tinaszheng, line 16: do you just want margin 12px on the right/left here? 
```suggestion
  margin: 0 12px;
```

src/components/Popups/FailedNetworkSwitchPopup.tsx
```diff
@@ -12,26 +12,32 @@ const RowNoFlex = styled(AutoRow)`
...
+`
+
+export const PopupAlertTriangle = styled(AlertTriangleFilled)`
+  flex-shrink: 0;
+  width: 32px;
+  height: 32px;
+  margin-right: 12px;
```
tinaszheng, line 23: prefer putting `gap` on the parent container to putting margin on child containers 

my general opinion is that layout should be determined by parent and not individually on the child components

src/components/Popups/TransactionPopup.tsx
```diff
@@ -1,46 +1,65 @@
...
+const Descriptor = styled(ThemedText.BodySmall)`
+  ${EllipsisStyle}
 `
 
+function TransactionPopupContent({ tx, chainId }: { tx: TransactionDetails; chainId: number }) {
+  const success = Boolean(tx?.receipt && tx?.receipt.status === 1)
+  const tokens = useCombinedActiveList()
+  const activity = tx ? parseLocalActivity(tx, chainId, tokens) : null
+  const { ENSName } = useENSName(activity?.otherAccount)
+  const explorerUrl = getExplorerLink(chainId, tx.hash, ExplorerDataType.TRANSACTION)
```
tinaszheng, line 25: you could move this line after line 27

src/components/WalletDropdown/MiniPortfolio/Activity/ActivityRow.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { Activity } from './types'
+
+const ActivityRowDescriptor = styled(ThemedText.BodySmall)`
+  color: ${({ theme }) => theme.textSecondary};
+  ${EllipsisStyle}
+`
+
+const StyledTimestamp = styled(ThemedText.Caption)`
+  color: ${({ theme }) => theme.textSecondary};
+  font-variant: small;
```
tinaszheng, line 22: i havent seen `font-variant` used before - what does this do?

src/components/WalletDropdown/MiniPortfolio/Activity/ActivityRow.tsx
```diff
@@ -0,0 +1,59 @@
...
+  font-feature-settings: 'tnum' on, 'lnum' on, 'ss02' on;
+`
+
+export function ActivityRow({ activity }: { activity: Activity }) {
+  const { chainId, status, title, descriptor, logos, otherAccount, currencies } = activity
+  const { ENSName } = useENSName(otherAccount)
+
+  const explorerUrl = getExplorerLink(activity.chainId, activity.hash, ExplorerDataType.TRANSACTION)
+  const timeSince = useTimeSince(activity.timestamp)
```
tinaszheng, line 31: nit: switch these so that the two hooks are together at the top

src/components/WalletDropdown/MiniPortfolio/Activity/ActivityRow.tsx
```diff
@@ -0,0 +1,59 @@
...
+`
+
+const StyledTimestamp = styled(ThemedText.Caption)`
+  color: ${({ theme }) => theme.textSecondary};
+  font-variant: small;
+  font-feature-settings: 'tnum' on, 'lnum' on, 'ss02' on;
+`
+
+export function ActivityRow({ activity }: { activity: Activity }) {
+  const { chainId, status, title, descriptor, logos, otherAccount, currencies } = activity
```
tinaszheng, line 27:
```suggestion
export function ActivityRow({ activity: { chainId, status, title, descriptor, logos, otherAccount, currencies, hash, timestamp }  }: { activity: Activity }) {
```

src/components/WalletDropdown/MiniPortfolio/Activity/ActivityRow.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const { ENSName } = useENSName(otherAccount)
+
+  const explorerUrl = getExplorerLink(activity.chainId, activity.hash, ExplorerDataType.TRANSACTION)
```
tinaszheng, line 30:
```suggestion
  const explorerUrl = getExplorerLink(chainId, hash, ExplorerDataType.TRANSACTION)
```

src/components/WalletDropdown/MiniPortfolio/Activity/ActivityRow.tsx
tinaszheng, line 31:
```suggestion
  const timeSince = useTimeSince(timestamp)
```

tinaszheng (changes requested):
just some small things!

src/components/Popups/TransactionPopup.tsx
```diff
@@ -1,46 +1,66 @@
...
+const Descriptor = styled(ThemedText.BodySmall)`
+  ${EllipsisStyle}
 `
 
-export default function TransactionPopup({ hash }: { hash: string }) {
-  const { chainId } = useWeb3React()
+function TransactionPopupContent({ tx, chainId }: { tx: TransactionDetails; chainId: number }) {
+  const success = Boolean(tx?.receipt && tx?.receipt.status === 1)
+  const tokens = useCombinedActiveList()
+  const activity = tx ? parseLocalActivity(tx, chainId, tokens) : null
```
tinaszheng, line 23: isnt `tx` always defined here

src/components/Popups/TransactionPopup.tsx
```diff
@@ -1,46 +1,66 @@
...
+import { PopupAlertTriangle } from './FailedNetworkSwitchPopup'
+
+const Descriptor = styled(ThemedText.BodySmall)`
+  ${EllipsisStyle}
 `
 
-export default function TransactionPopup({ hash }: { hash: string }) {
-  const { chainId } = useWeb3React()
+function TransactionPopupContent({ tx, chainId }: { tx: TransactionDetails; chainId: number }) {
+  const success = Boolean(tx?.receipt && tx?.receipt.status === 1)
```
tinaszheng: here too

src/components/WalletDropdown/MiniPortfolio/PortfolioRow.tsx
```diff
@@ -28,39 +26,30 @@ export default function PortfolioRow({
...
   descriptor,
   right,
-  setIsHover,
   onClick,
 }: {
   left: React.ReactNode
   title: React.ReactNode
   descriptor?: React.ReactNode
-  right: React.ReactNode
+  right?: React.ReactNode
```
tinaszheng, line 34: if `right` is undefined, maybe the `EndColumn` parent below should also not be rendered?

src/components/Popups/TransactionPopup.tsx
```diff
@@ -1,46 +1,66 @@
...
+import { PopupAlertTriangle } from './FailedNetworkSwitchPopup'
+
+const Descriptor = styled(ThemedText.BodySmall)`
+  ${EllipsisStyle}
 `
 
-export default function TransactionPopup({ hash }: { hash: string }) {
-  const { chainId } = useWeb3React()
+function TransactionPopupContent({ tx, chainId }: { tx: TransactionDetails; chainId: number }) {
+  const success = Boolean(tx.receipt?.status === 1)
```
tinaszheng, line 21: this is already a `Boolean` so no need for the wrapper

tinaszheng (changes requested):
ok just a fewww more small things, almost there !!!

tinaszheng (approved):
LGTM
image

PR #6234, opened 2023-03-22
fix: crash in nft listing flow

just-toby:
fixes a crash when trying to list an NFT 

the crash happens to me every time i try to list an NFT. repro steps:
- go to your NFTs
- click "list for sale" on an NFT
- click "continue" in sidebar

```
Error: opacify: provided color #98A1C03d was not in hexadecimal format (e.g. #000000)
```

### test plan

i also added the first half of an e2e test that would prevent this from happening in the future, but need to check w/ the team about they typically test goerli NFTs in e2e tests.

otherwise, i've manually tested and verified that this fixes the listing flow for me

### DIFF
--- a/src/nft/components/profile/list/NFTListRow.tsx
+++ b/src/nft/components/profile/list/NFTListRow.tsx
@@ -7,7 +7,6 @@ import { Dispatch, useEffect, useReducer, useState } from 'react'
 import { Trash2 } from 'react-feather'
 import styled, { css, useTheme } from 'styled-components/macro'
 import { BREAKPOINTS, ThemedText } from 'theme'
-import { opacify } from 'theme/utils'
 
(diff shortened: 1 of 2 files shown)


src/nft/components/profile/list/NFTListRow.tsx
```diff
@@ -20,7 +19,7 @@ const NFTListRowWrapper = styled(Row)`
   border-radius: 8px;
 
   &:hover {
-    background: ${({ theme }) => opacify(24, theme.backgroundOutline)};
+    background: ${({ theme }) => theme.backgroundOutline};
```
just-toby, line 22: note: the definition of `theme.backgroundOutline` is already `opacify(24, theme.backgroundOutline)`

unipadmini (approved):

PR #6236, opened 2023-03-23
fix: remove backend name check from badge flag

JFrankfurt:
The prior implementation assumed backend support would be rolled out comprehensively, but right now assets/portfolio info is supported, but search is not.

https://uniswaplabs.atlassian.net/browse/WEB-3014

## Updated
image

## Current
image

### DIFF
--- a/src/components/NavBar/SearchBarDropdown.tsx
+++ b/src/components/NavBar/SearchBarDropdown.tsx
@@ -360,7 +360,7 @@ export const SearchBarDropdown = ({
     searchHistory,
   ])
 
-  const showBNBComingSoonBadge = Boolean(chainId === SupportedChainId.BNB && !isLoading)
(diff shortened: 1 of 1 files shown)


src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -361,9 +360,7 @@ export const SearchBarDropdown = ({
     searchHistory,
   ])
 
-  const showBNBComingSoonBadge = Boolean(
-    chainId !== undefined && chainId === SupportedChainId.BNB && !isLoading && !CHAIN_ID_TO_BACKEND_NAME[chainId]
-  )
+  const showBNBComingSoonBadge = Boolean(chainId !== undefined && chainId === SupportedChainId.BNB && !isLoading)
```
vm, line 363: nit: do you need the chainId !== undefined? i think chainId === SupportedChainId.BNB is enough

PR #6240, opened 2023-03-23
feat: Web 2996 add fiat on ramp buy flow to swap modal on the interface

lynnshaoyu:
**Description**

Adding 'buy' button to main Swap UI component, as another entry point to fiat on ramp flow. Note: buy fiat is not available in NY state.

Screen Shot 2023-03-28 at 6 13 04 PM

If buy fiat is available in the region, the flow should be: 
1. Click 'buy' button
2. If account is not yet connected, open wallet modal. Connect account. 
3. Once account is connected, buy fiat modal should open. 

If buy fiat is unavailable in the region, should see the 'Buy' button disabled (greyed out) and also a tooltip upon hover indicating this.
Screen Shot 2023-03-28 at 6 16 07 PM

Note: there should be a dot next to the Buy button when user sees it for the first time. Once they complete the buy flow for the first time, the dot should disappear permanently (ie: persist to local storage). 

https://uniswaplabs.atlassian.net/browse/WEB-2996?atlOrigin=eyJpIjoiYTQwNjAzZGQ0ZTcwNGNiNDk4Y2U5ZDg3NmU1YWQ4N2MiLCJwIjoiaiJ9

figma: https://www.figma.com/file/HOfzU6EFq4H5HNAIiH6Faa/FOR-v1.1?node-id=431-13762&t=4qHZ3uy88ibpFuM2-0

**Testing**
Tested locally and on this vercel build: https://interface-83oxzw5om-uniswap.vercel.app/#/swap

Note: we need to greenlight URLs on the moonpay backend in order for the fiat modal to work. however each vercel build is a new link. so if you're testing the latest build and it is NOT this link: https://interface-83oxzw5om-uniswap.vercel.app/#/swap it might not be working yet.


src/components/swap/SwapHeader.tsx
```diff
@@ -13,14 +22,136 @@ const StyledSwapHeader = styled.div`
...
+const Dot = styled.span`
+  height: 8px;
+  width: 8px;
+  background-color: ${({ theme }) => theme.accentActive};
+  border-radius: 50%;
+  margin: 0px 4px;
+  display: inline-block;
+`
+
+const TextHeader = styled.div<{ color: string; marginLeft?: string; isClickable?: boolean; disabled?: boolean }>`
```
tinaszheng, line 34: use `button` here if this component is clickable

and then you can use just use the one `disabled` prop and derive hover states based on that

```
cursor: pointer 

&:disabled {
   cursor: unset
}
```

src/components/swap/SwapHeader.tsx
```diff
@@ -13,14 +22,136 @@ const StyledSwapHeader = styled.div`
...
+  margin-left: ${({ marginLeft }) => marginLeft};
+  margin-right: 8px;
+  display: flex;
+  flex-direction: row;
+  cursor: ${({ isClickable }) => isClickable && 'pointer'};
+  justify-content: center;
+  align-items: center;
+`
+
+const TOKEN_SAFETY_ARTICLE =
```
tinaszheng, line 46: the article name doesnt seem to match this var 🤔

src/components/swap/SwapHeader.tsx
```diff
@@ -13,14 +22,136 @@ const StyledSwapHeader = styled.div`
...
+    account,
+    checkFiatRegionAvailability,
+    fiatOnrampAvailabilityChecked,
+    fiatOnrampAvailable,
+    setBuyFiatClicked,
+    toggleWalletDrawer,
+    handleBuyCrypto,
+    setContinueBuyFiatFlowPostRegionCheck,
+    continueBuyFiatFlowPostWalletModal,
+    continueBuyFiatFlowPostRegionCheck,
```
tinaszheng, line 118: do you need all these dependencies in this array? seems like only 4 variables are used

src/components/swap/SwapHeader.tsx
```diff
@@ -13,14 +22,136 @@ const StyledSwapHeader = styled.div`
...
       <RowBetween>
         <RowFixed>
-          <ThemedText.DeprecatedBlack fontWeight={500} fontSize={16} style={{ marginRight: '8px' }}>
+          <TextHeader className={subhead} color={theme.textPrimary}>
             <Trans>Swap</Trans>
-          </ThemedText.DeprecatedBlack>
+          </TextHeader>
+          {fiatOnRampButtonEnabled && (
+            <MouseoverTooltipContent
+              wrap={true}
```
tinaszheng, line 130:
```suggestion
              wrap
```

src/components/swap/SwapHeader.tsx
```diff
@@ -13,14 +22,136 @@ const StyledSwapHeader = styled.div`
...
+              content={
+                <div>
+                  <Trans>Crypto purchases are not available in your region. </Trans>
+                  <ExternalLink href={TOKEN_SAFETY_ARTICLE} style={{ paddingLeft: '4px' }}>
```
tinaszheng, line 134: use `gap` on parent instead of defining padding here - in general layout should be parent's responsibility 🤔

tinaszheng: but also is just the regular sentence space not sufficient here?

src/components/swap/SwapHeader.tsx
```diff
@@ -13,14 +22,136 @@ const StyledSwapHeader = styled.div`
...
+              }
+              placement="bottom"
+              disableHover={!fiatOnrampAvailabilityChecked || fiatOnrampAvailable}
+            >
+              <TextHeader
+                className={subhead}
+                color={theme.textSecondary}
+                marginLeft="8px"
+                disabled={disableBuyCryptoButton}
+                isClickable={true}
```
tinaszheng, line 147: if `<TextHeader>` is always `isClickable` why is this prop here 🤔

src/components/swap/SwapHeader.tsx
```diff
@@ -13,14 +22,136 @@ const StyledSwapHeader = styled.div`
...
+                  </ExternalLink>
+                </div>
+              }
+              placement="bottom"
+              disableHover={!fiatOnrampAvailabilityChecked || fiatOnrampAvailable}
+            >
+              <TextHeader
+                className={subhead}
+                color={theme.textSecondary}
+                marginLeft="8px"
```
tinaszheng, line 145: same comment here about using `gap` on parent instead of defining margin on the children

src/state/user/reducer.ts
```diff
@@ -55,6 +57,7 @@ function pairKey(token0Address: string, token1Address: string) {
 }
 
 export const initialState: UserState = {
+  buyFiatClicked: false,
```
tinaszheng, line 60: doesnt this need a migration?

src/components/swap/SwapHeader.tsx
```diff
@@ -13,14 +22,136 @@ const StyledSwapHeader = styled.div`
...
+      openFiatOnRampModal()
+      setBuyFiatClicked(true)
+      setContinueBuyFiatFlowPostRegionCheck(false)
+      setContinueBuyFiatFlowPostWalletModal(false)
+    } else if (!fiatOnrampAvailable) {
+      setBuyFiatClicked(true)
+      setContinueBuyFiatFlowPostRegionCheck(false)
+      setContinueBuyFiatFlowPostWalletModal(false)
+    }
+  }, [
```
tinaszheng, line 93: can you explain what the behavior here is doing 🤔 

is this essentially like incrementing steps in progress in the flow?

src/components/Tooltip/index.tsx
```diff
@@ -103,13 +119,22 @@ export function MouseoverTooltipContent({
...
+  const close = () => {
+    setShow(false)
+  }
+
   return (
-    <TooltipContent {...rest} show={!disableHover && show} content={disableHover ? null : content}>
+    <TooltipContent
+      {...rest}
+      open={open}
+      close={close}
```
tinaszheng, line 134: you can technically just define this inline to be cleaner and get rid of lines 126-129

```
close={() => setShow(false)}
```

src/components/swap/SwapHeader.tsx
```diff
@@ -13,14 +22,136 @@ const StyledSwapHeader = styled.div`
...
+
+  const [checkFiatRegionAvailability, setCheckFiatRegionAvailability] = useState(false)
+  const {
+    available: fiatOnrampAvailable,
+    availabilityChecked: fiatOnrampAvailabilityChecked,
+    error,
+    loading: fiatOnrampAvailabilityLoading,
+  } = useFiatOnrampAvailability(checkFiatRegionAvailability)
+
+  const disableBuyCryptoButton = Boolean(
```
tinaszheng, line 69: nit: move this below all the hooks so that the hooks in the file are grouped together

src/state/user/hooks.tsx
```diff
@@ -68,6 +69,18 @@ export function useIsExpertMode(): boolean {
   return useAppSelector((state) => state.user.userExpertMode)
 }
 
+export function useBuyFiatClicked(): [boolean | undefined, (buyFiatClicked: boolean) => void] {
+  const dispatch = useAppDispatch()
+  const buyFiatClicked = useAppSelector((state) => state.user.buyFiatClicked)
+  const setBuyFiatClicked = useCallback(
+    (buyFiatClicked: boolean) => {
+      dispatch(updateUserBuyFiatClicked({ buyFiatClicked }))
```
tinaszheng, line 77: you can just pass in the `boolean` value as the payload here right?

```
dispatch(updateUserBuyFiatClicked(buyFiatClicked))`
```

and then the reducer below would just be

```
updateUserBuyFiatClicked(state, action) {
      state.buyFiatClicked = action.payload
    },
```

src/components/Tooltip/index.tsx
```diff
@@ -1,13 +1,16 @@
 import { transparentize } from 'polished'
-import { ReactNode, useCallback, useEffect, useState } from 'react'
+import { ReactNode, useEffect, useState } from 'react'
 import styled from 'styled-components/macro'
 
 import Popover, { PopoverProps } from '../Popover'
 
+const noOp = () => null
```
tinaszheng, line 7: nit, this should probably go into like a shared utils file since there are a couple of other one-off `noop` function definitions

also ive also always seen it spelled `noop` instead of `noOp` 🤔

tinaszheng (changes requested):

src/components/swap/SwapHeader.tsx
lynnshaoyu: sg done

src/components/Tooltip/index.tsx
lynnshaoyu: close is used again on line 141 so i feel like defining the function actually makes sense here. also same thing for open, so want to keep it consistent for readability.

src/components/swap/SwapHeader.tsx
lynnshaoyu: you are right yikes! i just didn't change cuz the linter wasn't complaining. thanks for pointing out.

src/components/swap/SwapHeader.tsx
lynnshaoyu: So TextHeader shouldn't always be clickable because Swap is also styled by TextHeader and it does nothing when clicked (thats the case where isClickable is false)

src/components/swap/SwapHeader.tsx
lynnshaoyu: Hmm so I don't want this component to always be clickable / be a button because the 'Swap' header is also styled by this component, and it is not clickable.

src/components/swap/SwapHeader.tsx
lynnshaoyu: yep exactly! it took me a few attempts to get this to reflect the flow we want without bugs XD 

so this is what is happening:
1. upon first attempt to click buy, make the region check (setCheckFiatRegionAvailability(true)) to see if fiat on ramps is available for the user.
2. if its UNAVAILABLE, we reach the 4th if statement case. this is the end of the flow. 
3. if it is AVAILABLE, we need to check if their account is connected. if not, open the wallet modal. however, we need to remember to open the fiat on ramp modal AFTER the wallet is connected. hence the continueBuyFiatFlowPostWalletModal state variable. 
4. if it is AVAILABLE and the account is now connected, we can finally show the fiat on ramp modal. this is also the end of the flow.

src/components/swap/SwapHeader.tsx
lynnshaoyu: adding gap in the parent doesn't work in this case, it will treat the link and the sentence as two separate entities and put a space between them. we want it to flow as if one paragraph. the sentence space doesn't show up unfortunately without padding :S I think padding Left here is the best / least code heavy solution.

src/state/user/reducer.ts
lynnshaoyu: we don't have migration yet on web 😅 (we are moving to redux persist soon from your suggestion remember :P) because of the way its being used in SwapHeader, undefined case is handled correctly / has same expected behavior as initial state false.

src/components/Tooltip/index.tsx
lynnshaoyu: sounds good I added a TODO for now, seems like something I can do in a separate PR for migration across web.

src/components/swap/SwapHeader.tsx
```diff
@@ -13,14 +22,120 @@ const StyledSwapHeader = styled.div`
...
+  'https://support.uniswap.org/hc/en-us/articles/11306664890381-Why-isn-t-MoonPay-available-in-my-region-'
+
+// TODO(lynnshaoyu): add analytics and logging
 export default function SwapHeader({ allowedSlippage }: { allowedSlippage: Percent }) {
+  const theme = useTheme()
+  const { account } = useWeb3React()
+  const openFiatOnRampModal = useOpenModal(ApplicationModal.FIAT_ONRAMP)
+  const fiatOnRampButtonEnabled = useFiatOnRampButtonEnabled()
+  const [buyFiatClicked, setBuyFiatClicked] = useBuyFiatClicked()
+  const [continueBuyFiatFlowPostRegionCheck, setContinueBuyFiatFlowPostRegionCheck] = useState(false)
```
tinaszheng, line 55: all of this new code should go in a separate `Buy` button component (up til line 138) so that the SwapHeader file itself can be relatively straightforward, and the FOR logic can be isolated to its own thing

src/components/swap/SwapHeader.tsx
tinaszheng: flow makes sense to me, a few small things that stand out to me:
1) it's not obvious from this code what happens if the user already has the wallet drawer open, isn't connected, and clicks on 'buy' - are they sort of "stuck"? maybe this is more of a product question
2) what happens if a user clicks the button again in the middle of the flow? the button should probably be disabled if in the middle of the flow, i think
3) it's a bit hard to read and it's not immediately obvious how this callback + the useeffect below work together to increment state.

a few suggestions on how to improve readability:

1. add this github comment as a comment in code
2. it's a bit confusing that both the button click and the useEffect use this function. to me it feels like it's comingling the concern of starting the flow vs incrementing the flow, and to me it feels like it should be different because starting the flow is a user action vs incrementing the flow is automatic
3. this is just a suggestion so feel free to change up as you see fit or not use at all: the number of variables `checkFiatRegionAvailability`, `continueBuyFiatFlowPostWalletModal`, and `continueBuyFiatFlowPostRegionCheck` could probably be condensed. essentially, there's only one possible state for a user to be in at a time, so instead of setting true/false on a bunch of different individual state variables, could you use strings to represent the curr step instead of boolean combinations? 
4. you could probably break out the `setBuyFiatClicked` handling into its own useEffect that only updates once availability status has been checked, and then add a comment there about what `setBuyFiatClicked` is doing (e.g. hide blue dot if user already attempted the flow)

tinaszheng: also you could probably break out these hooks into a separate one called like `useFiatOnRampStep` and then add unit tests for it since it's a rather complicated flow

src/components/swap/SwapHeader.tsx
```diff
@@ -13,14 +22,120 @@ const StyledSwapHeader = styled.div`
...
+  ])
+
+  useEffect(() => {
+    if (continueBuyFiatFlowPostRegionCheck || (account && continueBuyFiatFlowPostWalletModal)) {
+      handleBuyCrypto()
+    }
+  }, [account, handleBuyCrypto, continueBuyFiatFlowPostWalletModal, continueBuyFiatFlowPostRegionCheck])
+
+  const disableBuyCryptoButton = Boolean(
+    error || (!fiatOnrampAvailable && fiatOnrampAvailabilityChecked) || fiatOnrampAvailabilityLoading
```
tinaszheng, line 101: also does this `error` handling work correctly if it's a one-off error like internet issues? it seems like from this code that there's no way to 'reset' the error by clicking on the button again

src/components/swap/SwapHeader.tsx
JFrankfurt: I find modeling flows like this as a state machine helpful for readability

src/components/Tooltip/index.tsx
```diff
@@ -25,20 +29,23 @@ interface TooltipProps extends Omit<PopoverProps, 'content'> {
...
   disableHover?: boolean // disable the hover and content display
   timeout?: number
 }
 
 interface TooltipContentProps extends Omit<PopoverProps, 'content'> {
   content: ReactNode
   onOpen?: () => void
+  open?: () => void
+  close?: () => void
+  // time delay before showing tooltip upon hover
```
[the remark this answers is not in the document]
lynnshaoyu: sg done

src/components/swap/SwapHeader.tsx
lynnshaoyu: sounds good done

src/components/swap/SwapHeader.tsx
lynnshaoyu: good point, changed it to log error to console

src/components/Tooltip/index.tsx
```diff
@@ -100,21 +119,43 @@ export function MouseoverTooltipContent({
   children,
   onOpen: openCallback = undefined,
   disableHover,
+  delay,
   ...rest
 }: Omit<TooltipContentProps, 'show'>) {
   const [show, setShow] = useState(false)
-  const open = useCallback(() => {
+  // Default delay is undefined (no delay), so delayShow should initialize to be true.
+  const [delayShow, setDelayShow] = useState(delay ? false : true)
```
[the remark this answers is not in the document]
lynnshaoyu: resolved, discussed offline why we need delayShow variable

src/components/swap/SwapHeader.tsx
lynnshaoyu: sounds good done, separated BuyFiatButton into a separate component and used ButtonText component

src/components/swap/SwapHeader.tsx
lynnshaoyu: 1. Done
2.  I think re-using `handleBuyCrypto` is the most efficient way to code the desired behavior here, but understand the concern / how it can look confusing. To make it clearer, I added a comment above the second useEffect: `// Increment buy fiat flow state automatically, when requisite state changes have occured.` So the `handleBuyCrypto` is ALWAYS called when a user clicks on the Buy button, but is called conditionally when state is incremented automatically by the useEffect.
4. This makes sense, I condensed the state vars into the enum 'BuyFiatFlowState'
5. This makes sense too, I used another useEffect to handle "terminal" states of 'BuyFiatFlowState', which I defined as: `const TerminalBuyFiatFlowStates = new Set([
  BuyFiatFlowState.AVAILABLE_IN_REGION,
  BuyFiatFlowState.UNAVAILABLE_IN_REGION,
])`

lynnshaoyu: Also to answer the first 3 things that stand out: 

1. Good point, I disabled the Buy button now when the wallet drawer is open, and user is mid-flow.
2. Done.
3. I hope the comments, use of BuyFiatFlowState, and separation of concerns into multiple useEffects have made it clearer and easier to read 😁🤞

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+    loading: fiatOnrampAvailabilityLoading,
+  } = useFiatOnrampAvailability(checkFiatRegionAvailability)
+  const [buyFiatFlowState, setBuyFiatFlowState] = useState(BuyFiatFlowState.INACTIVE)
+  const [walletDrawerOpen, toggleWalletDrawer] = useWalletDrawer()
+
+  /* 
+  Once the buy fiat flow has been completed, reset BuyFiatFlowState and also save to 
+  local storage that user has completed this flow already.
+  */
+  const terminateBuyFiatFlow = useCallback(() => {
```
lynnshaoyu, line 52: @tinaszheng okayyy so separating this out into another useEffect actually introduced more bugs (the blue dot wasn't appearing when it should), because setState doesn't happen IMMEDIATELY in react (they're batched for perf reasons). Also, having this called in another useEffect introduced the need for another "terminal" state in BuyFiatFlowState (eg: when either the modal is available and ready to be shown to the user OR when it is unavailable). 

Thus I think using another function expressively called "terminateBuyFiatFlow" is better than using another useEffect.

src/components/swap/SwapHeader.tsx
lynnshaoyu: Ok wait @tinaszheng I removed the other useEffect for this reason: https://github.com/Uniswap/interface/pull/6240/files#r1156534032

src/components/Popover/index.tsx
```diff
@@ -117,17 +123,40 @@ export default function Popover({
   const updateCallback = useCallback(() => {
     update && update()
   }, [update])
-  useInterval(updateCallback, show ? 100 : null)
+  useInterval(updateCallback, show ? 200 : null)
+
+  useEffect(() => {
+    if (delayBeforeShow) {
```
tinaszheng, line 129: prefer guard statement here, so:

```
if (!delayBeforeShow) return 

...
```

and then you can avoid the nested if statements

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+      fiatOnrampAvailabilityLoading ||
+      // When wallet drawer is open AND user is in the connect wallet step
+      // of the buy fiat flow, disable buy fiat button.
+      (walletDrawerOpen && buyFiatFlowState === BuyFiatFlowState.ACTIVE_NEEDS_WALLET_CONNECTION)
+  )
+
+  return (
+    <MouseoverTooltipContent
+      wrap
+      delayBeforeShow={500}
```
tinaszheng, line 115: nit: move this to a constant that has the unit as a suffix, e.g. `POPOVER_DELAY_MS`

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+          </ExternalLink>
+        </div>
+      }
+      placement="bottom"
+      disableHover={!fiatOnrampAvailabilityChecked || (fiatOnrampAvailabilityChecked && fiatOnrampAvailable)}
+    >
+      <ButtonText
+        onClick={handleBuyCrypto}
+        color={theme.textSecondary}
+        noUnderline={true}
```
tinaszheng, line 130:
```suggestion
        noUnderline
```

src/components/swap/SwapHeader.tsx
```diff
@@ -13,14 +15,31 @@ const StyledSwapHeader = styled.div`
...
 
+const TextHeader = styled.div<{ color: string; marginLeft?: string; isClickable?: boolean; disabled?: boolean }>`
+  color: ${({ color }) => color};
+  opacity: ${({ disabled }) => disabled && '60%'};
+  margin-left: ${({ marginLeft }) => marginLeft};
+  margin-right: 8px;
+  display: flex;
+  line-height: 20px;
+  flex-direction: row;
+  cursor: ${({ isClickable }) => isClickable && 'pointer'};
```
tinaszheng, line 26: i think you should be able to remove these new props now!

src/state/user/reducer.ts
```diff
@@ -55,6 +57,7 @@ function pairKey(token0Address: string, token1Address: string) {
 }
 
 export const initialState: UserState = {
+  buyFiatFlowCompleted: false,
```
tinaszheng, line 60: since we don't have redux migrations yet, this initial state will technically just be `undefined`

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+  Depending on the current state of the buy fiat flow the user is in (buyFiatFlowState),
+  the desired behavior of clicking the 'Buy' button is different:
+  1) Initially upon first click, need to check the availability of the feature in the user's 
+  region, and continue the flow.
+  2) If the feature is available in the user's region, need to connect a wallet, and continue
+  the flow.
+  3) If the feature is available and a wallet account is connected, show fiat on ramp modal.
+  4) If the feature is unavailable, show feature unavailable tooltip. 
+  */
+  const handleBuyCrypto = useCallback(() => {
```
tinaszheng, line 67: nice this logic is much cleaner / more readable!

src/components/Button/index.tsx
```diff
@@ -227,19 +227,20 @@ export const ButtonEmpty = styled(BaseButton)`
   }
 `
 
-export const ButtonText = styled(BaseButton)`
+export const ButtonText = styled(BaseButton)<{ color?: string; noUnderline?: boolean }>`
```
tinaszheng, line 230: can the `SwapBuyFiatButton` component just inherit from this and then override color + the focus/active states there? i prefer that over trying to finesse these shared components into more generalized versions 😬

src/components/swap/SwapBuyFiatButton.tsx
lynnshaoyu: done

src/components/Popover/index.tsx
lynnshaoyu: makes sense ty done

src/components/swap/SwapBuyFiatButton.tsx
lynnshaoyu: sg done

src/components/swap/SwapBuyFiatButton.tsx
lynnshaoyu: removed, prop not needed anymore

src/components/Button/index.tsx
lynnshaoyu: yeppp that is wayyy cleaner, ty done.

src/components/swap/SwapHeader.tsx
lynnshaoyu: ah youre right, done!

src/state/user/reducer.ts
tinaszheng: yeah im saying you should just put `undefined` here to be more accurate

src/components/swap/SwapHeader.tsx
```diff
@@ -13,14 +15,28 @@ const StyledSwapHeader = styled.div`
...
+  const theme = useTheme()
+  const fiatOnRampButtonEnabled = useFiatOnRampButtonEnabled()
+
   return (
     <StyledSwapHeader>
       <RowBetween>
-        <RowFixed>
-          <ThemedText.DeprecatedBlack fontWeight={500} fontSize={16} style={{ marginRight: '8px' }}>
+        <RowFixed style={{ gap: '8px' }}>
+          <TextHeader className={subhead} color={theme.textPrimary}>
```
tinaszheng, line 36: if this is the only usage of `TextHeader` you can just add the `theme.textPrimary` color to the styled component directly, same with whatever properties are in `subhead`

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,150 @@
...
+  // Default initial state. User is not actively trying to buy fiat.
+  INACTIVE,
+  // Buy fiat flow is active and region availability has been checked.
+  ACTIVE_POST_REGION_CHECK,
+  // Buy fiat flow is active, feature is available in user's region & needs wallet connection.
+  ACTIVE_NEEDS_WALLET_CONNECTION,
+}
+
+const StyledTextButton = styled(ButtonText)<{ color?: string }>`
+  color: ${({ color }) => color};
```
tinaszheng, line 34: same here - you can just do `color: ${({ theme }) => theme.textSecondary};` here instead of passing the prop through since this is the only usage

src/components/swap/SwapBuyFiatButton.test.tsx
```diff
@@ -0,0 +1,11 @@
+import { render } from 'test-utils'
+
+import SwapBuyFiatButton from './SwapBuyFiatButton'
+
+// TODO(lynnshaoyu): add detailed unit tests for critical flows of component.
```
tinaszheng, line 5: i would recommend doing this before the launch :O

tinaszheng (changes requested):

src/components/swap/SwapBuyFiatButton.test.tsx
yyip-dev: +1 we should be adding unit tests for critical flows; adding post-launch defeats the purpose of helping increase our confidence in the code and functionality.

lynnshaoyu: sounds good does it have to go in with this PR or a follow up PR is ok?

src/components/swap/SwapBuyFiatButton.tsx
lynnshaoyu: sg done

src/components/swap/SwapBuyFiatButton.test.tsx
yyip-dev: It makes more sense to include here, as unit tests for new code should be included as part of a PR. It will also be easier to follow and understand if we wrote tests, and what tests were written, if there's an issue with the PR.

src/state/user/reducer.ts
lynnshaoyu: ooh sg can do, can change back to false after the migration left a todo.

src/components/swap/SwapBuyFiatButton.test.tsx
lynnshaoyu: makes sense, adding now!

src/components/swap/SwapHeader.tsx
tinaszheng: why not just use `<ThemedText.SubHeader>` as a font wrapper

lynnshaoyu: because ThemedText.SubHeader actually doesn't have the right properties for what Figma asks for 😬 i've found NFT's sprinkles to be more consistent / up to date with what we have in Figma. 

Specifically: 
Figma Line Height = 24px = sprinkles subhead line height
Figma Font Weight = 500 = sprinkles font weight

For ThemedText.Subheader, it has: Font weight = 600, and no specification for line height.

src/components/swap/SwapBuyFiatButton.test.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+const mockUseFiatOnRampsAvailable = (shouldCheck: boolean) => {
+  if (shouldCheck) {
+    return {
+      available: true,
+      availabilityChecked: true,
+      error: null,
+      loading: false,
+    }
+  } else {
```
tinaszheng, line 52: you don't need the `else` statement here if you're returning in the first `if`

see [here](https://github.com/airbnb/javascript#blocks--no-else-return)

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,149 @@
...
+    if (
+      buyFiatFlowState === BuyFiatFlowState.ACTIVE_POST_REGION_CHECK ||
+      (account && buyFiatFlowState === BuyFiatFlowState.ACTIVE_NEEDS_WALLET_CONNECTION)
+    ) {
+      handleBuyCrypto()
+    }
+  }, [account, handleBuyCrypto, buyFiatFlowState])
+
+  useEffect(() => {
+    if (error) console.error(error)
```
tinaszheng, line 117: it looks like there's already a console.error from within `useFiatOnrampAvailability` so not sure this is needed?

src/state/user/reducer.ts
```diff
@@ -55,6 +57,8 @@ function pairKey(token0Address: string, token1Address: string) {
 }
 
 export const initialState: UserState = {
+  // TODO(lynnshaoyu): change buyFiatFlowCompleted initial state to false after finishing redux migration.
```
tinaszheng, line 60: so you won't be able to change the initial state here later because some users will have already had `buyFiatFlowCompleted: undefined` as their persisted redux state

src/components/swap/SwapHeader.tsx
tinaszheng: aren't you setting the line-height to be 20 [here](https://github.com/Uniswap/interface/pull/6240/files#diff-020959a0e0e3d1c593444c25634669a19de67351f4af27bb32ba575469112eefR22)?

tinaszheng:
when i was testing earlier i kept running into the case where i'd move my mouse away after click and "nothing happens" because im not remaining hovering on the button to see the tooltip 
https://user-images.githubusercontent.com/59578595/230137608-7c30e735-b61a-4d01-aec0-0d0a92421482.mov

zzmp:
Tested from WA and it works as described.

src/components/Popover/index.tsx
```diff
@@ -98,6 +101,9 @@ export default function Popover({
   const [referenceElement, setReferenceElement] = useState<HTMLDivElement | null>(null)
   const [popperElement, setPopperElement] = useState<HTMLDivElement | null>(null)
   const [arrowElement, setArrowElement] = useState<HTMLDivElement | null>(null)
+  // Default delayBeforeShow is undefined (no delay before showing),
+  // so delayShow should initialize to be true.
+  const [delayShow, setDelayShow] = useState(delayBeforeShow ? false : true)
```
zzmp, line 106: This is confusing. Instead, can you default `delayBeforeShow = 0` (on L93), and then:

```suggestion
  const [delayShow, setDelayShow] = useState(Boolean(delayBeforeShow))
```

lynnshaoyu:
> when i was testing earlier i kept running into the case where i'd move my mouse away after click and "nothing happens" because im not remaining hovering on the button to see the tooltip

Yes I think this is as expected tbh :/ There was no specification to surface the tooltip upon CLICK (not hover). We used to have a toast but we deleted it upon rec from Cal and Fred because it was too busy. Unless @infredible wants diff behavior I think this is what we want?

src/components/Popover/index.tsx
zzmp: This variable name is confusing. Can you rename to `shouldShow` or `shouldDelayShow` so it's more obvious what it is doing?

src/components/Tooltip/index.tsx
```diff
@@ -1,13 +1,17 @@
 import { transparentize } from 'polished'
-import { ReactNode, useCallback, useEffect, useState } from 'react'
+import { ReactNode, useEffect, useState } from 'react'
 import styled from 'styled-components/macro'
 
 import Popover, { PopoverProps } from '../Popover'
 
+// TODO(lynnshaoyu): migrate noOps throughout web to a shared util file.
+const noOp = () => null
```
zzmp, line 8: `nop` or `noop` is more idiomatic

```suggestion
const noop = () => null
```

src/state/user/reducer.ts
lynnshaoyu: oooh I see gotcha, ok that works fine with our use case

src/components/swap/SwapBuyFiatButton.tsx
lynnshaoyu: true removed

src/components/swap/SwapBuyFiatButton.test.tsx
lynnshaoyu: thats true, removed

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,149 @@
...
+      fiatOnrampAvailabilityLoading ||
+      // When wallet drawer is open AND user is in the connect wallet step
+      // of the buy fiat flow, disable buy fiat button.
+      (walletDrawerOpen && buyFiatFlowState === BuyFiatFlowState.ACTIVE_NEEDS_WALLET_CONNECTION)
+  )
+
+  return (
+    <MouseoverTooltipContent
+      wrap
+      delayBeforeShow={POPOVER_DELAY_BEFORE_SHOW_MS}
```
zzmp, line 126: Why is this behavior necessary? It adds a lot of complexity to this PR.
Are we able to push back on design to remove this, for the sake of a simpler codebase?

src/components/swap/SwapBuyFiatButton.test.tsx
grabbou: As a general practice @lynnshaoyu, can you convert this into JIRA ticket and reference here as `TODO(WEB-XXX): description` so that we can track it later?

grabbou:
Hey, 

Added one comment here -> https://github.com/Uniswap/interface/pull/6240#discussion_r1158795286

Just wanted to explicitly call that out in case it got lost!

src/components/Popover/index.tsx
grabbou: `shouldShow` is fine, +1 on that change

src/components/Popover/index.tsx
```diff
@@ -117,17 +123,38 @@ export default function Popover({
...
+      const tooltipTimer = setTimeout(() => {
+        setDelayShow(true)
+      }, delayBeforeShow)
+
+      return () => {
+        clearTimeout(tooltipTimer)
+      }
+    } else {
+      setDelayShow(false)
+      return
```
grabbou, line 140: nit: `return` is redundant here, isn't it?

src/components/Popover/index.tsx
```diff
@@ -117,17 +123,38 @@ export default function Popover({
...
+    }
+  }, [show, delayBeforeShow])
 
   return showInline ? (
-    <PopoverContainer show={show}>{content}</PopoverContainer>
+    <PopoverContainer show={show && delayShow}>{content}</PopoverContainer>
```
grabbou, line 145: Can we incorporate `show` and `delayShow` into a single variable, something like `shouldShow`? It feels to me that `shouldShow` is always `true` when `show` is true, isn't it? I would think about that and try to simplify.

src/components/swap/SwapBuyFiatButton.tsx
lynnshaoyu: they specifically requested this so idts? unless @infredible ok's it - see thread convo here: https://uniswapteam.slack.com/archives/C04R3CV5887/p1680116729731119?thread_ts=1680116220.405679&cid=C04R3CV5887

src/components/Popover/index.tsx
grabbou: One of the ways one could think about it is creating a higher-order-compoent that takes `show` prop, adds timeout and passes `show` prop down to the final component (like this Popover), so Popover itself is not aware of the "timeout" and "interval". 

At the same time, I understand our codebase is far from being well organised as far as components, so it may seem like an overkill. Just a general comment.

src/components/Tooltip/index.tsx
```diff
@@ -100,21 +128,28 @@ export function MouseoverTooltipContent({
...
   return (
-    <TooltipContent {...rest} show={!disableHover && show} content={disableHover ? null : content}>
-      <div
-        style={{ display: 'inline-block', lineHeight: 0, padding: '0.25rem' }}
-        onMouseEnter={open}
-        onMouseLeave={close}
-      >
+    <TooltipContent
+      {...rest}
+      delayBeforeShow={delayBeforeShow}
```
grabbou, line 146: nit: I would propose renaming this property, b/c it sounds more like a boolean, but it's a timeout effectively. Maybe `delayShowTimeout`?

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,144 @@
...
+  &:active {
+    text-decoration: none;
+  }
+`
+
+const POPOVER_DELAY_BEFORE_SHOW_MS = 500
+
+export default function SwapBuyFiatButton() {
+  // Importing useWeb3React here is necessary for mocking different behavior in
+  // different tests in SwapBuyFiatButton.test.tsx
```
grabbou, line 47: Isn't there really any other way to mock an import? In web3-react, we don't have to import it this way and we still can use Jest to mock it. I would have another look, because this is effectively a dynamic require.

src/components/Popover/index.tsx
lynnshaoyu: I get a lint error: Not all code paths return a value.ts(7030)

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,144 @@
...
+  Once the buy fiat flow has been completed, reset BuyFiatFlowState and also save to 
+  local storage that user has completed this flow already.
+  */
+  const terminateBuyFiatFlow = useCallback(() => {
+    setBuyFiatFlowCompleted(true)
+    setBuyFiatFlowState(BuyFiatFlowState.INACTIVE)
+  }, [setBuyFiatFlowCompleted, setBuyFiatFlowState])
+
+  /*
+  Depending on the current state of the buy fiat flow the user is in (buyFiatFlowState),
```
grabbou, line 73: nit: please format this comment, e.g:
```
/*
 *
 */
```

Also, bonus points for being consistent - there's a mix of `//` (multiline) and this. I personally prefer `/* */`, but I know we don't have a good standard yet. Either way, we need to fix formatting here.

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,144 @@
...
+  the desired behavior of clicking the 'Buy' button is different:
+  1) Initially upon first click, need to check the availability of the feature in the user's 
+  region, and continue the flow.
```
grabbou, line 76: nit: would it be possible to break this comment down, so that each point is placed right above `if` condition, so it's easier to scan?

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,144 @@
...
+    terminateBuyFiatFlow,
+  ])
+
+  // Continue buy fiat flow automatically when requisite state changes have occured.
+  useEffect(() => {
+    if (
+      buyFiatFlowState === BuyFiatFlowState.ACTIVE_POST_REGION_CHECK ||
+      (account && buyFiatFlowState === BuyFiatFlowState.ACTIVE_NEEDS_WALLET_CONNECTION)
+    ) {
+      handleBuyCrypto()
```
grabbou, line 111: nit: I generally have nothing against `useEffect` for executing side-effects (it's one of the ways to do it React-way), but maybe... we can call `buyCrypto()` function directly from a click hnadler and don't really need state for it? Sounds like much easier way?

src/components/Popover/index.tsx
lynnshaoyu: sg, the boolean makes sense... done on the var rename too.

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,144 @@
...
+    }
+  }, [account, handleBuyCrypto, buyFiatFlowState])
+
+  const disableBuyCryptoButton = Boolean(
```
grabbou, line 115: nit: is `boolean()` conversion needed here? Looks like all variables are booleans down the road?

grabbou: nit: rename to `buyCryptoButtonDisabled` (`disabled` at the end) to stay consistent with the rest of the naming patterns in this file

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,144 @@
...
+      // of the buy fiat flow, disable buy fiat button.
+      (walletDrawerOpen && buyFiatFlowState === BuyFiatFlowState.ACTIVE_NEEDS_WALLET_CONNECTION)
+  )
+
+  return (
+    <MouseoverTooltipContent
+      wrap
+      delayBeforeShow={POPOVER_DELAY_BEFORE_SHOW_MS}
+      content={
+        <div data-testid="fiat-on-ramp-unavailable-tooltip">
```
grabbou, line 128: nit: maybe using a grid component with space instead of `div` (we rarely use them actually?). Not a big deal, just pointing out some options.

src/components/Tooltip/index.tsx
lynnshaoyu: done

src/state/user/reducer.ts
grabbou, line 60: Please make sure each new TODO has a corresponding JIRA ticket

grabbou (commented):
Hey! Just left a review - I don't really have anything strong against this PR, so I am going to leave most of my comments as nitpicks and recommendations. Personally, I would spend some time improving the abstraction here so it's reducing potential code debt, there are a few places where we can optimise (like the `Tooltip` component). Rather than that, this PR is solid work 🤝

(Not requesting changes, b/c I defer final approval to @tinaszheng who's been on it already)

src/components/swap/SwapBuyFiatButton.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+    mockUseWalletDrawer.mockImplementation(() => [false, toggleWalletDrawer])
+    mockUseOpenModal.mockImplementation(() => useOpenModal)
+    render(<SwapBuyFiatButton />)
+    await userEvent.click(screen.getByTestId('buy-fiat-button'))
+    expect(toggleWalletDrawer).toHaveBeenCalledTimes(1)
+    expect(screen.queryByTestId('fiat-on-ramp-unavailable-tooltip')).not.toBeInTheDocument()
+  })
+
+  it('fiat on ramps available in region, account connected', async () => {
+    jest.doMock('@web3-react/core', () => {
```
lynnshaoyu, line 91: here @grabbou

src/components/swap/SwapBuyFiatButton.tsx
lynnshaoyu: The reason why I had to do this is because I need different mock implementations a specific module (@web3-react/core, specifically the useWeb3React hook) in different tests in the same test file / suite. 

see this line of code: https://github.com/Uniswap/interface/pull/6240/files#r1158838190 

So in that specific test, I need account to be defined, but in the other tests in that file, I need account to be undefined. If I were just dealing with the useWeb3React hook that would be easy, I can do something like this: `jest.mock('@web3-react/core')
const mockUseWeb3React = useWeb3React as jest.MockedFunction<typeof useWeb3React>` and do different mockUseWeb3React.mockImplementation calls in different parts of the test. 

However, the trickiness here is that I NEED the REAL implementation of @web3-react/core (besides useWeb3React) for the tests to run properly. Hence I need to use jest.doMock in the individual test (not at the top level jest.mock), which only works with changing the original SwapBuyFiatButton.tsx file to import differently. With the original import at the top of the file, doMock is completely ignored. 

I tried many different ways of doing this yesterday and have gone deeeep into the github issue / stack overflow rabbithole, this is the only way I've found to work. The alternative is having 2 separate SwapBuyFiatButton.test.tsx files, one for account connected and one for not account connected. 

If someone reading this DOES know of a better way of doing this that is 100% confirmed to work, I would looveee to learn about it! 🙏

src/components/swap/SwapBuyFiatButton.tsx
lynnshaoyu: done

lynnshaoyu: done

src/components/swap/SwapBuyFiatButton.tsx
lynnshaoyu: I tried breaking it into the separate lines and I feel like it looks messier? :S its in order that matches the is / else if statements anyway so I woud prefer keeping it like this.

src/components/swap/SwapBuyFiatButton.tsx
tinaszheng: what error are you getting by just doing `jest.mock('@web3-react/core')` ?

if it needs to call specific functions, you can just mock those directly:

example:
```
jest.mock('@web3-react/core', () => ({
  ModuleName: { functionName: () => ({ on: jest.fn() }) },
}))
```

lynnshaoyu: jest.mock('@web3-react/core') makes the entire module mocked, so all the other imports we need from '@web3-react/core' don't have their real functionalities, which causes multiple breakages. 

You can see throughout the codebase's other tests, this is why we need to jest.requireActual the rest of the module when mocking useWeb3React hook, like this: 
```
jest.mock('@web3-react/core', () => {
  const web3React = jest.requireActual('@web3-react/core')
  return {
    ...web3React,
    useWeb3React: () => ({
      account: '123',
      isActive: true,
    }),
  }
})
```

I'm not sure how the second example is relevant in this use case? 

Also @tinaszheng  and @grabbou pls feel free to push a commit or an example as to how you would do this testing differently if you've found something that works🙏

tinaszheng: sounds good, if it's already used throughout the codebase as-is i won't make you change it as part of this PR haha

tinaszheng (approved):
@zzmp had a comment about removing the delayShow and looks like Fred approved, will leave it up to you whether you want to support the behavior or not!

src/components/swap/SwapBuyFiatButton.tsx
lynnshaoyu: whats the benefit of grid over div?

src/components/swap/SwapBuyFiatButton.tsx
lynnshaoyu: right, but i guess what i'm saying is it NEEDS to be done the way it currently is being done, or else all our tests will break because of needing the other core functionalities of @web3-react/core in the tests 😬

how can this be done differently in a way that works?

src/components/Tooltip/index.tsx
lynnshaoyu: sounds good done

src/components/Popover/index.tsx
```diff
@@ -117,17 +122,36 @@ export default function Popover({
...
+      }
+    } else {
+      setShouldShow(false)
+      return
+    }
+  }, [show, delayShowTimeout])
 
   return showInline ? (
-    <PopoverContainer show={show}>{content}</PopoverContainer>
+    <PopoverContainer show={shouldShow}>{content}</PopoverContainer>
```
lynnshaoyu, line 147: we no longer need to do `show && shouldShow` here now, shouldShow is the single source of truth for state here

src/components/Popover/index.tsx
lynnshaoyu: Okayyy actually upon second look, we do not need to use the variable `show` anymore in line 147 and line 154, we can just use `shouldShow`. I changed it so the useEffect now massages all information into making shouldShow the single source of truth for when the popover should appear and disappear. 

IMO this makes it simple/elegant enough to merge in 🤔 @zzmp @tinaszheng 

see comment here: https://github.com/Uniswap/interface/pull/6240/files#r1159048428

src/components/Popover/index.tsx
```diff
@@ -81,13 +81,16 @@ export interface PopoverProps {
   offsetY?: number
   hideArrow?: boolean
   showInline?: boolean
+  // time delay in milliseconds before the popover shows
```
zzmp, line 84: This should be JSDoc:

```suggestion
  /** time delay in milliseconds before the popover shows */
```

src/components/Popover/index.tsx
```diff
@@ -98,6 +101,8 @@ export default function Popover({
   const [referenceElement, setReferenceElement] = useState<HTMLDivElement | null>(null)
   const [popperElement, setPopperElement] = useState<HTMLDivElement | null>(null)
   const [arrowElement, setArrowElement] = useState<HTMLDivElement | null>(null)
+  // In default case where delayShowTimeout is 0 (no delay before showing) shouldShow initializes to true.
+  const [shouldShow, setShouldShow] = useState(Boolean(delayShowTimeout))
```
zzmp, line 105: Right now, it will initialize to `false`. I think you want:

```suggestion
  const [shouldShow, setShouldShow] = useState(show && delayShowTimeout === 0)
```

src/components/Popover/index.tsx
```diff
@@ -117,17 +122,36 @@ export default function Popover({
...
+  useEffect(() => {
+    if (show) {
+      if (!delayShowTimeout) {
+        setShouldShow(true)
+        return
+      }
+      const tooltipTimer = setTimeout(() => {
+        setShouldShow(true)
+      }, delayShowTimeout)
+
```
zzmp, line 136: Move the newline to delineate the two cases:

```suggestion

      const tooltipTimer = setTimeout(() => {
        setShouldShow(true)
      }, delayShowTimeout)
```

src/components/Popover/index.tsx
```diff
@@ -117,17 +122,36 @@ export default function Popover({
   const updateCallback = useCallback(() => {
     update && update()
   }, [update])
-  useInterval(updateCallback, show ? 100 : null)
+  useInterval(updateCallback, show ? 200 : null)
```
zzmp, line 125: Why are you updating this interval?

```suggestion
  useInterval(updateCallback, show ? 100 : null)
```

src/state/user/reducer.ts
```diff
@@ -9,6 +9,8 @@ import { SerializedPair, SerializedToken } from './types'
 const currentTimestamp = () => new Date().getTime()
 
 export interface UserState {
+  buyFiatFlowCompleted: boolean | undefined
```
zzmp, line 12: Simplify (because `undefined` is your stand-in for `false`, and a user can never set this to `false`), and propagate through the PR:

```suggestion
  buyFiatFlowCompleted?: true
```

src/state/user/reducer.ts
```diff
@@ -75,6 +78,9 @@ const userSlice = createSlice({
   name: 'user',
   initialState,
   reducers: {
+    updateUserBuyFiatFlowCompleted(state, action) {
+      state.buyFiatFlowCompleted = action.payload
+    },
```
zzmp, line 83:
```suggestion
    updateUserBuyFiatFlowCompleted(state) {
      state.buyFiatFlowCompleted = true
    },
```

src/state/user/hooks.tsx
```diff
@@ -68,6 +69,18 @@ export function useIsExpertMode(): boolean {
   return useAppSelector((state) => state.user.userExpertMode)
 }
 
+export function useBuyFiatFlowCompleted(): [boolean | undefined, (buyFiatFlowCompleted: boolean) => void] {
+  const dispatch = useAppDispatch()
+  const buyFiatFlowCompleted = useAppSelector((state) => state.user.buyFiatFlowCompleted)
+  const setBuyFiatFlowCompleted = useCallback(
+    (buyFiatFlowCompleted: boolean) => {
+      dispatch(updateUserBuyFiatFlowCompleted(buyFiatFlowCompleted))
```
zzmp, line 77:
```suggestion
export function useBuyFiatFlowCompleted(): [boolean, () => void] {
  const dispatch = useAppDispatch()
  const buyFiatFlowCompleted = Boolean(useAppSelector((state) => state.user.buyFiatFlowCompleted))
  const setBuyFiatFlowCompleted = useCallback(
    () => {
      dispatch(updateUserBuyFiatFlowCompleted())
```

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,141 @@
...
+  gap: 4px;
+  &:focus {
+    text-decoration: none;
+  }
+  &:active {
+    text-decoration: none;
+  }
+`
+
+const POPOVER_DELAY_BEFORE_SHOW_MS = 500
```
zzmp, line 43: Is there any case where this would not be 500? In which case, could this constant be moved into the Tooltip, and the "delayTimeout" be turned into a boolean?

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+export default function SwapBuyFiatButton() {
+  // Importing useWeb3React here is necessary for mocking different behavior in different tests in SwapBuyFiatButton.test.tsx
+
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const { useWeb3React } = require('@web3-react/core')
```
zzmp, line 49: This should be imported normally, not through require. What was the reasoning for requiring?

zzmp (changes requested):
Could you include a unit test for this new Popover behavior?
Alternatively, I still think it'd be better/easier to just remove the delay.

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,141 @@
...
+  const { account } = useWeb3React()
+  const openFiatOnRampModal = useOpenModal(ApplicationModal.FIAT_ONRAMP)
+  const [buyFiatFlowCompleted, setBuyFiatFlowCompleted] = useBuyFiatFlowCompleted()
+  const [checkFiatRegionAvailability, setCheckFiatRegionAvailability] = useState(false)
```
zzmp, line 53: `is`/`should` in the boolean variables gating behavior makes this easier to follow:

```suggestion
  const [isBuyFiatFlowCompleted, setBuyFiatFlowCompleted] = useBuyFiatFlowCompleted()
  const [shouldCheckFiatRegionAvailability, setCheckFiatRegionAvailability] = useState(false)
```

Also below, `isWalletDrawerOpen`.

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,141 @@
...
+`
+
+export const MOONPAY_REGION_AVAILABILITY_ARTICLE =
+  'https://support.uniswap.org/hc/en-us/articles/11306664890381-Why-isn-t-MoonPay-available-in-my-region-'
+
+enum BuyFiatFlowState {
+  // Default initial state. User is not actively trying to buy fiat.
+  INACTIVE,
+  // Buy fiat flow is active and region availability has been checked.
+  ACTIVE_POST_REGION_CHECK,
```
zzmp, line 27:
```suggestion
  ACTIVE_CHECKING_REGION,
```

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,141 @@
...
+    loading: fiatOnrampAvailabilityLoading,
+  } = useFiatOnrampAvailability(checkFiatRegionAvailability)
+  const [buyFiatFlowState, setBuyFiatFlowState] = useState(BuyFiatFlowState.INACTIVE)
+  const [walletDrawerOpen, toggleWalletDrawer] = useWalletDrawer()
+
+  /*
+   * Once the buy fiat flow has been completed, reset BuyFiatFlowState and also save to
+   * local storage that user has completed this flow already.
+   */
+  const terminateBuyFiatFlow = useCallback(() => {
```
zzmp, line 66: Can you move this into the other callback? I think it'll be clearer if all of the business logic is in one place. It's very confusing to me that you are calling `setBuyFiatFlowState` from multiple handlers.

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,141 @@
...
+export const MOONPAY_REGION_AVAILABILITY_ARTICLE =
+  'https://support.uniswap.org/hc/en-us/articles/11306664890381-Why-isn-t-MoonPay-available-in-my-region-'
+
+enum BuyFiatFlowState {
+  // Default initial state. User is not actively trying to buy fiat.
+  INACTIVE,
+  // Buy fiat flow is active and region availability has been checked.
+  ACTIVE_POST_REGION_CHECK,
+  // Buy fiat flow is active, feature is available in user's region & needs wallet connection.
+  ACTIVE_NEEDS_WALLET_CONNECTION,
```
zzmp, line 29:
```suggestion
  ACTIVE_NEEDS_ACCOUNT,
```

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,141 @@
...
+    walletDrawerOpen,
+    toggleWalletDrawer,
+    openFiatOnRampModal,
+    terminateBuyFiatFlow,
+  ])
+
+  // Continue buy fiat flow automatically when requisite state changes have occured.
+  useEffect(() => {
+    if (
+      buyFiatFlowState === BuyFiatFlowState.ACTIVE_POST_REGION_CHECK ||
```
zzmp, line 107: Should this occur if `buyFiatFlowState === BuyFiatFlowState.ACTIVE_POST_REGION_CHECK && fiatOnrampAvailable`, because that check occurs asynchronously and you may not know if `fiatOnrampAvailable` at first?

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -0,0 +1,141 @@
...
+      (account && buyFiatFlowState === BuyFiatFlowState.ACTIVE_NEEDS_WALLET_CONNECTION)
+    ) {
+      handleBuyCrypto()
+    }
+  }, [account, handleBuyCrypto, buyFiatFlowState])
+
+  const buyCryptoButtonDisabled =
+    (!fiatOnrampAvailable && fiatOnrampAvailabilityChecked) ||
```
zzmp, line 115: This is re-used in `disabled` - I think it could be pulled into a semantic variable to improve readability.

zzmp (changes requested):

src/components/swap/SwapBuyFiatButton.tsx
tinaszheng: well basically you want to just mock all the imports that are actually used in the flow, right? so in your above example, it's mocking the return value of `useWeb3React` imported from `@web3-react/core`, very similar to my example above where `ModuleName` is just the name of the import you want to mock. so one way you can figure out which ones you need to mock is to just run the test and see which imports are giving you errors, and then add those imports to the mocked object

src/components/Popover/index.tsx
lynnshaoyu: oof yea thats right done

src/components/swap/SwapBuyFiatButton.tsx
lynnshaoyu: thats inaccurate because this is the state when the region check is done.

src/components/swap/SwapBuyFiatButton.tsx
lynnshaoyu: done ty for showing how to get the unit tests to work in a different way without needing to do this!!!

src/components/swap/SwapBuyFiatButton.tsx
lynnshaoyu: uhhh this was originally in the other callback, it was asked to be separated out by @tinaszheng.

src/components/Popover/index.tsx
lynnshaoyu: deleted the delay

src/components/Popover/index.tsx
lynnshaoyu: done

src/components/Popover/index.tsx
lynnshaoyu: deleted the delay

src/components/Popover/index.tsx
lynnshaoyu: undone

src/components/swap/SwapBuyFiatButton.tsx
lynnshaoyu: deleted

src/components/swap/SwapBuyFiatButton.tsx
lynnshaoyu: react useState convention would necessitate changing the set functions to "setIsBuyFiatFlowCompleted" and "setShouldCheckFiatRegionAvailability", which seem overly wordy and confusing to me. also the use of "is" for booleans is highly inconsistent across the codebase so I dont really agree with needing this change.

src/components/Popover/index.tsx
lynnshaoyu: nvm removed this delay altogether

src/components/Popover/index.tsx
```diff
@@ -11,7 +11,7 @@ const PopoverContainer = styled.div<{ show: boolean }>`
   pointer-events: none;
   visibility: ${(props) => (props.show ? 'visible' : 'hidden')};
   opacity: ${(props) => (props.show ? 1 : 0)};
-  transition: visibility 150ms linear, opacity 150ms linear;
+  transition: visibility 125ms ease-in-out, opacity 125ms ease-in-out;
```
zzmp, line 14: Might want to back this out and include it in a future Popover PR.

src/components/Tooltip/index.tsx
```diff
@@ -25,25 +29,26 @@ interface TooltipProps extends Omit<PopoverProps, 'content'> {
...
   close?: () => void
-  noOp?: () => void
   disableHover?: boolean // disable the hover and content display
   timeout?: number
 }
 
 interface TooltipContentProps extends Omit<PopoverProps, 'content'> {
   content: ReactNode
   onOpen?: () => void
+  open?: () => void
```
zzmp, line 39: How is this different from `onOpen`? Why aren't these labeled as callbacks (eg `onOpen`/`onClose`)?

src/components/Tooltip/index.tsx
```diff
@@ -53,15 +58,28 @@ export default function Tooltip({ text, open, close, noOp, disableHover, ...rest
...
+    />
+  )
 }
 
 /** Standard text tooltip. */
 export function MouseoverTooltip({ text, disableHover, children, timeout, ...rest }: Omit<TooltipProps, 'show'>) {
   const [show, setShow] = useState(false)
-  const open = useCallback(() => text && setShow(true), [text, setShow])
-  const close = useCallback(() => setShow(false), [setShow])
+  const open = () => text && setShow(true)
```
zzmp, line 81: Shouldn't this still depend on `text`?

src/components/swap/SwapBuyFiatButton.tsx
zzmp: This is separating the flow logic into two places, while the majority of it is in `handleBuyCrypto`. I think it should be put back :(

zzmp (changes requested):

src/components/swap/SwapBuyFiatButton.tsx
lynnshaoyu: huddled offline - you're right, region check is async so technically it is still checking the region during this step.

zzmp (approved):

PR #6243, opened 2023-03-23
feat: sort tokens in selector by USD value

just-toby:
## Description

When loading the token selector, fetches the USD prices for all tokens with a balance so we can take into account the USD value when sorting.

https://uniswaplabs.atlassian.net/browse/WEB-2990?atlOrigin=eyJpIjoiZTYzZjZkMzRhNzI0NGQxNWI1MzIxYTJkMzUxZDJkMjkiLCJwIjoiaiJ9

## Screen Capture
|  image | image  |

#### Manual

- open the token selector on all Supported Chains and make sure the set of tokens with balances is unchanged, but the order is updated

#### Automated
- [x] Unit test - added for the sorting helper function
- [x] Integration/E2E test - we have e2e test (`swap-widget.cy.ts`) which test opening this component so it's covered a little bit

### DIFF
--- a/src/components/SearchModal/CurrencySearch.tsx
+++ b/src/components/SearchModal/CurrencySearch.tsx
@@ -15,7 +15,7 @@ import { ChangeEvent, KeyboardEvent, RefObject, useCallback, useEffect, useMemo,
 import AutoSizer from 'react-virtualized-auto-sizer'
 import { FixedSizeList } from 'react-window'
 import { Text } from 'rebass'
-import { useAllTokenBalances } from 'state/connection/hooks'
+import { useAllTokenBalances, useTokenPrices } from 'state/connection/hooks'
 import styled, { useTheme } from 'styled-components/macro'
 import { UserAddedToken } from 'types/tokens'
 
@@ -91,9 +91,11 @@ export function CurrencySearch({
   }, [defaultTokens, debouncedQuery])
 
   const [balances, balancesAreLoading] = useAllTokenBalances()
+  const balancePrices = useTokenPrices(balances)
+
   const sortedTokens: Token[] = useMemo(
     () =>
-      !balancesAreLoading
+      !balancesAreLoading && balancePrices
         ? filteredTokens
             .filter((token) => {
               if (onlyShowCurrenciesWithBalance) {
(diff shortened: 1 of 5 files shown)


src/components/SearchModal/CurrencySearch.tsx
```diff
@@ -91,9 +91,11 @@ export function CurrencySearch({
   }, [defaultTokens, debouncedQuery])
 
   const [balances, balancesAreLoading] = useAllTokenBalances()
+  const balancePrices = useTokenPrices(balances)
```
just-toby, line 94: maybe a perf hit for the UX since we're waiting for price queries now - we can't parallelize because we need to know which tokens have a balance and need prices

src/lib/hooks/useTokenList/sorting.ts
```diff
@@ -1,23 +1,43 @@
...
 
 /** Sorts currency amounts (descending). */
 function balanceComparator(a?: CurrencyAmount<Currency>, b?: CurrencyAmount<Currency>) {
+  return valueComparator(a, b) // defaults to price of 1, so this will just compare the balances directly
+}
+
+/** Sorts currency amounts by value (descending). */
+function valueComparator(a?: CurrencyAmount<Currency>, b?: CurrencyAmount<Currency>, aPrice?: number, bPrice?: number) {
+  const aBalance = currencyAmountToPreciseFloat(a) ?? 1
+  const bBalance = currencyAmountToPreciseFloat(b) ?? 1
```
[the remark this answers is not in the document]
JFrankfurt: `currencyAmountToPreciseFloat` is a formatting util used for string representation--it shouldn't be used for logical comparison

src/state/connection/hooks.ts
```diff
@@ -14,11 +15,35 @@ export {
...
+    async function loadPrices() {
+      const tokensWithBalance: CurrencyAmount<Token>[] = Object.values(balances).filter(
+        (balance) => balance?.greaterThan(0) // filters undefined values so we can cast safely
+      ) as CurrencyAmount<Token>[]
+      const pricesEntries = await Promise.all(
+        tokensWithBalance.map(async (token) => {
+          const price = await fetchUSDPrice(token)
+          return [token.currency.address, price]
+        })
+      )
```
JFrankfurt, line 43: This feels sketchy to me. Can you take a look at the way @cartcrom accomplished this in the mini portfolio? I think you'll be able to fetch a user's value-ordered balances using a single query instead of `Promise.all`'ing a potentially very large set of single token price requests.

JFrankfurt (changes requested):

JFrankfurt:
image
where did the $LINK dust go here?

cmcewen (changes requested):

just-toby:
i'm closing this and will re-open when the backend is ready to serve portfolio balances at a high volume

PR #6248, opened 2023-03-24
fix: modify mini portfolio pools row event properties

lynnshaoyu:
was sending in entire token object previously, modify to just address and symbol.

### DIFF
--- a/src/components/WalletDropdown/MiniPortfolio/Pools/index.tsx
+++ b/src/components/WalletDropdown/MiniPortfolio/Pools/index.tsx
@@ -118,7 +118,13 @@ function PositionListItem({ positionInfo }: { positionInfo: PositionInfo }) {
       events={[BrowserEvent.onClick]}
       name={SharedEventName.ELEMENT_CLICKED}
       element={InterfaceElementName.MINI_PORTFOLIO_POOLS_ROW}
-      properties={{ chain_id: chainId, pool_token_0: pool.token0, pool_token_1: pool.token1 }}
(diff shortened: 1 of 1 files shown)


src/components/WalletDropdown/MiniPortfolio/Pools/index.tsx
```diff
@@ -118,7 +118,13 @@ function PositionListItem({ positionInfo }: { positionInfo: PositionInfo }) {
...
       name={SharedEventName.ELEMENT_CLICKED}
       element={InterfaceElementName.MINI_PORTFOLIO_POOLS_ROW}
-      properties={{ chain_id: chainId, pool_token_0: pool.token0, pool_token_1: pool.token1 }}
+      properties={{
+        chain_id: chainId,
+        pool_token_0_symbol: pool.token0.symbol,
+        pool_token_1_symbol: pool.token1.symbol,
+        pool_token_0_address: pool.token0.address,
+        pool_token_1_address: pool.token1.address,
+      }}
```
JFrankfurt, line 127: should this be wrapped in `useMemo`?

JFrankfurt: https://github.com/Uniswap/analytics/blob/main/src/analytics/TraceEvent.tsx#L21

lynnshaoyu: ty for pointing this out!!

PR #6249, opened 2023-03-24
feat: reduce severity of phishing filter to allow url token names

JFrankfurt:
CX team has been getting some inbound regarding non-phishing tokens with domains as token symbol. This uses a more complicated approach that:
1. permits LP positions with any pair of tokens on the Labs token list
2. permits one url if one token is on the token list
3. disallows a pair of non-list tokens where either has a url symbol

https://uniswaplabs.atlassian.net/browse/WEB-2802?atlOrigin=eyJpIjoiMDM1OGMyZTg4ODMzNGFhMTk2NTQ4YjhjZTg1Y2UxNTkiLCJwIjoiaiJ9

test by walletconnecting with `0x11E4857Bb9993a50c685A79AFad4E6F65D518DDa` and viewing LP positions on Optimism


src/components/PositionListItem/index.tsx
```diff
@@ -203,10 +203,7 @@ export default function PositionListItem({
 
   const removed = liquidity?.eq(0)
 
-  const containsURL = useMemo(
-    () => [token0?.name, token0?.symbol, token1?.name, token1?.symbol].some((testString) => hasURL(testString)),
-    [token0?.name, token0?.symbol, token1?.name, token1?.symbol]
-  )
+  const containsURL = hasURL(token0?.symbol) || hasURL(token1?.symbol)
```
cmcewen, line 206: longer term is this something we plan to store a list in a DB or use another mechanism? i can understand that most URLs are probably not great but also it doesn't seem like 100%, and we are probably also not catching many phishing tokens that have non-url names

JFrankfurt: Here's the original jira issue:
https://uniswaplabs.atlassian.net/browse/WEB-2802?focusedCommentId=16655

and the discussion thread:
https://uniswapteam.slack.com/archives/C03BSGBKWA0/p1674574970331509

(note the false positives mentions--determination at the time was that this was an acceptable step)

There's no plan to add a "this is a scam" data source, and I don't think that we want to start bailing water on that particular sinking ship.  Annabel and I discussed the tradeoffs here and we're going to scale the current filter back to just symbols for now.

I'm thinking of two additional measures we might implement (both are simple enough that I could add them to this draft, imo):
1. check for presence of one or both LP tokens in a token list we index already. (possible perf issue w/ coingecko? will see)
2. add an analytics call to notify us of pairs that are blocked proactively instead of waiting for CX inbound.

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,62 @@
...
+
+const ERC20Interface = new Interface(ERC20ABI) as Erc20Interface
+
+export function useFilterPossiblyMaliciousPositions(positions: PositionDetails[]): PositionDetails[] {
+  const tokens = useDefaultActiveTokens()
+
+  const nonListPositionTokenAddresses = useMemo(
+    () =>
+      Array.from(
+        new Set(positions.reduce((acc, position) => acc.concat(position.token0, position.token1), [] as string[]))
```
JFrankfurt, line 20: `Set` deduplicates address set (i.e., if you've got 50 positions with the `WETH` address

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,62 @@
...
+      ).filter((address) => !tokens[address]),
```
JFrankfurt, line 21: remove tokens not in the list

src/components/PositionListItem/index.tsx
```diff
@@ -203,15 +202,6 @@ export default function PositionListItem({
...
 
-  const containsURL = useMemo(
-    () => [token0?.name, token0?.symbol, token1?.name, token1?.symbol].some((testString) => hasURL(testString)),
-    [token0?.name, token0?.symbol, token1?.name, token1?.symbol]
-  )
-
-  if (containsURL) {
-    return null
-  }
-
```
JFrankfurt, line 214: moved up to the parent so we can count the number rendered

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  const nonListPositionTokenAddresses = useMemo(
+    () =>
+      Array.from(
+        new Set(positions.reduce((acc, position) => acc.concat(position.token0, position.token1), [] as string[]))
+      ).filter((address) => !tokenList[address]),
+    [positions, tokenList]
+  )
+
+  const results = useMultipleContractSingleData(nonListPositionTokenAddresses, ERC20Interface, 'symbol')
```
JFrankfurt, line 25: call current chain rpc for info on addresses we don't know already from the list

src/hooks/useFilterPossiblyMaliciousPositions.ts
cmcewen: can you make this a separate function? like `getUniqueAddressesFromPositions` or something?

src/hooks/useFilterPossiblyMaliciousPositions.ts
cmcewen: also feel like this should be a separate function here? like there should be a general function that gives you token data without caring if you have it cached or not, which is separate than the purpose of this function

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,58 @@
+import { Interface } from '@ethersproject/abi'
+import { Token } from '@uniswap/sdk-core'
+import ERC20ABI from 'abis/erc20.json'
+import { Erc20Interface } from 'abis/types/Erc20'
+import { useMultipleContractSingleData } from 'lib/hooks/multicall'
+import { useMemo } from 'react'
+import { PositionDetails } from 'types/position'
+import { hasURL } from 'utils/urlChecks'
+
+import { useDefaultActiveTokens } from './Tokens'
```
grabbou, line 10: Is this a typo or expected that a file in `hooks` folder (not a component?) is uppercase?

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,58 @@
...
+  const addressesToSymbol: Record<string, string | undefined> = useMemo(
+    () =>
+      nonListPositionTokenAddresses.reduce(
+        (acc, address, i) => ({
+          ...acc,
+          [address]: results[i].result as string | undefined,
+        }),
+        {}
+      ),
+    [nonListPositionTokenAddresses, results]
```
grabbou, line 35: Are we sure that `useMultipleContractSingleData` will return same data if its arguments did not change? It sounds to me like it will always run. 

Maybe, as an alternative, we should put it inside this `useMemo` hook. So that we only re-run when `nonListPositionTokenAddresses` has changed.

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,58 @@
...
+  )
+  return useMemo(
+    () =>
+      positions.filter((position) => {
+        const token0FromList = tokenList[position.token0] as Token | undefined
+        const token1FromList = tokenList[position.token1] as Token | undefined
```
cmcewen, line 41: i think `activeTokenList` vs `tokenList` would make it more clear this is default tokens + user added tokens

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,58 @@
...
+      Array.from(
+        new Set(positions.reduce((acc, position) => acc.concat(position.token0, position.token1), [] as string[]))
+      ).filter((address) => !tokenList[address]),
+    [positions, tokenList]
+  )
+
+  const results = useMultipleContractSingleData(nonListPositionTokenAddresses, ERC20Interface, 'symbol')
+  const addressesToSymbol: Record<string, string | undefined> = useMemo(
+    () =>
+      nonListPositionTokenAddresses.reduce(
```
grabbou, line 28: nit: How about using `for/loop` here instead of reduce to avoid spreading objects in every iteration?

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,58 @@
...
+        (acc, address, i) => ({
+          ...acc,
+          [address]: results[i].result as string | undefined,
```
grabbou, line 31: When can `result` be undefined? Symbol not found? Call failed? Maybe we shouldn't include it in the object at all then? That way, check for presence instead of `undefined`.

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,58 @@
...
+      ),
+    [nonListPositionTokenAddresses, results]
+  )
+  return useMemo(
+    () =>
+      positions.filter((position) => {
+        const token0FromList = tokenList[position.token0] as Token | undefined
+        const token1FromList = tokenList[position.token1] as Token | undefined
+        const bothTokensInList = token0FromList && token1FromList
+        if (bothTokensInList) return true
```
grabbou, line 43: nit: Personally, I would recommend an inline `if` statement w/o a dedicated variable and a comment block with explanation if considered potentially confusing.

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,58 @@
...
+
+        const token0IsInList = !!token0FromList
+        const token1IsInList = !!token1FromList
+        const atLeastOneTokenIsInList = token0IsInList || token1IsInList
```
cmcewen, line 47: how about doing math and making it `tokensInList` = 0, 1, 2? then you can use it below instead of `maxOneUrlTokenSymbol`

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,58 @@
...
+        const token0HasUrlSymbol = hasURL(token0IsInList ? token0FromList.symbol : addressesToSymbol[position.token0])
+        const token1HasUrlSymbol = hasURL(token1IsInList ? token1FromList.symbol : addressesToSymbol[position.token1])
+        const maxOneUrlTokenSymbol = token0HasUrlSymbol ? !token1HasUrlSymbol : true
+        if (atLeastOneTokenIsInList && maxOneUrlTokenSymbol) return true
```
grabbou, line 51: Correct me if I am wrong, but can this entire `useMemo` be simplified to something like this:
```ts
// tokenSymbols contains both symbols from tokenList and RPC call
positions.filter(({token0, token1}) => {
   // both tokens are in the list
   if (tokenSymbols[token0] && tokenSymbols[token1]) { 
      return true
   }
   
   // none of the tokens are in the list
   if (!tokenSymbols[token0] && !tokenSymbols[token1]) {
      return false
   }

  // at this point "atLeastOneTokenIsInList" or both are in the list
  // I know this is a bit code-smell, but still better than ternary. 
  // "maxOne" - sum of two booleans should be 1
  //  "neitherTokenHasAUrlSymbol" - sum of two booleans will be 0
   if (Number(hasURL(tokenSymbols[token0])) + Number(hasURL(tokenSymbols[token1])) <= 1) {
      return true
   }

  // is false the default? when it can happen?
  return false
})
```

grabbou (changes requested):
Hey! Thanks for the PR. Just left some comments around booleans and if conditionals. Ideally, I'd like to see the clearness of this code to improve as much as possible. At the same time, I am fairly new to that particular code area, so please double-check my code suggestions for any inconsistencies.

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,58 @@
...
+          ...acc,
+          [address]: results[i].result as string | undefined,
+        }),
+        {}
+      ),
+    [nonListPositionTokenAddresses, results]
+  )
+  return useMemo(
+    () =>
+      positions.filter((position) => {
```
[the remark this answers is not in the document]
JFrankfurt: This did cross my mind while implementing, but in general my heuristic for loop perf is that constant scalars are generally okay if the result improves readability.

image

src/hooks/useFilterPossiblyMaliciousPositions.ts
JFrankfurt: Are you imagining something like this?

```
function useTokenContractsConstant(tokens: string[], field: 'name' | 'symbol' | 'decimals' | 'totalSupply') {
  return useMultipleContractSingleData(tokens, ERC20Interface, field, undefined, {
    blocksPerFetch: 10_000,
  })
}
```

grabbou: The only problem is that this is nothing more than a direct use of multi call API which is a hook! And I think at the end of the day this is a pure function, just that it's a hook b/c it uses some hook under the hood. And that's what makes it tricky to cache properly, b/c you can't do that inside `useMemo`.

src/hooks/useTokenContractsConstant.ts
```diff
@@ -0,0 +1,14 @@
...
+import { useMemo } from 'react'
+
+const ERC20Interface = new Interface(ERC20ABI) as Erc20Interface
+
+export function useTokenContractsConstant(tokens: string[], field: 'name' | 'symbol' | 'decimals' | 'totalSupply') {
+  const memoizedTokens = useMemo(() => tokens, [tokens])
+  return useMultipleContractSingleData(memoizedTokens, ERC20Interface, field, undefined, {
+    blocksPerFetch: 10_000,
+  })
+}
```
JFrankfurt, line 14: @cmcewen @grabbou 

moving the convo here

this util feels a little weird to me because
- accessing ad-hoc chain data w/ multicall is a pretty normal pattern
- wrapping the array to enforce a possibly-already-memo'd array is memo'd before the hook call when a user could have just written these two lines themselves feels jank

that said, I do like enforcing the very long re-fetch time on a subset of chain data that we can reasonably expect to never change!

src/hooks/useFilterPossiblyMaliciousPositions.ts
JFrankfurt: I agree this is weird and should be changed

src/components/WalletDropdown/MiniPortfolio/Pools/index.tsx
```diff
@@ -22,24 +22,42 @@ import { PositionInfo } from './cache'
...
+
+  return useMemo(
+    () => safePositionDetails.map((positionDetails) => tokenIdsToPositionInfo[positionDetails.tokenId.toString()]),
+    [safePositionDetails, tokenIdsToPositionInfo]
+  )
+}
+
 export default function Pools({ account }: { account: string }) {
   const { positions, loading } = useMultiChainPositions(account)
+  const safePositions = useFilterPossiblyMaliciousPositions(positions)
```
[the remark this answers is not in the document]
JFrankfurt: are you asking for a start on tests of `useMultiChainPositions ` or tests of `useFilterPossiblyMaliciousPositions`?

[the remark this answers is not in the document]
cmcewen: the second one - `useFilterPossiblyMaliciousPositions`

src/hooks/useFilterPossiblyMaliciousPositions.ts
JFrankfurt: resolving--superseded by @cmcewen 's `tokensInListCount` suggestion

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,58 @@
...
+        const token0FromList = tokenList[position.token0] as Token | undefined
+        const token1FromList = tokenList[position.token1] as Token | undefined
+        const bothTokensInList = token0FromList && token1FromList
+        if (bothTokensInList) return true
+
+        const token0IsInList = !!token0FromList
```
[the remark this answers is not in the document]
JFrankfurt: we could update the `const addressesToSymbol = useMemo(() ...` line to be something like this:

```
  useEffect(() => {
    for (let i = 0; i < nonListPositionTokenAddresses.length; i++) {
      const callResult = symbols[i].result
      if (!callResult) continue
      const address = nonListPositionTokenAddresses[i]
      activeTokensList[address].symbol = callResult as unknown as string
    }
  }, [activeTokensList, nonListPositionTokenAddresses, symbols])
```

and compress the data into the `activeTokensList[address].symbol` but we'd lose the information about list inclusion which we're using to make a decision about whether or not to allow a url symbol (see the first boolean return in the returned memo callback).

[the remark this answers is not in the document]
JFrankfurt: Given this feedback I've updated the `hasURL` call to be a little simpler so it requires one less var

```
        const token0HasUrlSymbol = hasURL(token0FromList?.symbol ?? addressesToSymbol[position.token0])
        const token1HasUrlSymbol = hasURL(token1FromList?.symbol ?? addressesToSymbol[position.token1])
```

src/hooks/useFilterPossiblyMaliciousPositions.ts
JFrankfurt: not quite

if neither token is in the list, then we don't allow any urls.
if one token is in the list, then we allow one url
if both tokens are in the list, then we allow two urls

re: `Number(hasURL(...)) <=1` I personally don't think relying on type coercion (`Number(true)===1`) is more clear than a standard logic language feature.

src/components/WalletDropdown/MiniPortfolio/Pools/index.tsx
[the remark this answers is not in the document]
JFrankfurt: done!

src/components/WalletDropdown/MiniPortfolio/Pools/index.tsx
```diff
@@ -22,24 +22,48 @@ import { PositionInfo } from './cache'
 import { useFeeValues } from './hooks'
 import useMultiChainPositions from './useMultiChainPositions'
 
+/*
+  This hook takes an array of PositionInfo objects (format used by the Uniswap Labs gql API).
+  The hook access PositionInfo.details (format used by the NFT position contract),
+  filters the PositionDetails data for malicious content,
+  and then returns the original data in its original format. 
+*/
+function useFilterPossiblyMaliciousPositionInfo(positions: PositionInfo[] | undefined): PositionInfo[] | undefined {
```
grabbou, line 31: When this function returns `undefined`? It is not clear for me from the code. Also, do we need explicit return type definition? TypeScript should infer it just right.

Also, if we really return `undefined`, I would recommend to not return `undefined`, but an empty array.

src/components/WalletDropdown/MiniPortfolio/Pools/index.tsx
```diff
@@ -22,24 +22,48 @@ import { PositionInfo } from './cache'
...
 export default function Pools({ account }: { account: string }) {
   const { positions, loading } = useMultiChainPositions(account)
+  const filteredPositions = useFilterPossiblyMaliciousPositionInfo(positions)
   const [showClosed, toggleShowClosed] = useReducer((showClosed) => !showClosed, false)
 
   const [openPositions, closedPositions] = useMemo(() => {
     const openPositions: PositionInfo[] = []
     const closedPositions: PositionInfo[] = []
-    positions?.forEach((position) => (position.closed ? closedPositions : openPositions).push(position))
+    filteredPositions?.forEach((position) => (position.closed ? closedPositions : openPositions).push(position))
```
grabbou, line 56: nit: how about using a simple for loop with a regular `if` statement?
```
if (position.closed) {
  closedPositions.push(position)
} else {
  openPositions.push(position)
}
```

src/components/WalletDropdown/MiniPortfolio/Pools/index.tsx
```diff
@@ -22,24 +22,48 @@ import { PositionInfo } from './cache'
...
     return [openPositions, closedPositions]
-  }, [positions])
+  }, [filteredPositions])
 
   const toggleWalletDrawer = useToggleWalletDrawer()
 
-  if (!positions || loading) {
+  if (!filteredPositions || loading) {
```
grabbou, line 62: not related to this PR, but general question: why do we render skeleton when this is undefined? Does that mean that filtered positions are still loading?

src/components/WalletDropdown/MiniPortfolio/Pools/index.tsx
```diff
@@ -22,24 +22,48 @@ import { PositionInfo } from './cache'
...
     return <PortfolioSkeleton />
   }
 
-  if (positions?.length === 0) {
+  if (filteredPositions?.length === 0) {
```
grabbou, line 66: Why is `?` needed here if we check for `!filteredPositions` earlier?

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,56 @@
...
+import { useDefaultActiveTokens } from './Tokens'
+import { useTokenContractsConstant } from './useTokenContractsConstant'
+
+function getUniqueAddressesFromPositions(positions: PositionDetails[]): string[] {
+  return Array.from(
+    new Set(positions.reduce<string[]>((acc, position) => acc.concat(position.token0, position.token1), []))
+  )
+}
+
+export function useFilterPossiblyMaliciousPositions(positions: PositionDetails[]): PositionDetails[] {
```
grabbou, line 15: nit: would be great to have a description here, what this function does, esp. in context of checking for presence in activeTokensList

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,56 @@
...
+  const nonListPositionTokenAddresses = useMemo(
+    () => getUniqueAddressesFromPositions(positions).filter((address) => !activeTokensList[address]),
+    [positions, activeTokensList]
+  )
+
+  const symbolCallStates = useTokenContractsConstant(nonListPositionTokenAddresses, 'symbol')
+
+  const addressesToSymbol: Record<string, string> = useMemo(() => {
+    const result: Record<string, string> = {}
+    for (let i = 0; i < nonListPositionTokenAddresses.length; i++) {
```
grabbou, line 27: nit: prefer `for... of` instead

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export function useFilterPossiblyMaliciousPositions(positions: PositionDetails[]): PositionDetails[] {
+  const activeTokensList = useDefaultActiveTokens()
+
+  const nonListPositionTokenAddresses = useMemo(
+    () => getUniqueAddressesFromPositions(positions).filter((address) => !activeTokensList[address]),
+    [positions, activeTokensList]
+  )
+
+  const symbolCallStates = useTokenContractsConstant(nonListPositionTokenAddresses, 'symbol')
```
grabbou, line 23: nit: what does `call states` mean in this context? To me, this is just `nonListTokenSymbols`

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  const addressesToSymbol: Record<string, string> = useMemo(() => {
+    const result: Record<string, string> = {}
+    for (let i = 0; i < nonListPositionTokenAddresses.length; i++) {
+      const callResult = symbolCallStates[i].result
+      if (!callResult) continue
+      const address = nonListPositionTokenAddresses[i]
+      result[address] = callResult as unknown as string
```
grabbou, line 31: nit: why casting is necessary here? can we improve the underlying hook?

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  return useMemo(
+    () =>
+      positions.filter((position) => {
+        let tokensInListCount = 0
+        const token0FromList = activeTokensList[position.token0] as Token | undefined
+        const token1FromList = activeTokensList[position.token1] as Token | undefined
+        if (token0FromList) tokensInListCount++
+        if (token1FromList) tokensInListCount++
+        if (tokensInListCount === 2) return true
```
grabbou, line 44: nit: would be great to annotate each `if` condition with a description of what is really happening here. I have to admit, personally, following this code is hard for me to understand.

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,56 @@
...
+      result[address] = callResult as unknown as string
+    }
+    return result
+  }, [nonListPositionTokenAddresses, symbolCallStates])
+
+  return useMemo(
+    () =>
+      positions.filter((position) => {
+        let tokensInListCount = 0
+        const token0FromList = activeTokensList[position.token0] as Token | undefined
```
grabbou, line 40: nit: is type casting necessary here? can we fix the return type from the hook itself?

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,56 @@
...
+        const token1FromList = activeTokensList[position.token1] as Token | undefined
+        if (token0FromList) tokensInListCount++
+        if (token1FromList) tokensInListCount++
+        if (tokensInListCount === 2) return true
+
+        const token0HasUrlSymbol = hasURL(token0FromList?.symbol ?? addressesToSymbol[position.token0])
+        const token1HasUrlSymbol = hasURL(token1FromList?.symbol ?? addressesToSymbol[position.token1])
+        const maxOneUrlTokenSymbol = token0HasUrlSymbol ? !token1HasUrlSymbol : true
+        if (tokensInListCount >= 1 && maxOneUrlTokenSymbol) return true
```
grabbou, line 49: Can this condition be greater than `1` ever? Sounds like this `tokensInListCount` is either 0, 1 or 2. You already checked for `2` on line 44. When is `>=` satisfied then, except for `=== 1`?

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,56 @@
...
+
+        const neitherTokenHasAUrlSymbol = !token0HasUrlSymbol && !token1HasUrlSymbol
+        return neitherTokenHasAUrlSymbol
```
grabbou, line 52: nit: why declare variable and return it instead of single return? I would say if this was `tokenSymbolCount` as per previous comments, having a return `=== 0` + annotated with a comment would be much clearer. But that's my personal opinion.

src/hooks/useFilterPossiblyMaliciousPositions.ts
grabbou: If both elements are an array, could you do something like:
```
Object.fromEntries(nonListPositionTokenAddresses, symbolCallStates)
```
but you would need to make sure `symbolCallStates` is previously mapped to flat array, instead of having a nested `result`.

grabbou (commented):
Thanks for another pass. I am leaving a bunch of nits that could improve the readability. I will ask @tinaszheng to have a second pass at the PR and will defer the review to her. I am leaving my review as a "comment" to not block the release.

tinaszheng (changes requested):
+1 to mikes comments on readability 🤔

src/hooks/useFilterPossiblyMaliciousPositions.ts
JFrankfurt: this is an array of async action states from chain calls

```
interface CallState {
    readonly valid: boolean;
    readonly result: CallStateResult | undefined;
    readonly loading: boolean;
    readonly syncing: boolean;
    readonly error: boolean;
}
```

src/hooks/useFilterPossiblyMaliciousPositions.ts
JFrankfurt: yes

if we add a tsconfig rule that doesn't let you assume a key has a valid result then this won't be necessary, but adding that rules would require major refactors

src/components/AccountDrawer/MiniPortfolio/Pools/index.tsx
```diff
@@ -22,24 +22,55 @@ import { PositionInfo } from './cache'
 import { useFeeValues } from './hooks'
 import useMultiChainPositions from './useMultiChainPositions'
 
+/*
+  This hook takes an array of PositionInfo objects (format used by the Uniswap Labs gql API).
```
grabbou, line 26: nit: missing `*` on multiline

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,71 @@
...
+      positions.filter((position) => {
+        let tokensInListCount = 0
+        const token0FromList = activeTokensList[position.token0] as Token | undefined
+        const token1FromList = activeTokensList[position.token1] as Token | undefined
+        if (token0FromList) tokensInListCount++
+        if (token1FromList) tokensInListCount++
+        // if both tokens are in the list, then we let both have url symbols (so we don't check)
+        if (tokensInListCount === 2) return true
+
+        // check the token symbols to see if they contain a url
```
grabbou, line 57: nit: prefer `/* */` for multiline

grabbou: And capital letters 😁

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,71 @@
...
+import { hasURL } from 'utils/urlChecks'
+
+import { useDefaultActiveTokens } from './Tokens'
+import { useTokenContractsConstant } from './useTokenContractsConstant'
+
+function getUniqueAddressesFromPositions(positions: PositionDetails[]): string[] {
+  return Array.from(
+    new Set(positions.reduce<string[]>((acc, position) => acc.concat(position.token0, position.token1), []))
+  )
+}
```
grabbou, line 13: nit: empty line space between this and comment

grabbou (approved):

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,71 @@
+import { Token } from '@uniswap/sdk-core'
+import { useMemo } from 'react'
+import { PositionDetails } from 'types/position'
+import { hasURL } from 'utils/urlChecks'
+
+import { useDefaultActiveTokens } from './Tokens'
+import { useTokenContractsConstant } from './useTokenContractsConstant'
+
+function getUniqueAddressesFromPositions(positions: PositionDetails[]): string[] {
+  return Array.from(
```
cmcewen, line 10: up to your discretion, but this is a little more readable to me
```
const tokenAddressArray = positions.flatMap(pos => [pos.token0, pos.token1])
return [... new Set(tokenAddressArray)]
```

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,71 @@
...
+  )
+
+  const symbolCallStates = useTokenContractsConstant(nonListPositionTokenAddresses, 'symbol')
+
+  const addressesToSymbol: Record<string, string> = useMemo(() => {
+    const result: Record<string, string> = {}
+    for (let i = 0; i < nonListPositionTokenAddresses.length; i++) {
+      const callResult = symbolCallStates[i].result
+      if (!callResult) continue
+      const address = nonListPositionTokenAddresses[i]
```
cmcewen, line 40: complete side note - but i think having to correlate two separate arrays isn't necessarily ideal. this is a result of our contract call hooks so not worth doing anything here, but curious if we can make this a little nicer over time

src/hooks/useFilterPossiblyMaliciousPositions.ts
```diff
@@ -0,0 +1,71 @@
...
+        // prioritize the token entity from the list if it exists
+        // if the token isn't in the list, then use the data returned from chain calls
+        let urlSymbolCount = 0
+        if (hasURL(token0FromList?.symbol ?? addressesToSymbol[position.token0])) urlSymbolCount++
+        if (hasURL(token1FromList?.symbol ?? addressesToSymbol[position.token1])) urlSymbolCount++
+        // if one token is in the list, then one token can have a url symbol
+        if (tokensInListCount === 1 && urlSymbolCount < 2) return true
+
+        // if neither token is in the list, then neither can have a url symbol
+        return urlSymbolCount === 0
```
cmcewen, line 67: this block is very readable, nice 👍

cmcewen (approved):

PR #6250, opened 2023-03-24
feat: remove unicon tooltip entirely

just-toby:
removes the tooltip from Unicons in the Authenticated Header. this is a [design feedback request](https://www.notion.so/uniswaplabs/Get-rid-of-the-Unicon-tooltip-9cba8e1752394d7194e89d393a712de9?pvs=4).

### test plan:

verified manually that the tooltip doesn't exist anymore:

https://user-images.githubusercontent.com/66155195/227617381-c977375d-bed6-420f-87ef-9252fb442e96.mov


cmcewen (approved):
ty for video!

PR #6251, opened 2023-03-24
fix: close MP drawer on nft nav

just-toby:
fixes the logic for closing the MP drawer when clicking on an NFT. the NFTCard component differentiates between the "click" event of the button (which is not shown in the MP drawer), and the card itself (which is actually a link to the NFT detail page). 

the default behavior of clicking the card is to navigate to the detail page - this is what we want, and this is what we need a callback for.

design feedback: https://www.notion.so/uniswaplabs/Clicking-an-NFT-doesn-t-close-the-MP-when-it-should-like-token-click-interaction-5aa1e9e8f7a243b581717cbc1682afab?pvs=4

### test plan

https://user-images.githubusercontent.com/66155195/227628944-73b7f65c-a6e1-47c4-a603-cb46bf8c7b37.mov


src/components/WalletDropdown/MiniPortfolio/NFT.tsx
```diff
@@ -62,10 +55,9 @@ export function NFT({
         display={{ disabledInfo: true }}
         isSelected={false}
         isDisabled={false}
-        selectAsset={navigateToNFTDetails}
-        unselectAsset={() => {
-          /* */
-        }}
+        selectAsset={() => null}
+        unselectAsset={() => null}
```
cmcewen, line 59: it seems like we were already passing a function that didn't do anything here, but i think that if we're okay with inaction on a select/unselect, these props shouldn't be a required function in `NftCard` and instead should be optional props (with a comensurate `if (selectAsset) selectAsset()` in the component click handler. what do you think?

src/nft/components/card/index.tsx
```diff
@@ -82,6 +84,7 @@ export const NftCard = ({
       testId={testId}
       onClick={() => {
         if (bagExpanded) setBagExpanded({ bagExpanded: false })
+        if (!doNotLinkToDetails) onNavigation?.()
```
cmcewen, line 87: i think this prop is confusingly named. in the card it also has `<StyledLink to={doNotLinkToDetails ? '' : detailsHref}>` which would be a link to nowhere, which doesn't seem like the right semantic naming. it also seems like the click behavior should be fully encapsulated in onClick instead of having two different paths. @JackShort @cbachmeier looks like this was added here: #6048 - can you think of a way to improve this component API?

src/components/WalletDropdown/MiniPortfolio/NFT.tsx
just-toby: i agree. the NFT card seems like one of those components that has grown in complexity so i subconsciously erred on the side of "don't mess with it". but i think making these props optional is pretty safe.

src/nft/components/card/index.tsx
just-toby: i'm proposing we rename `doNotLinkToDetails` to `linkToDetails` (true by default), but we can wait for the NFT team if you want.

i think the onClick actually should be two separate paths - there is an onClick for the button and for the card itself. this probably wasn't clear from this PR, but at other callsites the card looks like this:

image

to make this clearer, i renamed `onClick` to `onButtonClick`.

src/components/WalletDropdown/MiniPortfolio/NFT.tsx
cmcewen: that makes sense - i'd like to chip away at that complexity bit by bit too so it's less scary to change

cmcewen (approved):
tests pls 😄

src/nft/components/card/index.tsx
```diff
@@ -78,10 +92,11 @@ export const NftCard = ({
...
       detailsHref={detailsHref(asset)}
-      doNotLinkToDetails={doNotLinkToDetails}
+      linkToDetails={linkToDetails}
       testId={testId}
       onClick={() => {
         if (bagExpanded) setBagExpanded({ bagExpanded: false })
+        if (linkToDetails) onNavigation?.()
         sendAnalyticsEvent?.()
       }}
     >
```
grabbou, line 102: So, just to make it clear. The way I read this code and it's really confusing... when `linkToDetails` is `false`, this Container renders the children you see starting on line 103 - MediaContainer and inner components. When it's true, it does not render them, but a link instead?

src/nft/components/card/containers.tsx
```diff
@@ -186,22 +186,22 @@ const Container = ({
...
   detailsHref: string
-  doNotLinkToDetails: boolean
+  linkToDetails: boolean
   testId?: string
   children: ReactNode
   onClick?: (e: React.MouseEvent) => void
 }) => {
   return (
     <CardContainer isSelected={isSelected} isDisabled={isDisabled} testId={testId} onClick={onClick}>
-      <StyledLink to={doNotLinkToDetails ? '' : detailsHref}>{children}</StyledLink>
```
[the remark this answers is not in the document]
grabbou: We should also get rid of `detailsHref` entirely. This should just be a simple UI component.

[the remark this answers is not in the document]
grabbou: PS. Once you get rid of the two aforementioned props, you can just render a direct `CardContainer` instead (a Styled Component), you do not need a wrapper, because these props will be automatically passed to an underlying component!

src/nft/components/card/containers.tsx
```diff
@@ -186,22 +186,22 @@ const Container = ({
...
   isDisabled: boolean
   detailsHref: string
-  doNotLinkToDetails: boolean
+  linkToDetails: boolean
   testId?: string
   children: ReactNode
   onClick?: (e: React.MouseEvent) => void
 }) => {
   return (
     <CardContainer isSelected={isSelected} isDisabled={isDisabled} testId={testId} onClick={onClick}>
```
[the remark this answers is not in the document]
grabbou: I am going to post comments in other parts.

grabbou (approved):
I am going to approve, b/c this PR is generally a net positive change, but depends on some poorly structured code that is due to refactor. I will approve and send a follow-up resolving some of my comments.

PR #6252, opened 2023-03-24
fix: remove price tag from nft cards in the side drawer

just-toby:
hide the price tag on the NFT card in the mini portfolio

design feedback: https://www.notion.so/uniswaplabs/NFT-tile-remove-price-tag-pill-in-top-left-and-fix-play-button-54e382346d394c75927dccda9e455c1e?pvs=4

### test plan 

manually tested to verify that the price tag is gone in the side drawer AND that the profile page (owned NFTs) is unchanged

| Screenshot 2023-03-24 at 1 31 38 PM | Screenshot 2023-03-24 at 1 29 17 PM |


src/nft/components/card/index.tsx
```diff
@@ -71,7 +71,9 @@ export const NftCard = ({
   const tokenType = collectionNft ? asset.tokenType : profileNft ? asset.asset_contract.tokenType : undefined
   const marketplace = collectionNft ? asset.marketplace : undefined
   const listedPrice =
-    profileNft && !isDisabled && asset.floor_sell_order_price ? floorFormatter(asset.floor_sell_order_price) : undefined
+    profileNft && !isDisabled && !hideDetails && asset.floor_sell_order_price
```
cmcewen, line 74: this feels like mixing the concern of what the price is vs whether to show the price in a specific context. semantically i would prefer to make it an option on the tile component to display the price. 

you could imagine a case in the future where we want to give some special treatment to a tile based on a price threshold or something, and then that would require refactoring this without the benefit of design intent + context

just-toby: makes sense

cmcewen (commented):
woot much simpler! could you add a snapshot test for this component too to show diff states and then i'll approve 😄

tinaszheng (approved):

PR #6253, opened 2023-03-24
fix: align /pools CTAs to the top

vm:
This PR makes sure that the title and description for the "info" CTA in the pools page is aligned to the top. This makes sure it's even with the other CTA.

I simplified the code to just use a single CTA component, as we don't believe we need any style changes between the two. I also removed the display: flex as we don't actually need to use align-items or justify-content center here.

after:
Screen Shot 2023-03-24 at 4 48 38 PM

before:
Screen Shot 2023-03-24 at 4 48 46 PM

Test:
- Go to /pools in the preview
- Play around with how the CTAs look on different view widths

### DIFF
--- a/src/pages/Pool/CTACards.tsx
+++ b/src/pages/Pool/CTACards.tsx
@@ -19,14 +19,10 @@ const CTASection = styled.section`
   `};
 `
 
-const CTA1 = styled(ExternalLink)`
+const CTA = styled(ExternalLink)`
   padding: 16px;
   border-radius: 20px;
-  display: flex;
-  flex-direction: column;
   position: relative;
-  justify-content: center;
-  align-items: center;
   overflow: hidden;
   border: 1px solid ${({ theme }) => theme.deprecated_bg3};
 
@@ -39,30 +35,6 @@ const CTA1 = styled(ExternalLink)`
     border: 1px solid ${({ theme }) => theme.deprecated_bg4};
 
     text-decoration: none;
-    * {
-      text-decoration: none !important;
-    }
-  }
-`
-
-const CTA2 = styled(ExternalLink)`
-  position: relative;
-  overflow: hidden;
-  padding: 16px;
-  border-radius: 20px;
-  display: flex;
-  flex-direction: column;
-  justify-content: center;
-  border: 1px solid ${({ theme }) => theme.deprecated_bg3};
-
-  * {
-    color: ${({ theme }) => theme.textPrimary};
-    text-decoration: none !important;
-  }
-
-  :hover {
-    border: 1px solid ${({ theme }) => theme.deprecated_bg4};
-    text-decoration: none !important;
     * {
       text-decoration: none !important;
     }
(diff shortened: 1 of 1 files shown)


cmcewen (approved):
could we align on css grid vs flexbox? find it weird that we had flex nested below grid.

why is the breakpoint so jumpy here? could we not min/max-width this?

https://user-images.githubusercontent.com/1307633/227642929-79ebf7d2-7cb1-41a4-a30a-3cdea5d0b5a5.mov

src/pages/Pool/CTACards.tsx
```diff
@@ -47,19 +42,12 @@ const HeaderText = styled(ThemedText.DeprecatedLabel)`
 
   font-weight: 400;
   font-size: 16px;
-  ${({ theme }) => theme.deprecated_mediaWidth.deprecated_upToMedium`
```
vm, line 78: this seems unnecessary?

src/pages/Pool/CTACards.tsx
```diff
@@ -47,19 +42,12 @@ const HeaderText = styled(ThemedText.DeprecatedLabel)`
...
-    font-size: 16px;
-  `};
 `
 
 const ResponsiveColumn = styled(AutoColumn)`
   grid-template-columns: 1fr;
   width: 100%;
   gap: 8px;
-
-  ${({ theme }) => theme.deprecated_mediaWidth.deprecated_upToMedium`
```
vm, line 88: this seems unnecessary?

src/pages/Pool/CTACards.tsx
```diff
@@ -8,18 +8,13 @@ import { ThemedText } from 'theme'
 import { ExternalLink } from '../../theme'
 
 const CTASection = styled.section`
-  display: grid;
```
vm, line 11: removes grid in favor of flex

src/pages/Pool/CTACards.tsx
```diff
@@ -8,18 +8,13 @@ import { ThemedText } from 'theme'
...
   opacity: 0.8;
-
-  ${({ theme }) => theme.deprecated_mediaWidth.deprecated_upToSmall`
-    grid-template-columns: auto;
-    grid-template-rows: auto;
-  `};
 `
 
 const CTA = styled(ExternalLink)`
+  flex: 1;
```
vm, line 17: equal width for cards

src/pages/Pool/CTACards.tsx
```diff
@@ -8,18 +8,13 @@ import { ThemedText } from 'theme'
 import { ExternalLink } from '../../theme'
 
 const CTASection = styled.section`
-  display: grid;
-  grid-template-columns: 1fr 1fr;
+  display: flex;
   gap: 8px;
   opacity: 0.8;
-
-  ${({ theme }) => theme.deprecated_mediaWidth.deprecated_upToSmall`
```
vm, line 16: removes the jumpiness on breakpoint

cmcewen:
@vm tbh i prefer the original PR code, i don't feel like i understand the changes to responsive column and header text without a deeper investigation which i don't think should be bundled with this change

PR #6254, opened 2023-03-24
fix: l2 icon borders in MP

just-toby:
fixes [design feedback](https://www.notion.so/uniswaplabs/Network-icon-padding-looks-too-big-27d1254f60c346c8a3071dbd2633a9c5?pvs=4) about the outline here being too thick

i simplified the rendering logic here slightly

### test plan

|  image | Screenshot 2023-03-24 at 1 49 59 PM  |


src/components/WalletDropdown/MiniPortfolio/PortfolioLogo.tsx
```diff
@@ -60,31 +60,20 @@ const ENSAvatarImg = styled.img`
...
-  top: 68%;
-  left: 68%;
 `
 
-const ChainLogoSquareBackground = styled.div`
-  height: 18px;
-  width: 18px;
-  border-radius: 4px;
-  background-color: ${({ theme }) => theme.backgroundSurface};
+const L2LogoContainer = styled.div<{ nonSquareLogo?: boolean }>`
```
cmcewen, line 65: since there is a `isSquare` below, i think i'd prefer to keep the props not negative like squareLogo (or potentially the same name isSquare?) feels like a good general rule imo

src/components/WalletDropdown/MiniPortfolio/PortfolioLogo.tsx
```diff
@@ -60,31 +60,20 @@ const ENSAvatarImg = styled.img`
...
+  background-color: ${({ theme, nonSquareLogo }) => (nonSquareLogo ? theme.textPrimary : theme.backgroundSurface)};
+  border-radius: 2px;
+  height: 16px;
+  left: 60%;
```
cmcewen, line 69: don't change this code off of this - but imo percentages are always hard to work with

src/components/WalletDropdown/MiniPortfolio/PortfolioLogo.tsx
```diff
@@ -147,12 +136,10 @@ export function PortfolioLogo({
   }
 
   const L2Logo =
-    chainId === SupportedChainId.MAINNET ? null : (
-      <div>
-        {chainLogo && <ChainLogoSquareBackground />}
-        {!squareLogoUrl && logoUrl && <SquareBackgroundForNonSquareLogo />}
-        {chainLogo && <StyledChainLogo isSquare={!!squareLogoUrl} src={chainLogo} alt="chainLogo" />}
-      </div>
+    chainId === SupportedChainId.MAINNET || !chainLogo ? null : (
```
cmcewen, line 139: i'm finding this logic a bit hard to parse visually. usually i'll think "show the logo if the conditions for the logo to be shown are met, otherwise it's null". so that would look something like "show it if we're not on mainnet and we have a chainLogo"

src/components/WalletDropdown/MiniPortfolio/PortfolioLogo.tsx
tinaszheng: if you're passing in a non-native prop to an html component make sure to use the `$variable` notation

src/components/WalletDropdown/MiniPortfolio/PortfolioLogo.tsx
```diff
@@ -147,12 +136,10 @@ export function PortfolioLogo({
...
+      <L2LogoContainer nonSquareLogo={Boolean(!squareLogoUrl && logoUrl)}>
```
tinaszheng, line 140: in general im not a fan of dispersing the custom logic here within two components - so there's some logic on the `PortfolioLogo` level to determine the logo to be used and also `isSquare` logic, and then there's custom logic within `L2LogoContainer` on the background-color 

feels like they should all be prepared within just the `PortfolioLogo` container by determining the background color here instead of within `L2LogoContainer`

that way you also avoid passing in a non-native prop to an html element and also take out logic from css

tinaszheng (changes requested):

src/components/WalletDropdown/MiniPortfolio/PortfolioLogo.tsx
```diff
@@ -57,34 +57,23 @@ const ENSAvatarImg = styled.img`
...
-  height: ${({ isSquare }) => (isSquare ? '16px' : '14px')};
-  width: ${({ isSquare }) => (isSquare ? '16px' : '14px')};
-  margin-top: ${({ isSquare }) => (isSquare ? '0px' : '1px')};
-  margin-left: ${({ isSquare }) => (isSquare ? '0px' : '1px')};
-  position: absolute;
-  top: 68%;
-  left: 68%;
+const StyledChainLogo = styled.img<{ $padded: boolean }>`
+  height: ${({ $padded }) => ($padded ? '14px' : '16px')};
+  width: ${({ $padded }) => ($padded ? '14px' : '16px')};
```
tinaszheng, line 62: im confused by the convention here - why does `padded` just translate to there being smaller height and width?

just-toby: it does because it's centered relative to the parent, so making it smaller means there is "padding". anyway, i changed this to hopefully be clearer.

src/components/WalletDropdown/MiniPortfolio/PortfolioLogo.tsx
```diff
@@ -147,13 +137,11 @@ export function PortfolioLogo({
...
-      <div>
-        {chainLogo && <ChainLogoSquareBackground />}
-        {!squareLogoUrl && logoUrl && <SquareBackgroundForNonSquareLogo />}
-        {chainLogo && <StyledChainLogo isSquare={!!squareLogoUrl} src={chainLogo} alt="chainLogo" />}
-      </div>
-    )
+    chainId !== SupportedChainId.MAINNET && chainLogo ? (
+      <L2LogoContainer $backgroundColor={squareLogoUrl ? theme.backgroundSurface : theme.textPrimary}>
+        <StyledChainLogo $fillParent={Boolean(squareLogoUrl)} src={chainLogo} alt="chainLogo" />
+      </L2LogoContainer>
```
tinaszheng, line 143: honestly might be cleaner to just do two separate components here instead of trying to generalize for both square and non-square

src/components/WalletDropdown/MiniPortfolio/PortfolioLogo.tsx
```diff
@@ -57,34 +57,23 @@ const ENSAvatarImg = styled.img`
...
-  height: ${({ isSquare }) => (isSquare ? '16px' : '14px')};
-  width: ${({ isSquare }) => (isSquare ? '16px' : '14px')};
-  margin-top: ${({ isSquare }) => (isSquare ? '0px' : '1px')};
-  margin-left: ${({ isSquare }) => (isSquare ? '0px' : '1px')};
-  position: absolute;
-  top: 68%;
-  left: 68%;
+const StyledChainLogo = styled.img<{ $fillParent: boolean }>`
+  height: ${({ $fillParent }) => ($fillParent ? '100%' : '14px')};
+  width: ${({ $fillParent }) => ($fillParent ? '100%' : '14px')};
```
tinaszheng, line 62: this is also confusing because layout should generally be determined by parent instead of the child component - see comment below about just making two separate components here

PR #6258, opened 2023-03-27
fix: Revert to previous Nav menu

vm:
Reverts to the Nav UI from prior to https://github.com/Uniswap/interface/pull/6216

Before (app.uniswap.org):
Screen Shot 2023-03-27 at 9 37 17 AM
Note, the tax services 

After:
Screen Shot 2023-03-27 at 9 37 12 AM

Need to check:
- Are we keeping the tax services row?
- Do we need to show Pools in menu when on truncated view?

Test:
- Open menu on preview
- Click on the various icons to make sure they work


grabbou:
```
Need to check:

Are we keeping the tax services row?
Do we need to show Pools in menu when on truncated view?
```

@vm is this ready to review or still a draft then?

grabbou:
I will convert to `draft` just to keep things clean, but feel free to revert that change if you feel otherwise!

vm:
@grabbou np, re-opened now that PM approved design changes

src/nft/components/icons.tsx
```diff
@@ -4,7 +4,7 @@ import styled, { useTheme } from 'styled-components/macro'
 import { themeVars, vars } from '../css/sprinkles.css'
 
 // ESLint reports `fill` is missing, whereas it exists on an SVGProps type
-type SVGProps = React.SVGProps<SVGSVGElement> & { fill?: string }
+type SVGProps = React.SVGProps<SVGSVGElement> & { fill?: string; height?: string | number; width?: string | number }
```
vm, line 7: i'm surprised this was necessary tbh. if you feel it's good we can investigate if there's a better option here

grabbou (approved):
It's mostly reverting what was already working, so I think it's good to go!

PR #6261, opened 2023-03-28
fix: dont block trade when price impact is favorable

just-toby:
### context
- [jira ticket](https://uniswaplabs.atlassian.net/browse/WEB-3024?atlOrigin=eyJpIjoiMzcwZWE3MmI4MjU3NDVkYmFhZDY1NjgzOWJkMjIzZmIiLCJwIjoiaiJ9)
- [context for the bug report is here
](https://uniswapteam.slack.com/archives/C04D07TQBT4/p1678829853381139)

the Price Impact value we use to calculate this severity level is typically positive (above zero, not favorable). for a favorable trade the price impact is in the other direction, i.e. negative (less than zero).  note that we flip the sign when displaying the value.

### test plan

- added unit test for `warningSeverity` function to verify the expected behavior
- manually tested a typical trade (in which the slippage is not favorable) to verify that the warning severity is still correct

image

### DIFF
--- a/src/utils/prices.ts
+++ b/src/utils/prices.ts
@@ -84,7 +84,13 @@ const IMPACT_TIERS = [
 
 type WarningSeverity = 0 | 1 | 2 | 3 | 4
 export function warningSeverity(priceImpact: Percent | undefined): WarningSeverity {
-  if (!priceImpact || priceImpact.lessThan(0)) return 0
+  if (!priceImpact) return 0
+  // This function is used to calculate the Severity level for % changes in USD value and Price Impact.
+  // Price Impact is always an absolute value (conceptually always negative, but represented in code with a positive value)
+  // The USD value change can be positive or negative, and it follows the same standard as Price Impact (positive value is the typical case of a loss due to slippage).
+  // We don't want to return a warning level for a favorable/profitable change, so when the USD value change is negative we return 0.
+  // TODO (WEB-3133): Disambiguate Price Impact and USD value change, and flip the sign of USD Value change.
+  if (priceImpact.lessThan(0)) return 0
   let impact: WarningSeverity = IMPACT_TIERS.length as WarningSeverity
   for (const impactLevel of IMPACT_TIERS) {
     if (impactLevel.lessThan(priceImpact)) return impact
--- a/src/utils/prices.ts
+++ b/src/utils/prices.ts
@@ -84,7 +84,7 @@ const IMPACT_TIERS = [
 
 type WarningSeverity = 0 | 1 | 2 | 3 | 4
 export function warningSeverity(priceImpact: Percent | undefined): WarningSeverity {
-  if (!priceImpact) return 0
+  if (!priceImpact || priceImpact.lessThan(0)) return 0
   let impact: WarningSeverity = IMPACT_TIERS.length as WarningSeverity
   for (const impactLevel of IMPACT_TIERS) {
     if (impactLevel.lessThan(priceImpact)) return impact
--- a/src/utils/prices.test.ts
+++ b/src/utils/prices.test.ts
@@ -122,6 +122,9 @@ describe('prices', () => {
     it('0 for undefined', () => {
       expect(warningSeverity(undefined)).toEqual(0)
     })
+    it('0 for negative', () => {
+      expect(warningSeverity(new Percent(-1))).toEqual(0)
+    })
     it('correct for 0', () => {
       expect(warningSeverity(new Percent(0))).toEqual(0)
     })


src/utils/prices.ts
```diff
@@ -84,7 +84,7 @@ const IMPACT_TIERS = [
 
 type WarningSeverity = 0 | 1 | 2 | 3 | 4
 export function warningSeverity(priceImpact: Percent | undefined): WarningSeverity {
-  if (!priceImpact) return 0
+  if (!priceImpact || priceImpact.lessThan(0)) return 0
```
grabbou, line 87: Would `priceImpact?.lessThan(0)` work?

grabbou (approved):
That PR looks good to me overall, but I have some follow-up non-blocking questions that would be good to discuss beforehand!

tinaszheng (dismissed):
Wait blocking changes here since we discussed not doing this because negative price impact is a bug - did you find out why this is happening?

just-toby:
@tinaszheng yes here's what's happening 

first, a clarification:  at a conceptual level, Price Impact is always negative. and currently in the codebase we assume it's always negative, so we use positive `Percent` objects to represent it (and we just flip the sign when formatting to display it).

there's 2 different issues going on related to % changes:

1. (this diff addresses) when there is really low liquidity for a v3 pair and your trade actually does incur favorable change in USD values. if you check the slack thread i linked in the PR description, you'll see the example from BNB chain right after the launch. This is expected behavior from the router when there is low liquidity between pair A/B or A/stable or B/stable
2. in the widget (and maybe in interface too), we have an edge case bug where the `router-sdk` package calculates a *negative* (i.e. incorrect direction) *price impact*. this is unrelated to this PR

src/utils/prices.ts
just-toby: `if (priceImpact?.lessThan(0))` wouldn't return 0 when `priceImpact` is undefined, so i'd need to inverse the boolean logic or something which would be less readable imo. lmk if i misunderstood your point there.

for `undefined` .... it does seem to be a pattern throughout the codebase to allow it. i didn't write this code originally, so i don't know why this specific function accepts undefined. my only guess is that it silences some TS warnings

just-toby:
@tinaszheng see [here](https://github.com/Uniswap/interface/blob/main/src/pages/Swap/index.tsx#L404) - it looks like as of [#4217](https://github.com/Uniswap/interface/pull/4217) we are using the larger of [fiat value change, price impact] to determine the warning severity and block trade. either way, we're using this `warningSeverity` util function to calculate it

src/utils/prices.ts
cmcewen: i think this is a result of the hard commitment to hooks, where when data isn't available yet lots of things just pass undefined. so somewhere a hook calls this, and the hook accepts undefined

tinaszheng:
okay gotcha, yeah i think i was confused because we use `price impact` as a terminology in the codebase for both protocol price impact and usd value change, whereas IRL i've only heard the second one referred to as "USD value change" so sorry for the confusing message above

so i think with that in mind the negative/positive sign stuff makes this confusing to me:
- so price impact is an absolute number so it should technically always be positive (lets ignore the negative bug for now)
- usd value change can both be positive or negative

and 

big price impact and big negative usd value change = bad
low price impact and all 0/positive usd value change = good

so this code is confusing to me because what we're doing is actually:
if usd value change * -1 < 0 then it's actually positive => good
if price impact * -1 < 0 => ?? 

so i think in general this flow could use a refactor to get the positive/negative signs right, but for now i think your change is ok behavior wise, so could you just leave a big comment block above the `.lessThan(0)` case and explain the weird sign issues going on?

src/utils/prices.ts
```diff
@@ -85,6 +85,12 @@ const IMPACT_TIERS = [
 type WarningSeverity = 0 | 1 | 2 | 3 | 4
 export function warningSeverity(priceImpact: Percent | undefined): WarningSeverity {
   if (!priceImpact) return 0
+  // This function is used to calculate the Severity level for % changes in USD value and Price Impact.
+  // Price Impact is always an absolute value (conceptually always negative, but represented in code with a positive value)
+  // The USD value change can be positive or negative, and it follows the same standard as Price Impact (positive value is the typical case of a loss due to slippage).
+  // We don't want to return a warning level for a favorable/profitable change, so when the USD value change is negative we return 0.
+  // TODO (WEB-3133): Disambiguate Price Impact and USD value change, and flip the sign of USD Value change.
```
just-toby, line 92: https://uniswaplabs.atlassian.net/browse/WEB-3133?atlOrigin=eyJpIjoiNGMzOTUwMjc3OGMyNGIwOGFjZjA5MzQ5MDc1MjAzMGEiLCJwIjoiaiJ9

cmcewen (approved):
LGTM, Tina and I discussed before and agreed on leaving the other work as a followup

src/utils/prices.ts
```diff
@@ -85,6 +85,12 @@ const IMPACT_TIERS = [
 type WarningSeverity = 0 | 1 | 2 | 3 | 4
 export function warningSeverity(priceImpact: Percent | undefined): WarningSeverity {
   if (!priceImpact) return 0
+  // This function is used to calculate the Severity level for % changes in USD value and Price Impact.
+  // Price Impact is always an absolute value (conceptually always negative, but represented in code with a positive value)
+  // The USD value change can be positive or negative, and it follows the same standard as Price Impact (positive value is the typical case of a loss due to slippage).
```
tinaszheng, line 90: hm this is wrong btw, these values have nothing to do with slippage

tinaszheng: this is also confusing because the "standard practice" for price impact is an absolute value

PR #6263, opened 2023-03-28
fix: getCurrency crash when token not found

just-toby:
## Description

When a token doesn't exist in the default token lists, `getCurrency` is throwing an NPE and crashing the app. this PR updates `getCurrency` to return `undefined`, and all callsites/UI and dependent code to handle `undefined` values for these tokens.

- [sentry error](https://uniswap-labs.sentry.io/issues/4044560968/?environment=production&project=4504255148851200&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=1h&stream_index=9)
- [slack thread](https://uniswapteam.slack.com/archives/C05021F43HU/p1680042920041939)
- [slack thread](https://uniswapteam.slack.com/archives/C02T729PMQE/p1680041958264159)

## Screen Capture
Screenshot 2023-03-28 at 4 12 30 PM 

Screenshot 2023-03-28 at 5 37 29 PM
Screenshot 2023-03-28 at 5 36 42 PM
Screenshot 2023-03-28 at 5 38 08 PM

#### Manual

1. manually tested a swap between two tokens that are expected to be found, and verified that the tx notif behavior is unchanged (first screenshot above)
2. manually tested a swap with one known token and one unknown token (second screenshot above)
3. manually tested a swap with both tokens missing from the map (third screenshot above)
4. manually verified that the ActivityRows in the Mini Portfolio render correctly with one or both tokens missing.

#### Automated
- [x] Unit test
- [ ] Integration/E2E test

### DIFF
--- a/src/components/Popups/TransactionPopup.tsx
+++ b/src/components/Popups/TransactionPopup.tsx
@@ -1,13 +1,15 @@
+import { Currency } from '@uniswap/sdk-core'
 import { useWeb3React } from '@web3-react/core'
 import Column from 'components/Column'
 import { parseLocalActivity } from 'components/WalletDropdown/MiniPortfolio/Activity/parseLocal'
 import { PortfolioLogo } from 'components/WalletDropdown/MiniPortfolio/PortfolioLogo'
 import PortfolioRow from 'components/WalletDropdown/MiniPortfolio/PortfolioRow'
 import useENSName from 'hooks/useENSName'
+import { CheckCircle } from 'react-feather'
 import { useCombinedActiveList } from 'state/lists/hooks'
 import { useTransaction } from 'state/transactions/hooks'
 import { TransactionDetails } from 'state/transactions/types'
-import styled from 'styled-components/macro'
+import styled, { useTheme } from 'styled-components/macro'
 import { EllipsisStyle, ThemedText } from 'theme'
 import { ExplorerDataType, getExplorerLink } from 'utils/getExplorerLink'
 
@@ -22,22 +24,29 @@ function TransactionPopupContent({ tx, chainId }: { tx: TransactionDetails; chai
   const tokens = useCombinedActiveList()
   const activity = parseLocalActivity(tx, chainId, tokens)
   const { ENSName } = useENSName(activity?.otherAccount)
+  const theme = useTheme()
 
   if (!activity) return null
 
   const explorerUrl = getExplorerLink(chainId, tx.hash, ExplorerDataType.TRANSACTION)
+
+  const currencies = activity.currencies?.filter((c) => c) as Currency[]
 
   return (
     <PortfolioRow
       left={
         success ? (
           <Column>
-            <PortfolioLogo
-              chainId={chainId}
-              currencies={activity.currencies}
-              images={activity.logos}
-              accountAddress={activity.otherAccount}
-            />
+            {currencies.length > 0 ? (
+              <PortfolioLogo
+                chainId={chainId}
+                currencies={currencies}
+                images={activity.logos}
+                accountAddress={activity.otherAccount}
+              />
+            ) : (
+              <CheckCircle color={theme.accentSuccess} size={24} />
+            )}
           </Column>
         ) : (
           <PopupAlertTriangle />
--- a/src/components/Popups/TransactionPopup.tsx
+++ b/src/components/Popups/TransactionPopup.tsx
@@ -1,15 +1,13 @@
-import { Currency } from '@uniswap/sdk-core'
 import { useWeb3React } from '@web3-react/core'
 import Column from 'components/Column'
 import { parseLocalActivity } from 'components/WalletDropdown/MiniPortfolio/Activity/parseLocal'
 import { PortfolioLogo } from 'components/WalletDropdown/MiniPortfolio/PortfolioLogo'
 import PortfolioRow from 'components/WalletDropdown/MiniPortfolio/PortfolioRow'
 import useENSName from 'hooks/useENSName'
-import { CheckCircle } from 'react-feather'
 import { useCombinedActiveList } from 'state/lists/hooks'
 import { useTransaction } from 'state/transactions/hooks'
 import { TransactionDetails } from 'state/transactions/types'
-import styled, { useTheme } from 'styled-components/macro'
+import styled from 'styled-components/macro'
 import { EllipsisStyle, ThemedText } from 'theme'
 import { ExplorerDataType, getExplorerLink } from 'utils/getExplorerLink'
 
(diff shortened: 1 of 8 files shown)


src/components/Popups/TransactionPopup.tsx
```diff
@@ -22,22 +24,29 @@ function TransactionPopupContent({ tx, chainId }: { tx: TransactionDetails; chai
   const tokens = useCombinedActiveList()
   const activity = parseLocalActivity(tx, chainId, tokens)
   const { ENSName } = useENSName(activity?.otherAccount)
+  const theme = useTheme()
 
   if (!activity) return null
 
   const explorerUrl = getExplorerLink(chainId, tx.hash, ExplorerDataType.TRANSACTION)
 
+  const currencies = activity.currencies?.filter((c) => c) as Currency[]
```
zzmp, line 33: nit: I'm not sure if these will work in tandem, but you can use `Boolean` as an identity filter, and you can usually (depending on your filterFn) type the return type instead of coercing it, which I find to be safer.

```suggestion
  const currencies = activity.currencies?.filter<Currency>(Boolean)
```

src/components/Popups/TransactionPopup.tsx
```diff
@@ -22,22 +24,29 @@ function TransactionPopupContent({ tx, chainId }: { tx: TransactionDetails; chai
...
       left={
         success ? (
           <Column>
-            <PortfolioLogo
-              chainId={chainId}
-              currencies={activity.currencies}
-              images={activity.logos}
-              accountAddress={activity.otherAccount}
-            />
+            {currencies.length > 0 ? (
```
zzmp, line 40: `PortfolioLogo` already has logic to deal with undefined currencies. I think this file can be left as-is.

src/components/WalletDropdown/MiniPortfolio/Activity/ActivityRow.tsx
```diff
@@ -45,7 +49,16 @@ export function ActivityRow({
       <PortfolioRow
         left={
           <Column>
-            <PortfolioLogo chainId={chainId} currencies={currencies} images={logos} accountAddress={otherAccount} />
+            {definedCurrencies?.length > 0 ? (
+              <PortfolioLogo
```
zzmp, line 53: `PortfolioLogo` already has logic to deal with undefined currencies. I think this file can be left as-is.

src/components/WalletDropdown/MiniPortfolio/Activity/types.ts
```diff
@@ -12,7 +12,7 @@ export type Activity = {
   title: string
   descriptor?: string
   logos?: Array<string | undefined>
-  currencies?: Array<Currency>
+  currencies?: Array<Currency | undefined>
```
zzmp, line 15: These can be narrowed:

```
  logos?: [string | undefined, string | undefined],
  currencies?: [Currency | undefined, Currency | undefined],
```

src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.tsx
```diff
@@ -26,20 +26,22 @@ import {
...
   delimiter = t`for`
 ) {
-  const formattedA = formatCurrencyAmount(CurrencyAmount.fromRawAmount(currencyA, amtA))
-  const formattedB = formatCurrencyAmount(CurrencyAmount.fromRawAmount(currencyB, amtB))
-  return `${formattedA} ${currencyA.symbol} ${delimiter} ${formattedB} ${currencyB.symbol}`
+  const formattedA = currencyA ? formatCurrencyAmount(CurrencyAmount.fromRawAmount(currencyA, amtA)) : t`Unknown`
+  const symbolA = currencyA?.symbol ?? ''
+  const formattedB = currencyB ? formatCurrencyAmount(CurrencyAmount.fromRawAmount(currencyB, amtB)) : t`Unknown`
+  const symbolB = currencyB?.symbol ?? ''
+  return `${formattedA} ${symbolA} ${delimiter} ${formattedB} ${symbolB}`
```
zzmp, line 44: This adds too much space if the symbol is unknown. Can you refactor to avoid that? eg:

```suggestion
  return [formattedA, symbolA, delimiter, formattedB, symbolB].filter(Boolean).join(' ')
```

src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.tsx
```diff
@@ -79,7 +81,9 @@ function parseApproval(
 ): Partial<Activity> {
   // TODO: Add 'amount' approved to ApproveTransactionInfo so we can distinguish between revoke and approve
   const currency = getCurrency(approval.tokenAddress, chainId, tokens)
-  const descriptor = t`${currency.symbol ?? currency.name}`
+  const symbol = currency?.symbol
+  const name = currency?.name ?? 'Unknown'
+  const descriptor = t`${symbol ?? name}`
```
zzmp, line 86: Only 'Unknown' should be translated. You could also make this inline, which I think is more readable

```suggestion
  const descriptor = currency?.symbol ?? currency?.name ?? t`Unknown`
```

src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.tsx
```diff
@@ -120,8 +124,10 @@ function parseMigrateCreateV3(
   tokens: TokenAddressMap
 ): Partial<Activity> {
   const baseCurrency = getCurrency(lp.baseCurrencyId, chainId, tokens)
+  const baseSymbol = baseCurrency?.symbol ?? 'Unknown'
```
zzmp, line 127:
```suggestion
  const baseSymbol = baseCurrency?.symbol ?? t`Unknown`
```

src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.tsx
```diff
@@ -120,8 +124,10 @@ function parseMigrateCreateV3(
...
   const quoteCurrency = getCurrency(lp.baseCurrencyId, chainId, tokens)
-  const descriptor = t`${baseCurrency.symbol} and ${quoteCurrency.symbol}`
+  const quoteSymbol = quoteCurrency?.symbol ?? 'Unknown'
```
zzmp, line 129:
```suggestion
  const quoteSymbol = quoteCurrency?.symbol ?? t`Unknown`
```

zzmp (changes requested):

src/components/Popups/TransactionPopup.tsx
tinaszheng: i believe the type casting for `Currency[]` here is also wrong because `activity.currencies` can be `undefined`

src/components/WalletDropdown/MiniPortfolio/Activity/types.ts
just-toby: these can have any number of items depending on the tx type - i'd rather not make that big of a type change in this PR

src/components/Popups/TransactionPopup.tsx
just-toby: not relevant anymore because we can just pass arrays with undefined elements to `PortfolioLogo`

zzmp (approved):

tinaszheng (approved):

PR #6264, opened 2023-03-29
build: upgrade sentry

zzmp:
## Description

Upgrades Sentry integration to 7.45.0, which pulls in a fix for [INFRA-143](https://uniswaplabs.atlassian.net/browse/INFRA-143) ([release notes](https://github.com/getsentry/sentry-javascript/releases/tag/7.42.0#:~:text=fix(react)%3A%20Make%20redux%20integration%20be%20configurable%20via%20normalizeDepth%20(%237379))).

#### Manual

- [ ] Added an error, ran locally, and observed that the error reported included correctly expanded redux state

#### Automated
N/A


JFrankfurt (approved):
is the 10->3 change going to meaningfully reduce the amount of info we're getting?

zzmp:
> is the 10->3 change going to meaningfully reduce the amount of info we're getting?

no, it will only affect the session state that is logged from redux state.
I'm working on a separate filter to omit known "ignorable" event, which should reduce noise.

src/tracing/index.ts
```diff
@@ -17,24 +17,20 @@ const AMPLITUDE_DUMMY_KEY = '00000000000000000000000000000000'
 export const STATSIG_DUMMY_KEY = 'client-0000000000000000000000000000000000000000000'
 
 Sentry.init({
+  // general configuration:
   dsn: process.env.REACT_APP_SENTRY_DSN,
   release: process.env.REACT_APP_GIT_COMMIT_HASH,
   environment: getEnvName(),
+  // exception reporting:
   enabled: isSentryEnabled(),
+  // performance tracing:
```
grabbou, line 26: nit: capital letter

src/tracing/index.ts
```diff
@@ -17,24 +17,20 @@ const AMPLITUDE_DUMMY_KEY = '00000000000000000000000000000000'
 export const STATSIG_DUMMY_KEY = 'client-0000000000000000000000000000000000000000000'
 
 Sentry.init({
+  // general configuration:
   dsn: process.env.REACT_APP_SENTRY_DSN,
   release: process.env.REACT_APP_GIT_COMMIT_HASH,
   environment: getEnvName(),
+  // exception reporting:
```
grabbou: I personally don't think we need this comment and the one on line `20`, looks pretty self explanatory? I would prefer the comment to provide additional context if necessary.

src/tracing/index.ts
grabbou: I would drop this comment (no added value) or - alternatively - provide additional context what is sample rate and why we're setting it to this one. 

Also, are we satisfied with `0`? What does that mean in practice? Shouldn't we:
- keep it undefined
- throw an error when env variable not provided (to make sure we don't run into issues in production when one forgets to set the value - in dev, it will most likely exist anyway due to predefined `.env` file)

grabbou (approved):
Overall looks good, just a couple of comment-related nits. 

Also, I have one "requested change" for `tracesSampleRate`, but I think that got merged in the other PR, so not something blocking this PR specifically, but rather what we can discuss as a follow-up.

src/tracing/index.ts
zzmp: I think these comments have value in that they keep the configurations separated. Despite Sentry implementing both exception reporting and performance tracing, I think it's easier to think of configuring them separately, so these comments encourage devs to keep these configurations grouped separately.

src/tracing/index.ts
zzmp: In this case, `0` is functionally equivalent to `undefined`, and it cannot be guaranteed to exist in dev mode so I think its best to not throw (unless we decide to also define it in `.env`, but I think that's dangerous because then removing it in `.env.production` would cause us to use the dev value).

PR #6267, opened 2023-03-29
fix: omit failed eth_blockNumber calls from sentry

zzmp:
### Motivation

The noisiest Sentry issue is `eth_blockNumber` request failure, which is expected to happen (either spuriously or because they are rate-limited) because it is polled so frequently. Logging it floods our Sentry issues.

### Summary

Ignores `eth_blockNumber` request failures and network changes, which should not be fatal errors. Both of these are provider-originated, non-fatal, unhandled errors.

To avoid re-implementing a global exception handler - because Sentry's `GlobalHandlers` integration already does so, and includes edge cases that we would not - we instead intercept the Sentry events before they are sent to Sentry, and filter our block number and network change errors.

#### Manual

This is difficult to test without either debugging or modifying code. I tested this locally, with a modified build, using the debugging Sentry DSN.
- [x] Induced an eth_blockNumber request failure (using the _Network request blocking_ pane) and verified that it hit the codepath and did not report to Sentry
- [x] Induced a network change (by modifying code) and verified that it hit the codepath and did not report to Sentry
- [x] Induced a rejection to log to Sentry, and verified that it was still reported to Sentry

#### Automated
- [X] Unit test


grabbou (changes requested):
Is there a better way to do it, rather than working around Sentry tracing function? 

- Would this error be an "uncaught exception" that one could see in their dev tools console? If yes, could we instead have an `onerror` listener on `window` (or an alternative mechanism) where we catch these, so that they're not reported in the console either and do not require Sentry workaround? Sounds like something that would work for other kinds of errors too in the future.
- Alternatively, is there any particular line of code that can lead to this error, and wrapping it with a try/catch would fix it? If yes, why not do that instead?

zzmp:
> > Would this error be an "uncaught exception" that one could see in their dev tools console? If yes, could we instead have an onerror listener on window (or an alternative mechanism) where we catch these, so that they're not reported in the console either and do not require Sentry workaround? Sounds like something that would work for other kinds of errors too in the future.
> 
> I think you have a good point here, I'll re-tool.

Sentry's `GlobalHandlers` integration covers more edge cases, and after testing I don't think these errors are reported in the console, so there is nothing to gain by implementing our own global handler. Instead, I've opted here to hook into Sentry's infra to filter errors after they're caught by Sentry's global handler, because Sentry's global handler is more robust than one which we would write ourselves.

grabbou (approved):

src/tracing/errors.ts
```diff
@@ -0,0 +1,23 @@
+import { ErrorEvent, EventHint } from '@sentry/types'
+
+/** Identifies ethers request errors (as thrown by {@type import(@ethersproject/web).fetchJson}). */
+function isEthersRequestError(error: Error): error is Error & { requestBody: string } {
+  return 'requestBody' in error && typeof (error as unknown as Record<'requestBody', unknown>).requestBody === 'string'
+}
+
+export function filterKnownErrors(event: ErrorEvent, hint: EventHint): ErrorEvent | null {
```
just-toby, line 8: nit: we should document that returning `null` is how you prevent the error from being logged. i couldn't find this explicitly mentioned in the docs, but if you can then maybe we should link to it.

zzmp: It's documented through the type definition of beforeSend. I can enforce that `filterKnownErrors` matches that typedef!

just-toby (approved):

PR #6268, opened 2023-03-29
feat: add data-source agnostic token constant lookup cache mechanism (WIP)

JFrankfurt:
## Description

If a token constant field like name, symbol, or decimals exists on a token list it can be easily queried with negligible performance impact. But if the token at a given address is not already available then we have to use a provider to query chain state. We currently have no mechanism for persisting the result of that query such that future requests for the data resolve to a local cache and do not incur unnecessary network request overhead. This class aims to be a general query tool to address this problem.

### Context

original code which prompted this: https://github.com/Uniswap/interface/pull/6249/files#r1149720580

image

### Possible work to be done

- batch fetch mechanism
image

- decide between async getter to allow for chain data fetching and a standalone setter to be called by external data fetching mechanisms

#### Manual

_[Steps of how you are testing the change and ensuring no regression.]_

#### Automated
- [ ] Unit test
- [ ] Integration/E2E test

### DIFF
--- a/src/constants/TokenConstantsLookupTable.ts
+++ b/src/constants/TokenConstantsLookupTable.ts
@@ -1,0 +1,58 @@
+import store from 'state'
+
+import { SupportedChainId } from './chains'
+import { DEFAULT_LIST_OF_LISTS } from './lists'
+
+type lookupKey = `0x${string}:${SupportedChainId}`
+interface TokenConstants {
+  decimals?: number // uint8
+  name?: string
+  symbol?: string
+  chainChecked: boolean
+}
+class TokenConstantsLookupTable {
+  private dict: Record<lookupKey, TokenConstants | undefined> = {}
+  private initialized = false
+
+  initialize() {
+    const dict: Record<lookupKey, TokenConstants | undefined> = {}
+
+    DEFAULT_LIST_OF_LISTS.forEach((list) => {
+      const listData = store.getState().lists.byUrl[list]
+      if (!listData) {
+        return
+      }
+      listData.current?.tokens.forEach((token) => {
+        const lowercaseAddress = token.address.toLowerCase()
+        const key = `${lowercaseAddress}:${token.chainId}` as lookupKey
+        const entry: TokenConstants = { chainChecked: false }
+        if (token.decimals) entry.decimals = token.decimals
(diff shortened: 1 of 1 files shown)


src/constants/TokenConstantsLookupTable.ts
```diff
@@ -0,0 +1,58 @@
...
+  decimals?: number // uint8
+  name?: string
+  symbol?: string
+  chainChecked: boolean
+}
+class TokenConstantsLookupTable {
+  private dict: Record<lookupKey, TokenConstants | undefined> = {}
+  private initialized = false
+
+  initialize() {
```
cmcewen, line 17: when will this init? how do you ensure the store has been properly hydrated with the lists?

src/constants/TokenConstantsLookupTable.ts
```diff
@@ -0,0 +1,58 @@
...
+    const dict: Record<lookupKey, TokenConstants | undefined> = {}
+
+    DEFAULT_LIST_OF_LISTS.forEach((list) => {
+      const listData = store.getState().lists.byUrl[list]
+      if (!listData) {
+        return
+      }
+      listData.current?.tokens.forEach((token) => {
+        const lowercaseAddress = token.address.toLowerCase()
+        const key = `${lowercaseAddress}:${token.chainId}` as lookupKey
```
cmcewen, line 27: is it better to have a flat object map or to nest?

tinaszheng:
> We currently have no mechanism for persisting the result of that query such that future requests for the data resolve to a local cache and do not incur unnecessary network request overhead

isnt this what redux-multicall does tho? if you do `useSingleCallResult` with `NEVER_RELOAD` it will not make additional network requests

zzmp:
To add to @tinaszheng's comment: why are our current systems insufficient for caching network lookups? Is there some systemic or design reason that they fail? Has this been measured?

JFrankfurt:
this isn't strictly a network lookup cache--for most tokens a user might touch we already have this data in memory via token list, so this data structure would reference that instead of ever making an rpc call

tinaszheng:
I'm still confused what this is for and why we need it given all our existing cacheing solutions 🤔 this is a good opportunity for an eng doc imo

PR #6270, opened 2023-03-29
fix: remove deprecated default imports from zustand

just-toby:
## Description

the default import from `zustand` is deprecated, we should be using a named import instead. this fixes the console warnings and adds an eslint rule to prevent this from happening in the future.

https://uniswaplabs.atlassian.net/browse/WEB-1566?atlOrigin=eyJpIjoiMWUyZTM4NmExNDIwNGQ1MDkyZTVkZTVlYjFmNjcyOTkiLCJwIjoiaiJ9

## Screen Capture
Screenshot 2023-03-29 at 10 39 24 AM

#### Manual
- [X] Ran the app and verified that the warnings don't appear in the console anymore.
- [X] Tested the new lint rule.

#### Automated
- [ ] Unit test
- [ ] Integration/E2E test
- [x] lint rule

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,7 +15,7 @@ module.exports = {
               {
                 name: 'zustand',
                 importNames: ['default'],
-                message: 'Default import from zustand is depracated. Use named import instead.',
+                message: 'Default import from zustand is deprecated. Import `{ create }` instead.',
               },
             ],
           },
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,4 +4,23 @@ require('@uniswap/eslint-config/load')
 
 module.exports = {
   extends: '@uniswap/eslint-config/react',
+  overrides: [
+    {
+      files: ['**/*.ts', '**/*.tsx'],
+      rules: {
+        'no-restricted-imports': [
+          'error',
+          {
+            paths: [
+              {
+                name: 'zustand',
+                importNames: ['default'],
+                message: 'Default import from zustand is depracated. Use named import instead.',
+              },
+            ],
+          },
+        ],
+      },
+    },
+  ],
 }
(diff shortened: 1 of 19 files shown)


.eslintrc.js
```diff
@@ -4,4 +4,23 @@ require('@uniswap/eslint-config/load')
 
 module.exports = {
   extends: '@uniswap/eslint-config/react',
+  overrides: [
+    {
+      files: ['**/*.ts', '**/*.tsx'],
+      rules: {
+        'no-restricted-imports': [
```
zzmp, line 11: If there were to be an existing definition for `no-restricted-imports` in `@uniswap/eslint-config/react`, would this overwrite it or extend it? I think it would overwrite it. Is there a way to instead extend it?

.eslintrc.js
```diff
@@ -4,4 +4,23 @@ require('@uniswap/eslint-config/load')
...
+          'error',
+          {
+            paths: [
+              {
+                name: 'zustand',
+                importNames: ['default'],
+                message: 'Default import from zustand is depracated. Use named import instead.',
```
zzmp, line 18:
```suggestion
                message: 'Default import from zustand is deprecated. Import `{ create }` instead.',
```

zzmp (changes requested):

.eslintrc.js
just-toby: image

this merges the rules so we're good i think

zzmp: currently `@uniswap/eslint-config/react` uses [a different rule](https://github.com/Uniswap/eslint-config/blob/main/react.js#L28), `@typescript-eslint/no-restricted-imports`, so it may not be merged. This may lead to problems in the future.

zzmp (changes requested):
still concerned about overwriting vs extending

.eslintrc.js
zzmp: afaict there is no way to do this - if you want to extend vs override you need to bring in the old definition and redefine it at the override level. :(

zzmp (approved):

PR #6276, opened 2023-03-30
fix: injection detection bug

cartcrom:
## Description

Mini portfolio PR caused a regression in how we detect injected wallet that do not inject right away. This PR moves `Connection` fields that have dynamic value based on injection state to be getter functions rather than static vars.

This also adds more tests for edge cases for combinations of different browser environments / injections

#### Manual

_[Steps of how you are testing the change and ensuring no regression.]_

#### Automated
- [x] Unit test
- [ ] Integration/E2E test

### DIFF
--- a/src/components/WalletModal/index.tsx
+++ b/src/components/WalletModal/index.tsx
@@ -7,7 +7,7 @@ import { AutoColumn } from 'components/Column'
 import { AutoRow } from 'components/Row'
 import { useWalletDrawer } from 'components/WalletDropdown'
 import IconButton from 'components/WalletDropdown/IconButton'
-import { Connection, ConnectionType, networkConnection, useConnections } from 'connection'
+import { Connection, ConnectionType, getConnections, networkConnection } from 'connection'
 import { useGetConnection } from 'connection'
 import { ErrorCode } from 'connection/utils'
 import { isSupportedChain } from 'constants/chains'
@@ -91,7 +91,7 @@ export default function WalletModal({ openSettings }: { openSettings: () => void
   const [pendingConnection, setPendingConnection] = useState<Connection | undefined>()
   const [pendingError, setPendingError] = useState<any>()
 
-  const connections = useConnections()
+  const connections = getConnections()
   const getConnection = useGetConnection()
 
   useEffect(() => {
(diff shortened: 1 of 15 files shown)


src/components/WalletModal/index.tsx
```diff
@@ -108,39 +107,33 @@ export default function WalletModal({ openSettings }: { openSettings: () => void
...
-  ])
+  }, [connectedWallets, addWalletToConnectedWallets, lastActiveWalletAddress, account, connector, chainId, provider])
 
   // Used to track the state of the drawer in async function
   const drawerOpenRef = useRef(drawerOpen)
   drawerOpenRef.current = drawerOpen
 
   const tryActivation = useCallback(
     async (connection: Connection) => {
+      // Skips wallet connection if the connection should override the default behavior, i.e. install metamask or launch coinbase app
```
zzmp, line 134: This is new behavior. Why was this added? Is it related to the bug?

src/connection/utils.ts
```diff
@@ -1,15 +1,21 @@
-export const isInjected = Boolean(window.ethereum)
+import { isMobile } from 'utils/userAgent'
```
zzmp, line 1: Can you minimize risk by mapping the logic 1:1, instead of changing the abstraction (ie introducing `isMobile` in this file, and changing the mobile logic in `src/connection/index.ts`)

zzmp (changes requested):

src/connection/index.ts
```diff
@@ -50,73 +49,71 @@ const [web3Network, web3NetworkHooks] = initializeConnector<Network>(
...
-export const darkInjectedConnection: Connection = {
-  ...baseInjectedConnection,
-  icon: isGenericInjector ? INJECTED_DARK_ICON_URL : METAMASK_ICON_URL,
-}
-
-export const lightInjectedConnection: Connection = {
-  ...baseInjectedConnection,
-  icon: isGenericInjector ? INJECTED_LIGHT_ICON_URL : METAMASK_ICON_URL,
+  overrideActivate: () => {
+    if (!getIsMetaMaskWallet() && !getIsGenericInjector()) {
```
zzmp, line 77:
```suggestion
    if (getShouldAdvertiseMetaMask()) {
```

src/connection/index.ts
```diff
@@ -32,13 +30,14 @@ export enum ConnectionType {
...
   connector: Connector
   hooks: Web3ReactHooks
   type: ConnectionType
-  icon?: string
-  shouldDisplay?: boolean
-  overrideActivate?: () => void
+  // TODO(WEB-3130): add darkmode check for icons
+  getIcon?(): string
+  shouldDisplay(): boolean
+  overrideActivate?: () => boolean
```
zzmp, line 40: The callsite of `overrideActivate` must also be updated.

zzmp (changes requested):

zzmp (approved):

cmcewen (approved):

PR #6278, opened 2023-03-30
fix: catch json parse error in fetchTokenList

just-toby:
## Description

this is one of the top web sentry errors that happens when the response from fetching a token list is not valid JSON. currently the parsing error is unhandled so it's being reported to sentry. This PR updates `fetchTokenList` so that the parsing error is caught, we log a debug message similar to other errors in the function, and then continue attempting more URLs for this token list.

note that we still log an error if all the token lists fail for any reason.

#### Manual

manually cleared the token list cache and verified that the normal fetching behavior is unchanged. also, I manually hard-coded invalid json into this function to verify that we do continue to other URLs in the case of a parsing error 

#### Automated
- [x] Unit test
- [ ] Integration/E2E test

### DIFF
--- a/src/lib/hooks/useTokenList/fetchTokenList.ts
+++ b/src/lib/hooks/useTokenList/fetchTokenList.ts
@@ -51,9 +51,7 @@ export default async function fetchTokenList(
     throw new Error('Unrecognized list URL protocol.')
   }
 
-  /**
-   * Try each of the derived URLs until one succeeds.
-   */
+  // Try each of the derived URLs until one succeeds.
   for (let i = 0; i < urls.length; i++) {
     const url = urls[i]
     let response
--- a/src/lib/hooks/useTokenList/fetchTokenList.ts
+++ b/src/lib/hooks/useTokenList/fetchTokenList.ts
@@ -63,10 +63,17 @@ export default async function fetchTokenList(
       continue
     }
 
-    const json = await response.json()
-    const list = skipValidation ? json : await validateTokenList(json)
-    listCache?.set(listUrl, list)
-    return list
+    try {
+      const json = await response.json()
+      const list = skipValidation ? json : await validateTokenList(json)
+      listCache?.set(listUrl, list)
+      return list
+    } catch (error) {
+      const message = `failed to parse list response: ${listUrl}`
+      console.debug(message, error)
+      if (isLast) throw new Error(message)
+      continue
+    }
   }
 
   throw new Error('Unrecognized list URL protocol.')
--- a/src/lib/hooks/useTokenList/fetchTokenList.ts
+++ b/src/lib/hooks/useTokenList/fetchTokenList.ts
@@ -43,23 +43,24 @@ export default async function fetchTokenList(
     urls = uriToHttp(listUrl)
   }
 
+  if (urls.length === 0) {
+    throw new Error('Unrecognized list URL protocol.')
+  }
+
   for (let i = 0; i < urls.length; i++) {
     const url = urls[i]
-    const isLast = i === urls.length - 1
     let response
     try {
       response = await fetch(url, { credentials: 'omit' })
     } catch (error) {
       const message = `failed to fetch list: ${listUrl}`
       console.debug(message, error)
-      if (isLast) throw new Error(message)
       continue
     }
 
     if (!response.ok) {
       const message = `failed to fetch list: ${listUrl}`
       console.debug(message, response.statusText)
-      if (isLast) throw new Error(message)
       continue
     }
 
@@ -71,10 +72,9 @@ export default async function fetchTokenList(
     } catch (error) {
       const message = `failed to parse list response: ${listUrl}`
       console.debug(message, error)
-      if (isLast) throw new Error(message)
       continue
     }
   }
 
-  throw new Error('Unrecognized list URL protocol.')
+  throw new Error(`No valid list URL provided for ${listUrl}.`)
 }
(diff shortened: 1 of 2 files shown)


src/lib/hooks/useTokenList/fetchTokenList.ts
```diff
@@ -63,10 +63,17 @@ export default async function fetchTokenList(
...
+    try {
+      const json = await response.json()
+      const list = skipValidation ? json : await validateTokenList(json)
+      listCache?.set(listUrl, list)
+      return list
+    } catch (error) {
+      const message = `failed to parse list response: ${listUrl}`
+      console.debug(message, error)
+      if (isLast) throw new Error(message)
+      continue
```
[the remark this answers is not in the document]
tinaszheng: oh i see it's in line with the other parsers in the list, so you're not introducing a new pattern. still very odd to me though, do you know what's it's like that?

tinaszheng (approved):

src/lib/hooks/useTokenList/fetchTokenList.ts
```diff
@@ -43,31 +43,38 @@ export default async function fetchTokenList(
...
       console.debug(message, response.statusText)
-      if (isLast) throw new Error(message)
       continue
     }
 
-    const json = await response.json()
-    const list = skipValidation ? json : await validateTokenList(json)
-    listCache?.set(listUrl, list)
-    return list
+    try {
```
grabbou, line 67: Could we annotate either this function or this block with additional comments? I am feeling lost here and I think it may be confusing in the future for others too.

Why? 

My understanding (to be confirmed) is that we have an array of URLs and we resolve them on after another, unless it returned a correct response. Would be good to describe that behavior.

src/lib/hooks/useTokenList/fetchTokenList.ts
```diff
@@ -43,31 +43,38 @@ export default async function fetchTokenList(
...
+      const json = await response.json()
```
grabbou, line 68: If the response status was `ok`, when can this be an invalid json? Do we know that? I think it would be great to include that as a comment, b/c it's pretty confusing right now.

src/lib/hooks/useTokenList/fetchTokenList.ts
```diff
@@ -43,31 +43,38 @@ export default async function fetchTokenList(
...
+      const list = skipValidation ? json : await validateTokenList(json)
```
grabbou, line 69: Can `validateTokenList` function throw an error? If yes, I just wanted to point out that we swallow that error and report as `Failed to parse list response`, which can be confusing. 

I'd rather wrap just the `json` part, or update the error message.

src/lib/hooks/useTokenList/fetchTokenList.ts
```diff
@@ -43,31 +43,38 @@ export default async function fetchTokenList(
...
+      return list
+    } catch (error) {
+      const message = `failed to parse list response: ${listUrl}`
+      console.debug(message, error)
+      continue
+    }
   }
 
-  throw new Error('Unrecognized list URL protocol.')
+  throw new Error(`No valid list URL provided for ${listUrl}.`)
```
grabbou, line 79: nit: this is a bit confusing to me. No valid LIST URL provided for LIST URL. Could we rephrase that to be more specific?

grabbou (changes requested):

src/lib/hooks/useTokenList/fetchTokenList.ts
just-toby: yes i think it can. i'll update the error message to "parse and validate"

src/lib/hooks/useTokenList/fetchTokenList.ts
```diff
@@ -8,7 +8,11 @@ export const DEFAULT_TOKEN_LIST = 'https://gateway.ipfs.io/ipns/tokens.uniswap.o
 
 const listCache = new Map<string, TokenList>()
 
-/** Fetches and validates a token list. */
+/**
+ * Fetches and validates a token list.
+ * For a given token list URL, we try to fetch the list from all the possible HTTP URLs.
+ * For example, IPFS URLs can be fetched through multiple gateways.
+ **/
```
grabbou, line 15: nit: should be `*/` at the end (single `*`)

src/lib/hooks/useTokenList/fetchTokenList.ts
just-toby: the main failure case i'm trying to catch here is the one where `validateTokenList` fails, not when the `response.json()` itself is not parsable.

src/lib/hooks/useTokenList/fetchTokenList.ts
```diff
@@ -43,31 +47,40 @@ export default async function fetchTokenList(
     urls = uriToHttp(listUrl)
   }
 
+  if (urls.length === 0) {
+    throw new Error('Unrecognized list URL protocol.')
+  }
+
+  /**
+   * Try each of the derived URLs until one succeeds.
```
lynnshaoyu, line 55: I think the `/**` should be just `/*` unless this is intended to be a JSDoc which it doesn't seem like it should be?

lynnshaoyu: also since its one line could just be //

just-toby: sounds good

lynnshaoyu (approved):

PR #6281, opened 2023-03-31
fix: ignore known provider exceptions

zzmp:
### Motivation

The noisiest Sentry issue is eth_blockNumber request failure, which is expected to happen (either spuriously or because they are rate-limited) because it is polled so frequently. Logging it floods our Sentry issues.

### Summary

Ignores eth_blockNumber request failures and network changes, which should not be fatal errors. Both of these are provider-originated, non-fatal, and currently unhandled errors.

Instead of just omitting sending them to Sentry, this PR implements a global exception handler (`onunhandledrejection`) to replace the default Sentry integration (`GlobalHandlers`), so that we also do not log the unhandled error in the console.

For discussion, please use #web-sentry-squad in Slack.

#### Manual

This is difficult to test without either debugging or modifying code. I tested this locally, with a modified build, using the debugging Sentry DSN.
- [X] Induced an eth_blockNumber request failure (using the _Network request blocking_ pane) and verified that it hit the codepath and did not log nor report to Sentry
- [X] Induced a network change (by modifying code) and verified that it hit the codepath and logged but did not report to Sentry
- [x] Induced a rejection to log to Sentry, and verified that it was recorded as an `handled: no` with `mechanism: onunhandledrejection` (these are native fields to Sentry)

I then codified these behaviors with unit tests.

#### Automated
- [X] Unit test

### DIFF
--- a/src/tracing/errors.test.ts
+++ b/src/tracing/errors.test.ts
@@ -1,38 +1,22 @@
 import * as Sentry from '@sentry/react'
 
-import { beforeSendAddMechanism, onUnhandledRejection } from './errors'
+import { onUnhandledRejection } from './errors'
 
-jest.mock('@sentry/react', () => ({
-  captureException: jest.fn(),
-}))
-
-describe('beforeSendAddMechanism', () => {
-  it.each(['onunhandledrejection'])('adds mechanism to events with extra.mechanism set to "%s"', (mechanism) => {
-    const event = {
-      exception: { values: [{} as Record<'mechanism', unknown>] },
-      extra: { mechanism },
-    }
-    beforeSendAddMechanism(event as Sentry.Event)
-    expect(event.extra.mechanism).toBeUndefined()
-    expect(event.exception.values[0].mechanism).toEqual({ handled: false, type: mechanism })
-  })
-
-  it('does not add mechanism to other events', () => {
-    const event = {
-      exception: { values: [{} as Record<'mechanism', unknown>] },
-      extra: { mechanism: 'onother' },
-    }
-    beforeSendAddMechanism(event as Sentry.Event)
-    expect(event.extra.mechanism).toBe('onother')
-    expect(event.exception.values[0].mechanism).toBeUndefined()
-  })
+jest.mock('@sentry/react', () => {
+  return { getCurrentHub: jest.fn() }
 })
 
+const captureException = jest.fn()
+beforeEach(() => (Sentry.getCurrentHub as jest.Mock).mockReturnValue({ captureException }))
+
 describe('onUnhandledRejection', () => {
-  it('calls Sentry.captureException with the reason', () => {
+  it('calls Sentry.captureException', () => {
     const reason = new Error('test')
     onUnhandledRejection({ reason })
-    expect(Sentry.captureException).toHaveBeenCalledWith(reason, { extra: { mechanism: 'onunhandledrejection' } })
+    expect(Sentry.getCurrentHub().captureException).toHaveBeenCalledWith(reason, {
+      originalException: reason,
+      data: { mechanism: { handled: false, type: 'onunhandledrejection' } },
+    })
   })
 
   it('ignores eth_blockNumber exceptions', () => {
@@ -40,14 +24,14 @@ describe('onUnhandledRejection', () => {
       requestBody = JSON.stringify({ method: 'eth_blockNumber', params: [] })
     })()
     onUnhandledRejection({ reason })
-    expect(Sentry.captureException).not.toHaveBeenCalled()
+    expect(Sentry.getCurrentHub().captureException).not.toHaveBeenCalled()
   })
 
   it('ignores network change exceptions', () => {
     const reason = new Error('~~~underlying network changed~~~')
     jest.spyOn(console, 'warn').mockImplementation(() => undefined)
     onUnhandledRejection({ reason })
-    expect(Sentry.captureException).not.toHaveBeenCalled()
+    expect(Sentry.getCurrentHub().captureException).not.toHaveBeenCalled()
     expect(console.warn).toHaveBeenCalledWith(reason)
   })
 })
--- a/src/tracing/errors.test.ts
+++ b/src/tracing/errors.test.ts
@@ -1,24 +1,21 @@
 import * as Sentry from '@sentry/react'
 
-import { beforeSendAddMechanism, onerror, onunhandledrejection } from './errors'
+import { beforeSendAddMechanism, onUnhandledRejection } from './errors'
 
 jest.mock('@sentry/react', () => ({
   captureException: jest.fn(),
 }))
 
 describe('beforeSendAddMechanism', () => {
-  it.each(['onerror', 'onunhandledrejection'])(
-    'adds mechanism to events with extra.mechanism set to "%s"',
-    (mechanism) => {
-      const event = {
-        exception: { values: [{} as Record<'mechanism', unknown>] },
-        extra: { mechanism },
-      }
-      beforeSendAddMechanism(event as Sentry.Event)
-      expect(event.extra.mechanism).toBeUndefined()
-      expect(event.exception.values[0].mechanism).toEqual({ handled: false, type: mechanism })
+  it.each(['onunhandledrejection'])('adds mechanism to events with extra.mechanism set to "%s"', (mechanism) => {
+    const event = {
+      exception: { values: [{} as Record<'mechanism', unknown>] },
+      extra: { mechanism },
     }
-  )
+    beforeSendAddMechanism(event as Sentry.Event)
+    expect(event.extra.mechanism).toBeUndefined()
+    expect(event.exception.values[0].mechanism).toEqual({ handled: false, type: mechanism })
+  })
 
   it('does not add mechanism to other events', () => {
     const event = {
@@ -31,24 +28,10 @@ describe('beforeSendAddMechanism', () => {
   })
 })
 
-describe('onerror', () => {
-  it('calls Sentry.captureException with the error', () => {
-    const error = new Error('test')
-    onerror({} as Event, 'source', 1, 2, error)
-    expect(Sentry.captureException).toHaveBeenCalledWith(error, { extra: { mechanism: 'onerror' } })
-  })
-
-  it('calls Sentry.captureException with the event if there is no error', () => {
-    const event = {}
-    onerror({} as Event)
-    expect(Sentry.captureException).toHaveBeenCalledWith(event, { extra: { mechanism: 'onerror' } })
-  })
-})
-
-describe('onunhandledrejection', () => {
+describe('onUnhandledRejection', () => {
   it('calls Sentry.captureException with the reason', () => {
     const reason = new Error('test')
-    onunhandledrejection({ reason })
+    onUnhandledRejection({ reason })
     expect(Sentry.captureException).toHaveBeenCalledWith(reason, { extra: { mechanism: 'onunhandledrejection' } })
   })
 
@@ -56,14 +39,14 @@ describe('onunhandledrejection', () => {
     const reason = new (class extends Error {
       requestBody = JSON.stringify({ method: 'eth_blockNumber', params: [] })
     })()
-    onunhandledrejection({ reason })
+    onUnhandledRejection({ reason })
     expect(Sentry.captureException).not.toHaveBeenCalled()
   })
 
   it('ignores network change exceptions', () => {
     const reason = new Error('~~~underlying network changed~~~')
     jest.spyOn(console, 'warn').mockImplementation(() => undefined)
-    onunhandledrejection({ reason })
+    onUnhandledRejection({ reason })
     expect(Sentry.captureException).not.toHaveBeenCalled()
     expect(console.warn).toHaveBeenCalledWith(reason)
   })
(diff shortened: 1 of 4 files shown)


src/tracing/errors.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (event.extra.mechanism === ON_ERROR || event.extra.mechanism === ON_UNHANDLED_REJECTION) {
+      addExceptionMechanism(event, { handled: false, type: event.extra.mechanism })
+      delete event.extra.mechanism // delete this so it doesn't clutter the Sentry UI
+    }
+  }
+
+  return event
+}
+
+export function onerror(event: Event | string, source?: string, lineno?: number, colno?: number, error?: Error) {
```
grabbou, line 22: nit: prefer camel case instead of lower case, the usage of these utilities should not affect the internal naming?

src/tracing/errors.ts
```diff
@@ -0,0 +1,49 @@
...
+  Sentry.captureException(error ?? event, { extra: { mechanism: ON_ERROR } })
```
grabbou, line 23: Why do we need this if we're not filtering the errors? Can't we use the built-in mechanism then?

src/tracing/errors.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (reason instanceof Error) {
+    if (isEthersRequestError(reason)) {
+      const method = JSON.parse(reason.requestBody).method
+
+      // ethers aggressively polls for block number, and it sometimes fails (whether spuriously or through rate-limiting).
+      // If it fails, it should not be considered an exception.
+      if (method === 'eth_blockNumber') return
+    }
+
+    // If the error is a network change, it should not be considered an exception, but should still be breadcrumbed.
```
grabbou, line 36: >  but should still be breadcrumbed.

Why?

Also, little nit, I'd prefer a simpler wording than `breadcrumbed`, which may be confusing in the future.

src/tracing/errors.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (reason.message.match(/underlying network changed/)) {
+      console.warn(reason) // logging adds the message to the breadcrumbs
```
grabbou, line 38: This won't be reported unless there's an exception at the end of the day and console output is captured. In this case, you early return, so when there's no further exception, that will simply never reach Sentry.

src/tracing/errors.ts
```diff
@@ -0,0 +1,49 @@
...
+      return
+    }
+  }
+
+  Sentry.captureException(reason, { extra: { mechanism: ON_UNHANDLED_REJECTION } })
```
grabbou, line 43: Why not use "addExceptionMechanism" `https://github.com/getsentry/sentry-javascript/blob/develop/packages/utils/src/misc.ts#L86-L107` instead?

src/tracing/errors.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+/** Identifies ethers request errors (as thrown by {@type import(@ethersproject/web).fetchJson}). */
+function isEthersRequestError(error: Error): error is Error & { requestBody: string } {
+  return 'requestBody' in error && typeof (error as unknown as Record<'requestBody', unknown>).requestBody === 'string'
```
grabbou, line 48: This:
```
typeof (error as unknown as Record<'requestBody', unknown>).requestBody
```
should not be required. Since you already asserted on the presence of `requestBody`, you should just do:
```
const func = (e: Error) => {
    return 'requestBody' in e && typeof e.requestBody === 'string'
}
```
and that should pass. Here's TS repl to verify: https://www.typescriptlang.org/play?#code/MYewdgzgLgBAZgVzMGBeGAKApgLhgUQCdCRCBKNAPhgG8AoGRmQrKBQsGAchYEcEs0AEIgAJgE8uMAJacsMAGQKYUcQAcsIODCwA6PgOFjxaVOi7RCsgOZc6AXyA

src/tracing/errors.ts
just-toby, line 38: we're trying to avoid cluttering the console in prod with console logs. maybe use `debug` ?  warn is probably fine if you think it's important and doesn't happen too much though.

src/tracing/errors.test.ts
```diff
@@ -0,0 +1,70 @@
+import * as Sentry from '@sentry/react'
+
+import { beforeSendAddMechanism, onerror, onunhandledrejection } from './errors'
+
+jest.mock('@sentry/react', () => ({
+  captureException: jest.fn(),
+}))
+
+describe('beforeSendAddMechanism', () => {
```
just-toby, line 9: it looks like the coverage isn't being reported for this code. can you add it to the `collectCoverageFrom` set in package.json

just-toby (changes requested):

src/tracing/errors.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export function onunhandledrejection({ reason }: { reason: unknown }) {
+  if (reason instanceof Error) {
+    if (isEthersRequestError(reason)) {
+      const method = JSON.parse(reason.requestBody).method
+
+      // ethers aggressively polls for block number, and it sometimes fails (whether spuriously or through rate-limiting).
+      // If it fails, it should not be considered an exception.
+      if (method === 'eth_blockNumber') return
```
grabbou, line 33: For some reason, I can't access Sentry anymore, so I can't check the stack trace of this error. Could you share it here? 

What I am thinking: is there any place in the code in particular that triggers this, that we could possibly wrap with a try/catch and avoid all that?

src/tracing/errors.ts
```diff
@@ -0,0 +1,49 @@
...
+import { addExceptionMechanism } from '@sentry/utils'
+
+const ON_ERROR = 'onerror'
+const ON_UNHANDLED_REJECTION = 'onunhandledrejection'
+
+/** Adds a mechanism to events which have been sent from our global exception handlers. */
+export function beforeSendAddMechanism(event: Sentry.Event): Sentry.Event {
+  // Global exception handlers set an extra.mechanism, but the actual mechanism must be set on the event explicitly for
+  // Sentry to correctly process and display it in the Sentry UI. We still want mechanism to be set, despite using our
+  // own handler, because it aids in identifying unhandled errors and rejections.
```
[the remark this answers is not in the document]
zzmp: `/**` is reserved for doc-comments. `//` is still valid for multi-line, and unless we lint for this I prefer it for being more terse (doesn't require the trailing blank line).

> nit: shouldn't the comment be above the definition, not inside function body?
the doc comment above the function declaration explains what it does, but _how_ it does it is an implementation detail, so I put it at the point of implementation (within the body) to be closer to the code that it explains.

src/tracing/errors.ts
```diff
@@ -0,0 +1,49 @@
+import * as Sentry from '@sentry/react'
+import { addExceptionMechanism } from '@sentry/utils'
+
+const ON_ERROR = 'onerror'
+const ON_UNHANDLED_REJECTION = 'onunhandledrejection'
+
+/** Adds a mechanism to events which have been sent from our global exception handlers. */
```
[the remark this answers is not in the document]
zzmp: This is a doc-comment, not inline. If it is short enough for one line, I prefer the one-line form for terseness.

src/tracing/errors.ts
zzmp: yes! great callout.

src/tracing/errors.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  return event
+}
+
+export function onerror(event: Event | string, source?: string, lineno?: number, colno?: number, error?: Error) {
+  Sentry.captureException(error ?? event, { extra: { mechanism: ON_ERROR } })
+}
+
+export function onunhandledrejection({ reason }: { reason: unknown }) {
```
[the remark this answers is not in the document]
zzmp: per your other suggestion, I am using Sentry's default global handler for `onerror` (which is more tested - by them), so our implementation will be specialized to `onunhandledrejection`.

src/tracing/errors.ts
zzmp: excellent point, I'm referring to sentry breadcrumbs left by captureconsole. I will make that more clear in the comment.

src/tracing/errors.ts
zzmp: If an exception is reported, the last 100 breadcrumbs are captured (including logs and fetches), so this is expected behavior.

src/tracing/errors.ts
zzmp: `addExceptionMechanism` only acts on a `Sentry.Event`, and it is non-trivial to transform an exception to an event. My first draft created a custom event and called `addExceptionMechanism` within `onUnhandledRejection`, and I found it much cleaner to move that to `beforeSendAddMechanism`.

A `Sentry.Event` is not available as an output of `captureException` - it is hidden as an implementation detail. The only way to add the mechanism that I found was creating the event yourself (which is undocumented, and it seems likely that I'd miss some error context to send), or to intercept the event in `beforeSend` and augment it there.

src/tracing/errors.ts
zzmp: I agree with you in theory, and it passes in my local typechecker, but it fails somewhere in the craco pipeline.

src/tracing/errors.ts
```diff
@@ -0,0 +1,49 @@
+import * as Sentry from '@sentry/react'
+import { addExceptionMechanism } from '@sentry/utils'
+
+const ON_ERROR = 'onerror'
+const ON_UNHANDLED_REJECTION = 'onunhandledrejection'
+
+/** Adds a mechanism to events which have been sent from our global exception handlers. */
+export function beforeSendAddMechanism(event: Sentry.Event): Sentry.Event {
```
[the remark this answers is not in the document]
zzmp: This is difficult because we then have to build our own `Sentry.Event`, which is non-trivial and poorly documented.

src/tracing/errors.ts
```diff
@@ -0,0 +1,49 @@
...
+export function onerror(event: Event | string, source?: string, lineno?: number, colno?: number, error?: Error) {
+  Sentry.captureException(error ?? event, { extra: { mechanism: ON_ERROR } })
+}
+
+export function onunhandledrejection({ reason }: { reason: unknown }) {
+  if (reason instanceof Error) {
+    if (isEthersRequestError(reason)) {
+      const method = JSON.parse(reason.requestBody).method
+
+      // ethers aggressively polls for block number, and it sometimes fails (whether spuriously or through rate-limiting).
```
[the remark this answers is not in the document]
zzmp: `ethers` is a package name, and should not be capitalized.

src/tracing/errors.ts
zzmp: It is inconsistent and comes from polling (async operations), so I do not think there is a single point where we could wrap and catch.

src/tracing/errors.ts
zzmp: I put a briefer explanation in the doc-comment of `beforeSendAddMechanism`.

src/tracing/errors.ts
grabbou: I think this is a fair point tho @zzmp, what is the value of console logging in production? I always thought it was rather an anti-pattern. If we want to collect that data and act on it, we should probably collect it in a different way?

src/tracing/errors.ts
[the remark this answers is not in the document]
grabbou: > /** is reserved for doc-comments. 

Could you share a reference document for this? 

----

Did you end up finding out what is the recommended approach for multi-line comments as per Google / Airbnb? I personally think that we shouldn't use `//` for multiline comments, I don't think the trailing blank line outweighs better IDE support for multiline (which adds new lines automatically and makes it easier to refactor).

> the doc comment above the function declaration explains what it does, but how it does it is an implementation detail, so I put it at the point of implementation (within the body) to be closer to the code that it explains.

Overall I agree with you, but again, I think this in this case, your comment is not "how", it's more of a "why", esp.:
> Global exception handlers set an extra mechanism, but the actual mechanism must be set on the event explicitly for
>  // Sentry to correctly process and display it in the Sentry UI.

and that's why I think it should go above the function body. But again, this is a purely judgmental and personal preference. Maybe adding a function comment explaining why we need this hook, what's its purpose and where it's used would be good enough, as an addition to an existing comment.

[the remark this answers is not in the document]
grabbou: Okay, I see you added the top-level comment (didn't see it before), so I guess that makes more sense now.

grabbou (approved):
Thanks for applying the changes, the PR looks good to me. Not going to lie, I am not a fan - and I am unhappy that there's no better way in Sentry to ignore errors, like in our use case. However, this is currently our best take to achieve this and while I personally don't like some of the code in this PR, there's no other way to do it.

zzmp:
> Thanks for applying the changes, the PR looks good to me. Not going to lie, I am not a fan - wish there was a better way to ignore exceptions than having to build a custom reporter and most importantly, to use a special hook to attach some (to my understanding) internal properties for better reporting in the dashboard.
> 
> However, this is currently our best take to achieve this.

I've opened an issue through Sentry's GitHub to improve their API and allow this: https://github.com/getsentry/sentry-javascript/issues/7716.
Unfortunately, this does seem to be the best way to go about it for the time being.

zzmp:
I think this should be closed in favor of #6267, as that uses Sentry's `GlobalHandlers` integration which should have better handling of edge cases.

PR #6282, opened 2023-03-31
feat: reduce severity of phishing filter to allow url token names

JFrankfurt:
this is a lower-complexity version of #6249

https://uniswaplabs.atlassian.net/browse/WEB-2802?atlOrigin=eyJpIjoiMDM1OGMyZTg4ODMzNGFhMTk2NTQ4YjhjZTg1Y2UxNTkiLCJwIjoiaiJ9

test by walletconnecting with 0x11E4857Bb9993a50c685A79AFad4E6F65D518DDa and viewing LP positions on Optimism

### DIFF
--- a/src/utils/urlChecks.ts
+++ b/src/utils/urlChecks.ts
@@ -1,4 +1,4 @@
-export function checkStringForURL(str?: string): boolean {
+export function hasURL(str?: string): boolean {
   if (!str) return false
   const pattern = new RegExp(
     '([a-zA-Z0-9]+://)?' + // optional protocol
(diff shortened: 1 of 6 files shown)


cmcewen (commented):
this looks good to me but codecov would like some more coverage 🙏

tinaszheng (approved):

src/utils/urlChecks.ts
```diff
@@ -1,4 +1,4 @@
-export function hasURL(str?: string): boolean {
+export function checkStringForURL(str?: string): boolean {
```
tinaszheng, line 1: i kind of prefer the `hasURL` version tbh, it's more clear that it returns a boolean

this kind of feels like it's more of a validation function

JFrankfurt: ok, @tinaszheng gets `hasURL` and @cmcewen gets `shouldHidePosition`

deal? 😅

cmcewen: Deal 🤝

cmcewen (approved):

PR #6284, opened 2023-03-31
fix: only use local txs for current account

cartcrom:
## Description

Updates useLocalActivity to only return activity for TX's that are related to a given account

https://uniswaplabs.atlassian.net/browse/WEB-3106

## Screen Capture
| Before (local tx's from diff account)           | After           |
| --------- |-----------------|
|  image | image  |

#### Manual

1. Make a transaction (i.e. swap) and view it on the acitivty tab
2. Switch to another account
3. Ensure the previous account's transaction is no longer populating the activity tab

#### Automated
Added a test in parseLocal.test.tsx that confirms the right number of transactions appear for the input account when multiple accounts have tx history in redux store

### DIFF
--- a/src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.test.ts
+++ b/src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.test.ts
@@ -1,7 +1,7 @@
 // jest unit tests for the parseLocalActivity function
 
-import { SupportedChainId, Token, TradeType } from '@uniswap/sdk-core'
-import { DAI, USDC_MAINNET } from 'constants/tokens'
+import { SupportedChainId, Token, TradeType as MockTradeType } from '@uniswap/sdk-core'
+import { DAI as MockDAI, USDC_MAINNET as MockUSDC_MAINNET } from 'constants/tokens'
 import { TokenAddressMap } from 'state/lists/hooks'
 import { WrappedTokenInfo } from 'state/lists/wrappedTokenInfo'
 import {
@@ -10,23 +10,24 @@ import {
   TransactionDetails,
   TransactionType,
 } from 'state/transactions/types'
+import { renderHook } from 'test-utils'
 
-import { parseLocalActivity } from './parseLocal'
+import { parseLocalActivity, useLocalActivities } from './parseLocal'
 
 const oneUSDCRaw = '1000000'
 const oneDAIRaw = '1000000000000000000'
 
-function buildSwapInfo(
-  type: TradeType,
+function mockSwapInfo(
+  type: MockTradeType,
   inputCurrency: Token,
   inputCurrencyAmountRaw: string,
   outputCurrency: Token,
   outputCurrencyAmountRaw: string
 ): ExactInputSwapTransactionInfo | ExactOutputSwapTransactionInfo {
-  if (type === TradeType.EXACT_INPUT) {
+  if (type === MockTradeType.EXACT_INPUT) {
     return {
       type: TransactionType.SWAP,
-      tradeType: TradeType.EXACT_INPUT,
+      tradeType: MockTradeType.EXACT_INPUT,
       inputCurrencyId: inputCurrency.address,
       inputCurrencyAmountRaw,
       outputCurrencyId: outputCurrency.address,
@@ -36,7 +37,7 @@ function buildSwapInfo(
   } else {
     return {
       type: TransactionType.SWAP,
-      tradeType: TradeType.EXACT_OUTPUT,
+      tradeType: MockTradeType.EXACT_OUTPUT,
       inputCurrencyId: inputCurrency.address,
       expectedInputCurrencyAmountRaw: inputCurrencyAmountRaw,
       maximumInputCurrencyAmountRaw: inputCurrencyAmountRaw,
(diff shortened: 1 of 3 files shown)


src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.test.ts
```diff
@@ -10,23 +10,24 @@ import {
...
 
-import { parseLocalActivity } from './parseLocal'
+import { parseLocalActivity, useLocalActivities } from './parseLocal'
 
 const oneUSDCRaw = '1000000'
 const oneDAIRaw = '1000000000000000000'
 
-function buildSwapInfo(
-  type: TradeType,
+function mockSwapInfo(
```
cartcrom, line 20: var's have to be prefixed with "mock" in order to be used in then jest.mock fn scope below

src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.ts
```diff
@@ -182,24 +181,22 @@ export function parseLocalActivity(
...
 
-export function useLocalActivities(): ActivityMap | undefined {
+export function useLocalActivities(account: string): ActivityMap {
   const allTransactions = useMultichainTransactions()
-  const { chainId } = useWeb3React()
   const tokens = useCombinedActiveList()
 
   return useMemo(
     () =>
-      chainId
```
cartcrom, line 192: this check for useWeb3React.chainId was completely redundant / leftover from some refactor I made, so I removed it here

src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.ts
```diff
@@ -182,24 +181,22 @@ export function parseLocalActivity(
...
-          }, {})
-        : undefined,
-    [allTransactions, chainId, tokens]
+      allTransactions.reduce((acc: { [hash: string]: Activity }, [transaction, chainId]) => {
+        if (transaction.from !== account) return acc
+        try {
+          const localActivity = parseLocalActivity(transaction, chainId, tokens)
+          if (localActivity) acc[localActivity.hash] = localActivity
+        } catch (error) {
+          console.error('Failed to parse local activity', transaction)
```
just-toby, line 196: does this need to be an error, or can we use `console.debug`

just-toby: also can you try to cover this line in a test

src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.ts
```diff
@@ -182,24 +181,22 @@ export function parseLocalActivity(
   return { ...defaultFields, ...additionalFields }
 }
 
-export function useLocalActivities(): ActivityMap | undefined {
+export function useLocalActivities(account: string): ActivityMap {
   const allTransactions = useMultichainTransactions()
```
just-toby, line 185: why aren't we storing local transactions in redux on a per-address basis? is there any use case for combining transactions from all addresses? it could be another potential fix that doesn't require manual filtering.

cartcrom: I'm not sure why we store them this way, but I don't want to modify the redux tx flow at all until the backend is able to serve l2 tx's, since localStorage is currently our users only source of l2 tx's and I imagine changing how tx's are stored would result in client browsers loosing reference to that data

src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.ts
cartcrom: I was able to remove the try/catch it was redundant since everything is strictly typed and we're properly checking undefined from token maps now

cartcrom: GH is complaining abt about another line that is missing coverage now but I plan to cover that in [this ticket](https://uniswaplabs.atlassian.net/browse/WEB-3113)

src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.ts
```diff
@@ -182,24 +181,20 @@ export function parseLocalActivity(
...
-              const localActivity = parseLocalActivity(transaction, chainId, tokens)
-              if (localActivity) acc[localActivity.hash] = localActivity
-            } catch (error) {
-              console.error('Failed to parse local activity', transaction)
-            }
-            return acc
-          }, {})
-        : undefined,
-    [allTransactions, chainId, tokens]
+      allTransactions.reduce((acc: { [hash: string]: Activity }, [transaction, chainId]) => {
```
just-toby, line 190: can we rename `acc` to something more descriptive? maybe `filteredActivitiesByHash` or something like that

src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.ts
```diff
@@ -182,24 +181,20 @@ export function parseLocalActivity(
...
+        if (transaction.from === account) {
+          const localActivity = parseLocalActivity(transaction, chainId, tokens)
```
just-toby, line 192: are you confident that  `parseLocalActivity` can't throw anymore, at all? can it fail in some unexpected way unrelated to the tokens existing in the lists?

just-toby (changes requested):

src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.ts
```diff
@@ -182,24 +181,20 @@ export function parseLocalActivity(
...
-              if (localActivity) acc[localActivity.hash] = localActivity
-            } catch (error) {
-              console.error('Failed to parse local activity', transaction)
-            }
-            return acc
-          }, {})
-        : undefined,
-    [allTransactions, chainId, tokens]
+      allTransactions.reduce((acc: { [hash: string]: Activity }, [transaction, chainId]) => {
+        if (transaction.from === account) {
```
tinaszheng, line 191: would prefer to have this guard style so you can short-circuit if there's nothing to do, ie:

 ```
if (transaction.from !== account) return acc

const localActivity = ...
```

File: src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.ts (hunk 5 above)
Comment (tinaszheng) [inline src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.ts:190]: i also feel like this `.reduce` can just be turned into a for loop for readability. since you're just modifying the `acc` object in place this isn't a true reduce in the sense that it's a pure function 

would probably be better to do something like
```
const activityByHash = {}

for (const transaction of allTransactions) {
   // add to activityByHash
}

return activityByHash

src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.ts
cartcrom: I can add it back. I wrote it w/ the try/catch as a precaution but now it feels off to me that I call this code "parsing" and use try/catch, when ideally it'd be treated more as "adapting" & since these are all front-end types rely on the type-catcher rather than try/catch.

Or does it make sense to try/catch code that adapts types fetched from localStorage? I.e. in case a redux type is updated and the browser is providing objects w/ invalid fields or something like that? In that case I'm confident keeping the try/catch I just wanna make sure I'm following best practices

src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.ts
cartcrom: ty for this. I've been using reduce way too much for no good reason, this is way cleaner

src/components/WalletDropdown/MiniPortfolio/Activity/parseLocal.ts
tinaszheng: oh the error here is from looking up currencies that don't exist in our token map (like user imported ones) - we had an incident last week where we fixed the logic on tx notification by handling the `undefined` case but i'm not sure it's being handled here yet

cartcrom: @tinaszheng I think @just-toby fixed that by adding ? checks to getCurrency in this file. That fix should prevent errors for anything that uses parseLocalActivity (eddie's tx notification flow and the activity tab on MP). I'm wondering if the try/catch is still important tho for the other reasons I mentioned above

tinaszheng: ohh i see i see i think you're confident that no other issues can throw then fine to take out 🤔

tinaszheng (approved):
nice work carter!!

PR #6286, opened 2023-03-31
fix: e2e test coverage reporting

just-toby:
## Description

cypress coverage reporting wasn't actually working in builds after #6212 because craco always sets the babel environment to `production` when running yarn build.

when fixing this, I uncovered other issues with `babel-plugin-istanbul` causing the app to crash for various reasons. together with @zzmp  we decided the simplest solution would be to allow `unsafe-eval` in our CSP during e2e tests.

https://uniswaplabs.atlassian.net/browse/WEB-3061?atlOrigin=eyJpIjoiZmNiODkxMTYxMGY5NDk5Njg1MzBmNTczYTY3OWFhOWIiLCJwIjoiaiJ9

#### Manual

manually tested locally:
-  building and serving the instrumented version of the app, checking that `window.__coverage__` is defined and correct. 
- building the non-instrumented version and making sure there's no coverage data
- check the github action and codecov results

#### Automated
- [x] <s>Unit test</s>
- [x] Integration/E2E test


zzmp:
Note that we already disable chrome security for cypress tests, so this is not a big departure from that.

zzmp (approved):

.github/workflows/test.yml
```diff
@@ -49,6 +49,19 @@ jobs:
           path: build
           if-no-files-found: error
 
+  build-instrumented:
```
cmcewen, line 52: i'm not sure what instrumented means exactly. i would vote to name this
`build-e2e` and `yarn build:e2e` so that i know that this is a different build flavor meant specifically for e2e tests. instrumented could be for production or something else

craco.config.cjs
```diff
@@ -7,15 +7,30 @@ const commitHash = require('child_process').execSync('git rev-parse HEAD')
...
-      test: {
-        plugins: ['istanbul'],
-      },
-      development: {
-        plugins: ['istanbul'],
-      },
-    },
+    plugins: [
+      '@vanilla-extract/babel-plugin',
+      ...(process.env.REACT_APP_UNSAFE_EVAL
```
cmcewen, line 12: seems like you're mixing concerns here with using this env variable. could you set something like ENV=E2E_TEST or something like that?

package.json
```diff
@@ -19,6 +19,7 @@
     "prepare": "yarn contracts:compile && yarn graphql:fetch && yarn graphql:generate && yarn i18n:compile",
     "start": "craco start",
     "build": "craco build",
+    "build:instrumented": "REACT_APP_UNSAFE_EVAL=true craco --max_old_space_size=4096 build",
```
cmcewen, line 22: i'd prefer to name this something like `REACT_APP_OVERRIDE_CSP` or something, potentially with the value being the new CSP rule vs UNSAFE_EVAL. this current name requires knowledge of the CSP rule unsafe-eval, so i'd like it to be more obvious what the env variable is doing (overriding the csp).

cmcewen (changes requested):

cmcewen (approved):

.github/workflows/test.yml
```diff
@@ -85,6 +85,19 @@ jobs:
...
+      - run: yarn build:e2e
+      - uses: actions/upload-artifact@v2
+        with:
+          name: build-e2e
+          path: build
+          if-no-files-found: error
+
   size-tests:
     needs: [build]
     runs-on: ubuntu-latest
```
zzmp, line 103: Can you move this to be in the same step as `build`, and rename `build` to `size-tests`? Then you won't need the upload action in it, and it should make it run faster overall.

.github/workflows/test.yml
```diff
@@ -85,6 +85,19 @@ jobs:
           path: build
           if-no-files-found: error
 
+  build-e2e:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup
+      - run: yarn prepare
+      - run: yarn build:e2e
```
zzmp, line 94:
```suggestion
      - run: yarn build:e2e
        env:
          NODE_OPTIONS: "--max_old_space_size=4096"
```

zzmp (changes requested):

.github/workflows/test.yml
```diff
@@ -79,27 +79,25 @@ jobs:
...
-          if-no-files-found: error
+      - run: yarn test:size
 
-  size-tests:
-    needs: [build]
+  build-e2e:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup
```
zzmp, line 88: caching will speed this up by 2x.

```suggestion
      - uses: ./.github/actions/setup
      - uses: actions/cache@v3
        id: build-e2e-cache
        with:
          path: node_modules/.cache
          key: ${{ runner.os }}-build-e2e-${{ hashFiles('**/yarn.lock') }}-${{ github.run_id }}
```

zzmp (changes requested):

.github/workflows/test.yml
```diff
@@ -149,4 +169,4 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - if: needs.cypress-test-matrix.result != 'success'
-        run: exit 1
+        run: exit 1
```
zzmp, line 172: add back newline at EOF

zzmp (approved):

zzmp (approved):

PR #6293, opened 2023-04-03
refactor: token map types

cartcrom:
## Description

The existing type `TokenAddressMap` was misnomered as it actually was a nested map from `chainId` to `tokenAddress` to `{ token: WrappedTokenInfo; list?: TokenList }`

This PR removes the redundant renaming/re-exportation of `ChainTokenMap` to `TokenAddressMap` and renames `ChainTokenMap` to `ChainToTokenInfoMap`. 

Since the confusion of `TokenAddressMap` is now removed, this PR also adds  a simple `TokenMap` type to be used anywhere a map from token address to token is needed. `TokenMap` maps to `Token | undefined` to prevent undefined access errors

`ChainToTokenInfoMap` is also updated to map to `{ token: WrappedTokenInfo; list?: TokenList } | undefined` to prevent similar errors

#### Manual
This is a refactor PR that should not affect any functionality

### DIFF
--- a/src/components/SearchModal/CurrencySearch.tsx
+++ b/src/components/SearchModal/CurrencySearch.tsx
@@ -87,7 +87,8 @@ export function CurrencySearch({
 
   const defaultTokens = useDefaultActiveTokens()
   const filteredTokens: Token[] = useMemo(() => {
-    return Object.values(defaultTokens).filter(getTokenFilter(debouncedQuery))
+    const queryFilter = getTokenFilter(debouncedQuery)
+    return Object.values(defaultTokens).filter((token): token is Token => Boolean(token && queryFilter(token)))
   }, [defaultTokens, debouncedQuery])
 
   const [balances, balancesAreLoading] = useAllTokenBalances()
--- a/src/components/SearchModal/CurrencySearch.tsx
+++ b/src/components/SearchModal/CurrencySearch.tsx
@@ -19,7 +19,13 @@ import { useAllTokenBalances } from 'state/connection/hooks'
 import styled, { useTheme } from 'styled-components/macro'
 import { UserAddedToken } from 'types/tokens'
 
-import { useDefaultActiveTokens, useIsUserAddedToken, useSearchInactiveTokenLists, useToken } from '../../hooks/Tokens'
+import {
+  tokenMapToArray,
+  useDefaultActiveTokens,
+  useIsUserAddedToken,
+  useSearchInactiveTokenLists,
+  useToken,
+} from '../../hooks/Tokens'
 import { CloseIcon, ThemedText } from '../../theme'
 import { isAddress } from '../../utils'
 import Column from '../Column'
@@ -85,11 +91,11 @@ export function CurrencySearch({
     }
   }, [isAddressSearch])
 
-  const defaultTokens = useDefaultActiveTokens()
+  const defaultTokenMap = useDefaultActiveTokens()
   const filteredTokens: Token[] = useMemo(() => {
-    const queryFilter = getTokenFilter(debouncedQuery)
-    return Object.values(defaultTokens).filter((token): token is Token => Boolean(token && queryFilter(token)))
-  }, [defaultTokens, debouncedQuery])
+    const defaultTokens = tokenMapToArray(defaultTokenMap)
+    return defaultTokens.filter(getTokenFilter(debouncedQuery))
+  }, [defaultTokenMap, debouncedQuery])
 
   const [balances, balancesAreLoading] = useAllTokenBalances()
   const sortedTokens: Token[] = useMemo(
--- a/src/hooks/Tokens.ts
+++ b/src/hooks/Tokens.ts
@@ -5,7 +5,7 @@ import { SupportedChainId } from 'constants/chains'
 import { DEFAULT_INACTIVE_LIST_URLS, DEFAULT_LIST_OF_LISTS } from 'constants/lists'
 import { useCurrencyFromMap, useTokenFromMapOrNetwork } from 'lib/hooks/useCurrency'
 import { getTokenFilter } from 'lib/hooks/useTokenList/filtering'
-import { ChainToTokenInfoMap } from 'lib/hooks/useTokenList/utils'
+import { ChainTokenInfoMap } from 'lib/hooks/useTokenList/utils'
 import { useMemo } from 'react'
 import { isL2ChainId } from 'utils/chains'
 
@@ -15,22 +15,26 @@ import { useUserAddedTokens, useUserAddedTokensOnChain } from '../state/user/hoo
 import { useUnsupportedTokenList } from './../state/lists/hooks'
 
 export type TokenMap = { [address: string]: Token | undefined }
-// reduce token map into standard address <-> Token mapping, optionally include user added tokens
-function useTokensFromMap(tokenMap: ChainToTokenInfoMap): TokenMap {
+export function tokenMapToArray(tokens: TokenMap): Array<Token> {
+  return Object.values(tokens).filter((token): token is Token => token !== undefined)
+}
+
+// reduce ChainTokenInfoMap to standard TokenMap type
+function useTokensFromMap(chainTokenInfoMap: ChainTokenInfoMap): TokenMap {
   const { chainId } = useWeb3React()
   return useMemo(() => {
     if (!chainId) return {}
 
     // reduce to just tokens
-    return Object.keys(tokenMap[chainId] ?? {}).reduce<TokenMap>((newMap, address) => {
-      const entry = tokenMap[chainId][address]
+    return Object.keys(chainTokenInfoMap[chainId] ?? {}).reduce<TokenMap>((newMap, address) => {
+      const entry = chainTokenInfoMap[chainId]?.[address]
       if (entry) newMap[address] = entry.token
       return newMap
     }, {})
-  }, [chainId, tokenMap])
+  }, [chainId, chainTokenInfoMap])
 }
 
-export function useAllTokensMultichain(): ChainToTokenInfoMap {
+export function useAllTokensMultichain(): ChainTokenInfoMap {
   return useCombinedTokenMapFromUrls(DEFAULT_LIST_OF_LISTS)
 }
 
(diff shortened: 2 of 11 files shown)


src/lib/hooks/useTokenList/utils.ts
```diff
@@ -1,21 +1,21 @@
 import { TokenInfo, TokenList } from '@uniswap/token-lists'
 import { WrappedTokenInfo } from 'state/lists/wrappedTokenInfo'
 
-type TokenMap = Readonly<{ [tokenAddress: string]: { token: WrappedTokenInfo; list?: TokenList } }>
-export type ChainTokenMap = Readonly<{ [chainId: number]: TokenMap }>
+type TokenInfoMap = Readonly<{ [tokenAddress: string]: { token: WrappedTokenInfo; list?: TokenList } | undefined }>
+export type ChainToTokenInfoMap = Readonly<{ [chainId: number]: TokenInfoMap }>
```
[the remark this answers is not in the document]
zzmp: Could this typing be `SupportedChainId` to `TokenInfoMap`, instead of the looser `number`?

zzmp (changes requested):

src/lib/hooks/useTokenList/utils.ts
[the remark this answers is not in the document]
cartcrom: Will make TokenInfoMap undefineable (is there a good word equivalent to nullable for undefined...?). 

Did you mean to suggest both the optional `?` and the `| undefined`?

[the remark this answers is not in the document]
cartcrom: And yes, will change to ChainTokenInfoMap. I was trying to differentiate it somehow since it's a map to a map... because technically the fully correct version is ChainTokenInfoMapMap... but that feels worse.

src/lib/hooks/useTokenList/utils.ts
```diff
@@ -1,21 +1,21 @@
 import { TokenInfo, TokenList } from '@uniswap/token-lists'
 import { WrappedTokenInfo } from 'state/lists/wrappedTokenInfo'
 
-type TokenMap = Readonly<{ [tokenAddress: string]: { token: WrappedTokenInfo; list?: TokenList } }>
-export type ChainTokenMap = Readonly<{ [chainId: number]: TokenMap }>
+type TokenInfoMap = Readonly<{ [tokenAddress: string]: { token: WrappedTokenInfo; list?: TokenList } | undefined }>
```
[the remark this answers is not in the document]
cartcrom: I'm not sure what you mean? The `list` field is the only optional part of this

src/components/SearchModal/CurrencySearch.tsx
```diff
@@ -87,7 +87,8 @@ export function CurrencySearch({
 
   const defaultTokens = useDefaultActiveTokens()
   const filteredTokens: Token[] = useMemo(() => {
-    return Object.values(defaultTokens).filter(getTokenFilter(debouncedQuery))
+    const queryFilter = getTokenFilter(debouncedQuery)
+    return Object.values(defaultTokens).filter((token): token is Token => Boolean(token && queryFilter(token)))
```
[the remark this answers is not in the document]
cartcrom: I added a util for converting TokenMap -> Token[], as I suspect it'll be useful elsewhere in the future

src/lib/hooks/useTokenList/utils.ts
```diff
@@ -1,36 +1,38 @@
...
         return map
       }
-      if (!map[token.chainId]) {
-        map[token.chainId] = {}
+      const currentChainMap = map[token.chainId]
+      if (!currentChainMap) {
+        map[token.chainId] = { [token.address]: { token, list } }
+      } else {
+        currentChainMap[token.address] = { token, list }
       }
```
zzmp, line 30: I think this is more confusing, because you're referring to the value as `currentChainMap` in the conditional but `map[...]` in the gated assignment. Would you consider:

```suggestion
      const currentChainMap = map[token.chainId] || {}
      currentChainMap[token.address] = { token, list }
```

zzmp: It may need to be `... || {} as TokenInfoMap`.

nit: Also, consider renaming the variable to `tokenInfoMap` - its confusing to have `Chain` in there.

src/lib/hooks/useTokenList/utils.ts
```diff
@@ -1,36 +1,38 @@
...
-      map[token.chainId][token.address] = { token, list }
       return map
     } catch {
       return map
     }
-  }, {}) as ChainTokenMap
+  }, {}) as ChainTokenInfoMap
```
zzmp, line 35: I think you could reduce the coercions by casting the initial object, and be slightly safer:

```suggestion
  }, {} as ChainTokenInfoMap)
```

src/hooks/Tokens.ts
```diff
@@ -5,42 +5,49 @@ import { SupportedChainId } from 'constants/chains'
...
 
 import { useAllLists, useCombinedActiveList, useCombinedTokenMapFromUrls } from '../state/lists/hooks'
 import { WrappedTokenInfo } from '../state/lists/wrappedTokenInfo'
 import { useUserAddedTokens, useUserAddedTokensOnChain } from '../state/user/hooks'
-import { TokenAddressMap, useUnsupportedTokenList } from './../state/lists/hooks'
+import { useUnsupportedTokenList } from './../state/lists/hooks'
 
-// reduce token map into standard address <-> Token mapping, optionally include user added tokens
-function useTokensFromMap(tokenMap: TokenAddressMap): { [address: string]: Token } {
+export type TokenMap = { [address: string]: Token | undefined }
```
zzmp, line 17: Why isn't this just `TokenInfoMap`?

src/hooks/Tokens.ts
```diff
@@ -5,42 +5,49 @@ import { SupportedChainId } from 'constants/chains'
...
+export function tokenMapToArray(tokens: TokenMap): Array<Token> {
+  return Object.values(tokens).filter((token): token is Token => token !== undefined)
```
zzmp, line 19: Can you just use `Boolean` here?

```suggestion
  return Object.values(tokens).filter<Token>(Boolean)
```

src/hooks/Tokens.ts
```diff
@@ -5,42 +5,49 @@ import { SupportedChainId } from 'constants/chains'
...
+}
+
+// reduce ChainTokenInfoMap to standard TokenMap type
```
zzmp, line 22: This should be a doc-comment:

```suggestion
/** Reduces ChainTokenInfoMap to the standard TokenMap type. */
```

zzmp (changes requested):

src/state/user/hooks.tsx
```diff
@@ -275,7 +275,7 @@ export function useTrackedTokenPairs(): [Token, Token][] {
               (BASES_TO_TRACK_LIQUIDITY_FOR[chainId] ?? [])
                 // to construct pairs of the given token with each base
                 .map((base) => {
-                  if (base.address === token.address) {
+                  if (base.address === token?.address) {
```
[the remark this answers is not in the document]
cartcrom: this code is for v2 flows that I don't have any context on, so the goal here was to leave the behavior unchanged and change the code as little as possible

src/lib/hooks/useTokenList/utils.ts
cartcrom: It seems to have been written this way to allow the ChainTokenInfoMap to be mutable while being constructed, and then casts it as the readonly version afterwards.

src/hooks/Tokens.ts
cartcrom: TokenInfoMap is more complex/specific than needed for more general token mapping usage:
`type TokenInfoMap = Readonly<{ [tokenAddress in string]?: { token: WrappedTokenInfo; list?: TokenList } }>`

src/hooks/Tokens.ts
cartcrom: Boolean filtering doesn't appease the type-checker; some form of cast/guard is needed.  do you think
```suggestion
  return Object.values(tokens).filter(Boolean) as Array<Token>
```
is better?

cartcrom:
> hmmm so i think `export type TokenMap = { [address in string]?: Token }` is technically nottt correct here and the right move is to keep the `{ [address in string]: Token }` convention but enable the linter rule
> 
> while they would both make `tokenMap[a]` typed as `Token | undefined` the difference is that: Object.values(TokenMap instance) should be Token[] but yours would be `(Token | undefined)[]`, which makes some of the downstream types harder 🤔

turning `noUncheckedIndexedAccess` to true in tsconfig.json causes 383 problems in the vscode terminal / places that checks need to be added 😳. @tinaszheng @zzmp should this TokenMap change come after we do all of the refactors for this rule, or should we push this as is for now, and remove the ? from the types in a TODO later?

tinaszheng:
@cartcrom one way you can iterate through fixing a huge lint issue like that is to turn it on locally, fix the types associated with just `TokenMap` stuff, and then turn it off for merging in a PR. @zzmp was going to take the lead on enabling the lint rule, but that's one way we could break up fixing all the issues without one massive PR

src/state/user/hooks.tsx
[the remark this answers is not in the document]
zzmp: I can see this being an issue that's just being exposed by the new typing, so I think you should follow the codepath to resolve it as part of this PR.

zzmp (approved):

cartcrom:
> @cartcrom one way you can iterate through fixing a huge lint issue like that is to turn it on locally, fix the types associated with just `TokenMap` stuff, and then turn it off for merging in a PR. @zzmp was going to take the lead on enabling the lint rule, but that's one way we could break up fixing all the issues without one massive PR

@tinaszheng that sounds like a good strategy! But one worry: until we flip that lint rule on, other code could still be written elsewhere that accesses e.g. a `TokenMap` without checking undefined, potentially causing another bug/crash. Are we confident enough in PR review + the `noUncheckedIndexedAccess` timeline to take undefined out of the type definitions in this PR? cc @zzmp

PR #6296, opened 2023-04-03
fix: wallet connection analytics

zzmp:
## Description

Fixes wallet connection analytics, which were not being sent for eager connections. The analytics were being sent as part of the wallet connection modal, which was not used for eager connections.

- Moves the "active wallet change detection" used to trigger the event to Web3Provider
- Removes extra analytics from WalletConnect.ts
- https://uniswaplabs.atlassian.net/browse/WEB-2998

#### Manual
- [x] Induced an eager connection and observed the "Wallet Connect Transaction Completed" event in the _Network_ pane
- [x] Connected normally and observed the "Wallet Connect Transaction Completed" event in the _Network_ pane

#### Automated
- [x] Unit test


craco.config.cjs
```diff
@@ -24,6 +24,7 @@ module.exports = {
         moduleNameMapper: {
           '@uniswap/conedison/format': '@uniswap/conedison/dist/format',
           '@uniswap/conedison/provider': '@uniswap/conedison/dist/provider',
+          'node:events': 'eventemitter3',
```
grabbou, line 27: Why is this necessary? What kind of issue were you running into?

cartcrom (approved):

src/components/Web3Provider/index.tsx
```diff
@@ -40,6 +47,41 @@ function Tracer() {
     }
   }, [networkProvider, provider, shouldTrace])
 
+  // Send analytics events when the active account changes.
+  const activeAccount = useRef<string | undefined>()
+  const getConnection = useGetConnection()
+  const [connectedWallets, addConnectedWallet] = useConnectedWallets()
+  useEffect(() => {
```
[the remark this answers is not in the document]
zzmp: I think that's fine, as long as the code within it stays segmented. wdyt?

zzmp:
This fails tests because it modifies Web3Provider, and has effects during startup.

It _should_ probably be implemented at that level, and Web3Provider _should_ probably be mocked in test-utils/render, so I think the appropriate "fix" would be to replace Web3Provider in test-utils/render with a mock. This allows us to test _this_ behavior, and has the fortunate side effect of skipping web3-react logic for most unit tests - which is desired and a known bug of our unit testing.

src/components/Web3Provider/index.tsx
```diff
@@ -40,6 +47,41 @@ function Tracer() {
     }
   }, [networkProvider, provider, shouldTrace])
 
+  // Send analytics events when the active account changes.
+  const activeAccount = useRef<string | undefined>()
```
cartcrom, line 51: is this a use case for `usePrevious`?

cartcrom (approved):

src/components/Web3Provider/index.tsx
```diff
@@ -16,17 +22,19 @@ export default function Web3Provider({ children }: { children: ReactNode }) {
...
+      <Updater />
       {children}
     </Web3ReactProvider>
   )
 }
 
-function Tracer() {
-  const { chainId, provider } = useWeb3React()
+/** A component to run hooks under the Web3ReactProvider context. */
+function Updater() {
```
cmcewen, line 32: nit: easier to read to me if components are defined before they're used. also maybe `ConnectionUpdater` is more clear? looks like it is basically just used for debug/analytics of the connection

cmcewen (commented):
i'd just like to double check that this event name is correct before merge

src/components/Web3Provider/index.tsx
zzmp: this deserves to be a lint rule, because my general preference is the opposite.

src/components/Web3Provider/index.tsx
```diff
@@ -40,6 +48,40 @@ function Tracer() {
...
+      // so that the event contains the correct and up-to-date user properties.
+      user.set(CustomUserProperties.WALLET_ADDRESS, account)
+      user.set(CustomUserProperties.WALLET_TYPE, walletType)
+      user.set(CustomUserProperties.PEER_WALLET_AGENT, peerWalletAgent ?? '')
+      if (chainId) {
+        user.postInsert(CustomUserProperties.ALL_WALLET_CHAIN_IDS, chainId)
+      }
+      user.postInsert(CustomUserProperties.ALL_WALLET_ADDRESSES_CONNECTED, account)
+
+      sendAnalyticsEvent(InterfaceEventName.WALLET_CONNECT_TXN_COMPLETED, {
```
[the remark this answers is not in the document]
zzmp: This implies that a WalletConnect "connecting" tx has completed. It's been taken out of the context of the other WC events, so it makes less sense here, but I don't think we should change it in this PR because this is meant only to be a fix for previously broken behavior.

src/state/wallets/reducer.ts
```diff
@@ -18,13 +18,8 @@ const walletsSlice = createSlice({
   initialState,
   reducers: {
     addConnectedWallet(state, { payload }) {
-      const existsAlready = state.connectedWallets.find((wallet) => shallowEqual(payload, wallet))
-      if (!existsAlready) {
-        state.connectedWallets = state.connectedWallets.concat(payload)
-      }
-    },
-    removeConnectedWallet(state, { payload }) {
```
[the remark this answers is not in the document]
zzmp: It was never used - correct. This PR touches the src/state/wallets code path, so that code path has been cleaned up.

cmcewen (approved):

PR #6299, opened 2023-04-04
test: web3

zzmp:
## Description

Adds unit test for rendering Web3Provider

#### Manual
N/A

#### Automated
- [X] Unit test
- [ ] Integration/E2E test


src/components/Web3Provider/index.test.tsx
```diff
@@ -0,0 +1,51 @@
+import { act, render } from '@testing-library/react'
+import { initializeConnector } from '@web3-react/core'
+import { EIP1193 } from '@web3-react/eip1193'
+// TODO(INFRA-163): Export MockEIP1193Provider from @web3-react/eip1193 to facilitate testing.
+// eslint-disable-next-line @typescript-eslint/no-restricted-imports
+import { MockEIP1193Provider } from '@web3-react/eip1193/dist/mock'
```
zzmp, line 6: This should be fixed in web3-react before this PR lands

src/test-utils.tsx
```diff
@@ -50,6 +51,12 @@ const customRender = (ui: ReactElement) => render(ui, { wrapper: WithProviders }
 const customRenderHook = <Result, Props>(hook: (initialProps: Props) => Result) =>
   renderHook(hook, { wrapper: WithProviders })
 
+export function asMock<T extends (...args: any) => any>(fn: T) {
+  const mock = fn as unknown as jest.Mock<ReturnType<T>, Parameters<T>>
+  assert(mock.mock, 'fn is not a mock')
+  return mock
+}
+
```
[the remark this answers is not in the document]
zzmp: yes 100% this should have a jsdoc

yarn.lock
```diff
@@ -5404,13 +5404,32 @@
   optionalDependencies:
     "@ethersproject/providers" "^5"
 
+"@web3-react/core@8.1.3-beta.0":
```
zzmp, line 5407: These packages are duplicated because the widget depends on pinned versions.

cartcrom (approved):

src/components/Web3Provider/index.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+      connector,
+      hooks,
+      getName: jest.fn().mockReturnValue('test'),
+      type: 'INJECTED' as ConnectionType,
+      shouldDisplay: () => false,
+    }
+    asMock(useOrderedConnections).mockReturnValue([connection])
+  })
+
+  it('renders', async () => {
```
cmcewen, line 40: idk about test name convention but this looks like to me like this is testing that the `Web3Provider` `renders and eagerly connects`. and i feel like for something like this without a UI impact, we do actually want to test more for this connecting behavior

src/test-utils.tsx
```diff
@@ -50,6 +51,23 @@ const customRender = (ui: ReactElement) => render(ui, { wrapper: WithProviders }
...
+ * Use this in combination with jest.mock() to safely access function from mocked modules.
+ *
+ * @example
+ * import { useExample } from 'example'
+ * jest.mock('example', () => ({ useExample: jest.fn() }))
+ * beforeEach(() => {
+ *   asMock(useExample).mockImplementation(() => ...)
+ * })
+ */
+export function asMock<T extends (...args: any) => any>(fn: T) {
```
cmcewen, line 65: are you sure this is the only way to do this? i am surprised that if this is such a common use case, jest wouldn't include an easier to way to do this

cmcewen (commented):
i could be okay merging this but i am not super excited about duplicate packages and also that asMock util, so would like a look at each before we decide to merge

src/test-utils.tsx
zzmp: I have not found anything else, and I've looked for it many times (because it is a common use case). I _can_ improve it though, using https://jestjs.io/docs/jest-object#jestismockfunctionfn.

zzmp: This has been broken into its own PR: https://github.com/Uniswap/interface/pull/6310.

craco.config.cjs
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable @typescript-eslint/no-var-requires */
+/* eslint-env node */
```
zzmp, line 1: This has been broken out into its own PR: https://github.com/Uniswap/interface/pull/6311

src/components/Web3Provider/index.test.tsx
zzmp: This is just a simple test for `render`, it is _not_ a test for `useEagerlyConnect`. That would need to be tested separately, as it is specific to the connectors used in interface and will not function with the simple mocked connection used for the render test here.

This is just to lay a foundation of testing for this file. It is not meant to be exhaustive.

cmcewen: i get mocking something if it needs to be for it to be rendered, but the fact that you have `expect(useEagerlyConnect).toHaveBeenCalled()` means that you expect this component to call the hook, which means you're also testing more than just rendering. so technically i'd say you should either leave out that expect to have been called from this test, add the `and calls eagerly connect` to the description for this test, or have a separate test for render and this hook being called.

cmcewen (approved):

cmcewen (approved):

PR #6300, opened 2023-04-04
style: adding erc1155 batch buy design

JackShort:
Implements the design for erc1155 batch buying - keep in mind this is not adding the correct behavior just the design

* this is being pushed to a feature branch btw

Screen Shot 2023-04-04 at 1 56 39 PM

### DIFF
--- a/src/nft/components/card/containers.tsx
+++ b/src/nft/components/card/containers.tsx
@@ -31,7 +31,7 @@ const StyledDetailsContainer = styled(Column)`
   }
 `
 
-const StyledActionContainer = styled.div`
+const ActionContainer = styled.div`
   position: absolute;
   bottom: -32px;
   left: 8px;
@@ -46,10 +46,6 @@ const StyledActionContainer = styled.div`
     bottom: 8px;
   }
 `
-
-const ActionContainer = ({ children }: { children: ReactNode }) => (
-  <StyledActionContainer>{children}</StyledActionContainer>
-)
 
 const StyledActionButton = styled(ThemedText.BodySmall)<{
   selected: boolean
@@ -110,15 +106,21 @@ const ActionButton = ({
   )
 }
 
-const StyledErc1155ActionButton = styled.div<{ $disabled?: boolean }>`
+const StyledErc1155ActionButton = styled.button`
   position: relative;
   display: flex;
   width: 100%;
   height: 100%;
-  background: ${({ theme, $disabled }) => ($disabled ? theme.accentActionSoft : theme.accentAction)};
+  background: ${({ theme }) => theme.accentAction};
   padding: 4px 0px;
   justify-content: center;
-  cursor: ${({ $disabled }) => ($disabled ? 'default' : 'pointer')};
+  border: none;
+  cursor: pointer;
+
+  &:disabled {
+    cursor: default;
+    background: ${({ theme }) => theme.accentActionSoft};
+  }
 
   &:before {
     background-size: 100%;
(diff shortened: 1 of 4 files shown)


cgkol (commented):
High level (admittedly have not looked at the code at all), shall we add some tests?

JackShort:
> High level (admittedly have not looked at the code at all), shall we add some tests?

I think the tests should be added at the end when the whole feature is implemented

cgkol:
> > High level (admittedly have not looked at the code at all), shall we add some tests?
> 
> I think the tests should be added at the end when the whole feature is implemented - because a few things aren't accounted for here like loading the data etc.

Wait is this not finalized? Or is the next PR going to build on top of this one?

cgkol:
btw, make sure you tag me so i get the notif 😂 @JackShort

src/nft/components/card/containers.tsx
```diff
@@ -102,6 +110,91 @@ const ActionButton = ({
   )
 }
 
+const StyledErc1155ActionButton = styled.div<{ $disabled?: boolean }>`
```
tinaszheng, line 113: this is a `button` so this should just use `styled.button` 🤔 you'll get all the accessibility features associated with html buttons, and you'll also be able to just use the default `disabled` prop combined with 

```
styled.button`
   background: theme.accentAction 

   &:disabled {
      background: theme.accentActionSoft
   }
`
```

src/nft/components/card/icons.tsx
```diff
@@ -37,17 +37,29 @@ const ListPriceRowContainer = styled(Row)`
...
   hidePrice,
 }: {
   isSelected: boolean
+  quantitySelected?: number
   marketplace?: Markets
   tokenType?: NftStandard
   listedPrice?: string
   hidePrice?: boolean
 }) => {
+  if (quantitySelected > 0) {
```
tinaszheng, line 53: prefer guard-style returns, like:

```
if (quantitySelected === 0) return 

return (
   <StyledMarketplaceContainer> ...
)

File: src/nft/components/card/containers.tsx [hunk 3]
```diff
@@ -102,6 +110,91 @@ const ActionButton = ({
...
+  }
+`
+
+const StyledErc1155ActionContainer = styled(ThemedText.BodySmall)`
+  display: flex;
+  background: ${({ theme }) => theme.backgroundModule};
+  color: ${({ theme }) => theme.backgroundSurface};
+  border-radius: 8px;
+  justify-content: center;
+  font-weight: 600 !important;
```
tinaszheng, line 151: why do you need the !important here

src/nft/components/card/containers.tsx
```diff
@@ -30,34 +31,41 @@ const StyledDetailsContainer = styled(Column)`
...
+  will-change: transform;
+  visibility: hidden;
+
+  @media screen and (max-width: ${BREAKPOINTS.sm}px) {
+    visibility: visible;
+    bottom: 8px;
+  }
+`
+
+const ActionContainer = ({ children }: { children: ReactNode }) => (
```
tinaszheng, line 50: confused what this is doing - isn't this the same thing as just `StyledActionContainer` by itself?

src/nft/components/card/icons.tsx
JackShort: this really isn't applicable we only care here if quantity selected is passed if so we render it - if not it goes to the other conditional checks below it

PR #6301, opened 2023-04-04
chore: removing unnecessary data in assets query

JackShort:
Reducing initial response size by reducing the requested amount of data for the assets query + making mini portfolio only load nft balance when opened

### DIFF
--- a/src/graphql/data/nft/Asset.ts
+++ b/src/graphql/data/nft/Asset.ts
@@ -39,36 +39,18 @@ gql`
         node {
           id
           name
-          ownerAddress
           image {
             url
           }
           smallImage {
             url
           }
-          originalImage {
-            url
-          }
           tokenId
-          description
           animationUrl
           suspiciousFlag
           collection {
-            name
-            isVerified
-            image {
-              url
-            }
-            creator {
-              address
-              profileImage {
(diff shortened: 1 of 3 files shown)


src/graphql/data/nft/NftBalance.ts
```diff
@@ -122,7 +122,8 @@ export function useNftBalance(
   first?: number,
   after?: string,
   last?: number,
-  before?: string
+  before?: string,
+  disabled = false
```
cartcrom, line 126: nit: rather than `disabled` just keep the name `skip` for simplicity

cartcrom (approved):

tinaszheng (approved):

src/graphql/data/nft/Asset.ts
```diff
@@ -99,10 +84,7 @@ gql`
           }
           rarities {
             provider
-            rank
```
cbachmeier, line 102: Rarity no longer appearing on cards 
Screen Shot 2023-04-06 at 10 01 08

cbachmeier (changes requested):

cbachmeier (approved):
LGTM

PR #6303, opened 2023-04-04
fix: use backend to resolve NFT image for ENS avatar

just-toby:
## Description

Use our GraphQL API for fetching the NFT image URL for an ENS Avatar

https://uniswaplabs.atlassian.net/browse/WEB-3104?atlOrigin=eyJpIjoiZGZmZmIwMDQ4ZTdiNDJiYjgyYjBjZWZlODFjYjc5NGMiLCJwIjoiaiJ9

## Screen Capture

NFT metadata working fine:

| Before  | After        |
|  image |  image |

NFT metadata wasn't working:
| Before  | After   (when comparing owner to onchain owner     |
| [bug report](https://uniswaplabs.atlassian.net/browse/WEB-3104?atlOrigin=eyJpIjoiZTRjNzE2YzM2NTZmNDAxZDliMGRmNTJjZTUyMjRmODIiLCJwIjoiaiJ9)   |  image |

#### Manual

tested manually with an address that has an ENS avatar set to an owned NFT - it works the same for avatars that worked before, and it now works for avatars which were hitting cors issues on the previous implementation

#### Automated
- [ ] Unit test
- [x] Integration/E2E test

### DIFF
--- a/cypress/e2e/wallet-dropdown.test.ts
+++ b/cypress/e2e/wallet-dropdown.test.ts
@@ -77,4 +77,11 @@ describe('Wallet Dropdown', () => {
     cy.get(getTestSelector('theme-auto')).click()
     cy.get(getTestSelector('wallet-header')).should('have.css', 'color', 'rgb(119, 128, 160)')
   })
+
+  it('should render the ENS name and avatar', () => {
+    visit(true)
+    cy.get(getTestSelector('web3-status-connected')).click()
+    cy.contains('averylongtestaddress.eth').should('exist')
+    cy.get(getTestSelector('identicon-styled-avatar')).should('be.visible')
+  })
 })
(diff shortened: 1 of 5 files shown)


just-toby:
converting to draft because i've learned that we can't rely on backend ownership data

tinaszheng:
why do we want this 🤔

tinaszheng:
@just-toby i looked into the issue more and it's because the ENS avatar is resolving to a folia.app URL and their CORS policy is preventing us from fetching the token url, which is unfortunately not our fault

cal is going to reach out to folia to see if they can fix on their end! i'd say the bug is not actionable for now 🤔 im not sure moving to gql for ens lookups is the right solution rn

just-toby:
closing for now since we discussed offline and agreed that this needs to wait until our backend can accept more traffic.

In the absence of the backend limitation, I thought this was a reasonable change because we should have a unified rule on whether we can render an NFT image or not - if we can render the NFT image on the NFT details page, it's strange to just throw up our hands and say we can't render it as an ENS avatar because of a cors issue.

PR #6306, opened 2023-04-05
feat: remove amplitude swap error logging

just-toby:
## Description
removes the Amplitude event for "Swap Error" from the main page of the app. this was intended to be used to compare relative error counts between versions of the app with the widget and without. Since we're no longer using the widget on this page, we can remove this logging.

original PR: https://github.com/Uniswap/interface/pull/6008

### QA (ie manual testing)

- [x] triggered an error by adding a manual throw in the code, and ensured that this callback ran but no amplitude event was logged

### Automated testing

- [ ] <s>Unit test</s>


zzmp (approved):
lgtm - please make sure you clean up SWAP_ERROR from @uniswap/analytics-events as well.

Imebeez (commented):
👍

PR #6308, opened 2023-04-06
fix: do not double-report i18n exception

zzmp:
## Description

Removes the duplicate error reporting from a locale failing to load. This is already reported as a ChunkLoadError through the ErrorBoundary or the global onerror handler.

### QA (ie manual testing)

- [X] N/A

### Automated testing

- [ ] <s>Unit test</s>

### DIFF
--- a/src/lib/i18n.tsx
+++ b/src/lib/i18n.tsx
@@ -1,6 +1,5 @@
 import { i18n } from '@lingui/core'
 import { I18nProvider } from '@lingui/react'
-import * as Sentry from '@sentry/react'
 import { DEFAULT_LOCALE, SupportedLocale } from 'constants/locales'
 import {
   af,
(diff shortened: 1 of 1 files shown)


src/lib/i18n.tsx
```diff
@@ -83,9 +82,8 @@ export async function dynamicActivate(locale: SupportedLocale) {
     const catalog = await import(`locales/${locale}.js`)
     // Bundlers will either export it as default or as a named export named default.
     i18n.load(locale, catalog.messages || catalog.default.messages)
-  } catch (error) {
-    console.error(error)
-    Sentry.captureException(new Error(`Unable to load locale (${locale})`))
+  } catch (error: unknown) {
+    console.error(new Error(`Unable to load locale (${locale}): ${error}`))
```
tinaszheng, line 86: when should you just `console.error` and when should you do `Sentry.captureException`?

tinaszheng (approved):

src/lib/i18n.tsx
zzmp: In this case, `Sentry.captureException` is duplicating the exception, because it's a symptom of a different issue. Logging adds it to the breadcrumbs in Sentry in case another error is reported.

PR #6310, opened 2023-04-06
test: add jest.asMock

zzmp:
## Description

Adds a `mocked` utility for more ergonomic testing. This mirrors the [`mocked` function](https://kulshekhar.github.io/ts-jest/docs/27.0/guides/test-helpers/#mockedtitem-t-deep--false) exported by `jest-ts`.
I'd prefer this be available as `jest.mocked`, but I was unable to add a property to [the `jest` object](https://jestjs.io/docs/jest-object).
In order to isolate this from the render hooks, this change also refactors `test-utils` into its own directory. This prevents `import { mocked } from 'test-utils'` from bringing in un-mocked modules with side effects, because the render hooks set up contexts and bring in lots of dependencies.

### QA (ie manual testing)

- [X] N/A

### Automated testing

- [X] Unit test


src/tracing/trace.test.ts
```diff
@@ -11,10 +12,9 @@ jest.mock('@sentry/react', () => {
     startTransaction: jest.fn(),
   }
 })
-const startTransaction = Sentry.startTransaction as jest.Mock
 
 function getTransaction(index = 0): Transaction {
-  const transactions = startTransaction.mock.results.map(({ value }) => value)
+  const transactions = mocked(Sentry.startTransaction).mock.results.map(({ value }) => value)
```
cmcewen, line 17: it is a little strange to me that this function has side effect on the mock that requires a reset in between tests. like could it break tests if you call it twice in one test? but don't need to fix it as part of this PR

cmcewen (approved):

lynnshaoyu (changes requested):
pls fix failing unit tests

src/test-utils/render.tsx
```diff
@@ -13,8 +14,6 @@ import { HashRouter } from 'react-router-dom'
 import store from 'state'
 import ThemeProvider from 'theme'
 
-import catalog from './locales/en-US'
```
lynnshaoyu, line 16: why did this change?

zzmp: It was just moved to an absolute path for consistentcy.

PR #6312, opened 2023-04-06
feat: Use local reducer in Swap flow

tinaszheng:
## Description
Keep the swap state in a local reducer in the Swap flow instead of in one global state to enable multiple Swap components

- https://www.notion.so/uniswaplabs/Swap-widget-replacement-work-bd91b44494d547259b4f29f4c36b3dfc

### QA (ie manual testing)

- [x] manually go through the swap flow to ensure the inputs still work
- [x] navigate between pages to ensure the behavior is as expected

### Automated testing

- [ ] Unit test
- [x] Integration/E2E test - existing e2e tests pass, and added a new test case for navigating between pages


src/state/transactions/updater.tsx
```diff
@@ -81,17 +73,6 @@ export default function Updater() {
 
       const tx = transactions[chainId]?.[hash]
 
-      if (tx.info.type === TransactionType.SWAP && trade) {
```
tinaszheng, line 84: we'll have to fix how logging for analytics works for completed swaps - most likely by just adding in a serializable version of `trade` details to the `transactions` store

just-toby (approved):

tinaszheng:
oh nice, @just-toby we could also get this in before the "replacement" PR

zzmp (approved):

just-toby (approved):

zzmp (approved):

lynnshaoyu (approved):

PR #6313, opened 2023-04-06
refactor: rename WalletDropdown to AccountDrawer

just-toby:
## Description

Renames WalletDropdown to AccountDrawer. 

we don't even have a wallet dropdown anymore - it's more accurate to call this a PortfolioDrawer or just Portfolio for related code that is used in this view.

## Screen capture

| Before                  | After (Desktop)        | After (Mobile)         |
| ----------------------- |----------------------- | ---------------------- |
| image | image | image |

### QA (ie manual testing)

- [x] manually verified that opening the MP still works

### Automated testing

- <s>[ ] Unit test</s>
- <s>[ ] Integration/E2E test</s>


tinaszheng (approved):
hell yeah

cartcrom (changes requested):
I agree with updating the name (I've also been meaning to similarly rename WalletModal to WalletOptions since it's no longer a modal) but I'd highly prefer "WalletDrawer" over "Portfolio". 

Calling it Portfolio is potentially confusing because it's a drawer that contains multiple things, not all portfolio related (settings, account info, etc). Code paths like `/Portfolio/MiniPortfolio` and `/Portfolio/Settings` make less sense to me than something like `WalletDrawer`

zzmp:
> I agree with updating the name (I've also been meaning to similarly rename WalletModal to WalletOptions since it's no longer a modal) but I'd highly prefer "WalletDrawer" over "Portfolio".
> 
> Calling it Portfolio is potentially confusing because it's a drawer that contains multiple things, not all portfolio related (settings, account info, etc). Code paths like `/Portfolio/MiniPortfolio` and `/Portfolio/Settings` make less sense to me than something like `WalletDrawer`

It's not just `WalletDrawer` or `Portfolio`, because it also has `Settings` in it. Really, it's just a `RightHandMenu` at this point. If anything semantic, maybe a `ConnectMenu` or a `ConnectModal` (or `Connection*`) because the `Connect` button triggers it, but that could change in the future. Other names: `SlideOutMenu`, `FlyOutMenu`, `ControlPanel`, `StartMenu` (hello Win98), etc. (I'm personally fine with any name that's easy for the team to track, I'm just playing devil's advocate and throwing out add'l ideas in case they're useful. Maybe @cartcrom and @just-toby should huddle and just decide on one?)

cartcrom (approved):

tinaszheng (approved):
trying again

PR #6318, opened 2023-04-07
ci: Emit to slack on merge to releases/* branch

lavalamp-:
## Description

This PR introduces a CI action that triggers on merge to any of the `releases/*` branches. When code is merged to any of the branches in question a message will be sent to Slack containing relevant information as to...

- Who merged the code?
- What was in the code?
- Was the merge a force push?

This in turn provides the UL engineering team with visibility into code changes for their CI pipeline.

### QA (ie manual testing)

This GH action definition was first committed to the PR with a `pull_request` target definition, meaning that it runs every time new code is committed to this PR.

I then confirmed that a test message was correctly sent to the expected Slack channel:

image

The PR was then updated to have the correct information in the Slack message (which cannot be fully tested as the contents of the GH action context change depending on what the target type is).

Note that this updated Slack message content was copied from another UL repository where the formatting has already been tested and confirmed to work (please see #mobile-build-alerts channel for reference).

### DIFF
--- a/.github/workflows/notify-slack-on-release-merge.yml
+++ b/.github/workflows/notify-slack-on-release-merge.yml
@@ -1,0 +1,41 @@
+name: Slack notifications for releases/* merges
+
+# This CI job will push notifications to Slack whenever code is merged into any of the
+# release branches.
+
+on:
+  push:
+    branches:
+      - 'releases/*'
+  pull_request:
+    branches:
+      - '**'
+
+jobs:
+  notify-slack:
+    name: 'Emit Slack notification(s)'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Send custom JSON data to Slack workflow
+        id: slack
+        uses: slackapi/slack-github-action@007b2c3c751a190b6f0f040e47ed024deaa72844
+        with:
+          payload: |
+            {
+              "text": "This is a really cool test!",
+              "blocks": [
+                {
+                    "type": "divider"
+                },
+                {
+                  "type": "section",
+                  "text": {
+                    "type": "mrkdwn",
+                    "text": "Cool tests can be cool"
+                  }
+                }
+              ]
+            }
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+          SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
--- a/.github/workflows/notify-slack-on-release-merge.yml
+++ b/.github/workflows/notify-slack-on-release-merge.yml
@@ -12,7 +12,7 @@ jobs:
     name: 'Emit Slack notification(s)'
     runs-on: ubuntu-latest
     steps:
-      - name: Parse event to Slack messages
+      - name: Parse event to slug
         env:
           GITHUB_CONTEXT: ${{ toJson(github) }}
         # This is a bit of shell kung fu that formats the contents of the GitHub event into a multi-line Slack message
--- a/.github/workflows/notify-slack-on-release-merge.yml
+++ b/.github/workflows/notify-slack-on-release-merge.yml
@@ -20,9 +20,6 @@ on:
   push:
     branches:
       - 'releases/*'
-  pull_request:
-    branches:
-      - '**'
 
 jobs:
   notify-slack:
--- a/.github/workflows/notify-slack-on-release-merge.yml
+++ b/.github/workflows/notify-slack-on-release-merge.yml
@@ -20,9 +20,6 @@ on:
   push:
     branches:
       - 'releases/*'
-  pull_request:
-    branches:
-      - '**'
 
 jobs:
   notify-slack:
@@ -30,7 +27,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Parse event to slug
-        id: parse-slug
         env:
           GITHUB_CONTEXT: ${{ toJson(github) }}
         # Formats the contents of the GitHub event into slugs: one line per commit, formatted for Slack.
@@ -42,7 +38,7 @@ jobs:
           | awk -F';' '{print "• <"$1"|"$2"> (<https://github.com/"$3"|"$3">, "$4") - "$5}' \
           | sed 's/$/\\n/g' | tr -d '\n' \
           > /tmp/parsed_github_context
-          echo "SLACK_COMMITS=$(cat /tmp/parsed_github_context)" >> $GITHUB_OUTPUT
+          echo "SLACK_COMMITS=$(cat /tmp/parsed_github_context)" >> $GITHUB_ENV
       - name: Send custom JSON data to Slack workflow
         id: slack
         uses: slackapi/slack-github-action@007b2c3c751a190b6f0f040e47ed024deaa72844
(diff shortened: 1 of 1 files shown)


.github/workflows/notify-slack-on-release-merge.yml
```diff
@@ -0,0 +1,68 @@
+name: Slack notifications for releases/* merges
+
+# This CI job will push notifications to Slack whenever code is merged into any of the
+# release branches.
```
zzmp, line 4:
```suggestion
# This CI job will push notifications to Slack whenever code is merged into any releases/* branch.
```

.github/workflows/notify-slack-on-release-merge.yml
```diff
@@ -0,0 +1,68 @@
...
+jobs:
+  notify-slack:
+    name: 'Emit Slack notification(s)'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Parse event to Slack messages
+        env:
+          GITHUB_CONTEXT: ${{ toJson(github) }}
+        run: |
+          echo $GITHUB_CONTEXT | jq '.event.commits[] | [.url, .id[0:7], .author.username, .timestamp, .message] | @tsv' | sed 's/"//g' | sed 's/\\t/;/g' | sed 's/\\n/;/g' | sed 's/\\//g' | awk -F';' '{print "• <"$1"|"$2"> (<https://github.com/"$3"|"$3">, "$4") - "$5}' | sed 's/$/\\n/g' | tr -d '\n' > /tmp/foo
```
zzmp, line 20: Why do this inline? Would it be more readable if you called out to a script?
If you leave this inline, can you break it into a multi-line statement with comments?

.github/workflows/notify-slack-on-release-merge.yml
```diff
@@ -0,0 +1,68 @@
...
+          echo "SLACK_COMMITS=$(cat /tmp/foo)" >> $GITHUB_ENV
```
zzmp, line 21: Instead of /tmp/foo, what about /tmp/parsed_github_context?

.github/workflows/notify-slack-on-release-merge.yml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Send custom JSON data to Slack workflow
```
zzmp, line 22: Does this use the parsed event? Why is it being parsed?

zzmp (changes requested):

.github/workflows/notify-slack-on-release-merge.yml
lavalamp-: Sounds good - changed!

.github/workflows/notify-slack-on-release-merge.yml
lavalamp-: Yes! This takes the content of the [`$GITHUB_CONTEXT`](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) JSON blob and formats it into the proper format for the Slack message (check out the contents of the #mobile-build-alerts Slack channel to see what the output looks like).

.github/workflows/notify-slack-on-release-merge.yml
lavalamp-: I don't recall the context but inline was one of the only ways I could get this to work when I was iterating on this for the other repository where it originated. It could be that a shell script would work as well _but_ in the interest of getting the CI pipeline working my preference would be to add comments here explaining the commands rather than rework anything.

.github/workflows/notify-slack-on-release-merge.yml
lavalamp-: Done

zzmp (approved):
Looks great!

I've just added some suggestions to make it easier to maintain over time (formatting/comments), so that future readers can more easily tell what's being parsed out.

.github/workflows/notify-slack-on-release-merge.yml
```diff
@@ -0,0 +1,75 @@
...
+        # The steps are as follows:
+        # 1. Pull commit data from the Github context and format it as a TSV with jq
+        # 2. Clean up the TSV output so that it's more easily parsed by awk
+        # 3. Use awk to format the TSV into a Slack message
+        # 4. Use sed and tr to deal with new-line escaping issues
+        # 5. Write the output to a file
+        # 6. Use Github action syntax to load the contents of the file into an environment variable
+        run: |
+          echo $GITHUB_CONTEXT | jq '.event.commits[] | [.url, .id[0:7], .author.username, .timestamp, .message] | @tsv' | sed 's/"//g' | sed 's/\\t/;/g' | sed 's/\\n/;/g' | sed 's/\\//g' | awk -F';' '{print "• <"$1"|"$2"> (<https://github.com/"$3"|"$3">, "$4") - "$5}' | sed 's/$/\\n/g' | tr -d '\n' > /tmp/parsed_github_context
+          echo "SLACK_COMMITS=$(cat /tmp/parsed_github_context)" >> $GITHUB_ENV
```
zzmp, line 28: Why is this done using an environment variable instead of an output parameter?

eg you could `echo "SLACK_COMMITS=$(cat /tmp/parsed_github_context)" >> $GITHUB_OUTPUT` and then access it as `"${{ steps.parse-slug.outputs.SLACK_COMMITS }}"` so that it is more apparent where it is comming from. I think I'd prefer this method, as it uses localized variables so it seems easier to debug.

grabbou (approved):
I second Zach's review here and would like to see some of the comments applied before final merge.

lavalamp-:
Ok! I think I've got everything addressed.

I couldn't get the comments interspersed with the commands (as it's a multi-line shell command) but I went ahead and copied the steps into a comment at the start of the document and interspersed explanations there. Not the prettiest but I think it gives us what we want??

.github/workflows/notify-slack-on-release-merge.yml
```diff
@@ -0,0 +1,89 @@
...
+# Then we parse out the specific fields we want for our messages using jq and format it into tab-separated values
+#  | jq '.event.commits[] | [.url, .id[0:7], .author.username, .timestamp, .message] | @tsv' \
+# We need to do some cleaning on this output - specifically removing quotes and replacing newlines with something easier to split
+#  | sed 's/"//g' | sed 's/\\t/;/g' | sed 's/\\n/;/g' | sed 's/\\//g' \
+# We then use awk to format the TSV into a Slack message
+#  | awk -F';' '{print "• <"$1"|"$2"> (<https://github.com/"$3"|"$3">, "$4") - "$5}' \
+# Finally, we need to deal with some escaping issues with newlines so that we don't break the Slack message format
+#  | sed 's/$/\\n/g' | tr -d '\n' \
+# Then shove the bytes into a file to store them in their exact format
+#  > /tmp/parsed_github_context
```
zzmp, line 17: Move this comment into the step, just above run.

zzmp (approved):

.github/workflows/notify-slack-on-release-merge.yml
lavalamp-: Will do in a follow-on nit PR.

PR #6320, opened 2023-04-07
fix: pwat not setting correct spender

JackShort:
We ran into a bug on the NFT side where the backend upgraded the UR contract they are using and the frontend is still behind. This made it so the spender of the permit signature was the old UR contract but the one that was executing the transaction is the new one. Hardcoding the nft side ur contract manually and adding a test to make sure when the nft side is requesting the spender it is set to the correct one


src/nft/hooks/usePermit2Approval.ts
```diff
@@ -7,16 +7,23 @@ import usePermit2Allowance, { AllowanceState } from 'hooks/usePermit2Allowance'
 import { useCallback, useMemo, useState } from 'react'
 import invariant from 'tiny-invariant'
 
+const NFT_UNIVERSAL_ROUTER_MAINNET_ADDRESS = '0x4c60051384bd2d3c01bfc845cf5f4b44bcbe9de5'
```
cbachmeier, line 10: If we updated the UR sdk to the latest version would it have the correct address?

JackShort: discussed in #eng-web we decided this was the best solution

tinaszheng: can you add a TODO to update the the sdk and import address from there?

src/nft/hooks/usePermit2Approval.ts
```diff
@@ -7,16 +7,24 @@ import usePermit2Allowance, { AllowanceState } from 'hooks/usePermit2Allowance'
...
+  useNftRouter?: boolean
 ) {
   const { chainId } = useWeb3React()
 
   const allowance = usePermit2Allowance(
     enabled ? maximumAmount ?? (amount?.currency.isToken ? (amount as CurrencyAmount<Token>) : undefined) : undefined,
-    enabled && chainId ? UNIVERSAL_ROUTER_ADDRESS(chainId) : undefined
+    enabled && chainId
+      ? useNftRouter
+        ? NFT_UNIVERSAL_ROUTER_MAINNET_ADDRESS
```
tinaszheng, line 25: wondering how this would work when you're on a different chain rather than mainnet.. or is the NFT flow only on mainnet rn?

JackShort: We only offer NFTs on mainnet

tinaszheng: i would recommend handling the non mainnet + useNftRouter case in some way to prevent users from being able to call this flow if they're not on mainnet, few suggestions:

1) only pass in `useNftRouter` if user is on mainnet
2) throw Error in `usePermit2Approval` if user gets in a state with non mainnet + useNftRouter
3) default to looking up `UNIVERSAL_ROUTER_ADDRESSES` if not mainnet
4) pass in `undefined` spender if `useNftRouter` and not mainnet

JackShort: For context this cannot be called if the user is not on mainnet, the actual update approval can only be called by clicking the button which would default to switching chains over updating approval

cbachmeier (approved):
LGTM

cmcewen (approved):
given this is broken in prod, i'm okay to merge now. however i find the enabled flag in `usePermit2Approval` and lots of ternaries hard to read and not very ergonomic so i would appreciate the next PR to update the address taking another pass at that codepath

JackShort:
> given this is broken in prod, i'm okay to merge now. however i find the enabled flag in `usePermit2Approval` and lots of ternaries hard to read and not very ergonomic so i would appreciate the next PR to update the address taking another pass at that codepath

@cmcewen fortunately this might happen rather soon I think

cbachmeier (approved):

PR #6321, opened 2023-04-07
fix: update CurrencyList unit tests

just-toby:
## Description
Changes the test for this component from using a UI snapshot to checking for individual child elements. this is a workaround to an issue where the snapshot test was behaving differently in GH actions vs locally. and it improves the test because the functionality we're checking is written in the test.

also, removes an inline style from the component under test. css/class based styling is more performant than inline style objects.

from the react docs:

> We recommend using style only for dynamic styles where you don’t know the style values ahead of time. In other cases, applying plain CSS classes with className is more efficient. [Read more about className and style.](https://react.dev/reference/react-dom/components/common#applying-css-styles)

### Automated testing

- [x] Unit test
- <s>[ ] Integration/E2E test</s>


zzmp (approved):
This begs the question: in which situations are snapshot tests appropriate?

just-toby:
@zzmp they're appropriate when you want to verify details about a component's layout AND when they actually work. Until we figure out the root cause of the classname mismatch, it's not an option here.

both types of tests have their benefits - the new Jest version makes it easier to tell what's actually being tested. But the snapshots were easy to update and actually verified more details about the layouts - styling and hierarchy shape etc.

imo, without actual visual screenshots showing diffs between old/new snapshots, snapshot tests are less useful than they could be and act as more of a shortcut for an engineer not to write test code. not to say they shouldn't be an option.

lynnshaoyu (approved):

PR #6323, opened 2023-04-07
build: optimize build by splitting out typechecking and linting

zzmp:
## Description

Reduces build time for production builds by half (~6m in the GitHub Action), which in turn reduces the feedback loop for e2e tests.

- Splits linting and typechecking out of building. These are considered tests. They are not split out of development builds.
- Includes tests in typechecking (these had been omitted by CRA), and fixes erring types.
  - Typechecking is now configured by tsconfig.json, so that tests are included and checked.
- Updates eslint to ignore .gitignored files, and omits inappropriate linting from config and type files.
  - build/ and other .gitignored files had been included, making it prohibitively expensive to lint locally.

Follow-up steps:
- [ ] Once this is merged, set typechecking as a required test before merging other PRs.

### QA (ie manual testing)

- [x] Observe the improved test times in the GitHub Action summary.
- [x] Ensure linting/typechecking still work by triggering an error on a local build.

### Automated testing

- [ ] <s>Unit test</s>


.eslintrc.js
```diff
@@ -3,8 +3,14 @@
 require('@uniswap/eslint-config/load')
 
 module.exports = {
-  extends: '@uniswap/eslint-config/react',
+  extends: ['@uniswap/eslint-config/react'],
   overrides: [
+    {
+      files: ['**/*.config.*', '**/*.d.ts'],
+      rules: {
+        'import/no-unused-modules': 'off',
```
cmcewen, line 11: are you sure we should turn this off vs just comment things?

craco.config.cjs
```diff
@@ -29,25 +32,30 @@ module.exports = {
...
+        // Omit them from production builds, as they slow down the feedback loop.
+        webpackConfig.plugins = webpackConfig.plugins.filter((plugin) => {
+          if (plugin instanceof ForkTsCheckerWebpackPlugin) return false
+          if (plugin instanceof EsLintWebpackPlugin) return false
+          return true
+        })
+
+        // CSS ordering is mitigated through scoping / naming conventions, so we can ignore order warnings.
+        // See https://webpack.js.org/plugins/mini-css-extract-plugin/#remove-order-warnings.
+        webpackConfig.plugins.find((plugin) => plugin instanceof MiniCssExtractPlugin).options.ignoreOrder = true
```
cmcewen, line 53: do we lose our undefined check here on purpose?

craco.config.cjs
```diff
@@ -29,25 +32,30 @@ module.exports = {
...
-      if (instanceOfMiniCssExtractPlugin !== undefined) instanceOfMiniCssExtractPlugin.options.ignoreOrder = true
+      // Extend process.env with dynamic values (eg commit hash).
+      Object.assign(webpackConfig.plugins.find((plugin) => plugin instanceof DefinePlugin).definitions['process.env'], {
+        REACT_APP_GIT_COMMIT_HASH: JSON.stringify(commitHash),
+      })
+
+      if (isProduction) {
+        // Type checking and linting are only necessary as part of development and testing.
+        // Omit them from production builds, as they slow down the feedback loop.
+        webpackConfig.plugins = webpackConfig.plugins.filter((plugin) => {
```
cmcewen, line 45: is there a way we can do this via some config when we upgrade CRA or move to next? feels a bit weird to have to search through plugins to modify

craco.config.cjs
```diff
@@ -29,25 +32,30 @@ module.exports = {
...
-    ],
+    plugins: [new VanillaExtractPlugin({ identifiers: 'short' })],
     configure: (webpackConfig) => {
-      const instanceOfMiniCssExtractPlugin = webpackConfig.plugins.find(
-        (plugin) => plugin instanceof MiniCssExtractPlugin
-      )
-      if (instanceOfMiniCssExtractPlugin !== undefined) instanceOfMiniCssExtractPlugin.options.ignoreOrder = true
+      // Extend process.env with dynamic values (eg commit hash).
+      Object.assign(webpackConfig.plugins.find((plugin) => plugin instanceof DefinePlugin).definitions['process.env'], {
+        REACT_APP_GIT_COMMIT_HASH: JSON.stringify(commitHash),
```
cmcewen, line 39: why move this here? was it not working before?

cmcewen (commented):
excited for this change. while reviewing this PR i found myself trying to figure out why certain diffs were included (which when re-reading the description i got some answers), but still it's not clear why some formatting changes (array around eslint config extends, ordering of package.json commands, etc.) were added. i think it's borderline but fixing the tests in a separate PR first and then merging this would separate the diffs nicely

.eslintrc.js
```diff
@@ -3,8 +3,14 @@
 require('@uniswap/eslint-config/load')
 
 module.exports = {
-  extends: '@uniswap/eslint-config/react',
+  extends: ['@uniswap/eslint-config/react'],
```
zzmp, line 6: I can back out this array-ifying change - it's a no-op from a separate thing I was experimenting with locally.

.eslintrc.js
zzmp: yes: for config and typings files, it is typical to export things that are used outside of the transpiled package (eg not captured by the tsconfig), so I think it makes more sense to just turn this off entirely.

craco.config.cjs
zzmp: This was moved here to avoid adding another plugin to the stack, as it adds processing time. I assumed it was better to re-use the existing plugin, which is already installed by CRA.

craco.config.cjs
zzmp: No, this is the documented way to do so: the only way to do this, AFAIK, is to eject or use a package like craco that lets you modify it through a function exactly like this.

craco.config.cjs
zzmp: Yes: I've moved this into the `isProduction` check - it is always defined in production environments.

package.json
```diff
@@ -17,16 +17,17 @@
     "i18n:compile": "yarn i18n:extract && lingui compile",
     "i18n:pseudo": "lingui extract --locale pseudo && lingui compile",
     "prepare": "yarn contracts:compile && yarn graphql:fetch && yarn graphql:generate && yarn i18n:compile",
+    "postinstall": "patch-package",
```
zzmp, line 20: I reordered these to match the app lifecycle, and only felt comfortable doing so because I was already touching this file. I don't think it requires a separate PR.

craco.config.cjs
cmcewen: i'm okay with this but i think it's not necessarily clear that those two things are related - i would interpret this as only setting `ignoreOrder` in a production environment as a future reader, not that find will always be successful only in a prod environment

.eslintignore
```diff
@@ -1,5 +0,0 @@
-*.config.ts
-*.d.ts
-/src/graphql/data/__generated__/types-and-hooks.ts
-/src/graphql/thegraph/__generated__/types-and-hooks.ts
-/src/schema/schema.graphql
```
JFrankfurt, line 5: noticed these three files aren't mentioned elsewhere but CI is passing just fine--were these ignores just unnecessary? (it makes sense to me that generated files *should* be ignored.

package.json
```diff
@@ -17,16 +17,17 @@
...
     "start": "craco start",
     "build": "craco build",
     "serve": "serve build -l 3000",
-    "deduplicate": "yarn-deduplicate --strategy=highest",
-    "lint": "yarn eslint .",
+    "lint": "yarn eslint --ignore-path .gitignore .",
```
JFrankfurt, line 24: confused by the .gitignore reference when the file has been removed

src/state/application/reducer.test.ts
```diff
@@ -14,6 +14,7 @@ describe('application reducer', () => {
 
   beforeEach(() => {
     store = createStore(reducer, {
+      fiatOnramp: { available: false, availabilityChecked: false },
```
JFrankfurt, line 17: what changed to cause the linter to catch this now?

JFrankfurt (approved):

.eslintignore
zzmp: I think this was just a very stale file. By using .eslintignore, we were actually typechecking all of the build files. Instead, we should have been using .gitignore.

package.json
zzmp: .eslintignore was removed, which is used by default. .gitignore needs to be set explicitly as a CLI option (it is not available in the .eslintrc).

src/state/application/reducer.test.ts
zzmp: typechecking was being done as part of react-scripts, which seems to have only typechecked source files (not test files).
by breaking typechecking into its own step, it's applied to all files.

JFrankfurt (approved):

PR #6330, opened 2023-04-10
fix: ignores "user rejected transaction" errors

vm:
## Description
As part of a plan to reduce the number of Sentry errors, we are going through the most common issues and either fixing or catching errors.

This one is here: https://uniswap-labs.sentry.io/issues/4066659882/?project=4504255148851200&query=is%3Aunresolved&referrer=issue-stream&sort=freq&statsPeriod=14d&stream_index=17

In this case, it's not caught by the error boundary (as seen because it says `handled: no`). We don't need to log this error because user rejections are considered valid, and are not actually an error in our product.

JIRA ticket: https://uniswaplabs.atlassian.net/browse/WEB-3187

## Test plan
I added a test here to show that user rejected error matching strings should get filtered out.

### Automated testing
- [x] Unit test

### DIFF
--- a/src/tracing/errors.ts
+++ b/src/tracing/errors.ts
@@ -21,6 +21,9 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event: 
 
     // If the error is a network change, it should not be considered an exception.
     if (error.message.match(/underlying network changed/)) return null
+
+    // If the error is based on a user rejecting, it should not be considered an exception.
+    if (error.message.match(/user rejected transaction/)) return null
   }
 
   return event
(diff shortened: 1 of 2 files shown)


cmcewen (approved):
i don't think we should log an event on this one but worth considering which ones we may want to know about in terms of product analytics

src/tracing/errors.ts
```diff
@@ -21,6 +21,9 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event:
 
     // If the error is a network change, it should not be considered an exception.
     if (error.message.match(/underlying network changed/)) return null
+
+    // If the error is based on a user rejecting, it should not be considered an exception.
+    if (error.message.match(/user rejected transaction/)) return null
```
zzmp, line 26: Instead of re-implementing matching logic, can you replace this with `didUserReject` for x-wallet compatibility ([src](https://github.com/Uniswap/interface/blob/2ffc8a0bdfc66fc1372aa2f15cf6c64f4afdbf02/src/utils/swapErrorToUserReadableMessage.tsx#L12))?

zzmp (dismissed):

just-toby (approved):
makes sense

PR #6332, opened 2023-04-11
feat: use Swap Component on TDP

just-toby:
## Description

replace the widget on TDP with the regular Swap component. includes logic for automatically navigating between TDP when the input tokens change. 

_Relevant docs:_ https://www.notion.so/uniswaplabs/Swap-widget-replacement-work-bd91b44494d547259b4f29f4c36b3dfc
_Jira Ticket :_ https://uniswaplabs.atlassian.net/browse/WEB-3249?atlOrigin=eyJpIjoiOWUzOTk1NDRiMmViNDdjOTliYTlmODY1NDkxMDA1NTYiLCJwIjoiaiJ9

## Test plan

- [x] verify the token selections and automatic navigation are working as expected

### Automated testing

- [x] <s>Unit test</s>
- [x] Integration/E2E test


src/components/Tokens/TokenDetails/index.tsx
```diff
@@ -234,13 +226,11 @@ export default function TokenDetails({
 
         <RightPanel onClick={() => isBlockedToken && setOpenTokenSafetyModal(true)}>
           <div style={{ pointerEvents: isBlockedToken ? 'none' : 'auto' }}>
-            <Widget
-              defaultTokens={{
-                [Field.INPUT]: inputToken ?? undefined,
-                default: detailedToken ?? undefined,
+            <Swap
+              prefilledState={{
```
just-toby, line 230: note that the `inputCurrency` default is handled automatically by the default initialization of the SwapState reducer - it checks URL params for `inputCurrency` and `outputCurrency` on the main page, and we're just re-using the same variable name on TDP

src/components/Tokens/TokenDetails/index.tsx
```diff
@@ -234,13 +226,11 @@ export default function TokenDetails({
...
+                [Field.OUTPUT]: { currencyId: address === NATIVE_CHAIN_ID ? 'ETH' : address },
               }}
-              onDefaultTokenChange={navigateToWidgetSelectedToken}
-              onReviewSwapClick={onReviewSwapClick}
+              onCurrencyChange={handleCurrencyChange}
```
just-toby, line 233: opted for a single callback here instead of a callback for each of input and output.

this has the benefit of allowing us to check if the "Detailed" token is still one of the selections

src/components/Tokens/TokenDetails/index.tsx
```diff
@@ -234,13 +229,13 @@ export default function TokenDetails({
 
         <RightPanel onClick={() => isBlockedToken && setOpenTokenSafetyModal(true)}>
           <div style={{ pointerEvents: isBlockedToken ? 'none' : 'auto' }}>
-            <Widget
-              defaultTokens={{
-                [Field.INPUT]: inputToken ?? undefined,
-                default: detailedToken ?? undefined,
+            <Swap
+              defaultChainId={pageChainId}
+              prefilledState={{
```
tinaszheng, line 234: hmm so the chain id should always be coupled with `currencyId` right? can we just amend `currencyId` to include chain like mobile does

`{chainId}-{address}` format

tinaszheng: actually, might make more sense to leave the currencyIds as is but use a `useEffect` internal to the swap component that handles chain switching

tinaszheng (changes requested):

src/components/Tokens/TokenDetails/index.tsx
```diff
@@ -234,13 +229,13 @@ export default function TokenDetails({
 
         <RightPanel onClick={() => isBlockedToken && setOpenTokenSafetyModal(true)}>
           <div style={{ pointerEvents: isBlockedToken ? 'none' : 'auto' }}>
-            <Widget
-              defaultTokens={{
-                [Field.INPUT]: inputToken ?? undefined,
-                default: detailedToken ?? undefined,
+            <Swap
+              defaultChainId={pageChainId}
```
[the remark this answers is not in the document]
just-toby: `defaultChainId` refers to the Chain ID of the detailed token's chain, and if the user isn't connected to that chain we need to show a button to switch chains (this is how the widget works currently)

src/components/Tokens/TokenDetails/index.tsx
just-toby: we don't want to switch automatically, we want to show a button to switch chains in the swap component. 

as for combining the values, I worry that changing the format of a "Currency ID" in this one place would be confusing, since the ID is just the address everywhere else.

also, this "default" chain ID only applies to one token and doesn't really fit with the current approach of passing a subset of swap state into `prefilledState` so we'd have to change the API a bit. i'm leaning towards keeping this prop, but let me know what you think.

src/components/Tokens/TokenDetails/index.tsx
```diff
@@ -146,34 +144,31 @@ export default function TokenDetails({
...
-            inputAddress: tokens[Field.INPUT] ? getTokenAddress(tokens[Field.INPUT] as Currency) : null,
-          })
+  const handleCurrencyChange = useCallback(
+    (tokens: Partial<SwapState>) => {
+      if (tokens[Field.INPUT]?.currencyId !== address && tokens[Field.OUTPUT]?.currencyId !== address) {
+        const newDefaultTokenID = tokens[Field.OUTPUT]?.currencyId ?? tokens[Field.INPUT]?.currencyId
+        startTokenTransition(() =>
+          navigate(
+            getTokenDetailsURL({
+              address: newDefaultTokenID === 'ETH' ? 'NATIVE' : newDefaultTokenID,
```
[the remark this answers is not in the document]
just-toby: mainly because we only have the currency IDs (addresses) in scope

src/components/Tokens/TokenDetails/index.tsx
tinaszheng: shouldn't it just be a `chainId` prop then? default implies that it can change within the component internally, but seems like `chain` should be a fully controlled prop

src/pages/Swap/index.tsx
```diff
@@ -143,13 +143,35 @@ function largerPercentValue(a?: Percent, b?: Percent) {
...
+        <SwitchLocaleLink />
+      </>
+    </Trace>
+  )
+}
+
+export function Swap({
+  className,
+  prefilledState,
+  defaultChainId,
```
tinaszheng, line 163: hm still confused about the mixed usage of `defaultChainId` vs `useWeb3React().chainId` here 

so it looks like some hooks read from `defaultChainId` but some filtering later on is done through `chainId` - can it just read from the passed in prop instead of through the `useWeb3React` hook for consistency?

just-toby: yes, i think we can use the `pageChainId` (renamed from `defaultChainId`) in more places - unfortunately i think we still need references to both, because we need to check whether the user needs to switch chains in their wallet.

src/components/Tokens/TokenDetails/index.tsx
just-toby: i renamed to `pageChainId` which i think is clearer

tinaszheng (approved):

tinaszheng:
is moving all the swap component logic to a separate folder a followup task? i can do that after :P

grabbou:
> is moving all the swap component logic to a separate folder a followup task? i can do that after :P

@tinaszheng we should definitely do that sooner than later. I am also happy to help, as I was thinking of a folder for the menu things anyways.

cypress/utils/index.ts
```diff
@@ -2,4 +2,6 @@ export const getTestSelector = (selectorId: string) => `[data-testid=${selectorI
 
 export const getTestSelectorStartsWith = (selectorId: string) => `[data-testid^=${selectorId}]`
 
+// TODO: Remove this comment if you use the util.
+// eslint-disable-next-line import/no-unused-modules
```
grabbou, line 6: What does this mean in practice?

src/components/Tokens/TokenDetails/index.tsx
```diff
@@ -146,34 +145,35 @@ export default function TokenDetails({
...
+        addressesAreEquivalent(tokens[Field.OUTPUT]?.currencyId, address)
+      ) {
+        return
+      }
+      const newDefaultTokenID = tokens[Field.OUTPUT]?.currencyId ?? tokens[Field.INPUT]?.currencyId
       startTokenTransition(() =>
         navigate(
           getTokenDetailsURL({
-            address,
+            address: newDefaultTokenID === 'ETH' ? 'NATIVE' : newDefaultTokenID,
```
grabbou, line 160: nit: should we create a constant file in constants for this? ETH/NATIVE? Looks like a follow-up task, b/c we're not using that anywhere. Just looks strange to be using strings.

src/components/Tokens/TokenDetails/index.tsx
```diff
@@ -146,34 +145,35 @@ export default function TokenDetails({
...
             chain,
-            inputAddress: tokens[Field.INPUT] ? getTokenAddress(tokens[Field.INPUT] as Currency) : null,
+            inputAddress:
+              tokens[Field.INPUT] && tokens[Field.INPUT]?.currencyId !== newDefaultTokenID
```
grabbou, line 163: nit: could we have a comment here describing `tokens[Field.INPUT]?.currencyId !== newDefaultTokenID`? This may seem obvious, so if that's the case, just an explanation here will be enough.

src/components/Tokens/TokenDetails/index.tsx
```diff
@@ -234,13 +234,13 @@ export default function TokenDetails({
...
           <div style={{ pointerEvents: isBlockedToken ? 'none' : 'auto' }}>
-            <Widget
-              defaultTokens={{
-                [Field.INPUT]: inputToken ?? undefined,
-                default: detailedToken ?? undefined,
+            <Swap
+              chainId={pageChainId}
+              prefilledState={{
+                [Field.INPUT]: { currencyId: inputToken?.wrapped?.address },
+                [Field.OUTPUT]: { currencyId: address === NATIVE_CHAIN_ID ? 'ETH' : address },
```
grabbou, line 241: nit: this is unrelated to this PR, but I would consider this a good follow-up task (maybe we can write this down). I believe `Swap` component and `getTokenDetailsURL` should check INSIDE for `address === NATIVE_CHAIN_ID` and use `ETH` or `address` instead of the caller.

src/hooks/Tokens.ts
```diff
@@ -10,31 +10,33 @@ import { isL2ChainId } from 'utils/chains'
...
 import { TokenAddressMap, useUnsupportedTokenList } from './../state/lists/hooks'
 
 // reduce token map into standard address <-> Token mapping, optionally include user added tokens
-function useTokensFromMap(tokenMap: TokenAddressMap): { [address: string]: Token } {
-  const { chainId } = useWeb3React()
+function useTokensFromMap(tokenMap: TokenAddressMap, chainId?: SupportedChainId): { [address: string]: Token } {
+  const { chainId: connectedChainId } = useWeb3React()
   return useMemo(() => {
-    if (!chainId) return {}
+    if (!chainId && !connectedChainId) return {}
```
grabbou, line 20: why `chainId` is optional argument to `useTokensFromMap` if we seem to return from the function ALWAYS when it's not defined? We should mark it as required? We already have some issues with hooks-based APIs, but I think we're all used to using ugly ternary operators at the call-site, instead of accepting undefined as arguments and handling that inside a hook.

src/hooks/Tokens.ts
```diff
@@ -10,31 +10,33 @@ import { isL2ChainId } from 'utils/chains'
...
+
+    const targetChainId: number = chainId ?? (connectedChainId as number)
```
grabbou, line 22: nit: is `targetChainId: number` required?

src/state/swap/hooks.tsx
```diff
@@ -71,7 +72,10 @@ const BAD_RECIPIENT_ADDRESSES: { [address: string]: true } = {
 }
 
 // from the current swap inputs, compute the best trade and return it.
-export function useDerivedSwapInfo(state: SwapState): {
+export function useDerivedSwapInfo(
+  state: SwapState,
+  defaultChainId?: SupportedChainId
```
grabbou, line 77: nit: I would rephrase this to just `chainId`, because we want to get information for a given chain and there's no default chain anymore. I would rather let Swap page (main page) pass explicit `chainId` from web3react explicitly rather than handle that inside. And let's make it a required argument.

grabbou (commented):
PR overall looks good, I just have some high-level comments re; abstractions and making it easier to work with swaps in the future. Since we're now transitioning from `web3React`-driven `chainId` to a chainId that is arbitrary (b/c of the way we are browsing Swap component on a Tokens page), I would consider some of the comments I have left as changes to apply!

src/state/swap/hooks.tsx
grabbou: Then I would remove `chainId = useWeb3React` from `[src/hooks/Tokens.ts](https://github.com/Uniswap/interface/pull/6332/files/8ddb100fbc639d4191edd6137c9a87e648e89595#diff-3340e3e31b3f7eff7bb21323f4ee0e638c6f05d468225e0b46511ceee4462d8e)` and explicitly pass it on Swap page too.

cypress/utils/index.ts
just-toby: I could remove it, but I just felt like it was a useful enough util to leave in the codebase despite the linter complaining.

src/components/Tokens/TokenDetails/index.tsx
just-toby: actually good point, we have a constant used elsewhere for the "NATIVE" string , and `getTokenDetailsURL` uses it as the fallback. i guess i should be using `null` here instead!

src/hooks/Tokens.ts
just-toby: @grabbou we return early when both `chainId` *and* `useWeb3React().chainId` are undefined. the argument is optional because we fallback to the connected chain Id in this function.

do you think i should remove the fallback, and move the `useWeb3React()` out of this function to each callsite?

just-toby: ah nevermind, i see your other comment now!

cypress/utils/index.ts
tinaszheng: i vote to just remove it until it's useful

src/components/Tokens/TokenDetails/index.tsx
```diff
@@ -234,13 +237,14 @@ export default function TokenDetails({
...
         <RightPanel onClick={() => isBlockedToken && setOpenTokenSafetyModal(true)}>
           <div style={{ pointerEvents: isBlockedToken ? 'none' : 'auto' }}>
-            <Widget
-              defaultTokens={{
-                [Field.INPUT]: inputToken ?? undefined,
-                default: detailedToken ?? undefined,
+            <Swap
+              chainId={pageChainId}
+              prefilledState={{
+                [Field.INPUT]: { currencyId: inputToken?.wrapped?.address },
```
tinaszheng, line 243: wait why does this pass in the wrapped version?

src/components/Tokens/TokenDetails/index.tsx
```diff
@@ -234,13 +237,14 @@ export default function TokenDetails({
...
+                // TODO: move this check into the Swap component
```
tinaszheng, line 244: what does this comment mean

src/components/Tokens/TokenDetails/index.tsx
```diff
@@ -146,34 +144,38 @@ export default function TokenDetails({
...
       startTokenTransition(() =>
         navigate(
           getTokenDetailsURL({
-            address,
+            // The function falls back to "NATIVE" if the address is null
+            address: newDefaultTokenID === 'ETH' ? null : newDefaultTokenID,
             chain,
-            inputAddress: tokens[Field.INPUT] ? getTokenAddress(tokens[Field.INPUT] as Currency) : null,
+            inputAddress:
+              // If only one token was selected before we navigate, it becomes the output and we don't need to set an input token.
```
grabbou, line 164: Two questions:
- we don't need to set an input token. -> WHY?
-  If only one token was selected before we navigate -> when that selection happens? How can I test this flow?

src/lib/hooks/useNativeCurrency.ts
```diff
@@ -1,11 +1,9 @@
 import { NativeCurrency, Token } from '@uniswap/sdk-core'
-import { useWeb3React } from '@web3-react/core'
 import { SupportedChainId } from 'constants/chains'
 import { nativeOnChain } from 'constants/tokens'
 import { useMemo } from 'react'
 
-export default function useNativeCurrency(): NativeCurrency | Token {
-  const { chainId } = useWeb3React()
+export default function useNativeCurrency(chainId: SupportedChainId | undefined): NativeCurrency | Token {
```
grabbou, line 6: nit: shouldn't this be `chainId?: SupportedChainId` instead?

src/utils/addressesAreEquivalent.ts
```diff
@@ -1,4 +1,4 @@
-export function addressesAreEquivalent(a?: string, b?: string) {
+export function addressesAreEquivalent(a?: string | null, b?: string | null) {
```
grabbou, line 1: nit: I am generally confused with nulls vs undefined in our codebase sometimes, when this can be undefined, and when this can be null?

grabbou (approved):
LGTM. Left two more comments to understand some other parts better, but thanks for addressing my previous comments, it looks great!

src/lib/hooks/useNativeCurrency.ts
just-toby: in this case i agree

src/utils/addressesAreEquivalent.ts
just-toby: i think this function should allow nullish / undefined values, but it should be required to pass 2 arguments.

src/components/Tokens/TokenDetails/index.tsx
just-toby: let me try to improve the comment here

src/pages/Swap/index.tsx
```diff
@@ -194,16 +225,34 @@ export default function Swap({ className }: { className?: string }) {
...
   const [isExpertMode] = useExpertModeManager()
   // swap state
-  const [state, dispatch] = useReducer(swapReducer, initialSwapState)
+  const [state, dispatch] = useReducer(swapReducer, { ...initialSwapState, ...prefilledState })
   const { typedValue, recipient, independentField } = state
+
+  const previousConnectedChainId = usePrevious(connectedChainId)
+  useEffect(() => {
+    const combinedInitialState = { ...initialSwapState, ...prefilledState }
+    console.log({ combinedInitialState })
```
tinaszheng, line 234: pls delete

src/pages/Swap/index.tsx
```diff
@@ -194,16 +225,34 @@ export default function Swap({ className }: { className?: string }) {
...
+    if (previousConnectedChainId && previousConnectedChainId !== connectedChainId) {
+      dispatch(
+        replaceSwapState({
+          ...initialSwapState,
+          ...prefilledState,
+          field: combinedInitialState.independentField ?? Field.INPUT,
```
tinaszheng, line 240: is this needed? doesnt the spread above do the same? and isn't `independentField` always defined

src/pages/Swap/index.tsx
```diff
@@ -194,16 +225,34 @@ export default function Swap({ className }: { className?: string }) {
...
+          inputCurrencyId: combinedInitialState.INPUT.currencyId ?? undefined,
+          outputCurrencyId: combinedInitialState.OUTPUT.currencyId ?? undefined,
```
tinaszheng, line 242: same here - is this needed? also don't you want to clear and reset to the new chain's native token

src/pages/Swap/index.tsx
```diff
@@ -418,8 +470,14 @@ export default function Swap({ className }: { className?: string }) {
   const handleInputSelect = useCallback(
     (inputCurrency: Currency) => {
       onCurrencySelection(Field.INPUT, inputCurrency)
+      onCurrencyChange?.({
+        [Field.INPUT]: {
+          currencyId: inputCurrency.isToken ? inputCurrency.address : inputCurrency.isNative ? NATIVE_CHAIN_ID : '',
```
tinaszheng, line 475: this should be made into a helper function

src/pages/Swap/index.tsx
```diff
@@ -475,271 +539,245 @@ export default function Swap({ className }: { className?: string }) {
   const showDetailsDropdown = Boolean(
     !showWrap && userHasSpecifiedInputOutput && (trade || routeIsLoading || routeIsSyncing)
   )
+  const disableTokenInputs = chainId !== connectedChainId
```
tinaszheng, line 542: when would this happen on the swap page? isn't this only relevant on token details?

src/pages/Swap/index.tsx
```diff
@@ -143,13 +144,43 @@ function largerPercentValue(a?: Percent, b?: Percent) {
...
+  chainId,
+  onCurrencyChange,
+}: {
+  className?: string
+  prefilledState?: Partial<SwapState>
+  chainId: SupportedChainId | undefined
+  onCurrencyChange?: (selected: Pick<SwapState, Field.INPUT | Field.OUTPUT>) => void
+}) {
+  const { account, chainId: connectedChainId, connector } = useWeb3React()
   const loadedUrlParams = useDefaultsFromURLSearch()
```
tinaszheng, line 181: it still doesn't make sense to me that the URL parsing happens from within the shared component instead of on the page

imo this should be handled at the page level and the addresses should be passed in as part of `prefilledState`

src/pages/Swap/index.tsx
```diff
@@ -194,16 +225,34 @@ export default function Swap({ className }: { className?: string }) {
...
   // swap state
-  const [state, dispatch] = useReducer(swapReducer, initialSwapState)
+  const [state, dispatch] = useReducer(swapReducer, { ...initialSwapState, ...prefilledState })
   const { typedValue, recipient, independentField } = state
+
+  const previousConnectedChainId = usePrevious(connectedChainId)
+  useEffect(() => {
+    const combinedInitialState = { ...initialSwapState, ...prefilledState }
+    console.log({ combinedInitialState })
+    if (previousConnectedChainId && previousConnectedChainId !== connectedChainId) {
```
tinaszheng, line 235: why do the comparison like this instead of just comparing `connectedChainId !== chainId`?

tinaszheng (changes requested):

src/pages/Swap/index.tsx
just-toby: this is needed because the reducer action uses a different name (`field`) than the state type def (`independentField`). i wanted to avoid changing it to keep my changes minimal

src/pages/Swap/index.tsx
just-toby: that's correct. i can move it out to the callsite

src/pages/Swap/index.tsx
just-toby: i don't think that's the moment we want to clear the inputs. `connectedChainId !== chainId` would be true on the TDP if you're looking at tokens from other chains, and there's no need to run this reset action then (although i think it wouldn't be a noticable change to the user). 

conceptually, we only want to reset the inputs when the connected chain changes, which is what i'm checking here

src/pages/Swap/index.tsx
just-toby: same reason - the reducer action expects `inputCurrencyId` and `outputCurrencyId` instead of `[Field.INPUT]: { currencyId }`

and, `combinedInitialState` at this point has the tokens we want to default to, with the new chain's native token.

src/constants/tokens.ts
```diff
@@ -515,6 +515,16 @@ export function nativeOnChain(chainId: number): NativeCurrency | Token {
   return (cachedNativeCurrency[chainId] = nativeCurrency)
 }
 
+export function getSwapCurrencyId(currency: Currency): string {
+  if (currency.isToken) {
+    return currency.address
+  }
+  if (currency.isNative) {
```
tinaszheng, line 522: a currency can only be `isToken` or `isNative` 
```suggestion
return `NATIVE_CHAIN_ID`
```

src/lib/hooks/useCurrency.ts
```diff
@@ -3,6 +3,7 @@ import { parseBytes32String } from '@ethersproject/strings'
 import { Currency, Token } from '@uniswap/sdk-core'
 import { useWeb3React } from '@web3-react/core'
 import { isSupportedChain, SupportedChainId } from 'constants/chains'
+import { Maybe } from 'graphql/jsutils/Maybe'
```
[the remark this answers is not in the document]
just-toby: ok, removed it

src/components/Tokens/TokenDetails/index.tsx
```diff
@@ -234,13 +238,14 @@ export default function TokenDetails({
 
         <RightPanel onClick={() => isBlockedToken && setOpenTokenSafetyModal(true)}>
           <div style={{ pointerEvents: isBlockedToken ? 'none' : 'auto' }}>
-            <Widget
-              defaultTokens={{
-                [Field.INPUT]: inputToken ?? undefined,
-                default: detailedToken ?? undefined,
+            <Swap
```
tinaszheng, line 241: i forgot about this until today but are we still planning on feature flagging the replacement here?

as for the callback function changes, i would suggest keeping the old ones and using new ones for the `Swap` component :3 and we can delete the old ones once we get rid of feature flag logic

tinaszheng (changes requested):

tinaszheng (approved):
image

PR #6334, opened 2023-04-11
fix: ignore ethers call revert exceptions

vm:
As part of a plan to reduce the number of Sentry errors, we are going through the most common issues and either fixing or catching errors.

This one is here: https://uniswap-labs.sentry.io/issues/4076446793/?project=4504255148851200&query=is%3Aunresolved&referrer=issue-stream&sort=freq&statsPeriod=24h&stream_index=1

In this case, it's not caught by the error boundary (as seen because it says handled: no).

This is caused when there is an error thrown by an ethers call. The issue is we do not wrap all provider calls in try/catch, even though any of them are capable of failing. Over time we should start doing this but for now these errors are not actionable.

JIRA ticket: https://uniswaplabs.atlassian.net/browse/WEB-3187

Test plan
I added a test here to show that user rejected error matching strings should get filtered out.

Automated testing
 Unit test (need to fix this)
 Integration/E2E test ✅  (N/A)

### DIFF
--- a/src/tracing/errors.test.ts
+++ b/src/tracing/errors.test.ts
@@ -21,9 +21,9 @@ describe('filterKnownErrors', () => {
   })
 
   it('filters call revert errors', () => {
-    const originalException = new Error(
-      'call revert exception [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ]'
-    )
+    const originalException = new (class extends Error {
+      requestBody = JSON.stringify({ method: 'collect((uint256,address,uint128,uint128))' })
+    })()
     expect(filterKnownErrors(ERROR, { originalException })).toBe(null)
   })
 })
--- a/src/tracing/errors.test.ts
+++ b/src/tracing/errors.test.ts
@@ -19,4 +19,11 @@ describe('filterKnownErrors', () => {
     const originalException = new Error('underlying network changed')
     expect(filterKnownErrors(ERROR, { originalException })).toBe(null)
   })
+
+  it('filters call revert errors', () => {
+    const originalException = new Error(
+      'call revert exception [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ]'
+    )
+    expect(filterKnownErrors(ERROR, { originalException })).toBe(null)
+  })
 })
--- a/src/tracing/errors.test.ts
+++ b/src/tracing/errors.test.ts
@@ -22,7 +22,7 @@ describe('filterKnownErrors', () => {
 
   it('filters call revert errors', () => {
     const originalException = new (class extends Error {
-      requestBody = JSON.stringify({ method: 'collect((uint256,address,uint128,uint128))' })
+      requestBody = JSON.stringify({ code: 'CALL_EXCEPTION' })
     })()
     expect(filterKnownErrors(ERROR, { originalException })).toBe(null)
   })
--- a/src/tracing/errors.ts
+++ b/src/tracing/errors.ts
@@ -17,9 +17,10 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event: 
       // If block number polling fails, it should not be considered an exception.
       if (method === 'eth_blockNumber') return null
 
+      const code = JSON.parse(error.requestBody).code
       // For now, these errors are not actionable so we should not report them.
       // ethers exceptions are currently not caught in the codebase because we don't try/catch all provider calls.
-      if (method.startsWith('collect(')) return null
+      if (code.startsWith('CALL_EXCEPTION')) return null
     }
 
     // If the error is a network change, it should not be considered an exception.
--- a/src/tracing/errors.ts
+++ b/src/tracing/errors.ts
@@ -5,7 +5,6 @@ function isEthersRequestError(error: Error): error is Error & { requestBody: str
   return 'requestBody' in error && typeof (error as unknown as Record<'requestBody', unknown>).requestBody === 'string'
 }
 /**
-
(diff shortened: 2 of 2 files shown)


src/tracing/errors.test.ts
```diff
@@ -19,4 +19,11 @@ describe('filterKnownErrors', () => {
     const originalException = new Error('underlying network changed')
     expect(filterKnownErrors(ERROR, { originalException })).toBe(null)
   })
+
+  it('filters call revert errors', () => {
+    const originalException = new Error(
```
vm, line 24: need to look into how to make this match as an ethers exception

src/tracing/errors.ts
```diff
@@ -4,21 +4,25 @@ import { ClientOptions, ErrorEvent, EventHint } from '@sentry/types'
...
  * Intended as a {@link ClientOptions.beforeSend} callback. Returning null filters the error from Sentry.
  */
 export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event: ErrorEvent, hint: EventHint) => {
   const error = hint.originalException
   if (error instanceof Error) {
-    // ethers aggressively polls for block number, and it sometimes fails (whether spuriously or through rate-limiting).
-    // If block number polling, it should not be considered an exception.
     if (isEthersRequestError(error)) {
       const method = JSON.parse(error.requestBody).method
+      // ethers aggressively polls for block number, and it sometimes fails (whether spuriously or through rate-limiting).
```
vm, line 17: moved the comment in a bit because it was confusing for me

src/tracing/errors.ts
```diff
@@ -4,21 +4,25 @@ import { ClientOptions, ErrorEvent, EventHint } from '@sentry/types'
...
+      // If block number polling, it should not be considered an exception.
       if (method === 'eth_blockNumber') return null
     }
 
+    // for now, these errors are not actionable so we should not report them.
```
vm, line 22: unclear if we need this to live within the `isEthersRequestError` section above?

src/tracing/errors.ts
```diff
@@ -4,21 +4,25 @@ import { ClientOptions, ErrorEvent, EventHint } from '@sentry/types'
 function isEthersRequestError(error: Error): error is Error & { requestBody: string } {
   return 'requestBody' in error && typeof (error as unknown as Record<'requestBody', unknown>).requestBody === 'string'
 }
-
 /**
+
```
zzmp, line 8: unintentional newline?

src/tracing/errors.ts
```diff
@@ -4,21 +4,25 @@ import { ClientOptions, ErrorEvent, EventHint } from '@sentry/types'
...
  */
 export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event: ErrorEvent, hint: EventHint) => {
   const error = hint.originalException
   if (error instanceof Error) {
-    // ethers aggressively polls for block number, and it sometimes fails (whether spuriously or through rate-limiting).
-    // If block number polling, it should not be considered an exception.
     if (isEthersRequestError(error)) {
       const method = JSON.parse(error.requestBody).method
+      // ethers aggressively polls for block number, and it sometimes fails (whether spuriously or through rate-limiting).
+      // If block number polling, it should not be considered an exception.
```
zzmp, line 18:
```suggestion
      // If block number polling fails, it should not be considered an exception.
```

src/tracing/errors.ts
zzmp, line 22:
```suggestion
    // For now, these errors are not actionable so we should not report them.
```

zzmp: I think this should match an ethers request error, so you can include it in that section to increase specificity of the matching.

src/tracing/errors.test.ts
zzmp: I believe you can reuse the logic from the prior test, with `method: 'collect(...)'`.

zzmp (changes requested):
> even though any of them are capable of failing when the RPC provider is down.

This is failing with a call revert exception, so the RPC provider was up, but the call failed on-chain. Does that change whether we want to report it?

src/tracing/errors.ts
```diff
@@ -4,19 +4,22 @@ import { ClientOptions, ErrorEvent, EventHint } from '@sentry/types'
...
-    // If block number polling, it should not be considered an exception.
     if (isEthersRequestError(error)) {
       const method = JSON.parse(error.requestBody).method
+      // ethers aggressively polls for block number, and it sometimes fails (whether spuriously or through rate-limiting).
+      // If block number polling fails, it should not be considered an exception.
       if (method === 'eth_blockNumber') return null
+
+      // For now, these errors are not actionable so we should not report them.
+      // ethers exceptions are currently not caught in the codebase because we don't try/catch all provider calls.
+      if (method.startsWith('collect(')) return null
```
just-toby, line 22: `collect` here seems to be referring to a specific onchain interaction for LPs ([link](https://gist.github.com/Aboudjem/886f9fae8649d3801b4ffe6e9bac1564#file-uniswapinterfaces-sol-L1211) ) so maybe update the comment to reflect that, or use a more generic string match .

vm: updated to use CALL_EXCEPTION code, but we may be better off fixing the specific error here instead?

vm:
I think we should reconsider if this is necessary and get back to it, so i'll make it a draft.

vm:
Now that we know it's possible to see the specific page and contract that caused the revert, we no longer want to filter these out!

PR #6337, opened 2023-04-11
fix: last sale nan

JackShort:
fixes undefined last sale bug on the nft details page

Screen Shot 2023-04-11 at 11 30 54 AM

### DIFF
--- a/src/nft/components/details/AssetDetails.tsx
+++ b/src/nft/components/details/AssetDetails.tsx
@@ -313,15 +313,16 @@ export const AssetDetails = ({ asset, collection }: AssetDetailsProps) => {
     'no-cache'
   )
 
-  // TODO simplify typecasting when removing graphql flag
-  const lastSalePrice = isNftGraphqlEnabled ? gqlPriceData?.[0]?.price : priceData?.events[0]?.price
-  const formattedPrice = lastSalePrice
-    ? isNftGraphqlEnabled
-      ? formatEth(parseFloat(lastSalePrice ?? ''))
-      : putCommas(formatEthPrice(lastSalePrice)).toString()
-    : null
-  const [activeFilters, filtersDispatch] = useReducer(reduceFilters, initialFilterState)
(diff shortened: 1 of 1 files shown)


src/nft/components/details/AssetDetails.tsx
```diff
@@ -318,10 +318,10 @@ export const AssetDetails = ({ asset, collection }: AssetDetailsProps) => {
...
     ? formatEth(parseFloat(lastSalePrice ?? ''))
     : formatEthPrice(lastSalePrice) || 0
-  const formattedPrice = isNftGraphqlEnabled
-    ? formattedEthprice
-    : lastSalePrice
-    ? putCommas(parseFloat(formattedEthprice.toString())).toString()
+  const formattedPrice = lastSalePrice
+    ? isNftGraphqlEnabled
+      ? formattedEthprice
+      : putCommas(parseFloat(formattedEthprice.toString())).toString()
```
cbachmeier, line 324: I believe the correct order is still having the `isNftGraphqlEnabled` check first as the main branch for the if/else but then instead of checking for `formattedEthprice` check instead for `lastSalePrice`

cbachmeier: We actually can't do that simplification as it leads to an accidental truthy and nan being shown but the first example works as expected with no last sale price

cbachmeier: Honestly this whole block could be simplified. Let's replace this and `formattedEthprice` with the following which reduces a lot of redundancy
```
const formattedPrice = lastSalePrice
    ? isNftGraphqlEnabled
      ? formatEth(parseFloat(lastSalePrice ?? ''))
      : putCommas(formatEthPrice(lastSalePrice))
    : null

cbachmeier (approved):
LGTM. Tested for an NFT with and without previous sales

zzmp (approved):

PR #6338, opened 2023-04-11
fix: handle invalid chainId on position page

vm:
Makes sure that being on an unsupported chain does not crash the /pools/{positionId} pages.

The error is caused here: https://github.com/Uniswap/interface/blob/main/src/constants/tokens.ts#L492

I don't think we should change the behavior there to not throw an error, because there are other users of that method. We would need to test all of them. Similarly, returning undefined or mainnet ether is not reasonable behavior in my opinion. It should be a precondition that we are verified to be on a supported chain before getting to that call.

I need to separate the Invalid state into a separate component so the wrap hook isn't called before we check if it's an unsupported chain. This is a rule of hooks that we can't conditionally call them.

## Description
_Slack thread:_ https://uniswapteam.slack.com/archives/C047U65H422/p1681232054878529

## Screen capture

after:
invalid position on supported chain
Screen Shot 2023-04-11 at 12 04 51 PM

invalid position on unsupported chain
Screen Shot 2023-04-11 at 12 52 06 PM

valid position on supported chain
Screen Shot 2023-04-11 at 12 52 13 PM

before:
Screen Shot 2023-04-11 at 2 02 46 PM

### QA (ie manual testing)
- Go to /pools/12401824 and then switch to Sepolia in MetaMask
- Should see "go back to pools", not a crash

#### Devices
I tested on desktop Chrome.

### Automated testing
- [x] Unit test (n/a)
- [x] Integration/E2E test (n/a -- this is not doable without work to integrate Hardhat with Cypress, as we need to switch to an unsupported network to test this)


src/pages/Pool/PositionPage.tsx
```diff
@@ -346,7 +346,32 @@ const useInverter = ({
...
+        <Trans>To view a position, you must be connected to the network it belongs to.</Trans>
+      </ThemedText.BodyPrimary>
+      <PositionPageButtonPrimary as={Link} to="/pools" width="fit-content">
+        <Trans>Back to Pools</Trans>
+      </PositionPageButtonPrimary>
+    </div>
+  </PageWrapper>
+)
+
+export default function PositionPageWrapper() {
```
zzmp, line 365: `PositionPage.tsx` should export the `PositionPage`, so could you rename these?
`PositionPageWrapper` -> `PositionPage`
`PositionPage` -> `PositionPageContent`

src/pages/Pool/PositionPage.tsx
```diff
@@ -346,7 +346,32 @@ const useInverter = ({
   }
 }
 
-export function PositionPage() {
+const INVALID_STATE = (
```
zzmp, line 349: To be more idiomatic, could you render this as a function? I'm not sure what the perf implications are of fn vs const, but this seems non-standard.

```suggestion
function PositionNetworkMismatch() {
```

src/pages/Pool/PositionPage.tsx
```diff
@@ -346,7 +346,32 @@ const useInverter = ({
...
+      <PositionPageButtonPrimary as={Link} to="/pools" width="fit-content">
+        <Trans>Back to Pools</Trans>
+      </PositionPageButtonPrimary>
+    </div>
+  </PageWrapper>
+)
+
+export default function PositionPageWrapper() {
+  const { chainId } = useWeb3React()
+  if (isSupportedChain(chainId)) {
```
zzmp, line 367: Is it necessary to pull this into a separate component because the logic already in `PositionPage` will crash?

zzmp (changes requested):

src/pages/Pool/PositionPage.tsx
vm: yes exactly. without doing this the hooks in that component will run which we want to short circuit out of

src/pages/Pool/PositionPage.tsx
```diff
@@ -346,7 +346,32 @@ const useInverter = ({
   }
 }
 
-export function PositionPage() {
+const PositionPageUnsupportedContent = () => (
```
zzmp, line 349: Can you make this a function to be more idiomatic?

```suggestion
function PositionPageUnsupportedContent() {
```

zzmp (changes requested):
lgtm, just change the const to use the more idiomatic function (or justify the const's usage)

zzmp (changes requested):
lgtm, just change the const to use the more idiomatic function (or justify the const's usage)

zzmp (approved):

src/pages/Pool/PositionPage.tsx
```diff
@@ -588,21 +615,7 @@ export function PositionPage() {
...
-          <ThemedText.BodyPrimary style={{ marginBottom: '32px' }}>
-            <Trans>To view a position, you must be connected to the network it belongs to.</Trans>
-          </ThemedText.BodyPrimary>
-          <PositionPageButtonPrimary as={Link} to="/pools" width="fit-content">
-            <Trans>Back to Pools</Trans>
-          </PositionPageButtonPrimary>
-        </div>
-      </PageWrapper>
-    )
+    return <PositionPageUnsupportedContent />
```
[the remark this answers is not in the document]
vm: this is one thing i hate about hooks. there's no good way to only conditionally call them, or i'd need to make sure every one of those hooks can support an undefined state

JFrankfurt (approved):
Discussed tradeoffs of resolving this at the base fn level (`nativeOnChain` calling `ExtendedEther#wrapped`), but decided that the approach in this PR is sufficient, since each caller should handle unsupported chains on its own.

PR #6340, opened 2023-04-11
fix: handle fee collect error

vm:
## Description
This change solves a bug where a provider call fails and causes the error boundary to be reached. You will see the error reported here: https://uniswap-labs.sentry.io/issues/4079703285/?project=4504255148851200&query=is%3Aunresolved+call+revert+exception&referrer=issue-stream&sort=freq&statsPeriod=24h&stream_index=1

### Reproducing the error
1. Change the `.then` in related promise to throw an Error.
2. Confirm that it is caught and the collecting state goes back to false.

#### Devices
I only tested on Desktop Chrome for this.

### Automated testing

- [ ] Unit test (WIP)
- [x ] Integration/E2E test (N/A)

### DIFF
--- a/src/pages/Pool/PositionPage.tsx
+++ b/src/pages/Pool/PositionPage.tsx
@@ -514,6 +514,10 @@ export function PositionPage() {
               expectedCurrencyOwed1: CurrencyAmount.fromRawAmount(currency1ForFeeCollectionPurposes, 0).toExact(),
             })
           })
+          .catch((error) => {
+            setCollecting(false)
+            console.error(error)
+          })
       })
       .catch((error) => {
         setCollecting(false)


vm:
Gonna timebox a test for this

zzmp (changes requested):
I don't think you should convert this from promise to async-await styled async code in this PR. Keeping this PR to _just_ the relevant catch will de-risk the change, which is important considering it is core functionality that is rarely touched, so it doesn't have a lot of mindshare in case something _does_ go wrong.

vm:
After more investigation, this solution won't work because the error _is_ seemingly getting caught by the external promise. I am adding further logging in a different PR and will come back to this error.

PR #6342, opened 2023-04-11
test: add test for position page states

vm:
## Description
Adds tests for the position pages, both invalid and valid pools. This sets us up for the next PR, which will make sure that unsupported chains also show the invalid state vs crashing.

#### Devices
The test works on Chrome only right now, since it's through Cypress.

### Automated testing
- [x] Unit test (n/a)
- [x] Integration/E2E test

### DIFF
--- a/cypress/e2e/position.test.ts
+++ b/cypress/e2e/position.test.ts
@@ -1,0 +1,11 @@
+describe('Position', () => {
+  it('shows an valid state on a supported network', () => {
+    cy.visit('/pools/1')
+    cy.contains('UNI / ETH')
+  })
+
+  it('shows an invalid state on a supported network', () => {
+    cy.visit('/pools/788893')
+    cy.contains('To view a position, you must be connected to the network it belongs to.')
+  })
+})


cypress/e2e/position.test.ts
```diff
@@ -0,0 +1,11 @@
+describe('Position', () => {
+  it('shows an valid state on a supported network', () => {
+    cy.visit('/pools/1')
+    cy.contains('UNI / ETH')
+  })
+
+  it('shows an invalid state on a supported network', () => {
+    cy.visit('/pools/788893')
+    cy.contains('To view a position, you must be connected to the network it belongs to.')
```
JFrankfurt, line 9: I expected to see an assertion here instead of just a dom query. The failure case for `contains` is continuous retries until the test times out, so I think you should add a dedicated timeout and an explicit standalone assertion here.

JFrankfurt (changes requested):

cypress/e2e/position.test.ts
vm: i based on https://docs.cypress.io/guides/references/best-practices#Text-Content

JFrankfurt (approved):
changed my mind after some discussion w/r/t cypress best practices:

https://docs.cypress.io/guides/references/best-practices#Text-Content

JFrankfurt (approved):

cmcewen (approved):

PR #6345, opened 2023-04-11
build: add chainId as tag in sentry ErrorBoundary

vm:
## Description
Adds `chain_id` as a tag in sentry logs from the ErrorBoundary. This means "handled" events will have the tag, and async events will not yet. Will add that support in a separate PR so that we can start instrumenting this sooner to debug some issues. This will allow us to filter by chain_id for events, and also see which chains are causing errors most often for different errors.

https://uniswapteam.slack.com/archives/C050USFTS5B/p1681254541015739

### QA (ie manual testing)
Manually create an error, and verify that it is sent to Sentry with a chainId as a tag.
Screen Shot 2023-04-12 at 2 27 04 PM

### Automated testing
- [x] Unit test (N/A)

### DIFF
--- a/src/components/ErrorBoundary/index.tsx
+++ b/src/components/ErrorBoundary/index.tsx
@@ -1,5 +1,6 @@
 import { Trans } from '@lingui/macro'
 import * as Sentry from '@sentry/react'
+import { useWeb3React } from '@web3-react/core'
 import { ButtonLight, SmallButtonPrimary } from 'components/Button'
 import { ChevronUpIcon } from 'nft/components/icons'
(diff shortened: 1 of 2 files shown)


zzmp (commented):
Have you tested this with an invalid chainId? I want to make sure that the tag is set before any crash occurs.

vm:
The current implementation means we'll need to handle sending chain_id on async events, but I'd like to ship this first so I can start instrumenting some of the revert issues on production.

zzmp (approved):
can you update the PR description to reflect the updated implementation?

cartcrom (approved):

PR #6346, opened 2023-04-12
fix: show empty state on /add and /remove

vm:
Fixes behavior on /add and /remove to show an empty state when on unsupported chains.

## Screen capture
| Before       | After (Desktop)|
|Screen Shot 2023-04-19 at 1 01 27 PM | Screen Shot 2023-04-19 at 1 02 11 PM|
|Screen Shot 2023-04-19 at 5 59 01 PM|Screen Shot 2023-04-19 at 5 59 49 PM|

### QA (ie manual testing)
1. Switch wallet to Sepolia
2. Go to app.uniswap.org/#/add
3. See the state in both the test preview and production

#### Devices
I tested this on mobile and desktop chrome.

### Automated testing
I do not think this is easy to test for the same reason as https://github.com/Uniswap/interface/pull/6338. Until we have cypress that can change chain ids, this is not easy to test automated. It's also a minimal change so manual testing here should be enough.
- [x] Unit test
- [x] Integration/E2E test


src/pages/RemoveLiquidity/V3.tsx
```diff
@@ -58,6 +61,10 @@ export default function RemoveLiquidityV3() {
     return <Navigate to={{ ...location, pathname: '/pools' }} replace />
   }
 
+  if (!isSupportedChain(chainId)) {
+    return <PositionPageUnsupportedContent />
+  }
+
   return <Remove tokenId={parsedTokenId} />
```
zzmp, line 68: Don't negate:

```suggestion
  if (isSupportedChain(chainId)) {
    return <Remove tokenId={parsedTokenId} />
  } else {
    return <PositionPageUnsupportedContent />
  }
```

zzmp (changes requested):

grabbou (approved):
LGTM

zzmp (approved):

PR #6348, opened 2023-04-12
fix: catch approvals

vm:
## Description
Catches all the approvals on the LP side of the codebase. Instead of throwing on user rejection, we just console.debug for now.

Probably need a better UI state for this, but for now maybe this is better than the existing experience.

_JIRA ticket:_ https://linear.app/uniswap/issue/WEB-2398/catch-user-rejection-errors


vm:
This does not seem like the best use of time ATM.

natew (approved):
just avoids errors in the ui? agree it would be better to send them to some sort of toast right? how would i trigger this to test? probably fine though overall so approving.

vm:
Well I think if there's a real error we should show an error state, but that's totally unhandled atm. If the user rejects the transaction I think going back to the previous state is fine, which is what this does.

To reproduce you can go to the LP pages and try to create a position, or remove liquidity. Then when the tx shows up in wallet just decline it. Previously, that'd throw an error and now it doesn't.

JFrankfurt:
can you give repro steps or a screenshot of the prior experience?

also, this feels like a good candidate for an e2e test

JFrankfurt (changes requested):
code looks good, but should add an e2e test to confirm intended behavior

PR #6350, opened 2023-04-12
feat: remove /wallet route

cartcrom:
## Description
Replaces /wallet route w/ links to the external microsite

Removes the wallet_microsite feature flag

Adds specific analytics for microsite opens vs app store opens

## Test plan
* Enable the mgtm feature flag
* Open the uniswap wallet connector modal
* Click download while on desktop, expect microsite
* Click download while on iOS, expect app store

### DIFF
--- a/src/components/AccountDrawer/UniwalletModal.tsx
+++ b/src/components/AccountDrawer/UniwalletModal.tsx
@@ -1,5 +1,6 @@
 import { Trans } from '@lingui/macro'
 import { sendAnalyticsEvent } from '@uniswap/analytics'
+import { InterfaceElementName } from '@uniswap/analytics-events'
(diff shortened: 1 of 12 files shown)


src/components/AccountDrawer/UniwalletModal.tsx
```diff
@@ -122,7 +122,7 @@ function InfoSection({ onClose }: { onClose: () => void }) {
         </ThemedText.Caption>
       </AutoColumn>
       <Column>
-        <DownloadButton onClick={onClose} />
```
cartcrom, line 125: Will H mentioned we should leave the modal open, rather than close it after download is clicked

lynnshaoyu (approved):

cmcewen (approved):

PR #6355, opened 2023-04-12
feat: updated banner

cartcrom:
## Description

_JIRA ticket:_
_Slack thread:_
_Relevant docs:_

## Screen capture

### Reproducing the error

1. 

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test


src/components/NavBar/MenuDropdown.tsx
```diff
@@ -140,24 +146,37 @@ export const MenuDropdown = () => {
                 <Box display={{ sm: 'none', lg: 'flex', xxl: 'none' }}>
                   <PrimaryMenuRow to="/pool" close={toggleOpen}>
                     <Icon>
-                      <PoolIcon width={24} height={24} color={theme.textSecondary} />
+                      <PoolIcon width={24} height={24} fill={theme.textPrimary} />
```
cartcrom, line 149: color was incorrect for these

src/components/NavBar/MenuDropdown.tsx
```diff
@@ -140,24 +146,37 @@ export const MenuDropdown = () => {
...
+                >
+                  <PrimaryMenuRow close={toggleOpen}>
+                    <Icon>
+                      <AppleLogo width="24px" height="24px" fill={theme.textPrimary} />
+                    </Icon>
+                    <PrimaryMenuRow.Text>
+                      <Trans>Download Uniswap Wallet</Trans>
+                    </PrimaryMenuRow.Text>
+                  </PrimaryMenuRow>
+                </Box>
```
cartcrom, line 168: Fred/lindsay wanted this download link added back here. Ideally this would be decoupled into a sep PR, but time crunch 🤯

src/components/NavBar/MenuDropdown.tsx
```diff
@@ -140,24 +146,37 @@ export const MenuDropdown = () => {
...
-                      <PoolIcon width={24} height={24} color={theme.textSecondary} />
+                      <PoolIcon width={24} height={24} fill={theme.textPrimary} />
                     </Icon>
                     <PrimaryMenuRow.Text>
                       <Trans>Pool</Trans>
                     </PrimaryMenuRow.Text>
                   </PrimaryMenuRow>
                 </Box>
+                <Box
+                  display={mgtmEnabled ? 'flex' : 'none'}
```
zzmp, line 157: Are we intentionally showing this for all devices, and not just for iOS devices? eg in the banner the language changes if !iOS to "Learn more" instead of "Download".

src/components/NavBar/MenuDropdown.tsx
```diff
@@ -140,24 +146,37 @@ export const MenuDropdown = () => {
...
                     <Icon>
-                      <PoolIcon width={24} height={24} color={theme.textSecondary} />
+                      <PoolIcon width={24} height={24} fill={theme.textPrimary} />
                     </Icon>
                     <PrimaryMenuRow.Text>
                       <Trans>Pool</Trans>
                     </PrimaryMenuRow.Text>
                   </PrimaryMenuRow>
                 </Box>
+                <Box
```
zzmp, line 156: Is there a reason this uses a custom `<Box>` instead of the `<DownloadButton>`?

src/components/Banner/UniswapWalletBanner.tsx
```diff
@@ -1,111 +1,142 @@
...
+  border-radius: 16px;
+  ${OpacityHoverState};
 `
 
 export default function UniswapWalletBanner() {
   const [hideUniswapWalletBanner, toggleHideUniswapWalletBanner] = useHideUniswapWalletBanner()
-  const [walletDrawerOpen] = useAccountDrawer()
   const mgtmEnabled = useMgtmEnabled()
+  const location = useLocation()
+  const isLandingScreen = location.search === '?intro=true' || location.pathname === '/'
```
zzmp, line 91: There's no existing hook for this?

src/components/Banner/UniswapWalletBanner.tsx
```diff
@@ -1,111 +1,142 @@
...
-            onClick={() => toggleHideUniswapWalletBanner()}
-            element={InterfaceElementName.UNISWAP_WALLET_BANNER_DOWNLOAD_BUTTON}
-          />
-        </ButtonRow>
-      </InnerContainer>
+      <StyledXButton
+        data-testid="uniswap-wallet-banner"
+        size={20}
+        onClick={(e) => {
+          e.preventDefault()
```
zzmp, line 101: Why is the prevent/stop necessary?

zzmp (changes requested):

src/components/Banner/UniswapWalletBanner.tsx
cartcrom: Not that I know of. I'm not sure why it was written in such a way that both the paths of "?intro=true" and "/" are the landing page, feels wrong to me. But this is the only piece of code that needs to detect the landing page it seems

src/components/Banner/UniswapWalletBanner.tsx
cartcrom: This was already there. I believe there were scenarios in the past where there was potentially something else underneath the advertisement. I.e. maybe you click the x button but don't want the click to propagate to the token row underneath if you're on /tokens

src/components/NavBar/MenuDropdown.tsx
cartcrom: I'm following this file's convention of using the display property of Box for responsiveness. <DownloadButton /> is visually very different than this inline overflow menu item.

src/components/NavBar/MenuDropdown.tsx
cartcrom: Yes, `openDownloadApp` opens the microsite if not on iOS, which is the desired functionality.

src/components/Banner/UniswapWalletBanner.tsx
zzmp: I think you should add a comment explaining this so people like me don't try to remove these method calls :)

src/components/NavBar/MenuDropdown.tsx
zzmp: This feels inconsistent. Could you check in with design on this?

cartcrom: They're in the loop on this functionality.

zzmp (changes requested):
When testing on an iPhone 13 Mini, "Download App" has a linebreak so that each word is on its own line. It still fits within the button. Is this acceptable?

cartcrom:
> When testing on an iPhone 13 Mini, "Download App" has a linebreak so that each word is on its own line. It still fits within the button. Is this acceptable?

I added repsonsiveness for "Download App" to shrink to just "Download" past a certain screen size, and also decreased padding on the buttons

src/components/NavBar/MenuDropdown.tsx
yyip-dev: +1 can confirm this is desired behavior

src/components/Banner/UniswapWalletBanner.tsx
yyip-dev: The `?intro=true` logic was specifically for when users click on the top left nav icon to bring back up the landing page (the expected behavior there is to pull up the landing page).

zzmp (approved):
lgtm - please ask design why we're not using the app store badge provided by apple 😕

yyip-dev (approved):
Responded to/added context to a few comments. Also tested on desktop + iOS and lgtm.

src/components/Banner/UniswapWalletBanner.tsx
```diff
@@ -122,9 +122,9 @@ export default function UniswapWalletBanner() {
               backgroundColor="white"
               onClick={() => openDownloadApp(InterfaceElementName.UNISWAP_WALLET_BANNER_DOWNLOAD_BUTTON)}
             >
-              <AppleLogo width={16} height={16} />
-              <ThemedText.LabelSmall color="black" marginLeft="4px">
-                {!screenSize['xs'] ? <Trans>Download</Trans> : <Trans>Download App</Trans>}
+              <AppleLogo width={14} height={14} />
+              <ThemedText.LabelSmall color="black" marginLeft="5px">
+                {!screenSize['xs'] ? <Trans>Download</Trans> : <Trans>Download app</Trans>}
```
zzmp, line 127: Should it also be "Learn more"?

zzmp (changes requested):

src/components/Banner/UniswapWalletBanner.tsx
cartcrom: yes ty

cmcewen (approved):

zzmp (approved):

PR #6356, opened 2023-04-12
fix: ignore error caused by Cloudflare HTML response

vm:
## Description
This is caused by HTML being returned for a chunk from Cloudflare. Usually, it's the result of a 499 exception right before it, which should be handled. Therefore, this can be ignored.

In the future, there will be a PR that handles the 499 exceptions via Cloudflare to ignore those exceptions in logging as well.

_JIRA ticket:_ https://uniswaplabs.atlassian.net/browse/INFRA-172
_Slack thread:_ https://uniswapteam.slack.com/archives/C050USFTS5B/p1681335647517689

### QA (ie manual testing)
This is not easily testable locally, so instead we can see if these errors are not showing up tomorrow in Setry.

### Automated testing
- [x] Unit test (added)

### DIFF
--- a/src/tracing/errors.test.ts
+++ b/src/tracing/errors.test.ts
@@ -24,4 +24,9 @@ describe('filterKnownErrors', () => {
     const originalException = new Error('user rejected transaction')
     expect(filterKnownErrors(ERROR, { originalException })).toBe(null)
   })
+
+  it('filters invalid HTML response errors', () => {
+    const originalException = new Error("Unexpected token '<'")
+    expect(filterKnownErrors(ERROR, { originalException })).toBe(null)
+  })
 })
--- a/src/tracing/errors.test.ts
+++ b/src/tracing/errors.test.ts
@@ -26,7 +26,7 @@ describe('filterKnownErrors', () => {
   })
 
   it('filters invalid HTML response errors', () => {
-    const originalException = new Error("Unexpected token '<'")
+    const originalException = new SyntaxError("Unexpected token '<'")
     expect(filterKnownErrors(ERROR, { originalException })).toBe(null)
   })
 })
--- a/src/tracing/errors.ts
+++ b/src/tracing/errors.ts
@@ -28,7 +28,7 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event: 
 
     // Cloudflare may serve HTML error pages (eg from a 499), which are already raised as ChunkLoadError.
     // Parsing the HTML results in a second SyntaxError. This not the root error (it is already reported as ChunkLoadError), so it should not be reported.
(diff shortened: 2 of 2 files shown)


src/tracing/errors.ts
```diff
@@ -25,6 +25,11 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event:
 
     // If the error is based on a user rejecting, it should not be considered an exception.
     if (didUserReject(error)) return null
+
+    // This is caused by HTML being returned for a chunk from Cloudflare.
+    // Usually, it's the result of a 499 exception right before it, which should be handled.
+    // Therefore, this can be ignored.
```
zzmp, line 31: If it's "Usually" then it shouldn't be filtered out, because it might hide other legitimate errors.

```suggestion
    // Cloudflare may serve HTML error pages (eg from a 499), which are already raised as ChunkLoadError.
    // Parsing the HTML results in a second SyntaxError. This not the root error (it is already reported as ChunkLoadError), so it should not be reported.
```

src/tracing/errors.ts
```diff
@@ -25,6 +25,11 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event:
...
+    if (error.message.match(/Unexpected token '<'/)) return null
```
zzmp, line 32: This should be a SyntaxError - I think you should be as specific as possible here:

```suggestion
    if (error instanceof SyntaxError && error.message.match(/Unexpected token '<'/)) return null
```

src/tracing/errors.test.ts
```diff
@@ -24,4 +24,9 @@ describe('filterKnownErrors', () => {
     const originalException = new Error('user rejected transaction')
     expect(filterKnownErrors(ERROR, { originalException })).toBe(null)
   })
+
+  it('filters invalid HTML response errors', () => {
+    const originalException = new Error("Unexpected token '<'")
```
zzmp, line 29:
```suggestion
    const originalException = new SyntaxError("Unexpected token '<'")
```

zzmp (changes requested):

src/tracing/errors.ts
```diff
@@ -25,6 +25,10 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event:
 
     // If the error is based on a user rejecting, it should not be considered an exception.
     if (didUserReject(error)) return null
+
+    // Cloudflare may serve HTML error pages (eg from a 499), which are already raised as ChunkLoadError.
```
lynnshaoyu, line 29: I think @zzmp wanted to standardize multiline developer docs with `/* some comment here */` instead of `// some comment here`

lynnshaoyu (approved):

zzmp (approved):

PR #6357, opened 2023-04-13
fix: incorrect line wrapping of text in menus in CJK

kecrily:
## Description

Fixes incorrect line wrapping of text in menus in CJK languages

## Screen capture

| Before       | After |
| image | image |
| image | image |
| image | image  |

### Reproducing the error

1. Open <https://app.uniswap.org/#/swap?lng=ja-JP>

### QA (ie manual testing)

- [x] N/A

#### Devices

Possibly all devices. I simulated these devices with Chrome Device Mode

### Automated testing

- [x] Unit test **N/A**
- [x] Integration/E2E test **N/A**

### DIFF
--- a/src/components/NavBar/style.css.ts
+++ b/src/components/NavBar/style.css.ts
@@ -22,8 +22,6 @@ export const logo = style([
 export const baseSideContainer = style([
   sprinkles({
     display: 'flex',
-    width: 'full',
-    flex: '1',
     flexShrink: '2',
   }),
 ])
@@ -38,7 +36,6 @@ export const leftSideContainer = style([
 
 export const searchContainer = style([
   sprinkles({
-    flex: '1',
     flexShrink: '1',
     justifyContent: { lg: 'flex-end', xl: 'center' },
     display: { sm: 'none', navSearchInputVisible: 'flex' },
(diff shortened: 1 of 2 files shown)


JFrankfurt (approved):

src/components/NavBar/style.css.ts
```diff
@@ -38,7 +36,6 @@ export const leftSideContainer = style([
 
 export const searchContainer = style([
```
cbachmeier, line 37: The searchbar is no longer centered in ENG or JPN
Screen Shot 2023-05-16 at 10 32 27
Screen Shot 2023-05-16 at 10 33 34

cbachmeier (changes requested):
Thanks for the PR Submission. Everything other than the centering of the searchbar from 1100px up looks good from my testing

src/components/NavBar/style.css.ts
cbachmeier: Compared to current production:
Screen Shot 2023-05-16 at 10 37 22

kecrily: @cbachmeier If the search is always centered then it will cause the left side to be squeezed, which is the case now. So I used `space-between` to balance the space between the left and the right.

cbachmeier: As a design choice, we will always want the searchbar to be centered on the swap modal and instead would prefer the text in the navbar to wrap.

cartcrom:

PR #6361, opened 2023-04-13
test: swap modal header unit test

lynnshaoyu:
unit test for SwapModalHeader.tsx.


src/components/swap/SwapModalHeader.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+        shouldLogModalCloseEvent={false}
+        showAcceptChanges={false}
+        setShouldLogModalCloseEvent={noop}
+        onAcceptChanges={noop}
+        recipient={testRecipientAddress}
+      />
+    )
+    expect(asFragment()).toMatchSnapshot()
+    expect(screen.getByText(/Output is estimated. You will receive at least /i)).toBeInTheDocument()
+    expect(screen.getByTestId('input-symbol')).toHaveTextContent('ABC')
```
cartcrom, line 47: Would it make more sense to use values from the input props here?

```suggestion
    expect(screen.getByTestId('input-symbol')).toHaveTextContent(testTradeExactInput. inputAmount.currency.symbol)
```

src/components/swap/SwapModalHeader.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+    expect(screen.getByTestId('output-symbol')).toHaveTextContent('DEF')
+    expect(screen.getByTestId('input-amount')).toHaveTextContent('0.000000000000001')
```
cartcrom, line 49: similarly, should we consider using the input amount instead of the hardcoded string?

lynnshaoyu: great idea, done

zzmp (approved):

cartcrom (approved):

PR #6364, opened 2023-04-13
build: utilize the node_modules/.cache

zzmp:
## Description
Caches build steps - locally and in test.yml, our testing GH Action - to greatly reduce runtimes on repeated actions.
For most builds (eg for PRs with no changes to the lockfile), this halves the CI e2e run time.

1. Adds caching to the following build tools (all to the standard `node_modules/.cache/<build_tool_name>/` directory:
  - eslint: configured in craco.config.js and the `yarn lint` script
  - jest: configured in craco.config.js
  - tsc (typescript): configured in tsconfig.json
 
2. Updates caching in the setup/action.yml and test.yml GitHub Action definitions:
  - Explicitly omits node_modules/.cache from setup/action.yml caching, as no build steps have occured
  - Adds node_modules/.cache to all jobs in test.yml, as each tool has its own cache
  - Fixes cypress binary caching to reduce amortized run time - it will now run as part of testing, to avoid a separate setup job - and pins cypress to avoid version churn
  - Also updates actions/cache to v3 to eliminate warnings from the outdated v2 usage

3. Cleans up craco.config.js, as long as its being modified to specify caching.

_Relevant docs:_ https://jongleberry.medium.com/speed-up-your-ci-and-dx-with-node-modules-cache-ac8df82b7bb0#:~:text=your%20entire%20node_modules-,node_modules%2F.,use%20this%20folder%20by%20default.

### QA (ie manual testing)

- [x] Run `yarn test`, `yarn lint`, `yarn typecheck`, and note the improvement for repeated runs.

### Automated testing

- [x] Unit test N/A
- [x] Integration/E2E test
  - Runs in test.yml GitHub Action

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -10,7 +10,9 @@ runs:
       with:
         node-version: 14
         registry-url: https://registry.npmjs.org
+        cache: yarn
 
+    # node_modules/.cache is intentionally omitted, as this is used for build tool caches.
     - uses: actions/cache@v3
       id: install-cache
       with:
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,4 +1,6 @@
 name: Setup
+description: checkout repo, setup node, and install node_modules
+
 runs:
   using: composite
   steps:
@@ -8,14 +10,14 @@ runs:
       with:
         node-version: 14
         registry-url: https://registry.npmjs.org
-        cache: yarn
 
     - uses: actions/cache@v3
       id: install-cache
       with:
-        path: node_modules/
+        path: |
+          node_modules
+          !node_modules/.cache
         key: ${{ runner.os }}-install-${{ hashFiles('**/yarn.lock') }}
-
     - if: steps.install-cache.outputs.cache-hit != 'true'
       run: yarn install --frozen-lockfile --ignore-scripts
       shell: bash
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,9 @@
 name: Test
 
+# Many build steps have their own caches, so each job has its own cache to improve subsequent build times.
+# Build tools are configured to cache cache to node_modules/.cache, so this is cached independently of node_modules.
+# See https://jongleberry.medium.com/speed-up-your-ci-and-dx-with-node-modules-cache-ac8df82b7bb0.
+
 on:
   push:
     branches:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,6 +14,12 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup
+      - uses: actions/cache@v3
+        id: eslint-cache
+        with:
+          path: node_modules/.cache
+          key: ${{ runner.os }}-eslint-${{ hashFiles('**/yarn.lock') }}-${{ github.run_id }}
+          restore-keys: ${{ runner.os }}-eslint-${{ hashFiles('**/yarn.lock') }}-
       - run: yarn lint
   
   typecheck:
@@ -21,6 +27,12 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup
+      - uses: actions/cache@v3
+        id: tsc-cache
+        with:
+          path: node_modules/.cache
+          key: ${{ runner.os }}-tsc-${{ hashFiles('**/yarn.lock') }}-${{ github.run_id }}
+          restore-keys: ${{ runner.os }}-tsc-${{ hashFiles('**/yarn.lock') }}-
       - run: yarn prepare
       - run: yarn typecheck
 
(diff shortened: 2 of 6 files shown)


tsconfig.json
```diff
@@ -23,6 +24,7 @@
     "strict": true,
     "strictNullChecks": true,
     "target": "es5",
+    "tsBuildInfoFile": "node_modules/.cache/.tsbuildinfo",
```
JFrankfurt, line 27: I looked up the docs for this setting and don't understand exactly why it's being used here in this way--help?

package.json
```diff
@@ -21,7 +21,7 @@
     "start": "craco start",
     "build": "craco build",
     "serve": "serve build -l 3000",
-    "lint": "yarn eslint --ignore-path .gitignore .",
+    "lint": "yarn eslint --ignore-path .gitignore --cache --cache-location node_modules/.cache/eslint/ .",
```
JFrankfurt, line 24: should lint/type cache files live outside node_modules?

JFrankfurt:
I'm confused by the eslint/ts/jest usage of `node_modules/.cache`

package.json
zzmp: `node_modules/.cache` is a community-driven standard for where to put tooling-related caches - see [find-cache-dir](https://github.com/sindresorhus/find-cache-dir).
We already use this, it's just not explicit yet. For example, if you log the eslint config from react-scripts, it is caching in `node_modules/.cache/.eslintcache`; if you check `node_modules/.cache`, you'll also see babel and terser caching there.
I don't want to use a new arbitrary folder for doing so, and have to manage upkeep for that as well (ie `.gitignore`) - I'd rather use an existing standard.

tsconfig.json
zzmp: `tsBuildInfoFile` is a file written by `tsc` when `incremental: true`, so that it can save compiler state to disk and use it to speed up future builds. This option just specifies where that file should be written.

package.json
zzmp, line 24: This should be changed to match the lint location set by react-scripts. I will make that explicit in craco.config.js so that maintenance is simpler.

JFrankfurt (approved):

.github/workflows/test.yml
```diff
@@ -14,13 +14,25 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup
+      - uses: actions/cache@v3
```
cmcewen, line 17: Could you add that blog post as a comment in this file somewhere? Useful as a reference if someone needs to touch these

.github/actions/setup/action.yml
```diff
@@ -8,14 +10,14 @@ runs:
       with:
         node-version: 14
         registry-url: https://registry.npmjs.org
-        cache: yarn
```
cmcewen, line 11: why are you removing this? it is probably an noop for us but it says it caches global packages data, not node_modules https://github.com/actions/setup-node#caching-global-packages-data

.github/workflows/test.yml
```diff
@@ -49,9 +67,15 @@ jobs:
...
+        id: build-cache
+        with:
+          path: node_modules/.cache
+          key: ${{ runner.os }}-build-${{ hashFiles('**/yarn.lock') }}-${{ github.run_id }}
+          restore-keys: ${{ runner.os }}-build-${{ hashFiles('**/yarn.lock') }}-
       - run: yarn prepare
       - run: yarn build
-      - uses: actions/upload-artifact@v2
+      - uses: actions/upload-artifact@v3
```
cmcewen, line 78: what changed here?

.github/workflows/test.yml
```diff
@@ -63,51 +87,38 @@ jobs:
...
-      - uses: actions/download-artifact@v2
+      - uses: actions/download-artifact@v3
         with:
           name: build
           path: build
       - run: yarn test:size
 
 
-  cypress-build:
+  cypress-test-matrix:
```
cmcewen, line 97: cypress-build was basically a dupe step?

craco.config.cjs
```diff
@@ -21,9 +23,22 @@ module.exports = {
       },
     },
   },
+  eslint: {
```
cmcewen, line 26: knew there had to be a better way 😄

package.json
```diff
@@ -100,7 +100,7 @@
     "@uniswap/eslint-config": "^1.1.1",
     "@vanilla-extract/babel-plugin": "^1.1.7",
     "@vanilla-extract/webpack-plugin": "^2.1.11",
-    "cypress": "^10.3.1",
+    "cypress": "10.3.1",
```
cmcewen, line 103: why did you lock this?

tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "downlevelIteration": true,
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
+    "incremental": true,
```
cmcewen, line 10: consider this too maybe: https://www.typescriptlang.org/tsconfig#composite

cmcewen (commented):
🐎. i would appreciate if some of the changes were mentioned in the description as it's not clear to me how they impact the caching addition

.github/actions/setup/action.yml
zzmp: couldn't find those docs, thought it was duplicating the node_modules cache.

.github/workflows/test.yml
zzmp: upload-artifact was logging warnings related to node v12, so I upgraded it to eliminate warnings.

.github/workflows/test.yml
zzmp: cypress-build was building the cypress binary and caching it, and cypress-test was using that cached binary.
however, cypress-build added setup time to the total e2e testing time. moving the build to the cypress-test step means it will run slower with a cache miss but faster overall because it can skip a whole job's setup.

package.json
cmcewen: could you document this somewhere in a followup or on notion?

PR #6365, opened 2023-04-13
fix: ignore ChunkLoadingErrors if related to 499 or missing data

vm:
## Description
This makes any browsers that support `responseStatus` on performance entries make sure that we ignore 499 related errors when loading chunks. This is an error code from Cloudflare that we want to ignore.
For any browsers that don't support this feature, we'll just ignore the error completely. While this may hide legitimate errors, for now it makes sense so we can hide false positives in logging.

_JIRA ticket:_ https://uniswaplabs.atlassian.net/browse/INFRA-172

### Automated testing
- [x] Unit test


src/tracing/errors.test.ts
```diff
@@ -29,4 +29,145 @@ describe('filterKnownErrors', () => {
...
+        return []
+      })
+      window.performance.getEntriesByType = mockGetEntriesByType
+      const originalException = new Error(
+        'Loading chunk 20 failed. (error: https://app.uniswap.org/static/js/20.d55382e0.chunk.js)'
+      )
+      expect(filterKnownErrors(ERROR, { originalException })).toBeNull()
+    })
+
+    it('filters out error when performance is undefined', () => {
```
grabbou, line 125: nit: is this unit test title correct?

vm, line 125: gonna remove this test

grabbou (approved):
LGTM! +1 for comments and making it super explicit!

src/tracing/errors.ts
```diff
@@ -40,6 +40,17 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event:
     // If the error is based on a user rejecting, it should not be considered an exception.
     if (didUserReject(error)) return null
 
+    if (error.message.match(/Loading chunk \d+ failed\. \(error: .+\.chunk\.js\)/)) {
```
grabbou, line 43: nit: could you preface this `if` condition with a comment explaining why this error happens and why it's okay to filter out `499`? Is it false positive on Cloudflare, but application still works OK?

src/tracing/errors.test.ts
vm: sorry i just fixed it

grabbou (approved):

src/tracing/errors.test.ts
```diff
@@ -29,4 +29,145 @@ describe('filterKnownErrors', () => {
...
+        ] as any
+      })
+      window.performance.getEntriesByType = mockGetEntriesByType
+      const originalException = new Error(
+        'Loading chunk 20 failed. (error: https://app.uniswap.org/static/js/20.d55382e0.chunk.js)'
+      )
+      expect(filterKnownErrors(ERROR, { originalException })).toBeNull()
+    })
+
+    it('keeps 200 error coded chunk error', () => {
```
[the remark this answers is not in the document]
vm: true i will change

src/tracing/errors.ts
vm: yes i'll do

src/tracing/errors.ts
```diff
@@ -40,6 +40,17 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event:
...
+      const asset = error.message.match(/https?:\/\/.+?\.chunk\.js/)?.[0]
+      const resource = [...(performance?.getEntriesByType('resource') ?? [])].find(({ name }) => name === asset)
```
just-toby, line 45: woah, cool usage of `window.performance`

just-toby (approved):

src/tracing/errors.ts
```diff
@@ -40,6 +40,21 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event:
     // If the error is based on a user rejecting, it should not be considered an exception.
     if (didUserReject(error)) return null
 
+    /*
+     * This ignores 499 errors, which are caused by Cloudflare when a request is cancelled.
+     * CF claims that some number of these is expected, and that they should be ignored.
```
zzmp, line 45:
```suggestion
     * CF claims that some number of these is expected, and that they should be ignored.
     * See https://groups.google.com/a/uniswap.org/g/cloudflare-eng/c/t3xvAiJFujY.
```

src/tracing/errors.ts
```diff
@@ -40,6 +40,21 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event:
...
+     */
+    if (error.message.match(/Loading chunk \d+ failed\. \(error: .+\.chunk\.js\)/)) {
```
zzmp, line 47: This will only catch js chunk errors. You should catch all errors (eg css, etc).

```suggestion
    if (error.message.match(/Loading chunk \d+ failed\. \(error: .+\.chunk\.\w{2+}\)/)) {
```

src/tracing/errors.ts
```diff
@@ -40,6 +40,21 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event:
...
+      const asset = error.message.match(/https?:\/\/.+?\.chunk\.js/)?.[0]
+      const resource = [...(performance?.getEntriesByType('resource') ?? [])].find(({ name }) => name === asset)
+      // `responseStatus` is not on the `ResponseStatus` type, because it's only supported on some browsers at the moment.
+      const status = (resource as any)?.responseStatus
+      // If the status if 499, then we ignore. If there's no status, for now we'll ignore all chunk loading errors also.
```
zzmp, line 52: Why are we ignoring unknown errors? Why not _just_ 499s?

src/tracing/errors.ts
```diff
@@ -40,6 +40,21 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event:
     // If the error is based on a user rejecting, it should not be considered an exception.
     if (didUserReject(error)) return null
 
+    /*
+     * This ignores 499 errors, which are caused by Cloudflare when a request is cancelled.
+     * CF claims that some number of these is expected, and that they should be ignored.
+     */
+    if (error.message.match(/Loading chunk \d+ failed\. \(error: .+\.chunk\.js\)/)) {
+      const asset = error.message.match(/https?:\/\/.+?\.chunk\.js/)?.[0]
+      const resource = [...(performance?.getEntriesByType('resource') ?? [])].find(({ name }) => name === asset)
```
zzmp, line 49: I'd prefer you cast the returned type, instead of casting it on L51.
Instead of casting to any, can you do a narrower cast? eg just add the property you need.

src/tracing/errors.test.ts
```diff
@@ -29,4 +29,141 @@ describe('filterKnownErrors', () => {
...
+            secureConnectionStart: 254.19999998807907,
+            connectEnd: 254.19999998807907,
+            requestStart: 255.09999999403954,
+            responseStart: 256,
+            responseEnd: 256.8999999910593,
+            transferSize: 0,
+            encodedBodySize: 227688,
+            decodedBodySize: 227688,
+            responseStatus: 499,
+            serverTiming: [],
```
zzmp, line 65: Only mock what you need:

```suggestion
            name: 'https://app.uniswap.org/static/js/20.d55382e0.chunk.js',
            responseStatus: 499,
```

src/tracing/errors.test.ts
```diff
@@ -29,4 +29,141 @@ describe('filterKnownErrors', () => {
...
+            secureConnectionStart: 254.19999998807907,
+            connectEnd: 254.19999998807907,
+            requestStart: 255.09999999403954,
+            responseStart: 256,
+            responseEnd: 256.8999999910593,
+            transferSize: 0,
+            encodedBodySize: 227688,
+            decodedBodySize: 227688,
+            responseStatus: 200,
+            serverTiming: [],
```
zzmp, line 103:
```suggestion
            name: 'https://app.uniswap.org/static/js/20.d55382e0.chunk.js',
            responseStatus: 200,
```

src/tracing/errors.test.ts
```diff
@@ -29,4 +29,141 @@ describe('filterKnownErrors', () => {
...
+          },
+        ] as any
```
zzmp, line 105: shouldn't this be typed? why `any`? same for preceding test.

src/tracing/errors.test.ts
```diff
@@ -29,4 +29,141 @@ describe('filterKnownErrors', () => {
...
+      })
+      window.performance.getEntriesByType = mockGetEntriesByType
+      const originalException = new Error(
+        'Loading chunk 20 failed. (error: https://app.uniswap.org/static/js/20.d55382e0.chunk.js)'
+      )
+      expect(filterKnownErrors(ERROR, { originalException })).toBeNull()
+    })
+
+    it('keeps error when status is different than 499', () => {
+      const mockGetEntriesByType = jest.fn(() => {
```
zzmp, line 77: Instead of defining a mock and then assigning it, do it in one line. It's easier to read / keeps it all in one place:

```
jest.mock(window.performance, 'getEntriesByType').mockResolves([...])
```

zzmp (changes requested):

src/tracing/errors.ts
vm: That won't actually be sufficient because the error message is slightly different for example. So I'd like to leave that for a separate PR.

src/tracing/errors.ts
grabbou: Also related to my comment https://github.com/Uniswap/interface/pull/6365#discussion_r1172885316

vm: we're ignoring them bc responseStatus is only supported on Chrome. otherwise we would be logging ChunkErrors from other browsers that are likely just 499s.

src/tracing/errors.test.ts
vm: I tried doing this but getEntriesByType isn't defined in this execution env so that won't work properly

vm: i did it on one line now instead tho

src/tracing/errors.test.ts
```diff
@@ -2,6 +2,8 @@ import { ErrorEvent } from '@sentry/types'
 
 import { filterKnownErrors } from './errors'
 
+jest.mock('./errors')
```
zzmp, line 5: whoops this should go away too

zzmp (approved):

src/tracing/errors.test.ts
```diff
@@ -2,6 +2,11 @@ import { ErrorEvent } from '@sentry/types'
 
 import { filterKnownErrors } from './errors'
 
+Object.defineProperty(window.performance, 'getEntriesByType', {
+  writable: true,
+  value: jest.fn(),
+})
+
```
zzmp, line 9: This is expected on the DOM. Could it be put in `setupTests.ts` instead?

zzmp (changes requested):

src/tracing/errors.test.ts
vm: we're gonna talk through this separately

PR #6366, opened 2023-04-13
ci: GH action for promoting staging -> prod

lavalamp-:
## Description

This PR introduces a Github action that will cut a PR to merge the HEAD of the `releases/staging` branch to the `releases/prod` branch. This PR...

- Will bring `releases/prod` into sync with `releases/staging`
- Is configured to automatically merge once sufficient approvals have been given
- Merges using the `merge` strategy, which maintains initial commit hashes (this is necessary to generate future PRs that bring the two branches into sync)

Through this configuration the merge is performed by the associated service account which enables us to use branch protection rules limiting the users that can push to the `releases/prod` branch.

### QA (ie manual testing)

As the service account token necessary to test this is only available in the `main` branch we'll need to wait until this code gets merged to `main` to test it. That said, a close variation of this code is already successfully humming along in another repository.

### DIFF
--- a/.github/workflows/pr-staging-to-prod.yml
+++ b/.github/workflows/pr-staging-to-prod.yml
@@ -1,0 +1,61 @@
+name: 3 | Generate PR for releases/staging to releases/prod
+
+# This CI job is responsible for generating PRs that bring the HEAD of releases/staging into releases/prod.
+# These PRs are meant to be the only (standard) way that code is merged into the releases/prod branch.
+#
+# https://github.com/peter-evans/create-pull-request/blob/main/docs/examples.md#keep-a-branch-up-to-date-with-another
+
+on:
+  workflow_dispatch:
+
+jobs:
+  prod-gen-pr:
+    name: 'Generate PR for merging to releases/prod branch'
+    runs-on: ubuntu-latest
+    environment:
+      name: release
+    steps:
+      - uses: actions/checkout@7884fcad6b5d53d10323aee724dc68d8b9096a2e
+        with:
+          token: ${{ secrets.RELEASE_SERVICE_ACCESS_TOKEN }}
+          ref: main
+      - uses: actions/checkout@7884fcad6b5d53d10323aee724dc68d8b9096a2e
+        with:
+          token: ${{ secrets.RELEASE_SERVICE_ACCESS_TOKEN }}
+          ref: releases/prod
+      - name: Reset promotion branch
+        run: |
+          git fetch origin releases/staging:releases/staging
+          git reset --hard releases/staging
+      - name: Setup git
+        run: |
+          git config user.name "UL Mobile Service Account"
+          git config user.email "hello-happy-puppy@users.noreply.github.com"
+      - uses: peter-evans/create-pull-request@ea54357f43e3d1cf1125471d0814f4d02cc0d364
+        id: cpr
+        with:
+          token: ${{ secrets.RELEASE_SERVICE_ACCESS_TOKEN }}
+          commit-message: "This PR merges latest commits of `releases/staging` to `releases/prod`. Upon approval and successful merge, a Github action associated with the prod Fastlane setup will deploy to TestFlight, where the app can be submitted to Apple."
+          base: 'releases/prod'
+          title: 'ci: Release Staging to Prod'
+          delete-branch: 'true'
+          committer: 'UL Service Account <hello-happy-puppy@users.noreply.github.com>'
+          author: 'UL Service Account <hello-happy-puppy@users.noreply.github.com>'
+          branch: 'approvals/stagingToProd'
+      - uses: peter-evans/enable-pull-request-automerge@60812ab1c2c6c6a8932b4d6e059becafaf386256
+        with:
+          token: ${{ secrets.RELEASE_SERVICE_ACCESS_TOKEN }}
+          pull-request-number: ${{ steps.cpr.outputs.pull-request-number }}
+          merge-method: merge
+      - name: Update PR body
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          echo "### Description" > /tmp/pr_desc
+          echo "" >> /tmp/pr_desc
+          echo "This PR promotes the following commits from releases/staging to prod." >> /tmp/pr_desc
+          echo "" >> /tmp/pr_desc
+          gh pr view ${{ steps.cpr.outputs.pull-request-number }} --json commits | jq '.commits[] | [.oid, .messageHeadline] | @tsv' | sed 's/"//g' | sed 's/\\t/ - /g' >> /tmp/pr_desc
+          echo "" >> /tmp/pr_desc
+          echo "**Once approved this PR will automatically merge via a merge commit.**" >> /tmp/pr_desc
+          gh pr edit ${{ steps.cpr.outputs.pull-request-number }} -b "$(cat /tmp/pr_desc)"


.github/workflows/pr-staging-to-prod.yml
```diff
@@ -0,0 +1,61 @@
+name: 3 | Generate PR for releases/staging to releases/prod
+
+# This CI job is responsible for generating PRs that bring the HEAD of releases/staging into releases/prod.
```
zzmp, line 3:
```suggestion
# This CI job is responsible for generating PRs that bring the HEAD of `releases/staging` into `releases/prod`.
```

.github/workflows/pr-staging-to-prod.yml
```diff
@@ -0,0 +1,61 @@
...
+# These PRs are meant to be the only (standard) way that code is merged into the releases/prod branch.
```
zzmp, line 4:
```suggestion
# These PRs are meant to be the only (standard) way that code is merged into the `releases/prod` branch.
```

.github/workflows/pr-staging-to-prod.yml
```diff
@@ -0,0 +1,61 @@
...
+  workflow_dispatch:
+
+jobs:
+  prod-gen-pr:
+    name: 'Generate PR for merging to releases/prod branch'
+    runs-on: ubuntu-latest
+    environment:
+      name: release
+    steps:
+      - uses: actions/checkout@7884fcad6b5d53d10323aee724dc68d8b9096a2e
```
zzmp, line 18: Why not use a newer version of checkout?

.github/workflows/pr-staging-to-prod.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Reset promotion branch
+        run: |
+          git fetch origin releases/staging:releases/staging
+          git reset --hard releases/staging
+      - name: Setup git
+        run: |
+          git config user.name "UL Mobile Service Account"
+          git config user.email "hello-happy-puppy@users.noreply.github.com"
+      - uses: peter-evans/create-pull-request@ea54357f43e3d1cf1125471d0814f4d02cc0d364
+        id: cpr
```
zzmp, line 35:
```suggestion
        id: create-pull-request
```

.github/workflows/pr-staging-to-prod.yml
```diff
@@ -0,0 +1,61 @@
...
+        with:
+          token: ${{ secrets.RELEASE_SERVICE_ACCESS_TOKEN }}
+          commit-message: "This PR merges latest commits of `releases/staging` to `releases/prod`. Upon approval and successful merge, a Github action associated with the prod Fastlane setup will deploy to TestFlight, where the app can be submitted to Apple."
+          base: 'releases/prod'
+          title: 'ci: Release Staging to Prod'
```
zzmp, line 40:
```suggestion
          title: 'ci: promotes staging to prod'
```

.github/workflows/pr-staging-to-prod.yml
```diff
@@ -0,0 +1,61 @@
...
+          git reset --hard releases/staging
+      - name: Setup git
+        run: |
+          git config user.name "UL Mobile Service Account"
+          git config user.email "hello-happy-puppy@users.noreply.github.com"
+      - uses: peter-evans/create-pull-request@ea54357f43e3d1cf1125471d0814f4d02cc0d364
+        id: cpr
+        with:
+          token: ${{ secrets.RELEASE_SERVICE_ACCESS_TOKEN }}
+          commit-message: "This PR merges latest commits of `releases/staging` to `releases/prod`. Upon approval and successful merge, a Github action associated with the prod Fastlane setup will deploy to TestFlight, where the app can be submitted to Apple."
```
zzmp, line 38:
```suggestion
          commit-message: "Merges latest commits of `releases/staging` into `releases/prod`."
```

You should update this to reflect web, not mobile.

.github/workflows/pr-staging-to-prod.yml
```diff
@@ -0,0 +1,61 @@
...
+          base: 'releases/prod'
+          title: 'ci: Release Staging to Prod'
+          delete-branch: 'true'
+          committer: 'UL Service Account <hello-happy-puppy@users.noreply.github.com>'
+          author: 'UL Service Account <hello-happy-puppy@users.noreply.github.com>'
+          branch: 'approvals/stagingToProd'
```
zzmp, line 44:
```suggestion
          branch: 'approvals/staging-to-prod'
```

.github/workflows/pr-staging-to-prod.yml
```diff
@@ -0,0 +1,61 @@
...
+      - uses: peter-evans/enable-pull-request-automerge@60812ab1c2c6c6a8932b4d6e059becafaf386256
+        with:
+          token: ${{ secrets.RELEASE_SERVICE_ACCESS_TOKEN }}
+          pull-request-number: ${{ steps.cpr.outputs.pull-request-number }}
+          merge-method: merge
```
zzmp, line 49: Prefer to use the GitHub CLI - I'd expect it to have a higher level of trust. See https://github.com/peter-evans/enable-pull-request-automerge.

.github/workflows/pr-staging-to-prod.yml
```diff
@@ -0,0 +1,61 @@
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          echo "### Description" > /tmp/pr_desc
+          echo "" >> /tmp/pr_desc
+          echo "This PR promotes the following commits from releases/staging to prod." >> /tmp/pr_desc
+          echo "" >> /tmp/pr_desc
+          gh pr view ${{ steps.cpr.outputs.pull-request-number }} --json commits | jq '.commits[] | [.oid, .messageHeadline] | @tsv' | sed 's/"//g' | sed 's/\\t/ - /g' >> /tmp/pr_desc
+          echo "" >> /tmp/pr_desc
+          echo "**Once approved this PR will automatically merge via a merge commit.**" >> /tmp/pr_desc
```
zzmp, line 60:
```suggestion
          echo "**Once approved this PR will be automatically merge via a merge commit.**" >> /tmp/pr_desc
```

.github/workflows/pr-staging-to-prod.yml
```diff
@@ -0,0 +1,61 @@
...
+          token: ${{ secrets.RELEASE_SERVICE_ACCESS_TOKEN }}
+          pull-request-number: ${{ steps.cpr.outputs.pull-request-number }}
+          merge-method: merge
+      - name: Update PR body
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          echo "### Description" > /tmp/pr_desc
+          echo "" >> /tmp/pr_desc
+          echo "This PR promotes the following commits from releases/staging to prod." >> /tmp/pr_desc
```
zzmp, line 56:
```suggestion
          echo "This PR promotes the following commits from `releases/staging` to `releases/prod`." >> /tmp/pr_desc
```

.github/workflows/pr-staging-to-prod.yml
zzmp: Can this be left blank, considering that it is updated later?

zzmp (changes requested):

.github/workflows/pr-staging-to-prod.yml
grabbou: Didn't you want to pin actions to specific commits before? However, in order to stay aligned with the rest, I would discuss whether pinning to `v3` or similar isn't more appropriate here

.github/workflows/pr-staging-to-prod.yml
lavalamp-: Oh nice find! Will update accordingly.

.github/workflows/pr-staging-to-prod.yml
lavalamp-: All of the pinned versions are copied from the `mobile` repository. I could go either way on whether we want to pin to a newer commit hash. On one hand (1) newer versions are likely better but on the other hand (2) having more versions of the same action running in different UL repositories means there's more risk that we've pinned to a vulnerable version.

My $0.02 would be that it's fine to pin to a newer commit hash _but we don't want to just use a tag like `v3`_. Defer to your judgment here though.

grabbou: Makes sense, and overall I agree. I don't know whether we actively tack this right now, I remember there was a JIRA ticket, but one day I was working on pinning all the actions too. I think that's good move.

lavalamp-: All of the pinned versions are copied from the `mobile` repository. I could go either way on whether we want to pin to a newer commit hash. On one hand (1) newer versions are likely better but on the other hand (2) having more versions of the same action running in different UL repositories means there's more risk that we've pinned to a vulnerable version.

My $0.02 would be that it's fine to pin to a newer commit hash _but we don't want to just use a tag like `v3`_. Defer to your judgment here though.

.github/workflows/pr-staging-to-prod.yml
lavalamp-: Needed to update this in a few places so not committing this suggestion but have made the changes in a separate commit.

.github/workflows/pr-staging-to-prod.yml
lavalamp-: Yeah I think so - deleting for now.

.github/workflows/pr-staging-to-prod.yml
lavalamp-: Sounds good - I'll pin to a newer version commit hash accordingly.

.github/workflows/pr-staging-to-prod.yml
lavalamp-: Not entirely grammatically correct - will make a similar grammatically correct commit to address.

.github/workflows/pr-staging-to-prod.yml
```diff
@@ -0,0 +1,64 @@
+name: 3 | Generate PR for releases/staging to releases/prod
```
zzmp, line 1: rename the file to match: 3-staging-to-prod.yml

.github/workflows/pr-staging-to-prod.yml
```diff
@@ -0,0 +1,64 @@
...
+          echo "This PR promotes the following commits from `releases/staging` to `releases/prod`." >> /tmp/pr_desc
+          echo "" >> /tmp/pr_desc
+          gh pr view ${{ steps.create-pull-request.outputs.pull-request-number }} --json commits | jq '.commits[] | [.oid, .messageHeadline] | @tsv' | sed 's/"//g' | sed 's/\\t/ - /g' >> /tmp/pr_desc
+          echo "" >> /tmp/pr_desc
+          echo "**Once approved this PR will be automatically merged via a merge commit.**" >> /tmp/pr_desc
+          gh pr edit ${{ steps.create-pull-request.outputs.pull-request-number }} -b "$(cat /tmp/pr_desc)"
+      - name: Enable PR automerge
+        env:
+          GH_TOKEN: ${{ secrets.RELEASE_SERVICE_ACCESS_TOKEN }}
+        run: gh pr merge --delete-branch --merge --auto "${{ steps.create-pull-request.outputs.pull-request-number }}"
```
zzmp, line 64: add newline

.github/workflows/pr-staging-to-prod.yml
```diff
@@ -0,0 +1,64 @@
+name: 3 | Generate PR for releases/staging to releases/prod
+
+# This CI job is responsible for generating PRs that bring the HEAD of `releases/staging` into `releases/prod`.
+# These PRs are meant to be the only (standard) way that code is merged into the `releases/prod` branch.
+#
+# https://github.com/peter-evans/create-pull-request/blob/main/docs/examples.md#keep-a-branch-up-to-date-with-another
+
+on:
+  workflow_dispatch:
+
```
zzmp, line 10:
```suggestion

on:
  workflow_dispatch:

# https://github.com/peter-evans/create-pull-request/blob/main/docs/examples.md#keep-a-branch-up-to-date-with-another
```

.github/workflows/pr-staging-to-prod.yml
```diff
@@ -0,0 +1,64 @@
...
+          git reset --hard releases/staging
+      - name: Setup git
+        run: |
+          git config user.name "UL Mobile Service Account"
+          git config user.email "hello-happy-puppy@users.noreply.github.com"
+      - name: Add CODEOWNERS file
+        run: |
+          echo "@uniswap/web-reviewers" > CODEOWNERS
+          git add CODEOWNERS
+          git commit -m "ci: add CODEOWNERS file"
```
zzmp, line 38: Couldn't we just add this file to the repo, and not have to do it as part of a CI step?

zzmp (changes requested):

.github/workflows/pr-staging-to-prod.yml
lavalamp-: The file needs to be added only for a single branch unfortunately :( If we added this to `main` then the `@uniswap/web-reviewers` group would be required to approve every PR targeting `main`.

Not a lot of good documentation on it but I found this instructive - https://stackoverflow.com/questions/75380368/github-codeowners-for-each-branch

.github/workflows/3-staging-to-prod.yml
```diff
@@ -0,0 +1,64 @@
...
+          echo "" >> /tmp/pr_desc
+          gh pr view ${{ steps.create-pull-request.outputs.pull-request-number }} --json commits | jq '.commits[] | [.oid, .messageHeadline] | @tsv' | sed 's/"//g' | sed 's/\\t/ - /g' >> /tmp/pr_desc
+          echo "" >> /tmp/pr_desc
+          echo "**Once approved this PR will be automatically merged via a merge commit.**" >> /tmp/pr_desc
+          gh pr edit ${{ steps.create-pull-request.outputs.pull-request-number }} -b "$(cat /tmp/pr_desc)"
+      - name: Enable PR automerge
+        env:
+          GH_TOKEN: ${{ secrets.RELEASE_SERVICE_ACCESS_TOKEN }}
+        run: gh pr merge --delete-branch --merge --auto "${{ steps.create-pull-request.outputs.pull-request-number }}"
+
```
zzmp, line 64: add newline

.github/workflows/pr-staging-to-prod.yml
zzmp: We already require that `@Uniswap/web-reviewers` approve every PR targeting `main`, so I think that would be fine. They own the codebase, so they should make all of the PR _and_ deploy approvals.

We can also segment off sub-sections with nested CODEOWNERS

zzmp (changes requested):
just held up on CODEOWNERS

zzmp (approved):

grabbou (approved):

PR #6367, opened 2023-04-14
chore: add constants for testing

lynnshaoyu:
add a file of shared constants to test-utils. 

also resolves this: https://uniswaplabs.atlassian.net/browse/WEB-3163


JFrankfurt (approved):

src/utils/noop.ts
```diff
@@ -0,0 +1,3 @@
+export default function noop() {
```
zzmp, line 1: why null instead of undefined or void?
I would expect this to return void:

```
export default function noop(): void {
  // Intentionally left empty.
}
```

src/test-utils/constants.tsx
```diff
@@ -0,0 +1,29 @@
+import { CurrencyAmount, Token } from '@uniswap/sdk-core'
+import { FeeAmount, Pool } from '@uniswap/v3-sdk'
+import JSBI from 'jsbi'
+
+export const testToken1 = new Token(1, '0x0000000000000000000000000000000000000001', 18)
```
zzmp, line 5: nit: I prefer constants to be in caps (eg all of the test tokens and pools):

```suggestion
export const TEST_TOKEN_1 = new Token(1, '0x0000000000000000000000000000000000000001', 18)
```

src/test-utils/constants.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const testToken2 = new Token(1, '0x0000000000000000000000000000000000000002', 18)
+export const testToken3 = new Token(1, '0x0000000000000000000000000000000000000003', 18)
+export const testRecipientAddress = '0x0fF2D1eFd7A57B7562b2bf27F3f37899dB27F4a5'
```
zzmp, line 8: Why is this a real address? Should it not be something like:

```suggestion
export const testRecipientAddress = '0x0000000000000000000000000000000000000004'
```

src/test-utils/constants.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const testPool13 = new Pool(
+  testToken1,
+  testToken3,
+  FeeAmount.MEDIUM,
+  '2437312313659959819381354528',
+  '10272714736694327408',
+  -69633
+)
+
+export const testCurrencyAmount = (token: Token, amount: number) =>
```
zzmp, line 28: More descriptive:

```suggestion
export const toCurrencyAmount = (token: Token, amount: number) =>
```

zzmp, line 28: I'd expect a matching `toEthAmount = (amount: number) =>`, or for `toCurrencyAmount` to take `Token | 'NATIVE'` or something.

src/test-utils/constants.tsx
```diff
@@ -0,0 +1,29 @@
+import { CurrencyAmount, Token } from '@uniswap/sdk-core'
```
zzmp, line 1: These are all token-related testing constants. Would you consider naming this file `src/test-utils/tokens.ts`?

zzmp: Definitely `ts`, not `tsx`.

src/test-utils/constants.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { FeeAmount, Pool } from '@uniswap/v3-sdk'
+import JSBI from 'jsbi'
+
+export const testToken1 = new Token(1, '0x0000000000000000000000000000000000000001', 18)
+export const testToken2 = new Token(1, '0x0000000000000000000000000000000000000002', 18)
+export const testToken3 = new Token(1, '0x0000000000000000000000000000000000000003', 18)
```
cartcrom, line 7: for these to be fully-reusable in other future tests, would it help for these to have names and symbols?

cartcrom: i.e. snapshots of components that render token symbols / names

src/utils/noop.ts
zzmp: If it needs to be `null` to satisfy typing, just leave that as a doc-comment:

```
/** No-op function. Returns `null` to satisfy most React typings. */
```

In the future maybe there can be other noops that return `undefined` or `void`, but for right now this looks great.

zzmp (changes requested):

cartcrom (approved):

src/test-utils/constants.tsx
lynnshaoyu: good call, I add symbols in later PRs (the actual test prs) but makes sense to add them all here, done.

src/test-utils/constants.tsx
lynnshaoyu: can add that if / when it is used in a test.

zzmp (approved):

PR #6375, opened 2023-04-14
build: tighten tsconfig

zzmp:
## Description

- Extends cypress tsconfig from base tsconfig
- Omits node_modules from watched types

### Automated testing

- [x] Unit test N/A

### DIFF
--- a/cypress/tsconfig.json
+++ b/cypress/tsconfig.json
@@ -1,8 +1,6 @@
 {
+  "extends": "../tsconfig.json",
   "compilerOptions": {
-    "strict": true,
-    "target": "es5",
-    "lib": ["es5", "dom"],
     "types": ["cypress"]
   },
   "include": ["**/*.ts"]
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -26,6 +26,9 @@
     "types": ["jest"],
     "useUnknownInCatchVariables": false
   },
-  "exclude": ["node_modules", "cypress"],
-  "include": ["src/**/*"]
+  "include": ["src/**/*"],
+  "watchOptions": {
+    "excludeDirectories": ["node_modules"],
+    "excludeFiles": ["node_modules/**"]
+  }
 }


tsconfig.json
```diff
@@ -26,6 +26,9 @@
     "types": ["jest"],
     "useUnknownInCatchVariables": false
   },
-  "exclude": ["node_modules", "cypress"],
-  "include": ["src/**/*"]
+  "include": ["src/**/*"],
+  "watchOptions": {
+    "excludeDirectories": ["node_modules"],
```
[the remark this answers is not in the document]
zzmp: It should only crawl imports to get this, so I think it's fine to omit

package.json
```diff
@@ -22,7 +22,7 @@
     "build": "craco build",
     "serve": "serve build -l 3000",
     "lint": "yarn eslint --ignore-path .gitignore --cache --cache-location node_modules/.cache/eslint/ .",
-    "typecheck": "tsc --noEmit",
+    "typecheck": "tsc",
```
grabbou, line 25: Is this intended? 
> The noEmit option tells TypeScript that we only want to run type checking and do not want the compiler to output any transpiled code

tsconfig.json
```diff
@@ -24,10 +24,13 @@
     "strict": true,
     "strictNullChecks": true,
     "target": "ESNext",
-    "tsBuildInfoFile": "node_modules/.cache/.tsbuildinfo",
+    "tsBuildInfoFile": "node_modules/.cache/tsbuildinfo/build",
```
grabbou, line 27: Was this not working previously?

cypress/tsconfig.json
```diff
@@ -1,9 +1,17 @@
...
+    "esModuleInterop": true,
+    "incremental": true,
+    "lib": ["DOM", "DOM.Iterable", "ESNext"],
+    "noEmit": true,
     "strict": true,
-    "target": "es5",
-    "lib": ["es5", "dom"],
-    "types": ["cypress"]
+    "target": "ES5",
+    "tsBuildInfoFile": "../node_modules/.cache/tsbuildinfo/cypress",
```
grabbou, line 9: I think the PR description is a bit misleading. This is not "extending" TSConfig from the Cypress config. It would be great if you documented the use case of `tsBuildInfoFile` in regards to that configuration.

grabbou (changes requested):

package.json
zzmp: Yes! noEmit is included in our tsconfig.

tsconfig.json
zzmp: this _was_ working, but is clobbered by the cypress/tsconfig.json value. Now, they'll both have their own caches :)

grabbou (approved):
LGTM. I would recommend slightly amending the PR title from `tighten` to be more specific as to what is this PR doing (build cache)

yyip-dev (approved):

PR #6379, opened 2023-04-17
build: target ESNext for dev

zzmp:
## Description

Updates tsconfig to target ESNext. This allows babel to reduce transforms for dev builds.
Because ESNext is less restrictive than our browserslist, this results in babel respecting the browserslist configuration, which reduces compilation time, especially for `yarn start`.
Effectively, this eliminates many transforms for dev builds, but still requires them for production builds.

_NOTE: This will make our production build use our browserslist config instead of using es5, which is more modern, and may cause very old browsers to no longer work with uniswap. This is further detailed in the attached doc._

_Relevant docs:_ https://www.notion.so/uniswaplabs/Ship-modern-JavaScript-code-instead-of-ES5-depending-on-current-browser-market-share-add1532f37184554a067c2c1c7c0d62c

### QA (ie manual testing)

- [x] N/A

### Automated testing

- [x] Unit test N/A

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,5 @@
+*.config.ts
+*.d.ts
+/src/graphql/data/__generated__/types-and-hooks.ts
+/src/graphql/thegraph/__generated__/types-and-hooks.ts
+/src/schema/schema.graphql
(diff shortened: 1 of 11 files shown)


cmcewen (approved):

yyip-dev (approved):
Thanks for linking the doc, very helpful explanations

PR #6380, opened 2023-04-17
build: cache webpack sources

zzmp:
## Description

Caches webpack sources (from node_modules) to the filesystem for faster subsequent builds.
This will result in faster builds across the board: `yarn start`, `yarn build`, and GitHub Actions.

This reduced the time to build in a GitHub Action from 5-6m to 4-5m.

NB: webpack 5 supports native caching of the whole build, but webpack 4 does not - this seems like the simplest and safest solution to use as a stopgap to improve build times until webpack is upgraded.

### QA (ie manual testing)

- [x] N/A

### Automated testing

- [x] Unit test N/A

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -75,6 +75,8 @@ jobs:
           restore-keys: ${{ runner.os }}-build-${{ hashFiles('**/yarn.lock') }}-
       - run: yarn prepare
       - run: yarn build
+        env:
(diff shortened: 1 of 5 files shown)


zzmp:
This will need to be done with cache-loader, not hard-sources-webpack-plugin. Empirically, this is not improving GH Action build times, and it still logs lots of warnings.

PR #6384, opened 2023-04-17
chore: remove graphql feature flag and old query methods

cbachmeier:
## Description

- Remove the nft graphql feature flag now that the change has been live for a week with no regression needed
- Removes the old queries of nft activity, trending nft collections, and nft collection search
- Removes conditional logic that was needed to handle the dual states

_JIRA ticket:_
_Slack thread:_
_Relevant docs:_

## Screen capture
Feature Flag Modal
Screen Shot 2023-04-17 at 10 13 43
NFT Activity Functionality
activity
NFT Collection Search Functionality
searchBar
Trending NFT Table and Caorusel
trendingTable

### QA (ie manual testing)

- All relevant flows (activity, trending, and search) have been tested as shown in the screen gifs


src/components/NavBar/SearchBar.tsx
```diff
@@ -52,44 +49,20 @@ export const SearchBar = () => {
...
-      ? {
-          gatedCollections: gqlCollections,
-          gatedCollectionsAreLoading: gqlCollectionsAreLoading,
-        }
-      : {
-          gatedCollections: queryCollections,
-          gatedCollectionsAreLoading: queryCollectionsAreLoading,
-        }
-  }, [gqlCollections, gqlCollectionsAreLoading, isNftGraphqlEnabled, queryCollections, queryCollectionsAreLoading])
+  const { data: gqlCollections, loading } = useCollectionSearch(debouncedSearchValue)
```
JackShort, line 58: Can we just name this to collections since the specificity is no longer needed?

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -149,34 +145,27 @@ export const SearchBarDropdown = ({
...
-  const isNftGraphqlEnabled = useNftGraphqlEnabled()
   const isTokenPage = pathname.includes('/tokens')
   const [resultsState, setResultsState] = useState<ReactNode>()
 
-  const { data: trendingCollectionResults, isLoading: trendingCollectionsAreLoading } = useQuery(
-    ['trendingCollections', 'eth', 'twenty_four_hours'],
-    () => fetchTrendingCollections({ volumeType: 'eth', timePeriod: 'ONE_DAY' as TimePeriod, size: 3 })
-  )
-
   const { data: gqlData, loading } = useTrendingCollections(3, HistoryDuration.Day)
```
JackShort, line 151: Thinking that gql is the norm now we should rename this

src/nft/components/collection/Activity.tsx
```diff
@@ -66,46 +63,8 @@ export const reduceFilters = (state: typeof initialFilterState, action: { eventT
...
-      },
-      refetchInterval: 15000,
-      refetchIntervalInBackground: false,
-      refetchOnWindowFocus: false,
-      refetchOnMount: false,
-    }
-  )
-
   const {
     nftActivity: gqlEventsData,
```
JackShort, line 69: Same comment as above here

src/nft/components/details/AssetDetails.tsx
```diff
@@ -486,27 +387,27 @@ export const AssetDetails = ({ asset, collection }: AssetDetailsProps) => {
             <Filter eventType={ActivityEventType.Transfer} />
             <Filter eventType={ActivityEventType.CancelListing} />
           </ActivitySelectContainer>
-          {gatedLoading ? (
+          {loading ? (
             <LoadingAssetActivity rowCount={10} />
-          ) : events && events.length > 0 ? (
+          ) : gqlEventsData && gqlEventsData.length > 0 ? (
```
JackShort, line 392: Same comment here

src/nft/components/explore/Banner.tsx
```diff
@@ -117,34 +113,17 @@ const TRENDING_COLLECTION_SIZE = 5
...
-      })
-    },
-    {
-      refetchOnReconnect: false,
-      refetchOnWindowFocus: false,
-      refetchOnMount: false,
-    }
-  )
+
   const { data: gqlData } = useTrendingCollections(
```
JackShort, line 117: Also here

src/nft/components/explore/Cells/Cells.tsx
```diff
@@ -8,7 +7,7 @@ import { ReactNode } from 'react'
 import styled from 'styled-components/macro'
 import { ThemedText } from 'theme'
 
-import { ethNumberStandardFormatter, formatWeiToDecimal } from '../../../utils/currency'
+import { ethNumberStandardFormatter } from '../../../utils/currency'
```
JackShort, line 10: Can we use the actual route of the file?

src/nft/components/explore/TrendingCollections.tsx
```diff
@@ -90,20 +88,6 @@ function convertTimePeriodToHistoryDuration(timePeriod: TimePeriod): HistoryDura
...
-    },
-    {
-      refetchOnReconnect: false,
-      refetchOnWindowFocus: false,
-      refetchOnMount: false,
-      refetchInterval: 5000,
-    }
-  )
 
   const { data: gqlData, loading } = useTrendingCollections(100, convertTimePeriodToHistoryDuration(timePeriod))
```
JackShort, line 92: Same comment here

JackShort (approved):
Just marked a bunch of places where I think we should change the naming convention from `gqlData` to `data` since gql is the norm now

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -149,34 +145,27 @@ export const SearchBarDropdown = ({
...
+  const { data, loading } = useTrendingCollections(3, HistoryDuration.Day)
 
   const trendingCollections = useMemo(() => {
-    const gatedTrendingCollections = isNftGraphqlEnabled ? gqlData : trendingCollectionResults
-    return gatedTrendingCollections && (!isNftGraphqlEnabled || !loading)
-      ? gatedTrendingCollections
-          .map((collection) => ({
+    return !loading
+      ? data
+          ?.map((collection) => ({
```
zzmp, line 156: Is it intentional to return `undefined` if the `data` is empty? If `!loading`, can we assert that `data` is non-empty?

src/graphql/data/nft/CollectionSearch.ts
```diff
@@ -80,9 +79,8 @@ function useCollectionQuerySearch(query: string, skip?: boolean): useCollectionS
 }
 
 export function useCollectionSearch(queryOrAddress: string): useCollectionSearchReturnProps {
-  const isNftGraphqlEnabled = useNftGraphqlEnabled()
   const isName = !isAddress(queryOrAddress.toLowerCase())
-  const queryResult = useCollectionQuerySearch(queryOrAddress, isNftGraphqlEnabled ? !isName : true)
-  const addressResult = useCollection(queryOrAddress, isNftGraphqlEnabled ? isName : true)
+  const queryResult = useCollectionQuerySearch(queryOrAddress, !isName)
+  const addressResult = useCollection(queryOrAddress, isName)
```
zzmp, line 84: The second parameter of both of these methods should be changed to an options pack (eg `{ isQuery: boolean }`) - as is, it's very unclear that they're being called correctly.

src/nft/components/collection/Activity.tsx
```diff
@@ -66,52 +63,9 @@ export const reduceFilters = (state: typeof initialFilterState, action: { eventT
...
-    }
-  )
-
-  const {
-    nftActivity: gqlEventsData,
-    hasNext,
-    loadMore,
-    loading,
-  } = useNftActivity(
+  const { nftActivity, hasNext, loadMore, loading } = useNftActivity(
```
zzmp, line 68: Use more explicit variables.

src/nft/components/collection/ActivityCells.tsx
```diff
@@ -188,16 +186,7 @@ const PriceTooltip = ({ price }: { price: string }) => (
...
-  const formattedPrice = useMemo(
-    () =>
-      price
-        ? isNftGraphqlEnabled
-          ? formatEth(parseFloat(price?.toString()))
-          : putCommas(formatEthPrice(price.toString()))?.toString()
-        : null,
-    [isNftGraphqlEnabled, price]
-  )
+  const formattedPrice = useMemo(() => (price ? formatEth(parseFloat(price?.toString())) : null), [price])
```
zzmp, line 189: Should this be pulled into a util?

src/nft/components/collection/ActivityCells.tsx
```diff
@@ -269,25 +258,16 @@ export const EventCell = ({
...
   return (
     <Column height="full" justifyContent="center" gap="4">
       <Row className={styles.eventDetail} color={eventColors(eventType)}>
         {renderEventIcon(eventType)}
         {ActivityEventTypeDisplay[eventType]}
       </Row>
-      {eventTimestamp && (isValidDate(eventTimestamp) || isNftGraphqlEnabled) && !isMobile && !eventOnly && (
+      {eventTimestamp && !isMobile && !eventOnly && (
```
zzmp, line 268: Why is it no longer being validated (`isValidDate`)?

src/nft/components/collection/ActivityCells.tsx
```diff
@@ -412,10 +391,7 @@ export const ItemCell = ({ event, rarityVerified, collectionName, eventTimestamp
             collectionName={collectionName}
           />
         )}
-        {isMobile &&
-          eventTimestamp &&
-          (isValidDate(eventTimestamp) || isNftGraphqlEnabled) &&
-          getTimeDifference(eventTimestamp.toString(), isNftGraphqlEnabled)}
+        {isMobile && eventTimestamp && getTimeDifference(eventTimestamp.toString())}
```
zzmp, line 394: Why is it no longer being validated?

src/nft/components/details/AssetDetails.tsx
```diff
@@ -344,52 +306,7 @@ export const AssetDetails = ({ asset, collection }: AssetDetailsProps) => {
...
-  )
-
-  const {
-    nftActivity: gqlEventsData,
-    hasNext,
-    loadMore,
-    loading,
-    error,
-  } = useNftActivity(
+  const { nftActivity, hasNext, loadMore, loading, error } = useNftActivity(
```
zzmp, line 309: More explicit variable names, please.

src/nft/components/explore/Banner.tsx
```diff
@@ -117,34 +113,14 @@ const TRENDING_COLLECTION_SIZE = 5
...
-      refetchOnWindowFocus: false,
-      refetchOnMount: false,
-    }
-  )
-  const { data: gqlData } = useTrendingCollections(
-    TRENDING_COLLECTION_SIZE + EXCLUDED_COLLECTIONS.length,
-    HistoryDuration.Day
-  )
+
+  const { data } = useTrendingCollections(TRENDING_COLLECTION_SIZE + EXCLUDED_COLLECTIONS.length, HistoryDuration.Day)
```
zzmp, line 117:
```suggestion
  const { data: trendingCollections } = useTrendingCollections(TRENDING_COLLECTION_SIZE + EXCLUDED_COLLECTIONS.length, HistoryDuration.Day)
```

src/nft/components/explore/TrendingCollections.tsx
```diff
@@ -90,22 +88,8 @@ function convertTimePeriodToHistoryDuration(timePeriod: TimePeriod): HistoryDura
...
-    {
-      refetchOnReconnect: false,
-      refetchOnWindowFocus: false,
-      refetchOnMount: false,
-      refetchInterval: 5000,
-    }
-  )
 
-  const { data: gqlData, loading } = useTrendingCollections(100, convertTimePeriodToHistoryDuration(timePeriod))
+  const { data, loading } = useTrendingCollections(100, convertTimePeriodToHistoryDuration(timePeriod))
```
zzmp, line 92:
```suggestion
  const { data: collections, loading: collectionsLoading } = useTrendingCollections(100, convertTimePeriodToHistoryDuration(timePeriod))
```

zzmp (changes requested):
You often call `useTrendingCollections` and assign it to `collections`, and then transform it and call that transformation `trendingCollections`. This naming scheme is confusing - could it be something like `trendingCollections` -> `collectionColumns`, that's more akin to what the actual transformation is?

src/components/NavBar/SearchBarDropdown.tsx
cbachmeier: Yes, then no suggestions will load which is the desire behaviour. We can't assert `data` is non-empty if `!loading` because it is possible the query returns empty

src/graphql/data/nft/CollectionSearch.ts
cbachmeier: Can you go into more detail here? The second param states whether or not to skip the query and save on a call. The first call should only run if given a name to search for matched results, the second is if given an address which should only return 1 matching collection with the corresponding address

src/nft/components/collection/ActivityCells.tsx
cbachmeier: formatEth is already a util, otherwise we're just making sure price is passed in as a number

src/nft/components/collection/ActivityCells.tsx
cbachmeier: This has to do with the format passed back to us from the BE. From our now deprecated heroku api, date was passed as a string, non-unix timestamp, so this "validate" function made sure it could be parsed into unix. The gql endpoint returns dates now as straight unix so the validate function isn't needed

src/nft/components/collection/ActivityCells.tsx
cbachmeier: see above

cbachmeier:
> You often call useTrendingCollections and assign it to collections, and then transform it and call that transformation trendingCollections. This naming scheme is confusing - could it be something like trendingCollections -> collectionColumns, that's more akin to what the actual transformation is?
There's only 1 instance where it gets formatted to match the columns in the Trending Collections table, the other instance formats the trending collections to the same format as collections returned from search. Both instances have had their names updated

src/graphql/data/nft/CollectionSearch.ts
zzmp: My point was just that I can't tell what the second parameter is meant to be from the callsite. Considering it's a boolean, I'd prefer it look like `, /* paramName= */ isName)` or, even more preferable, like `, { paramName: isName })`, for future readers.

cbachmeier: I would need to completely refactor the arguments for both the `useCollectionQuerySearch` and `useCollection` to allow for the named arguments and these functions are currently used in several locations throughout the app. To avoid scope creep and potentially introducing bugs into this chore PR, I've used your first suggestion

zzmp (approved):

zzmp (approved):

src/components/NavBar/SearchBarDropdown.tsx
zzmp: This must be an empty array, not null, or else L319 may err at runtime.

zzmp (changes requested):

zzmp (approved):

PR #6394, opened 2023-04-19
test: add cypress-hardhat

zzmp:
## Description

Installs, but does not configure, cypress-hardhat.
This adds a hardhat fork running from the cypress server process for cypress e2e tests. Once configured, it can be used to set up and assert on blockchain state for those tests.

Installing without configuring keeps the commits smaller, and separates the complexity. Configuration will involve mocking an injected provider, so it is not trivial.

_JIRA ticket:_ https://uniswaplabs.atlassian.net/browse/INFRA-121
_Relevant docs:_ https://github.com/Uniswap/cypress-hardhat#readme

### QA (ie manual testing)

- [x] Verified manually that cypress-hardhat runs in this installation. This will be verified via automated testing once configured - for now it is only installed.

### Automated testing

- [x] Unit test N/A
- [x] Integration/E2E test: cypress tests continue to chug along


package.json
```diff
@@ -140,6 +142,7 @@
     "@reduxjs/toolkit": "^1.6.1",
     "@sentry/react": "^7.45.0",
     "@sentry/tracing": "^7.45.0",
+    "@sentry/types": "^7.45.0",
```
zzmp, line 145: `@sentry/types` is imported from as part of `src/tracing`, but it was "shadow-installed" through other sentry packages. `hardhat` uses an older version of sentry, so it was necessary to install it so that typechecking would not break - not to mention it shouldn't have been "shadow-installed" in the first place.

package.json
```diff
@@ -101,8 +101,10 @@
     "@vanilla-extract/babel-plugin": "^1.1.7",
     "@vanilla-extract/webpack-plugin": "^2.1.11",
     "cypress": "10.3.1",
+    "cypress-hardhat": "^1.0.0",
     "env-cmd": "^10.1.0",
     "eslint": "^7.11.0",
+    "hardhat": "^2.14.0",
```
zzmp, line 107: `hardhat` is a _large_ dependency, so it is expected that there is a lot of churn in `yarn.lock`.

hardhat.config.js
```diff
@@ -0,0 +1,22 @@
+/* eslint-env node */
+require('dotenv').config()
+
+const mainnetFork = {
+  url: `https://mainnet.infura.io/v3/${process.env.REACT_APP_INFURA_KEY}`,
+  blockNumber: 17023328,
```
yyip-dev, line 6: Curious why this block number in particular? Or will this get re-configured?

yyip-dev (approved):

hardhat.config.js
zzmp: It's arbitrary. It was close to the latest when I initially wrote this, that's all.

hardhat.config.js
```diff
@@ -0,0 +1,22 @@
+/* eslint-env node */
+require('dotenv').config()
+
+const mainnetFork = {
+  url: `https://mainnet.infura.io/v3/${process.env.REACT_APP_INFURA_KEY}`,
```
JFrankfurt, line 5: confirming `REACT_APP_INFURA_KEY` is the correct key in the hardhat env

zzmp: yes -  hardhat needs to fork off of a mainnet archival node, and the infura JSON-RPC endpoint is used to access that archival node.

hardhat.config.js
cmcewen: could you add a comment here? or should this query the latest block at runtime and use it?

cmcewen (approved):

PR #6396, opened 2023-04-19
fix: filter OneKey related errors

vm:
## Description
Filters out OneKey related errors. This is a desktop crypto wallet which runs our dapp in a browser.

An example error is here: https://uniswap-labs.sentry.io/issues/4009216608/?project=4504255148851200&query=is%3Aunresolved&referrer=issue-stream&sort=freq&statsPeriod=24h&stream_index=0

But this is very likely to be related to this: https://uniswap-labs.sentry.io/issues/4008758184/?project=4504255148851200&query=is%3Aunresolved&referrer=issue-stream&sort=freq&statsPeriod=24h&stream_index=3

Therefore, we don't get a benefit from keeping these specific errors. We can re-consider turning these back on once sourcemaps are working properly.

### Automated testing

- [x] Unit test


vm:
Talked to Connor. We think it'd be better to try and solve the sourcemaps first to see what the root cause is here vs ignoring these now, even if that takes a bit longer.

vm:
I'm re-considering this now that we know we can't see the stack trace properly in Sentry even when sourcemaps are enabled. This is due to the filesystem being different in OneKey.

https://uniswap-labs.sentry.io/issues/4009216608/?project=4504255148851200&referrer=release-issue-stream

zzmp (changes requested):

zzmp (approved):

PR #6397, opened 2023-04-19
fix: filter CSP unsafe-eval errors

vm:
## Description
Filters out unsafe-eval errors caused by our content-security policy. This is what prevents unsafe javascript evaluation on our site which can be an attack vector. Since these are expected errors, we don't need to send them to Sentry.

### Automated testing
- [x] Unit test

### DIFF
--- a/src/tracing/errors.ts
+++ b/src/tracing/errors.ts
@@ -48,8 +48,9 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event: 
     if (error.message.match(/Unexpected token '<'/)) return null
 
     /*
-     * Content security policy errors can be filtered out because there are expected failures.
+     * Content security policy 'unsafe-eval' errors can be filtered out because there are expected failures.
      * For example, if a user runs an eval statement in console this error would still get thrown.
+     * TODO(INFRA-176): We should extend this to filter out any type of CSP error.
      */
     if (
       error.message.match(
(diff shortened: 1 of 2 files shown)


src/tracing/errors.ts
```diff
@@ -46,6 +46,18 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event:
      * Therefore, this can be ignored.
      */
     if (error.message.match(/Unexpected token '<'/)) return null
+
+    /*
+     * Content security policy errors can be filtered out because there are expected failures.
```
zzmp, line 51: This will not filter out all CSP errors, only `unsafe-eval`.

We _should_ be reporting these violations somewhere, but it's not yet configured. You should ping @cmcewen to see if that's planned. Once that's done, you can expand the regex to match all reported errors. Could you ticket that and link it in a TODO(INFRA-123)-style comment?

zzmp (changes requested):

grabbou (approved):
Looks good to me! I don't have all the knowledge here, but just wanted to point out whether all `unsafe-eval` errors are expected to have this error message across different browsers? If not, do we care? I guess the regex pattern is from Sentry, based on what is actually reported, so while this may not cover all, it certainly does the trick for now.

vm:
I do think they may be different by browser, but I think we should handle that as part of the ticket I made above.

cbachmeier (approved):
LGTM

zzmp (approved):

PR #6399, opened 2023-04-19
feat: [DetailsV2] build containers for data page

cbachmeier:
## Description
- Added containers and placeholder files for all the components to be built on the data page of the new NFT details page. For more information see the [spec document](https://www.notion.so/uniswaplabs/Details-Redesign-0b14fc9f47d946c8af5523b27812205e).
- Added a page break spacer for the landing page
- Added extensible tabs for the containers that have different content depending on active tab
  - The tabs for Traits, Description and Details were not given their own files due to their simplicity and our plan for them to be completed as part of the same task
  - To track which tab is active, I purposefully didn't use the index of the `tabs` array. This decision was based on the fact that the pattern of using the tab's key better aligns with object oriented principles and allows for active tab state management to be disconnected from the data integrity of the array. Active state can be determined solely based off the tabs themselves. This comes at a tradeoff of having to do an O(n) function to get the active content on render, but with max number of tabs currently set to 3 this tradeoff is negligible. Additionally, if performance suffers due to the find function, `activeContent` could be memoized to only run the function call on a relevant data change.
- Containers wrap when on mobile screen
- This will allow future parallel component work with little chance for conflict
- Moved all DetailsV2 related files to their own folder
- Note: The uses of `style={{height: XXXpx}}` is just a visual placeholder and will be updated when the component is implemented.

_JIRA ticket:_ https://uniswaplabs.atlassian.net/browse/NFT-1168
_Slack thread:_
_Relevant docs:_ [Details Redesign Implementation Plan](https://www.notion.so/uniswaplabs/NFT-Details-Redesign-Implementation-Plan-f28dd956c8ae4005b40a164e18722259)

## Screen capture

New containers:
Screen Shot 2023-04-19 at 10 12 01

Mobile:
Screen Shot 2023-04-19 at 10 16 05

Tab switching
tabs


src/nft/components/details/detailsV2/DataPageDescription.tsx
```diff
@@ -0,0 +1,41 @@
+import { Trans } from '@lingui/macro'
+import styled from 'styled-components/macro'
+
+import { containerStyles } from './shared'
+import { Tab, TabbedComponent } from './TabbedComponent'
+
+const DescriptionContainer = styled.div`
+  height: 288px;
+  padding: 16px 20px 20px;
```
JackShort, line 9: I think the differing bottom padding on this one is a design mistake and they should all be the same with 16px.

JackShort: I think moving `padding: 16px 20px` and setting `padding-left: 0px` and `padding-right: 0px` for the table might be the best. Will talk to design tho

src/nft/components/details/detailsV2/DataPageTraits.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+const TraitsContainer = styled.div`
+  height: 528px;
+  padding: 16px 20px;
+
+  ${containerStyles}
+`
+
+export const DataPageTraits = () => {
+  return <TraitsContainer>Traits</TraitsContainer>
```
JackShort, line 13: I think it might actually be better to put this in a Tabbed component so it can share the same styling as the others. Just make it an array of 1

src/nft/components/details/detailsV2/DataPage.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  @media screen and (max-width: ${BREAKPOINTS.xs}px) {
+    padding: 24px 20px;
+  }
+`
+
+const ContentContainer = styled(Row)`
+  gap: 24px;
+  padding-bottom: 45px;
```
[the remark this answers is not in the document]
cbachmeier: true for the containers which I'll ping Phil about, but this specific instance is for the entire page bottom padding

src/nft/components/details/detailsV2/DataPageDescription.tsx
```diff
@@ -0,0 +1,41 @@
+import { Trans } from '@lingui/macro'
+import styled from 'styled-components/macro'
+
+import { containerStyles } from './shared'
+import { Tab, TabbedComponent } from './TabbedComponent'
+
+const DescriptionContainer = styled.div`
+  height: 288px;
```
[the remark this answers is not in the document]
cbachmeier: correct, this is just for visual testing while no content is present in the containers

src/nft/components/details/detailsV2/DataPage.tsx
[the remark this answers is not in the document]
cbachmeier: confirmed 16px

JackShort (approved):

src/nft/components/details/detailsV2/TabbedComponent.tsx
```diff
@@ -0,0 +1,62 @@
...
+const TabsRow = styled(Row)`
+  gap: 32px;
+  margin-bottom: 12px;
+  width: 100;
+`
+
+const Tab = styled(ThemedText.SubHeader)<{ isActive: boolean; numTabs: number }>`
+  color: ${({ theme, isActive }) => (isActive ? theme.textPrimary : theme.textTertiary)};
+  line-height: 24px;
+  cursor: ${({ numTabs }) => (numTabs > 1 ? 'pointer' : 'default')};
```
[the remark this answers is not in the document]
cbachmeier: I chose numTabs deliberately so that it's flexible if we want to add different behaviour depending on number of tabs down the road

src/nft/components/details/detailsV2/TabbedComponent.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const [activeTab, setActiveTab] = useState(defaultTab.key)
+  const activeContent = tabs.find((tab) => tab.key === activeTab)?.content
+  return (
+    <TabbedComponentContainer style={style}>
+      <TabsRow>
+        {tabs.map((tab) => (
+          <Tab
+            isActive={activeTab === tab.key}
+            numTabs={tabs.length}
+            onClick={() => setActiveTab(tab.key)}
```
cmcewen, line 52: so typically the .map pattern you pass the `setActiveTab` function and then a key to the child, and then you don't create new functions in each render. in this case because the tab is a styled component that would require a new wrapper, i don't think it's worth it - but just something to keep in mind.

src/nft/components/details/detailsV2/TabbedComponent.tsx
```diff
@@ -0,0 +1,62 @@
...
+}
+
+interface TabbedComponentProps {
+  tabs: Tab[]
+  defaultTab: Tab
+  style?: React.CSSProperties
+}
+
+export const TabbedComponent = ({ tabs, defaultTab, style }: TabbedComponentProps) => {
+  const [activeTab, setActiveTab] = useState(defaultTab.key)
```
cmcewen, line 43: why not use array index?

src/nft/components/details/detailsV2/TabbedComponent.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const activeContent = tabs.find((tab) => tab.key === activeTab)?.content
```
cmcewen, line 44: array index also makes this easier - you just do tabs[activeTab].content

src/nft/components/details/detailsV2/TabbedComponent.tsx
```diff
@@ -0,0 +1,62 @@
...
+  cursor: ${({ numTabs }) => (numTabs > 1 ? 'pointer' : 'default')};
+
+  &:hover {
+    opacity: ${({ numTabs, theme }) => numTabs > 1 && theme.opacity.hover}};
+  }
+`
+
+export interface Tab {
+  title: React.ReactNode
+  key: string
```
cmcewen, line 32: it looks like you only use this as a unique identifier for rendering, so you could nuke this and just use array length and pass index in your map function. however i don't think it's necessarily bad to have this as like a "url slug" in case you want to like change the default tab to render somehow. but then i might rename it to `slug` or something

cmcewen (changes requested):
i think the style vars are fine if you don't have an answer yet, but i'd like a change in tab props/index management

src/nft/components/details/detailsV2/DataPage.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { DataPageHeader } from './DataPageHeader'
+import { DataPageTable } from './DataPageTable'
+import { DataPageTraits } from './DataPageTraits'
+
+const DataPageContainer = styled(Column)`
+  padding: 24px 64px;
+  height: 100vh;
+  width: 100%;
+  gap: 36px;
+  max-width: 1200px;
```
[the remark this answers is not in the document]
cbachmeier: this is a const only for the Data Page as used in the Figma

src/nft/components/details/detailsV2/DataPage.tsx
```diff
@@ -0,0 +1,55 @@
...
+  margin: 0 auto;
+
+  @media screen and (max-width: ${BREAKPOINTS.sm}px) {
+    padding: 24px 48px;
```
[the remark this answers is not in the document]
cbachmeier: The X padding for sm and xs screen sizes could be re-used for the Landing page but otherwise these are only for the Data Page.

src/nft/components/details/detailsV2/TabbedComponent.tsx
cbachmeier: By using the `key` we can know which is the active tab without outside reference to the entire `tabs` array. I think this is a better object oriented pattern than relying on the tab's position in the array.

src/nft/components/details/detailsV2/TabbedComponent.tsx
cbachmeier: There would also be tradeoffs, later on when we map the tabs to React components we'd also have to track index for knowing which tab is currently active.

src/nft/components/details/detailsV2/TabbedComponent.tsx
cbachmeier: Again I think it's a better object oriented pattern to control state based on a unique identifier for the element than it's array index.

src/nft/components/details/detailsV2/DataPage.tsx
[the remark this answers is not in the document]
grabbou: Is it referenced somewhere else in styles? Worth extracting as a theme value? Is there any other container width used by other pages? Similar? Maybe we can unify?

src/nft/components/details/detailsV2/DataPage.tsx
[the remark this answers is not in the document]
grabbou: How different are they from the rest of the website? Is this something we can try to unify?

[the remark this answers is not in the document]
grabbou: Also, if those values are repeated in other components, maybe we can have another component, like `Padding` that handles breakpoints on its own.

src/nft/components/details/detailsV2/DataPageTable.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import { ActivityTableContent } from './ActivityTableContent'
+import { ListingsTableContent } from './ListingsTableContent'
+import { OffersTableContent } from './OffersTableContent'
+import { Tab, TabbedComponent } from './TabbedComponent'
+
+const TableTabs: Array<Tab> = [
+  {
+    title: <Trans>Activity</Trans>,
+    key: 'activity',
```
grabbou, line 11: Should this be a constant (maybe)?

src/nft/components/details/detailsV2/TabbedComponent.tsx
grabbou: Yeah, agreed that unique identifier might be better as with keys we may have some unexpected behavior should we hide tabs dynamically in the future.

src/nft/components/details/detailsV2/TabbedComponent.tsx
grabbou: Why not use `object` or `map` instead? That way, you can still access by object notation. For iteration, you can always use `Object.values`

src/nft/components/details/detailsV2/DataPage.tsx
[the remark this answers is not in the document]
cbachmeier: I just parsed the code base and didn't find another example where `padding: 24px 48px` was used. This may warrant a large conversation with Design to set standards but for this PR given it's the only use I lean towards current implementation.

src/nft/components/details/detailsV2/DataPage.tsx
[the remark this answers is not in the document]
cbachmeier: I dug and found that this max-width is used in a couple of locations in NFTs and Token Explore. Token Explore has set up their own set of breakpoints in `src/components/Tokens/constants.ts` including `MAX_WIDTH_MEDIA_BREAKPOINT = '1200px'`. I've set a `maxWidth` in the theme file and implemented it where used in NFT pages. To avoid scope creep and introducing regressions I did not alter the Token explore code.

src/nft/components/details/detailsV2/TabbedComponent.tsx
cbachmeier: Converted to map

src/nft/components/details/detailsV2/DataPageTable.tsx
cbachmeier: made it an enum in the parent file

cmcewen (approved):

PR #6400, opened 2023-04-19
build: add yarn-analyze

zzmp:
## Description

Adds a yarn script to analyze source maps. This can be helpful for size analyses and source map debugging.

_Relevant docs:_ https://create-react-app.dev/docs/analyzing-the-bundle-size/

## Test plan
N/A


package.json
```diff
@@ -20,6 +20,7 @@
     "postinstall": "patch-package",
     "start": "craco start",
     "build": "craco build",
+    "analyze": "source-map-explorer 'build/static/js/*.js'",
```
vm, line 23: analyze seems a bit vague?

zzmp: `analyze` is the command used as part of CRA, so it's idiomatic here.

grabbou: correct! it's not present by default, but docs use that name there too https://create-react-app.dev/docs/analyzing-the-bundle-size/

grabbou (approved):

PR #6401, opened 2023-04-19
build: upgrade web3-react@^8.2

zzmp:
## Description

Upgrades web3-react to the stable 🥳 version.

Adds `resolutions` that require the stable version for dependencies (ie widgets). This should be a nop, because nothing has changed since v8.1.2-beta.0, the version previously required. It is more risky and onerous to first upgrade widgets, because other things have gone into widgets, such as upgrading the SOR package.

## Test plan

Manually confirmed wallet connection still works.


grabbou (approved):
FWIW, this wasn't previously required b/c widgets was in sync with interface. Now, that widgets is still on `beta` and deduping doesn't work with beta versions, we need to manually set the version via `resolutions`. Should be good to drop it once widgets is upgraded in this project.

cmcewen (approved):

PR #6403, opened 2023-04-20
chore: removing unused files in nft public folder

JackShort:
we are currently not using these files in the public/nft folder

* Main use of testing was searching for the string '/nft/' in the codebase to see if anything was referencing the public folder -> then visually testing
* also removes fallback provider code since that is not being used anymore


cbachmeier (approved):
Thanks for the clean up

cgkol (approved):

cgkol:
fwiw, i am assuming you have confirmed that it works as expected still? Try to add ways you tested on the PR description so people know for future reference if something breaks!

cbachmeier (approved):

zzmp (approved):

PR #6405, opened 2023-04-20
draft: v2 containers

JackShort:


src/nft/components/details/detailsV2/DataPageTraits.tsx
```diff
@@ -0,0 +1,19 @@
+import { Trans } from '@lingui/macro'
+
+import { Tab, TabbedComponent } from './TabbedComponent'
+
+const TraitsContent = () => {
+  return <div>Traits</div>
+}
+
+const DescriptionTabs: Array<Tab> = [
```
cbachmeier, line 9: Traits component doesn't have tabs it just has the one main page. We wouldn't want the Traits title to appear to the user as clickable

src/nft/components/details/detailsV2/DataPageTraits.tsx
```diff
@@ -0,0 +1,19 @@
...
+  {
+    title: <Trans>Traits</Trans>,
+    key: 'description',
```
cbachmeier, line 12: traits key

src/nft/components/details/detailsV2/TabbedComponent.tsx
```diff
@@ -0,0 +1,62 @@
...
+export interface Tab {
+  title: React.ReactNode
+  key: string
+  content: JSX.Element
+}
+
+interface TabbedComponentProps {
+  tabs: Tab[]
+  defaultTab: Tab
+  noContentPadding?: boolean
```
cbachmeier, line 44: I don't think it's worth having such a specific parameter be required across all the components since we see such variation, I think it might be better to handle it at atomic container level

PR #6408, opened 2023-04-20
chore: removing fallback rarity provider

JackShort:
* removing fallback rarity provider because we no longer use this


src/nft/components/collection/ActivityCells.tsx
```diff
@@ -351,12 +348,10 @@ const Ranking = ({ rarity, collectionName, rarityVerified }: RankingProps) => {
         text={
           <Row>
             <Box display="flex" marginRight="4">
-              <img src={rarityProviderLogo} alt="cardLogo" width={16} />
+              <img src="/nft/svgs/gem.svg" alt="cardLogo" width={16} />
```
cbachmeier, line 351: we could also move this gem svg (and potentially any others) to the nft `icons` file

cbachmeier (approved):
LGTM

src/nft/components/collection/ActivityCells.tsx
JackShort: yeah turns out its impossible to actually get a rarity sniper svg - it has to be in here. The svg file is pretty bad

cgkol (approved):
we should have probably have had some tests for this to begin with but 🤷‍♂️

PR #6411, opened 2023-04-20
refactor: keep constants insular

zzmp:
## Description

Refactors constants to not require sibling directories, so that it can be used from within cypress tests.

## Test plan
N/A


JFrankfurt (approved):

cartcrom (approved):

just-toby (approved):

cmcewen (approved):
`not require other folds` is not clear to me but seems to be something to do with import structure

PR #6412, opened 2023-04-20
feat: add apple store banner

just-toby:
## Description

Adds a basic Apple Store meta tag to the site, which links to the Mobile Wallet (if installed) or the App store. 

_Slack thread:_ https://uniswapteam.slack.com/archives/CKX7SR7E1/p1682028116824229
_Relevant docs:_ https://developer.apple.com/documentation/webkit/promoting_apps_with_smart_app_banners

## Screen capture

| paste_before | past_after      | image    |

### QA (ie manual testing)

- [x] manually verified that the banner appears and links to the app from safari, but not from chrome
- [x] manually verified that the banner links to the app store when the app is not installed

### Automated testing

- [x] <s>Unit test</s>


grabbou (approved):
LGTM.

Nit: Might make sense to prefix that line with a little comment, like we do for `manifest.json`. Chances are that comment was automatically generated, but adding a small one for iTunes wouldn't hurt.

lynnshaoyu (approved):

PR #6413, opened 2023-04-20
refactor: activation hook w/ global state

cartcrom:
## Description

Moves pending transaction state from `<WalletModal />` to a new hook: `useActivateConnection`. Also relocates some connection utility code for reusability

Updates `<UniwalletModal />` to use `useActivateConnection` to cancel the pending activation upon modal close, fixing [WEB-3219](https://uniswaplabs.atlassian.net/browse/WEB-3219)

Adds new tests for the pending connection/activation functions & state variables returned by `useActivateConnection`

### Reproducing the error

1. Open the Uniswap Wallet Modal
2. Close it
3. The connection options should be stuck in a loading state

### QA (ie manual testing)

- [ ] Test connecting with all connection options to ensure refactors did not introduce bugs
- [ ] Open the Uniswap Wallet modal, close it, reopen it & connect

[WEB-3219]: https://uniswaplabs.atlassian.net/browse/WEB-3219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/WalletModal/index.tsx
```diff
@@ -185,9 +128,9 @@ export default function WalletModal({ openSettings }: { openSettings: () => void
         <ThemedText.SubHeader fontWeight={500}>Connect a wallet</ThemedText.SubHeader>
         <IconButton Icon={Settings} onClick={openSettings} data-testid="wallet-settings" />
       </AutoRow>
-      {pendingError ? (
-        pendingConnection && (
-          <ConnectionErrorView openOptions={openOptions} retryActivation={() => tryActivation(pendingConnection)} />
+      {pending.error ? (
+        pending.error && (
```
grabbou, line 132: Is this expected? Shouldn't there be just one `pending.error` check?

src/components/WalletModal/index.tsx
```diff
@@ -135,46 +111,13 @@ export default function WalletModal({ openSettings }: { openSettings: () => void
...
-
-          sendAnalyticsEvent(InterfaceEventName.WALLET_CONNECT_TXN_COMPLETED, {
-            result: WalletConnectionResult.FAILED,
-            wallet_type: connection.getName(),
-          })
-        }
-      }
-    },
-    [dispatch, setPendingError, toggleWalletDrawer]
+  const activate = useCallback(
```
grabbou, line 118: Do we need `useCallback` here? Can't we just pass it as an arrow function? Looks a bit unnecessary, given that we still pass an arrow function in line `133`. If we removed this, we would get rid of two arrow functions and just use one (pass everything inline on line `133`)

src/components/WalletModal/index.tsx
```diff
@@ -185,9 +128,9 @@ export default function WalletModal({ openSettings }: { openSettings: () => void
         <ThemedText.SubHeader fontWeight={500}>Connect a wallet</ThemedText.SubHeader>
         <IconButton Icon={Settings} onClick={openSettings} data-testid="wallet-settings" />
       </AutoRow>
-      {pendingError ? (
-        pendingConnection && (
-          <ConnectionErrorView openOptions={openOptions} retryActivation={() => tryActivation(pendingConnection)} />
+      {pending.error ? (
+        pending.error && (
+          <ConnectionErrorView openOptions={cancelActivation} retryActivation={() => activate(pending.error)} />
```
grabbou, line 133: I personally find it a bit confusing that `openOptions` here is passed `cancelActivation`. What does `cancelActivation` do that it opens options? Is there a better / more explicit way to express it? For example, clearing errors and then, opening the modal once again? I guess (but it's just my guess and I would like some help here) that canceling activation works here only because we have an assumption that modal is already open? 

Could this potentially break if modal wasn't open? For connections such as WalletConnect, we may hide the modal while connection is "pending" to avoid z-index issues. Could this break?

src/components/WalletModal/index.tsx
```diff
@@ -198,8 +141,8 @@ export default function WalletModal({ openSettings }: { openSettings: () => void
                 <Option
                   key={connection.getName()}
                   connection={connection}
-                  activate={() => tryActivation(connection)}
-                  pendingConnectionType={pendingConnection?.type}
+                  activate={() => activate(connection)}
+                  pendingConnectionType={pending.connection?.type}
```
grabbou, line 145: Could we have a follow-up to refactor this component to get rid of `pendingConnectionType` in favor of two new props, `pending` and `disabled`? That would make it more explicit than passing this prop. Wasn't initially clear to me what's the purpose of it.

src/connection/activate.ts
```diff
@@ -0,0 +1,94 @@
...
+          return
+        }
+
+        updatePendingError(error)
+      }
+    },
+    [dispatch, updatePendingConnection, updatePendingError]
+  )
+
+  return { pending, tryActivation, cancelActivation }
```
grabbou, line 93: I think it would be a bit better if this exported React Query-like API, that is: `isPending`, `isError`, `error`, `tryActivation` and `cancelActivation` instead of having `pending.error` under it.

I would be open to naming it: `isPendingActivation`, `isActivationError` and `activationError` to avoid renaming later on the call-site, so it's more explicit.

src/connection/activate.ts
```diff
@@ -0,0 +1,94 @@
...
+    async (connection: Connection, onSuccess: () => void) => {
+      // Skips wallet connection if the connection should override the default behavior, i.e. install metamask or launch coinbase app
+      if (connection.overrideActivate?.()) return
+
+      try {
+        updatePendingConnection(connection)
+        await connection.connector.activate()
+
+        console.debug(`connection activated: ${connection.getName()}`)
+        dispatch(updateSelectedWallet({ wallet: connection.type }))
```
grabbou, line 71: Suggestion for future refactor: Would be great if this `updateSelectedWallet` and `pendingConnectionAtom` was a single thing. `isPending`, `isSuccess`, `isError`. Feels like we're doing a bit of redundant work here. Out of scope for this PR, but just leaving as something we could discuss as a future improvement.

src/connection/activate.ts
```diff
@@ -0,0 +1,94 @@
...
+  )
+
+  const cancelActivation = useCallback(async () => {
+    await pending.connection?.connector.deactivate?.()
+    updatePendingConnection(undefined)
+  }, [pending, updatePendingConnection])
+
+  const tryActivation = useCallback(
+    async (connection: Connection, onSuccess: () => void) => {
+      // Skips wallet connection if the connection should override the default behavior, i.e. install metamask or launch coinbase app
```
grabbou, line 63: nit: break into multi-line (too long) and fix spelling, e.g. `MetaMask` or `Coinbase`

src/connection/activate.ts
```diff
@@ -0,0 +1,94 @@
...
+      if (connection.overrideActivate?.()) return
```
grabbou, line 64: Could this be prefaced with a comment? What does this code do? If `overrideActivate` is present and returns a boolean, we return?

src/connection/activate.ts
```diff
@@ -0,0 +1,94 @@
...
+
+      try {
+        updatePendingConnection(connection)
```
grabbou, line 67: nit: `setPendingConnection` instead of `update` for consistency

src/connection/activate.ts
```diff
@@ -0,0 +1,94 @@
...
+  connection: Connection | undefined
+  error: any
+}
+
+const pendingConnectionAtom = atom<PendingConnectionState>({ connection: undefined, error: undefined })
+
+export function useActivateConnection() {
+  const dispatch = useAppDispatch()
+
+  const [pending, setPending] = useAtom(pendingConnectionAtom)
```
grabbou, line 22: I would rename this from `pending` to something else, so it doesn't suggest the value is `boolean`, but an object. Maybe we could have two atoms, one for `connection` and another one for `error`. Also, any particular reason to use Jotai here instead of `useState`? AFAIK if we're not referring to this from the outside of this component, I don't think there's any reason for using that over state. But I may be wrong, so I am leaving this as an open question!

src/connection/activate.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  const updatePendingConnection = useCallback(
+    (connection: Connection | undefined) => {
+      // log selected wallet
+      if (connection !== undefined) {
```
grabbou, line 27: I know we have moved this code over from previous implementation, but I just wanted to ask whether logging an event here makes sense. This occurs while connection is pending. Not while connection was completed. In case it errors, we will send an event of changed wallet, but that did not complete successfully.

src/connection/activate.ts
```diff
@@ -0,0 +1,94 @@
...
+      // Setting a new connection clears any pending errors
+      setPending({ connection, error: undefined })
+    },
+    [setPending]
+  )
+
+  const updatePendingError = useCallback(
+    (error: any) => {
+      setPending(({ connection }) => {
+        if (error !== undefined && connection !== undefined) {
```
grabbou, line 43: Why `if (error !== undefined && connection !== undefined) {` here? Shouldn't the caller make sure to call `updatePendingError` when there's an error, so we don't need to check here for its presence and noop otherwise?

src/connection/activate.ts
```diff
@@ -0,0 +1,94 @@
...
+          sendAnalyticsEvent(InterfaceEventName.WALLET_CONNECT_TXN_COMPLETED, {
```
grabbou, line 44: Question: how is `sendAnalyticsEvent` different from `sendEvent`?

src/components/AccountDrawer/UniwalletModal.tsx
```diff
@@ -39,44 +38,32 @@ const Divider = styled.div`
 `
 
 export default function UniwalletModal() {
-  const open = useModalIsOpen(ApplicationModal.UNIWALLET_CONNECT)
-  const toggle = useToggleUniwalletModal()
-
+  const { pending, cancelActivation } = useActivateConnection()
   const [uri, setUri] = useState<string>()
+  const open = pending.connection?.type === ConnectionType.UNIWALLET && !isIOS && !!uri
```
grabbou, line 43: Question: why do we have `!isIOS` here and `!!uri`? Is `open` a noop in other cases? Maybe it's better to handle this outside of the component. Otherwise, seeing this code is a bit confusing as to how the code behaves when `open` is false (despite `pending.connection?.type === ConnectionType.UNIWALLET` being true).

grabbou (changes requested):
Hey! Thanks for the PR! I requested some changes and also, left some open questions to better understand the code. Overall, I like the unification this PR is trying to achieve, so it's net positive change. I just want to make sure we're on the same page as far as the changes and that we make sure there's nothing confusing arising from this code in the future!

src/components/AccountDrawer/UniwalletModal.tsx
cartcrom: Good point, I'll move the isIOS check to the parent & add a comment explaining the check for URI here

src/components/WalletModal/index.tsx
cartcrom: Ahhh thanks for catching, I was refactoring and accidentally replaced checks for connection with checks for error

src/components/WalletModal/index.tsx
cartcrom: Great point

src/components/WalletModal/index.tsx
cartcrom: Good point, renaming from `openOptions` to `close`. `cancelActivation` resets the pending connection and error, causing this component to render the connection options again rather than showing the error view

src/components/WalletModal/index.tsx
cartcrom: Sounds good, agree this is confusing

src/connection/activate.ts
cartcrom: I'm not sure if it makes sense either. I noticed that as well when I was moving it over, what do you think we should do here?

src/connection/activate.ts
cartcrom: sendEvent goes to Google Analytics (which I'm not sure if we still use?) vs sendAnalyticsEvent goes to Amplitude

src/connection/activate.ts
cartcrom: Good catch, I don't think I actually needed this check for error at all, removed it

src/connection/activate.ts
cartcrom: I'm a bit confused, there is a prefacing comment right above it 😅 if overrideActivate exists and has an action described in the comment above to execute, then it returns true and we skip activation

src/connection/activate.ts
cartcrom: I think `selectedWallet` refers to a successfully connected wallet that the user is currently accessing though, while `pendingConnectionAtom` only has to do with pending state

src/connection/activate.ts
cartcrom: I think adding those might feel too boilerplate-y for this hook's use-case & I'm not sure the two components that use this hook will benefit from shaping this more like an api

src/connection/activate.ts
cartcrom: Will rename 👍 I originally had it as two sep atoms, but it felt cleaner to combine into an object since we're generally updating them at the same time

These are moved into an atom because if this hook used `useState`, then both callers of this hook `<WalletModal />` and `<UniwalletModal />` would have their own pending connection state, when the desired functionality is for closing the `<UniwalletModal />` to update the same pending state that `<WalletModal />` uses

src/connection/activate.ts
grabbou: Yeah, in the future, `selectedWallet` itself could have `pending` flag to make it less confusing! Not something for this PR, I think we could have this as a follow-up if at all!

src/connection/activate.ts
grabbou: Yeah, I must have got confused by the comment! Overall, I am just curious when `connection.overrideActivate?.()` returns `false`, in that case, this will not early return!

src/connection/activate.ts
grabbou: Sounds good! Thanks for explaining that! If we have two call-sites, that definitely makes sense!

src/connection/activate.ts
grabbou: Makes sense! A bit confusing name, but that clarifies it a bit!

src/components/AccountDrawer/UniwalletModal.tsx
```diff
@@ -39,44 +37,35 @@ const Divider = styled.div`
...
 export default function UniwalletModal() {
-  const open = useModalIsOpen(ApplicationModal.UNIWALLET_CONNECT)
-  const toggle = useToggleUniwalletModal()
-
+  const { pendingConnection, error, cancelActivation } = useActivateConnection()
   const [uri, setUri] = useState<string>()
+
+  const isUniwapWallet = pendingConnection?.type === ConnectionType.UNIWALLET
+  // Displays the modal if a Uniswap Wallet Connection is pending without error, & qrcode URI is available
+  const open = isUniwapWallet && error === undefined && uri !== undefined
```
grabbou, line 45: A few things:
1. Can we manage the state inside `useActivateConnection` in a way that error is automatically cleared when connection is pending, so that here, we can actually use `isPending` instead of `error === undefined`? Just feels a bit more idiomatic 
2. Can `uri` be null and if yes, does that have any other meaning? If not, maybe a simple `&& !!uri` does the trick.

src/components/WalletModal/index.tsx
```diff
@@ -185,9 +126,9 @@ export default function WalletModal({ openSettings }: { openSettings: () => void
         <ThemedText.SubHeader fontWeight={500}>Connect a wallet</ThemedText.SubHeader>
         <IconButton Icon={Settings} onClick={openSettings} data-testid="wallet-settings" />
       </AutoRow>
-      {pendingError ? (
-        pendingConnection && (
-          <ConnectionErrorView openOptions={openOptions} retryActivation={() => tryActivation(pendingConnection)} />
+      {error ? (
+        pendingConnection !== undefined && (
```
grabbou, line 130: If we manage to clear `error` automatically when `isPending`, we don't need this nested `if` condition, right? We can assume that errors will get automatically cleared when activation is called. This is what often happens with forms, where `error` is cleared as soon as you start typing, so you don't check for text input length (I hope that analogy works here)

src/connection/activate.ts
```diff
@@ -0,0 +1,95 @@
...
+          return
+        }
+
+        setPendingError(error)
+      }
+    },
+    [dispatch, setPendingConnection, setPendingError]
+  )
+
+  return { pendingConnection: pendingState.connection, error: pendingState.error, tryActivation, cancelActivation }
```
grabbou, line 94: Is the presence of `pendingConnection` equivalent to `isPending: true`?

grabbou (commented):
Good work! Thanks for addressing my previews review! This is looking good! Did another pass

src/connection/activate.ts
cartcrom: For most connection options, overrideActivate will be undefined. For something like MetaMask, we have to determine at runtime (b/c of injection timing) whether we should prompt metamask installation or not. So if metamaskConnection.overrideActivate?.() is called w/ injection present, false is returned.

src/connection/activate.ts
cartcrom: I see! Makes sense

src/connection/activate.ts
cartcrom: Turns out we no longer need this google-analytics event! Removed 👍

src/components/WalletModal/index.tsx
cartcrom: I refactored the return type to make this cleaner 👍

src/components/AccountDrawer/UniwalletModal.tsx
cartcrom: 1. Yes, I updated to use a stricter `ActivationState` type that makes this cleaner.

2. Technically the WalletConnect eventListener returns `any` for the uri. It's possible it could return null, would have to look into web3-react and/or the WalletConnect lib but will use `&& !!uri` for now 👍

grabbou: Yeah, in case of 2nd, I think that's a correct assumption, I am not aware of any differentiation there.

src/connection/activate.ts
```diff
@@ -0,0 +1,86 @@
...
+    () =>
+      setActivationState((activationState) => {
+        if (activationState.status !== ActivationStatus.EMPTY) activationState.connection.connector.deactivate?.()
+        return EMPTY_ACTIVATION_STATE
+      }),
+    [setActivationState]
+  )
+}
+
+export function useActivationState() {
```
grabbou, line 80: nit: I think something like `useActivateConnection` (previous name) would work here a bit more, b/c we're not just returning the state, but also tryActivation and cancelActivation hooks.

grabbou (commented):
Thanks for addressing my review! This is looking good, we're getting closer, leaving another round of feedback!

src/connection/activate.ts
cartcrom: I felt like `useActivateConnection` implied it was returning a single function. I like `useActivationState` here since it returns state + functions to update state, similar to useState (although I agree this isn't perfect, since it returns an object rather than an array)

src/components/WalletModal/Option.tsx
```diff
@@ -62,25 +64,26 @@ const IconWrapper = styled.div`
...
       name={InterfaceEventName.WALLET_SELECTED}
       properties={{ wallet_type: connection.getName() }}
       element={InterfaceElementName.WALLET_TYPE_OPTION}
     >
       <OptionCardClickable
-        onClick={!pendingConnectionType ? activate : undefined}
-        clickable={!pendingConnectionType}
-        disabled={Boolean(!isPending && !!pendingConnectionType)}
+        onClick={!isSomeConnectorPending ? activate : undefined}
+        clickable={!isSomeConnectorPending}
```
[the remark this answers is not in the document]
cartcrom: Yeah this is definitely overly complicated ty

I refactored the css so that `!clickable` disables pointer events, and then replaced `disabled` w/ a more simple `selected` var

src/components/WalletModal/Option.tsx
```diff
@@ -14,21 +16,22 @@ const OptionCardLeft = styled.div`
...
   width: 100% !important;
-  border-color: ${({ theme, isActive }) => (isActive ? theme.accentActive : 'transparent')};
 
   display: flex;
   flex-direction: row;
   align-items: center;
   justify-content: space-between;
-  margin-top: 2rem;
-  padding: 1rem;
+  padding: 18px;
```
cartcrom, line 28: `isActive` was unused/leftover from old code, so I removed the border completely since there is no border in our current design. I upped the padding from 16px (idk why we had rem here) to 18px to fix the size change caused by zero border

grabbou: Sounds good, thanks for explanation!

src/components/WalletModal/Option.tsx
```diff
@@ -62,25 +65,26 @@ const IconWrapper = styled.div`
...
       name={InterfaceEventName.WALLET_SELECTED}
       properties={{ wallet_type: connection.getName() }}
       element={InterfaceElementName.WALLET_TYPE_OPTION}
     >
       <OptionCardClickable
-        onClick={!pendingConnectionType ? activate : undefined}
-        clickable={!pendingConnectionType}
-        disabled={Boolean(!isPending && !!pendingConnectionType)}
+        onClick={activate}
+        clickable={!isSomeConnectorPending}
```
grabbou, line 86: Removing `disabled` was a good call, I still don't understand why we had both. Anyway, do you think removing `clickable` and sticking to native `disabled` would actually make more sense here? That way we don't even need to set `pointer-events: none` for clickable, because `disabled` will disable the button entirely.

src/components/WalletModal/Option.tsx
```diff
@@ -62,25 +65,26 @@ const IconWrapper = styled.div`
...
 `
 
-type OptionProps = {
-  connection: Connection
-  activate: () => void
-  pendingConnectionType?: ConnectionType
-}
-export default function Option({ connection, pendingConnectionType, activate }: OptionProps) {
-  const isPending = pendingConnectionType === connection.type
+export default function Option({ connection }: { connection: Connection }) {
```
grabbou, line 68: BTW, while we're at it, could you create a simple snapshot test for this component to make sure appropriate state / options for active option, pending option and other pending option? 🙏

src/components/WalletModal/Option.tsx
cartcrom: Yes that makes way more sense, will use disabled instead thank you.

Previously, clickable meant whether the hover state should activate, and disabled meant whether the opacity should be lowered (because when one connector is pending, all should _not_ be clickable, but the selected one should not have it's opacity lowered)

image

So we still have to pass in two vars to accomplish this, but I think we've arrived at the much easier-to-follow  `selected` and `disabled` compared to the previous `disabled` and `clickable`

src/components/WalletModal/Option.tsx
cartcrom: Great idea, will do!

src/connection/activate.ts
```diff
@@ -0,0 +1,88 @@
...
+  IDLE,
+}
+
+type ActivationPendingState = { status: ActivationStatus.PENDING; connection: Connection }
+type ActivationErrorState = { status: ActivationStatus.ERROR; connection: Connection; error: any }
+const IDLE_ACTIVATION_STATE = { status: ActivationStatus.IDLE } as const
+
+type ActivationState = ActivationPendingState | ActivationErrorState | typeof IDLE_ACTIVATION_STATE
+
+const pendingConnectionStateAtom = atom<ActivationState>(IDLE_ACTIVATION_STATE)
```
zzmp, line 24: Why is this _pending_, and also why does the name mismatch the type?

Why not `activationStateAtom`?

src/connection/activate.ts
```diff
@@ -0,0 +1,88 @@
...
+        // TODO(WEB-3162): re-add special treatment for already-pending injected errors
+        console.debug(`web3-react connection error: ${JSON.stringify(error)}`)
+
+        // Gracefully handles errors from the user rejecting a connection attempt
+        if (didUserReject(connection, error)) {
+          setActivationState(IDLE_ACTIVATION_STATE)
+          return
+        }
+
+        sendAnalyticsEvent(InterfaceEventName.WALLET_CONNECT_TXN_COMPLETED, {
```
zzmp, line 59: Should there be an analytics event on a successful connection? Will it duplicate the event already sent by the Web3Provider?

src/components/WalletModal/Option.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    expect(option).toMatchSnapshot()
+  })
+
+  it('renders pending state when selected', () => {
+    const result = renderHook(useActivationState).result
+    result.current.tryActivation(mockConnection1, jest.fn())
+
+    const component = render(<Option connection={mockConnection1} />)
+    const option = component.getByTestId('wallet-modal-option')
+    option.click()
```
zzmp, line 43: Should these clicks be wrapped in an `act`?

src/components/AccountDrawer/index.tsx
```diff
@@ -18,7 +17,7 @@ const DRAWER_MARGIN = '8px'
 const DRAWER_OFFSET = '10px'
 const DRAWER_TOP_MARGIN_MOBILE_WEB = '72px'
 
-const accountDrawerOpenAtom = atom(false)
+const accountDrawerOpenAtom = atomWithReset(false)
```
zzmp, line 20: If its just a boolean, i think a reset atom may be overkill.

zzmp (changes requested):
mostly just questions - this definitely feels like a simplification and improvement :)

src/components/TopLevelModals/index.tsx
```diff
@@ -26,7 +27,7 @@ export default function TopLevelModals() {
       <AddressClaimModal isOpen={addressClaimOpen} onDismiss={addressClaimToggle} />
       <ConnectedAccountBlocked account={account} isOpen={accountBlocked} />
       <Bag />
-      <UniwalletModal />
+      {!isIOS && <UniwalletModal />}
```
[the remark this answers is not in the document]
cartcrom: Great callout. That is the originally intended behaviour, but thinking about it now, if an iOS user doesn't have the app installed and they click the Uniswap Wallet Connector, this modal is the user's immediate entry-point to the download button / App Store.

I'll remove this change for now and check w/ product for future change. Thank you!

src/connection/activate.ts
cartcrom: neglected to update after renaming the type, ty for catching

src/connection/activate.ts
cartcrom: Yes the Web3Provider will be sending the successful connections here

src/connection/utils.ts
```diff
@@ -25,3 +27,11 @@ export enum ErrorCode {
   WC_MODAL_CLOSED = 'Error: User closed modal',
   CB_REJECTED_REQUEST = 'Error: User denied account authorization',
 }
+
+export function didUserReject(connection: Connection, error: any): boolean {
```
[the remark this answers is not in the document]
cartcrom: Yes I think it can be, but since that will mean updating the didUserReject function there and potentially renaming `swapErrorToUserReadableMessage`, probably makes sense to do in a sep PR. Added a TODO w/ a [ticket](https://uniswaplabs.atlassian.net/browse/WEB-3279)

grabbou (approved):
AWESOME!

src/components/WalletModal/Option.tsx
```diff
@@ -62,25 +63,26 @@ const IconWrapper = styled.div`
...
       name={InterfaceEventName.WALLET_SELECTED}
       properties={{ wallet_type: connection.getName() }}
       element={InterfaceElementName.WALLET_TYPE_OPTION}
     >
       <OptionCardClickable
-        onClick={!pendingConnectionType ? activate : undefined}
-        clickable={!pendingConnectionType}
-        disabled={Boolean(!isPending && !!pendingConnectionType)}
+        onClick={activate}
+        disabled={isSomeConnectorPending}
```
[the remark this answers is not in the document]
cartcrom: done!

src/connection/activate.ts
zzmp: Can you add a comment explaining this (and @link-ing to that code)?

src/connection/activate.ts
```diff
@@ -0,0 +1,88 @@
...
+       * Skips wallet connection if the connection should override the default
+       * behavior, i.e. install MetaMask or launch Coinbase app
+       */
+      if (connection.overrideActivate?.()) return
+
+      try {
+        setActivationState({ status: ActivationStatus.PENDING, connection })
+        await connection.connector.activate()
+
+        console.debug(`Connection activated: ${connection.getName()}`)
```
zzmp, line 41: Additional logging would make any future issues easier to debug:

```suggestion

        console.debug(`Connection activating: ${connection.getName()}`)
        await connection.connector.activate()
        console.debug(`Connection activated: ${connection.getName()}`)
```

src/connection/activate.ts
```diff
@@ -0,0 +1,88 @@
...
+        dispatch(updateSelectedWallet({ wallet: connection.type }))
+
+        // Clears pending connection state
+        setActivationState(IDLE_ACTIVATION_STATE)
+
+        onSuccess()
+      } catch (error) {
+        // TODO(WEB-3162): re-add special treatment for already-pending injected errors
+        console.debug(`web3-react connection error: ${JSON.stringify(error)}`)
+        console.log(error.message)
```
zzmp, line 51: This is overkill, and a `warn` should be used instead:

```suggestion
        console.debug(`Connection failed: ${connection.getName()}`)
        console.warn(error)
```

zzmp: You should not log if the user rejected. Move this logging to _after_ the `didUserReject` check.

zzmp (changes requested):

src/components/WalletModal/Option.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+  it('renders default state', () => {
+    const component = render(<Option connection={mockConnection1} />)
+    const option = component.getByTestId('wallet-modal-option')
+
+    expect(option).toMatchSnapshot()
+  })
+
+  it('renders pending state when selected', () => {
+    const result = renderHook(useActivationState).result
+    result.current.tryActivation(mockConnection1, jest.fn())
```
zzmp, line 39: Any change to state must be wrapped in act.

```suggestion
    await act(() => result.current.tryActivation(mockConnection1, jest.fn()))
```

src/components/WalletModal/Option.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    const component = render(<Option connection={mockConnection1} />)
+    const option = component.getByTestId('wallet-modal-option')
+    act(() => option.click())
```
zzmp, line 43: `act` must be `await`ed.

Also, why are you testing a click when it's already pending?
I think you need to _not_ `tryActivation`, but instead, render, check it, click, check it again, and possible resolve and ensure that it closes the drawer.

src/components/WalletModal/Option.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    expect(option).toMatchSnapshot()
+  })
+
+  it('renders disabled state when another connection is pending', () => {
+    const result = renderHook(useActivationState).result
+    result.current.tryActivation(mockConnection2, jest.fn())
+
+    const component = render(<Option connection={mockConnection1} />)
+    const option = component.getByTestId('wallet-modal-option')
+    act(() => option.click())
```
zzmp, line 54: `act` must be `await`ed.

src/connection/activate.ts
cartcrom: Ahhh I had added the log temporarily and forgotten to remove. Thanks for catching, I'll start using the `console.log('cartcrom', ...)` notation more now.

As for the existing debug, this is set up right now to help debug that MetaMask loading state issue from a few weeks ago that I've been unable to repro if you remember. Lynn was able to repro it, but it logged `[Object object]` to the console and we were not able to determine what the issue was. Hence the stringify for now. I updated the TODO comment above to note that the logging should be moved under the `didUserReject` check later

src/connection/activate.test.ts
```diff
@@ -0,0 +1,256 @@
...
+    getName: () => 'Test Connection',
+    hooks: {} as unknown as Web3ReactHooks,
+    type,
+    shouldDisplay: () => true,
+    connector: new MockConnector(activate, deactivate),
+  }
+}
+
+beforeEach(() => {
+  console.error = jest.fn()
```
zzmp, line 59: You should update this to squelch anything you log so it doesn't spam the console. Right now you're using debug and log.
You should also include assertions for these in your tests if you want to make sure they're not removed.

zzmp (changes requested):

src/connection/activate.test.ts
```diff
@@ -88,6 +74,8 @@ it('Should call activate function on a connection', async () => {
 
   expect(result.current.activationState).toEqual({ status: ActivationStatus.IDLE })
   expect(mockConnection.connector.activate).toHaveBeenCalledTimes(1)
+  expect(console.debug).toHaveBeenLastCalledWith(`Connection activated: ${mockConnection.getName()}`)
+  expect(console.debug).toHaveBeenCalledTimes(2)
```
zzmp, line 78: unnecessary - you've already checked it was called with the strings you wanted.

src/connection/activate.test.ts
```diff
@@ -111,6 +99,8 @@ it('Should properly deactivate pending connection attempts', async () => {
 
   expect(result.current.activationState).toEqual({ status: ActivationStatus.IDLE })
   expect(mockConnection.connector.deactivate).toHaveBeenCalledTimes(1)
+  expect(console.debug).not.toHaveBeenLastCalledWith(`Connection activated: ${mockConnection.getName()}`)
+  expect(console.debug).toHaveBeenCalledTimes(1)
```
zzmp, line 103: unnecessary - you've already checked it was called with the strings you wanted.

src/components/WalletModal/Option.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+    expect(option2).toHaveProperty('selected', false)
+
+    await act(() => option1.click())
+
+    expect(option1).toBeDisabled()
+    expect(option1).toHaveProperty('selected', true)
+    expect(option2).toBeDisabled()
+    expect(option2).toHaveProperty('selected', false)
+
+    await act(async () => activationResponse.resolve())
```
zzmp, line 65: You could add a mock and check that the drawer is closed also.

zzmp (approved):

zzmp (changes requested):
there's still a console.debug showing up in the console during unit tests.

zzmp (approved):

src/components/WalletModal/Option.test.tsx
```diff
@@ -1,12 +1,20 @@
...
 import { createDeferredPromise } from 'test-utils/promise'
 import { act, render } from 'test-utils/render'
 
 import Option from './Option'
 
+const mockCloserDrawer = jest.fn()
+
+beforeEach(() => {
+  jest.spyOn(console, 'debug').mockReturnValue()
+  jest.spyOn(accountDrawer, 'useCloseAccountDrawer').mockReturnValue(mockCloserDrawer)
```
zzmp, line 15: You should mock this from the file:

```
import { useCloseAccountDrawer } from '...'
...
jest.mock('components/AccountDrawer')
...

    mocked(useCloseAccountDrawer).mockReturnValue(jest.fn())

zzmp (changes requested):

zzmp (approved):

TDG510 (commented):
TDG510

PR #6416, opened 2023-04-21
chore: remove miniportfolio and mgtm flags

cartcrom:
## Description

Removes mgtm and miniPortfolio flags from the interface, as neither are used anymore. This will also allow us to remove the mgtm statsig gate, which redundantly passes 100% of users.


just-toby (approved):

src/components/NavBar/MenuDropdown.tsx
```diff
@@ -48,7 +47,7 @@ const PrimaryMenuRow = ({
           <Row onClick={close}>{children}</Row>
         </NavLink>
       ) : (
-        <Row as="a" href={href} target="_blank" rel="noopener noreferrer" className={styles.MenuRow}>
+        <Row cursor="pointer" as="a" href={href} target="_blank" rel="noopener noreferrer" className={styles.MenuRow}>
```
cartcrom, line 50: fixes this tiny bug since I'm touching this file:

image

zzmp: Should the `cursor="pointer" also be removed from `SecondaryLinkedText`?

zzmp (approved):

PR #6420, opened 2023-04-21
feat: [DetailsV2] more shared component content

cbachmeier:
## Description
Iteratively adding more shared styles, components, and functions to be used across the Data page components. These are:
- Some tabs have bubbles corresponding to a number related to the content
- Removed the `styles` param and instead added placeholder height to the content themselves which can be replaced as those components are worked on.

Dependent on #6399 

## Screen capture
Screen Shot 2023-04-21 at 10 53 05

### Automated testing

- [x] Updated snapshot

### DIFF
--- a/src/nft/components/details/detailsV2/ActivityTableContent.tsx
+++ b/src/nft/components/details/detailsV2/ActivityTableContent.tsx
@@ -1,3 +1,5 @@
+import { TableContentHeight } from './DataPageTable'
+
 export const ActivityTableContent = () => {
-  return <div>Activity Content</div>
+  return <div style={{ height: `${TableContentHeight}px` }}>Activity Content</div>
 }
--- a/src/nft/components/details/detailsV2/DataPageDescription.tsx
+++ b/src/nft/components/details/detailsV2/DataPageDescription.tsx
@@ -2,12 +2,14 @@ import { Trans } from '@lingui/macro'
 
 import { Tab, TabbedComponent } from './TabbedComponent'
 
+const DescriptionContentHeight = 252
+
 const DescriptionContent = () => {
-  return <div>Description Content</div>
+  return <div style={{ height: `${DescriptionContentHeight}px` }}>Description Content</div>
 }
 
 const DetailsContent = () => {
-  return <div>Details Content</div>
+  return <div style={{ height: `${DescriptionContentHeight}px` }}>Details Content</div>
 }
 
 const DescriptionTabs: Array<Tab> = [
@@ -24,5 +26,5 @@ const DescriptionTabs: Array<Tab> = [
 ]
 
 export const DataPageDescription = () => {
-  return <TabbedComponent tabs={DescriptionTabs} style={{ height: '288px' }} />
+  return <TabbedComponent tabs={DescriptionTabs} />
 }
--- a/src/nft/components/details/detailsV2/DataPageTable.tsx
+++ b/src/nft/components/details/detailsV2/DataPageTable.tsx
@@ -3,7 +3,7 @@ import { Trans } from '@lingui/macro'
 import { ActivityTableContent } from './ActivityTableContent'
 import { ListingsTableContent } from './ListingsTableContent'
 import { OffersTableContent } from './OffersTableContent'
-import { Tab, TabbedComponent } from './TabbedComponent'
+import { Tab, TabbedComponent, TabTitleWithBubble } from './TabbedComponent'
 
 export const TableContentHeight = 568
 
@@ -14,12 +14,12 @@ const TableTabs: Array<Tab> = [
     content: <ActivityTableContent />,
   },
   {
-    title: <Trans>Offers</Trans>,
+    title: <TabTitleWithBubble title={<Trans>Offers</Trans>} bubbleNumber={/* placeholder */ 11} />,
     key: 'offers',
     content: <OffersTableContent />,
   },
   {
-    title: <Trans>Listings</Trans>,
+    title: <TabTitleWithBubble title={<Trans>Listings</Trans>} bubbleNumber={/* placeholder */ 11} />,
     key: 'listings',
     content: <ListingsTableContent />,
   },
--- a/src/nft/components/details/detailsV2/DataPageTable.tsx
+++ b/src/nft/components/details/detailsV2/DataPageTable.tsx
@@ -5,6 +5,8 @@ import { ListingsTableContent } from './ListingsTableContent'
 import { OffersTableContent } from './OffersTableContent'
 import { Tab, TabbedComponent } from './TabbedComponent'
 
+export const TableContentHeight = 568
+
 const TableTabs: Array<Tab> = [
   {
     title: <Trans>Activity</Trans>,
@@ -24,5 +26,5 @@ const TableTabs: Array<Tab> = [
 ]
 
 export const DataPageTable = () => {
-  return <TabbedComponent tabs={TableTabs} style={{ height: '604px' }} />
+  return <TabbedComponent tabs={TableTabs} />
 }
--- a/src/nft/components/details/detailsV2/DataPageTraits.tsx
+++ b/src/nft/components/details/detailsV2/DataPageTraits.tsx
@@ -1,9 +1,7 @@
 import { Trans } from '@lingui/macro'
-import Row from 'components/Row'
 import { GenieAsset } from 'nft/types'
 
-import { getBubbleText, TabNumBubble } from './shared'
-import { Tab, TabbedComponent } from './TabbedComponent'
+import { Tab, TabbedComponent, TabTitleWithBubble } from './TabbedComponent'
 
 const TraitsContent = () => {
   return <div style={{ height: '492px' }}>Traits Content</div>
@@ -12,12 +10,7 @@ const TraitsContent = () => {
 export const DataPageTraits = ({ asset }: { asset: GenieAsset }) => {
   const TraitTabs: Array<Tab> = [
     {
-      title: (
-        <Row gap="8px">
-          <Trans>Traits</Trans>
-          {asset.traits && <TabNumBubble>{getBubbleText(asset.traits.length)}</TabNumBubble>}
-        </Row>
-      ),
+      title: <TabTitleWithBubble title={<Trans>Traits</Trans>} bubbleNumber={asset.traits?.length} />,
       key: 'traits',
       content: <TraitsContent />,
     },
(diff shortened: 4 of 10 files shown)


src/nft/components/details/detailsV2/DataPageTable.tsx
```diff
@@ -3,7 +3,9 @@ import { Trans } from '@lingui/macro'
 import { ActivityTableContent } from './ActivityTableContent'
 import { ListingsTableContent } from './ListingsTableContent'
 import { OffersTableContent } from './OffersTableContent'
-import { Tab, TabbedComponent } from './TabbedComponent'
+import { Tab, TabbedComponent, TabTitleWithBubble } from './TabbedComponent'
+
+export const TableContentHeight = 568
```
JackShort, line 8: I don't think this should be in this file

src/nft/components/details/detailsV2/ActivityTableContent.tsx
```diff
@@ -1,3 +1,5 @@
+import { TableContentHeight } from './DataPageTable'
+
 export const ActivityTableContent = () => {
-  return <div>Activity Content</div>
+  return <div style={{ height: `${TableContentHeight}px` }}>Activity Content</div>
```
JackShort, line 4: this should just be a styled component - I don't know if `TableContentHeight` needs to be a constant in the first place

src/nft/components/details/detailsV2/DataPageDescription.tsx
```diff
@@ -2,12 +2,14 @@ import { Trans } from '@lingui/macro'
...
+const DescriptionContentHeight = 252
+
 const DescriptionContent = () => {
-  return <div>Description Content</div>
+  return <div style={{ height: `${DescriptionContentHeight}px` }}>Description Content</div>
 }
 
 const DetailsContent = () => {
-  return <div>Details Content</div>
+  return <div style={{ height: `${DescriptionContentHeight}px` }}>Details Content</div>
```
JackShort, line 12: Same comment here this should just be a styled component

src/nft/components/details/detailsV2/DataPageTraits.tsx
```diff
@@ -1,26 +1,26 @@
 import { Trans } from '@lingui/macro'
+import { GenieAsset } from 'nft/types'
 
-import { Tab, TabbedComponent } from './TabbedComponent'
+import { Tab, TabbedComponent, TabTitleWithBubble } from './TabbedComponent'
 
 const TraitsContent = () => {
-  return <div>Traits Content</div>
+  return <div style={{ height: '492px' }}>Traits Content</div>
```
JackShort, line 7: same comment with the styled component here

src/nft/components/details/detailsV2/DataPageTraits.tsx
```diff
@@ -1,26 +1,26 @@
...
-      key: TraitTabsKeys.Traits,
-      content: <TraitsContent />,
-    },
-  ],
-])
-
-export const DataPageTraits = () => {
-  return <TabbedComponent tabs={TraitTabs} style={{ height: '528px' }} />
+export const DataPageTraits = ({ asset }: { asset: GenieAsset }) => {
+  const TraitTabs: Map<string, Tab> = new Map([
```
JackShort, line 15: I think this should be in a `useMemo`

src/nft/components/details/detailsV2/ListingsTableContent.tsx
```diff
@@ -1,3 +1,5 @@
+import { TableContentHeight } from './DataPageTable'
+
 export const ListingsTableContent = () => {
-  return <div>Listings Content</div>
+  return <div style={{ height: `${TableContentHeight}px` }}>Listings Content</div>
```
JackShort, line 4: Same comment here maybe export the styled component

src/nft/components/details/detailsV2/OffersTableContent.tsx
```diff
@@ -1,3 +1,5 @@
+import { TableContentHeight } from './DataPageTable'
+
 export const OffersTableContent = () => {
-  return <div>Offers Content</div>
+  return <div style={{ height: `${TableContentHeight}px` }}>Offers Content</div>
```
JackShort, line 4: Same comment here

src/nft/components/details/detailsV2/TabbedComponent.tsx
```diff
@@ -27,6 +25,23 @@ const Tab = styled(ThemedText.SubHeader)<{ isActive: boolean; numTabs: number }>
   }
```
JackShort, line 25: there is an extra brace here

src/nft/components/details/detailsV2/TabbedComponent.tsx
```diff
@@ -27,6 +25,23 @@ const Tab = styled(ThemedText.SubHeader)<{ isActive: boolean; numTabs: number }>
...
 
+const TabNumBubble = styled(ThemedText.UtilityBadge)`
+  background: ${({ theme }) => theme.backgroundOutline};
+  border-radius: 4px;
+  padding: 2px 4px;
+  color: ${({ theme }) => theme.textSecondary};
+  line-height: 12px;
+`
+
+export const TabTitleWithBubble = ({ title, bubbleNumber }: { title: React.ReactNode; bubbleNumber?: number }) => {
```
JackShort, line 36: I am wondering if this is the best way to do this - or if this should be the default tab component that is rendered and you can just add another prop to the `Tab` interface like `count` which is then passed into `bubbleNumber`

src/nft/components/details/detailsV2/shared.ts
```diff
@@ -7,3 +7,8 @@ export const containerStyles = css`
   width: 100%;
   align-self: flex-start;
 `
+
+export function getBubbleText(num: number) {
```
JackShort, line 11: don't really think this belongs in the shared folder considering its only going to be used in one file

src/nft/components/details/detailsV2/ActivityTableContent.tsx
cbachmeier: This is just a temporary height set for visualization purposes and will be removed by the implementing engineer for that component. I made it a const for consistency across the tabs and to reduce the number of magic number heights

src/nft/components/details/detailsV2/DataPageTable.tsx
cbachmeier: Named the height to reduce magic number use and for consistency across the tabs. This seems like the best central source for it imo.

src/nft/components/details/detailsV2/DataPageTraits.tsx
cbachmeier: There is no complex operations happening in this initialization, just map construction. I believe useMemo in this case would be less efficient

src/nft/components/details/detailsV2/shared.ts
cbachmeier: good callout, I previously hadn't made the bubble text part of tabbedComponent so it was more distributed. Given it's only used in one location I've deprecated it as a helper fn

src/nft/components/details/detailsV2/TabbedComponent.tsx
cbachmeier: The Titles are currently evenly split 3:3 on whether or not they include the count. Given that I have no strong preference but this solution does allow for flexibility

src/nft/components/details/detailsV2/DataPageTable.tsx
JackShort: yeah I understand I just think that if its being used in the same context of `<div style={{ height: TableContentHeight }} />` then it just might be better off to make a styled component with that set height and import the styled component

src/nft/components/details/detailsV2/DataPageDescription.tsx
```diff
@@ -1,13 +1,18 @@
 import { Trans } from '@lingui/macro'
+import styled from 'styled-components/macro'
 
 import { Tab, TabbedComponent } from './TabbedComponent'
 
+const DescriptionContentContainer = styled.div`
```
cmcewen, line 6: would this be a `TabbedContentContainer` given it is used in both description and details?

src/nft/components/details/detailsV2/TabbedComponent.tsx
cmcewen: for me this decision would be based around how tightly the title area is coupled to the overall tab component. if you can change the title very independently of the tab and have it render fine, i think it makes sense to have it here. if changing the tab component impacts title rendering significantly, i think having a `bubbleCount` as an optional tab prop will ensure that if someone messes with tab styling, they also have to update the title rendering

src/nft/components/details/detailsV2/DataPageDescription.tsx
cbachmeier: The entire component is referred to as `Description` in the spec, figma, and jira

JackShort (approved):
lgtm

src/nft/components/details/detailsV2/TabbedComponent.tsx
cbachmeier: Had a discussion with design. When the count is 0 we'll want to hide the bubble anyway so I've updated the Tab interface to have an optional `count` field and made the bubble with count the default. If that count is 0 or undefined, the bubble will be hidden.

src/nft/components/details/detailsV2/DataPageTable.tsx
```diff
@@ -1,10 +1,15 @@
...
+import styled from 'styled-components/macro'
 
 import { ActivityTableContent } from './ActivityTableContent'
 import { ListingsTableContent } from './ListingsTableContent'
 import { OffersTableContent } from './OffersTableContent'
 import { Tab, TabbedComponent } from './TabbedComponent'
 
+export const TableContentContainer = styled.div`
+  height: 568px;
+`
```
zzmp, line 11: By defining this here, you're creating a cyclical dependency between the files using it and this file. Consider moving this to a separate file so that the build is less complex.

zzmp (approved):

src/nft/components/details/detailsV2/DataPageTable.tsx
cbachmeier: Moved to the shared file

zzmp (approved):

PR #6422, opened 2023-04-21
fix: mock useWeb3React

zzmp:
## Description

Mocks useWeb3React. This eliminates many `act` warnings, but is also necessary in order to add more functionality to our wrapping provider (see #6296).

In addition, mocks are often reset, which is unnecessary per our jest configuration, so this PR also removes all cases where mocks are reset in a beforeEach hook (which was often related).

_JIRA ticket:_ Partial fix for https://uniswaplabs.atlassian.net/browse/WEB-3132?search_id=f82d1daf-33c0-40da-8e38-097a920b30eb

### Automated testing

- [x] Unit test


src/test-utils/render.tsx
```diff
@@ -31,23 +31,34 @@ const WithProviders = ({ children }: { children?: ReactNode }) => {
...
-              <MockedProvider>
-                <BlockNumberProvider>
-                  <ThemeProvider>{children}</ThemeProvider>
-                </BlockNumberProvider>
-              </MockedProvider>
-            </Web3Provider>
+            {/*
+             * Web3Provider is mocked through setupTests.ts
+             * To test behavior that depends on Web3Provider, use jest.unmock('@web3-react/core')
+             */}
```
grabbou, line 37: +1 on the comment. My worry is that this may become confusing in the future when analyzed outside of the context of this PR, where I can see where the `<Web3Provider />` used to be. Someone reading this file in the future may not be aware that such provider is needed in first place for our application to work.

Some alternatives:
- Define an empty component `const Web3Provider = (children) => children` that is just an empty pass-through and comment that component instead. That way, it provides direct explanation why we're not using Web3Provider from `web3-react`, but this one instead.
- Mock `components/Web3Provider` in `setupTests` next to where we mock `web3React` import.

src/test-utils/render.tsx
```diff
@@ -31,23 +31,34 @@ const WithProviders = ({ children }: { children?: ReactNode }) => {
...
       </Provider>
     </MockedI18nProvider>
   )
 }
 
-const customRender = (ui: ReactElement) => render(ui, { wrapper: WithProviders })
-const customRenderHook = <Result, Props>(hook: (initialProps: Props) => Result) =>
-  renderHook(hook, { wrapper: WithProviders })
+type CustomRenderOptions = Omit<RenderOptions, 'wrapper'>
+const customRender = (ui: ReactElement, options?: CustomRenderOptions) => {
```
grabbou, line 51: Correct me if I am wrong, but I didn't see you pass 2nd argument to `render` anywhere in this PR. Is this something you plan to use in the future?

src/tracing/trace.test.ts
```diff
@@ -25,13 +25,11 @@ function getTransaction(index = 0): Transaction {
...
 describe('trace', () => {
   beforeEach(() => {
-    mocked(Sentry.startTransaction)
-      .mockReset()
-      .mockImplementation((context) => {
-        const transaction: Transaction = jest.requireActual('@sentry/react').startTransaction(context)
-        transaction.initSpanRecorder()
-        return transaction
-      })
+    mocked(Sentry.startTransaction).mockImplementation((context) => {
```
grabbou, line 28: This change looks like something that could be a good candidate for a follow-up PR, as it seems to be unrelated?

grabbou (changes requested):
Great work! The PR overall looks good to me. I mostly left questions and nits, but I would like to read the answers before this gets merged, hence using "request changes" instead of "approve".

src/components/Identicon/StatusIcon.test.tsx
```diff
@@ -8,17 +10,42 @@ jest.mock('../../hooks/useSocksBalance', () => ({
 }))
 
 describe('StatusIcon', () => {
-  it('renders children in correct order, with no account and with socks', () => {
-    const supportedConnections = getConnections()
-    const injectedConnection = supportedConnections[1]
-    const component = render(<StatusIcon connection={injectedConnection} />)
-    expect(component.getByTestId('StatusIconRoot')).toMatchSnapshot()
+  describe('with no account', () => {
+    it('renders children in correct order, with with socks', () => {
```
JFrankfurt, line 14:
```suggestion
    it('renders children in correct order with socks', () => {
```

src/components/Identicon/StatusIcon.test.tsx
```diff
@@ -8,17 +10,42 @@ jest.mock('../../hooks/useSocksBalance', () => ({
...
   })
 
-  it('renders with no account and showMiniIcons=false', () => {
-    const supportedConnections = getConnections()
-    const injectedConnection = supportedConnections[1]
-    const component = render(<StatusIcon connection={injectedConnection} showMiniIcons={false} />)
-    expect(component.getByTestId('StatusIconRoot').children.length).toEqual(0)
+  describe('with account', () => {
+    beforeEach(() => {
+      mocked(useWeb3React).mockReturnValue({
```
JFrankfurt, line 31: is there a way to get `mocked` to infer `as ReturnType<typeof useWeb3React>`?

src/components/Identicon/StatusIcon.test.tsx
```diff
@@ -8,17 +10,42 @@ jest.mock('../../hooks/useSocksBalance', () => ({
...
+        account: '0x52270d8234b864dcAC9947f510CE9275A8a116Db',
+        isActive: true,
+      } as ReturnType<typeof useWeb3React>)
+    })
+
+    it('renders children in correct order, with with socks', () => {
```
JFrankfurt, line 37:
```suggestion
    it('renders children in correct order with socks', () => {
```

src/components/Identicon/StatusIcon.test.tsx
```diff
@@ -8,17 +10,42 @@ jest.mock('../../hooks/useSocksBalance', () => ({
...
-    expect(component.getByTestId('StatusIconRoot')).toMatchSnapshot()
+  describe('with no account', () => {
+    it('renders children in correct order, with with socks', () => {
+      const supportedConnections = getConnections()
+      const injectedConnection = supportedConnections[1]
+      const component = render(<StatusIcon connection={injectedConnection} />)
+      expect(component.getByTestId('StatusIconRoot')).toMatchSnapshot()
+    })
+
+    it('renders with showMiniIcons=false', () => {
```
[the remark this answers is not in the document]
zzmp: I'm not updating these tests, I'm just updating the test framework. I've added an add'l test for 'account' but I think anything else is out of scope here.

src/components/Identicon/StatusIcon.test.tsx
```diff
@@ -8,17 +10,42 @@ jest.mock('../../hooks/useSocksBalance', () => ({
...
-    const supportedConnections = getConnections()
-    const injectedConnection = supportedConnections[1]
-    const component = render(<StatusIcon connection={injectedConnection} showMiniIcons={false} />)
-    expect(component.getByTestId('StatusIconRoot').children.length).toEqual(0)
+  describe('with account', () => {
+    beforeEach(() => {
+      mocked(useWeb3React).mockReturnValue({
+        account: '0x52270d8234b864dcAC9947f510CE9275A8a116Db',
+        isActive: true,
+      } as ReturnType<typeof useWeb3React>)
```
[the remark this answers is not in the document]
zzmp: this is standard and should not need a comment: mocks expect to recieve the same shape that the fn would return, and we're missing many props, so we must cast it to satisfy tsc.

src/components/SearchModal/CurrencyList/index.test.tsx
```diff
@@ -23,17 +23,6 @@ jest.mock(
       `CurrencyLogo currency=${currency.symbol}`
 )
 
-jest.mock('@web3-react/core', () => {
```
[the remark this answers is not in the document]
zzmp: this mock has been made global. it is now it setupTests!

src/connection/index.test.tsx
```diff
@@ -3,11 +3,6 @@ import INJECTED_LIGHT_ICON from 'assets/svg/browser-wallet-light.svg'
 import { ConnectionType, getConnections, useGetConnection } from 'connection'
 import { renderHook } from 'test-utils/render'
 
-beforeEach(() => {
```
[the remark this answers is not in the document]
zzmp: this is just cleanup, yes. this was a no op with our current config.

src/lib/hooks/useTokenList/fetchTokenList.test.ts
```diff
@@ -10,7 +10,6 @@ describe('fetchTokenList', () => {
 
   beforeEach(() => {
     jest.spyOn(console, 'debug').mockReturnValue(undefined)
-    resolver.mockReset()
```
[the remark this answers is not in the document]
zzmp: mocks are always reset, so this was a no op. it's just misleading, so i took all mock resets out as part of this pr.

src/serviceWorker/document.test.ts
```diff
@@ -46,7 +46,6 @@ describe('document', () => {
     })
 
     beforeEach(() => {
-      fetch.mockReset()
```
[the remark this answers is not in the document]
zzmp: mocks are always reset before each test. this was misleading, so i took all resets out as part of this pr.

src/test-utils/render.tsx
zzmp: I think both alternatives add extra and unnecessary complexity. I'd prefer to just leave it as this comment. This is the place where the providers are setup, so I think it's appropriate to just leave aa comment here.

src/test-utils/render.tsx
zzmp: I added this here for completeness with the wrapped library. It is currently unused.

src/tracing/trace.test.ts
zzmp: I've included removing mock resets in this PR.

src/components/Identicon/StatusIcon.test.tsx
zzmp: `mocked` actually expects `ReturnType<typeof useWeb3React>`, and is "upset" because we only pass a `Partial`.
This _could_ be fixed by updating the typing of `mocked` to accept a `Partial`, but that would be a heavier lift and is out of scope here. This is - afaik - an idiomatic cast in a test file.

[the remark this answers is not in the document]
zzmp: the most common tests case was inactive, so that's what I put as the default.

JFrankfurt (approved):
I noticed a few removed mocks included account addresses--were those just totally unnecessary?

src/components/Identicon/StatusIcon.test.tsx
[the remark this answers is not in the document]
grabbou: Agreed, if they were written in this fashion before.

src/components/Identicon/StatusIcon.test.tsx
grabbou: Yup, like it or not, I am doing that in a few other test cases myself too. A bit verbose and strange, but gets the job done.

grabbou (approved):
Approving, my questions were addressed.

grabbou (approved):

PR #6423, opened 2023-04-21
fix: handle other types of chunk error codes

vm:
## Description
This makes the regex matching more generic. This is because the text can say `error:`, but also things like `timeout:` or `missing:`. These should all be checked for 499s.

See: https://uniswap-labs.sentry.io/issues/4009148084/events/5e960da931d44c19bb54e977f8d651a4/?project=4504255148851200&query=is%3Aunresolved+release%3A95b9624bca2325ffc28bb513426120c9e4a1c8bb+%22timeout%3A%22&referrer=previous-event&sort=freq&statsPeriod=14d&stream_index=0

### Automated testing
- [x] Unit test

### DIFF
--- a/src/tracing/errors.ts
+++ b/src/tracing/errors.ts
@@ -67,7 +67,7 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event: 
      * CF claims that some number of these is expected, and that they should be ignored.
      * See https://groups.google.com/a/uniswap.org/g/cloudflare-eng/c/t3xvAiJFujY.
      */
-    if (error.message.match(/Loading chunk \d+ failed\. \(([^:]+):[^:]+\.chunk\.js\)/)) {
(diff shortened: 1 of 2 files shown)


src/tracing/errors.ts
```diff
@@ -67,7 +67,7 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event:
      * CF claims that some number of these is expected, and that they should be ignored.
      * See https://groups.google.com/a/uniswap.org/g/cloudflare-eng/c/t3xvAiJFujY.
      */
-    if (error.message.match(/Loading chunk \d+ failed\. \(error: .+\.chunk\.js\)/)) {
+    if (error.message.match(/Loading chunk \d+ failed\. \(([^:]+): .+\.chunk\.js\)/)) {
```
just-toby, line 70: this is really generic ... what about just letters ? a-zA-Z

just-toby (approved):

zzmp (approved):

PR #6424, opened 2023-04-21
chore: updating marketplace icons

JackShort:
updating marketplace icons across app besides the listing side

* changing it so whenever a marketplace icon is needed it gets passed through the `getMarketplaceIcon` util function
* also adds the 'Unavailable' text to the buy now button for marketplaces that purchasing isn't supported
* the real point of this pr is to get the activity endpoint setup for migration to nxyz


src/nft/components/icons.tsx
```diff
@@ -4,7 +4,12 @@ import styled, { useTheme } from 'styled-components/macro'
 import { themeVars, vars } from '../css/sprinkles.css'
```
zzmp, line 4: Why not put these in as separate SVG files?

cbachmeier:
Can you make use of these icons for the listing side as well? It should be a fairly simple refactor and prevent this dual state problem which will be tech debt to address.

src/nft/components/collection/ActivityCells.tsx
```diff
@@ -57,6 +58,15 @@ const AddressLink = styled(ExternalLink)`
   }
 `
 
+const isPurchasableOrder = (orderStatus?: OrderStatus, marketplace?: string): boolean => {
```
cbachmeier, line 61: this is separate from updating the marketplace icons, can we split this out into it's own PR or update the title and description?

src/nft/components/collection/ActivityCells.tsx
```diff
@@ -66,7 +76,7 @@ const formatListingStatus = (status: OrderStatus): string => {
     case OrderStatus.Expired:
       return 'Expired'
     case OrderStatus.Valid:
-      return 'Add to Bag'
+      return 'Unavailable'
```
cbachmeier, line 79: should this check isPurchaseableOrder before marking as unavailable?

cbachmeier:
It also looks like we use the old icon format in `src/nft/components/explore/CarouselCard.tsx` however these are formatted differently (grayscale and round), can you check with design if we want to update those as well?

src/nft/components/details/AssetPriceDetails.tsx
```diff
@@ -245,9 +243,9 @@ const OwnerContainer = ({ asset }: { asset: WalletAsset }) => {
           {listing ? 'Your Price' : 'List for Sale'}
         </ThemedText.SubHeader>
         {listing && (
-          <ExternalLink href={listing.marketplaceUrl}>
-            <MarketplaceIcon alt={listing.marketplace} src={getMarketplaceIcon(listing.marketplace)} />
-          </ExternalLink>
+          <MarketplaceIcon href={listing.marketplaceUrl}>
+            {getMarketplaceIcon(listing.marketplace, '20', '20', uuidv4())}
```
cbachmeier, line 247: Gradient is looking strange on asset I have for sale on x2y2
Screen Shot 2023-04-24 at 15 19 24
Screen Shot 2023-04-24 at 15 19 20

src/nft/components/collection/ActivityCells.tsx
```diff
@@ -57,6 +58,15 @@ const AddressLink = styled(ExternalLink)`
   }
 `
 
+const isPurchasableOrder = (orderStatus?: OrderStatus, marketplace?: string): boolean => {
+  if (!marketplace || !orderStatus) return false
+  const purchasableMarkets = Object.keys(NftMarketplace).map((market) => market.toLowerCase())
```
cbachmeier, line 63: this seems to include all markets, is the plan to note the non-puchaseable Nfts in a later PR?

JackShort:
> Can you make use of these icons for the listing side as well? It should be a fairly simple refactor and prevent this dual state problem which will be tech debt to address.

I actually started doing that but its actually a little larger than what I wanted to put in one pr

JackShort:
> It also looks like we use the old icon format in `src/nft/components/explore/CarouselCard.tsx` however these are formatted differently (grayscale and round), can you check with design if we want to update those as well?

this should prob be done in a sep pr

src/nft/components/collection/ActivityCells.tsx
JackShort: it actually already checks it before this function is called. Considering this isn't going to be used outside of this component I don't believe that the extra check would be too necessary

src/nft/components/details/AssetPriceDetails.tsx
JackShort: going to fix did not see that x2y2 icon was using gradients

src/nft/components/collection/ActivityCells.tsx
JackShort: actually this is not quite the case I talked to mz about this and he said that the enum for `NftMarketplace` is only used for purchasable markets thats why this works, but our enum for markets has all of the markets in it

JackShort: thats why it is actually done like this - previously it was an array until mz informed me of this

cbachmeier:
Can you add tech debt tickets to the backlog for updating the Carousel and Listing marketplace icons for our tracking?

src/nft/components/collection/ActivityCells.tsx
cmcewen: is there any chance this would ever change on the backend? it sounds a bit brittle to me because the object is `NftMarketplace` but there is an assumption you'll only get purchasable ones. i think either naming it `PurchasableNftMarketplace` or adding a purchasable field per marketplace could help - or is there any other way to get this info?

src/nft/components/collection/ActivityCells.tsx
```diff
@@ -57,6 +58,15 @@ const AddressLink = styled(ExternalLink)`
...
+
+  const validOrder = orderStatus === OrderStatus.Valid
+  const purchasableMarket = purchasableMarkets.includes(marketplace.toLowerCase())
```
cmcewen, line 66: nit: naming it `purchasableMarket` when `purchasableMarkets` is an array of strings implies this would be a string too, but i think it's a boolean

src/nft/utils/asset.tsx
```diff
@@ -25,8 +40,43 @@ export const getAssetHref = (asset: GenieAsset | WalletAsset, origin?: DetailsOr
   return `/nfts/asset/${address}/${asset.tokenId}${origin ? `?origin=${origin}` : ''}`
 }
 
-export const getMarketplaceIcon = (marketplace: string) => {
-  return `/nft/svgs/marketplaces/${marketplace.toLowerCase()}.svg`
+export const getMarketplaceIcon = (
+  marketplace: string,
+  width: string | number,
+  height: string | number,
+  gradientId?: string
```
cmcewen, line 47: if you're passing in uuidv4 here every time, do you need to pass it in? would there be a case you don't use it for the gradientId?

cmcewen (changes requested):
looking good overall - i'd like to see if there's a way to make sure the purchasable market info doesn't change, and also charlie's suggestion to have a different PR for it would make sense to me

src/nft/components/collection/ActivityCells.tsx
JackShort: Going to move this to another pr, but I talked to BE and this is what they define as purchasable on their side. So it might change but that means something is now purchasable or not so I think the fact that it could change is good

cmcewen (approved):

src/nft/utils/asset.tsx
```diff
@@ -25,8 +41,38 @@ export const getAssetHref = (asset: GenieAsset | WalletAsset, origin?: DetailsOr
   return `/nfts/asset/${address}/${asset.tokenId}${origin ? `?origin=${origin}` : ''}`
 }
 
-export const getMarketplaceIcon = (marketplace: string) => {
-  return `/nft/svgs/marketplaces/${marketplace.toLowerCase()}.svg`
+export const getMarketplaceIcon = (marketplace: string, size: string | number = '16') => {
```
cbachmeier, line 44: nit: a nice/simple add would be a unit test for this helper fn

cbachmeier (approved):
LGTM

JackShort:
@cbachmeier i also removed the unnecessary marketplaces can you confirm they are not needed?

cbachmeier (approved):
LGTM couldn't find any use of the removed markets

PR #6425, opened 2023-04-21
test: transform vanilla-extract for jest

zzmp:
## Description

Adds the vanilla-extract transform to jest.

Profiling jest reveals that most of the time is spent transforming vanilla extract files as if they were proper typescript files. Instead, they should go through their documented transformer (see https://vanilla-extract.style/documentation/test-environments/).

### Automated testing

- [x] Unit test


craco.config.cjs
```diff
@@ -54,14 +54,23 @@ module.exports = {
...
         cacheDirectory: 'node_modules/.cache/jest',
-        transformIgnorePatterns: ['@uniswap/conedison/format', '@uniswap/conedison/provider'],
+        transformIgnorePatterns: [
+          // Ignore node_modules, except for modules with known issues, to speed up the test builds.
+          '/node_modules/(?!(@uniswap/conedison|d3.*|delaunator|internmap|robust-predicates))/',
+        ],
         moduleNameMapper: {
           '@uniswap/conedison/format': '@uniswap/conedison/dist/format',
           '@uniswap/conedison/provider': '@uniswap/conedison/dist/provider',
+          ...jestConfig.moduleNameMapper,
```
grabbou, line 70: I understand this change, but my question is, wasn't this working before?

craco.config.cjs
```diff
@@ -54,14 +54,23 @@ module.exports = {
...
         },
       })
+      return config
```
grabbou, line 73: Why this change?

vanilla.transform.cjs
```diff
@@ -0,0 +1,9 @@
+/***
```
grabbou, line 1: nit: `***`

vanilla.transform.cjs
```diff
@@ -0,0 +1,9 @@
...
+ * @vanilla-extract/jest-transform incorrectly maps its default export to "default", and so must be transformed itself
```
grabbou, line 2: nit: I don't understand how this comment maps to the underlying code, what is that we do to "map its default export to default" here?

grabbou (commented):
Leaving some comments ahead of approving. Let's keep discussing.

vanilla.transform.cjs
```diff
@@ -0,0 +1,12 @@
+/* eslint-env node */
+/**
+ * @file Re-exports the vanilla-extract jest transform, so that jest can properly transform .css.ts files.
+ * @vanilla-extract/jest-transform incorrectly maps its default export, so that `import *` does not work; and expects
```
grabbou, line 4: Would annotate this with ` so that `@vanilla` is not highlighted as JSDoc component.

grabbou (approved):

grabbou (approved):
LGTM!

PR #6426, opened 2023-04-21
fix: handle addEventListener being unsupported on Safari <14

vm:
## Description
`addEventListener` is not defined on Safari versions below 14. This adds code to fallback to `addListener`.
https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener

https://uniswap-labs.sentry.io/issues/4044597264/?end=2023-04-25T15%3A57%3A59&project=4504255148851200&query=release%3A963121f19b0ea66590d648f83f940c9b9928f75b&referrer=issue-stream&sort=freq&start=2023-04-25T12%3A11%3A00&stream_index=19

### Reproducing the error
Use Safari 13 (I haven't actually tried this) and try changing the dark mode preference on the system.

### Automated testing
- [x] Unit test

### DIFF
--- a/src/theme/components/ThemeToggle.tsx
+++ b/src/theme/components/ThemeToggle.tsx
@@ -29,10 +29,16 @@ const themeModeAtom = atomWithStorage<ThemeMode>('interface_color_theme', ThemeM
 export function SystemThemeUpdater() {
   const setSystemTheme = useUpdateAtom(systemThemeAtom)
 
+  const listener = (event: MediaQueryListEvent) => {
+    setSystemTheme(event.matches ? ThemeMode.DARK : ThemeMode.LIGHT)
+  }
+
   useEffect(() => {
-    DARKMODE_MEDIA_QUERY.addEventListener?.('change', (event: MediaQueryListEvent) => {
-      setSystemTheme(event.matches ? ThemeMode.DARK : ThemeMode.LIGHT)
-    })
+    try {
+      DARKMODE_MEDIA_QUERY.addEventListener('change', listener)
+    } catch (e) {
+      DARKMODE_MEDIA_QUERY.addListener(listener)
+    }
   }, [setSystemTheme])
 
   return null
--- a/src/theme/components/ThemeToggle.tsx
+++ b/src/theme/components/ThemeToggle.tsx
@@ -29,10 +29,16 @@ const themeModeAtom = atomWithStorage<ThemeMode>('interface_color_theme', ThemeM
 export function SystemThemeUpdater() {
   const setSystemTheme = useUpdateAtom(systemThemeAtom)
 
+  const listener = (event: MediaQueryListEvent) => {
+    setSystemTheme(event.matches ? ThemeMode.DARK : ThemeMode.LIGHT)
+  }
+
   useEffect(() => {
-    DARKMODE_MEDIA_QUERY.addEventListener('change', (event) => {
-      setSystemTheme(event.matches ? ThemeMode.DARK : ThemeMode.LIGHT)
-    })
+    try {
+      DARKMODE_MEDIA_QUERY.addEventListener('change', listener)
+    } catch (e) {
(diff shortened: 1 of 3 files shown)


src/theme/components/ThemeToggle.tsx
```diff
@@ -29,10 +29,16 @@ const themeModeAtom = atomWithStorage<ThemeMode>('interface_color_theme', ThemeM
...
+  const listener = (event: MediaQueryListEvent) => {
+    setSystemTheme(event.matches ? ThemeMode.DARK : ThemeMode.LIGHT)
+  }
+
   useEffect(() => {
-    DARKMODE_MEDIA_QUERY.addEventListener('change', (event) => {
-      setSystemTheme(event.matches ? ThemeMode.DARK : ThemeMode.LIGHT)
-    })
+    try {
+      DARKMODE_MEDIA_QUERY.addEventListener('change', listener)
```
zzmp, line 38: This should include a comment linking to documentation of why the try-catch is required. Does MDN have a compatibility table?
_What_ is unsupported? `addEventListener` _should_ be supported - is it just the `change` event?

zzmp (changes requested):

src/theme/components/ThemeToggle.tsx
vm: `addListener` is what's supported on older browsers on `matchMedia`. we could either fallback like this, or we could just not worry about this for older browsers (but not fail). lmk what you think?

zzmp: I think you should fallback if it's just one line.
Either way, could you add an in-code comment explaining why this is necessary?

zzmp (changes requested):
could you fallback and add an in-code comment?

zzmp (changes requested):
- You should return a fn that unregisters the listener from the effect
- Can you extract this into a function (not a hook) and add unit tests for it?

vm:
We're gonna polyfill it instead

src/utils/matchMedia.ts
```diff
@@ -0,0 +1,18 @@
+// Need to fallback to support older browsers (mainly, Safari < 14).
+// See: https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList#browser_compatibility
+
+export function addListener(mediaQuery: MediaQueryList, listener: (event: MediaQueryListEvent) => void) {
```
zzmp, line 4: Can you name these addMediaQueryListener and removeMediaQueryListener?

src/utils/matchMedia.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    const listener = jest.fn()
+    removeListener(mediaQuery as unknown as MediaQueryList, listener)
+
+    expect(mediaQuery.removeEventListener).toHaveBeenCalledWith('change', listener)
+    expect(mediaQuery.removeListener).not.toHaveBeenCalled()
+  })
+
+  test('falls back to removeListener for older browsers', () => {
+    const mediaQuery = {
+      removeEventListener: jest.fn(() => {
```
zzmp, line 49: This should be one lined without specifying a full implementation.

src/utils/matchMedia.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    addListener(mediaQuery as unknown as MediaQueryList, listener)
+
+    expect(mediaQuery.addEventListener).toHaveBeenCalledWith('change', listener)
+    expect(mediaQuery.addListener).toHaveBeenCalledWith(listener)
+  })
+})
+
+describe('removeListener', () => {
+  test('removes event listener from media query', () => {
+    const mediaQuery = {
```
zzmp, line 35: Type this at the declaration instead of coercing it later. This is safer in case you expand the test in the future.

zzmp (changes requested):

zzmp (approved):

just-toby (approved):
lgtm

PR #6427, opened 2023-04-21
test: prevent async debounce updates

zzmp:
## Description

Prevents async debounce updates from the popper library, which were significantly slowing down our tests.
This was discovered through profiling, but is a known issue with popper (see https://github.com/floating-ui/react-popper/issues/368#issuecomment-1340360233).

_JIRA ticket:_ Partial fix for https://uniswaplabs.atlassian.net/browse/WEB-3132?search_id=f82d1daf-33c0-40da-8e38-097a920b30eb

### Automated testing

- [x] Unit test


zzmp:
This is functional, but does not prevent the first debounce. I'm going to try to improve it.

src/components/swap/SwapBuyFiatButton.test.tsx
```diff
@@ -76,7 +76,7 @@ describe('SwapBuyFiatButton.tsx', () => {
     mockuseAccountDrawer.mockImplementation(() => [false, toggleWalletDrawer])
     mockUseOpenModal.mockImplementation(() => useOpenModal)
     render(<SwapBuyFiatButton />)
-    await userEvent.click(screen.getByTestId('buy-fiat-button'))
+    await act(() => userEvent.click(screen.getByTestId('buy-fiat-button')))
```
JFrankfurt, line 79: the way I'm reading the docs, useEvent should already handle everything `act` does: 
- https://legacy.reactjs.org/docs/test-utils.html#act
- https://testing-library.com/docs/user-event/intro/

src/setupTests.ts
```diff
@@ -19,3 +22,19 @@ global.matchMedia =
...
+// This is necessary to avoid warnings during tests, as popper will asynchronously update state outside of test setup.
+jest.mock('@popperjs/core', () => {
+  const core = jest.requireActual('@popperjs/core')
+  return {
+    ...core,
+    createPopper: (...args: Parameters<typeof createPopper>) => {
+      const [referenceElement, popperElement, options] = args
+      if (options?.modifiers) {
+        options.modifiers = options.modifiers.filter((modifier) => !modifier.fn)
+      }
```
JFrankfurt, line 36: is it possible that wrapping the modifiers in `act` is a better approach here? I'm not super familiar w/ popper, but I remember at one point we had modifiers with a meaningful ux impact

JFrankfurt (approved):

src/setupTests.ts
```diff
@@ -19,3 +22,29 @@ global.matchMedia =
       removeEventListener: jest.fn(),
     }
   }
+
+// Prevent popper from making state updates asynchronously.
+// This is necessary to avoid warnings during tests, as popper will asynchronously update state outside of test setup.
```
grabbou, line 27: nit: what is our current strategy for multiline? Did we agree to `/*` already?

grabbou (approved):
Looks good, but I have to admit I am not well experienced with Popper. If this is the best way to enable `synchronousUpdate` (rather than passing some configuration with `true` as a value), it's fine.

JFrankfurt (approved):

PR #6436, opened 2023-04-24
fix: eslintrc breakage

lynnshaoyu:
`yarn eslint --fix` was breaking, resulting in the error: 

Error: .eslintrc.js#overrides[1]:
	Configuration for rule "import/no-restricted-paths" is invalid:
	Value ["src/**/*[!.test].ts","src/**/*[!.test].tsx"] should be string.

This PR fixes it.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -21,7 +21,11 @@ module.exports = {
           {
             zones: [
               {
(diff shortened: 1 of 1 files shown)


zzmp (changes requested):
You should be able to fix this with `yarn install` - I think you're using an old version of the eslint import plugin.

lynnshaoyu:
@zzmp is right, no need

PR #6438, opened 2023-04-24
fix: deduplicate remote and local tx lists

JFrankfurt:
## Description
this uses the nonce field on a transaction to remove outdated transactions from the pending tx set

_JIRA ticket:_ https://uniswaplabs.atlassian.net/browse/WEB-3168

## Test plan
submit a transaction on goerli with a very low gas price, then immediately "speed it up" or "cancel" the transaction from your wallet. Observe that once the second transaction is confirmed you'll see the pending tx from the interface removed.

### Reproducing the error
1. send a swap with a very low fee
2. send a replacement/cancellation transaction
3. observe the pending transaction (1) hanging around forever

### Automated testing
- [x] Integration/E2E test


cmcewen (changes requested):
looks like you have some type errors to fix here. also - i like adding the nonce to the transactions, but i'm wondering if there is a different want to do this other than the hook. even though you useEffect it, i don't love a hook that mutates getting called in the render lifecycle. is there any other way to do this? like look in the reducer when you get confirmation of the replacement tx? keep the localtx but dedupe it in the view to only show a confirmed one with that nonce?

zzmp:
> is there any other way to do this? like look in the reducer when you get confirmation of the replacement tx? keep the localtx but dedupe it in the view to only show a confirmed one with that nonce?

I think this should be handled in the reducer. Your duplicate tx is new state, and once you add it via dispatch, the reducer should take care of de-duping it, so it doesn't need to run in a hook / in the render cycle.

Whether or not you also show the canceled tx is something that I'd defer to the design team for.

JFrankfurt:
> I think this should be handled in the reducer. Your duplicate tx is new state, and once you add it via dispatch, the reducer should take care of de-duping it, so it doesn't need to run in a hook / in the render cycle.

The remote transactions don't ever enter the reducer--they're maintained in component state. Only transactions generated by the application get stored in the reducer, so we can't use the `addTransaction` handler on the reducer to dedupe by nonce, since it never gets run w/ the replacement tx.

src/lib/hooks/transactions/updater.tsx
```diff
@@ -39,16 +41,17 @@ const RETRY_OPTIONS_BY_CHAIN_ID: { [chainId: number]: RetryOptions } = {
 const DEFAULT_RETRY_OPTIONS: RetryOptions = { n: 1, minWait: 0, maxWait: 0 }
 
 interface UpdaterProps {
-  pendingTransactions: { [hash: string]: Transaction }
+  pendingTransactions: { [hash: string]: TransactionDetails }
```
JFrankfurt, line 44: this was incorrectly typed

src/lib/hooks/transactions/updater.tsx
```diff
@@ -58,6 +61,15 @@ export default function Updater({ pendingTransactions, onCheck, onReceipt }: Upd
         () =>
           provider.getTransactionReceipt(hash).then((receipt) => {
             if (receipt === null) {
+              if (account) {
+                provider.getTransactionCount(account).then((transactionCount) => {
```
cmcewen, line 65: nit but i prefer async/await 😄

src/state/transactions/reducer.test.ts
```diff
@@ -28,7 +29,7 @@ describe('transaction reducer', () => {
         },
       })
       store.dispatch(updateVersion())
-      expect(store.getState()[1]['abc']).toBeUndefined()
+      expect(store.getState()[SupportedChainId.MAINNET]['abc']).toBeUndefined()
```
cmcewen, line 32: 💯

src/state/transactions/reducer.ts
```diff
@@ -11,24 +12,40 @@ export interface TransactionState {
   }
 }
 
+interface AddTransactionPayload {
```
cmcewen, line 15: ❤️

src/state/transactions/updater.tsx
```diff
@@ -7,14 +7,21 @@ import { useAppDispatch, useAppSelector } from 'state/hooks'
...
   // speed up popup dismisall time if on L2
   const isL2 = Boolean(chainId && L2_CHAIN_IDS.includes(chainId))
   const transactions = useAppSelector((state) => state.transactions)
+  const pendingTransactions = useMemo(() => {
+    if (!chainId || !transactions[chainId]) return {}
+    return Object.values(transactions[chainId]).reduce((acc, tx) => {
+      if (!tx.receipt) acc[tx.hash] = tx
+      return acc
+    }, {} as Record<string, TransactionDetails>)
+  }, [chainId, transactions])
```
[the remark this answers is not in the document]
cmcewen: can you verify that nothing else relied on this incorrect behavior?

src/lib/hooks/transactions/updater.tsx
```diff
@@ -58,6 +61,15 @@ export default function Updater({ pendingTransactions, onCheck, onReceipt }: Upd
...
+                  const tx = pendingTransactions[hash]
+                  if (tx.nonce && tx.nonce < transactionCount) {
```
cmcewen, line 67: can you add a comment to make it clear to keep the tx.nonce check because old saved txs won't have a tx.nonce?

cmcewen (approved):

src/state/transactions/updater.tsx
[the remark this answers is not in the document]
JFrankfurt: confirmed--a more involved modification was happening subsequently on line 18 of `src\lib\hooks\transactions\updater.tsx` before the incorrect data was used

cmcewen (approved):

src/components/AccountDrawer/MiniPortfolio/Activity/index.tsx
```diff
@@ -87,10 +87,20 @@ function combineActivities(localMap: ActivityMap = {}, remoteMap: ActivityMap =
...
-    const remoteActivity = remoteMap?.[hash] ?? {}
-    // TODO(cartcrom): determine best logic for which fields to prefer from which sources, i.e. prefer remote exact swap output instead of local estimated output
-    acc.push({ ...remoteActivity, ...localActivity } as Activity)
+    const localActivity = (localMap?.[hash] ?? {}) as Activity
+    const remoteActivity = (remoteMap?.[hash] ?? {}) as Activity
+
+    // Check for nonce collision
+    const isNonceCollision =
+      localActivity.nonce !== undefined &&
+      Object.keys(remoteMap).some((remoteHash) => remoteMap[remoteHash]?.nonce === localActivity.nonce)
```
JFrankfurt, line 96: I added this so that the local nonce never renders. Before the duplicate was rendering until the next block triggered a refresh.

cypress/e2e/mini-portfolio/activity-history.test.ts
```diff
@@ -0,0 +1,133 @@
...
+      .then((hardhat) => hardhat.wallet.getTransactionCount())
+      .then((currentNonce) => {
+        const nextNonce = currentNonce + 1
+        // Mock graphql response to include a specific nonce.
+        cy.intercept(
+          {
+            method: 'POST',
+            url: 'https://beta.api.uniswap.org/v1/graphql',
+          },
+          {
```
just-toby, line 27: suggestion: move this to a separate file so this test is easier to read

src/lib/hooks/transactions/updater.tsx
```diff
@@ -39,25 +41,36 @@ const RETRY_OPTIONS_BY_CHAIN_ID: { [chainId: number]: RetryOptions } = {
...
+          provider.getTransactionReceipt(hash).then(async (receipt) => {
             if (receipt === null) {
+              if (account) {
+                const transactionCount = await provider.getTransactionCount(account)
+                const tx = pendingTransactions[hash]
+                // We check for the presence of a nonce because we haven't always saved them,
+                //   so this code may run against old store state where nonce is undefined.
+                if (tx.nonce && tx.nonce < transactionCount) {
+                  // We remove pending transactions from redux if they are no longer the latest nonce.
+                  removeTransaction(hash)
```
just-toby, line 71: if this happens, do we still want to throw the `RetryableError` below?

just-toby (approved):

PR #6439, opened 2023-04-24
fix: moonpay modal height

just-toby:
## Description

removes the `maxHeight` property from the moonpay modal in favor of setting the modal height using `vh` units which properly account for top/bottom padding necessary on mobile devices

_JIRA ticket:_ https://uniswaplabs.atlassian.net/browse/WEB-3244?atlOrigin=eyJpIjoiYmU0MDg2MGQwZDFmNGI3OWEyNzEyMzNkYjAwOTBkNzgiLCJwIjoiaiJ9

## Screen capture

| image 
image | Pasted Graphic      | Pasted Graphic 1    |

### Reproducing the error

1. open moonpay modal in a browser window with smaller dimensions, like iPhone SE

### QA (ie manual testing)

- [x] manually inspected the modal on web and mobile, and it's not cut off anymore

#### Devices

- iPhone SE
- Web Chrome full screen width
- ios simulator

### Automated testing

- [x] <s>Unit test</s>
- [x] Integration/E2E test


src/components/FiatOnrampModal/index.tsx
```diff
@@ -125,7 +135,7 @@ export default function FiatOnrampModal() {
   }, [fetchSignedIframeUrl])
 
   return (
-    <Modal isOpen={fiatOnrampModalOpen} onDismiss={closeModal} maxHeight={720}>
+    <Modal isOpen={fiatOnrampModalOpen} onDismiss={closeModal} maxHeight={modalHeight} minHeight={modalHeight}>
```
just-toby, line 138: i'm basically copying the same approach that the currency search modal is using - calculate a `vh` value for the modal height which is mobile-compatible

JFrankfurt: I think if this is the necessary solution you should prob add a `height` prop and pass it through the way max/min currently are:

image

just-toby: sounds good!

src/components/FiatOnrampModal/index.tsx
```diff
@@ -79,6 +80,15 @@ export default function FiatOnrampModal() {
...
   const [loading, setLoading] = useState(false)
 
+  const { height: windowHeight } = useWindowSize()
+  const modalHeight = useMemo(() => {
+    if (!windowHeight) {
+      return 620 // use this as a fallback if we can't get the window height
+    }
+    const modalHeightPx = windowHeight - theme.navHeight
+    return Math.min(Math.round((modalHeightPx / windowHeight) * 100), 80)
+  }, [theme.navHeight, windowHeight])
```
tinaszheng, line 90: slightly confused by what's happening here - so do you want it to be 80% of the window height? why cant you just pass in `80vh` with css instead of doing math

just-toby: @tinaszheng this math was intended to calculate the height of the screen minus the navbar height. but after testing some more, i realized we were always falling back to 80vh because it was smaller. i think using a constant 80vh every time is honestly good enough for this fix, it works well consistently on all devices

JFrankfurt (approved):

cypress/e2e/buy-crypto-modal.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  it('should open and close', () => {
+    cy.visit('/')
+
+    cy.get(getTestSelector('buy-fiat-button'))
+      .click()
+      .then(() => {
+        cy.get(getTestSelector('fiat-onramp-modal')).should('be.visible')
+      })
+
+    cy.get('body')
```
zzmp, line 13: Add comments (preferably before every block except `cy.visit` in your test) explaining what they're doing. eg

    // Click outside the modal to dismiss it.

cypress/e2e/buy-crypto-modal.test.ts
```diff
@@ -0,0 +1,36 @@
...
+
+describe('Buy Crypto Modal', () => {
+  it('should open and close', () => {
+    cy.visit('/')
+
+    cy.get(getTestSelector('buy-fiat-button'))
+      .click()
+      .then(() => {
+        cy.get(getTestSelector('fiat-onramp-modal')).should('be.visible')
+      })
```
zzmp, line 11: Can this be done with less nesting? eg would it perform the same if expressed as

    cy.get(getTestSelector('buy-fiat-button')).click()
    cy.get(getTestSelector('fiat-onramp-modal')).should('be.visible')
    
If so, I think the latter should be preferred, as it makes it much easier to read without having to keep scoping/nesting/closures in mind.

zzmp (changes requested):

zzmp (approved):

PR #6440, opened 2023-04-24
build: set version on sentry release

vm:
## Description
Right now we have cases where this error shows up.
Screen Shot 2023-04-24 at 6 05 45 PM

I believe it's because we're not adding the release version into the github action which actually uploads the sourcemaps.
https://github.com/getsentry/action-release#parameters

The version is set here to the current commit hash. https://github.com/Uniswap/interface/blob/main/src/tracing/index.ts#L23

## Test plan
I believe we have to merge this to see what happens on deploy tomorrow honestly.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -123,4 +123,5 @@ jobs:
         with:
           environment: production
           sourcemaps: './build/static/js'
+          version: ${{ github.sha }}
           url_prefix: '~/static/js'


cmcewen (changes requested):
is there a way to make sure these things are updated at the same time? for example if we were to start using a different way to version releases, how could we ensure this wouldn't break?

vm:
We'd need a way to save a `SENTRY_RELEASE` variable somewhere, then read it in the Github action. Doable though...

.github/workflows/release.yaml
```diff
@@ -123,4 +123,5 @@ jobs:
         with:
           environment: production
           sourcemaps: './build/static/js'
+          version: ${{ github.sha }}
```
zzmp, line 126: Isn't this the default? Why would this have not worked before?

zzmp (changes requested):

vm:
Turns out this is the default as zach said.

PR #6395, opened 2023-04-19
test: add a 'hardhat' mock window.ethereum

zzmp:
## Description

Adds a `'hardhat'` mock `window.ethereum` for cypress e2e tests, using the cypress-hardhat plugin.
Updates the Cypress.VisitOptions to allow selection between `'hardhat'` and `'goerli'`, and to clarify existing usages (eg `userState`).

_JIRA ticket:_ https://uniswaplabs.atlassian.net/browse/INFRA-121
_Slack thread:_ #web-e2e-testing

### QA (ie manual testing)

- [x] Verified that tests run locally. Have not yet fully verified hardhat network usage. This will be done as part of drafting new e2e tests.

### Automated testing

- [x] Unit test N/A
- [x] Integration/E2E tests chugging along 🚂

PR #6442, opened 2023-04-25
feat: e2e testing - Handle wallet rejection

JFrankfurt:
# Description
This adds an e2e test that confirms the correct UX for rejection at the user's wallet.

This will be merged after #6395, which it depends on

_JIRA ticket:_ https://uniswaplabs.atlassian.net/browse/WEB-3259
_Relevant docs:_ https://www.notion.so/uniswaplabs/Chain-interaction-test-flows-6ba2178c23574a5fb7d492188e20955c?pvs=4

- [x] Integration/E2E test


cypress/e2e/swap.test.ts
```diff
@@ -154,6 +154,20 @@ describe('Swap', () => {
     cy.get('#swap-currency-output .token-amount-input').should('have.value', '0.01')
   })
 
+  it('should render and dismiss the wallet rejection modal', () => {
+    cy.stub(hardhat.wallet, 'sendTransaction').rejects(new Error('user cancelled'))
+    cy.get('#swap-currency-output .open-currency-select-button').click()
+    cy.get(getTestSelector('token-search-input')).clear().type(USDC_MAINNET.address)
+    cy.contains('USDC').click()
+    cy.get('#swap-currency-output .token-amount-input').clear().type('1')
+    cy.get('#swap-currency-input .token-amount-input').should('not.equal', '')
```
zzmp, line 163: we should know _exactly_ what it will equal, because we are on a fork

cypress/e2e/swap.test.ts
```diff
@@ -154,6 +154,20 @@ describe('Swap', () => {
     cy.get('#swap-currency-output .token-amount-input').should('have.value', '0.01')
   })
 
+  it('should render and dismiss the wallet rejection modal', () => {
+    cy.stub(hardhat.wallet, 'sendTransaction').rejects(new Error('user cancelled'))
```
zzmp, line 158:
```suggestion
    cy.stub(hardhat.wallet, 'sendTransaction').rejects(new Error('user cancelled'))
    
```

zzmp (changes requested):

cypress/e2e/swap.test.ts
JFrankfurt: I think this is a reasonable fudge, since in order to know exactly what we'd get here we would need to run the router and get the input amount for this exact output swap, and the routing api isn't looking at the fork

zzmp (approved):

zzmp (approved):

just-toby (approved):

PR #6444, opened 2023-04-25
test(e2e): deadline passed swap error state

JFrankfurt:
## Description
mines blocks past the transaction deadline before it is confirmed

_JIRA ticket:_ https://uniswaplabs.atlassian.net/browse/WEB-3257
_Relevant docs:_ https://www.notion.so/uniswaplabs/Chain-interaction-test-flows-6ba2178c23574a5fb7d492188e20955c?pvs=4#460f6635ad714316893a15d74b5a9375


cypress/e2e/swap.test.ts
```diff
@@ -112,6 +115,53 @@ describe('Swap', () => {
...
+      })
+      .then(() => hardhat.utils.getBalance(hardhat.wallet.address, USDC_MAINNET))
+      .then((balance) => Number(balance.toFixed(1)))
+      .then((initialBalance) => {
+        // input swap info
+        cy.get('#swap-currency-output .open-currency-select-button').click()
+        cy.get(getTestSelector('token-search-input')).clear().type(MAINNET_USDC_ADDRESS)
+        cy.contains('USDC').click()
+        cy.get('#swap-currency-output .token-amount-input').clear().type(BALANCE_INCREMENT.toString())
+        cy.get('#swap-currency-input .token-amount-input').should('not.equal', '')
```
zzmp, line 133: Why not assert what it _will_ equal?

cypress/e2e/swap.test.ts
```diff
@@ -112,6 +115,53 @@ describe('Swap', () => {
...
+
+        cy.get('#swap-button').click()
+        cy.get('#confirm-swap-or-send').click()
+
+        cy.then(() => hardhat.send('eth_getBlockByNumber', ['latest']))
+          .then((block) => block.timestamp * 2)
+          .then((timestampIncrease) => {
+            hardhat.send('evm_increaseTime', [hexValue(timestampIncrease)])
+            hardhat.send('hardhat_mine', [hexValue(1), hexValue(timestampIncrease)])
+          })
```
zzmp, line 149: What is this doing?

cypress/e2e/swap.test.ts
```diff
@@ -112,6 +115,53 @@ describe('Swap', () => {
...
+
+        cy.then(() => hardhat.send('hardhat_mine', ['0x1', '0xc'])).then(() => {
+          cy.get(getTestSelector('dismiss-tx-confirmation')).click()
+
+          // ui check
```
zzmp, line 154: These comments should be more explicit, eg: `// Check that the balance has not changed`

zzmp (changes requested):
I think that your comments should be more explicit. A good rule of thumb is to use complete sentences, not just fragments, to ensure it's understandable to someone with less context.

cypress/e2e/swap.test.ts
JFrankfurt: That requires complicated pre-calc of routing stuff given chain state at the fork point

cypress/e2e/swap.test.ts
JFrankfurt: increases chain time, then mines a block

cypress/e2e/swap.test.ts
```diff
@@ -75,35 +77,91 @@ describe('Swap', () => {
...
-            // ui check
+    cy.visit('/swap', { ethereum: 'hardhat' })
+      .hardhat()
+      .then((hardhat) => {
+        hardhat
+          .reset()
+          .then(() => hardhat.getBalance(hardhat.wallet.address, USDC_MAINNET))
+          .then((balance) => Number(balance.toFixed(1)))
+          .then((initialBalance) => {
+            hardhat.provider.send('evm_setAutomine', [true])
```
JFrankfurt, line 89: updating to use the new syntax and enable auto-mine

cypress/e2e/swap.test.ts
```diff
@@ -75,35 +77,88 @@ describe('Swap', () => {
...
-          cy.get(getTestSelector('dismiss-tx-confirmation')).click()
-
-          cy.then(() => hardhat.provider.send('hardhat_mine', ['0x1', '0xc'])).then(() => {
-            // ui check
+    cy.visit('/swap', { ethereum: 'hardhat' })
+      .hardhat()
+      .then((hardhat) => {
+        hardhat
+          .reset()
+          .then(() => hardhat.provider.send('evm_setAutomine', [true]))
```
zzmp, line 86: Doesn't this need to be backed out after being set? How do you know it won't affect the subsequent test?

cypress/e2e/swap.test.ts
```diff
@@ -75,35 +77,88 @@ describe('Swap', () => {
...
+            cy.get('#swap-currency-output .open-currency-select-button').click()
+            cy.contains('USDC').click()
+            cy.get('#swap-currency-output .token-amount-input').clear().type(BALANCE_INCREMENT.toString())
+            cy.get('#swap-currency-input .token-amount-input').should('not.equal', '')
+
+            // set deadline to minimum (1 minute)
+            cy.get(getTestSelector('open-settings-dialog-button')).click()
+            cy.get(getTestSelector('deadline-input')).clear().type(DEADLINE_MINUTES.toString())
+            cy.get('body').click('topRight')
+            cy.get(getTestSelector('deadline-input')).should('not.exist')
```
zzmp, line 133: Is this necessary, or can we test this with the default deadline?

cypress/e2e/swap.test.ts
```diff
@@ -75,35 +77,88 @@ describe('Swap', () => {
...
+
+            // Mine a block past the deadline
+            cy.then(() => hardhat.provider.send('eth_getBlockByNumber', ['latest']))
+              .then((block) => block.timestamp * 2)
+              .then((timestampIncrease) => {
+                hardhat.provider.send('evm_increaseTime', [hexValue(timestampIncrease)])
+                hardhat.provider.send('hardhat_mine', [hexValue(1), hexValue(timestampIncrease)])
+              })
+
+            cy.then(() => hardhat.provider.send('hardhat_mine', ['0x1', '0xc'])).then(() => {
```
zzmp, line 146: I think this can all be replaced by `hardhat.mine(1, DEADLINE_MINUTES * 60 + 1)`.

cypress/e2e/swap.test.ts
```diff
@@ -75,35 +77,88 @@ describe('Swap', () => {
...
+              cy.get(getTestSelector('dismiss-tx-confirmation')).click()
```
zzmp, line 147: Shouldn't there be an error?

zzmp (changes requested):

cypress/e2e/swap.test.ts
zzmp: You may need to update `cypress-hardhat` for this method to be available.

cypress/e2e/swap.test.ts
JFrankfurt: I think in this case since we're planning on having automine set to true by default it's just a defensive safety check

JFrankfurt: If anything I think it could be removed (on the assumption that *other* tests which set it to false unwind that by setting it to true at the end.

cypress/e2e/swap.test.ts
```diff
@@ -135,28 +135,43 @@ describe('Swap', () => {
...
 
-            // Mine a block past the deadline
-            cy.then(() => hardhat.provider.send('eth_getBlockByNumber', ['latest']))
+            // Mine a block past the deadline.
+            cy.then(() => hardhat.provider.send('eth_getBlockByNumber', ['latest', false]))
               .then((block) => block.timestamp * 2)
               .then((timestampIncrease) => {
                 hardhat.provider.send('evm_increaseTime', [hexValue(timestampIncrease)])
                 hardhat.provider.send('hardhat_mine', [hexValue(1), hexValue(timestampIncrease)])
               })
```
zzmp: Why not use `hardhat.mine`? Why do you need to predicate this off of the data from `eth_getBlockByNumber`?

zzmp (changes requested):

zzmp (approved):

cartcrom (approved):

PR #6445, opened 2023-04-25
fix: filter WebAssembly compile errors

vm:
## Description
These are also a form of CSP errors, which prevent unsafe evaluation on the interface. https://uniswap-labs.sentry.io/issues/4009129049/?end=2023-04-25T15%3A57%3A59&project=4504255148851200&query=release%3A963121f19b0ea66590d648f83f940c9b9928f75b&referrer=issue-stream&sort=freq&start=2023-04-25T12%3A11%3A00&stream_index=16

We should eventually find a better way to catch all different strings for CSP errors, but for now we can just filter the ones that are showing up in sentry often enough to worry about.

### Automated testing
- [x] Unit test

### DIFF
--- a/src/tracing/errors.test.ts
+++ b/src/tracing/errors.test.ts
@@ -22,22 +22,22 @@ describe('filterKnownErrors', () => {
     const originalException = new (class extends Error {
       requestBody = JSON.stringify({ method: 'eth_blockNumber' })
     })()
-    expect(filterKnownErrors(ERROR, { originalException })).toBe(null)
+    expect(filterKnownErrors(ERROR, { originalException })).toBeNull()
   })
 
   it('filters network change errors', () => {
     const originalException = new Error('underlying network changed')
-    expect(filterKnownErrors(ERROR, { originalException })).toBe(null)
+    expect(filterKnownErrors(ERROR, { originalException })).toBeNull()
   })
 
   it('filters user rejected request errors', () => {
     const originalException = new Error('user rejected transaction')
-    expect(filterKnownErrors(ERROR, { originalException })).toBe(null)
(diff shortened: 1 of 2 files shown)


src/tracing/errors.test.ts
```diff
@@ -22,22 +22,22 @@ describe('filterKnownErrors', () => {
     const originalException = new (class extends Error {
       requestBody = JSON.stringify({ method: 'eth_blockNumber' })
     })()
-    expect(filterKnownErrors(ERROR, { originalException })).toBe(null)
+    expect(filterKnownErrors(ERROR, { originalException })).toBeNull()
```
zzmp, line 25: Did you catch this from linting?

src/tracing/errors.ts
```diff
@@ -93,6 +93,11 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event:
     if (error.message.match(/'unsafe-eval'.*content security policy/i)) {
       return null
     }
+
+    // WebAssembly compication requires 'unsafe-eval' in CSP, meaning these errors can be filtered out.
```
zzmp, line 97: typo: compication

zzmp, line 97: This comment is unclear. What about:

```suggestion
    // WebAssembly compilation fails because we do not allow 'unsafe-eval' in our CSP.
    // Any thrown errors are due to 3P extensions/applications, so we do not need to handle them.
```

zzmp (approved):

src/tracing/errors.test.ts
vm: no i didn't i just noticed it

cartcrom (approved):

PR #6447, opened 2023-04-25
fix: filter cross-origin frame errors

vm:


vm:
Gonna just ignore in sentry instead of filtering here.

PR #6450, opened 2023-04-26
feat: mining mode switch to disable auto-mine

JFrankfurt:
This allows us to make assertions about pending transaction state


hardhat.config.js
```diff
@@ -21,6 +21,10 @@ module.exports = {
       accounts: {
         count: 1,
       },
+      mining: {
+        auto: false,
+        interval: 8000,
```
zzmp, line 26: why not 12_000?
also, pls format as X_000 for readability.

zzmp (changes requested):

hardhat.config.js
JFrankfurt: 12_000 seems like a good default, but I was concerned about making these tests too long, so went with a slightly shorter default.

zzmp: this won't have auto-mining, so this should just configure the apparent time between blocks. ie if you inspect the timestamp on blocks, they'll be `interval` seconds apart, but they'll _actually_ just mine when you tell them to, they shouldn't add that delay to these tests.

I think you should use `12000` because that's the expected block time on mainnet.

zzmp (changes requested):

hardhat.config.js
zzmp: I'm 100% wrong here - I just checked the docs: https://hardhat.org/hardhat-network/docs/reference#mining-modes.

Why do we want interval mining at all? Why not just mine when we call `hardhat.mine()`?

zzmp (changes requested):

cypress/e2e/swap.test.ts
```diff
@@ -92,16 +92,19 @@ describe('Swap', () => {
...
-          'have.text',
-          `Balance: ${initialBalance + BALANCE_INCREMENT}`
-        )
-
-        // chain state check
-        cy.then(() => hardhat.utils.getBalance(hardhat.wallet.address, USDC_MAINNET))
-          .then((balance) => Number(balance.toFixed(1)))
-          .should('eq', initialBalance + BALANCE_INCREMENT)
+
+        cy.then(() => hardhat.send('hardhat_mine', ['0x1', '0xc'])).then(() => {
```
JFrankfurt, line 96: `hardhat.send` is an async action, so it needs to be wrapped in `cy.then`

zzmp: I think this is fine as an unblocker, but should really be exposed through cypress-hardhat's Utils class.

hardhat.config.js
```diff
@@ -21,6 +21,10 @@ module.exports = {
       accounts: {
         count: 1,
       },
+      mining: {
```
zzmp, line 24: Include a comment explaining why you are doing this.

```suggestion
      // Disable auto-mining so that e2e tests can explicitly test pending states.
      mining: {
```

zzmp (changes requested):

zzmp (approved):

vm (approved):

PR #6451, opened 2023-04-26
feat: new review design

just-toby:
## Description

Implement new design for the "review swap" modal. this was implemented in the widget, so the code is copied and modified from that repo. 

design spec here: https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?type=design&node-id=1708-26043&t=HDoFkKDloxhc8us5-0

_JIRA ticket:_ https://uniswaplabs.atlassian.net/browse/WEB-3251?atlOrigin=eyJpIjoiZGY1M2NjYzEwN2QwNGUyMTk1NDRiMzJkNWNkYzg2NGMiLCJwIjoiaiJ9
_Relevant docs:_ https://www.notion.so/uniswaplabs/Swap-widget-replacement-work-bd91b44494d547259b4f29f4c36b3dfc

## Screen capture

| Before      | Before (mobile) | After (Desktop) | After (Mobile) |
| ------------| ---------------- |---------------- | -------------- |
| image |  image | image      |  image   image |

### QA (ie manual testing)

- [x] manually comparing the new implementation to the figma, and adhoc design review

#### Devices

- chrome desktop
- chrome devtools mobile layout simulator
- ios simulator on mac

### Automated testing

- [x] <s>Unit test</s>


src/components/swap/SwapModalHeaderAmount.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { ReactNode, useState } from 'react'
+import { Info } from 'react-feather'
+import styled, { useTheme } from 'styled-components/macro'
+import { ThemedText } from 'theme'
+
+const MAX_AMOUNT_STR_LENGTH = 9
+
+export const Label = styled.span`
+  color: ${({ theme }) => theme.textSecondary};
+  margin-right: 0.5rem;
```
grabbou, line 16: To be honest, I wasn't aware we're using `rem` up to this point. Apparently, we have around 150 occurrences of `rem` throughout the codebase vs ~2000 occurrences of `px`. Doesn't look like `rem` is the way forward, given our codebase is mostly filled with pixels. 

One of the things I would like to make sure, of while we're at this widget replacement work, is not to reduce code debt if we can do it while adding new features -> great.

I would suggest we discuss this at a meeting, which is the way we want to go forward and whether we want to refactor e.g. to pixels while copying the code over.

src/components/swap/SwapModalFooter.tsx
```diff
@@ -120,46 +154,136 @@ export default function SwapModalFooter({
...
+      <Column gap="md" padding="0 0.5rem">
+        {details.map(({ label, value, color, labelTooltipText }, i) => (
+          <SwapModalDetailRow key={i} label={label} value={value} color={color} labelTooltipText={labelTooltipText} />
+        ))}
+      </Column>
+      {showAcceptChanges ? (
+        <SwapShowAcceptChanges justify="flex-start" gap="0px" data-testid="show-accept-changes">
+          <RowBetween>
+            <RowFixed>
+              <AlertTriangle size={20} style={{ marginRight: '8px', minWidth: 24 }} />
```
grabbou, line 240: Shouldn't we wrap this with Styled Components to pass that style? Also, looks like `px` is missing for `minWidth`.

src/components/swap/SwapModalFooter.tsx
```diff
@@ -120,46 +154,136 @@ export default function SwapModalFooter({
...
-            disabled={disabledConfirm}
-            style={{ margin: '10px 0 0 0' }}
-            id={InterfaceElementName.CONFIRM_SWAP_BUTTON}
+      <Column gap="md" padding="0 0.5rem">
+        {details.map(({ label, value, color, labelTooltipText }, i) => (
+          <SwapModalDetailRow key={i} label={label} value={value} color={color} labelTooltipText={labelTooltipText} />
+        ))}
+      </Column>
+      {showAcceptChanges ? (
+        <SwapShowAcceptChanges justify="flex-start" gap="0px" data-testid="show-accept-changes">
```
grabbou, line 237: nit: Is the `0px` gap necessary?

src/components/swap/SwapModalFooter.tsx
```diff
@@ -120,46 +154,136 @@ export default function SwapModalFooter({
...
+          <RowBetween>
+            <RowFixed>
+              <AlertTriangle size={20} style={{ marginRight: '8px', minWidth: 24 }} />
+              <ThemedText.DeprecatedMain color={theme.accentAction}>
+                <Trans>Price Updated</Trans>
+              </ThemedText.DeprecatedMain>
+            </RowFixed>
+            <ButtonPrimary
+              style={{ padding: '.5rem', width: 'fit-content', fontSize: '0.825rem', borderRadius: '12px' }}
```
grabbou, line 246: You should most likely wrap this with `styled` and overwrite relevant styles. Otherwise, you're passing inline HTML attribute styles?

grabbou, line 246: nit: Consider replacing `.5rem` with `8px` (base font size is `16px` as far as I am aware - checked with Google Inspector) and fontSize with `13px`. It would be a good opportunity to double-check with Figma, whether these values are correct. I don't recall `13px` being ever used and that's what `0.825rem` equals to.

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -183,26 +183,26 @@ export function ConfirmationModalContent({
...
   return (
-    <Wrapper>
-      <Section>
-        <RowBetween>
-          <Text fontWeight={500} fontSize={16}>
+    <ConfirmationModalWrapper>
+      <AutoColumn gap="0.5rem" padding="1em 0.5rem 0">
+        <Row>
+          <BackArrowIcon style={{ position: 'absolute' }} onClick={onDismiss} data-cy="confirmation-close-icon" />
+          <Text fontWeight={500} fontSize={16} marginX="auto">
```
grabbou, line 190: Sounds like `ThemedText.SubHeader`. Also, what is the purpose of `marginX` here? Centering text? Can we/should we do it with `Row` flex property instead?

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -183,26 +183,26 @@ export function ConfirmationModalContent({
   bottomContent?: () => ReactNode | undefined
 }) {
   return (
-    <Wrapper>
-      <Section>
-        <RowBetween>
-          <Text fontWeight={500} fontSize={16}>
+    <ConfirmationModalWrapper>
+      <AutoColumn gap="0.5rem" padding="1em 0.5rem 0">
```
grabbou, line 187: nit: I already wrote it down somewhere, but I would consider `<AutoColumn gap="8px" padding="16px 8px 0">` much more readable and actually consistent with the rest of the codebase.

grabbou: Screenshot 2023-04-27 at 10 19 48
Then, you can further replace that with `grid="sm"` which is equivallent to `8px`. Before I got moved from `web` team, I wanted to do a find&replace in the entire codebase to get rid of arbitrary values, so we're more aligned with Figma.

grabbou (changes requested):
Thanks for opening up the draft PR! As we discussed, I left some preliminary feedback. Since we're just getting started in bringing the `widgets` code back and porting some code over, I am mostly concerned about ensuring we maintain overall code quality. My comments/requests are mostly about things I consider inconsistent or something we can improve going forward. 

Feel free to respond/dismiss/agree/disagree, let's keep it an open discussion because we need to come up with a way to work not just with this PR, but others in the future too (will only get worse, as dependencies across different components kick-in).

src/components/swap/SwapModalFooter.tsx
just-toby: it looks like i can just use an existing `Button` component we have in this codebase

cartcrom (commented):
design issue i found:

Long token name (could also be caused if there were trailing decimals):
image

token amount should be right-text-aligned & I think the exchange rate should still be one line at this length

cartcrom (changes requested):
love the new design, noticed two small css things:

| current | design |
| ---- | ---- |
| image | image | 

font weight is 600 instead of 500
\+ we need 4px margin between "you pay" and the amount

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -50,6 +50,14 @@ const StyledLogo = styled.img`
   margin-left: 6px;
 `
 
+const ConfirmationModalContentWrapper = styled(AutoColumn)`
+  padding: 0 0 12px 0;
```
tinaszheng, line 54: can just do `padding-bottom: 12px`

src/components/swap/SwapModalDetailRow.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { Label } from './SwapModalHeaderAmount'
+
+const Value = styled.span<{ color?: string }>`
+  color: ${({ color, theme }) => color ?? theme.textPrimary};
+  text-align: end;
+`
+
+const DetailValue = styled(Value)`
+  max-width: 45%;
+  overflow-wrap: break-word;
```
tinaszheng, line 16: can these two be consolidated?

src/components/swap/SwapModalDetailRow.tsx
```diff
@@ -0,0 +1,41 @@
...
+`
+
+interface DetailProps {
+  label: string | ReactNode
+  value: string | ReactNode
```
tinaszheng, line 21: same here

src/components/swap/SwapModalFooter.tsx
```diff
@@ -121,46 +171,130 @@ export default function SwapModalFooter({
...
-        >
-          <ButtonError
-            data-testid="confirm-swap-button"
-            onClick={onConfirm}
-            disabled={disabledConfirm}
-            style={{ margin: '10px 0 0 0' }}
-            id={InterfaceElementName.CONFIRM_SWAP_BUTTON}
+      <DetailsContainer gap="md">
+        {details.map(({ label, value, color, labelTooltipText }, i) => (
+          <SwapModalDetailRow key={i} label={label} value={value} color={color} labelTooltipText={labelTooltipText} />
```
tinaszheng, line 245: `key` here should not be the index because the order of these things may change in the future if something doesn't have price impact - `label` should be good enough for the React key

src/components/swap/SwapModalFooter.tsx
```diff
@@ -121,46 +171,130 @@ export default function SwapModalFooter({
...
-          </ButtonError>
-        </TraceEvent>
-        {swapErrorMessage ? <SwapCallbackError error={swapErrorMessage} /> : null}
-      </AutoRow>
+            <ButtonError
+              data-testid="confirm-swap-button"
+              onClick={onConfirm}
+              disabled={disabledConfirm}
+              style={{ margin: '10px 0 0 0' }}
+              $borderRadius="12px"
```
tinaszheng, line 286: can you make a new styled component that wraps `ButtonError` and has these attributes?

src/components/swap/SwapModalFooter.tsx
```diff
@@ -121,46 +171,130 @@ export default function SwapModalFooter({
...
+              id={InterfaceElementName.CONFIRM_SWAP_BUTTON}
+            >
+              <Text fontSize={20} fontWeight={600}>
```
tinaszheng, line 289: can this be `HeadlineSmall` instead

src/components/swap/SwapModalHeader.tsx
```diff
@@ -1,216 +1,87 @@
...
-  left: calc(50% - 16px);
-  display: flex;
-  justify-content: center;
-  align-items: center;
-  background-color: ${({ theme }) => theme.backgroundSurface};
-  border: 4px solid;
-  border-color: ${({ theme }) => theme.backgroundModule};
-  z-index: 2;
+const RuleWrapper = styled.div`
+  margin: 16px 2px 24px;
```
tinaszheng, line 13: same here on the 3-item thing

src/components/swap/SwapModalHeader.tsx
```diff
@@ -1,216 +1,87 @@
...
-    trade.inputAmount.currency.chainId === trade.outputAmount.currency.chainId
-      ? trade.inputAmount.currency.chainId
-      : undefined,
-  response,
-  token_in_symbol: trade.inputAmount.currency.symbol,
-  token_out_symbol: trade.outputAmount.currency.symbol,
-  price_update_basis_points: priceUpdate,
-})
+const HeaderContainer = styled(AutoColumn)`
+  margin: 16px 0 0 0;
```
tinaszheng, line 17:
```suggestion
  margin-top: 16px;
```

src/components/swap/SwapModalHeaderAmount.tsx
```diff
@@ -0,0 +1,69 @@
...
+  }
+  return (
+    <ThemedText.HeadlineLarge fontWeight={500} {...textProps}>
+      {children}
+    </ThemedText.HeadlineLarge>
+  )
+}
+
+interface AmountProps {
+  field: 'input' | 'output'
```
tinaszheng, line 34:
```suggestion
import { Field } from @/state/swap/actions

  field: Field
```

src/components/swap/SwapModalHeaderAmount.tsx
```diff
@@ -0,0 +1,69 @@
...
+          {tooltipText ? (
+            <MouseoverTooltip text={tooltipText}>
+              <Label cursor="help">{label}</Label>
+            </MouseoverTooltip>
+          ) : (
+            <Label>{label}</Label>
+          )}
+        </ThemedText.BodySecondary>
+        <Column gap="xs">
+          <ResponsiveHeadline color="primary" data-testid={`${field}-amount`}>
```
tinaszheng, line 60: you might want to update the ids in the tests if you use the Field enum

tinaszheng (changes requested):
mostly looks good, just some nits!

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -84,18 +72,24 @@ export default function ConfirmSwapModal({
         swapQuoteReceivedDate={swapQuoteReceivedDate}
         fiatValueInput={fiatValueInput}
         fiatValueOutput={fiatValueOutput}
+        shouldLogModalCloseEvent={shouldLogModalCloseEvent}
+        setShouldLogModalCloseEvent={setShouldLogModalCloseEvent}
```
tinaszheng, line 76: why are these now in the ModalBottom instead of in the header

src/components/swap/SwapModalDetailRow.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { ReactNode } from 'react'
+import styled from 'styled-components/macro'
+import { ThemedText } from 'theme'
+
+import { Label } from './SwapModalHeaderAmount'
+
+const Value = styled.span<{ color?: string }>`
+  color: ${({ color, theme }) => color ?? theme.textPrimary};
+  text-align: end;
+  max-width: 45%;
```
tinaszheng, line 12: if this component has a max-width then it should be a `div` and not a `span` - spans are for non-layout wrappers

src/components/swap/SwapModalFooter.tsx
```diff
@@ -1,30 +1,44 @@
...
-import { AutoRow } from '../Row'
-import { SwapCallbackError } from './styleds'
+import { ButtonError, SmallButtonPrimary } from '../Button'
+import { AutoRow, RowBetween, RowFixed } from '../Row'
+import { SwapCallbackError, SwapShowAcceptChanges } from './styleds'
+import { SwapModalDetailRow } from './SwapModalDetailRow'
 
 interface AnalyticsEventProps {
-  trade: InterfaceTrade<Currency, Currency, TradeType>
+  trade: InterfaceTrade<Currency, Currency, TradeType> | undefined
```
tinaszheng, line 41: why are u allowing `undefined` trades here now

src/components/swap/SwapModalFooter.tsx
```diff
@@ -111,8 +161,12 @@ export default function SwapModalFooter({
   swapQuoteReceivedDate,
   fiatValueInput,
   fiatValueOutput,
+  shouldLogModalCloseEvent,
+  setShouldLogModalCloseEvent,
+  showAcceptChanges,
+  onAcceptChanges,
 }: {
-  trade: InterfaceTrade<Currency, Currency, TradeType>
+  trade: InterfaceTrade<Currency, Currency, TradeType> | undefined
```
tinaszheng, line 169: shouldn't trade always be defined by time you open the review modal?

src/components/swap/SwapModalHeader.tsx
```diff
@@ -1,216 +1,88 @@
...
-  setShouldLogModalCloseEvent,
   allowedSlippage,
-  recipient,
-  showAcceptChanges,
-  onAcceptChanges,
 }: {
-  trade: InterfaceTrade<Currency, Currency, TradeType>
-  shouldLogModalCloseEvent: boolean
-  setShouldLogModalCloseEvent: (shouldLog: boolean) => void
+  trade: InterfaceTrade<Currency, Currency, TradeType> | undefined
```
tinaszheng, line 36: same here

src/components/swap/SwapModalHeader.tsx
```diff
@@ -1,216 +1,88 @@
...
-  useEffect(() => {
-    if (shouldLogModalCloseEvent && showAcceptChanges) {
-      sendAnalyticsEvent(
-        SwapEventName.SWAP_PRICE_UPDATE_ACKNOWLEDGED,
-        formatAnalyticsEventProperties(trade, priceUpdate, SwapPriceUpdateUserResponse.REJECTED)
-      )
-    }
-    setShouldLogModalCloseEvent(false)
-  }, [shouldLogModalCloseEvent, showAcceptChanges, setShouldLogModalCloseEvent, trade, priceUpdate])
+  const estimateMessage = useMemo(() => {
```
tinaszheng, line 42: you don't need to prepare the component up here and then wrap it with a `useMemo` just use the ternary inside the React tree

tinaszheng: eitehr that or just make it its own component

src/components/swap/SwapModalHeaderAmount.tsx
```diff
@@ -0,0 +1,70 @@
...
+  color: ${({ theme }) => theme.textSecondary};
+  margin-right: 8px;
+`
+
+const ResponsiveHeadline = ({ children, ...textProps }: PropsWithChildren<TextProps>) => {
+  const { width } = useWindowSize()
+  if (width && width < BREAKPOINTS.xs) {
+    return <ThemedText.HeadlineMedium {...textProps}>{children}</ThemedText.HeadlineMedium>
+  }
+  return (
```
tinaszheng, line 27: nit: newline above

src/components/swap/SwapModalHeaderAmount.tsx
```diff
@@ -0,0 +1,70 @@
...
+      {children}
+    </ThemedText.HeadlineLarge>
+  )
+}
+
+interface AmountProps {
+  field: Field
+  tooltipText?: ReactNode
+  label: string
+  amount: CurrencyAmount<Currency> | undefined
```
tinaszheng, line 38: how can amount be undefined

src/components/swap/SwapModalHeaderAmount.tsx
```diff
@@ -0,0 +1,70 @@
...
+            </MouseoverTooltip>
+          ) : (
+            <Label>{label}</Label>
+          )}
+        </ThemedText.BodySecondary>
+        <Column gap="xs">
+          <ResponsiveHeadline color="primary" data-testid={`${field}-amount`}>
+            {formattedAmount} {amount?.currency.symbol}
+          </ResponsiveHeadline>
+          {usdAmount && <ThemedText.BodySmall color="textTertiary">${usdAmount.toFixed(2)}</ThemedText.BodySmall>}
```
tinaszheng, line 64: outside the scope of this PR but we should be using the formatters from conedison for string formatting

tinaszheng (changes requested):

src/components/swap/SwapModalDetailRow.tsx
```diff
@@ -0,0 +1,38 @@
+import Row from 'components/Row'
+import { MouseoverTooltip } from 'components/Tooltip'
+import { ReactNode } from 'react'
+import styled from 'styled-components/macro'
+import { ThemedText } from 'theme'
+
+import { Label } from './SwapModalHeaderAmount'
+
+const Value = styled.span<{ color?: string }>`
```
grabbou: Please wrap ThemedText appropriately and pass `color` prop as string to it. Note that `Text` is a div, so it automatically resolves @tinaszheng comment.

src/components/swap/SwapModalDetailRow.tsx
```diff
@@ -0,0 +1,38 @@
...
+  value: ReactNode
+  color?: string
+  labelTooltipText?: string
+}
+
+export function SwapModalDetailRow({ label, value, color, labelTooltipText }: DetailProps) {
+  return (
+    <ThemedText.BodySmall>
+      <Row align="flex-start" justify="space-between">
+        {labelTooltipText ? (
```
grabbou, line 27: Conditional is not required. You can leverage `disabled` property of `MouseoverTooltip` and make it disabled when `text` is undefined.

src/components/swap/SwapModalFooter.tsx
grabbou, line 41: Please revert this change and make sure that this component is not rendered when `trade` is null.

just-toby: the UI was flickering because trade was briefly undefined, i think because the quote was reloading. this only happened when you waited long enough or switched focus back to the window.

src/components/swap/SwapModalFooter.tsx
```diff
@@ -63,7 +77,22 @@ const formatRoutesEventProperties = (routes: RoutingDiagramEntry[]) => {
...
+    trade.inputAmount.currency.chainId === trade.outputAmount.currency.chainId
+      ? trade.inputAmount.currency.chainId
+      : undefined,
+  response,
+  token_in_symbol: trade.inputAmount.currency.symbol,
+  token_out_symbol: trade.outputAmount.currency.symbol,
+  price_update_basis_points: priceUpdate,
+})
+
+const formatSwapButtonClickEventProperties = ({
```
grabbou, line 95: Can these two formatters be moved out to a separate file, such as utility?

src/components/swap/SwapModalFooter.tsx
grabbou: +1. Let's make sure we do not pass undefined/nulls as they're confusing when it comes to `trade`.

src/components/swap/SwapModalHeaderAmount.tsx
just-toby: amount is undefined if the trade is undefined while the modal is still open - handling it here helps us handle that moment without the modal changing height. we show a placeholder here instead

src/components/swap/SwapModalFooter.tsx
```diff
@@ -121,46 +175,135 @@ export default function SwapModalFooter({
...
+    if (trade && shouldLogModalCloseEvent && showAcceptChanges) {
+      sendAnalyticsEvent(
+        SwapEventName.SWAP_PRICE_UPDATE_ACKNOWLEDGED,
+        formatSwapPriceUpdatedEventProperties(trade, priceUpdate, SwapPriceUpdateUserResponse.REJECTED)
+      )
+    }
+    setShouldLogModalCloseEvent(false)
+  }, [shouldLogModalCloseEvent, showAcceptChanges, setShouldLogModalCloseEvent, trade, priceUpdate])
+
+  const details = useMemo(() => {
```
grabbou, line 210: I think we should remove this and generally avoid defining content of components as arrays as long as they're not dynamic / coming from the API.

Instead, we should explicitly define each `detail` in JSX, such as:
```
<SwapModalDetailRow label={""} value={""} />
```
and so on.

That gives us more control and flexibility as well as it's more light-weight and React friendly.

src/components/swap/SwapModalFooter.tsx
```diff
@@ -121,46 +175,135 @@ export default function SwapModalFooter({
...
+    const label = `${trade?.executionPrice.baseCurrency?.symbol} `
+    const labelInverted = `${trade?.executionPrice.quoteCurrency?.symbol}`
+    const formattedPrice = formatTransactionAmount(priceToPreciseFloat(trade?.executionPrice))
+    const details: Array<DetailItem> = [
+      { label: t`Exchange rate`, value: `${'1 ' + labelInverted + ' = ' + formattedPrice ?? '-'} ${label}` },
+      {
+        label: t`Network fee`,
+        value: `~${formatCurrencyAmount(trade?.gasUseEstimateUSD, NumberType.FiatGasPrice)}`,
```
grabbou, line 218: When converting to a static JSX content, please make sure to not call `formatCurrencyAmount` when `gasUseEstimateUSD` is false.

src/components/swap/SwapModalFooter.tsx
```diff
@@ -121,46 +175,135 @@ export default function SwapModalFooter({
...
-        >
-          <ButtonError
-            data-testid="confirm-swap-button"
-            onClick={onConfirm}
-            disabled={disabledConfirm}
-            style={{ margin: '10px 0 0 0' }}
-            id={InterfaceElementName.CONFIRM_SWAP_BUTTON}
+      <DetailsContainer gap="md">
+        {details.map(({ label, value, color, labelTooltipText }) => (
+          <SwapModalDetailRow
```
grabbou, line 249: I don't think we need this to live in a separate file / be a separate component. If you implement my aforementioned request to make this explicit, you can just write the whole body yourself. That way, you don't need to hardcode a bunch of conditionals in `SwapModalDetailRow` to factor in for different detail rows. 

Please have a look at `MaxSlippageSettings`, `TransactionDeadlineSettings`, `RouterPreferenceSettings` for inspiration how I did it for Settings menu (explicit, manual over array) and `Settings` component (top-level) for how I organised that into a single and readable component.

src/components/swap/SwapModalFooter.tsx
```diff
@@ -121,46 +175,135 @@ export default function SwapModalFooter({
...
+            key={label}
+            label={label}
+            value={value}
+            color={color}
+            labelTooltipText={labelTooltipText}
+          />
+        ))}
+      </DetailsContainer>
+      {showAcceptChanges ? (
+        <SwapShowAcceptChanges justify="flex-start" data-testid="show-accept-changes">
```
grabbou, line 259: nit: Isn't `flex-start` the default? Also, `SwapShowAcceptChanges` is a styled component, so you can just write it as a CSS rule, instead of using a prop. It's a bit more idiomatic in my opinion.

src/components/swap/SwapModalFooter.tsx
```diff
@@ -121,46 +175,135 @@ export default function SwapModalFooter({
...
-          </ButtonError>
-        </TraceEvent>
-        {swapErrorMessage ? <SwapCallbackError error={swapErrorMessage} /> : null}
-      </AutoRow>
+            <ConfirmButton
+              data-testid="confirm-swap-button"
+              onClick={onConfirm}
+              disabled={disabledConfirm}
+              $borderRadius="12px"
+              id={InterfaceElementName.CONFIRM_SWAP_BUTTON}
```
grabbou, line 296: Why do you need both `id` and `data-testid`?

src/components/swap/SwapModalFooter.tsx
```diff
@@ -121,46 +175,135 @@ export default function SwapModalFooter({
...
-            </Text>
-          </ButtonError>
-        </TraceEvent>
-        {swapErrorMessage ? <SwapCallbackError error={swapErrorMessage} /> : null}
-      </AutoRow>
+            <ConfirmButton
+              data-testid="confirm-swap-button"
+              onClick={onConfirm}
+              disabled={disabledConfirm}
+              $borderRadius="12px"
```
grabbou, line 295: Why `borderRadius` is prefixed here with `$`?  Also, its value is hardcoded and this is Styled Component, just define it inside `styled` macro.

src/components/swap/SwapModalHeader.tsx
```diff
@@ -1,216 +1,88 @@
...
-  response,
-  token_in_symbol: trade.inputAmount.currency.symbol,
-  token_out_symbol: trade.outputAmount.currency.symbol,
-  price_update_basis_points: priceUpdate,
-})
+const HeaderContainer = styled(AutoColumn)`
+  margin-top: 16px;
+`
+
+const Rule = styled.hr<{ padded?: true; scrollingEdge?: 'top' | 'bottom' }>`
```
grabbou, line 20: Is this component copied from somewhere? Looks like it. Please note that while doing so, it's always good to look at the properties that such component has and whether we are going to need them.

In your particular example, you're not passing `padded` and `scrollingEdge` to it at all, just rendering `<Rule />`. In that case, you should just remove unnecessary properties / configurability and define a simple component.

src/components/swap/SwapModalHeader.tsx
```diff
@@ -1,216 +1,88 @@
...
+  border: none;
+  border-bottom: 1px solid ${({ theme }) => theme.backgroundOutline};
```
grabbou, line 22: Also, consider moving `Divider` from `components/Settings` (I did it in my previous PR) into reusable components directory (`theme/components`) so that you don't have to create a new one but we can reuse instead. It has all the same properties. 

As a rule of thumb, especially when copying from another codebase, I always do a search over existing codebase to look for potential synergies. Often times, we have something that can be reused, leading to smaller code debt at the end of the day.

src/components/swap/SwapModalHeader.tsx
```diff
@@ -1,216 +1,88 @@
...
-  margin-bottom: -18px;
-  left: calc(50% - 16px);
-  display: flex;
-  justify-content: center;
-  align-items: center;
-  background-color: ${({ theme }) => theme.backgroundSurface};
-  border: 4px solid;
-  border-color: ${({ theme }) => theme.backgroundModule};
-  z-index: 2;
+const RuleWrapper = styled.div`
```
grabbou, line 12: You don't need both  `RuleWrapper` and `Rule` (it's only child). Just merge properties together or pass padding to the parent.

src/components/swap/SwapModalHeader.tsx
grabbou: Agree. Please make sure to do that in this and other places.

src/components/swap/SwapModalHeader.tsx
```diff
@@ -1,216 +1,88 @@
...
-    if (shouldLogModalCloseEvent && showAcceptChanges) {
-      sendAnalyticsEvent(
-        SwapEventName.SWAP_PRICE_UPDATE_ACKNOWLEDGED,
-        formatAnalyticsEventProperties(trade, priceUpdate, SwapPriceUpdateUserResponse.REJECTED)
-      )
-    }
-    setShouldLogModalCloseEvent(false)
-  }, [shouldLogModalCloseEvent, showAcceptChanges, setShouldLogModalCloseEvent, trade, priceUpdate])
+  const estimateMessage = useMemo(() => {
+    return trade?.tradeType === TradeType.EXACT_INPUT ? (
```
grabbou: Once `trade` is made required, we can safely drop `?`.

src/components/swap/SwapModalHeader.tsx
```diff
@@ -1,216 +1,88 @@
...
-            </Trans>
-          </ThemedText.DeprecatedMain>
-        </AutoColumn>
-      ) : null}
-    </AutoColumn>
+    <HeaderContainer gap="sm">
+      <Column gap="lg">
+        <SwapModalHeaderAmount
+          field={Field.INPUT}
+          label={t`You pay`}
```
grabbou, line 46: Can you pass label as component? `label={<Trans>...</Trans>}`? This is more common pattern in our codebase.

src/components/swap/SwapModalHeaderAmount.tsx
grabbou: I actually think since we're rewriting that file, it's a good thing to do.

src/components/swap/SwapModalHeaderAmount.tsx
```diff
@@ -0,0 +1,70 @@
...
+        </Column>
+      </Column>
+      <CurrencyLogo currency={amount?.currency} size="36px" />
```
grabbou, line 69: Isn't it better to do {amount?.currency ? <CurrencyLogo />}? We don't want to render component for undefined currency/amount, right?

src/theme/components/index.tsx
```diff
@@ -467,17 +467,19 @@ export const SpinnerSVG = styled.svg`
...
-    <BackArrowLink to={to}>
-      <ArrowLeft />
-    </BackArrowLink>
+    <StyledInternalLink to={to}>
+      <BackArrowIcon />
+    </StyledInternalLink>
   )
 }
 
+export const BackArrowIcon = styled(ArrowLeft)`
```
grabbou, line 478: Where is `BackArrowIcon` used? I can't find a reference in the diff.

src/theme/components/index.tsx
```diff
@@ -467,17 +467,19 @@ export const SpinnerSVG = styled.svg`
...
+  color: ${({ theme }) => theme.textPrimary};
+  cursor: pointer;
```
grabbou, line 480: I don't think `cursor: pointer` should be set on non-clickable elements by default. You can safely remove it, because `StyledInternalLink` that you wrap `BackArrowIcon` with has `cursor: pointer` already set up and that value will be propagated to the children too.

src/theme/index.tsx
```diff
@@ -67,6 +72,8 @@ const fonts = {
   code: 'courier, courier new, serif',
 }
 
+export type Gap = keyof DefaultTheme['grids']
```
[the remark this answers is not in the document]
grabbou: Also, are you sure this code does work? To me, the value of that is silently `any`, because you have now created a circular reference - `DefaultTheme` defined in `styled.d.ts` imports from `theme/index.tsx` to check for `typeof`. I am not sure if TypeScript can handle that.

grabbou (changes requested):
Thanks for the PR! Left a bunch of nits and some requests for changes. Happy to follow up on Slack.

src/theme/index.tsx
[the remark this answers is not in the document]
just-toby: yes it was working like this, but i'll make the change

image

src/components/Row/index.tsx
```diff
@@ -24,7 +23,7 @@ const Row = styled(Box)<{
   padding: ${({ padding }) => padding};
   border: ${({ border }) => border};
   border-radius: ${({ borderRadius }) => borderRadius};
-  gap: ${({ gap, theme }) => gap && (theme.grids[gap as Gap] || gap)};
+  gap: ${({ gap, theme }) => (gap && theme.grids[gap as Gap]) ?? gap};
```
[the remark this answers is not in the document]
just-toby: this was just a merge conflict since i made the same change in parallel

src/components/swap/SwapModalFooter.tsx
tinaszheng: ah yes that is a bug. can you rebase this after i land https://github.com/Uniswap/interface/pull/6499 ? trade should not be undefined once this fix goes on.

src/components/swap/SwapModalFooter.tsx
just-toby: the prefix prevents the prop from being passed to the dom element, which is sometimes needed. in this case `BaseButtonProps` already had this prop name, so i'm just adding a usage. i didn't create the prop. also, i'm using the prop instead of CSS because the radius doesn't get applied correctly unless I use this prop.

src/components/swap/SwapModalFooter.tsx
just-toby: i'm not sure, i didn't add these and i'd prefer not to remove them in this PR

src/theme/components/index.tsx
just-toby: it was used before the design changed, but this refactor is the suggestion you made in your first review

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -59,21 +57,11 @@ export default function ConfirmSwapModal({
   }, [isOpen, onDismiss])
 
   const modalHeader = useCallback(() => {
```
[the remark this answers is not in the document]
just-toby: noted in my todo list

src/components/swap/SwapModalFooter.tsx
grabbou: If it wasn't undefined previously, it means that the UI was flickering already before you started working on the PR? In that case, I would consider that as non-issue and just move on without `undefined`. If after working here, there's a bug in the code to be fixed instead.

src/components/swap/SwapModalHeaderAmount.tsx
grabbou: > amount is undefined if the trade is undefined while the modal is still open 

When can this happen? Can we solve this in a different way?

src/components/Row/index.tsx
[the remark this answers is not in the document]
grabbou: That doesn't answer my question - there should be no diff here. I think the current version on main is cleaner.

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -277,7 +285,7 @@ function L2Content({
           )}
         </ConfirmedIcon>
         <AutoColumn gap="md" justify="center">
-          <Text fontWeight={500} fontSize={20} textAlign="center">
+          <ThemedText.SubHeaderLarge textAlign="center">
```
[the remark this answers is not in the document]
just-toby: i want it to be center here, so i'm not sure what you mean

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -300,7 +308,7 @@ function L2Content({
           ) : (
             <div style={{ height: '17px' }} />
           )}
-          <Text color={theme.textTertiary} style={{ margin: '20px 0 0 0' }} fontSize="14px">
+          <ThemedText.SubHeaderSmall color={theme.textTertiary} marginTop="20px">
```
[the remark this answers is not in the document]
just-toby: in this case i am choosing to use the prop

src/theme/components/index.tsx
grabbou: Makes sense, but my question is why do we need to `export` it outside of this module?

src/components/Row/index.tsx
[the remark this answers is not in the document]
just-toby: i already reverted this change

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -50,6 +50,14 @@ const StyledLogo = styled.img`
   margin-left: 6px;
 `
 
+const ConfirmationModalContentWrapper = styled(AutoColumn)`
+  padding-bottom: 12px;
+`
+
+const ConfirmationModalTopContentWrapper = styled(AutoColumn)`
```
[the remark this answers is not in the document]
just-toby: i already fixed this

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -74,15 +82,15 @@ function ConfirmationPendingContent({
           <CustomLightSpinner src={Circle} alt="loader" size={inline ? '40px' : '90px'} />
         </ConfirmedIcon>
         <AutoColumn gap="md" justify="center">
-          <Text fontWeight={500} fontSize={20} color={theme.textPrimary} textAlign="center">
+          <ThemedText.SubHeaderLarge color={theme.textPrimary} textAlign="center">
             <Trans>Waiting for confirmation</Trans>
-          </Text>
-          <Text fontWeight={600} fontSize={16} color={theme.textPrimary} textAlign="center">
+          </ThemedText.SubHeaderLarge>
+          <ThemedText.SubHeader color={theme.textPrimary} textAlign="center">
```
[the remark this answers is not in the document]
just-toby: this is done

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -287,7 +295,7 @@ function L2Content({
             ) : (
               <Trans>Error</Trans>
             )}
-          </Text>
+          </ThemedText.SubHeaderLarge>
           <Text fontWeight={400} fontSize={16} textAlign="center">
```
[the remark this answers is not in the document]
just-toby: this is done

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -311,14 +319,14 @@ function L2Content({
                 </span>
               </div>
             )}
-          </Text>
+          </ThemedText.SubHeaderSmall>
           <ButtonPrimary onClick={onDismiss} style={{ margin: '4px 0 0 0' }}>
-            <Text fontWeight={500} fontSize={20}>
+            <ThemedText.SubHeaderLarge fontWeight={500} fontSize={20}>
```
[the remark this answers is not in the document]
just-toby: i removed the extra props from the ThemedTexts

src/components/swap/SwapModalDetailRow.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { MouseoverTooltip } from 'components/Tooltip'
+import { ReactNode } from 'react'
+import styled from 'styled-components/macro'
+import { ThemedText } from 'theme'
+
+import { Label } from './SwapModalHeaderAmount'
+
+const Value = styled.span<{ color?: string }>`
+  color: ${({ color, theme }) => color ?? theme.textPrimary};
+  text-align: end;
```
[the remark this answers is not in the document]
just-toby: i changed this

src/components/swap/SwapModalDetailRow.tsx
just-toby: i made this change

src/components/swap/SwapModalDetailRow.tsx
just-toby: this is updated

src/components/TransactionConfirmationModal/index.tsx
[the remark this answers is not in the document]
grabbou: Sorry, for some reason I think I saw `textAlign=left` lol

src/components/swap/SwapModalFooter.tsx
just-toby: this is reverted

src/components/swap/SwapModalFooter.tsx
just-toby: done

src/components/swap/SwapModalFooter.tsx
just-toby: i inlined this UI

src/theme/components/index.tsx
just-toby: https://github.com/Uniswap/interface/blob/main/src/nft/components/collection/TransactionCompleteModal.tsx#L378

just-toby: oh wait, that's a different icon! i think i can remove the export

src/components/swap/SwapModalHeaderAmount.tsx
```diff
@@ -0,0 +1,76 @@
...
+  let formattedAmount = formatCurrencyAmount(amount, NumberType.TokenTx)
+  if (formattedAmount.length > MAX_AMOUNT_STR_LENGTH) {
+    formattedAmount = formatCurrencyAmount(amount, NumberType.SwapTradeAmount)
+  }
+
+  return (
+    <Row align="center" justify="space-between" gap="md">
+      <Column gap="xs">
+        <ThemedText.BodySecondary>
+          {tooltipText ? (
```
grabbou, line 54: Let's always render MouseoverTooltip and pass `disabled` property when needed instead. The reason for this is simple. MouseOverTooltip is not just a simple component, it has some wrappers inside. So, when we render `<Label />` alone, we're increasing chances of UI glitch, b/c MouseoverTooltip may do something with paddings/margins that changes the appearance.

src/components/swap/SwapModalHeaderAmount.tsx
```diff
@@ -0,0 +1,76 @@
...
+            <MouseoverTooltip text={tooltipText}>
+              <Label cursor="help">{label}</Label>
+            </MouseoverTooltip>
+          ) : (
+            <Label>{label}</Label>
+          )}
+        </ThemedText.BodySecondary>
+        <Column gap="xs">
+          <ResponsiveHeadline color="primary" data-testid={`${field}-amount`}>
```
grabbou, line 63: What kind of color is `primary`? Are you sure it shouldn't be `textPrimary`?

src/components/swap/SwapModalFooter.tsx
```diff
@@ -121,46 +75,155 @@ export default function SwapModalFooter({
...
+  setShouldLogModalCloseEvent: (shouldLog: boolean) => void
+  showAcceptChanges: boolean
+  onAcceptChanges: () => void
 }) {
   const transactionDeadlineSecondsSinceEpoch = useTransactionDeadline()?.toNumber() // in seconds since epoch
   const isAutoSlippage = useUserSlippageTolerance()[0] === 'auto'
   const [routerPreference] = useRouterPreference()
-  const routes = getRoutingDiagramEntries(trade)
+  const routes = trade ? getRoutingDiagramEntries(trade) : []
+  const [lastExecutionPrice, setLastExecutionPrice] = useState(trade?.executionPrice)
```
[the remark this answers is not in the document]
just-toby: removed null check

src/components/swap/SwapModalFooter.tsx
```diff
@@ -121,46 +75,155 @@ export default function SwapModalFooter({
...
+  const [priceUpdate, setPriceUpdate] = useState<number>()
+  const theme = useTheme()
+  const { chainId } = useWeb3React()
+  const nativeCurrency = useNativeCurrency(chainId)
+
+  useEffect(() => {
+    if (trade && lastExecutionPrice && !trade.executionPrice.equalTo(lastExecutionPrice)) {
```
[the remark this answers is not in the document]
just-toby: removed null check

src/components/swap/SwapModalFooter.tsx
```diff
@@ -121,46 +75,155 @@ export default function SwapModalFooter({
...
+      setPriceUpdate(getPriceUpdateBasisPoints(lastExecutionPrice, trade.executionPrice))
+      setLastExecutionPrice(trade.executionPrice)
+    }
+  }, [lastExecutionPrice, setLastExecutionPrice, trade])
+
+  useEffect(() => {
+    if (trade && shouldLogModalCloseEvent && showAcceptChanges) {
```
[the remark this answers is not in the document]
just-toby: removed null check

src/components/swap/SwapModalFooter.tsx
```diff
@@ -121,46 +75,155 @@ export default function SwapModalFooter({
...
+          </Row>
+        </ThemedText.BodySmall>
+        <ThemedText.BodySmall>
+          <Row align="flex-start" justify="space-between">
+            <MouseoverTooltip text={t`The impact your trade has on the market price of this pool.`}>
+              <Label cursor="help">
+                <Trans>Price impact</Trans>
+              </Label>
+            </MouseoverTooltip>
+            <DetailRowValue color={trade ? getPriceImpactWarning(trade.priceImpact) : undefined}>
```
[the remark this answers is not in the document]
just-toby: removed null check

src/components/swap/SwapModalFooter.tsx
```diff
@@ -121,46 +75,155 @@ export default function SwapModalFooter({
...
+            <Label>
+              <Trans>Exchange rate</Trans>
+            </Label>
+            <DetailRowValue>{`${'1 ' + labelInverted + ' = ' + formattedPrice ?? '-'} ${label}`}</DetailRowValue>
+          </Row>
+        </ThemedText.BodySmall>
+        <ThemedText.BodySmall>
+          <Row align="flex-start" justify="space-between">
+            <MouseoverTooltip
+              text={t`The fee paid to miners who process your transaction. This must be paid in ${nativeCurrency.symbol}.`}
```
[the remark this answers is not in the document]
just-toby: updated

src/components/swap/SwapModalFooter.tsx
```diff
@@ -121,46 +75,155 @@ export default function SwapModalFooter({
...
-            disabled={disabledConfirm}
-            style={{ margin: '10px 0 0 0' }}
-            id={InterfaceElementName.CONFIRM_SWAP_BUTTON}
+      <DetailsContainer gap="md">
+        <ThemedText.BodySmall>
+          <Row align="flex-start" justify="space-between">
+            <Label>
+              <Trans>Exchange rate</Trans>
+            </Label>
+            <DetailRowValue>{`${'1 ' + labelInverted + ' = ' + formattedPrice ?? '-'} ${label}`}</DetailRowValue>
```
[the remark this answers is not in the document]
just-toby: sure

src/components/swap/SwapModalFooter.tsx
```diff
@@ -121,46 +75,155 @@ export default function SwapModalFooter({
...
+
+  useEffect(() => {
+    if (trade && shouldLogModalCloseEvent && showAcceptChanges) {
+      sendAnalyticsEvent(
+        SwapEventName.SWAP_PRICE_UPDATE_ACKNOWLEDGED,
+        formatSwapPriceUpdatedEventProperties(trade, priceUpdate, SwapPriceUpdateUserResponse.REJECTED)
+      )
+    }
+    setShouldLogModalCloseEvent(false)
+  }, [shouldLogModalCloseEvent, showAcceptChanges, setShouldLogModalCloseEvent, trade, priceUpdate])
```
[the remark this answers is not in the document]
just-toby: yes, i just moved this to the parent

src/components/swap/SwapModalHeader.tsx
```diff
@@ -1,216 +1,72 @@
...
+                <Trans>
+                  Output is estimated. You will receive at least{' '}
+                  <b>
+                    {trade.minimumAmountOut(allowedSlippage).toSignificant(6)} {trade.outputAmount.currency.symbol}
+                  </b>{' '}
+                  or the transaction will revert.
+                </Trans>
+              </ThemedText.Caption>
+            ) : (
+              <ThemedText.Caption textAlign="left" style={{ width: '100%' }}>
```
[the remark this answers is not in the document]
just-toby: removed

src/components/swap/SwapModalHeaderAmount.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+const ResponsiveHeadline = ({ children, ...textProps }: PropsWithChildren<TextProps>) => {
+  const { width } = useWindowSize()
+
+  if (width && width < BREAKPOINTS.xs) {
+    return <ThemedText.HeadlineMedium {...textProps}>{children}</ThemedText.HeadlineMedium>
+  }
+
+  return (
+    <ThemedText.HeadlineLarge fontWeight={500} {...textProps}>
```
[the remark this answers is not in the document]
just-toby: it's probably from the figma, but i'll just remove it

src/components/swap/SwapModalHeaderAmount.tsx
just-toby: ok, done

src/components/swap/SwapModalHeaderAmount.tsx
just-toby: i just removed this since it wasn't doing anything

grabbou (approved):
LGTM

src/pages/Swap/index.tsx
```diff
@@ -570,22 +570,23 @@ export function Swap({
...
-        onConfirm={handleSwap}
-        swapErrorMessage={swapErrorMessage}
-        onDismiss={handleConfirmDismiss}
-        swapQuoteReceivedDate={swapQuoteReceivedDate}
-        fiatValueInput={fiatValueTradeInput}
-        fiatValueOutput={fiatValueTradeOutput}
-      />
+      {trade && (
+        <ConfirmSwapModal
+          isOpen={showConfirm}
```
tinaszheng, line 575: it's weird that show state of this modal here is controlled both by `trade` and `showConfirm` - can this just be:

```
{trade && showConfirm && <ConfirmSwapModal ... />}
```

and make the modal internally always show if it's rendered?

tinaszheng: it would also save computation by not running the hooks and preparing React components if the modal is not rendered

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -123,13 +123,13 @@ export default function ConfirmSwapModal({
   return (
     <Trace modal={InterfaceModalName.CONFIRM_SWAP}>
       <TransactionConfirmationModal
-        isOpen={isOpen}
+        isOpen={true}
```
tinaszheng, line 126:
```suggestion
        isOpen
```

tinaszheng (approved):

tinaszheng (approved):

PR #6452, opened 2023-04-26
fix: disable bnb network option on uniwallet

JFrankfurt:
_Slack thread:_ https://uniswapteam.slack.com/archives/C04D07TQBT4/p1682535435459049

## Screen capture

| image | image        |

## Test plan
connect to the app with the uniswap wallet and confirm both options are disabled

connect to the app with anything else and confirm all options are enabled

### DIFF
--- a/src/components/NavBar/ChainSelector.tsx
+++ b/src/components/NavBar/ChainSelector.tsx
@@ -76,7 +76,7 @@ export const ChainSelector = ({ leftAlign }: ChainSelectorProps) => {
       <Column paddingX="8">
         {NETWORK_SELECTOR_CHAINS.map((chainId: SupportedChainId) => (
           <ChainSelectorRow
-            disabled={isUniWallet && chainId === SupportedChainId.CELO}
+            disabled={isUniWallet && [SupportedChainId.CELO, SupportedChainId.BNB].includes(chainId)}
             onSelectChain={onSelectChain}
             targetChain={chainId}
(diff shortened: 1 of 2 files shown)


tinaszheng (approved):

src/components/NavBar/ChainSelector.tsx
```diff
@@ -76,7 +76,7 @@ export const ChainSelector = ({ leftAlign }: ChainSelectorProps) => {
       <Column paddingX="8">
         {NETWORK_SELECTOR_CHAINS.map((chainId: SupportedChainId) => (
           <ChainSelectorRow
-            disabled={isUniWallet && chainId === SupportedChainId.CELO}
+            disabled={isUniWallet && !UniWalletSupportedChains.includes(chainId)}
```
grabbou, line 79: I find it generally a bit confusing that this feature - displayed in the UI as a generic "not supported by your wallet" feature - is specific to Uniwallet. In that case, I'd rather see the wording "not supported by Uniwallet". 

Also, it would be great if this feature was expanded in the future into more generic abstraction, especially as we add WalletConnect multi-chain support where all chains become optional and when not present after connecting - it will mean that a give chain is unsupported.

I understand it's out of scope for this PR, so would rather have a follow-up ticket. If you agree, I would appreciate if you could create one!

grabbou (approved):

PR #6455, opened 2023-04-27
ci: Staging deploy GH action definition

lavalamp-:
## Description

This PR introduces a new GitHub action that will run off pushes to the `releases/staging` branch. Specifically it:

- Prepares the build environment
- Builds a "staging" version of the `interface` code (presently only code difference is Sentry reporting)
- Deploys the "staging" version of the `interface` code to a staging Cloudflare Page
- Sends the source maps to Sentry

### QA (ie manual testing)

Once all of the new CI PRs are landed we will test the system end-to-end.


src/utils/env.ts
```diff
@@ -21,7 +21,7 @@ export function isAppUniswapOrg({ hostname }: { hostname: string }): boolean {
 
 export function isSentryEnabled(): boolean {
   // Disable in e2e test environments
-  if (isStagingEnv() || (isProductionEnv() && !isAppUniswapOrg(window.location))) return false
```
lavalamp-, line 24: I'm thinking this is necessary... No point in updating Sentry environment to staging if we're not actually going to do any reporting??

zzmp: yes, I just wonder where else this is being set. Probably in our vercel builds, which we do _not_ want reporting to sentry. I think we may want an additional check on window.location, but that can be done in a follow-up.

zzmp: Could you change this to `(isStagingEnv() && !isAppUniswapStagingOrg(window.location))`? That way we'll only send sentry events from the staging build.

zzmp (changes requested):

.github/workflows/deploy-to-staging.yml
```diff
@@ -0,0 +1,41 @@
...
+on:
+  push:
+    branches:
+      - 'releases/staging'
+
+jobs:
+  deploy-to-staging:
+    runs-on: ubuntu-latest
+    environment:
+      name: release
```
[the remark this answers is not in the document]
lavalamp-: Depends on where we're storing the new Cloudflare credentials! If they're stored in a GH environment titled `staging` then yes sounds good :)

src/utils/env.ts
lavalamp-: Sounds good :)

.github/workflows/deploy-to-staging.yml
```diff
@@ -0,0 +1,41 @@
+name: Deploy to staging
```
zzmp, line 1: rename the file to match other ordinal actions: `2-deploy-to-staging.yml`?

.github/workflows/deploy-to-staging.yml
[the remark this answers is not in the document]
zzmp: yes, I'm going to create an environment called staging for this 🤝

zzmp (changes requested):

.github/workflows/deploy-to-staging.yml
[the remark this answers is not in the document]
zzmp: done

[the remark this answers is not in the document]
lavalamp-: 🙏

.github/workflows/deploy-to-staging.yml
lavalamp-: Done!

zzmp (approved):

PR #6456, opened 2023-04-27
chore: moving nft test fixtures to their own file

JackShort:
Moving the nft test fixtures to their own file to cleanup the test constants file. There are going to be a few more nft fixtures and I feel like moving them to their own file will be cleaner


src/test-utils/nftFixtures.ts
```diff
@@ -0,0 +1,99 @@
+import { NftActivityType, NftStandard, OrderStatus } from 'graphql/data/__generated__/types-and-hooks'
```
cbachmeier, line 1: nit: to mimic our component file structure I would move/name this file `test-utils/nft/constants.ts`

JackShort: yeah i thought about that but idk if we really want to add folders to test-utils considering the breadth isn't very big, and I don't really agree with the constants naming convention as these are more fixtures

JackShort: but I am not really too opinionated about it those were just my thoughts with this current naming convention

zzmp: in case nft usage of test-utils continues to grow, I'd encourage this. I think it's worth refactoring before merging this in.

zzmp (changes requested):

PR #6458, opened 2023-04-27
chore: updating listing marketplace icons

JackShort:
Updates the listing page to use the `getMarketplaceIcon` util function instead of accessing the svgs directly


src/nft/components/profile/list/RoyaltyTooltip.tsx
```diff
@@ -17,16 +17,20 @@ const RoyaltyContainer = styled(Column)`
...
 `
 
-const CollectionIcon = styled(MarketIcon)`
+const CollectionIcon = styled.img`
+  width: 16px;
+  height: 16px;
+  object-fit: cover;
+  outline: 1px solid ${({ theme }) => theme.backgroundInteractive};
+  margin-right: 8px;
   border-radius: 50%;
```
cbachmeier, line 34: to prevent redundancy can you move the shared styles to a `css` object used in both components

src/nft/components/profile/list/utils.ts
```diff
@@ -109,7 +109,7 @@ const getListings = (sellAssets: WalletAsset[]): [CollectionRow[], ListingRow[]]
         )
       ) {
         const newCollectionRow = {
-          images: [asset.asset_contract.image_url, marketplace.icon],
+          images: [asset.asset_contract.image_url],
```
cbachmeier, line 112: nit: this could be refactored as the array is no longer needed, `image: asset.asset_contract.image_url`

cbachmeier (approved):
LGTM minor nits

cmcewen (approved):
would love to see some snapshot tests here so we could see the diff!

JackShort:
> would love to see some snapshot tests here so we could see the diff!

we are currently in the process of adding snapshot tests across all nft pages!

PR #6459, opened 2023-04-27
build: upgrade to webpack 5

zzmp:
## Description

Upgrades the toolchain from Webpack 4 to Webpack 5. This includes some other changes required for the upgrade:
- Upgrading craco to match
- Updating the craco.config.js to deal with new requirements
- Importing JSON files using only default imports

This also includes configuring a vendors bundle, as the CRA/Webpack upgrade changed the bundle splitting behavior. Without configuring a vendors bundle, the initial JS chunk would be significantly larger and would trigger size warnings from workbox. This _should_ be more performant, as the vendors chunk _shouldn't_ change more frequently than our deploy frequency.

_Relevant docs:_ [Design doc](https://www.notion.so/uniswaplabs/Optimizing-interface-build-performance-29adef9205ae44928bac0e04ed50e382#edf3cdae66fa4e299441373bba945024), [Bug bash](https://www.notion.so/uniswaplabs/Bug-Bash-Webpack-5-Migration-8754184029394129811ecd640f1103d0)

### QA (ie manual testing)

- [x] N/A

### Automated testing

_Existing tests should continue running fine :)_

- [x] Unit test
- [x] Integration/E2E test


craco.config.cjs
```diff
@@ -69,7 +79,13 @@ module.exports = {
     },
   },
   webpack: {
-    plugins: [new VanillaExtractPlugin({ identifiers: 'short' })],
+    plugins: [
+      // react-markdown requires process, and Webpack 5 does not polyfill, so we provide it explicitly.
+      new ProvidePlugin({ process: 'process/browser' }),
+      // vanilla-extract has poor performance on M1 machines with 'debug' identifiers.
```
grabbou, line 85: nit: It would be great if the comment explained the difference between `debug` and `short`.

craco.config.cjs
```diff
@@ -81,6 +97,15 @@ module.exports = {
             })
           }
 
+          // Increase the cacheable file size (by the ServiceWorker) so that main.<hash>.js is cached.
+          if (plugin instanceof InjectManifest) {
```
grabbou, line 101: Is this a feature (separate PR) or something to provide Webpack 4-like functionality due to change in cacheable file size as of Webpack 5?

craco.config.cjs
```diff
@@ -81,6 +97,15 @@ module.exports = {
...
+            plugin.config.maximumFileSizeToCacheInBytes = 7.5 * 1024 * 1024 // 7.5 MB
```
grabbou, line 102: Do you have context why the cacheable size was changed (e.g. performance/memory) and what does that mean for us, if overwritten like this?

craco.config.cjs
```diff
@@ -100,10 +125,35 @@ module.exports = {
...
         })
 
-      // We're currently on Webpack 4.x which doesn't support the `exports` field in package.json.
-      // Instead, we need to manually map the import path to the correct exports path (eg dist or build folder).
-      // See https://github.com/webpack/webpack/issues/9509.
-      webpackConfig.resolve.alias['@uniswap/conedison'] = '@uniswap/conedison/dist'
+      webpackConfig.resolve.plugins = webpackConfig.resolve.plugins.map((plugin) => {
+        // Allow vanilla-extract in production builds.
+        // See https://sandroroth.com/blog/vanilla-extract-cra#production-build.
+        if (plugin instanceof ModuleScopePlugin) {
```
grabbou, line 131: Could you explain the relation between this and Webpack upgrade?

craco.config.cjs
```diff
@@ -100,10 +125,35 @@ module.exports = {
...
+          plugin.allowedPaths.push(path.join(__dirname, 'node_modules/@vanilla-extract/webpack-plugin'))
+        }
+
+        return plugin
+      })
+
+      const rules = webpackConfig.module.rules[1].oneOf
```
grabbou, line 138: nit: This section is missing a high-level explanation of what it does. I assume it has something to do with `@uniswap/analytics` package, but I would appreciate additional explanation.

craco.config.cjs
```diff
@@ -100,10 +125,35 @@ module.exports = {
...
+          // The class properties transform is required for @uniswap/analytics to build.
+          rule.options.plugins.push('@babel/plugin-proposal-class-properties')
+        }
+        return rule
+      })
+
+      // react-markdown requires path, and Webpack 5 does not polyfill, so we polyfill it.
+      webpackConfig.resolve.fallback = { path: require.resolve('path-browserify') }
+
+      // Ignore failed source mappings to avoid spamming the console.
```
grabbou, line 154: nit: I would appreciate if comment could be updated to mention when this can happen and why we're not interested in warnings

cypress/e2e/service-worker.test.ts
```diff
@@ -53,8 +53,11 @@ describe('Service Worker', () => {
   it('installs a ServiceWorker', () => {
     cy.visit('/', { serviceWorker: true })
       .get('#swap-page')
-      .wait('@NotInstalled', { timeout: 20000 })
-      .window({ timeout: 20000 })
+      .wait('@NotInstalled', {
+        // This is a large file, and takes some time to load.
```
grabbou, line 57: Does the increased timeout means that we're having downgraded performance when it comes to loading a service worker now?

scripts/test-size.js
```diff
@@ -15,42 +14,23 @@ try {
...
 }
 
-// The last recorded size for these assets, as reported by `yarn build`.
-const LAST_SIZE_MAIN_KB = 435
-
-// This is the async-loaded js, called <number>.<hash>.js, with a matching css file.
-const LAST_SIZE_ENTRY_KB = 1442
-
+// The last recorded size for main.<hash>.js, as reported by `yarn build`.
+const SIZE_MAIN_KB = 1790
```
grabbou, line 18: What causes the chunk size to grow from 435 to 1790? Also, that is in KB, previously you mentioned `~7MB` in craco.config.js, so I am confused - what is the correct size here?

scripts/test-size.js
```diff
@@ -15,42 +14,23 @@ try {
...
 const SIZE_TOLERANCE_KB = 10
 
-const jsEntrypoints = entrypoints.filter((entrypoint) => entrypoint.endsWith('js'))
-assert(jsEntrypoints.length === 3)
```
grabbou, line 27: Why we don't want to assert on this anymore?

src/state/claim/hooks.ts
```diff
@@ -1,5 +1,5 @@
 import type { TransactionResponse } from '@ethersproject/providers'
-import { abi as MERKLE_DISTRIBUTOR_ABI } from '@uniswap/merkle-distributor/build/MerkleDistributor.json'
+import MerkleDistributorJSON from '@uniswap/merkle-distributor/build/MerkleDistributor.json'
```
grabbou, line 2: Not related to the PR, but why do we import from `build` folder? Does this package just happen to have a `build` folder or are we actually importing from `build`? Is this due to something else not working properly, such as `package.exports` not being fully supported?

grabbou (changes requested):
Looking good! Great to see our infrastructure moving towards Webpack 5.x finally and dropping some of the workarounds. Left some comments, mostly nits, but I do have some questions to get more context before final approval. Looking forward.

craco.config.cjs
```diff
@@ -100,10 +125,35 @@ module.exports = {
           return true
         })
 
-      // We're currently on Webpack 4.x which doesn't support the `exports` field in package.json.
-      // Instead, we need to manually map the import path to the correct exports path (eg dist or build folder).
-      // See https://github.com/webpack/webpack/issues/9509.
-      webpackConfig.resolve.alias['@uniswap/conedison'] = '@uniswap/conedison/dist'
+      webpackConfig.resolve.plugins = webpackConfig.resolve.plugins.map((plugin) => {
```
cbachmeier, line 128: q: what's the difference between the resolve.plugins and the plugins we mapped through on ln 90-91?

cbachmeier (approved):
LGTM. Checked all pages/styles load as expected including VE global styles. Major flows such as swap, pool, purchase nft, list, work as expected on tenderly

craco.config.cjs
zzmp: This is a required part of the migration.
Webpack 4 split the entry file into entry (very small) and main (initial assets), but Webpack 5 optimizes by serving it all as a single file. As a result, that file is much larger, and so we need to bump the limit on cacheable file size by the workbox InjectManifest plugin.

craco.config.cjs
```diff
@@ -100,10 +125,35 @@ module.exports = {
...
+          rule.resolve = { fullySpecified: false }
+
+          // The class properties transform is required for @uniswap/analytics to build.
+          rule.options.plugins.push('@babel/plugin-proposal-class-properties')
+        }
+        return rule
+      })
+
+      // react-markdown requires path, and Webpack 5 does not polyfill, so we polyfill it.
+      webpackConfig.resolve.fallback = { path: require.resolve('path-browserify') }
```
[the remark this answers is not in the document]
zzmp: No, because they are accessed two different ways. `ProvidePlugin` adds `process` to the global scope, but `path` is not on the global scope - it is still imported as a package.

craco.config.cjs
zzmp: I think that create-react-app 5 has additional restrictions to prevent bad practices, and this is one of them - it prevents you from including files in your build which shouldn't obviously be there.

craco.config.cjs
zzmp: These are resolver-specific plugins, whereas the other plugins have to do with the global pipeline.

src/state/claim/hooks.ts
zzmp: This is due to how our legacy packages are built. They export their abis in build folders. They do not use `package.exports`.

This is much older infra, so `exports` was probably just not considered at the time. I don't think we want to touch these - they are too foundational.

craco.config.cjs
zzmp: The CRA / Webpack 5 configuration no longer splits the entry chunk, so the entry chunk (main.js) is now much larger. Instead of allowing a larger cached file, I'm going to make a vendor bundle to better mimic the Webpack 4 behavior and for better performance.

craco.config.cjs
zzmp: See https://github.com/Uniswap/interface/pull/6459#discussion_r1180898182

cypress/e2e/service-worker.test.ts
zzmp: See https://github.com/Uniswap/interface/pull/6459#discussion_r1180898182

scripts/test-size.js
zzmp: https://github.com/Uniswap/interface/pull/6459#discussion_r1180898182

scripts/test-size.js
zzmp: https://github.com/Uniswap/interface/pull/6459#discussion_r1180898182

craco.config.cjs
```diff
@@ -44,8 +46,13 @@ module.exports = {
     pluginOptions(eslintConfig) {
       return Object.assign(eslintConfig, {
         cache: true,
-        cacheLocation: 'node_modules/.cache/eslint/',
+        cacheLocation: 'node_modules/.cache/.eslintcache',
         ignorePath: '.gitignore',
+        // Clear the create-react-app overrides to use the defaults.
+        formatter: undefined,
```
grabbou, line 52: nit: why was this not required before?

grabbou, line 52: nit: what do you mean by `the defaults` in this context? Craco defaults? Webpack defaults?

craco.config.cjs
```diff
@@ -69,8 +78,16 @@ module.exports = {
     },
   },
   webpack: {
-    plugins: [new VanillaExtractPlugin({ identifiers: 'short' })],
+    plugins: [
+      // react-markdown requires path to be global, and Webpack 5 does polyfill node globals, so we polyfill it.
+      new ProvidePlugin({ process: 'process/browser' }),
+      // vanilla-extract has poor performance on M1 machines with 'debug' identifiers, so we use 'short' instead.
```
grabbou, line 84: nit: https://uniswapteam.slack.com/archives/C04KGMR1ACW/p1682955163281759

grabbou (approved):

craco.config.cjs
zzmp: I've updated the comment to be more explicit. CRA overrides these properties to use the eslint that ships with react-scripts, but doing so invalidates our lint cache.

zzmp: I've updated the comment to be more explicit. CRA overrides these properties to use the eslint that ships with react-scripts.

grabbou (approved):

cbachmeier (approved):

cbachmeier (approved):

cbachmeier (approved):

PR #6460, opened 2023-04-27
feat: [DetailsV2] Add Trait component content

cbachmeier:
## Description
- Add content to trait component on data page
- Hide trait component for assets with 0 traits
- Added randomly generated rarities, counts, and trait floors while waiting on BE to support the new required data
- Display a dynamic graph of trait rarity
  - You can learn more about the graph spec in the Relevant Doc linked below
- When scrolling traits, show a scrim on the top and bottom of the container. Don't show top/bottom scrim if at the top/bottom.
- Added purpleVibrant, color from figma missing in theme

_JIRA ticket:_ https://uniswaplabs.atlassian.net/browse/NFT-1171
_Relevant docs:_https://www.notion.so/Details-Redesign-0b14fc9f47d946c8af5523b27812205e#11a16ae42cdb4a649a203266696bc6e7

## Screen capture
Trait container while scrolling
Screen Shot 2023-04-27 at 11 17 31
No scroll available
Screen Shot 2023-04-27 at 11 17 45
No traits
Screen Shot 2023-04-27 at 11 18 32

### Automated testing

- Added a snapshot test for an asset with no traits
  - Will add one for an asset with trait when rarity isn't randomly generated as snapshots will never match currently


src/nft/components/details/detailsV2/DataPageTraits.tsx
```diff
@@ -1,16 +1,104 @@
...
+
+const TraitRowContainer = styled.div`
+  position: relative;
+`
+
+const TraitRowScrollableContainer = styled.div`
+  overflow-y: auto;
+  overflow-x: hidden;
+  max-height: 412px;
+  width: calc(100% + 6px);
```
cbachmeier, line 34: this added width is to allow the scrollbar to exist in the right padding

src/nft/components/details/detailsV2/DataPageTraits.tsx
```diff
@@ -1,16 +1,104 @@
...
+
+  const scrollHandler = (event: React.UIEvent<HTMLDivElement>) => {
+    const scrollTop = event.currentTarget.scrollTop
+    const containerHeight = event.currentTarget.clientHeight
+    const scrollHeight = event.currentTarget.scrollHeight
+
+    setScrollProgress(scrollTop ? ((scrollTop + containerHeight) / scrollHeight) * 100 : 0)
+  }
+
+  // This is needed to prevent rerenders when handling scrolls
```
cbachmeier, line 72: Here's a fun gif without the memo where the random number generation occurs with each rerender caused by scrolling
randomGraphScroll

src/nft/components/details/detailsV2/DataPageTraits.tsx
```diff
@@ -1,16 +1,104 @@
...
+
+const Scrim = styled.div<{ isBottom?: boolean }>`
+  position: absolute;
+  height: 88px !important;
+  left: 0px;
+  right: 12px;
+  ${({ isBottom }) => !isBottom && 'top: 0px'};
+  ${({ isBottom }) => !isBottom && 'transform: matrix(1, 0, 0, -1, 0, 0)'};
+
+  ${({ isBottom }) => isBottom && 'bottom: 0px'};
```
JackShort, line 47: I think these can be transformed into this

```
  ${({ isBottom }) =>
    isBottom
      ? `
        bottom: 0px
    `
      : `
      top: 0px;
      transform: matrix(1, 0, 0, -1, 0, 0)
    `};
```

src/nft/components/details/detailsV2/DataPageTraits.tsx
```diff
@@ -1,16 +1,104 @@
...
+    }
+  }
+
+  const scrollHandler = (event: React.UIEvent<HTMLDivElement>) => {
+    const scrollTop = event.currentTarget.scrollTop
+    const containerHeight = event.currentTarget.clientHeight
+    const scrollHeight = event.currentTarget.scrollHeight
+
+    setScrollProgress(scrollTop ? ((scrollTop + containerHeight) / scrollHeight) * 100 : 0)
+  }
```
JackShort, line 70: I know this same code is used in bag - I think it might be good to make this into its own hook then later we can also refactor the bag to use the same hook

src/nft/components/details/detailsV2/DataPageTraits.tsx
JackShort, line 72: I am going to assume this comment is going to be deleted before merge?

src/nft/components/details/detailsV2/DataPageTraits.tsx
```diff
@@ -1,16 +1,104 @@
...
+
+  // This is needed to prevent rerenders when handling scrolls
+  const traitRows = useMemo(() => {
+    return traits?.map((trait) => <TraitRow trait={trait} key={trait.trait_type + ':' + trait.trait_value} />)
```
JackShort, line 74: nit: the `trait_type` should be unique right so the key can just be `trait_type`?

src/nft/components/details/detailsV2/DataPageTraits.tsx
```diff
@@ -1,16 +1,104 @@
...
+  }, [traits])
+
+  return (
+    <Column>
+      <TraitsHeaderContainer>
+        <TraitsHeader $flex={3}>
+          <Trans>Trait</Trans>
+        </TraitsHeader>{' '}
```
JackShort, line 82: What do `{' '}` do here?

src/nft/components/details/detailsV2/RarityGraph.tsx
```diff
@@ -0,0 +1,93 @@
...
+    RarityLevel.ExtremelyRare,
+    {
+      threshold: 0,
+      color: colors.magentaVibrant,
+    },
+  ],
+])
+
+function getRarityLevel(rarity: number) {
+  switch (true) {
```
JackShort, line 65: fun fact: since there are only 5 steps you can actually do something like this which has better performance - at a downside of initial understanding so I don't really recommend doing it

```
function getRarityLevel(rarity: number) {
  switch (Math.floor(rarity * 5)) {
    case (RarityLevels.get(RarityLevel.VeryCommon)?.threshold ?? Number.MAX_SAFE_INTEGER) * 5:
      return RarityLevels.get(RarityLevel.VeryCommon)
    case (RarityLevels.get(RarityLevel.Common)?.threshold ?? Number.MAX_SAFE_INTEGER) * 5:
      return RarityLevels.get(RarityLevel.Common)
    case (RarityLevels.get(RarityLevel.Rare)?.threshold ?? Number.MAX_SAFE_INTEGER) * 5:
      return RarityLevels.get(RarityLevel.Rare)
    case (RarityLevels.get(RarityLevel.VeryRare)?.threshold ?? Number.MAX_SAFE_INTEGER) * 5:
      return RarityLevels.get(RarityLevel.VeryRare)
    case (RarityLevels.get(RarityLevel.ExtremelyRare)?.threshold ?? Number.MAX_SAFE_INTEGER) * 5:
      return RarityLevels.get(RarityLevel.ExtremelyRare)
    default:
      return RarityLevels.get(RarityLevel.VeryCommon)
  }
}
```

src/nft/components/details/detailsV2/RarityGraph.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+enum RarityLevel {
+  VeryCommon = 'Very Common',
+  Common = 'Common',
+  Rare = 'Rare',
+  VeryRare = 'Very Rare',
+  ExtremelyRare = 'Extremely Rare',
+}
+
+const RarityLevels: Map<string, RarityValue> = new Map([
```
JackShort, line 26: this might be better as an object to avoid the conditionals

src/nft/components/details/detailsV2/TraitRow.tsx
```diff
@@ -0,0 +1,47 @@
...
+  font-weight: 600;
+  color: ${({ theme }) => theme.textSecondary};
+`
+
+const TraitValue = styled(Column)`
+  gap: 4px;
+  flex: 3;
+`
+
+const TraitRowValue = styled(ThemedText.BodySmall)<{ $flex?: number; alignRight?: boolean }>`
```
JackShort, line 22: same comment here about the `alignRight`

src/nft/components/details/detailsV2/TraitRow.tsx
```diff
@@ -0,0 +1,47 @@
...
+`
+
+export const TraitRow = ({ trait }: { trait: Trait }) => {
+  // TODO: Replace with actual rarity, count, and floor price when BE supports
+  // rarity eventually should be number of items with this trait / total number of items, smaller rarity means more rare
+  const randomRarity = Math.random()
+  return (
+    <Row padding="12px 8px 12px 0px">
+      <TraitValue>
+        <SubheaderTiny>{trait.trait_type}</SubheaderTiny>{' '}
```
JackShort, line 37: same question here about the `{' '}`

src/nft/components/details/detailsV2/DataPage.tsx
```diff
@@ -47,7 +47,7 @@ export const DataPage = ({ asset }: { asset: GenieAsset }) => {
       <DataPageHeader />
       <ContentContainer>
         <LeftColumn>
-          <DataPageTraits asset={asset} />
+          {!!asset.traits?.length && <DataPageTraits asset={asset} />}
```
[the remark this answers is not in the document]
cbachmeier: traits is possibly undefined so `!!` covers both that it is defined and has a value greater than 0

src/nft/components/details/detailsV2/DataPageTraits.tsx
```diff
@@ -1,16 +1,104 @@
...
+  position: relative;
+`
+
+const TraitRowScrollableContainer = styled.div`
+  overflow-y: auto;
+  overflow-x: hidden;
+  max-height: 412px;
+  width: calc(100% + 6px);
+
+  ${ScrollBarStyles}
```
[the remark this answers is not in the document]
cbachmeier: design is currently reviewing the deploy

src/nft/components/details/detailsV2/DataPageTraits.tsx
```diff
@@ -1,16 +1,104 @@
...
+`
+
+const Scrim = styled.div<{ isBottom?: boolean }>`
+  position: absolute;
+  height: 88px !important;
```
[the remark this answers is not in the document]
cbachmeier: no good catch, scrim was originally having a height of 0px but it was caused by not having display set to flex

src/nft/components/details/detailsV2/DataPageTraits.tsx
cbachmeier: Our usual practice is to include maps in the render function so I think it's worth keeping since this breaks the convention. We could also remove this once we stop using random values for rarity

src/nft/components/details/detailsV2/DataPageTraits.tsx
cbachmeier: I don't believe there is any built in protection that guarantees `trait_type` is always unique for an item

src/nft/components/details/detailsV2/RarityGraph.tsx
cbachmeier: Does the performance increase come from the comparison of > vs being equally to an int? Interesting to know but I think I lean towards the better readability of the first implementation personally.

src/nft/components/details/detailsV2/RarityGraph.tsx
cbachmeier: The reason I didn't use an object is that with an object I can't use the `RarityLevel` enum as the key for the respective `RarityValues`. Having that is worth the tradeoff of needing the superfluous conditional when getting the threshold

src/nft/components/details/detailsV2/DataPageTraits.tsx
cbachmeier: great suggestion, I've added the `useSubscribeScrollState` hook

src/nft/components/details/detailsV2/RarityGraph.tsx
JackShort: Yeah it comes from removing the extra branch that is created in the comparison

src/nft/components/details/detailsV2/RarityGraph.tsx
JackShort: can you not not do it like this?

```
const RarityLevels: { [key in RarityLevel]: RarityValue } = {
  [RarityLevel.VeryCommon]: {
    threshold: 0.8,
    color: colors.gray500,
  },
  [RarityLevel.Common]: {
    threshold: 0.6,
    color: colors.green300,
  },
  [RarityLevel.Rare]: {
    threshold: 0.4,
    color: colors.blueVibrant,
  },
  [RarityLevel.VeryRare]: {
    threshold: 0.2,
    color: colors.purpleVibrant,
  },
  [RarityLevel.ExtremelyRare]: {
    threshold: 0,
    color: colors.magentaVibrant,
  },
}

function getRarityLevel(rarity: number) {
  switch (true) {
    case rarity > RarityLevels[RarityLevel.VeryCommon].threshold:
      return RarityLevels[RarityLevel.VeryCommon]
    case rarity > RarityLevels[RarityLevel.Common].threshold:
      return RarityLevels[RarityLevel.Common]
    case rarity > RarityLevels[RarityLevel.Rare].threshold:
      return RarityLevels[RarityLevel.Rare]
    case rarity > RarityLevels[RarityLevel.VeryRare].threshold:
      return RarityLevels[RarityLevel.VeryRare]
    case rarity >= RarityLevels[RarityLevel.ExtremelyRare].threshold:
      return RarityLevels[RarityLevel.ExtremelyRare]
    default:
      return RarityLevels[RarityLevel.VeryCommon]
  }
}
```

JackShort (approved):

src/nft/components/details/detailsV2/DataPage.tsx
```diff
@@ -47,7 +47,7 @@ export const DataPage = ({ asset }: { asset: GenieAsset }) => {
       <DataPageHeader />
       <ContentContainer>
         <LeftColumn>
-          <DataPageTraits asset={asset} />
+          {!!asset.traits?.length && <DataPageTraits traits={asset.traits} />}
```
grabbou, line 50: nit: I think `asset.traits?.length > 0` would look a bit cleaner

src/nft/components/details/detailsV2/DataPageTraits.test.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+import { DataPageTraits } from './DataPageTraits'
+
+it('data page trait component does not load with asset with no traits', () => {
+  const { asFragment } = render(<DataPageTraits asset={TEST_NFT_ASSET} />)
+  expect(asFragment()).toMatchSnapshot()
+})
+
+// TODO: add test for trait component with asset with traits when rarity is not randomly generated
+// while rarities are randomly generated, snapshots will never match
```
grabbou, line 12: nit: what I usually do is I create a JIRA ticket for each todo and write `TODO(WEB-XXX)` instead so that it's easier to track. I would encourage you to do the same, unless already handled, so that this doesn't end up living there forever!

src/nft/components/details/detailsV2/DataPageTraits.tsx
```diff
@@ -1,23 +1,101 @@
...
+  height: 88px;
+  left: 0px;
+  right: 6px;
+
+  ${({ isBottom }) =>
+    isBottom
+      ? 'bottom: 0px'
+      : `
+        top: 0px;
+        transform: matrix(1, 0, 0, -1, 0, 0);
```
grabbou, line 51: nit: this is really interesting and I have to admit, a bit of a blackbox to me. Do you think it would make sense to annotate this component with a comment, what it is and how it works?

src/nft/components/details/detailsV2/RarityGraph.tsx
cbachmeier: I stand corrected, great solution

src/nft/components/details/detailsV2/DataPageTraits.tsx
grabbou: Thanks for the gif and the catch, that definitely makes sense with the scroll handler in mind.

grabbou: Which comment are you referring to? The one that describes the use of `useMemo` is actually quite handy here IMO, I would consider extending to mention it's due to scrollHandler causing re-renders.

src/nft/components/details/detailsV2/DataPageTraits.tsx
```diff
@@ -1,23 +1,101 @@
...
+
+  // This is needed to prevent rerenders when handling scrolls
+  const traitRows = useMemo(() => {
+    return traits?.map((trait) => <TraitRow trait={trait} key={trait.trait_type + ':' + trait.trait_value} />)
+  }, [traits])
+
+  return (
+    <Column>
+      <TraitsHeaderContainer>
+        <TraitsHeader $flex={3}>
```
grabbou, line 70: nit: when did we start using `$` for variables? (I know the way $ works with Styled components). What makes `flex` different from e.g. `justify` (used elsewhere for TraitsHeader)

src/nft/components/details/detailsV2/DataPageTraits.tsx
grabbou: Are you sure you didn't lave accidental spaces there that were annotated as such? I am pretty sure if you remove this, it won't re-appear, unless there's actually space there and is needed for the UI.

src/nft/components/details/detailsV2/RarityGraph.tsx
```diff
@@ -0,0 +1,93 @@
...
+      return RarityLevels.get(RarityLevel.ExtremelyRare)
+    default:
+      return RarityLevels.get(RarityLevel.VeryCommon)
+  }
+}
+
+export const RarityGraph = ({ trait, rarity }: { trait: Trait; rarity: number }) => {
+  const rarityLevel = getRarityLevel(rarity)
+  return (
+    <Row gap="1.68px" justify="flex-end">
```
grabbou, line 69: nit: This is a very odd value and I definitely doesn't fall under any of the XS, sm, md etc. that we have for `gap`. Is that value coming from Figma? Does it by any chance have an appropriate name, such as `xxxs`?

src/nft/components/details/detailsV2/TraitRow.tsx
```diff
@@ -0,0 +1,47 @@
...
+  gap: 4px;
+  flex: 3;
+`
+
+const TraitRowValue = styled(ThemedText.BodySmall)<{ $flex?: number; justify?: string }>`
+  display: flex;
+  line-height: 20px;
+  padding-top: 20px;
+  flex: ${({ $flex }) => $flex ?? 1};
+  justify-content: ${({ justify }) => justify};
```
grabbou, line 27: nit: I wonder whether `$justifyContent` prop wouldn't be more appropriate, just to avoid confusion (I see we're differentiating between configurable CSS properties and the rest).

src/nft/components/details/detailsV2/TraitRow.tsx
```diff
@@ -0,0 +1,47 @@
...
+import Row from 'components/Row'
+import { Trait } from 'nft/types'
+import { formatEth } from 'nft/utils'
+import styled from 'styled-components/macro'
+import { ThemedText } from 'theme'
+
+import { RarityGraph } from './RarityGraph'
+
+const SubheaderTiny = styled.div`
+  font-size: 10px;
```
grabbou, line 11: nit: Do we have any ThemedText that works here? If not, does the font size used in Figma have an actual name? Can we create a new ThemedText for it then? If yes, can we already pin down other occurrences of such configuration where we could use ThemedText too?

src/nft/hooks/useSubscribeScrollState.ts
```diff
@@ -0,0 +1,21 @@
...
+  const [userCanScroll, setUserCanScroll] = useState(false)
+  const [scrollProgress, setScrollProgress] = useState(0)
+  const scrollRef = (node: HTMLDivElement) => {
+    if (node !== null) {
+      const canScroll = node.scrollHeight > node.clientHeight
+      canScroll !== userCanScroll && setUserCanScroll(canScroll)
+    }
+  }
+
+  const scrollHandler = (event: React.UIEvent<HTMLDivElement>) => {
```
grabbou, line 13: Did you copy this from `nft/components/Bag`? If yes, why not to reuse that hook there too?

grabbou (changes requested):
Thanks for the PR! Left some questions and a few nits!

src/nft/components/details/detailsV2/DataPage.tsx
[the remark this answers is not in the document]
cbachmeier: @grabbou see this comment for using !! over > 0

src/nft/components/details/detailsV2/DataPageTraits.tsx
cbachmeier: We use `$` when the param name matches an exact css style such as `$flex` or `$color`

src/nft/components/details/detailsV2/TraitRow.tsx
cbachmeier: This doesn't fit any current ThemedText styles. It's also the first occurrence I have personally seen of this typography and it is also not specced out in the global design system. If Design plans on making this a standard than we can add it but given this is currently a one-off I think current implementation is sufficient

src/nft/components/details/detailsV2/DataPageTraits.test.tsx
cbachmeier: As a heads up in case you missed the slack thread, this whole page redesign is in progress with unfinished UX being expected. There is a outlined implementation plan in the description.

src/nft/components/details/detailsV2/DataPageTraits.tsx
cbachmeier: I've added a comment. You can also reference the Scrim object in the figma if you'd like to go more in depth for the styles

src/nft/components/details/detailsV2/DataPage.tsx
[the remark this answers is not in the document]
grabbou: That makes sense to me, however, still, it's a question what we consider more readable. Personally, I would still do `> 0` to embrace optional chaining (at the disadvantage of having `length` being undefined, converted to `NaN` and compared against 0, which always returns false). Your approach is also valid, in a way that it will convert both `undefined` and `0` (when `traits` is defined). I let you pick the right choice.

src/nft/components/details/detailsV2/DataPageTraits.test.tsx
grabbou: Sounds good. That means there's JIRA ticket in progress. I would myself use the umbrella ticket as a reference to quickly find all the remaining work under that ticket, but I guess you're deeper into the NFT part of the codebase, so do what works better for you!

src/nft/components/details/detailsV2/DataPageTraits.tsx
grabbou: Sounds good. If it's in Figma, that is good. +1 for keeping the name.

src/nft/components/details/detailsV2/TraitRow.tsx
grabbou: Yes, that makes sense - I agree. I have raised this with the design team tho.

src/nft/hooks/useSubscribeScrollState.ts
grabbou: Sounds good to me! If you could either have a JIRA ticket for it or just write it down somewhere on a list of follow-up tickets so it doesn't get lost - it would be appreciated. Feel free to CC me on that PR when it's up.

cbachmeier: The ticket is tagged at the top of the hook `TODO(NFT-1190)`

grabbou (approved):

PR #6462, opened 2023-04-27
test: upgrade cypress-hardhat

zzmp:
## Description

Upgrades cypress-hardhat to a version which includes cy.hardhat() and cy.provider().

### Automated testing

- [x] Integration/E2E test


JFrankfurt (approved):

cypress/e2e/swap.test.ts
```diff
@@ -81,35 +77,33 @@ describe('Swap', () => {
...
-            'have.text',
-            `Balance: ${initialBalance + BALANCE_INCREMENT}`
-          )
-
-          // chain state check
-          cy.then(() => hardhat.utils.getBalance(hardhat.wallet.address, USDC_MAINNET))
-            .then((balance) => Number(balance.toFixed(1)))
-            .should('eq', initialBalance + BALANCE_INCREMENT)
+    cy.hardhat().then((hardhat) => {
+      cy.then(() => hardhat.getBalance(hardhat.wallet.address, USDC_MAINNET))
```
cmcewen, line 81: can this take an async function? i find a deeply nested `then` tree difficult

cypress/e2e/swap.test.ts
```diff
@@ -81,35 +77,33 @@ describe('Swap', () => {
...
+          cy.get('#swap-currency-output .open-currency-select-button').click()
+          cy.get(getTestSelector('token-search-input')).clear().type(TOKEN_ADDRESS)
+          cy.contains('USDC').click()
+          cy.get('#swap-currency-output .token-amount-input').clear().type(BALANCE_INCREMENT.toString())
+          cy.get('#swap-currency-input .token-amount-input').should('not.equal', '')
+          cy.get('#swap-button').click()
+          cy.get('#confirm-swap-or-send').click()
+          cy.get(getTestSelector('dismiss-tx-confirmation')).click()
+
+          cy.then(() => hardhat.provider.send('hardhat_mine', ['0x1', '0xc'])).then(() => {
```
cmcewen, line 93: why did you set the interval value here?

package.json
```diff
@@ -103,7 +103,7 @@
     "@vanilla-extract/webpack-plugin": "^2.1.11",
     "babel-plugin-istanbul": "^6.1.1",
     "cypress": "10.3.1",
-    "cypress-hardhat": "^1.0.1",
+    "cypress-hardhat": "^2.0.0",
```
cmcewen, line 106: did we 1.0 too early here?

cmcewen (approved):

cypress/e2e/swap.test.ts
zzmp: No, that's not how cypress operates.

Cypress does not use promises, but instead uses Chainable, which forces you to use closures to maintain state across actions.

cypress/e2e/swap.test.ts
zzmp: This is legacy - it's on our TODO list to move this to a hardhat function. The interval is required to explicitly run over deadline, I think.

package.json
zzmp: We just had a breaking change, so we bumped the major. I'd consider this working-as-intended, I don't think we should assign too much value to "hitting 1.0".

cypress/e2e/swap.test.ts
zzmp: I'll make sure to include this in a readme in the cypress folder for future test-writers :)

PR #6465, opened 2023-04-28
feat: adding feat flag to update pwat with new contract

JackShort:
The UR contract on the backend is being updated to a new version on Tuesday and this pr is syncing up a statsig gate so we can easily switch to the contract whenever backend does

### Automated testing

- [x] Unit test


src/nft/hooks/usePermit2Approval.ts
```diff
@@ -17,14 +18,22 @@ export default function usePermit2Approval(
...
+  const canPermit = enabled && chainId
+  let contractToApprove: string | undefined
+
+  if (canPermit) {
+    const forceNftRouter = shouldUseNftRouter && chainId === 1
+    const mainnetContractToApprove = shouldUseUpdatedContract
+      ? NFT_UNIVERSAL_ROUTER_MAINNET_ADDRESS
+      : DEPRECATED_NFT_UNIVERSAL_ROUTER_MAINNET_ADDRESS
+    contractToApprove = forceNftRouter ? mainnetContractToApprove : UNIVERSAL_ROUTER_ADDRESS(chainId)
+  }
```
zzmp, line 32: This is pretty confusing with all the ternaries.
Can you move this to an unexported hook that just returns the contract, so you can use if/else statements instead of ternaries?

zzmp (changes requested):

src/nft/hooks/usePermit2Approval.ts
```diff
@@ -17,14 +40,11 @@ export default function usePermit2Approval(
   shouldUseNftRouter?: boolean
 ) {
   const { chainId } = useWeb3React()
+  const contractToApprove = useUniversalRouterAddress(enabled, shouldUseNftRouter, chainId)
```
zzmp, line 43: nit: call this `universalRouterAddress`

src/nft/hooks/usePermit2Approval.ts
```diff
@@ -3,12 +3,35 @@ import { InterfaceEventName } from '@uniswap/analytics-events'
...
+
+  if (forceNftRouter) {
+    contractToApprove = shouldUseUpdatedContract
+      ? NFT_UNIVERSAL_ROUTER_MAINNET_ADDRESS
+      : DEPRECATED_NFT_UNIVERSAL_ROUTER_MAINNET_ADDRESS
+  } else if (canPermit) {
+    contractToApprove = UNIVERSAL_ROUTER_ADDRESS(chainId)
+  }
+
+  return useMemo(() => contractToApprove, [contractToApprove])
```
zzmp, line 33: This is a string, so it does not need to be memo-ized. You can simplify by just returning earli in L25-31.

src/nft/hooks/usePermit2Approval.ts
```diff
@@ -3,12 +3,35 @@ import { InterfaceEventName } from '@uniswap/analytics-events'
...
+  chainId?: number
+): string | undefined => {
+  let contractToApprove: string | undefined
+  const shouldUseUpdatedContract = usePwatNewContractEnabled()
+
+  const canPermit = enabled && chainId
+  const forceNftRouter = canPermit && shouldUseNftRouter && chainId === 1
+
+  if (forceNftRouter) {
+    contractToApprove = shouldUseUpdatedContract
```
zzmp, line 26: Just use if statements, and return early - avoid ternary here.

src/nft/hooks/usePermit2Approval.ts
```diff
@@ -3,12 +3,35 @@ import { InterfaceEventName } from '@uniswap/analytics-events'
...
-const NFT_UNIVERSAL_ROUTER_MAINNET_ADDRESS = '0x4c60051384bd2d3c01bfc845cf5f4b44bcbe9de5'
+const DEPRECATED_NFT_UNIVERSAL_ROUTER_MAINNET_ADDRESS = '0x4c60051384bd2d3c01bfc845cf5f4b44bcbe9de5'
+const NFT_UNIVERSAL_ROUTER_MAINNET_ADDRESS = '0x3fC91A3afd70395Cd496C647d5a6CC9D4B2b7FAD'
+
+const useUniversalRouterAddress = (
+  enabled?: boolean,
+  shouldUseNftRouter?: boolean,
+  chainId?: number
+): string | undefined => {
+  let contractToApprove: string | undefined
```
zzmp, line 19: not needed - just return early.

zzmp (changes requested):

src/nft/hooks/usePermit2Approval.ts
```diff
@@ -3,12 +3,31 @@ import { InterfaceEventName } from '@uniswap/analytics-events'
...
 
-// TODO: This should be removed when the sdk is updated to include the new UR address
-const NFT_UNIVERSAL_ROUTER_MAINNET_ADDRESS = '0x4c60051384bd2d3c01bfc845cf5f4b44bcbe9de5'
+const DEPRECATED_NFT_UNIVERSAL_ROUTER_MAINNET_ADDRESS = '0x4c60051384bd2d3c01bfc845cf5f4b44bcbe9de5'
+const NFT_UNIVERSAL_ROUTER_MAINNET_ADDRESS = '0x3fC91A3afd70395Cd496C647d5a6CC9D4B2b7FAD'
+
+const useUniversalRouterAddress = (
+  enabled?: boolean,
+  shouldUseNftRouter?: boolean,
+  chainId?: number
```
zzmp, line 17: These shouldn't be optional - they're always used.

zzmp (approved):

src/featureFlags/flags/featureFlags.ts
```diff
@@ -9,4 +9,5 @@ export enum FeatureFlag {
   swapWidget = 'swap_widget_replacement_enabled',
   statsigDummy = 'web_dummy_gate_amplitude_id',
   detailsV2 = 'details_v2',
+  newContractForPWAT = 'new_contract_for_pwat',
```
cbachmeier, line 12: nit: get the latest UR contract version number and use that in the flag name such as `pwat_contract_vXX.XX`. This avoids when the next version comes out, us having to use the flag `new_new_contract_for_pwat`

src/nft/hooks/usePermit2Approval.ts
```diff
@@ -3,12 +3,31 @@ import { InterfaceEventName } from '@uniswap/analytics-events'
 import { CurrencyAmount, Token } from '@uniswap/sdk-core'
 import { UNIVERSAL_ROUTER_ADDRESS } from '@uniswap/universal-router-sdk'
 import { useWeb3React } from '@web3-react/core'
+import { usePwatNewContractEnabled } from 'featureFlags/flags/pwatNewContract'
 import usePermit2Allowance, { AllowanceState } from 'hooks/usePermit2Allowance'
 import { useCallback, useMemo, useState } from 'react'
 import invariant from 'tiny-invariant'
 
-// TODO: This should be removed when the sdk is updated to include the new UR address
```
cbachmeier, line 10: q: is the sdk upgrade still blocked by BE updates?

src/nft/hooks/usePermit2Approval.ts
```diff
@@ -3,12 +3,31 @@ import { InterfaceEventName } from '@uniswap/analytics-events'
...
-const NFT_UNIVERSAL_ROUTER_MAINNET_ADDRESS = '0x4c60051384bd2d3c01bfc845cf5f4b44bcbe9de5'
+const DEPRECATED_NFT_UNIVERSAL_ROUTER_MAINNET_ADDRESS = '0x4c60051384bd2d3c01bfc845cf5f4b44bcbe9de5'
+const NFT_UNIVERSAL_ROUTER_MAINNET_ADDRESS = '0x3fC91A3afd70395Cd496C647d5a6CC9D4B2b7FAD'
```
cbachmeier, line 12: nit: I would use version numbers for the naming instead of deprecated vs not

cbachmeier (approved):
LGTM

src/featureFlags/flags/featureFlags.ts
JackShort: I actually think this is a better way to do it because with two prs we could update the contract without changing the feature flag. By changing both deprecated and nft router address to the current address we have now, then turn off feature flag, then update the new nft router to the new address. I think this beats making a new feature flag for every new contract upgrade - when we can just reuse this one

src/nft/hooks/usePermit2Approval.ts
JackShort: yeah but now we have a like a feature flag setup and stuff so I think its a little more clear whats going on vs a hardcoded string

src/featureFlags/flags/featureFlags.ts
JackShort: resolved on slack - but will keep this naming convention since it has to match the current naming convention on statsig -however for future prs will change

PR #6469, opened 2023-04-28
test: keep test output clean

zzmp:
## Description

Fails tests if they log to the console. Tests should instead spyOn the console object's methods and assert that they are called, if need be.

_JIRA ticket:_ https://uniswaplabs.atlassian.net/browse/WEB-3132

### Automated testing

- [x] Unit test

### DIFF
--- a/src/setupTests.ts
+++ b/src/setupTests.ts
@@ -3,6 +3,7 @@ import 'jest-styled-components' // adds style diffs to snapshot tests
 
 import type { createPopper } from '@popperjs/core'
 import { useWeb3React } from '@web3-react/core'
+import failOnConsole from 'jest-fail-on-console'
 import ResizeObserver from 'resize-observer-polyfill'
 import { Readable } from 'stream'
 import { mocked } from 'test-utils/mocked'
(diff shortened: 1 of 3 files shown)


cartcrom (approved):
awesome

src/setupTests.ts
```diff
@@ -77,3 +78,22 @@ beforeEach(() => {
   // Mock useWeb3React to return a chainId of 1 by default.
   mocked(useWeb3React).mockReturnValue({ chainId: 1 } as ReturnType<typeof useWeb3React>)
 })
+
+/**
+ * Fail tests if anything is logged to the console. This keeps the console clean and ensures test output stays readable.
+ * If something should log to the console, it should be stubbed and asserted:
+ * @example
+ * beforeEach(() => jest.spyOn(console, 'error').mockReturnsValue())
```
grabbou, line 86: nit: Historically, I often surrounded my code blocks with triple `, like on Github. You could consider doing that too.

grabbou (approved):
Awesome. Great way to be more in charge of what and where we `console.log` to avoid accidents.

src/setupTests.ts
zzmp: this is not required for jsdoc. See https://jsdoc.app/tags-example.html.
jsdoc examples should automatically be displayed as highlighted code.

grabbou: You are right. However, I think in context of a TypeScript file, we should refer to TSDoc instead, which is the standard used by different tools within TypeScript ecosystem (including Visual Studio Code) to work with comments. 

While JSDoc comments may work, it is not guaranteed:
> As much as possible, TSDoc should preserve the familiar style of JSDoc/Markdown. This also maximizes the likelihood that legacy comments will parse correctly as TSDoc

That said, TSDoc [uses ```  to annotate code snippets](https://tsdoc.org/pages/tags/example/). This is also a format [used by Docosaurus](https://docusaurus.io/docs/markdown-features/code-blocks), which increases interoperability, in case we ever have code snippets in other parts of our codebase that serve as documentation.

PR #6475, opened 2023-05-01
fix: filter chrome-extension errors

vm:
## Description
https://uniswap-labs.sentry.io/issues/4122994488/?project=4504255148851200&referrer=release-issue-stream
https://uniswap-labs.sentry.io/issues/4144076511/?end=2023-05-01T15%3A50%3A59&project=4504255148851200&query=release%3Ad0a10fcf8dce6d8f9b1c06c0f640921b7d5ab33b+chrome-extension&referrer=issue-stream&sort=freq&start=2023-05-01T12%3A09%3A00&stream_index=1

Errors like these can be filtered out because we can't control what scripts try to inject onto the page. If they break the rules of our CSP or try to read values that don't exist, these errors should not be sent to Sentry.

### Automated testing
- [x] Unit test


zzmp (changes requested):
I don't know that we want to ignore this entire class of errors. We interact with wallets, a large portion of which are injected through chrome extensions. If we filter out all of these errors, we may be filtering out wallet integration errors too.

zzmp (approved):

cartcrom (approved):

PR #6476, opened 2023-05-01
test(lint): separate multiline comments

zzmp:
## Description

Enforces separate-line [multiline-comment-style](https://eslint.org/docs/latest/rules/multiline-comment-style) for this codebase. This disambiguates implementation notes (`//`) from TSDoc (`/**`), _which are still allowed as multiline comments_.
This should eventually be moved upstream to @uniswap/eslint-config.

_Relevant docs:_ https://eslint.org/docs/latest/rules/multiline-comment-style

### Automated testing

- [x] eslint


cbachmeier (changes requested):
What's the rationale for switching to star block comments over separate lines? Based on the repo, we have as a team seemed to heavily favor separate lines. Currently we have 100+ instances of using separate line multiblock comments and only 17 instances of star block.

I also think separate lines is a better devEx as it's the standard for Vs Code's multiline comment shortcut `CMD + /` and it reduces empty lines. For example the PR above just to convert adds 300+lns.

grabbou:
@cbachmeier 

> I also think separate lines is a better devEx as it's the standard for Vs Code's multiline comment shortcut CMD + / and it reduces empty lines. For example the PR above just to convert adds 300+lns.

That depends on your workflow. With multiline comments, there's no need for `CMD + /` as lines are appended automatically on each new line. It is also required syntax by JSDoc/TSDoc in order to provide tooltips when working with different parts of the codebase.

grabbou:
> What's the rationale for switching to star block comments over separate lines?

The rationale is pretty much to settle down on a standard that will be easy for everyone to remember and to use. Right now, we don't really have a clear pattern for writing comments, and we tend to use both `/*` for inline and multiline comments as well as `//`. 

Now, what is the standard, is another question. The outcome of this PR should be reducing the ambiguity.

zzmp:
> It is also required syntax by JSDoc/TSDoc in order to provide tooltips when working with different parts of the codebase.

This is a non-issue, as JSDoc/TSDoc comments are exempted from this rule by design.

src/constants/tokenSafety.tsx
```diff
@@ -56,8 +56,7 @@ export function getWarningCopy(warning: Warning | null, plural = false) {
 export type Warning = {
   level: WARNING_LEVEL
   message: JSX.Element
-  /* canProceed determines whether triangle/slash alert icon is used, and
-    whether this token is supported/able to be traded */
+  /** Determines whether triangle/slash alert icon is used, and whether this token is supported/able to be traded. */
```
cbachmeier, line 59: Nit: this one could probably be swapped to //

src/hooks/useTokenInfoFromActiveList.ts
```diff
@@ -3,9 +3,7 @@ import { useWeb3React } from '@web3-react/core'
 import { useMemo } from 'react'
 import { useCombinedActiveList } from 'state/lists/hooks'
 
-/**
- * Returns a WrappedTokenInfo from the active token lists when possible,
- * or the passed token otherwise. */
+/** Returns a WrappedTokenInfo from the active token lists when possible, or the passed token otherwise. */
```
cbachmeier, line 6: Nit: this one could also be swapped with //

cbachmeier (approved):
LGTM

src/constants/tokenSafety.tsx
zzmp: This is a doc-comment - using `/**` ensures an IDE will surface this as documentation for the property.

src/hooks/useTokenInfoFromActiveList.ts
zzmp: This is a doc-comment - using `/**` ensures an IDE will surface this as documentation for the function.

cmcewen (approved):

PR #6480, opened 2023-05-01
feat: new Settings menu for Swap

grabbou:
This PR implements new Settings menu design as per Figma.

| Before  | After
|---|---|
| Screenshot 2023-05-03 at 17 18 24  |  Screenshot 2023-05-03 at 17 17 05

Additional changes made in the PR:
- Added new primitive components (Radio, Expand) to achieve desired UI effect
- Replaced `useClientSideRouter` with `useRouterPreference` and added a new `RouterPreference.AUTO` that defaults to `API`
- Added support for grid primitives (e.g. `sm` or `md`) in a Row component
- Replaced `auto` with `SlippageTolerance.AUTO`
- Renamed `placeholderSlippage` to `autoSlippage` to pass down default slippage from trade, instead of user value with default (required for the UI to work)

PR includes a few TODOs as comments, will convert to JIRA tickets and update comments once PR and TODOs get approved.

I will followup with animations in next PR.


src/components/Row/index.tsx
```diff
@@ -18,7 +26,7 @@ const Row = styled(Box)<{
   padding: ${({ padding }) => padding};
   border: ${({ border }) => border};
   border-radius: ${({ borderRadius }) => borderRadius};
-  gap: ${({ gap }) => gap};
+  gap: ${({ gap, theme }) => gap && (theme.grids[gap as Gap] || gap)};
```
tinaszheng, line 29: since none of the other props here are theme tokens thoughts on just letting the consumer figure out the right value?

grabbou: For the `gap`? The primary motivation was to align with `Column`, which already has a `gap` property that accepts theme tokens. Not sure why this wasn't available for `Row` yet.

tinaszheng: omg 😂

src/state/user/reducer.ts
```diff
@@ -20,7 +21,8 @@ export interface UserState {
 
   userExpertMode: boolean
 
-  userClientSideRouter: boolean // whether routes should be calculated with the client side router only
+  // which route should be used to calculate trades
```
tinaszheng, line 24:
```suggestion
  // which router should be used to calculate trades
```

:3

tinaszheng:
so beautiful

src/components/Settings/Input/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  background: transparent;
+  text-align: right;
+  &::-webkit-outer-spin-button,
+  &::-webkit-inner-spin-button {
+    -webkit-appearance: none;
+  }
+  ::placeholder {
+    color: ${({ theme }) => theme.textTertiary};
+  }
+  :invalid {
```
tinaszheng, line 21: v nice usage of defaults here

src/components/Settings/MaxSlippageSettings/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+const Option = styled(Row)<{ isActive: boolean }>`
+  width: auto;
+  cursor: pointer;
+  padding: 6px 12px;
+  text-align: center;
+  gap: 4px;
+  border-radius: 12px;
+  background: ${({ isActive, theme }) => (isActive ? theme.backgroundInteractive : 'transparent')};
+  pointer-events: ${({ isActive }) => isActive && 'none'};
```
tinaszheng, line 25: hmm feels more correct to do this based on the inverse  `isDisabled` instead since that's more consistent with things like button active states?

src/components/Settings/MaxSlippageSettings/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+const NUMBER_WITH_MAX_TWO_DECIMAL_PLACES = /^(?:\d*\.\d{0,2}|\d+)$/
+
+export default function MaxSlippageSettings({ placeholder }: { placeholder: Percent }) {
+  const [userSlippageTolerance, setUserSlippageTolerance] = useUserSlippageTolerance()
+
+  // If user has previously entered a custom deadline, we want to show that value in the input field
+  // instead of a placeholder by defualt
+  const [slippageInput, setSlippageInput] = useState(
+    userSlippageTolerance !== 'auto' && !userSlippageTolerance.equalTo(placeholder)
```
tinaszheng, line 43: can the 'auto' be made into an enum instead

src/components/Settings/MaxSlippageSettings/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+        setSlippageError(SlippageError.InvalidInput)
+      } else {
+        setUserSlippageTolerance(new Percent(parsed, 10_000))
+      }
+    } catch (e) {
+      setSlippageError(SlippageError.InvalidInput)
+    }
+  }
+
+  const tooLow = userSlippageTolerance !== 'auto' && userSlippageTolerance.lessThan(new Percent(5, 10_000))
```
tinaszheng, line 81: id define the 5% and 1% as constants outside so no new object on every render

src/components/Settings/MaxSlippageSettings/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+            value={slippageInput}
+            onChange={(e) => parseSlippageInput(e.target.value)}
+            onBlur={() => {
+              // When the input field is blurred, reset the input field to the current slippage tolerance
+              setSlippageInput(userSlippageTolerance !== 'auto' ? userSlippageTolerance.toFixed(2) : '')
+              setSlippageError(false)
+            }}
+          />
+          <ThemedText.BodyPrimary>
+            <Trans>%</Trans>
```
tinaszheng, line 142: does this need to be translated 😅

src/components/Settings/RouterPreferenceSettings/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+const PreferencesContainer = styled(Column)`
+  gap: 1.5px;
+  border-radius: 12px;
+  overflow: hidden;
+`
+
+export default function RouterPreferenceSettings() {
+  const [routerPreference, setRouterPreference] = useRouterPreference()
+
+  const isAutoRoutingActive = routerPreference === RouterPreference.PRICE
```
tinaszheng, line 25: there should probs be a new `RouterPreference.AUTO` that is the default :3

tinaszheng: and for now it can just alias to be the same as `RouterPreference.API` :3

src/components/Settings/TransactionDeadlineSettings/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+  InvalidInput = 'InvalidInput',
+}
+
+const THREE_DAYS_IN_SECONDS = ms`3 days` / 1000
+const NUMBERS_ONLY = /^[0-9\b]+$/
+
+export default function TransactionDeadlineSettings() {
+  const [deadline, setDeadline] = useUserTransactionTTL()
+
+  const defaultInputValue = deadline && deadline !== DEFAULT_DEADLINE_FROM_NOW ? (deadline / 60).toString() : ''
```
tinaszheng, line 22: why dont we just store deadline as minute-based and then just convert to `second` when we submit the tx 🤔  maybe outside the scope of this PR tho lol

src/components/Settings/TransactionDeadlineSettings/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  // If user has previously entered a custom deadline, we want to show that value in the input field
+  // instead of a placeholder by defualt
+  const [deadlineInput, setDeadlineInput] = useState(defaultInputValue)
+  const [deadlineError, setDeadlineError] = useState<DeadlineError | false>(false)
+
+  function parseCustomDeadline(value: string) {
```
tinaszheng, line 29: this whole function is very clever

src/components/Settings/TransactionDeadlineSettings/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+        <Row width="auto">
+          <ThemedText.BodySecondary>
+            <Trans>Transaction deadline</Trans>
+          </ThemedText.BodySecondary>
+          <QuestionHelper
+            text={<Trans>Your transaction will revert if it is pending for more than this period of time.</Trans>}
+          />
+        </Row>
+      }
+      button={<Trans>{deadline / 60}m</Trans>}
```
tinaszheng, line 69: same question about the storing in min vs seconds here 😅 seems lame that we have to convert when we only use seconds at the final step

src/components/Settings/RouterPreferenceSettings/index.tsx
tinaszheng: `RouterPreference.PRICE` is just used internally for our token -> USDC trades to get a USD value. its kind of confusing that it's a "Router preference" type when it should really just be a "quote type"

tinaszheng (commented):
LOOKS GREAT

src/components/Settings/MaxSlippageSettings/index.tsx
grabbou: Done!

src/components/Settings/MaxSlippageSettings/index.tsx
grabbou: Good idea, converted.

src/components/Settings/TransactionDeadlineSettings/index.tsx
grabbou: Not sure if that's a good thing! 😅

src/components/Settings/TransactionDeadlineSettings/index.tsx
grabbou: Yeah, I feel like this needs to be refactored, but it spans across different component. Will add a TODO for follow-ups.

src/components/Settings/TransactionDeadlineSettings/index.tsx
grabbou: Responded in another comment, marking as a TODO.

src/components/Radio/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+  border: 2px solid ${({ theme, isActive }) => (isActive ? theme.accentAction : theme.backgroundOutline)};
+  border-radius: 50%;
+  cursor: pointer;
+  display: flex;
+  outline: none;
+  padding: 5px;
+  width: fit-content;
+`
+
+const ToggleElementHoverStyle = (hasBgColor: boolean, theme: any, isActive?: boolean) =>
```
just-toby, line 18: i think you can type `theme` as `DefaultTheme`

src/components/Settings/MaxSlippageSettings/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+            }}
+          />
+          <ThemedText.BodyPrimary>%</ThemedText.BodyPrimary>
+        </InputContainer>
+      </RowBetween>
+      {tooLow || tooHigh ? (
+        <RowBetween
+          style={{
+            fontSize: '14px',
+            paddingTop: '7px',
```
just-toby, line 160: 7px seems wrong, should we round it to 8? also, i think we wanted to avoid custom `style` props and use styled components for everything.

just-toby (approved):
lgtm

src/components/Settings/index.tsx
```diff
@@ -97,89 +71,32 @@ const MenuFlyout = styled.span`
...
-  border-radius: 20px;
+  border-width: 0;
+  margin: 0;
+  background-color: ${({ theme }) => theme.backgroundOutline};
 `
 
-export default function SettingsTab({ placeholderSlippage }: { placeholderSlippage: Percent }) {
+export default function SettingsTab({ autoSlippage }: { autoSlippage: Percent }) {
   const { chainId } = useWeb3React()
+  const showDeadlineSettings = Boolean(chainId && !L2_CHAIN_IDS.includes(chainId))
```
tinaszheng, line 90: we need deadline settings for L2s as well

src/components/swap/SwapModalFooter.tsx
```diff
@@ -139,7 +140,7 @@ export default function SwapModalFooter({
             allowedSlippage,
             transactionDeadlineSecondsSinceEpoch,
             isAutoSlippage,
-            isAutoRouterApi: !clientSideRouter,
+            isAutoRouterApi: routerPreference === RouterPreference.AUTO,
```
tinaszheng, line 143: i think you'll also want `|| routerPreference === RouterPreference.API` here

tinaszheng: its kind of confusing but rn "auto router" refers to using our routing api 🤔

tinaszheng (changes requested):
just some small things but ALMOST THERE

src/components/Settings/index.tsx
tinaszheng: JK ignore meee :)

src/components/swap/SwapModalFooter.tsx
grabbou: Yes, you are correct. Overlooked it!

src/components/Radio/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+  className?: string
+  isActive: boolean
+  toggle: () => void
+}
+
+export default function Radio({ id, className, isActive, children, toggle }: PropsWithChildren<RadioProps>) {
+  return (
+    <Container className={className} onClick={toggle}>
+      {children}
+      <Button id={id} isActive={isActive}>
```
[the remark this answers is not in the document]
grabbou: I will go ahead and remove it, left-over from code I copied.

tinaszheng (approved):
image
LGTM

src/components/Settings/Input/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+        :focus-within {
+          border-color: ${theme.accentFailureSoft};
+        }
+      `
+      : `
+        border: 1px solid ${theme.backgroundOutline};
+        :focus-within {
+          border-color: ${theme.accentActiveSoft};
+        }
+      `}
```
zzmp, line 45: These should use the `css` macro

src/state/user/reducer.ts
```diff
@@ -20,13 +21,14 @@ export interface UserState {
...
+  // which router should be used to calculate trades
+  userRouterPreference: RouterPreference
 
   // hides closed (inactive) positions across the app
   userHideClosedPositions: boolean
 
   // user defined slippage tolerance in bips, used in all txns
-  userSlippageTolerance: number | 'auto'
+  userSlippageTolerance: number | SlippageTolerance.Auto
   userSlippageToleranceHasBeenMigratedToAuto: boolean // temporary flag for migration status
```
zzmp, line 32: Can this temporary flag be removed?

src/components/Settings/RouterPreferenceSettings/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+  return (
+    <Column gap="md">
+      <RowBetween>
+        <RowFixed>
+          <Column gap="xs">
+            <ThemedText.BodySecondary>
+              <Trans>Auto Routing</Trans>
+            </ThemedText.BodySecondary>
+            <ThemedText.Caption color="textSecondary">
+              <Trans>Finds the best route across liquidity sources.</Trans>
```
tinaszheng, line 36: dont forget to update this to @infredible's suggestions :D

src/components/Settings/MaxSlippageSettings/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+const MINIMUM_RECOMMENDED_SLIPPAGE = new Percent(5, 10_000)
+const MAXIMUM_RECOMMENDED_SLIPPAGE = new Percent(1, 100)
+
+export default function MaxSlippageSettings({ autoSlippage }: { autoSlippage: Percent }) {
+  const [userSlippageTolerance, setUserSlippageTolerance] = useUserSlippageTolerance()
+
+  // In order to trigger `custom` mode, we need to set `userSlippageTolerance` to a value that is not `auto`.
+  // To do so, we use `autoSlippage` value. However, since users are likely to change that value,
+  // we render it as a placeholder instead of a value.
+  const defaultSlippageInputValue =
```
zzmp, line 46: Call this `placeholderSlippageValue`

src/components/Settings/MaxSlippageSettings/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+      : ''
+
+  // If user has previously entered a custom slippage, we want to show that value in the input field
+  // instead of a placeholder.
+  const [slippageInput, setSlippageInput] = useState(defaultSlippageInputValue)
+  const [slippageError, setSlippageError] = useState<SlippageError | false>(false)
+
+  const parseSlippageInput = (value: string) => {
+    // Do not allow non-numerical characters in the input field or more than two decimals
+    if (value.length > 0 && !NUMBER_WITH_MAX_TWO_DECIMAL_PLACES.test(value)) {
```
zzmp, line 58: Why not inline this regex?

src/components/Settings/MaxSlippageSettings/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+        <RowBetween gap="md">
+          <CautionTriangle />
+          <ThemedText.Caption color="accentWarning">
+            {tooLow ? (
+              <Trans>
+                Slippage below {MINIMUM_RECOMMENDED_SLIPPAGE.toFixed(2)}% may result in a failed transaction
+              </Trans>
+            ) : (
+              <Trans>Your transaction may be frontrun and result in an unfavorable trade.</Trans>
+            )}
```
zzmp, line 168: I think this would be clearer if you guarded: `{ tooLow && <...` so that you could see both values.

src/components/Expand/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+}: PropsWithChildren<{
+  header: ReactElement
+  button: ReactElement
+}>) {
+  const [isExpanded, setExpanded] = useState(false)
+  return (
+    <Column gap="md">
+      <RowBetween>
+        {header}
+        <ButtonContainer gap="xs" onClick={() => setExpanded(!isExpanded)}>
```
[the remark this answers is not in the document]
grabbou: Good idea, however I don't think we have something like this formally written down. Thats why I didn't consider adding anything around accessibility in first place (I don't think it's in Widget either: https://github.com/Uniswap/widgets/blob/main/src/components/Expando.tsx)

src/state/user/types.ts
```diff
@@ -10,3 +10,7 @@ export interface SerializedPair {
   token0: SerializedToken
   token1: SerializedToken
 }
+
+export enum SlippageTolerance {
+  Auto = 'auto',
+}
```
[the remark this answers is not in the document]
grabbou: Yeah, we can do it this way too.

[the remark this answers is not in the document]
grabbou: I prefer `enum` usage vs doing `typeof AutoSlippageTolerance` every time when it is expected to be passed as an argument, which may turn out to be confusing (as to why it's not a primitive type)

src/components/Expand/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+          {button}
+          <ExpandIcon $isExpanded={isExpanded} />
+        </ButtonContainer>
+      </RowBetween>
+      {isExpanded && children}
```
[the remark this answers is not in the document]
grabbou: I will double check with the design team and worst case scenario, I will ship follow-up. However, I have shared the link with a few people and did not see this mentioned. I know Widget had animation, so should be straight-forward to apply!

src/components/Settings/RouterPreferenceSettings/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+          <Preference
+            isActive={routerPreference === RouterPreference.API}
+            toggle={() => setRouterPreference(RouterPreference.API)}
+          >
+            <Column gap="xs">
+              <ThemedText.BodyPrimary>
+                <Trans>Uniswap API</Trans>
+              </ThemedText.BodyPrimary>
+              <ThemedText.Caption color="textSecondary">
+                <Trans>Finds the best route on the Uniswap Protocol using the Uniswap Labs Routing API.</Trans>
```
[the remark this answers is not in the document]
grabbou: The copy was provided as a part of the design. Let me forward this suggestion.

src/components/Radio/index.tsx
```diff
@@ -0,0 +1,61 @@
+import { RowBetween } from 'components/Row'
```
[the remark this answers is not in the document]
grabbou: I did not see it required to implement a grouping component since the "active" state is controlled outside of the Radio component anyway (left up to the parent that renders and lays things out). We don't really have any occurrences of Radio anywhere outside of Settings menu, so I am expecting this component to evolve as we continue adding new features.

src/state/user/reducer.ts
grabbou: It would be best to follow-up with the original author, that has more context - I am definitely not aware of the migration and how was it performed. I will follow-up with whoever authored this code early next week. Meanwhile, I've added a TODO with Jira ticket.

src/components/Settings/MaxSlippageSettings/index.tsx
grabbou: Why not? My preference is to keep constant values outside of React component

src/components/Settings/MaxSlippageSettings/index.tsx
grabbou: It is not possible for both values to be present at the same time, since one value can't be both too low and too large. That said, I updated the code to your suggested pattern and added a test case to guard against two showing up at the same time. It's a left-over from the previous code.

src/components/Settings/MaxSlippageSettings/index.tsx
grabbou: I find `placeholder` name confusing in this context, because this value is not being used as a `placeholder` for the Input. This value is used as a `Input` value, showing the slippage value, except for default and auto mode.

src/components/Radio/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+    ? {
+        opacity: '0.8',
+      }
+    : {
+        background: isActive ? darken(0.05, theme.accentAction) : darken(0.05, theme.deprecated_bg4),
+        color: isActive ? theme.white : theme.textTertiary,
+      }
+
+const ButtonFill = styled.span<{ isActive?: boolean; bgColor?: string }>`
+  animation: 0.1s opacity ease-in;
```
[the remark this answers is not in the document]
grabbou: You are right, there are constants, I will re-use.

src/components/Radio/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const ToggleElementHoverStyle = (hasBgColor: boolean, theme: DefaultTheme, isActive?: boolean) =>
+  hasBgColor
+    ? {
+        opacity: '0.8',
+      }
+    : {
+        background: isActive ? darken(0.05, theme.accentAction) : darken(0.05, theme.deprecated_bg4),
+        color: isActive ? theme.white : theme.textTertiary,
+      }
```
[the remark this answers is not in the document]
grabbou: Correct. Will rewrite this and in `Expand`, I borrowed that code from `Toggle` component since it had similar UI behavior.

src/components/Settings/Input/index.tsx
grabbou: Why would you use a macro in this case? We don't wrap it with `css` when we do something like this:
```
${({ theme, error }) => border: 1px solid ${theme.accentFailure}
```
so I am not sure what is different about two properties. I personally use `css` when I want to reuse some styling across components, but for inline definitions like this - I don't.

src/components/Settings/RouterPreferenceSettings/index.tsx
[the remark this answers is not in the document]
grabbou: Update: sticking to Figma

src/components/Radio/index.tsx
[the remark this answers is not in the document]
grabbou: Update: I have removed this entirely, this function was not needed as `bgColor` was not supported here.

src/components/Settings/Input/index.tsx
grabbou: https://styled-components.com/docs/api#css

src/components/Radio/index.tsx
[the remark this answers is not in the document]
grabbou: For some reason, I am having a hard time debugging this animation. I will remove this from the PR, so that we can ship it, and follow-up immediately to do this and Expand animation.

src/components/Expand/index.tsx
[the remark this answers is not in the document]
grabbou: Animation is preferred. I will send a follow-up to split the workload and reduce the size of the PR.

src/state/user/reducer.ts
grabbou: After doing some research, turns out that this flag is unlikely to be removed in the future, unless we're absolutely certain that all users of Uniswap have successfully loaded their app at least once since this feature was added. This flag is used to migrate / hydrate pre-existing localStorage state with the default values required to work. Removing it will make a very very small percentage (but still) of users experience issues, unless they do hard reload.

grabbou: Perhaps in the future, we should be a bit more careful around how we roll out those features and avoid adding new flags/properties as much as possible.

src/state/user/reducer.ts
```diff
@@ -132,9 +132,11 @@ const userSlice = createSlice({
     },
   },
   extraReducers: (builder) => {
+    // After adding a new property to the state, its value will be `undefined` (instead of the default)
+    // for all existing users with a previous version of the state in their localStorage.
```
tinaszheng, line 136: oooooh this is because we don't have migrations huh

grabbou:
Update: test failure of one e2e test is not related. For some reason, the popup started to appear and covers the UI. I will open JIRA ticket for this.

image

tinaszheng (approved):

zzmp (approved):

grabbou:
Update: Unfortuantely, merging is blocked with a failing test. I tried to cherry-pick @zzmp PR that was fixing the tests, but for some reason, they keep failing here. I asked @zzmp and @JFrankfurt for help here, let's see if we can get the test fixed.

tinaszheng (approved):

bpsib:
Now there's no expert mode option anymore. You have now ruined the interface you morons. Why the hell remove such a useful option?? It's so annoying having to click confirm all the time.

PR #6481, opened 2023-05-01
chore: bubbling up pwat spender from route query

JackShort:
This is to fix the problem of hardcoding the nft backend ur contract address - we can make a dummy call to route to get the contract address and bubble it up to be used for the approval process

### Automated testing

- [x] Unit test
- [x] Manual testing


src/nft/hooks/usePermit2Approval.ts
```diff
@@ -7,26 +7,28 @@ import usePermit2Allowance, { AllowanceState } from 'hooks/usePermit2Allowance'
...
-// TODO: This should be removed when the sdk is updated to include the new UR address
-const NFT_UNIVERSAL_ROUTER_MAINNET_ADDRESS = '0x4c60051384bd2d3c01bfc845cf5f4b44bcbe9de5'
-
 export default function usePermit2Approval(
-  amount?: CurrencyAmount<Token>,
-  maximumAmount?: CurrencyAmount<Token>,
-  shouldUseNftRouter?: boolean
+  amount: CurrencyAmount<Token> | undefined,
+  maximumAmount: CurrencyAmount<Token> | undefined,
+  nftUniversalRouterContractAddress?: string
```
cbachmeier, line 13: the first two params use `| undefined` and the last uses `?` I would stick with one consistent notation (? preferred)

src/nft/components/bag/BagFooter.tsx
```diff
@@ -316,6 +317,7 @@ export const BagFooter = ({ setModalIsOpen, eventProperties }: BagFooterProps) =
   const isPending = PENDING_BAG_STATUSES.includes(bagStatus)
   const activeCurrency = inputCurrency ?? defaultCurrency
   const usingPayWithAnyToken = !!inputCurrency && chainId === SupportedChainId.MAINNET
+  const universalRouterAddress = useNftUniversalRouterAddress()
```
cbachmeier, line 320: `useNftUniversalRouterAddressQuery` also as an optional `loading` field you could pass through this hook as `universalRouterAddress` would be undefined while loading. While`universalRouterAddressIsLoading` is true you should set `tradeState` to Loading to prevent an inaccurate UR address from being used because the query was still loading.

cbachmeier (commented):
Is there anyway I can test this to get the next UR address? Connecting to mz's endpoint? Successfully getting the current UR address `0x3fC91A3afd70395Cd496C647d5a6CC9D4B2b7FAD` just want to make sure this will work for next release

src/graphql/data/nft/NftUniversalRouterAddress.ts
```diff
@@ -0,0 +1,20 @@
...
+  query NftUniversalRouterAddress($chain: Chain = ETHEREUM) {
+    nftRoute(chain: $chain, senderAddress: "", nftTrades: []) {
+      toAddress
+    }
+  }
+`
+
+export function useNftUniversalRouterAddress(): string | undefined {
+  const { data } = useNftUniversalRouterAddressQuery({
+    fetchPolicy: 'no-cache',
```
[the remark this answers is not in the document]
JackShort: since this is the same endpoint that calls the routing and it will get called around similar times - i think its better for this not to mess with the cache

src/nft/hooks/usePermit2Approval.ts
JackShort: actually this is following zachs comment in the other pr `?` defines it as optional and `| undefined` requires you to pass in undefined and it not optional

JackShort:
> `0x3fC91A3afd70395Cd496C647d5a6CC9D4B2b7FAD`

that is actually the newly deployed router so thats actually a good test

JackShort:
> Is there anyway I can test this to get the next UR address? Connecting to mz's endpoint? Successfully getting the current UR address `0x3fC91A3afd70395Cd496C647d5a6CC9D4B2b7FAD` just want to test this for the next release

that is actually the newly deployed router so thats actually a good test

cbachmeier:
> > Is there anyway I can test this to get the next UR address? Connecting to mz's endpoint? Successfully getting the current UR address `0x3fC91A3afd70395Cd496C647d5a6CC9D4B2b7FAD` just want to test this for the next release
> 
> that is actually the newly deployed router so thats actually a good test

My bad, meant to say I'm getting the current address **0x4C60051384bd2d3C01bfc845Cf5F4b44bcbE9de5** and wanted to try and get the updated one `0x3fC91A3afd70395Cd496C647d5a6CC9D4B2b7FAD`

JackShort:
> > > Is there anyway I can test this to get the next UR address? Connecting to mz's endpoint? Successfully getting the current UR address `0x3fC91A3afd70395Cd496C647d5a6CC9D4B2b7FAD` just want to test this for the next release
> > 
> > 
> > that is actually the newly deployed router so thats actually a good test
> 
> My bad, meant to say I'm getting the current address **0x4C60051384bd2d3C01bfc845Cf5F4b44bcbE9de5** and wanted to try and get the updated one `0x3fC91A3afd70395Cd496C647d5a6CC9D4B2b7FAD`

I shall ask BE if they have an endpoint that uses the updated contract - but I don't think they do

cbachmeier (approved):
Tested for updated contract and got the expected address

src/graphql/data/nft/NftUniversalRouterAddress.ts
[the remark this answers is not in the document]
tinaszheng: @JackShort i'd move this comment into the code :P

src/nft/hooks/usePermit2Approval.ts
```diff
@@ -7,26 +7,28 @@ import usePermit2Allowance, { AllowanceState } from 'hooks/usePermit2Allowance'
...
 ) {
   const { chainId } = useWeb3React()
 
-  let contractToApprove: string | undefined
-  if (chainId) {
-    const forceNftRouter = shouldUseNftRouter && chainId === 1
-    contractToApprove = forceNftRouter ? NFT_UNIVERSAL_ROUTER_MAINNET_ADDRESS : UNIVERSAL_ROUTER_ADDRESS(chainId)
+  let universalRouterAddress = nftUniversalRouterContractAddress
+  switch (chainId) {
```
tinaszheng, line 18: this switch statement is kinda awkward with the `undefined` case in between - can you just define a helper function to simplify?

```
function getURAddress(chainId?: number, nftURAddress?: string) {
   if (!chainId) return
 
   // if mainnet and on NFT flow, use the contract address returned by GQL
   if (chainId === SupportedChainId.Mainnet) {
       return nftURAddress ?? UNIVERSAL_ROUTER_ADDRESS(chainId)
   }

  return UNIVERSAL_ROUTER_ADDRESS(chainId)
}
```

tinaszheng (changes requested):
just a few nits for readability!

tinaszheng (approved):
image
ship ittt

src/graphql/data/nft/NftUniversalRouterAddress.ts
```diff
@@ -0,0 +1,26 @@
+import gql from 'graphql-tag'
+import { useMemo } from 'react'
+
+import { useNftUniversalRouterAddressQuery } from '../__generated__/types-and-hooks'
+
+gql`
+  query NftUniversalRouterAddress($chain: Chain = ETHEREUM) {
+    nftRoute(chain: $chain, senderAddress: "", nftTrades: []) {
+      toAddress
```
dannydaniil, line 9: Wondering if we considered avoiding the extra call by getting `toAddress` to the `nftRoute` call we make regardless during purchase and kicking off the approval process after that, based on that.

JackShort: they thought it should be approve -> pay isntead of pay -> approve -> pay

PR #6482, opened 2023-05-01
ci: GH action for deploying to production

lavalamp-:
## Description

This PR effectively duplicates our standard `release.yml` to deploy from our `releases/prod` branch for the new CI configuration.

Note that this PR intoduces the `prod` environment which will contain the deployment secrets necessary for this new deployment pipeline.

### QA (ie manual testing)

Once all of the CI-related PRs have been merged we will do manual end-to-end testing of the whole staging and production CI flow.

### DIFF
--- a/.github/workflows/4-deploy-to-prod.yml
+++ b/.github/workflows/4-deploy-to-prod.yml
@@ -52,7 +52,7 @@ jobs:
 
             You can also access the Uniswap Interface from an IPFS gateway.
             **BEWARE**: The Uniswap interface uses [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) to remember your settings, such as which tokens you have imported.
-            **You should always use an IPFS gateway that enforces origin separation**, or our alias to the latest release at [app.uniswap.org](https://app.uniswap.org).
+            **You should always use an IPFS gateway that enforces origin separation**, or our hosted deployment of the latest release at [app.uniswap.org](https://app.uniswap.org).
             Your Uniswap settings are never remembered across different URLs.
 
             IPFS gateways:
--- a/.github/workflows/4-deploy-to-prod.yml
+++ b/.github/workflows/4-deploy-to-prod.yml
@@ -7,7 +7,6 @@ on:
 jobs:
 
   tag:
-    needs: wait-on-tests
     runs-on: ubuntu-latest
     outputs:
       new_tag: ${{ steps.github-tag-action.outputs.new_tag }}
--- a/.github/workflows/4-deploy-to-prod.yml
+++ b/.github/workflows/4-deploy-to-prod.yml
@@ -1,0 +1,104 @@
+name: Deploy to prod
+on:
+  push:
+    branches:
+      - 'releases/prod'
+
+jobs:
+
+  tag:
+    needs: wait-on-tests
+    runs-on: ubuntu-latest
+    outputs:
+      new_tag: ${{ steps.github-tag-action.outputs.new_tag }}
+      changelog: ${{ steps.github-tag-action.outputs.changelog }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Bump and tag
+        id: github-tag-action
+        uses: mathieudutour/github-tag-action@v6.0
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          release_branches: .*
+          default_bump: patch
+
+  deploy-to-prod:
+    environment:
+      name: prod
+    steps:
+
+      - uses: actions/checkout@v3
+
+      - uses: ./.github/actions/setup
+
+      - run: yarn prepare
+
+      - run: yarn build
+
+      - name: Pin to IPFS
+        id: pinata
+        uses: anantaramdas/ipfs-pinata-deploy-action@39bbda1ce1fe24c69c6f57861b8038278d53688d
+        with:
+          pin-name: Uniswap ${{ needs.tag.outputs.new_tag }}
+          path: './build'
+          pinata-api-key: ${{ secrets.PINATA_API_KEY }}
+          pinata-secret-api-key: ${{ secrets.PINATA_API_SECRET_KEY }}
+
+      - name: Convert CIDv0 to CIDv1
+        id: convert-cidv0
+        uses: uniswap/convert-cidv0-cidv1@v1.0.0
+        with:
+          cidv0: ${{ steps.pinata.outputs.hash }}
+
+      - name: Release
+        uses: actions/create-release@v1.1.0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ needs.tag.outputs.new_tag }}
+          release_name: Release ${{ needs.tag.outputs.new_tag }}
+          body: |
+            IPFS hash of the deployment:
+            - CIDv0: `${{ steps.pinata.outputs.hash }}`
+            - CIDv1: `${{ steps.convert-cidv0.outputs.cidv1 }}`
+
+            The latest release is always accessible via our alias to the Cloudflare IPFS gateway at [app.uniswap.org](https://app.uniswap.org).
+
+            You can also access the Uniswap Interface directly from an IPFS gateway.
+            **BEWARE**: The Uniswap interface uses [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) to remember your settings, such as which tokens you have imported.
+            **You should always use an IPFS gateway that enforces origin separation**, or our alias to the latest release at [app.uniswap.org](https://app.uniswap.org).
+            Your Uniswap settings are never remembered across different URLs.
+
+            IPFS gateways:
+            - https://${{ steps.convert-cidv0.outputs.cidv1 }}.ipfs.dweb.link/
+            - https://${{ steps.convert-cidv0.outputs.cidv1 }}.ipfs.cf-ipfs.com/
+            - [ipfs://${{ steps.pinata.outputs.hash }}/](ipfs://${{ steps.pinata.outputs.hash }}/)
+
+            ${{ needs.tag.outputs.changelog }}
+
+      - name: Setup node@16 (required by Cloudflare Pages)
+        uses: actions/setup-node@v3
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/4-deploy-to-prod.yml
```diff
@@ -0,0 +1,103 @@
...
+  push:
+    branches:
+      - 'releases/prod'
+
+jobs:
+
+  tag:
+    runs-on: ubuntu-latest
+    outputs:
+      new_tag: ${{ steps.github-tag-action.outputs.new_tag }}
```
zzmp, line 12: If we're aliasing it anyway, why not just call it `tag`?

```suggestion
      tag: ${{ steps.github-tag-action.outputs.new_tag }}
```

.github/workflows/4-deploy-to-prod.yml
```diff
@@ -0,0 +1,103 @@
...
+      changelog: ${{ steps.github-tag-action.outputs.changelog }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Bump and tag
+        id: github-tag-action
+        uses: mathieudutour/github-tag-action@v6.0
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          release_branches: .*
```
zzmp, line 21: Should this be made `releases/prod`?

.github/workflows/4-deploy-to-prod.yml
```diff
@@ -0,0 +1,103 @@
...
+      name: prod
+    steps:
+
+      - uses: actions/checkout@v3
+
+      - uses: ./.github/actions/setup
+
+      - run: yarn prepare
+
+      - run: yarn build
```
zzmp, line 35:
```suggestion
      - uses: actions/checkout@v3
      - uses: ./.github/actions/setup
      - run: yarn prepare
      - run: yarn build
```

.github/workflows/4-deploy-to-prod.yml
```diff
@@ -0,0 +1,103 @@
...
+      - name: Bump and tag
+        id: github-tag-action
+        uses: mathieudutour/github-tag-action@v6.0
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          release_branches: .*
+          default_bump: patch
+
+  deploy-to-prod:
+    environment:
```
zzmp, line 25: If they are specified as separate jobs, this will need `needs: tag`. However, I don't know what is gained by separating them. Why not include `Bump and tag` as a step in one single job?

```suggestion
    needs: tag
    environment:
```

.github/workflows/4-deploy-to-prod.yml
```diff
@@ -0,0 +1,103 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ needs.tag.outputs.new_tag }}
+          release_name: Release ${{ needs.tag.outputs.new_tag }}
+          body: |
+            IPFS hash of the deployment:
+            - CIDv0: `${{ steps.pinata.outputs.hash }}`
+            - CIDv1: `${{ steps.convert-cidv0.outputs.cidv1 }}`
+
+            The latest release is always accessible via our alias to the Cloudflare IPFS gateway at [app.uniswap.org](https://app.uniswap.org).
```
zzmp, line 64: We should amend this.

```suggestion
            The latest release is always mirrored at [app.uniswap.org](https://app.uniswap.org).
```

.github/workflows/4-deploy-to-prod.yml
```diff
@@ -0,0 +1,103 @@
...
+
+            You can also access the Uniswap Interface directly from an IPFS gateway.
```
zzmp, line 66:
```suggestion
            You can also access the Uniswap Interface from an IPFS gateway.
```

.github/workflows/4-deploy-to-prod.yml
```diff
@@ -0,0 +1,103 @@
...
+            - [ipfs://${{ steps.pinata.outputs.hash }}/](ipfs://${{ steps.pinata.outputs.hash }}/)
+
+            ${{ needs.tag.outputs.changelog }}
+
+      - name: Setup node@16 (required by Cloudflare Pages)
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
+
+      - name: Update Cloudflare Pages deployment
```
zzmp, line 83: If you're going to split anything into a separate job, it should be cloudflare/sourcemaps, but I don't know if this is worthwhile. It'd require saving the build artifact between jobs, and I think it's out of scope.

.github/workflows/4-deploy-to-prod.yml
```diff
@@ -0,0 +1,103 @@
...
+        uses: cloudflare/pages-action@364c7ca09a4b57837c5967871d64a2c31adb8c0d
```
zzmp, line 84: Do we need to override the branch, because it is currently configured to release from `main`? Or should we be updating the CF setting when we are ready to deploy this?

.github/workflows/4-deploy-to-prod.yml
```diff
@@ -0,0 +1,103 @@
...
+        with:
+          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+          projectName: ${{ secrets.CLOUDFLARE_PROJECT_NAME }}
+          directory: build
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Upload source maps to Sentry
+        uses: getsentry/action-release@4744f6a65149f441c5f396d5b0877307c0db52c7
+        continue-on-error: true
```
zzmp, line 94: Should we be continuing-on-error for IPFS, considering that we could have a CF release without a successful IPFS release? Should they be decoupled?

zzmp (changes requested):

.github/workflows/4-deploy-to-prod.yml
lavalamp-: I don't _think_ so but I'm not certain. The documentation can be found here:

https://github.com/mathieudutour/github-tag-action#filter-branches

My read on this is that it's a separate action-level filter. So the workflow definition could trigger on _any_ branch and if you wanted to limit the branches that this tag job applied to then you could specify branches here in this argument.

Thus in this action's current state, the job only kicks off from `releases/prod` and then when the tagging step happens `.*` matches `releases/prod` so the tagging action proceeds.

.github/workflows/4-deploy-to-prod.yml
lavalamp-: Yeah good call - I'll fold them into the same job.

.github/workflows/4-deploy-to-prod.yml
lavalamp-: Got it - makes sense!

.github/workflows/4-deploy-to-prod.yml
lavalamp-: Mmmmm I'm not sure I'm qualified to make that judgment call.

In the spirit of transparency / community I'd think that failing to pin to IPFS but successfully deploying via a centralized platform would be poorly received.

Does this failure mode happen often? If not then probably not worth considering decoupling.

.github/workflows/4-deploy-to-prod.yml
lavalamp-: I think the cleaner way would be to configure the Cloudflare page to accept deploys from both `main` branch and `releases/prod`. If we can't configure the Cloudflare page like that then overriding branch would be my second choice.

.github/workflows/4-deploy-to-prod.yml
lavalamp-: Changed in my own commit.

.github/workflows/4-deploy-to-prod.yml
lavalamp-: Changed in my own commit.

.github/workflows/4-deploy-to-prod.yml
zzmp: We can only configure a single branch to be "production".

.github/workflows/4-deploy-to-prod.yml
zzmp: I don't think I've ever seen it happen, so let's leave as-is.

.github/workflows/4-deploy-to-prod.yml
```diff
@@ -0,0 +1,90 @@
...
+          body: |
+            IPFS hash of the deployment:
+            - CIDv0: `${{ steps.pinata.outputs.hash }}`
+            - CIDv1: `${{ steps.convert-cidv0.outputs.cidv1 }}`
+
+            The latest release is always mirrored at [app.uniswap.org](https://app.uniswap.org).
+
+            You can also access the Uniswap Interface from an IPFS gateway.
+            **BEWARE**: The Uniswap interface uses [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) to remember your settings, such as which tokens you have imported.
+            **You should always use an IPFS gateway that enforces origin separation**, or our alias to the latest release at [app.uniswap.org](https://app.uniswap.org).
```
zzmp, line 55:
```suggestion
            **You should always use an IPFS gateway that enforces origin separation**, or our hosted deployment of the latest release at [app.uniswap.org](https://app.uniswap.org).
```

zzmp (changes requested):
down to nits: overriding branch names and one language nit in the release summary.

grabbou (approved):

.github/workflows/4-deploy-to-prod.yml
```diff
@@ -0,0 +1,91 @@
...
+      - 'releases/prod'
+
+jobs:
+  deploy-to-prod:
+    environment:
+      name: prod
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup
+      - run: yarn prepare
```
zzmp, line 14: `yarn prepare` is now broken into its components in `actions/setup` to better utilize caching, but I think it should remain here so that the build is not dependent on caching.

zzmp (approved):

grabbou (approved):

PR #6486, opened 2023-05-02
fix: update service worker to precache css/js files

vm:
## Description
We are getting a significant number of errors where chunks are not being found:
https://uniswap-labs.sentry.io/issues/4009148084/?project=4504255148851200&referrer=release-issue-stream
https://uniswap-labs.sentry.io/issues/4131879787/?project=4504255148851200&referrer=release-issue-stream

We think this is caused by cache misses from the service worker. After some investigation, it seems like the logic for separating the pre-cached and on-demand cached content could be leading to the issue. Previously, we only pre-cached the document, but now we want to pre-cache all JS/CSS files, while all media assets can be on-demand cached.

The downside here is many files that may not be needed will get stored to disk, but this is a compromise we are willing to make for now. Over time we could make smarter decisions like not precaching locales, but we'd need to make improvements to the build system to make that happen.

I needed to make a change to the service worker e2e test because it was being flakey. This is unrelated to this code change but needed to make sure the tests passed before merging.

### Reproducing the error
1. add console.logs for `self.__WB_MANIFEST`, `assets` and `entries` in `src/serviceWorker/index.ts` to see what the values look like.
2. run `yarn build` and `yarn serve`
3. Check console to see that entries only has one element in it.

After changes:
Screen Shot 2023-05-03 at 12 07 02 PM
Screen Shot 2023-05-03 at 12 07 07 PM

### Automated testing
- [x] Unit test

### DIFF
--- a/cypress/e2e/service-worker.test.ts
+++ b/cypress/e2e/service-worker.test.ts
@@ -53,7 +53,8 @@ describe('Service Worker', () => {
   it('installs a ServiceWorker', () => {
     cy.visit('/', { serviceWorker: true })
       .get('#swap-page')
-      .wait('@NotInstalled', { timeout: 20000 })
+      // This is emitted after caching the entry file, which takes some time to load.
+      .wait('@NotInstalled', { timeout: 60000 })
       .window({ timeout: 20000 })
       .and((win) => {
         expect(win.navigator.serviceWorker.controller?.state).to.equal('activated')
--- a/cypress/e2e/service-worker.test.ts
+++ b/cypress/e2e/service-worker.test.ts
@@ -55,7 +55,7 @@ describe('Service Worker', () => {
       .get('#swap-page')
       // This is emitted after caching the entry file, which takes some time to load.
       .wait('@NotInstalled', { timeout: 60000 })
-      .window({ timeout: 20000 })
+      .window()
       .and((win) => {
         expect(win.navigator.serviceWorker.controller?.state).to.equal('activated')
       })
--- a/cypress/e2e/service-worker.test.ts
+++ b/cypress/e2e/service-worker.test.ts
@@ -8,8 +8,8 @@ describe('Service Worker', () => {
       if (!isValid) {
         throw new Error(
           '\n' +
-          'Service Worker tests must be run on a production-like build\n' +
-          'To test, build with `yarn build:e2e` and serve with `yarn serve`'
+            'Service Worker tests must be run on a production-like build\n' +
+            'To test, build with `yarn build:e2e` and serve with `yarn serve`'
         )
       }
     })
--- a/cypress/e2e/service-worker.test.ts
+++ b/cypress/e2e/service-worker.test.ts
@@ -8,8 +8,8 @@ describe('Service Worker', () => {
       if (!isValid) {
         throw new Error(
           '\n' +
-            'Service Worker tests must be run on a production-like build\n' +
-            'To test, build with `yarn build:e2e` and serve with `yarn serve`'
+          'Service Worker tests must be run on a production-like build\n' +
+          'To test, build with `yarn build:e2e` and serve with `yarn serve`'
         )
       }
     })
--- a/src/serviceWorker/index.ts
+++ b/src/serviceWorker/index.ts
@@ -20,15 +20,18 @@ self.skipWaiting()
 registerRoute(new DocumentRoute())
 
 // Splits entries into assets, which are loaded on-demand; and entries, which are precached.
-// Effectively, this precaches the document, and caches all other assets on-demand.
+// Effectively, this caches all media assets on-demand and pre-caches everything else.
 const { assets, entries } = self.__WB_MANIFEST.reduce<{ assets: string[]; entries: PrecacheEntry[] }>(
   ({ assets, entries }, entry) => {
     if (typeof entry === 'string') {
+      // If the entry is a string, it's the index.html file.
       return { entries, assets: [...assets, entry] }
-    } else if (entry.revision) {
-      return { entries: [...entries, entry], assets }
-    } else {
+    } else if (entry.url.includes('/media/')) {
+      // If the entry is a media file, it's an asset.
       return { entries, assets: [...assets, toURL(entry)] }
+    } else {
+      // Otherwise, it's an entry.
+      return { entries: [...entries, entry], assets }
     }
   },
   { assets: [], entries: [] }
--- a/src/serviceWorker/index.ts
+++ b/src/serviceWorker/index.ts
@@ -7,7 +7,7 @@ import { registerRoute, Route } from 'workbox-routing'
 import { CacheFirst } from 'workbox-strategies'
 
 import { DocumentRoute } from './document'
-import { groupEntries } from './utils'
+import { groupEntries, toURL } from './utils'
 
 declare const self: ServiceWorkerGlobalScope
 
@@ -23,9 +23,10 @@ registerRoute(new DocumentRoute())
 const { onDemandCacheEntries, precacheEntries } = groupEntries(self.__WB_MANIFEST)
 
 // Registers the onDemandCacheEntries' routes for on-demand caching.
+const onDemandCacheURLs = onDemandCacheEntries.map(toURL)
 registerRoute(
   new Route(
-    ({ url }) => onDemandCacheEntries.includes('.' + url.pathname),
+    ({ url }) => onDemandCacheURLs.includes('.' + url.pathname),
     new CacheFirst({
       cacheName: 'assets',
       plugins: [new ExpirationPlugin({ maxEntries: 16 })],
--- a/src/serviceWorker/index.ts
+++ b/src/serviceWorker/index.ts
@@ -7,7 +7,7 @@ import { registerRoute, Route } from 'workbox-routing'
 import { CacheFirst } from 'workbox-strategies'
 
 import { DocumentRoute } from './document'
-import { splitAssetsAndEntries } from './utils'
+import { groupEntries } from './utils'
 
 declare const self: ServiceWorkerGlobalScope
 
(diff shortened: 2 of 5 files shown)


src/serviceWorker/utils.ts
```diff
@@ -7,15 +7,35 @@ declare const self: ServiceWorkerGlobalScope
...
   return typeof entry === 'string' ? entry : entry.url
 }
+
+export const splitAssetsAndEntries = (
+  resources: (string | PrecacheEntry)[]
+): { assets: string[]; entries: PrecacheEntry[] } => {
+  return resources.reduce<{ assets: string[]; entries: PrecacheEntry[] }>(
+    ({ assets, entries }, entry) => {
+      if (typeof entry === 'string') {
+        // If the entry is a string, it's the index.html file.
```
vm, line 29: wait, why are we doing this actually?

src/serviceWorker/index.ts
```diff
@@ -20,19 +19,8 @@ self.skipWaiting()
...
-    } else if (entry.revision) {
-      return { entries: [...entries, entry], assets }
-    } else {
-      return { entries, assets: [...assets, toURL(entry)] }
-    }
-  },
-  { assets: [], entries: [] }
-)
+// Effectively, this caches all media assets on-demand and pre-caches everything else.
+const { assets, entries } = splitAssetsAndEntries(self.__WB_MANIFEST)
```
zzmp, line 23: Change these names from `assets`/`entries` to make this more clear.

zzmp (changes requested):

src/serviceWorker/index.ts
```diff
@@ -20,24 +19,13 @@ self.skipWaiting()
...
+const { onDemandCacheEntries, precacheEntries } = groupEntries(self.__WB_MANIFEST)
 
-// Registers the assets' routes for on-demand caching.
+// Registers the onDemandCacheEntries' routes for on-demand caching.
 registerRoute(
   new Route(
-    ({ url }) => assets.includes('.' + url.pathname),
+    ({ url }) => onDemandCacheEntries.includes('.' + url.pathname),
     new CacheFirst({
       cacheName: 'assets',
```
vm, line 30: do we wanna rename the cache?

src/serviceWorker/index.ts
```diff
@@ -20,24 +19,14 @@ self.skipWaiting()
 registerRoute(new DocumentRoute())
 
 // Splits entries into assets, which are loaded on-demand; and entries, which are precached.
```
zzmp, line 21: This comment (and L22) needs to be updated.

src/serviceWorker/utils.ts
```diff
@@ -19,3 +19,17 @@ export function isDevelopment() {
 export function toURL(entry: string | PrecacheEntry): string {
   return typeof entry === 'string' ? entry : entry.url
 }
+
+type GroupedEntries = { onDemandCacheEntries: (string | PrecacheEntry)[]; precacheEntries: PrecacheEntry[] }
```
zzmp, line 23: I think it'll simplify readability if you return this as `mediaUrls: string[]`. It'll avoid further transforms, and let you name things knowing its acting on media.

```suggestion
type GroupedEntries = { mediaUrls: string[]; precacheEntries: PrecacheEntry[] }
```

src/serviceWorker/utils.ts
```diff
@@ -19,3 +19,17 @@ export function isDevelopment() {
...
+export const groupEntries = (resources: (string | PrecacheEntry)[]): GroupedEntries => {
+  return resources.reduce<GroupedEntries>(
+    ({ onDemandCacheEntries, precacheEntries }, entry) => {
+      if (typeof entry === 'string' || entry.url.includes('/media/')) {
+        return { precacheEntries, onDemandCacheEntries: [...onDemandCacheEntries, entry] }
```
zzmp, line 28:
```suggestion
      if (typeof entry === 'string') {
        return { precacheEntries, mediaUrls: [...mediaUrls, entry] }
      } else if (entry.url.includes('/media/')) {
        return { precacheEntries, mediaUrls: [...onDemandCacheEntries, entry.url] }
```

src/serviceWorker/index.ts
zzmp: `media`

src/serviceWorker/utils.test.ts
```diff
@@ -0,0 +1,67 @@
...
+import { groupEntries } from './utils'
+
+describe('groupEntries', () => {
+  test('splits resources into onDemandCacheEntries and precacheEntries', () => {
+    const resources = [
+      './static/whitepaper.pdf',
+      { url: './static/js/main.js', revision: 'abc123' },
+      { url: './static/css/styles.css', revision: 'def456' },
+      { url: './static/media/image.jpg', revision: 'ghi789' },
+      { url: '/other-page.html', revision: 'jkl012' },
```
zzmp, line 12:
```suggestion
      { url: '/index.html', revision: 'jkl012' },
```

src/serviceWorker/utils.test.ts
```diff
@@ -0,0 +1,67 @@
+import { PrecacheEntry } from 'workbox-precaching/_types'
+
+import { groupEntries } from './utils'
+
+describe('groupEntries', () => {
+  test('splits resources into onDemandCacheEntries and precacheEntries', () => {
+    const resources = [
+      './static/whitepaper.pdf',
```
zzmp, line 8: Is this a realistic resource entry?

src/serviceWorker/utils.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  test('handles input with only precacheEntries', () => {
+    const resources = [
+      { url: '/other-page.html', revision: 'abc123' },
+      { url: '/about.html', revision: 'def456' },
+    ]
+
+    const result = groupEntries(resources)
+
+    expect(result).toEqual({ onDemandCacheEntries: [], precacheEntries: resources })
+  })
```
zzmp, line 66: This doesn't happen, so I don't think it needs to be tested. You can simplify by just including the first test.

```suggestion
```

zzmp (changes requested):

src/serviceWorker/utils.test.ts
vm: think so yes. it's just an example of a string url without a revision.

src/serviceWorker/index.ts
```diff
@@ -19,31 +18,20 @@ self.skipWaiting()
...
+const { mediaURLs, precacheEntries } = groupEntries(self.__WB_MANIFEST)
 
-// Registers the assets' routes for on-demand caching.
 registerRoute(
   new Route(
-    ({ url }) => assets.includes('.' + url.pathname),
+    ({ url }) => mediaURLs.includes('.' + url.pathname),
     new CacheFirst({
-      cacheName: 'assets',
+      cacheName: 'media',
```
zzmp, line 30: Will this remove the old `assets` cache?

src/serviceWorker/utils.ts
```diff
@@ -16,6 +16,18 @@ export function isDevelopment() {
   )
 }
 
-export function toURL(entry: string | PrecacheEntry): string {
-  return typeof entry === 'string' ? entry : entry.url
+type GroupedEntries = { mediaURLs: string[]; precacheEntries: PrecacheEntry[] }
+export const groupEntries = (entries: (string | PrecacheEntry)[]): GroupedEntries => {
```
zzmp, line 20: nit:

```suggestion
export function groupEntries(entries: (string | PrecacheEntry)[]): GroupedEntries {
```

zzmp (approved):

src/serviceWorker/index.ts
vm: it will because we call `skipWaiting` at the top which clears everything

zzmp (approved):

src/serviceWorker/index.ts
```diff
@@ -19,31 +18,20 @@ self.skipWaiting()
...
-    } else {
-      return { entries, assets: [...assets, toURL(entry)] }
-    }
-  },
-  { assets: [], entries: [] }
-)
+// Splits entries into:
+// - mediaURLs: loaded on-demand
+// - precacheEntries
+const { mediaURLs, precacheEntries } = groupEntries(self.__WB_MANIFEST)
```
cmcewen, line 24: nit: i think the naming isn't really clear here - it seems like you basically just want to group URLs into either precache or on-demand (no cache), but i don't know why mediaURLs aren't cached as a reader of this. i think something like `mediaURLs` and `pageAssetURLs` or `ondemandEntries` and `precacheEntries` seems like a more clear grouping to me

cmcewen (approved):
`The downside here is many files that may not be needed will get stored to disk, but this is a compromise we are willing to make for now.` could you say more about this? how did you quantify the compromise?

src/serviceWorker/index.ts
vm: i had that before but zach felt differently about it. ill ask him again

cypress/e2e/service-worker.test.ts
```diff
@@ -53,8 +53,9 @@ describe('Service Worker', () => {
   it('installs a ServiceWorker', () => {
     cy.visit('/', { serviceWorker: true })
       .get('#swap-page')
-      .wait('@NotInstalled', { timeout: 20000 })
-      .window({ timeout: 20000 })
+      // This is emitted after caching the entry file, which takes some time to load.
+      .wait('@NotInstalled', { timeout: 60000 })
```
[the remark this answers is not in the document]
vm: added

src/serviceWorker/index.ts
vm: if he agrees i'll change in a diff pr.

zzmp: I like `onDemandEntries` and `precacheEntries`, but we are looking at the url of the on demand entries, so I think it simplifies the downstream code to use `{ onDemandURLs, precacheEntries }`.

vm: personally, think it'd be better to do `onDemandEntries` because then the function doesn't need to be coupled to the use of it. it's not complicated to them get the URLs from the list of onDemandEntries.

vm: OG cmment here https://github.com/Uniswap/interface/pull/6486#discussion_r1184016228

PR #6487, opened 2023-05-03
test(e2e): reset hardhat between tests

zzmp:
## Description

Resets cypress between e2e tests.

### Automated testing

- [x] Integration/E2E test


cartcrom (approved):

JFrankfurt (approved):

hardhat.config.js
```diff
@@ -22,8 +22,8 @@ module.exports = {
         count: 1,
       },
       mining: {
-        auto: true,
-        interval: 0,
+        auto: true, // automine to make tests easier to write.
+        interval: 0, // do not interval mine so that tests remain deterministic
```
JFrankfurt, line 26: nit: these lines can be removed, since we're using the defaults for mining config--or you could just remove the interval setting if you want to be explicit about auto

zzmp: I like that this is more explicit, in case someone tries to be clever with the settings later on.

PR #6489, opened 2023-05-03
fix: add amplitude page context to swap flow events

just-toby:
## Description

Add analytics context to swap flow events as event properties. specifically we care mostly about the `page` property because we want to differentiate between TDP and main swap page.

## Screen capture

Screenshot 2023-05-03 at 11 18 18 AM
Screenshot 2023-05-03 at 11 18 42 AM
Screenshot 2023-05-03 at 11 22 09 AM
Screenshot 2023-05-03 at 11 30 03 AM

### Reproducing the error

1.  set `debug: true` on the amplitude initialization and view the events in chrome inspector - see that the `page` is not included

### QA (ie manual testing)

- [x] manually verified that `page` is included in the events now

### Automated testing

- [x] <s>Unit test</s>

### DIFF
--- a/src/hooks/useUniversalRouter.ts
+++ b/src/hooks/useUniversalRouter.ts
@@ -1,7 +1,7 @@
 import { TransactionResponse } from '@ethersproject/abstract-provider'
 import { BigNumber } from '@ethersproject/bignumber'
 import { t } from '@lingui/macro'
-import { sendAnalyticsEvent } from '@uniswap/analytics'
+import { sendAnalyticsEvent, useTrace } from '@uniswap/analytics'
 import { SwapEventName } from '@uniswap/analytics-events'
 import { Trade } from '@uniswap/router-sdk'
 import { Currency, Percent, TradeType } from '@uniswap/sdk-core'
@@ -49,6 +49,7 @@ export function useUniversalRouterSwapCallback(
   options: SwapOptions
 ) {
   const { account, chainId, provider } = useWeb3React()
+  const analyticsContext = useTrace()
 
   return useCallback(async (): Promise<TransactionResponse> => {
     return trace(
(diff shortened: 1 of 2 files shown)


src/hooks/useUniversalRouter.ts
```diff
@@ -49,6 +49,7 @@ export function useUniversalRouterSwapCallback(
   options: SwapOptions
 ) {
   const { account, chainId, provider } = useWeb3React()
+  const analyticsContext = useTrace()
```
lynnshaoyu, line 52: seems like its called "trace" throughout the interface codebase (and also in src/pages/Swap/index.tsx where you just added). could we rename it to trace here to keep consistent?

just-toby: i used a different name here to disambiguate from the `trace` function being used to do perf monitoring

zzmp: I'd love for a followup to change the export in `@uniswap/analytics` to just be `useAnalyticsContext` so that it can cleanly be named `analyticsContext`. Could you ticket that and follow up on it?

If we _always_ want the trace to be included with an analytics event, it may also be useful to have a `useSendAnalyticsEvent` hook exported which includes the trace in the sent event properties.

src/hooks/useUniversalRouter.ts
```diff
@@ -90,12 +91,20 @@ export function useUniversalRouterSwapCallback(
...
-              )
+              const eventProperties = {
+                ...formatSwapSignedAnalyticsEventProperties({
+                  trade,
+                  fiatValues,
+                  txHash: response.hash,
+                }),
+                ...analyticsContext,
+              }
+              sendAnalyticsEvent(SwapEventName.SWAP_SIGNED, eventProperties)
```
zzmp, line 102: nit:

```suggestion
              sendAnalyticsEvent(SwapEventName.SWAP_SIGNED, {                
                ...formatSwapSignedAnalyticsEventProperties({
                  trade,
                  fiatValues,
                  txHash: response.hash,
                }),
                ...analyticsContext,
              })
```

zzmp (approved):
lgtm, with some ideas for future cleanup

src/hooks/useUniversalRouter.ts
just-toby: https://linear.app/uniswap/issue/WEB-17/analytics-improvements

adjkant: maybe naming here: analyticsTrace?

lynnshaoyu (approved):

PR #6493, opened 2023-05-04
feat: update useAllTokensMultichain usage

cartcrom:
## Description

Updates local tx parser to use useAllTokensMultichain, and updates useAllTokensMultichain to merge tokens from all lists w/ userAdded tokens.

This fixes an existing problem where Mini Portfolio and & the TX Notification toast would only render tokens from the default list.

[Slack thread](https://uniswapteam.slack.com/archives/C04QKTF9T35/p1683220935782969)

## Screen capture

| ----------- |---------------- |
| Screen Shot 2023-05-04 at 4 52 03 PM | image      |

### Reproducing the error

1. Swap a token that is not included on the default list & view the tx notification or Activity tab

### DIFF
--- a/src/hooks/Tokens.ts
+++ b/src/hooks/Tokens.ts
@@ -5,6 +5,7 @@ import { SupportedChainId } from 'constants/chains'
 import { DEFAULT_INACTIVE_LIST_URLS, DEFAULT_LIST_OF_LISTS } from 'constants/lists'
 import { useCurrencyFromMap, useTokenFromMapOrNetwork } from 'lib/hooks/useCurrency'
 import { getTokenFilter } from 'lib/hooks/useTokenList/filtering'
+import { TokenAddressMap } from 'lib/hooks/useTokenList/utils'
 import { useMemo } from 'react'
 import { useAppSelector } from 'state/hooks'
 import { isL2ChainId } from 'utils/chains'
@@ -12,7 +13,7 @@ import { isL2ChainId } from 'utils/chains'
 import { useAllLists, useCombinedActiveList, useCombinedTokenMapFromUrls } from '../state/lists/hooks'
 import { WrappedTokenInfo } from '../state/lists/wrappedTokenInfo'
 import { deserializeToken, useUserAddedTokens, useUserAddedTokensOnChain } from '../state/user/hooks'
-import { TokenAddressMap, useUnsupportedTokenList } from './../state/lists/hooks'
+import { useUnsupportedTokenList } from './../state/lists/hooks'
 
 type Maybe<T> = T | null | undefined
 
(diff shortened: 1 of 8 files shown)


src/hooks/Tokens.ts
```diff
@@ -28,8 +29,36 @@ function useTokensFromMap(tokenMap: TokenAddressMap, chainId: Maybe<SupportedCha
   }, [chainId, tokenMap])
 }
 
-export function useAllTokensMultichain(): TokenAddressMap {
-  return useCombinedTokenMapFromUrls(DEFAULT_LIST_OF_LISTS)
+export type ChainTokenMap = { [chainId in number]?: { [address in string]?: Token } }
```
zzmp, line 32: How is this different from useDefaultActiveTokens? Could you just reuse that instead to i improve code reuse?

zzmp (changes requested):

src/hooks/Tokens.ts
cartcrom: `DefaultActiveTokens` are tokens on the default list + user added tokens. 

`AllTokensMultichain` are tokens from any list + user added tokens

src/hooks/Tokens.ts
```diff
@@ -28,8 +29,36 @@ function useTokensFromMap(tokenMap: TokenAddressMap, chainId: Maybe<SupportedCha
...
+export function useAllTokensMultichain(): ChainTokenMap {
```
zzmp, line 33: If this file already has tests, this should include a unit test.

zzmp (approved):

src/hooks/Tokens.ts
```diff
@@ -28,8 +29,36 @@ function useTokensFromMap(tokenMap: TokenAddressMap, chainId: Maybe<SupportedCha
...
+      .map(Number)
+      .forEach((chainId) => {
+        const tokenMap = {} as { [address in string]?: Token }
+        Object.values(userAddedTokensMap[chainId]).forEach((serializedToken) => {
+          tokenMap[serializedToken.address] = deserializeToken(serializedToken)
+        })
+        chainTokenMap[chainId] = tokenMap
+      })
+
+    Object.keys(allTokensFromLists)
```
just-toby, line 50: to easily unit test this, you could put this logic in a regular function, write tests for that function, and call it from this useMemo hook

src/hooks/Tokens.ts
```diff
@@ -28,8 +29,36 @@ function useTokensFromMap(tokenMap: TokenAddressMap, chainId: Maybe<SupportedCha
...
+export type ChainTokenMap = { [chainId in number]?: { [address in string]?: Token } }
+export function useAllTokensMultichain(): ChainTokenMap {
+  const allTokensFromLists = useCombinedTokenMapFromUrls(DEFAULT_LIST_OF_LISTS)
+  const userAddedTokensMap = useAppSelector(({ user: { tokens } }) => tokens)
+
+  return useMemo(() => {
+    const chainTokenMap: ChainTokenMap = {}
+
+    Object.keys(userAddedTokensMap)
+      .map(Number)
```
just-toby, line 41: maybe you should cast these keys to `Number` in the forEach body itself, instead of `map` - then we would iterate through the array once instead of twice, right?

zzmp (approved):

JFrankfurt (approved):

PR #6495, opened 2023-05-05
build: cache generated files across builds

zzmp:
## Description

Splits the `yarn prepare` step so that its components can utilize caching in CI builds.
Updates the setup action workflow to cache compiled contracts, generated graphql schema, extracted translations, and compiled localizations across builds if they have not changed.

`yarn prepare` will continue to work locally, but it's intention is now to be used to setup local development. CI, which was partially bottle-necked by these cacheable steps, will now call directly into `yarn prepare`s component steps.

### QA (ie manual testing)
- [ ] Integration/E2E test


lingui.config.ts
```diff
@@ -1,8 +1,60 @@
+import { default as babelExtractor } from '@lingui/cli/api/extractors/babel'
+import { createHash } from 'crypto'
+import { mkdirSync, readFileSync, writeFileSync } from 'fs'
+import * as path from 'path'
+import * as pkgUp from 'pkg-up'
```
JFrankfurt, line 5: you might consider find-up instead, as it's the same functionality with more usage https://www.npmjs.com/package/find-up

lingui.config.ts
```diff
@@ -1,8 +1,60 @@
...
+    const root = path.dirname(pkg)
+
+    const filePath = path.join(root, filename)
+    const file = readFileSync(filePath)
+    const hash = createHash('sha256').update(file).digest('hex')
+
+    const cacheRoot = path.join(root, 'node_modules/.cache/lingui')
+    mkdirSync(cacheRoot, { recursive: true })
+    const cachePath = path.join(cacheRoot, filename.replace(/\//g, '-'))
+    if (!process.argv.includes('--clean')) {
```
JFrankfurt, line 32: is clean a new arg?

JFrankfurt: ah, I see it used in the script 👍

lingui.config.ts
```diff
@@ -1,8 +1,60 @@
...
+      try {
+        const cache = readFileSync(cachePath, 'utf8')
+        if (cache === hash) return
+      } catch (e) {
+        // It should not be considered an error if there is no cache file.
+      }
+    }
```
JFrankfurt, line 39: async readFile might read better here

JFrankfurt (approved):

lingui.config.ts
zzmp: deserves a comment, though

lingui.config.ts
zzmp: I think this function still needs to return synchronously, though

lingui.config.ts
zzmp: `pkgUp` is used internally by `lingui`, so I used it here so I could match the code from lingui's own extractors

cmcewen: could you add this as a comment? useful info

lingui.config.ts
```diff
@@ -1,8 +1,63 @@
+import { default as babelExtractor } from '@lingui/cli/api/extractors/babel'
```
cmcewen, line 1: is this something you could upstream to this package?

cmcewen (approved):

lingui.config.ts
zzmp: yes, but it'd need to be a little more fully-featured.
but if we made time for it I'd love to do it.

JFrankfurt (approved):

PR #6496, opened 2023-05-05
feat: [DetailsV2] background for nft details

JackShort:
** creates the background for the new nft details page

Screen Shot 2023-05-05 at 2 45 13 PM
Screen Shot 2023-05-05 at 2 45 22 PM

### DIFF
--- a/src/nft/components/details/detailsV2/NftDetails.tsx
+++ b/src/nft/components/details/detailsV2/NftDetails.tsx
@@ -1,4 +1,5 @@
 import { CollectionInfoForAsset, GenieAsset } from 'nft/types'
+import styled from 'styled-components/macro'
 
 import { DataPage } from './DataPage'
 import { LandingPage } from './LandingPage'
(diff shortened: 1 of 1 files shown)


src/nft/components/details/detailsV2/NftDetails.tsx
```diff
@@ -8,11 +9,29 @@ interface NftDetailsProps {
   collection: CollectionInfoForAsset
 }
 
+const DetailsBackground = styled.div<{ backgroundImage: string }>`
+  position: fixed;
+  top: ${({ theme }) => `-${theme.navHeight}px`};
```
cmcewen, line 14: i feel like negative spacing can lead to unexpected problems - any other way to do this?

src/nft/components/details/detailsV2/NftDetails.tsx
```diff
@@ -8,11 +9,29 @@ interface NftDetailsProps {
...
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background-image: ${({ backgroundImage }) => `url(${backgroundImage})`};
+  filter: blur(100px);
+  opacity: ${({ theme }) => (theme.darkMode ? 0.2 : 0.24)};
+`
+
+const DetailsContentContainer = styled.div`
+  z-index: 2;
```
cmcewen, line 24: we have a Z_INDEX constant defined - could you use that?

JackShort: My understanding is that we use bootstraps z-index rules: https://getbootstrap.com/docs/5.0/layout/z-index/ - their rules say to use 1,2, or 3 for handling z-indexes inside of components. Is that not appropriate to apply here. All of the z-indexes we have in themes are large

src/nft/components/details/detailsV2/NftDetails.tsx
JackShort: forgot that since this is `fixed` it doesn't even need the spacing

src/nft/components/details/detailsV2/NftDetails.tsx
cmcewen: oh i guess you are right. idk what do you think about just still making them constants in there? it says `1, 2, and 3 for default, hover, and active states` so maybe naming them something like that?

cmcewen (approved):

src/nft/components/details/detailsV2/NftDetails.tsx
JackShort: I like that

cbachmeier (approved):
LGTM

PR #6497, opened 2023-05-05
fix: filter AbortError from Sentry

vm:
## Description
It is possible for an AbortError to be sent to Sentry even if it's caused by the user navigating away from the page, because Sentry captures errors and exceptions occurring in your application, including those caused by user actions.

In the case of an AbortError, it typically occurs when a fetch request is aborted, often due to the user navigating away from the page or closing the browser tab. When this happens, the browser cancels the ongoing fetch request, and the JavaScript Promise associated with the request rejects with an AbortError.

image

### Automated testing
- [x] Unit test

### DIFF
--- a/src/tracing/errors.ts
+++ b/src/tracing/errors.ts
@@ -96,6 +96,9 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event: 
     if (error.message.match(/WebAssembly.instantiate\(\): Wasm code generation disallowed by embedder/)) {
       return null
     }
+
+    // These are caused by user navigation away from the page before a request has finished.
+    if (error.message.startsWith('AbortError')) return null
   }
 
   return event
--- a/src/tracing/errors.test.ts
+++ b/src/tracing/errors.test.ts
@@ -201,4 +201,9 @@ describe('filterKnownErrors', () => {
       expect(filterKnownErrors(ERROR, { originalException })).toBeNull()
     })
   })
+
+  it('filters AbortErrors', () => {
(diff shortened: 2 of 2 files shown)


just-toby (approved):

src/tracing/errors.ts
```diff
@@ -96,6 +96,9 @@ export const filterKnownErrors: Required<ClientOptions>['beforeSend'] = (event:
     if (error.message.match(/WebAssembly.instantiate\(\): Wasm code generation disallowed by embedder/)) {
       return null
     }
+
+    // These are caused by user navigation away from the page before a request has finished.
+    if (error.message.startsWith('AbortError')) return null
```
zzmp, line 101: nit: This is a typed error. Can you be more specific and match the type/name?
See https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort#:~:text=Note%3A%20When%20abort()%20is%20called%2C%20the%20fetch()%20promise%20rejects%20with%20an%20Error%20of%20type%20DOMException%2C%20with%20name%20AbortError.

zzmp (approved):

src/tracing/errors.ts
vm: it doesn't seem to be typed? https://uniswap-labs.sentry.io/issues/3788567314/?query=is%3Aunresolved+aborterror&referrer=issue-stream&statsPeriod=14d&stream_index=0

vm: i'll merge but if you think i'm misunderstanding this lmk and i'll happily make a new pr haha

PR #6498, opened 2023-05-05
fix: ensure event.request is defined in trace

vm:
## Description
This may or may not fix https://uniswap-labs.sentry.io/issues/4008758184/?end=2023-05-05T19%3A09%3A59&project=4504255148851200&query=release%3A33c099119da8d87f45de8dd207f2adccc622fe8b&referrer=issue-stream&sort=freq&start=2023-05-04T12%3A10%3A00&stream_index=1

Regardless, better to check for request being defined here imo to make sure it can be destructured.

### QA (ie manual testing)
I called the function in the code with an event without request and it didn't error out.

### Automated testing
- [ ] Unit test
- [ ] Integration/E2E test


src/components/Web3Provider/index.tsx
```diff
@@ -86,7 +86,7 @@ function Updater() {
 }
 
 function trace(event: any) {
-  if (event.action !== 'request') return
```
JFrankfurt, line 89: do we not need the prior check as well? this changes the behavior in addition to adding the new guard

vm: i guess i'm not sure why we had this check before. @zzmp could you explain?

zzmp: This is a debugging method that is flag-guarded, used for logging JSON-RPC requests to the console.
The conditional used in the guard was selected through observation and duck-typing. If this seems to work, then it's fine to change.

zzmp (approved):

JFrankfurt (approved):

PR #6499, opened 2023-05-05
chore: Refactor swap request flow

tinaszheng:
Note to reviewer: Most of these line diffs are just unit test updates 🤔 

## Description
This PR refactors the swap flow to be more like the request code in Widget and contains these improvements:
- Request arguments are now coupled with the trade objects they are transformed into, so no more mismatch between args.currencyIn and trade.currencyIn
- Removes the need for the state `TradeState.SYNCING` because of this ^
- Renames `Trade<Currency, Currency, TradeType>` to just be `InterfaceTrade` 
- Passes `gasUseEstimateUSD` as a string instead of `CurrencyAmount` because the currency amount version is only used in one place - the autoSlippageTolerance, and we were formatting it specially everywhere

## Screen capture

No UI changes!

### QA (ie manual testing)
- Test ETH -> USDC quote request
- Test imported token -> USDC quote request
- Test changing chains 
- Test client side routing 

### Automated testing
- [x] Update unit tests


src/components/Widget/transactions.ts
```diff
@@ -106,6 +106,7 @@ export function useSyncWidgetTransactions() {
 
         const eventProperties = {
           ...formatSwapSignedAnalyticsEventProperties({
+            // @ts-ignore because we're removing this soon
```
tinaszheng, line 109: could wait for the widget removal stuff to go in first

src/hooks/useUSDPrice.ts
```diff
@@ -41,8 +41,8 @@ function useETHValue(currencyAmount?: CurrencyAmount<Currency>): {
     }
   }
 
-  if (!trade || !currencyAmount?.currency || !isGqlSupportedChain(chainId)) {
-    return { data: undefined, isLoading: state === TradeState.LOADING || state === TradeState.SYNCING }
+  if (!trade || state === TradeState.LOADING || !currencyAmount?.currency || !isGqlSupportedChain(chainId)) {
```
tinaszheng, line 44: adding in this case because it wasn't caught before - if the state is loading, then that means the USD price could be displayed for an outdated currency and will give a flash of the wrong value

src/hooks/useUSDPrice.ts
```diff
@@ -41,8 +41,8 @@ function useETHValue(currencyAmount?: CurrencyAmount<Currency>): {
     }
```
zzmp, line 41: Why aren't this function's return values memoized?

src/hooks/useUSDPrice.ts
```diff
@@ -41,8 +41,8 @@ function useETHValue(currencyAmount?: CurrencyAmount<Currency>): {
...
   }
 
-  if (!trade || !currencyAmount?.currency || !isGqlSupportedChain(chainId)) {
-    return { data: undefined, isLoading: state === TradeState.LOADING || state === TradeState.SYNCING }
+  if (!trade || state === TradeState.LOADING || !currencyAmount?.currency || !isGqlSupportedChain(chainId)) {
+    return { data: undefined, isLoading: state === TradeState.LOADING }
```
zzmp, line 45: This should be a file-defined constant so that memoization is not necessary.

src/lib/utils/analytics.ts
```diff
@@ -70,7 +70,7 @@ export const formatSwapQuoteReceivedEventProperties = (
     token_in_address: getTokenAddress(trade.inputAmount.currency),
     token_out_address: getTokenAddress(trade.outputAmount.currency),
     price_impact_basis_points: trade ? formatPercentInBasisPointsNumber(computeRealizedPriceImpact(trade)) : undefined,
-    estimated_network_fee_usd: gasUseEstimateUSD ? formatToDecimal(gasUseEstimateUSD, 2) : undefined,
+    estimated_network_fee_usd: gasUseEstimateUSD ?? undefined,
```
zzmp, line 73: Isn't this already `undefined` - do you need the nullish operator?

src/nft/components/bag/BagFooter.tsx
```diff
@@ -218,7 +218,7 @@ const InputCurrencyValue = ({
     )
   }
 
-  if (tradeState === TradeState.LOADING) {
+  if (tradeState === TradeState.LOADING && !trade) {
```
zzmp, line 221: Will there ever be a valid case where `tradeState === TradeState.LOADING && trade`? If not, this could remain as-is to be simpler/easier-to-read.

src/pages/Swap/index.tsx
```diff
@@ -123,11 +122,11 @@ const DetailsSwapSection = styled(SwapSection)`
 `
 
 function getIsValidSwapQuote(
-  trade: InterfaceTrade<Currency, Currency, TradeType> | undefined,
+  trade: InterfaceTrade | undefined,
   tradeState: TradeState,
   swapInputError?: ReactNode
 ): boolean {
-  return !!swapInputError && !!trade && (tradeState === TradeState.VALID || tradeState === TradeState.SYNCING)
+  return !!swapInputError && !!trade && tradeState === TradeState.VALID
```
zzmp, line 129: Instead of `!!`, what do you think of `Boolean(swapInputError && trade && tradeState === TradeState.VALID)`?

src/pages/Swap/index.tsx
```diff
@@ -302,7 +301,7 @@ export function Swap({
   const fiatValueOutput = useUSDPrice(parsedAmounts[Field.OUTPUT])
 
   const [routeNotFound, routeIsLoading, routeIsSyncing] = useMemo(
-    () => [!trade?.swaps, TradeState.LOADING === tradeState, TradeState.SYNCING === tradeState],
+    () => [!trade?.swaps, TradeState.LOADING === tradeState, TradeState.LOADING === tradeState && Boolean(trade)],
```
zzmp, line 304: nit: everywhere else you check `tradeState === TradeState.XYZ`, here it is flipped.

src/state/index.ts
```diff
@@ -14,7 +14,15 @@ const store = configureStore({
   reducer,
   enhancers: (defaultEnhancers) => defaultEnhancers.concat(sentryEnhancer),
   middleware: (getDefaultMiddleware) =>
-    getDefaultMiddleware({ thunk: true })
+    getDefaultMiddleware({
+      thunk: true,
+      serializableCheck: {
```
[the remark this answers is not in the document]
zzmp: is this actually necessary? are we still getting serializable check errors?

src/state/routing/utils.ts
```diff
@@ -37,88 +55,113 @@ export function computeRoutes(
...
+        outputAmount: CurrencyAmount.fromRawAmount(parsedCurrencyOut, rawAmountOut),
       }
     })
   } catch (e) {
-    // `Route` constructor may throw if inputs/outputs are temporarily out of sync
-    // (RTK-Query always returns the latest data which may not be the right inputs/outputs)
-    // This is not fatal and will fix itself in future render cycles
-    console.error(e)
-    return undefined
+    console.error('computeRoutes error', e)
```
zzmp, line 73:
```suggestion
    console.error('Error computing routes:', e)
```

src/state/routing/utils.ts
```diff
@@ -37,88 +55,113 @@ export function computeRoutes(
         throw new Error('Expected both amountIn and amountOut to be present')
       }
 
-      const routeProtocol = getRouteProtocol(route)
+      const isOnlyV2 = isVersionedRoute<V2PoolInRoute>(PoolType.V2Pool, route)
+      const isOnlyV3 = isVersionedRoute<V3PoolInRoute>(PoolType.V3Pool, route)
```
zzmp, line 59: Because the type is tied to the first parameter, you shouldn't need to pass it:

```suggestion
      const isOnlyV2 = isVersionedRoute(PoolType.V2Pool, route)
      const isOnlyV3 = isVersionedRoute(PoolType.V3Pool, route)
```

src/state/routing/utils.ts
```diff
@@ -37,88 +55,113 @@ export function computeRoutes(
...
+}
+
+export function transformRoutesToTrade(args: GetQuoteArgs, data: GetQuoteResult): TradeResult {
+  const { tokenInAddress, tokenOutAddress, tradeType } = args
+  const tokenInIsNative = Object.values(SwapRouterNativeAssets).includes(tokenInAddress as SwapRouterNativeAssets)
+  const tokenOutIsNative = Object.values(SwapRouterNativeAssets).includes(tokenOutAddress as SwapRouterNativeAssets)
+  const { gasUseEstimateUSD, blockNumber } = data
+  const routes = computeRoutes(tokenInIsNative, tokenOutIsNative, data.route)
+
+  const trade = new InterfaceTrade({
```
zzmp, line 96: Why is this type filtering necessary? Couldn't you just pass in the whole thing if you'd named them routev2, etc from computeRoutes?

src/state/routing/utils.ts
```diff
@@ -37,88 +55,113 @@ export function computeRoutes(
...
         )
-        .map(({ mixedRoute, inputAmount, outputAmount }) => ({ mixedRoute, inputAmount, outputAmount })) ?? [],
+        .map(({ mixedRoute, inputAmount, outputAmount }) => ({
+          mixedRoute,
+          inputAmount,
+          outputAmount,
+        })) ?? [],
     tradeType,
-    gasUseEstimateUSD,
+    gasUseEstimateUSD: parseFloat(gasUseEstimateUSD).toFixed(2).toString(),
```
zzmp, line 129: I think this should be formatting it for display (eg if it's <0.01, return `<$0.01`; include the `$`).

src/state/routing/utils.ts
```diff
@@ -37,88 +55,113 @@ export function computeRoutes(
...
   new Pair(
     CurrencyAmount.fromRawAmount(parseToken(reserve0.token), reserve0.quotient),
     CurrencyAmount.fromRawAmount(parseToken(reserve1.token), reserve1.quotient)
   )
 
-const genericPoolPairParser = (pool: V3PoolInRoute | V2PoolInRoute): Pool | Pair => {
-  return pool.type === 'v3-pool' ? parsePool(pool) : parsePair(pool)
+export function isExactInput(tradeType: TradeType): boolean {
+  return tradeType === TradeType.EXACT_INPUT
 }
```
zzmp, line 159: As a follow-up, this could be moved to a utility and used _everywhere_.

src/state/routing/useRoutingAPITrade.ts
```diff
@@ -44,83 +46,34 @@ export function useRoutingAPITrade<TTradeType extends TradeType>(
...
-    }
-
-    if (isLoading && !quoteResult) {
-      // only on first hook render
+    if (!amountSpecified || isError || !queryArgs) {
+      return TRADE_INVALID
+    } else if (tradeResult?.state === QuoteState.NOT_FOUND && isCurrent) {
+      return TRADE_NOT_FOUND
+    } else if (!tradeResult?.trade) {
+      // TODO: use `isLoading` returned by rtk-query hook instead of checking for `trade` status
```
zzmp, line 67: This should be ticketed.

src/state/routing/useRoutingAPITrade.ts
```diff
@@ -44,83 +46,34 @@ export function useRoutingAPITrade<TTradeType extends TradeType>(
...
+      return TRADE_LOADING
+    } else {
+      const isCurrent = currentTradeResult === tradeResult
```
zzmp, line 70: This shadows `isCurrent` on L59, I think it can be removed.

src/state/routing/types.ts
```diff
@@ -68,20 +67,20 @@ export interface GetQuoteResult {
   routeString: string
 }
 
-export class InterfaceTrade<
-  TInput extends Currency,
-  TOutput extends Currency,
-  TTradeType extends TradeType
-> extends Trade<TInput, TOutput, TTradeType> {
-  gasUseEstimateUSD: CurrencyAmount<Token> | null | undefined
+class ClassicTrade<TInput extends Currency, TOutput extends Currency, TTradeType extends TradeType> extends Trade<
```
zzmp, line 70: Is this used anywhere? Is there any reason it can't be called `InterfaceTrade`, and extend a concrete type of `Trade`, instead of re-aliasing it just underneath?

src/state/routing/types.ts
```diff
@@ -105,3 +104,43 @@ export class InterfaceTrade<
...
+      trade: InterfaceTrade
+    }
+
+export enum PoolType {
+  V2Pool = 'v2-pool',
+  V3Pool = 'v3-pool',
+}
+
+// swap router API special cases these strings to represent native currencies
+// all chains have "ETH" as native currency symbol except for polygon
```
zzmp, line 142: Is this due to a recent swap router upgrade, or is native polygon trading broken in its current state?

src/lib/hooks/routing/clientSideSmartOrderRouter.ts
```diff
@@ -19,50 +21,44 @@ export function isSupportedChainId(chainId: ChainId | undefined): boolean {
...
     amount: BigintIsh
   },
   router: AlphaRouter,
   config: Partial<AlphaRouterConfig>
-): Promise<{ data: GetQuoteResult; error?: unknown }> {
-  const currencyIn = new Token(tokenIn.chainId, tokenIn.address, tokenIn.decimals, tokenIn.symbol)
-  const currencyOut = new Token(tokenOut.chainId, tokenOut.address, tokenOut.decimals, tokenOut.symbol)
+): Promise<QuoteResult> {
+  if (!amountRaw) {
+    return { state: QuoteState.INITIALIZED }
```
zzmp, line 38: This will not actually initialize the client-side smart order router! You _must_ call `router.route` with dummy values to do so. Otherwise, the first call with actual value will take significantly longer. This is why widgets defaults to using a value of `1` and then returns after calling `router.route`.

Another option (to improve on widgets) is to call `router.route`, and return without awaiting, although you risk not actually initializing in this case, too.

src/lib/hooks/routing/clientSideSmartOrderRouter.ts
```diff
@@ -19,50 +21,44 @@ export function isSupportedChainId(chainId: ChainId | undefined): boolean {
...
     amount: amountRaw,
   }: {
-    type: 'exactIn' | 'exactOut'
+    tradeType: TradeType
     tokenIn: { address: string; chainId: number; decimals: number; symbol?: string }
     tokenOut: { address: string; chainId: number; decimals: number; symbol?: string }
     amount: BigintIsh
   },
   router: AlphaRouter,
   config: Partial<AlphaRouterConfig>
```
zzmp, line 35: nit: rename this to `routerConfig` to disambiguate with `swapConfig`.

src/state/routing/slice.ts
```diff
@@ -85,7 +93,7 @@ export const routingApi = createApi({
     baseUrl: 'https://api.uniswap.org/v1/',
   }),
   endpoints: (build) => ({
-    getQuote: build.query<GetQuoteResult, GetQuoteArgs>({
+    getQuote: build.query<TradeResult, GetQuoteArgs>({
```
zzmp, line 96: Why no `SkipToken`? I think (from widgets code) that you may also need to check for the skip token and return early, but I am not sure. See https://github.com/Uniswap/widgets/blob/main/src/state/routing/slice.ts#L31.

```suggestion
    getQuote: build.query<TradeResult, GetQuoteArgs | SkipToken>({
```

src/state/routing/slice.ts
```diff
@@ -114,12 +122,11 @@ export const routingApi = createApi({
           }
         )
       },
-      async queryFn(args, _api, _extraOptions, fetch) {
-        const { tokenInAddress, tokenInChainId, tokenOutAddress, tokenOutChainId, amount, routerPreference, type } =
-          args
-
-        try {
-          if (routerPreference === RouterPreference.API) {
+      async queryFn(args: GetQuoteArgs, _api, _extraOptions, fetch) {
```
zzmp, line 125:
```suggestion
      async queryFn(args: GetQuoteArgs | SkipToken, _api, _extraOptions, fetch) {
```

Do you need to handle the skip token? I'm not sure if this is baked in, but widgets does so. See https://github.com/Uniswap/widgets/blob/main/src/state/routing/slice.ts#L58.

src/state/routing/slice.ts
```diff
@@ -129,21 +136,41 @@ export const routingApi = createApi({
...
+                // NO_ROUTE should be treated as a valid response to prevent retries.
+                if (typeof errorData === 'object' && errorData?.errorCode === 'NO_ROUTE') {
+                  return { data: { state: QuoteState.NOT_FOUND } }
+                }
+              } catch {
+                throw response.error
+              }
+            }
+
+            const quoteData = response.data as GetQuoteResult
```
zzmp, line 153: I'm surprised this isn't already typed.

zzmp, line 153:
```suggestion
            const getQuoteResult = response.data as GetQuoteResult
```

src/state/routing/types.ts
```diff
@@ -68,20 +67,20 @@ export interface GetQuoteResult {
   routeString: string
```
zzmp, line 67: nit: rename `GetQuoteResult` to simpler `QuoteData`.

zzmp (changes requested):

src/lib/hooks/routing/clientSideSmartOrderRouter.ts
zzmp: either way this is done, you should add a test guaranteeing that we actually initialize the SOR instead of returning early, so this isn't broken in the future.

src/state/index.ts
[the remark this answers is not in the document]
tinaszheng: yes

src/components/swap/GasEstimateBadge.tsx
```diff
@@ -37,15 +36,15 @@ export default function GasEstimateBadge({
...
-  trade: InterfaceTrade<Currency, Currency, TradeType> | undefined | null // dollar amount in active chain's stablecoin
+  trade: InterfaceTrade | undefined | null // dollar amount in active chain's stablecoin
   loading: boolean
   showRoute?: boolean // show route instead of gas estimation summary
   disableHover?: boolean
 }) {
   const formattedGasPriceString = trade?.gasUseEstimateUSD
-    ? trade.gasUseEstimateUSD.toFixed(2) === '0.00'
+    ? trade.gasUseEstimateUSD === '0.00'
       ? '<$0.01'
```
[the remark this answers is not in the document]
tinaszheng: i would also love to do that, but we do actually parse this value in one place - in `useAutoSlippageTolerance` and it converts it into a CurrencyAmount. so keeping this logic for now

src/hooks/useAutoSlippageTolerance.ts
```diff
@@ -72,15 +72,14 @@ const MAX_AUTO_SLIPPAGE_TOLERANCE = new Percent(5, 100) // 5%
...
-  trade: InterfaceTrade<Currency, Currency, TradeType> | undefined
-): Percent {
+export default function useAutoSlippageTolerance(trade: InterfaceTrade | undefined): Percent {
   const { chainId } = useWeb3React()
   const onL2 = chainId && L2_CHAIN_IDS.includes(chainId)
   const outputDollarValue = useStablecoinValue(trade?.outputAmount)
   const nativeGasPrice = useGasPrice()
 
   const gasEstimate = guesstimateGas(trade)
+  const gasEstimateUSD = useStablecoinAmountFromFiatValue(trade?.gasUseEstimateUSD) ?? null
```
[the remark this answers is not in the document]
tinaszheng: one is a string and one is a `CurrencyAmount` - could use a better variable name

src/hooks/useUSDPrice.ts
tinaszheng: i feel like these types of arguments should be the other way around since the default is _not_ memoized - like there needs to be a reason for memoizing vs the other. in this case the consumers destructure the props, so memoizing is unnecessary for maintaining referential stability

src/components/swap/GasEstimateBadge.tsx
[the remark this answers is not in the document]
zzmp: rather than include parsing logic here, can this be added as a class method on `ClassicTrade` to keep the logic in one place? eg

```
  gasUseEstimate: CurrencyAmount<X, Y>
  gasUseEstimateDisplay: string
```

src/lib/utils/analytics.ts
tinaszheng: nice catch!

src/state/routing/utils.ts
tinaszheng: we can take a look at this logic more closely and move to sdk! i'd rather not make any changes for now, because this code is consistent with both widgets and mobile right now

src/state/routing/types.ts
tinaszheng: it is now used in one place!

src/state/routing/types.ts
tinaszheng: its because native polygon is MATIC since they are a JOKE OF A CHAIN

src/lib/hooks/routing/clientSideSmartOrderRouter.ts
tinaszheng: switched the ordering back to original and also left a comment

src/state/routing/slice.ts
tinaszheng: you don't need to explicitly handle that - it's baked into rtk-query!

src/state/routing/slice.ts
tinaszheng: quoteData is more accurate in my mind

src/state/routing/types.ts
tinaszheng: it's not broken because we always pass in `.wrapped` into the quote request, which effectively gives you the same result for native tokens (because on-chain they need to be wrapped anyway). but for [redacted] it wouldn't work simialrly

src/state/routing/utils.ts
tinaszheng: it unfortunately gives me another type error if i do that - will leave it alone for now

src/components/swap/SwapModalFooter.tsx
```diff
@@ -24,7 +24,7 @@ import { SwapCallbackError } from './styleds'
 import { getTokenPath, RoutingDiagramEntry } from './SwapRoute'
 
 interface AnalyticsEventProps {
-  trade: InterfaceTrade<Currency, Currency, TradeType>
+  trade: InterfaceTrade
   hash: string | undefined
```
grabbou, line 28: nit: not related to this PR, but I wonder why this isn't just `hash?: string`

src/components/swap/SwapModalFooter.tsx
```diff
@@ -74,7 +74,7 @@ const formatAnalyticsEventProperties = ({
   fiatValueInput,
   fiatValueOutput,
 }: AnalyticsEventProps) => ({
-  estimated_network_fee_usd: trade.gasUseEstimateUSD ? formatToDecimal(trade.gasUseEstimateUSD, 2) : undefined,
+  estimated_network_fee_usd: trade.gasUseEstimateUSD ?? undefined,
```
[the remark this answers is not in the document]
grabbou: To add to this, maybe worth annotating with a comment explaining the use of `??` and ideally ,a link where one can read further about when `gasUseEstimateUSD` is null. I am still quite confused why we have so many `null` and `undefined`s in a lot of places.

src/hooks/useAutoSlippageTolerance.ts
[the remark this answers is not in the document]
grabbou: I think it might not be necessary/worth it to expose both if it's only used here for now.

src/hooks/useAutoSlippageTolerance.ts
```diff
@@ -72,15 +72,14 @@ const MAX_AUTO_SLIPPAGE_TOLERANCE = new Percent(5, 100) // 5%
 /**
  * Returns slippage tolerance based on values from current trade, gas estimates from api, and active network.
  */
-export default function useAutoSlippageTolerance(
-  trade: InterfaceTrade<Currency, Currency, TradeType> | undefined
-): Percent {
+export default function useAutoSlippageTolerance(trade: InterfaceTrade | undefined): Percent {
```
grabbou, line 75: nit: any preference in particular over `trade?: InterfaceTrade`

src/hooks/useAutoSlippageTolerance.ts
```diff
@@ -100,9 +99,7 @@ export default function useAutoSlippageTolerance(
     // NOTE - dont use gas estimate for L2s yet - need to verify accuracy
     // if not, use local heuristic
     const dollarCostToUse =
-      chainId && SUPPORTED_GAS_ESTIMATE_CHAIN_IDS.includes(chainId) && trade?.gasUseEstimateUSD
-        ? trade.gasUseEstimateUSD
-        : dollarGasCost
+      chainId && SUPPORTED_GAS_ESTIMATE_CHAIN_IDS.includes(chainId) && gasEstimateUSD ? gasEstimateUSD : dollarGasCost
```
grabbou, line 102: This is one thing that makes me often confused. If the trade is calculated against a particular chainId (say I am swapping on Polygon), then I would expect `gasEstimateUSD` to be null if the chain does not support it. 

My point is, I find `chainId && SUPPORTED_GAS_ESTIMATE_CHAIN_IDS.includes(chainId)` unnecessary and feel like it would be great to do where we create `trade` object. That way, we could remove A LOT of conditionals from the codebase, I have seen that one around in particular.

src/lib/hooks/routing/clientSideSmartOrderRouter.ts
```diff
@@ -19,50 +21,45 @@ export function isSupportedChainId(chainId: ChainId | undefined): boolean {
...
+    : new Token(tokenOut.chainId, tokenOut.address, tokenOut.decimals, tokenOut.symbol)
 
-  const swapRoute = await router.route(
-    amount,
-    quoteCurrency,
-    type === 'exactIn' ? TradeType.EXACT_INPUT : TradeType.EXACT_OUTPUT,
-    /*swapConfig=*/ undefined,
-    config
-  )
+  const baseCurrency = tradeType === TradeType.EXACT_INPUT ? currencyIn : currencyOut
```
grabbou, line 47: What does `TradeType.EXACT_INPUT` mean? It's interesting that in/out values are just flipped. I am confused, because I don't see any correlation with `exact`.

src/lib/hooks/routing/clientSideSmartOrderRouter.ts
```diff
@@ -19,50 +21,45 @@ export function isSupportedChainId(chainId: ChainId | undefined): boolean {
...
+  const quoteCurrency = tradeType === TradeType.EXACT_INPUT ? currencyOut : currencyIn
 
-  if (!swapRoute) throw new Error('Failed to generate client side quote')
+  const amount = CurrencyAmount.fromRawAmount(baseCurrency, JSBI.BigInt(amountRaw))
+  const swapRoute = await router.route(amount, quoteCurrency, tradeType, /*swapConfig=*/ undefined, routerConfig)
 
-  return { data: transformSwapRouteToGetQuoteResult(type, amount, swapRoute) }
-}
+  // this check is intentionally left delayed until after the `router.route` call so that on empty input, the
+  // client side routing will begin initialization
```
grabbou, line 54: I think I need a better explanation here - what does it mean that "client side routing will begin initialization"? AFAIK we're not doing any work here, we're just returning state. What kind of "initialisation" do you mean?

src/nft/components/bag/BagFooter.tsx
grabbou: So, for "syncing" state, we're not going to have any UI response? BTW, would be great to add a comment

src/pages/Swap/index.tsx
```diff
@@ -123,11 +122,11 @@ const DetailsSwapSection = styled(SwapSection)`
 `
 
 function getIsValidSwapQuote(
-  trade: InterfaceTrade<Currency, Currency, TradeType> | undefined,
+  trade: InterfaceTrade | undefined,
   tradeState: TradeState,
   swapInputError?: ReactNode
 ): boolean {
-  return !!swapInputError && !!trade && (tradeState === TradeState.VALID || tradeState === TradeState.SYNCING)
+  return Boolean(swapInputError && trade && tradeState === TradeState.VALID)
```
grabbou, line 129: This is confusing (and it was before too) - "isValidSwapQuote" will return false when swapInputError is null, right? Is this code correct?

src/pages/Swap/index.tsx
```diff
@@ -303,7 +302,7 @@ export function Swap({
   const fiatValueOutput = useUSDPrice(parsedAmounts[Field.OUTPUT])
 
   const [routeNotFound, routeIsLoading, routeIsSyncing] = useMemo(
```
grabbou, line 304: Do we need this to be `useMemo`?

grabbou: Can we try removing `routeIsSyncing` entirely? There are some conditionals in this component that are super confusing and I think not necessary.

For example here, I think we can easily just replace with `tradeState === TradeState.LOADING`
Screenshot 2023-05-11 at 18 36 34
Screenshot 2023-05-11 at 18 36 36
Screenshot 2023-05-11 at 18 37 44

here - we have `trade` and that prop should be enough. We don't even need to pass `loading` and `syncing` because you can easily read that off `trade.state`
Screenshot 2023-05-11 at 18 38 03

grabbou (commented):
Great work! I am 70% through the PR, going to submit what I have right now and will return back!

src/state/routing/slice.ts
```diff
@@ -133,21 +140,41 @@ export const routingApi = createApi({
...
-              // TODO(zzmp): Use PRICE_PARAMS for RouterPreference.PRICE.
-              // This change is intentionally being deferred to first see what effect router caching has.
-              CLIENT_PARAMS
+            const response = await fetch(`quote?${query}`)
+            if (response.error) {
+              try {
+                // cast as any here because we do a runtime check on it being an object before indexing into .errorCode
+                const errorData = response.error.data as any
+                // NO_ROUTE should be treated as a valid response to prevent retries.
+                if (typeof errorData === 'object' && errorData?.errorCode === 'NO_ROUTE') {
```
grabbou, line 149: nit: Do we have any other types of errors? Are we expecting to have more in the future? If yes to any of these, maybe good opportunity for enum?

src/state/routing/types.ts
```diff
@@ -68,20 +67,20 @@ export interface GetQuoteResult {
...
 }
 
-export class InterfaceTrade<
+export class ClassicTrade<
   TInput extends Currency,
```
grabbou, line 71: nit: Can someone create another type based on ClassicTrade and supplement different values here? Do we do that or plan to do it in the future?

If not, I would vote to make it a simple ClassicTrace class that extends `Trade<Currency, Currency, TradeType>`

src/state/routing/types.ts
```diff
@@ -68,20 +67,20 @@ export interface GetQuoteResult {
...
   constructor({
     gasUseEstimateUSD,
     blockNumber,
     ...routes
   }: {
-    gasUseEstimateUSD?: CurrencyAmount<Token> | undefined | null
+    gasUseEstimateUSD?: string | undefined | null
     blockNumber?: string | null | undefined
     v2Routes: {
       routev2: V2Route<TInput, TOutput>
```
grabbou, line 86: nit: I don't think we should re-define these types here. They are already defined by the `Trade` type - that's why it's generic and accepts those three parameters. Instead, we should have a type like this:
```
{ gasUseEstimateUSD?: string | null, blockNumber?: string | null } & TradeConstructorArgs }
```
that we can extract with something like `ConstructorParameters` -> https://www.typescriptlang.org/docs/handbook/utility-types.html#constructorparameterstype

src/state/routing/types.ts
```diff
@@ -105,3 +104,43 @@ export class InterfaceTrade<
...
+export enum QuoteState {
+  SUCCESS = 'Success',
+  INITIALIZED = 'Initialized',
+  NOT_FOUND = 'Not found',
+}
+
+export type QuoteResult =
+  | {
+      state: QuoteState.INITIALIZED | QuoteState.NOT_FOUND
+      data?: undefined
```
grabbou, line 119: nit: `data: undefined`

src/state/routing/types.ts
```diff
@@ -105,3 +104,43 @@ export class InterfaceTrade<
...
+    }
+
+export enum PoolType {
+  V2Pool = 'v2-pool',
+  V3Pool = 'v3-pool',
+}
+
+// swap router API special cases these strings to represent native currencies
+// all chains have "ETH" as native currency symbol except for polygon
+export enum SwapRouterNativeAssets {
```
grabbou, line 143: nit: Why is this named `NativeAssets`? According to the comment, it's something like native token for a given chain.

src/state/routing/types.ts
```diff
@@ -105,3 +104,43 @@ export class InterfaceTrade<
...
+  MATIC = 'MATIC',
```
grabbou, line 144: nit: Can we unify this with `src/nft/utils/fetchPrice.ts`? Apparently NFT team has similar enum called `Currency` for that particular purpose. I wonder if we have more occurrences like this.

src/state/routing/useRoutingAPITrade.ts
```diff
@@ -3,14 +3,16 @@ import { Currency, CurrencyAmount, TradeType } from '@uniswap/sdk-core'
...
 import { useMemo } from 'react'
 import { RouterPreference, useGetQuoteQuery } from 'state/routing/slice'
 
-import { InterfaceTrade, TradeState } from './types'
-import { computeRoutes, transformRoutesToTrade } from './utils'
+import { InterfaceTrade, QuoteState, TradeState } from './types'
+
+const TRADE_INVALID = { state: TradeState.INVALID, trade: undefined }
+const TRADE_NOT_FOUND = { state: TradeState.NO_ROUTE_FOUND, trade: undefined }
+const TRADE_LOADING = { state: TradeState.LOADING, trade: undefined }
```
grabbou, line 15: nit: Since we're returning those objects to the outside world, it might be good to add `as const` to each declaration to ensure they're truly read-only. Otherwise, we risk (very unlikely, but still) interference across different calls. It's theoretical, but still.

src/state/routing/useRoutingAPITrade.ts
```diff
@@ -25,7 +27,7 @@ export function useRoutingAPITrade<TTradeType extends TradeType>(
   routerPreference: RouterPreference
 ): {
   state: TradeState
-  trade: InterfaceTrade<Currency, Currency, TTradeType> | undefined
+  trade: InterfaceTrade | undefined
```
grabbou, line 30: nit: `trade?: InterfaceTrade` personally cleaner and could be used throughout the rest of the PR, but I leave the convention choice up to you!

src/state/routing/useRoutingAPITrade.ts
```diff
@@ -44,83 +46,33 @@ export function useRoutingAPITrade<TTradeType extends TradeType>(
   })
 
   const {
-    isLoading,
     isError,
-    data: quoteResult,
-    currentData,
+    data: tradeResult,
+    currentData: currentTradeResult,
```
grabbou, line 51: Question: how is `data` different from `currentData`? Could we have a comment there possibly?

grabbou: I see we're comparing them to know whether things are syncing or not. Are these coming from different APIs or sources?

src/state/routing/utils.ts
```diff
@@ -1,32 +1,50 @@
...
-  tradeType: TradeType,
-  quoteResult: Pick<GetQuoteResult, 'route'> | undefined
-) {
-  if (!quoteResult || !quoteResult.route || !currencyIn || !currencyOut) return undefined
+  tokenInIsNative: boolean,
+  tokenOutIsNative: boolean,
+  routes: QuoteData['route']
+):
+  | {
+      routev3: V3Route<Currency, Currency> | null
```
grabbou, line 30: nit: I think you could possibly infer those types / read them from Trade type, like suggested in the other comment. That way, don't have to redefine `routev3` `routev2` and `mixedRoute` types again. Makes it easier if we change the types we pass to generics.

grabbou (commented):
Another pass.

zzmp (commented):
Please re-request once this has had `main` merged in (and please merge, don't rebase).

src/state/routing/utils.ts
tinaszheng: the `tradeType === TradeType.EXACT_INPUT` line is in a couple of other places in the codebase - will add a TODO to convert them to using the util

src/components/swap/AdvancedSwapDetails.tsx
```diff
@@ -21,7 +21,7 @@ const StyledCard = styled(Card)`
 `
 
 interface AdvancedSwapDetailsProps {
-  trade?: InterfaceTrade<Currency, Currency, TradeType>
+  trade?: InterfaceTrade
```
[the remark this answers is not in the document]
tinaszheng: talked about this over slack:)

src/lib/hooks/routing/clientSideSmartOrderRouter.ts
tinaszheng: @zzmp can you help me with wording here?

src/pages/Swap/index.tsx
tinaszheng: ill make a ticket to do this in a follow-up PR bcuz this one is getting bloated already 😓

src/state/routing/slice.ts
```diff
@@ -133,21 +140,41 @@ export const routingApi = createApi({
...
-              args,
-              router,
-              // TODO(zzmp): Use PRICE_PARAMS for RouterPreference.PRICE.
-              // This change is intentionally being deferred to first see what effect router caching has.
-              CLIENT_PARAMS
+            const response = await fetch(`quote?${query}`)
+            if (response.error) {
+              try {
+                // cast as any here because we do a runtime check on it being an object before indexing into .errorCode
+                const errorData = response.error.data as any
```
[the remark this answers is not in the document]
tinaszheng: image
this ends up being quite messy 🤔

src/state/routing/types.ts
tinaszheng: this one you actually need to have `data?: undefined` here or else it will explicitly require that consumers pass in `{ state: whatever, data: undefined }`

src/state/routing/types.ts
tinaszheng: that one is used for Coinbase API though - i would prefer to keep separate

src/state/routing/useRoutingAPITrade.ts
tinaszheng: it's a thing from rtk-query - `data` is the cached result from the previous query run, and while refetching, `data` and `currentData` can be different

https://redux-toolkit.js.org/rtk-query/usage/queries#frequently-used-query-hook-return-values

src/hooks/useAutoSlippageTolerance.ts
zzmp: I agree - why can't this check just be `trade?.gasEstimateUSD ? gasEstimateUSD : dollarGasCost`

src/pages/Swap/index.tsx
zzmp: When I read this, I want to read "when the tradeState is LOADING". Everywhere else, you check `tradeState === TradeState.LOADING`, which follows my mental model. However, here, you check `TradeState.LOADING === tradeState`: "when LOADING is the tradeState".

In general, I like my conditionals to be "when VARIABLE is CONSTANT", and not the other way around. Can I lint for this?

src/state/routing/slice.ts
```diff
@@ -119,11 +127,10 @@ export const routingApi = createApi({
         )
       },
       async queryFn(args, _api, _extraOptions, fetch) {
-        const { tokenInAddress, tokenInChainId, tokenOutAddress, tokenOutChainId, amount, routerPreference, type } =
-          args
-
-        try {
-          if (routerPreference === RouterPreference.API || routerPreference === RouterPreference.AUTO) {
+        if (args.routerPreference === RouterPreference.API || args.routerPreference === RouterPreference.AUTO) {
```
zzmp, line 130: What about `INTERNAL_ROUTER_PREFERENCE_PRICE`?
What codepath is that meant to use?

src/state/routing/slice.ts
```diff
@@ -133,21 +140,40 @@ export const routingApi = createApi({
...
+          const router = getRouter(args.tokenInChainId)
+          const quoteResult = await getClientSideQuote(args, router, CLIENT_PARAMS)
+          if (quoteResult.state === QuoteState.SUCCESS) {
+            return { data: transformRoutesToTrade(args, quoteResult.data) }
+          } else {
+            return { data: quoteResult }
+          }
+        } catch (error: any) {
+          console.warn(`GetQuote failed on client: ${error}`)
+          return { error: { status: 'CUSTOM_ERROR', error: error?.detail ?? error?.message ?? error } }
```
zzmp, line 176: Why did you switch detail and message?

src/lib/hooks/routing/clientSideSmartOrderRouter.ts
zzmp: It means that the user has specified an exact amount for their input, and output is variable.

src/lib/hooks/routing/clientSideSmartOrderRouter.ts
```diff
@@ -19,50 +21,45 @@ export function isSupportedChainId(chainId: ChainId | undefined): boolean {
...
-    quoteCurrency,
-    type === 'exactIn' ? TradeType.EXACT_INPUT : TradeType.EXACT_OUTPUT,
-    /*swapConfig=*/ undefined,
-    config
-  )
+  const baseCurrency = tradeType === TradeType.EXACT_INPUT ? currencyIn : currencyOut
+  const quoteCurrency = tradeType === TradeType.EXACT_INPUT ? currencyOut : currencyIn
 
-  if (!swapRoute) throw new Error('Failed to generate client side quote')
+  const amount = CurrencyAmount.fromRawAmount(baseCurrency, JSBI.BigInt(amountRaw))
```
zzmp, line 50: Do you need to initialize here with a default, in case `amountRaw` is undefined? `...JSBI.BigInt(amountRaw ?? 1))`

src/lib/hooks/routing/clientSideSmartOrderRouter.ts
zzmp: What if you simplify this by making it explicit in code.

```
const isInitialization = !amountRaw
if (isInitialization) {
  // Initialization should use a dummy amount so that the router can initialize.
  // The router makes on-chain calls to initialize a route. Calling `router.route` to initialize
  // significantly improves the lag for the first *actual* route.
  const amount = CurrencyAmount.fromRawAmount(baseCurrency, JSBI.BigInt(1))
  router.route(amount, quoteCurrency, tradeType, /*swapConfig=*/ undefined, routerConfig)
  return { state: QuoteState.INITIALIZED }
}
```

zzmp (changes requested):

src/hooks/useAutoSlippageTolerance.ts
tinaszheng: this feels outside of the scope of this PR

src/state/routing/slice.ts
tinaszheng: mike's pr comment above :P

tinaszheng:
@zzmp opted to remove client side router initialization for now - made a TODO (web-2055) to see if it's worth adding as a followup

src/state/routing/utils.ts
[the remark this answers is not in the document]
tinaszheng: i haven't really touched this code, just moved it around in this file, so this change feels out of scope for this PR 🤔

zzmp (changes requested):
once the `amountRaw === null` case is handled I think this is ready to merge

zzmp (approved):

src/lib/hooks/routing/clientSideSmartOrderRouter.ts
```diff
@@ -48,13 +48,9 @@ async function getQuote(
   const quoteCurrency = tradeType === TradeType.EXACT_INPUT ? currencyOut : currencyIn
 
   const amount = CurrencyAmount.fromRawAmount(baseCurrency, JSBI.BigInt(amountRaw))
+  // TODO (WEB-2055): explore initializing client side routing on first load (when amountRaw is null) if there are enough users using client-side router preference.
   const swapRoute = await router.route(amount, quoteCurrency, tradeType, /*swapConfig=*/ undefined, routerConfig)
 
-  // this check is intentionally left delayed until after the `router.route` call so that on empty input, the
-  // client side routing will begin initialization
-  if (!amountRaw) {
-    return { state: QuoteState.INITIALIZED }
```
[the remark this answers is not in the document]
tinaszheng: yes you're right!

grabbou (approved):

PR #6504, opened 2023-05-08
feat: add tokenlist validation

matteenm:
## Description
moving tokenlist validation code from widgets repo to interface. 
referenced [widgets](https://github.com/Uniswap/widgets) package closely, most code is copied over directly. 
specific PRs referenced:
https://github.com/Uniswap/widgets/pull/273/files

_JIRA ticket:_
_Slack thread:_
_Relevant docs:_

## Screen capture

### QA (ie manual testing)

- searched for several tokens in swap selector to ensure current lists are still being loaded properly. 

### Automated testing

- [x] Unit test
- [x] Integration/E2E test


package.json
```diff
@@ -158,7 +159,7 @@
     "@uniswap/router-sdk": "^1.3.0",
     "@uniswap/sdk-core": "^3.2.2",
     "@uniswap/smart-order-router": "^3.6.1",
-    "@uniswap/token-lists": "^1.0.0-beta.30",
+    "@uniswap/token-lists": "^1.0.0-beta.31",
```
tinaszheng, line 162: does this require an upgrade in the same PR or is it unrelated

matteenm: on second thought I can separate this change out into another PR. this PR can just be moving the same logic over. ill followup with a PR that upgrade the version to address the CG list issue.

.eslintrc.js
```diff
@@ -4,6 +4,7 @@ require('@uniswap/eslint-config/load')
 
 module.exports = {
   extends: '@uniswap/eslint-config/react',
+  ignorePatterns: ['src/utils/__generated__/'],
```
zzmp, line 7: nit: big oof that generated files are linted at all. If you're up for it, I think that `@uniswap/eslint-config` should ignore any generated file: `**/__generated__/`. If you're not up for it, could you ping me (in Slack) and I can own or ticket it for the near future?

.gitignore
```diff
@@ -12,6 +12,10 @@
 /src/graphql/**/__generated__
 schema.graphql
 
+# generated ajv schema
+/src/utils/__generated__/*
+!/src/utils/__generated__/.gitkeep
```
zzmp, line 17: nit: You can update this to exclude all generated files, which is desired, if you'd like:

```
# generated files
/src/**/__generated__

# schema
schema.graphql
```

Separately - why is the `.gitkeep` required here?

scripts/compile-ajv-validators.js
```diff
@@ -0,0 +1,20 @@
...
+addFormats(ajv)
+const validate = ajv.compile(schema)
+let moduleCode = standaloneCode(ajv, validate)
+fs.writeFileSync(path.join(__dirname, '../src/utils/__generated__/validateTokenList.js'), moduleCode)
+
+const ajv2 = new Ajv({ code: { source: true, esm: true } })
+addFormats(ajv2)
+const validate2 = ajv2.compile(Object.assign({}, schema, { required: ['tokens']}))
+let moduleCode2 = standaloneCode(ajv2, validate2)
+fs.writeFileSync(path.join(__dirname, '../src/utils/__generated__/validateTokens.js'), moduleCode2)
```
zzmp, line 20: nit: newline at end of file (I don't know why this isn't linted :/)

scripts/compile-ajv-validators.js
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable */
```
zzmp, line 1: Don't disable eslint. In this case, you probably want node:

```suggestion
/* eslint-env node */
```

src/utils/validateTokenList.ts
```diff
@@ -0,0 +1,61 @@
...
+  let validatorImport
+  switch (schema) {
+    case ValidationSchema.LIST:
+      validatorImport = import('utils/__generated__/validateTokenList')
+      break
+    case ValidationSchema.TOKENS:
+      validatorImport = import('utils/__generated__/validateTokens')
+      break
+    default:
+      throw new Error('No validation function specified for schema')
```
zzmp, line 26:
```suggestion
      throw new Error('No validation function specified for token list schema')
```

src/utils/validateTokenList.ts
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Validates an array of tokens.
+ * @param json the TokenInfo[] to validate
+ */
+export async function validateTokens(json: TokenInfo[]): Promise<TokenInfo[]> {
+  try {
+    await validate(ValidationSchema.TOKENS, { tokens: json })
+    return json
+  } catch (err) {
```
zzmp, line 45: nit: call the caught errors `error` unless there is a shadowing issue (here and below).

scripts/compile-ajv-validators.js
```diff
@@ -0,0 +1,20 @@
...
+const path = require('path')
+const Ajv = require('ajv')
+const standaloneCode = require('ajv/dist/standalone').default
+const addFormats = require('ajv-formats')
+const schema = require('@uniswap/token-lists/dist/tokenlist.schema.json')
+
+const ajv = new Ajv({ code: { source: true, esm: true } })
+addFormats(ajv)
+const validate = ajv.compile(schema)
+let moduleCode = standaloneCode(ajv, validate)
```
zzmp, line 13: I'd rather these be renamed to disambiguate, instead of tacking `2` onto the latter:
- `validateTokenList`, `validateTokenListCode` or `tokenListValidationCode`
- `validateTokens`, `validateTokensCode` or `tokensValidationCode`

scripts/compile-ajv-validators.js
```diff
@@ -0,0 +1,20 @@
...
+fs.writeFileSync(path.join(__dirname, '../src/utils/__generated__/validateTokenList.js'), moduleCode)
+
+const ajv2 = new Ajv({ code: { source: true, esm: true } })
+addFormats(ajv2)
+const validate2 = ajv2.compile(Object.assign({}, schema, { required: ['tokens']}))
```
zzmp, line 18: Why did you switch from a spread operator to `Object.assign` (which I think broadens typings, so is less strict)?

zzmp (changes requested):

zzmp (changes requested):
Run `yarn` and commit the lockfile changes to resolve your CI issues with `yarn prepare`. Although you've added `ajv` to the `package.json`, CI will only install what is in the lockfile.

matteenm:
> Run `yarn` and commit the lockfile changes to resolve your CI issues with `yarn prepare`. Although you've added `ajv` to the `package.json`, CI will only install what is in the lockfile.

yarn doesn't generate changes to the lockfile unfortunately :/

scripts/compile-ajv-validators.js
matteenm: temp workaround to linter not liking the spread operator ([related stack overflow](https://stackoverflow.com/questions/45120789/spread-operator-and-eslint)). I think we need to specify ecmaVersion to be >=2018 in the eslint-config package to fix that.

scripts/compile-ajv-validators.js
matteenm: agree, will rename

.gitignore
matteenm: was referencing widget package's .gitignore file. will see if it's really needed.

.eslintrc.js
matteenm: yeah np, I can update eslint-config

tinaszheng (approved):

PR #6507, opened 2023-05-08
fix: handle switchChain failure in swap

vm:
## Description
https://uniswap-labs.sentry.io/issues/3788318733/events/6d4f1062f25a40db900229797c48e1fe/?project=4504255148851200&referrer=next-event

`message: User rejected the request.` means the user rejected in their wallet when we prompt to switch chains. This is an example where that rejection is not caught.

Right now this ignores user rejected errors (meaning the user stays on the same chain) and continues to throw for other errors. This adds a note to a JIRA ticket to improve the failure case UX.

### Reproducing the error
1. Connect to mainnet
2. Go to an arbitrum token details page
3. Click "connect to ethereum"
4. Cancel the request in wallet
5. See the error in console

### QA (ie manual testing)
- [x] Run through the same steps and see no error

### Automated testing
- [ ] Unit test
- [ ] Integration/E2E test

I didn't add these because we can't switch chain via tests easily yet.

### DIFF
--- a/src/pages/Swap/index.tsx
+++ b/src/pages/Swap/index.tsx
@@ -712,6 +712,8 @@ export function Swap({
                   await switchChain(connector, chainId)
                 } catch (error) {
                   // Ignore error, which keeps the user on the previous chain.
+                  // TODO(WEB-3306): This UX could be improved to handle user rejected errors
+                  // differently compared to other failures.
                 }
               }}
             >
--- a/src/pages/Swap/index.tsx
+++ b/src/pages/Swap/index.tsx
@@ -707,8 +707,12 @@ export function Swap({
             </TraceEvent>
           ) : chainId && chainId !== connectedChainId ? (
             <ButtonPrimary
-              onClick={() => {
-                switchChain(connector, chainId)
+              onClick={async () => {
+                try {
+                  await switchChain(connector, chainId)
+                } catch (error) {
+                  // Ignore error, which keeps the user on the previous chain.
+                }
               }}
             >
               Connect to {getChainInfo(chainId)?.label}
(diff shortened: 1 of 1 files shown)


src/pages/Swap/index.tsx
```diff
@@ -707,8 +707,12 @@ export function Swap({
             </TraceEvent>
           ) : chainId && chainId !== connectedChainId ? (
             <ButtonPrimary
-              onClick={() => {
-                switchChain(connector, chainId)
+              onClick={async () => {
+                try {
+                  await switchChain(connector, chainId)
+                } catch (error) {
+                  // Ignore error, which keeps the user on the previous chain.
```
zzmp, line 714: This should be checking if its a rejection from the user before ignoring the error.
You can re-use the existing utility for swap errors to disambiguate user rejections.

zzmp (changes requested):

src/pages/Swap/index.tsx
vm: Even if the chain switch fails for other reasons, I think the way to handle it is the same? I talk about this in my PR description also.

zzmp: > I'd suggest to ticket that as part of the web team backlog and come back to it, as this is a good incremental improvement.

You should add a TODO referencing the ticket as a comment here.

zzmp (changes requested):

src/pages/Swap/index.tsx
just-toby: i agree we shouldn't start ignoring all errors here, just the user rejected ones

just-toby (changes requested):

just-toby (changes requested):
test needed (sorry to block on this, but it's the web team protocol now)

vm:
no problem that's fair, i'll add tests. i just realized i don't need to actually switch chain here, just mock to return specific errors.

zzmp (approved):

just-toby (approved):

PR #6508, opened 2023-05-08
feat: refactor Swap Button and  state management

just-toby:
## Description

Refactor the Swap Button into separate components, and move more state into `DerivedSwapInfo`, and make all the `DerivedSwapInfo` state available to the component and children via a Context. 

Motivation:
All of this state was just living in the body of the Swap Component, which makes the file long and difficult to parse. Since it's state, it was already causing the entire Swap Component to re-render, so this is more of a re-organization of code than a functional change.
Also, a refactor like this is needed if we're going to start modularizing different parts of the UI - e.g. separating the Button/Footer, inputs, header instead of keeping them in this monolithic component. The Context allows children at any level below the Swap Component to access swap state without passing tons of props (for an example, see `ConfirmSwapModal`)

As you can see by the added tests, modularizing like this makes unit testing easier too.

Next Steps:
This was just a first step in this direction, only focused on the SwapButton and state that it needs. We can continue to modularize pieces of the Swap Component and add more shared state to the `DerivedSwapInfoContext`. 

### QA (ie manual testing)

- [x] manually swap and verify that the Swap Component behaves the same

### Automated testing

- [x] Unit test (added several new test suites)
- [x] Integration/E2E test (automated test runs)

### DIFF
--- a/src/pages/Swap/SwapButton.tsx
+++ b/src/pages/Swap/SwapButton.tsx
@@ -112,7 +112,7 @@ export function SwapButton({ chainId, onSwapClick }: SwapButtonProps) {
 
   if (swapIsUnsupported) {
     return (
-      <ButtonPrimary disabled={true}>
+      <ButtonPrimary disabled={true} data-testid="swap-button">
         <ThemedText.DeprecatedMain mb="4px">
           <Trans>Unsupported Asset</Trans>
         </ThemedText.DeprecatedMain>
@@ -128,7 +128,7 @@ export function SwapButton({ chainId, onSwapClick }: SwapButtonProps) {
         properties={{ received_swap_quote: getIsValidSwapQuote(trade, tradeState, swapInputError) }}
         element={InterfaceElementName.CONNECT_WALLET_BUTTON}
       >
-        <ButtonLight onClick={toggleWalletDrawer} fontWeight={600}>
+        <ButtonLight data-testid="swap-button" onClick={toggleWalletDrawer} fontWeight={600}>
           <Trans>Connect Wallet</Trans>
         </ButtonLight>
       </TraceEvent>
@@ -138,6 +138,7 @@ export function SwapButton({ chainId, onSwapClick }: SwapButtonProps) {
   if (chainId && chainId !== connectedChainId) {
     return (
       <ButtonPrimary
+        data-testid="swap-button"
         onClick={() => {
           switchChain(connector, chainId)
         }}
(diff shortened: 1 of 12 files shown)


src/pages/Swap/SwapButton.tsx
```diff
@@ -0,0 +1,233 @@
...
+import invariant from 'tiny-invariant'
+import { getIsValidSwapQuote } from 'utils/getIsValidSwapQuote'
+import { switchChain } from 'utils/switchChain'
+
+import { Field } from '../../state/swap/actions'
+import { WrapButton } from './WrapButton'
+
+interface SwapButtonProps {
+  chainId: SupportedChainId | undefined
+  onSwapClick: () => void
```
just-toby, line 30: why pass `onSwapClick` as a prop when we can derive it from hooks like the Wrap callback below - maybe remove this

src/pages/Swap/SwapButton.tsx
```diff
@@ -0,0 +1,233 @@
...
+      <GrayCard style={{ textAlign: 'center' }} data-testid="swap-button">
+        <ThemedText.DeprecatedMain mb="4px">
+          <Trans>Insufficient liquidity for this trade.</Trans>
+        </ThemedText.DeprecatedMain>
+      </GrayCard>
+    )
+  }
+
+  if (isValid && allowance.state === AllowanceState.REQUIRED) {
+    return (
```
just-toby, line 166: This should have it's own component and set of tests - todo or update this PR?

src/pages/Swap/index.tsx
```diff
@@ -686,129 +620,24 @@ export function Swap({
...
+              if (isExpertMode) {
+                handleSwap()
+              } else {
+                setSwapState({
+                  tradeToConfirm: trade,
+                  attemptingTxn: false,
+                  swapErrorMessage: undefined,
+                  showConfirm: true,
+                  txHash: undefined,
+                })
```
just-toby, line 637: move this into SwapButton

src/state/swap/hooks.tsx
```diff
@@ -85,8 +85,85 @@ export function useDerivedSwapInfo(
...
+    data: undefined,
+    isLoading: false,
+  },
+  fiatValueTradeOutput: {
+    data: undefined,
+    isLoading: false,
+  },
+})
+
+export const DerivedSwapInfoContextProvider = ({
```
just-toby, line 148: this is not a hook - should i move this to a different file in this directory that's just for the Derived State Context ?

src/state/swap/hooks.tsx
```diff
@@ -85,8 +85,85 @@ export function useDerivedSwapInfo(
...
+}>) => {
+  return (
+    <DerivedSwapInfoContext.Provider value={useDerivedSwapInfo(state, chainId)}>
+      {children}
+    </DerivedSwapInfoContext.Provider>
+  )
+}
+
+// from the current swap inputs, compute the best trade and return it.
+function useDerivedSwapInfo(state: SwapState, chainId: SupportedChainId | undefined): DerivedSwapInfo {
```
just-toby, line 164: `const trade = useBestTrade(...)` is obviously a very important bit of state that was already in this hook. does it make sense to split up `useDerivedSwapInfo` into multiple hooks to organize the state further? is there any perf benefit to keeping expensive hooks like `useBestTrade` separate from the rest of the derived state?

vm:
This is a very large PR. Is there any way we could add this in pieces under a feature flag? Maybe we add the components even if they're not integrated yet first?

I think it'd be difficult to give a great review on a diff this large.

PR #6509, opened 2023-05-08
fix: add chunkResponseStatus tag

vm:
## Description
- Adds `chunkResponseStatus` as a tag on chunk request failure errors.
- Removes filtering for 499 errors so we can see which statuses are occurring in Sentry directly.

### Automated testing
- [x] Unit test


src/tracing/errors.ts
```diff
@@ -15,26 +15,46 @@ function isEthersRequestError(error: Error): error is Error & { requestBody: str
...
 
 export function beforeSend(event: ErrorEvent, hint: EventHint) {
-  // Since the interface currently uses HashRouter, URLs will have a # before the path.
-  // This leads to issues when we send the URL into Sentry, as the path gets parsed as a "fragment".
-  // Instead, this logic removes the # part of the URL.
-  // See https://romain-clement.net/articles/sentry-url-fragments/#url-fragments
-  if (event.request?.url) {
-    event.request.url = event.request.url.replace('/#', '')
-  }
+  if (event) {
```
zzmp, line 18: event should always be defined - do we need this top-level conditional check?

src/tracing/errors.ts
```diff
@@ -15,26 +15,46 @@ function isEthersRequestError(error: Error): error is Error & { requestBody: str
...
+        const status = getChunkResponseStatus(asset)
+        if (status) {
+          if (!event.tags) {
+            event.tags = {}
+          }
+          event.tags.chunkResponseStatus = status
+        }
+      }
+    }
+  }
```
zzmp, line 50: I think you should move all of this into a helper function.

src/tracing/errors.test.ts
```diff
@@ -1,12 +1,68 @@
...
+          responseStatus: 200,
+        } as PerformanceEntry,
+      ])
+      const originalException = new Error('Loading CSS chunk 12 failed. (./static/css/12.d5b3cfe3.chunk.css)')
+      expect((beforeSend(ERROR, { originalException }) as Event).tags?.chunkResponseStatus).toBe(200)
+    })
+  })
+})
+
 describe('filterKnownErrors', () => {
```
zzmp, line 66: Isn't this exported solely for testing?
Should it be tested through beforeSend? Alternatively, could you export your chunk helper method for isolated testing?

zzmp (changes requested):

src/tracing/errors.test.ts
vm: probably, i can change that all in a separate pr

src/setupTests.ts
```diff
@@ -20,6 +20,8 @@ if (typeof global.TextEncoder === 'undefined') {
 
 global.ResizeObserver = ResizeObserver
 
+global.origin = 'https://app.uniswap.org'
```
zzmp, line 23: Why is this necessary? You should leave a comment explaining it.

src/tracing/errors.ts
```diff
@@ -9,45 +9,51 @@ declare global {
   }
 }
 
-/** Identifies ethers request errors (as thrown by {@type import(@ethersproject/web).fetchJson}). */
+// Identifies ethers request errors (as thrown by {@type import(@ethersproject/web).fetchJson}).
```
zzmp, line 12: nit: This is a doc-comment, not an implementation comment, even though it's not exported, so you should prefer JSDoc syntax: `/**` instead of `//`.

src/tracing/errors.ts
```diff
@@ -9,32 +9,58 @@ declare global {
...
+      }
+    }
+  }
+}
 
+export function beforeSend(event: ErrorEvent, hint: EventHint) {
+  updateRequestUrl(event)
+  addChunkResponseStatusTag(event, hint)
   return filterKnownErrors(event, hint)
 }
```
zzmp, line 58: nit: put this at the top of the file - everything else is in scope due to hoisting. You already use hoisting for `getChunkResponseStatus`, so moving the exported bit to the very top (or the very bottom, without any hoisting) improves readability instead of leaving it in the middle.

zzmp (changes requested):

src/tracing/errors.ts
vm: should we add a lint for this?

src/tracing/errors.ts
vm: idk if there's a way to lint for this also?

src/tracing/errors.test.ts
vm: tbh we could probably lint for this too

vm: https://github.com/Uniswap/interface/pull/6513

just-toby (approved):

zzmp (approved):

PR #6510, opened 2023-05-08
feat: new routing diagram

grabbou:
## Description

_JIRA ticket:_
_Slack thread:_
_Relevant docs:_

## Screen capture

### Reproducing the error

1. 

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test


src/components/Popover/index.tsx
```diff
@@ -99,18 +99,15 @@ export default function Popover({
...
-      strategy: 'fixed',
-      modifiers: [
-        { name: 'offset', options: { offset: [offsetX, offsetY] } },
-        { name: 'arrow', options: { element: arrowElement } },
-        { name: 'preventOverflow', options: { padding: 8 } },
-      ],
-    }),
-    [arrowElement, offsetX, offsetY, placement]
-  )
+  const options: Options = {
```
tinaszheng, line 102: oh i feel like you may want to still use the `memo` here - are u sure `usePopper` doesn't care about the new object every time?

src/components/swap/AdvancedSwapDetails.tsx
```diff
@@ -45,119 +39,92 @@ function TextWithLoadingPlaceholder({
...
+            }
+          >
+            <ThemedText.BodySmall color="textSecondary">
+              <Trans>Expected output</Trans>
+            </ThemedText.BodySmall>
+          </MouseoverTooltip>
+        </RowFixed>
+        <TextWithLoadingPlaceholder syncing={syncing} width={65}>
+          <ThemedText.BodySmall>
+            {`${trade.outputAmount.toSignificant(6)} ${trade.outputAmount.currency.symbol}`}
```
cmcewen, line 107: nit: how should we treat formatting? i know we have a function some places but not all

src/components/swap/GasEstimateTooltip.tsx
```diff
@@ -0,0 +1,61 @@
...
+import styled from 'styled-components/macro'
+import { ThemedText } from 'theme'
+
+import { ReactComponent as GasIcon } from '../../assets/images/gas-icon.svg'
+import SwapRoute from './SwapRoute'
+
+const StyledGasIcon = styled(GasIcon)`
+  margin-right: 4px;
+  height: 18px;
+  & > * {
```
cmcewen, line 17: my css knowledge may be a bit weak but i think a comment like `// style all elements in the svg icon` would be useful - i had to look up what this syntax did

src/components/swap/GasEstimateTooltip.tsx
```diff
@@ -0,0 +1,61 @@
...
+export default function GasEstimateTooltip({
+  trade,
+  loading,
+  disabled,
+}: {
+  trade: InterfaceTrade<Currency, Currency, TradeType> // dollar amount in active chain's stablecoin
+  loading: boolean
+  disabled?: boolean
+}) {
+  const formattedGasPriceString = trade?.gasUseEstimateUSD
```
cmcewen, line 31: another place where formatting function might be nice

src/components/swap/GasEstimateTooltip.tsx
```diff
@@ -0,0 +1,61 @@
...
+    : undefined
+
+  return (
+    <MouseoverTooltip
+      disabled={disabled}
+      size={TooltipSize.Large}
+      // TODO(WEB-3304)
+      // Most of Swap-related components accept either `syncing`, `loading` or both props at the same time.
+      // We are often using them interchangeably, or pass both values as one of them (`syncing={loading || syncing}`).
+      // This is confusing and can lead to unpredicted UI behavior. We should refactor and unify this.
```
cmcewen, line 44: ➕

src/components/swap/PriceImpactWarning.tsx
```diff
@@ -40,7 +39,7 @@ export default function PriceImpactWarning({ priceImpact }: PriceImpactWarningPr
               </ThemedText.DeprecatedSubHeader>
             </RowFixed>
             <ThemedText.DeprecatedLabel textAlign="right" fontSize={14} color={theme.accentFailure}>
-              {formatPriceImpact(priceImpact)}
+              {`${priceImpact.multiply(-1).toFixed(2)}%`}
```
cmcewen, line 42: as a reader this is not how i would expect it to work (having to multiply by negative 1) so i do think continuing to have that named function would be nice

src/state/routing/slice.ts
```diff
@@ -13,11 +13,12 @@ export enum RouterPreference {
...
   API = 'api',
   CLIENT = 'client',
-
-  // Used internally for token -> USDC trades to get a USD value.
-  PRICE = 'price',
 }
 
+// This is excluded from `RouterPreference` enum because it's only used
+// internally for token -> USDC trades to get a USD value.
+export const INTERNAL_ROUTER_PREFERENCE_PRICE = 'price' as const
```
cmcewen, line 20: nit: something like USD_PRICE might be easier, took me scrolling till here to understand the var change in the files above

src/utils/getTokenPath.ts
```diff
@@ -0,0 +1,43 @@
+import { Protocol } from '@uniswap/router-sdk'
```
cmcewen, line 1: i think a utils folder under swap would be more meaningful - i think the path `src/utils` gives you no real information about what belongs in here

src/utils/getTokenPath.ts
```diff
@@ -0,0 +1,43 @@
...
+  path: [Currency, Currency, FeeAmount][]
+  protocol: Protocol
+}
+
+const V2_DEFAULT_FEE_TIER = 3000
+
+/**
+ * Loops through all routes on a trade and returns an array of diagram entries.
+ */
+export default function getTokenPath(trade: InterfaceTrade<Currency, Currency, TradeType>): RoutingDiagramEntry[] {
```
cmcewen, line 18: good function for a test?

src/components/swap/SwapRoute.tsx
```diff
@@ -1,157 +1,71 @@
...
+  const routes = getTokenPath(trade)
 
-  const formattedGasPriceString = trade?.gasUseEstimateUSD
-    ? trade.gasUseEstimateUSD.toFixed(2) === '0.00'
-      ? '<$0.01'
-      : '$' + trade.gasUseEstimateUSD.toFixed(2)
-    : undefined
+  const gasPrice =
+    // TODO(WEB-3303)
+    // Can `trade.gasUseEstimateUSD` be defined when `chainId` is not in `SUPPORTED_GAS_ESTIMATE_CHAIN_IDS`?
```
cmcewen, line 29: i also had this question 😄

cmcewen (approved):
all of my comments are nits - but i do think the useMemo in Popover should still be there. will defer to @tinaszheng's approval on that one

src/components/swap/AdvancedSwapDetails.tsx
tinaszheng: we should do a full refresh of swap flow formatting in a later proj!

src/components/swap/GasEstimateTooltip.tsx
tinaszheng: https://github.com/Uniswap/interface/pull/6499 done here :))

src/components/Popover/index.tsx
grabbou: My bad, double checked in the docs! https://popper.js.org/react-popper/v2/faq/#why-i-get-render-loop-whenever-i-put-a-function-inside-the-popper-configuration

src/state/routing/useRoutingAPITrade.ts
```diff
@@ -22,7 +22,7 @@ export function useRoutingAPITrade<TTradeType extends TradeType>(
   tradeType: TTradeType,
   amountSpecified: CurrencyAmount<Currency> | undefined,
   otherCurrency: Currency | undefined,
-  routerPreference: RouterPreference
+  routerPreference: RouterPreference | typeof INTERNAL_ROUTER_PREFERENCE_PRICE
```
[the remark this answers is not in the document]
grabbou: Agreed. Already added as a TODO. Happy to do that follow-up, but need to look at your PR too!

src/components/Tooltip/index.tsx
```diff
@@ -1,14 +1,29 @@
...
-import { ReactNode, useEffect, useState } from 'react'
+import { PropsWithChildren, ReactNode, useEffect, useState } from 'react'
 import styled from 'styled-components/macro'
 import noop from 'utils/noop'
 
 import Popover, { PopoverProps } from '../Popover'
 
-export const TooltipContainer = styled.div`
-  max-width: 256px;
+export enum TooltipSize {
```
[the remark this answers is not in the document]
grabbou: Yeah, it's not in the Styled Component `theme` (if that's the one you meant), rather, living as an inline value.

src/components/Tooltip/index.tsx
```diff
@@ -1,14 +1,29 @@
...
+const getPaddingForSize = (size: TooltipSize) => {
+  switch (size) {
+    case TooltipSize.Small:
+      return '12px'
+    case TooltipSize.Large:
+      return '16px 20px'
+  }
+}
+
+const TooltipContainer = styled.div<{ size: TooltipSize }>`
```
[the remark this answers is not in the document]
grabbou: Yeah, it is, however, I have run across mixed style guides so far in our codebase when to prefix with `$`. For example, we have `Column` and `Row` components that both take `gap`. Historically, I have only prefixed with `$` to avoid setting an HTML attribute when there was a name collision.

Either way, I would be happy to change, but would be great to establish a common way forward.

src/components/swap/AdvancedSwapDetails.tsx
```diff
@@ -1,30 +1,24 @@
...
-import { AutoColumn } from '../Column'
+import Column from '../Column'
 import { RowBetween, RowFixed } from '../Row'
-import { MouseoverTooltip } from '../Tooltip'
-import FormattedPriceImpact from './FormattedPriceImpact'
-
-const StyledCard = styled(Card)`
-  padding: 0;
-`
+import { MouseoverTooltip, TooltipSize } from '../Tooltip'
```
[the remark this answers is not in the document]
grabbou: Good catch! Personally, I use relatives, but in this codebase, we should stick to absolute. For some reason, VSCode failed here. Will check this and other imports.

src/components/swap/AdvancedSwapDetails.tsx
grabbou: Agreed, based on experience working on mobile, this way is not the beast for many reasons

src/state/routing/slice.ts
grabbou: FWIW we will be moving that completely in a follow-up PR. I agree it is not the best name, but it is being used as a substitute for `RouterPreference` enum, so I wanted to make it as similar in name as possible. That way `RouterPreference | typeof INTERNAL_ROUTER_PREFERENCE_PRICE` reads relatively okay while scanning the codebase.

src/utils/getTokenPath.ts
grabbou: Might be something to look at in the future - right now, we have all utilities in `src/utils`, such as `calculateSlippageAmount` that would be also very useful to be grouped as swap/token/trade related. Maybe we should consider having that form of grouping in the future?

src/components/swap/PriceImpactWarning.tsx
grabbou: To be honest, after playing around with the UI, the equivallent (self explanatory) is:
```
{`-${priceImpact.toFixed(2)}%`}
```
so I just went ahead and modified it.

src/utils/getTokenPath.ts
grabbou: Good call. Added a snapshot test and actually ended up renaming it to something more explicit.

src/components/swap/GasEstimateTooltip.tsx
grabbou: Yay!

src/components/swap/GasEstimateTooltip.tsx
grabbou: Yeah, we're doing that a couple of times. FWIW Tina is tackling in her follow-up by refactoring this logic.

src/components/swap/RouterLabel.tsx
```diff
@@ -1,58 +1,15 @@
...
-  color: ${({ theme }) => theme.accentSuccess};
-
-  @supports (-webkit-background-clip: text) and (-webkit-text-fill-color: transparent) {
-    background-image: linear-gradient(90deg, #2172e5 0%, #54e521 163.16%);
-    -webkit-background-clip: text;
-    -webkit-text-fill-color: transparent;
+  switch (routerPreference) {
+    case RouterPreference.AUTO:
+    case RouterPreference.API:
+      return <ThemedText.BodySmall>Uniswap API</ThemedText.BodySmall>
```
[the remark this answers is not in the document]
grabbou: I think these are proprietary names (we don't translate them in other places as well). I will ask Fred in the meantime, but let's not block on this.

grabbou:
Thanks @tinaszheng and @cmcewen for the review! I have updated some of the nits and will need your re-approval before final merging!

cmcewen (approved):

PR #6511, opened 2023-05-08
test(e2e): mock uniswap token lists from ipfs

zzmp:
## Description

Fixes flakiness in tests caused by slow loading of token lists over IPFS, by serving the default token list through an HTTP interceptor in cypress E2E tests.

### Automated testing
- [x] Integration/E2E test

### DIFF
--- a/cypress/support/e2e.ts
+++ b/cypress/support/e2e.ts
@@ -117,15 +117,9 @@ beforeEach(() => {
       )
     })
     // Mock our own token list responses to avoid the latency of IPFS.
-    .intercept('https://gateway.ipfs.io/ipns/tokens.uniswap.org', (req) => {
-      req.reply(TokenListJSON)
-    })
-    .intercept('https://gateway.ipfs.io/ipns/extendedtokens.uniswap.org', (req) => {
-      req.reply(200)
-    })
-    .intercept('https://gateway.ipfs.io/ipns/unsupportedtokens.uniswap.org', (req) => {
-      req.reply(200)
-    })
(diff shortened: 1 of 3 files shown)


cypress/support/e2e.ts
```diff
@@ -115,6 +116,10 @@ beforeEach(() => {
         })
       )
     })
+    // Mock our own token list responses to avoid the latency of IPFS.
+    .intercept('https://gateway.ipfs.io/ipns/tokens.uniswap.org', TokenListJSON)
+    .intercept('https://gateway.ipfs.io/ipns/extendedtokens.uniswap.org', { statusCode: 201, body: { tokens: [] } })
+    .intercept('https://gateway.ipfs.io/ipns/unsupportedtokens.uniswap.org', { statusCode: 201, body: { tokens: [] } })
```
just-toby, line 122: can these be overridden in individual tests? there may be a use case for that especially the unsupported list

cypress/support/e2e.ts
```diff
@@ -9,6 +9,7 @@ import '@cypress/code-coverage/support'
 import 'cypress-hardhat/lib/browser'
 
 import { Eip1193Bridge } from '@ethersproject/experimental/lib/eip1193-bridge'
+import TokenListJSON from '@uniswap/default-token-list'
```
just-toby, line 12: should we try to prevent this from being imported into product code? not sure how large it is

just-toby (approved):

tinaszheng (approved):

PR #6512, opened 2023-05-08
feat: add retry logic for dynamic imports

vm:
## Description
Attempts to resolve chunk errors by making multiple attempts to fetch the chunk. Currently, any network error on lazy loaded chunks will cause the error page to show up. Instead, we should try 3 times with backoff before showing the error state.

See more here: https://github.com/webpack/webpack/issues/14115#issuecomment-1079883725

### Reproducing the error
Screen Shot 2023-05-12 at 9 51 22 AM

- Go to app.uniswap.org
- Clear out network requests
- Go to a token details page
- Filter for `chunk`
- Block both chunk URLs
- See error when you reload

### Automated testing
- [x] Unit test
- [x] Integration/E2E test

I also added a lint rule to prevent future imports not using retry.
Screen Shot 2023-05-16 at 12 50 08 PM
Screen Shot 2023-05-16 at 12 50 17 PM


JFrankfurt (approved):

zzmp (changes requested):
Are there other assets which are loaded (like initial water-falled assets), which may not use retries because they aren't going through `lazy`. Is there a way to capture those and force them to use `retry` as well?

zzmp:
Could you test all branches? codecov only report 50% coverage of branching.

src/components/TopLevelModals/index.tsx
```diff
@@ -8,10 +8,11 @@ import useAccountRiskCheck from 'hooks/useAccountRiskCheck'
...
 import { useModalIsOpen, useToggleModal } from 'state/application/hooks'
 import { ApplicationModal } from 'state/application/reducer'
+import { retry } from 'utils/retry'
 
-const Bag = lazy(() => import('nft/components/bag/Bag'))
-const TransactionCompleteModal = lazy(() => import('nft/components/collection/TransactionCompleteModal'))
-const AirdropModal = lazy(() => import('components/AirdropModal'))
+const Bag = lazy(retry(() => import('nft/components/bag/Bag')))
+const TransactionCompleteModal = lazy(retry(() => import('nft/components/collection/TransactionCompleteModal')))
+const AirdropModal = lazy(retry(() => import('components/AirdropModal')))
```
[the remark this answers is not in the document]
vm: i actually can't make it do that because then it will fail our `no-unused-exports` lint rule. because then the imports can't be statically analyized.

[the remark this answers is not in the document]
vm: I'll make a ticket to add a separate PR for the lint rule?

vm:
I don't think we need to change anything about other forms of fetching for now because only chunk loading failures on lazy load are going to crash the app. The rest should be caught normally.

package.json
```diff
@@ -24,7 +24,7 @@
     "build:e2e": "REACT_APP_CSP_ALLOW_UNSAFE_EVAL=true REACT_APP_ADD_COVERAGE_INSTRUMENTATION=true craco build",
     "analyze": "source-map-explorer 'build/static/js/*.js' --only-mapped",
     "serve": "serve build -l 3000",
-    "lint": "yarn eslint --ignore-path .gitignore --cache --cache-location node_modules/.cache/eslint/ .",
+    "lint": "yarn eslint --ignore-path .gitignore --cache --cache-location node_modules/.cache/eslint/ --rulesdir ./rules .",
```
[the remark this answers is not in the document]
vm: apparently the only way to do this would be to eject because rulesdir isn't an option on craco or yarn start

.eslintrc.js
```diff
@@ -2,13 +2,18 @@
 
 require('@uniswap/eslint-config/load')
 
+const rulesDirPlugin = require('eslint-plugin-rulesdir')
+rulesDirPlugin.RULES_DIR = 'rules'
```
zzmp, line 6: change this directory name to eslint-rules

rules/enforce-retry-on-import.js
```diff
@@ -0,0 +1,34 @@
...
+            grandParent &&
+            grandParent.type === 'CallExpression' &&
+            grandParent.callee.name === 'retry' &&
+            grandParent.arguments.length === 1 &&
+            grandParent.arguments[0].type === 'ArrowFunctionExpression'
+          )
+        ) {
+          context.report({
+            node,
+            message: 'Dynamic import should be wrapped in the pattern retry(() => import(...))',
```
zzmp, line 28:
```suggestion
            message: 'Dynamic import should be wrapped in retry: `retry(() => import(...))`',
```

rules/enforce-retry-on-import.js
```diff
@@ -0,0 +1,34 @@
...
+  },
+  create(context) {
+    return {
+      ImportExpression(node) {
+        const grandParent = node.parent.parent
+        if (
+          !(
+            grandParent &&
+            grandParent.type === 'CallExpression' &&
+            grandParent.callee.name === 'retry' &&
```
zzmp, line 21:
```suggestion
            // Technically, we are only checking that a function named `retry` wraps the dynamic import.
            // We do not go as far as enforcing that it is import('utils/retry').retry
            grandParent.callee.name === 'retry' &&
```

src/utils/retry.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * Executes a Promise-based function multiple times with retries.
+ * If the function fails, it will retry a specified number of times, with an increasing delay between each attempt.
+ *
+ * @typeParam T - The type of the Promise's resolved value.
```
zzmp, line 5: You don't need this line, it's already documented through Typescript.

```suggestion
```

src/utils/retry.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * Executes a Promise-based function multiple times with retries.
```
zzmp, line 2: Promise-based isn't a thing - call it an async fn.

```suggestion
 * Executes an async function multiple times with retries.
```

src/utils/retry.ts
```diff
@@ -0,0 +1,36 @@
...
+ * If the function fails, it will retry a specified number of times, with an increasing delay between each attempt.
+ *
+ * @typeParam T - The type of the Promise's resolved value.
+ * @param fn - The Promise-based function that should be retried upon failure.
```
zzmp, line 6: Unnecessary, already in TypeScript.

```suggestion
```

src/utils/retry.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @param delay - The initial delay between retry attempts in milliseconds. This delay doubles after each failed attempt. Defaults to 1000 if not specified.
+ *
+ * @returns A Promise that resolves with the result of the original function if it succeeds within the specified number of retries, and rejects with the error from the last attempt otherwise.
+ *
+ * @example
+ * ```ts
+ * const fetchWithRetry = retry(fetchData, 5, 2000);
+ * fetchWithRetry.then(data => console.log(data)).catch(error => console.error(error));
+ * ```
+ */
```
zzmp, line 17:
```suggestion
/**
 * Executes a Promise-based function multiple times with exponential backoff (doubling).
 * @returns the result of the original function's final attempt.
 *
 * @example
 * ```ts
 * const fetchWithRetry = retry(fetchData, 5, 2000);
 * fetchWithRetry.then(data => console.log(data)).catch(error => console.error(error));
 * ```
 */
```

.eslintrc.js
```diff
@@ -2,13 +2,18 @@
 
 require('@uniswap/eslint-config/load')
 
+const rulesDirPlugin = require('eslint-plugin-rulesdir')
+rulesDirPlugin.RULES_DIR = 'eslint_rules'
```
zzmp, line 6: we don't use underscores anywhere - this should be `eslint-rules` for consistency

src/utils/retry.test.ts
```diff
@@ -0,0 +1,33 @@
+import { retry } from './retry'
+
+describe('retry function', () => {
+  jest.useRealTimers()
```
zzmp, line 4: You shouldn't need this - it should be the default.

```suggestion
```

src/utils/retry.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  let mockFn: jest.Mock
```
zzmp, line 5: This isn't shared state. You should re-declare it in each test: `const mockFn = jest.fn().mockResolvedValue(...)`. Trying to share it makes it harder to read the tests in isolation.

src/utils/retry.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    mockFn.mockResolvedValue(expectedResult)
+    const result = await retry(mockFn)
+    expect(result).toEqual(expectedResult)
+    expect(mockFn).toHaveBeenCalledTimes(1)
+  })
+
+  it('should retry the specified number of times before rejecting', async () => {
+    const error = new Error('Failure')
+    mockFn.mockRejectedValue(error)
+    await expect(retry(mockFn, 3, 1)).rejects.toEqual(error)
```
zzmp, line 22: Check that this takes at least 1 + 2 + 4 = 7 ms.

src/utils/retry.ts
```diff
@@ -0,0 +1,29 @@
...
+  return new Promise((resolve, reject) => {
+    const attempt = async (attempts: number, currentDelay: number) => {
+      try {
+        const result = await fn()
+        resolve(result)
+      } catch (error) {
+        if (attempts === retries) {
+          reject(error)
+        } else {
+          const exponentialBackoff = currentDelay * 2
```
zzmp, line 21: nit: `exponentialBackoffDelay`

zzmp (changes requested):

src/utils/retry.test.ts
vm: there isn't really a way to do this without using fake timers, which we haven't figured out yet

.eslintrc.js
vm: discussed in dms that we have `node_modules` already at the top level as an example

src/utils/retry.test.ts
zzmp: you can do this by just checking `Date.now()` before and after the call to `retry`, and ensuring the difference is >= 7.

zzmp (approved):

PR #6516, opened 2023-05-09
test(e2e): token details and test cleanup

zzmp:
## Description

Refactors and simplifies tests in service of making them faster and more debuggable:
- Improves support files by factoring out commands, and silencing spammy logs
  - All request logs and some stubs are silenced, as they flood the cypress log
- No longer waits on animations, so tests can execute faster
- Refactors tests to use explicit checks, instead of branching helper methods
  - Avoids branching logic in helper methods
  - Moves token details tests to the token details spec, as overloading the swap spec slows down total testing time

### Automated testing

- [x] Integration/E2E test


cypress/e2e/swap.test.ts
```diff
@@ -2,62 +2,22 @@ import { BigNumber } from '@ethersproject/bignumber'
...
 import { SupportedChainId, WETH9 } from '@uniswap/sdk-core'
 
-import { UNI as UNI_MAINNET, USDC_MAINNET } from '../../src/constants/tokens'
-import { FeatureFlag } from '../../src/featureFlags/flags/featureFlags'
-import { WETH_GOERLI } from '../fixtures/constants'
+import { UNI, USDC_MAINNET } from '../../src/constants/tokens'
 import { getTestSelector } from '../utils'
 
-describe('Swap', () => {
-  const verifyAmount = (field: 'input' | 'output', amountText: string | null) => {
```
zzmp, line 11: Branching logic makes it harder to follow what is actually being verified in tests.

cypress/e2e/swap.test.ts
```diff
@@ -2,62 +2,22 @@ import { BigNumber } from '@ethersproject/bignumber'
...
-        if ($body.find(getTestSelector('TokenSafetyWrapper')).length) {
-          return 'I understand'
-        }
-
-        return 'no-op' // Don't click on anything, a no-op
-      })
-      .then((content) => {
-        if (content !== 'no-op') {
-          cy.contains(content).click()
-        }
```
zzmp, line 44: This was causing flakiness, and should be fixed by a combination of using `ethereum: 'hardhat'` and stubbing the supported tokens list.

cypress/e2e/swap.test.ts
```diff
@@ -333,7 +278,7 @@ describe('Swap', () => {
     cy.visit('/swap', { ethereum: 'hardhat' })
       .hardhat()
       .then((hardhat) => {
-        cy.stub(hardhat.wallet, 'sendTransaction').rejects(new Error('user cancelled'))
+        cy.stub(hardhat.wallet, 'sendTransaction').log(false).rejects(new Error('user cancelled'))
```
zzmp, line 281: Wallet transactions dump pages of text to the log, so I silenced it.

cypress/e2e/swap.test.ts
```diff
@@ -431,16 +312,12 @@ describe('Swap', () => {
...
-          cy.contains('1inch').should('exist')
-
           // Select UNI as output token
           cy.get(getTestSelector('token-search-input')).clear().type('Uniswap')
           cy.get(getTestSelector('currency-list-wrapper'))
             .contains(/^Uniswap$/)
             .first()
-            .should('exist')
-            .click()
+            .click({ force: true })
```
zzmp, line 320: This was flaky because our scrolling library was mid-scroll when cypress kicked in, so there was a `pointer-events: none` on the element. I got around it with `force: true`.

cypress/support/e2e.ts
```diff
@@ -6,125 +6,17 @@
...
+import './setupTests'
+
+// Squelch logs from fetches, as they clutter the logs so much as to make them unusable.
+// See https://docs.cypress.io/api/commands/intercept#Disabling-logs-for-a-request.
+// TODO(https://github.com/cypress-io/cypress/issues/26069): Squelch only wildcard logs once Cypress allows it.
+const log = Cypress.log
+Cypress.log = function (options, ...args) {
+  if (options.displayName === 'script' || options.name === 'request') return
+  return log(options, ...args)
+} as typeof log
```
JFrankfurt, line 19: will this make it difficult to debug issues related to failing requests? I agree the volume is not manageable currently, but my first thought is that a more selective approach would be better.

cypress/e2e/swap.test.ts
```diff
@@ -431,16 +312,14 @@ describe('Swap', () => {
...
 
-          // Wait for the currency list to load
-          cy.contains('1inch').should('exist')
-
           // Select UNI as output token
           cy.get(getTestSelector('token-search-input')).clear().type('Uniswap')
           cy.get(getTestSelector('currency-list-wrapper'))
             .contains(/^Uniswap$/)
             .first()
-            .should('exist')
```
JFrankfurt, line 442: this `should` was here to allow time for the list to load--are you sure it's not necessary?

JFrankfurt (approved):

just-toby (approved):

cypress/e2e/swap.test.ts
zzmp: I removed this and saw no difference in the test.

cypress/support/setupTests.ts
```diff
@@ -0,0 +1,41 @@
...
+        payload_size_bytes: byteSize,
+        events_ingested: req.body.events.length,
+      })
+    )
+  }).intercept('https://*.sentry.io', { statusCode: 200 })
+
+  // Mock our own token list responses to avoid the latency of IPFS.
+  cy.intercept('https://gateway.ipfs.io/ipns/tokens.uniswap.org', TokenListJSON)
+    .intercept('https://gateway.ipfs.io/ipns/extendedtokens.uniswap.org', { statusCode: 404 })
+    .intercept('https://gateway.ipfs.io/ipns/unsupportedtokens.uniswap.org', { statusCode: 404 })
```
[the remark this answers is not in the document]
zzmp: I've confirmed through the (now removed) logs.

cypress/support/e2e.ts
zzmp: Yes, it will. I can update this in a follow-up to be more selective.

cypress/support/setupTests.ts
```diff
@@ -0,0 +1,41 @@
...
+import TokenListJSON from '@uniswap/default-token-list'
+
+beforeEach(() => {
+  // Many API calls enforce that requests come from our app, so we must mock Origin and Referer.
+  cy.intercept('*', (req) => {
+    req.headers['referer'] = 'https://app.uniswap.org'
+    req.headers['origin'] = 'https://app.uniswap.org'
+  })
+
+  // Infura uses a test endpoint, which allow-lists http://localhost:3000 instead.
```
tinaszheng, line 11: these comments are great!

tinaszheng (approved):

PR #6519, opened 2023-05-09
fix: properly filter AbortErrors

vm:
## Description
According to MDN:
```
Note: When abort() is called, the fetch() promise rejects with an Error of type DOMException, with name AbortError.
```

https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort

Therefore, my previous filtering was failing because I checked for AbortError: to be the start of the error message. Really, that only happens when toString is called later on.

## Test plan
- [x] Unit test

### DIFF
--- a/src/tracing/errors.test.ts
+++ b/src/tracing/errors.test.ts
@@ -203,7 +203,8 @@ describe('filterKnownErrors', () => {
   })
 
   it('filters AbortErrors', () => {
-    const originalException = new Error('AbortError: The user aborted a request.')
+    const originalException = new DOMException('The user aborted a request.') as any
+    originalException.name = 'AbortError'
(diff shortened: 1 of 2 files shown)


src/tracing/errors.test.ts
```diff
@@ -203,7 +203,8 @@ describe('filterKnownErrors', () => {
   })
 
   it('filters AbortErrors', () => {
-    const originalException = new Error('AbortError: The user aborted a request.')
+    const originalException = new DOMException('The user aborted a request.') as any
```
vm, line 206: name is read-only on DOMException

JFrankfurt (approved):

zzmp (approved):

PR #6524, opened 2023-05-10
chore: Replace widget skeleton with internal swap skeleton

tinaszheng:
## Description
Removes the dependency for rendering a Swap skeleton imported from @uniswap/widgets 

Received some new designs from @infredible - https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?type=design&node-id=13335-127826&t=CAu5FD8kCowy3z7c-0 

I haven't finished the whole thing ^ but close enough equivalent for now. Ticket for finishing it: WEB-2030

|image|image| not shown on mobile  |

Note I updated the Swap arrow to be the same one as in the Swap component

### QA (ie manual testing)
If you want to see this on Vercel, I recommend going to the token details page, then going to Network > Throttling > No network > REFRESH to see the loading state show

### Automated testing
Added snapshot test


src/components/swap/SwapSkeleton.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+function FloatingButton() {
+  return <Blob marginTop={0.55} height="3.5rem" width="100%" radius={0.75} />
+}
+
+export function SwapSkeleton() {
+  const theme = useTheme()
+
+  return (
+    <LoadingWrapper>
```
tinaszheng, line 85: note the general structure was copied over from https://github.com/Uniswap/widgets/blob/main/src/components/Swap/Skeleton.tsx - but made some simplifications

tinaszheng:
@zzmp would you mind taking a look at this failing cypress test? it seems vaguely related to mike's test failure from earlier today, but im unsure

src/components/swap/SwapSkeleton.tsx
```diff
@@ -0,0 +1,103 @@
...
+  justify-content: space-between;
+
+  padding: 8px;
+  border: ${({ theme }) => `1px solid ${theme.backgroundOutline}`};
+  border-radius: 1rem;
+`
+
+const Blob = styled.div<{ height: string; width: string; radius?: number; marginTop?: number }>`
+  background-color: ${({ theme }) => theme.backgroundModule};
+  border-radius: ${({ radius }) => (radius ?? 0.25) + 'rem'};
```
cmcewen, line 19: it seems like we settled on `rem` - is there anywhere in readme/notion we can point people to as the "why" and some migration tips

src/components/swap/SwapSkeleton.tsx
```diff
@@ -0,0 +1,103 @@
...
+  return (
+    <Row>
+      <ModuleBlob height="2rem" width="3.75rem" />
+      <ModuleBlob height="2rem" width="7.25rem" />
+    </Row>
+  )
+}
+
+function FloatingButton() {
+  return <Blob marginTop={0.55} height="3.5rem" width="100%" radius={0.75} />
```
cmcewen, line 78: this component api is a little confusing to me - it's not clear that `marginTop` is a number in rem, and it's inconsistent with the type of height/width which are strings. i'd vote either making everything a string or every prop a number with a `Rem` at the end

cmcewen (approved):
it's not a blocker for me but `Blob` props could be more consistent

cypress.config.ts
```diff
@@ -5,7 +5,6 @@ import { setupHardhatEvents } from 'cypress-hardhat'
 export default defineConfig({
   projectId: 'yp82ef',
   videoUploadOnPasses: false,
-  waitForAnimations: false,
```
grabbou, line 8: Is this intentional? I think it was added recently by Zach to speed things up

src/components/swap/SwapSkeleton.tsx
grabbou: I still use pixels in my PR 😆

cypress.config.ts
tinaszheng: yeah it was to fix a flakey test but this has been added on main already so I can rebase

src/pages/Swap/index.tsx
```diff
@@ -65,7 +65,8 @@ import { computeFiatValuePriceImpact } from '../../utils/computeFiatValuePriceIm
 import { maxAmountSpend } from '../../utils/maxAmountSpend'
 import { computeRealizedPriceImpact, warningSeverity } from '../../utils/prices'
 import { supportedChainId } from '../../utils/supportedChainId'
-const ArrowContainer = styled.div`
+
+export const ArrowContainer = styled.div`
```
grabbou, line 69: nit: can we move this to a separate file? Ideally, we could have the `ArrowDown` component exported from `theme/components` and the fact that it uses `ArrowContainer` would just be an implementation detail. It's quite confusing to import just the container to wrap your icon.

Update: Maybe `styleds.tsx` can work too.

src/components/swap/SwapSkeleton.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const theme = useTheme()
+
+  return (
+    <LoadingWrapper>
+      <FloatingTitle />
+      <InputColumn>
+        <FloatingInput />
+      </InputColumn>
+      <div>
+        <ArrowWrapper clickable={false}>
```
grabbou, line 91: nit: I would prefer if this was just an `Arrow` component, the fact we need wrapper/container/arrowdown is super confusing to me.

src/pages/Swap/index.tsx
```diff
@@ -65,7 +65,8 @@ import { computeFiatValuePriceImpact } from '../../utils/computeFiatValuePriceIm
...
   display: inline-block;
```
grabbou, line 70: Why does this component have display defined twice? I now it's not touched in this diff, but would be cool to clean up. Might be worth checking whether width and height 100% are required to be present as well, b/c that looks like a left-over.

src/components/Widget/index.tsx
```diff
@@ -198,7 +198,7 @@ export default function Widget({
   )
 }
 
-export function WidgetSkeleton({ width = DEFAULT_WIDGET_WIDTH }: { width?: number | string }) {
+function WidgetSkeleton({ width = DEFAULT_WIDGET_WIDTH }: { width?: number | string }) {
   const theme = useWidgetTheme()
   return <SwapWidgetSkeleton theme={theme} width={width} />
```
grabbou, line 203: Here, I think we're still rendering skeleton from Widget. Is this intentional?

src/components/swap/SwapSkeleton.tsx
grabbou: To be honest, this is perfect example where `rem` is confusing. I have no idea what `3.5rem` is here.

src/components/swap/SwapSkeleton.tsx
```diff
@@ -0,0 +1,103 @@
...
+          <ArrowContainer color={theme.textPrimary}>
```
grabbou, line 92: nit: do we really need `color` for `ArrowContainer`?

src/components/swap/SwapSkeleton.tsx
```diff
@@ -0,0 +1,103 @@
...
+            <ArrowDown size="16" color={theme.textTertiary} />
```
grabbou, line 93: nit: I would recommend wrapping this with a StyledComponent and passing `color` property there. I know it's kind of a nit pick, but if you remember my previous PR, I did the same for the Gas Icon, so it just keeps it more consistent.

src/components/Widget/index.tsx
```diff
@@ -36,7 +36,7 @@ import { useSyncWidgetSettings } from './settings'
 import { DARK_THEME, LIGHT_THEME } from './theme'
 import { useSyncWidgetTransactions } from './transactions'
 
-export const DEFAULT_WIDGET_WIDTH = 360
+const DEFAULT_WIDGET_WIDTH = 360
```
grabbou, line 39: nit: it would be great to keep this in sync with `TokenDetails/Skeleton`, so they never run out of sync. 

In reality, this is just bad design, because it seems to me that we have two different components, TokenDetails "main page" and TokenDetails "skeleton".

I believe, instead, we should have a single component, TokenDetails, where we would have "RightPanel" (with fixed width) and inside, based on the loading state, we would either render WidgetSkeleton or Widget. That way, both WidgetSkeleton and Widget would not have to set width, but use 100% of available space and delegate it back to RightPanel to decide. 

I don't know whether this change is possible within this PR, probably not, but I would leave a TODO comment + Jira ticket to refactor this, because I think this can be improved.

grabbou (commented):
Looks good! Left some comments, mostly nits, I think that we could do some further tweaking here and there (if time allows), otherwise, it's fine!

cypress.config.ts
grabbou: Ah, sorry, I misread. I thought it was the opposite. Yeah, this is good change!

src/components/swap/SwapSkeleton.tsx
tinaszheng: can i refactor this in a separate PR :D beacuse it's the same in the main Swap component and i don't want to make any changes there for this PR yet.

src/components/Widget/index.tsx
tinaszheng: agreed i was also really confused by it when i was refactoring! will make a linear ticket

src/pages/Swap/index.tsx
tinaszheng: can i do in a followup :D since this is already existing convention in the main Swap flow.

src/pages/Swap/index.tsx
tinaszheng: omg 😂

tinaszheng:
@cmcewen @grabbou made a few changes, please take a lewwwk~~~~

mainly i received some updated designs from Fred, so i incorporated some of that into this PR. 
- moved to using px because we actually don't have a strongly defined reason for moving to rem right now, so sticking with px until we come up with a migration strategy. as it is, the mixture of rem + px doesn't quite make sense to me, so i'm siding with @grabbou to stick with px for now
- refactored the components a bit to take advantage of flex gap instead of using one-off padding values to do layout. i feel like the new ver is much cleaner~
- removed the need for some props

cmcewen (approved):
🚢

src/pages/Swap/index.tsx
grabbou: Yes, please! Just make sure to not forget (I'll check hahaha jk)

src/components/Widget/index.tsx
grabbou: Sounds good!

src/components/Widget/index.tsx
grabbou: I usually add it to the code too, but don't want you to re-request another review from Connor so just for the future haha

grabbou (approved):

PR #6527, opened 2023-05-10
fix: portfolio balances after switching accounts

cartcrom:
## Description
Updates `PrefetchBalancesWrapper` to refetch balances after switching accounts. This component conservatively controls when new balances are fetched, and previously was not accounting for account change.

Also adds an e2e test for account switching in mini portfolio.

[JIRA ticket](https://uniswaplabs.atlassian.net/browse/WEB-3210)

## Screen capture

Before

https://user-images.githubusercontent.com/39385577/231274641-59716b07-c80a-4297-aaca-7b0302a77748.mov

After

https://user-images.githubusercontent.com/39385577/231274598-e9e76579-9592-482a-b805-d22e0f5d89eb.mov

### Reproducing the error

1. View your mini portfolio & then switch accounts in wallet (i.e. using second account on metamask)
2. Second account's portfolio will be stuck in loading state

### QA (ie manual testing)

- [ ] Test viewing balances after switching accounts with account drawer open
- [ ] Test viewing balances after switching accounts with account drawer closed

### Automated testing

Added e2e test

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -1,9 +1,6 @@
 import codeCoverageTask from '@cypress/code-coverage/task'
 import { defineConfig } from 'cypress'
 import { setupHardhatEvents } from 'cypress-hardhat'
-import * as dotenv from 'dotenv'
-
-dotenv.config()
 
 export default defineConfig({
   projectId: 'yp82ef',
@@ -25,5 +22,4 @@ export default defineConfig({
     baseUrl: 'http://localhost:3000',
     specPattern: 'cypress/e2e/**/*.{js,jsx,ts,tsx}',
   },
-  env: process.env,
 })
--- a/cypress/e2e/mini-portfolio/accounts.test.ts
+++ b/cypress/e2e/mini-portfolio/accounts.test.ts
@@ -2,7 +2,7 @@ import { getTestSelector } from '../../utils'
 
 describe('Mini Portfolio account drawer', () => {
   beforeEach(() => {
-    cy.intercept(Cypress.env('REACT_APP_AWS_API_ENDPOINT'), cy.spy().as('gqlSpy'))
+    cy.intercept(/api.uniswap.org\/v1\/graphql/, cy.spy().as('gqlSpy'))
     cy.visit('/swap', { ethereum: 'hardhat' })
   })
 


cypress/e2e/mini-portfolio.test.ts
```diff
@@ -0,0 +1,59 @@
...
+
+        // A shortened version of the first account's address should be shown
+        cy.contains(accountA.slice(0, 6)).should('exist')
+
+        // Stores the current portfolio balance to later compare to next account's balance
+        cy.get(getTestSelector('portfolio-total-balance'))
+          .invoke('text')
+          .then((originalBalance) => {
+            // Simulates the wallet changing accounts via eip-1193 event
+            cy.window().then((win) => win.ethereum.emit('accountsChanged', [accountB]))
```
JFrankfurt, line 48: @zzmp I'm wondering if there's a way to add account switching directly to the hardhat utils

src/components/AccountDrawer/PrefetchBalancesWrapper.tsx
```diff
@@ -27,12 +27,12 @@ function useHasUpdatedTx() {
   const previousPendingTxs = usePrevious(pendingTxs)
 
   return useMemo(() => {
-    if (!previousPendingTxs) return false
+    if (!previousPendingTxs || !account) return false
     return Object.values(currentChainTxs).some(
-      (tx) => !isTxPending(tx) && wasPending(previousPendingTxs, tx),
+      (tx) => tx.from === account && !isTxPending(tx) && wasPending(previousPendingTxs, tx),
```
JFrankfurt, line 32: what about `tx.to`?

src/components/AccountDrawer/PrefetchBalancesWrapper.tsx
```diff
@@ -49,16 +49,19 @@ export default function PrefetchBalancesWrapper({ children }: PropsWithChildren)
...
-    if (!hasUpdatedTx) return
-
-    // If the drawer is open, fetch balances immediately, else set a flag to fetch on next hover
-    if (drawerOpen) fetchBalances()
-    else setHasUnfetchedBalances(true)
-  }, [drawerOpen, fetchBalances, hasUpdatedTx])
+    if (hasUpdatedTx || accountChanged) {
+      // If the drawer is open, fetch balances immediately, else set a flag to fetch on next hover
+      if (drawerOpen) fetchBalances()
+      else setHasUnfetchedBalances(true)
```
zzmp, line 62: Use brackets for this if-else

cypress/e2e/mini-portfolio.test.ts
zzmp: Could you request a util via an infra ticket and add a `TODO(INFRA-123): ` note above this line?

zzmp (approved):
looks great, 2 nits

tinaszheng (changes requested):

src/components/AccountDrawer/PrefetchBalancesWrapper.tsx
JFrankfurt: image
I think you should be able to add `tx.to` to the transaction adder from the ethers `TransactionResponse`

cartcrom: @tinaszheng I think that is out of scope for this PR as it would mean switching chain would remove any indication of a pending transaction, which might be desired but seems like a product choice beyond this bug fix.

cartcrom: @JFrankfurt transactions on interface that are put into the redux store will still only be "from" and not "to" the current account

cartcrom: we don't have any use for a "to" field currently

src/components/AccountDrawer/PrefetchBalancesWrapper.tsx
```diff
@@ -49,16 +49,22 @@ export default function PrefetchBalancesWrapper({ children }: PropsWithChildren)
...
+  const hasUpdatedTx = useHasUpdatedTx(account)
+  // Listens for account changes & recently updated transactions to keep portfolio balances fresh in apollo cache
   useEffect(() => {
-    if (!hasUpdatedTx) return
-
-    // If the drawer is open, fetch balances immediately, else set a flag to fetch on next hover
-    if (drawerOpen) fetchBalances()
-    else setHasUnfetchedBalances(true)
-  }, [drawerOpen, fetchBalances, hasUpdatedTx])
+    if (hasUpdatedTx || accountChanged) {
```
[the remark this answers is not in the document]
cartcrom: oooh ty good catch, I'll add account to the dependency array

src/components/AccountDrawer/PrefetchBalancesWrapper.tsx
tinaszheng: yeah don't think we need to do `tx.to` here - that's mostly for contract addresses in the context of swap

tinaszheng: @cartcrom it's not based on chain switching right? it's based on account switching, and removing pending tx based on account seems logical to me. you can also define a separate hook that just filters out based on active address and use that exclusively for this flow

tinaszheng (dismissed):

src/components/AccountDrawer/PrefetchBalancesWrapper.tsx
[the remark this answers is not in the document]
tinaszheng: if you add `account` here you shouldn't need to also calculate `accountChanged` on the outside

tinaszheng (changes requested):

src/components/AccountDrawer/PrefetchBalancesWrapper.tsx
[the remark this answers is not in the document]
cartcrom: oh so tru ty

src/components/AccountDrawer/PrefetchBalancesWrapper.tsx
cartcrom: ahh my bad I had totally conflated the two when I read your comment. yes will update or add new hook

cartcrom: talked to you in office abt adding a todo comment instead 🫡

tinaszheng (approved):

JFrankfurt (approved):

zzmp (approved):

PR #6533, opened 2023-05-10
fix: propagates user rejection errors

vm:
## Description
This reverts this PR: https://github.com/Uniswap/interface/pull/6330

I've changed my mind here, and think with the added context in Sentry these errors are worth monitoring again. We should know where we're not catching these errors, and address them.

## Test plan
- [x] Unit test

### DIFF
--- a/src/tracing/errors.test.ts
+++ b/src/tracing/errors.test.ts
@@ -74,6 +74,11 @@ describe('filterKnownErrors', () => {
     expect(filterKnownErrors(ERROR, { originalException })).toBe(ERROR)
   })
 
+  it('propagates user rejected request errors', () => {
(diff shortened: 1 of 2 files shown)


just-toby (approved):

tinaszheng:
can you say more about `the added context in Sentry these errors are worth monitoring again`?

tinaszheng:
still don't see how these are actionable though 🤔 isnt user rejection just.. user rejection?

vm:
we should be catching the user rejection errors still, even if that means we don't do anything with it. ex: https://github.com/Uniswap/interface/pull/6507

zzmp (approved):

PR #6538, opened 2023-05-10
feat: permit2 flow updates

just-toby:
## Description

Updates the Permit2 approval flow to happen in the `ConfirmSwapModal`, with the following changes:

1. remove the approval button from the top-level Swap component
2. restructure the `ConfirmSwapModal` to track whether the user is "reviewing", "allowing", or "pending confirmation". each of these states have their own UI associated, and the "Allowing" state has multiple steps depending on the value of the current permit2 token `allowance` 
3. split up `allowance.approveAndPermit()` into two functions so we can accurately track the user's progress through the full flow, with intermediate steps.
4. implement the new UI for "approval" and "pending confirmation" steps in the modal - there are some more recent design updates that need to be applied here, and will be coming in separate PRs

Next steps:
- implement the animations between the `PendingModalContent` steps
- update the loader icon and pending states to match new designs

_JIRA ticket:_ https://linear.app/uniswap/issue/WEB-2032/build-new-permit2-flow

### intermediate steps (NOTE: these are subject to change with new designs): 
| no approval needed | token and permit approval needed | only token needed | only permit needed |
| ------------------- | ----------------------------------- | ------------------- | -------------- |
| image | image  image     |  image    | image | paste_permit |

### confirm swap prompt:

image

### success state (NOTE: this is subject to change with new designs):
image

### QA (ie manual testing)

- [x] extensive manually testing to catch edge cases and each different variant of the flow

### Automated testing

- [x] Unit test
- [x] Integration/E2E test: https://github.com/Uniswap/interface/pull/6541


tinaszheng:
image
do you think we still need the blue dot step counter if theres only 1 step

just-toby:
> image do you think we still need the blue dot step counter if theres only 1 step

yes, i talked to fred about this

infredible:
@just-toby sry, i dont remember that convo. I think we should remove the dot if it's only the one step. Could be confusing.

cartcrom (changes requested):
I think having two sets of state, `currentStep` and `confirmModalState` is confusing since they are very related.

I think this ConfirmSwapModal.tsx would read better like this:

split `ConfirmModalState.ALLOWING` apart again: something like `ConfirmModalState.APPROVING_TOKEN` & `ConfirmModalState.PERMITTING`, then you can use confirmModalState instead of currentStep.

& I think PendingModalContent ideally should look something like:
```
<PendingModalContent
    stepsNeeded={<Array of ConfirmModalState>}
    trade={trade}
    currentState={confirmModalState}
    error={approvalError}
/>
```
I think it'd make more sense for PendingModalContent to contain all of the PendingModalStep instances, using the props you pass in to map from state to content

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -46,24 +60,175 @@ export default function ConfirmSwapModal({
...
+    async function startFlow() {
+      prepareSwapFlow()
+      if (allowance.state === AllowanceState.REQUIRED) {
+        await updateAllowance()
+      } else {
+        setConfirmModalState(ConfirmModalState.PENDING_CONFIRMATION)
+        onConfirm()
+      }
+    }
+    startFlow()
```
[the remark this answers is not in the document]
just-toby: it's actually not needed here, thanks

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -46,24 +60,175 @@ export default function ConfirmSwapModal({
...
+      })
+    } catch (e) {
+      setConfirmModalState(ConfirmModalState.REVIEWING)
+      if (didUserReject(e)) {
+        return
+      }
+      console.error(e)
+      setApprovalError('permit_approval')
+    }
+    return true
```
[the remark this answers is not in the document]
just-toby: also not needed, good catch

src/components/TransactionConfirmationModal/AnimatedConfirmation.tsx
```diff
@@ -38,11 +38,11 @@ const PolyLine = styled.polyline`
   animation: ${dashCheck} 0.9s 0.35s ease-in-out forwards;
 `
 
-export default function AnimatedConfirmation() {
+export default function AnimatedConfirmation({ size }: { size?: string }) {
   const theme = useTheme()
 
   return (
-    <Wrapper className="w4rAnimated_checkmark">
+    <Wrapper className="w4rAnimated_checkmark" size={size}>
```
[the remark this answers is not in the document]
just-toby: i can test that out - i assumed it was used to make the animation work but we'll see

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -46,86 +68,197 @@ export default function ConfirmSwapModal({
...
+      } else {
+        setConfirmModalState(ConfirmModalState.PERMITTING)
+        await allowance.permit()
+      }
+      sendAnalyticsEvent(InterfaceEventName.APPROVE_TOKEN_TXN_SUBMITTED, {
+        chain_id: chainId,
+        token_symbol: maximumAmountIn?.currency.symbol,
+        token_address: maximumAmountIn?.currency.address,
+        ...trace,
+      })
```
cartcrom, line 145: This will send an APPROVE_TOKEN_TXN_SUBMITTED event even when only a permit signature occurred

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -46,86 +68,197 @@ export default function ConfirmSwapModal({
...
+    } catch (e) {
+      setConfirmModalState(ConfirmModalState.REVIEWING)
+      if (didUserReject(e)) {
+        return
+      }
+      console.error(e)
+      setApprovalError(PendingModalError.PERMIT_ERROR)
+    }
+  }, [allowance, chainId, maximumAmountIn?.currency.address, maximumAmountIn?.currency.symbol, trace])
```
cartcrom, line 154: `updateAllowance` and the `requestSignature useEffect` above it seem a lot more complicated than useReviewState hooks in the widgets repo. I don't think you should need two separate call-sites for `allowance.permit()`, you can probably just combine the `requestSignature useEffect` into `updateAllowance`

cartcrom (changes requested):

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -1,39 +1,172 @@
...
+        return
+      }
+      console.error(e)
+      setApprovalError(
+        allowance.needsPermit2Approval ? PendingModalError.TOKEN_APPROVAL_ERROR : PendingModalError.PERMIT_ERROR
+      )
+    }
+  }, [allowance, chainId, maximumAmountIn?.currency.address, maximumAmountIn?.currency.symbol, trace])
+
+  const onStartSwapFlow = useCallback(async () => {
```
tinaszheng, line 106: i feel like one big change i'd prefer is, instead of having the main modal control updating the state for every screen based on the combination of these callbacks and useEffects, it would be better for each child screen to control its own `confirm` method and give it a `nextStep()` callback they can call when it's done. lmk if this makes sense, i can provide a code snippet if not

src/components/swap/ConfirmSwapModal.tsx
just-toby: we discussed this offline and i've made updates

src/components/swap/ConfirmSwapModal.tsx
just-toby: updated

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -1,39 +1,172 @@
...
+  allowedSlippage: Percent
+  onSwap: () => void
+  allowance: Allowance
+  doesTradeDiffer: boolean
+}) {
+  const [confirmModalState, setConfirmModalState] = useState<ConfirmModalState>(ConfirmModalState.REVIEWING)
+  const [approvalError, setApprovalError] = useState<PendingModalError>()
+  const [pendingModalSteps, setPendingModalSteps] = useState<PendingConfirmModalState[]>([])
+
+  const prepareSwapFlow = useCallback(() => {
```
grabbou, line 62: I don't think this needs to be a function. You could make it an array with `useMemo` and it would declaratively update itself when `allowance` is set. That way, it is not required to set it imperatively.

src/components/swap/ConfirmSwapModal.tsx
grabbou: Do we have separate child screens tho? The problem, which I outlined in the other PR is that we have a dynamic Modal component and we just replace the bottomContent/content callbacks.

I think this is slightly cleaner, at least we have unified control flow. I fear that if we spread it across multiple components, we may loose a sight of what's going on in an already quite complex file.

grabbou, line 106: nit: Also, we don't need `useCallback`. This function is not passed as a prop anywhere. Please consider cleaning up this line and other occurrences where reference equality is not important to the component.

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -1,39 +1,172 @@
...
+    }
+  }, [allowance, chainId, maximumAmountIn?.currency.address, maximumAmountIn?.currency.symbol, trace])
+
+  const onStartSwapFlow = useCallback(async () => {
+    setApprovalError(undefined)
+    if (allowance.state === AllowanceState.REQUIRED) {
+      await updateAllowance()
+    } else {
+      setConfirmModalState(ConfirmModalState.PENDING_CONFIRMATION)
+      onSwap()
```
grabbou, line 112: I think instead of calling `await updateAllowance()`, you should just do `await allowance.approve()` and just move the body of that function here instead.

My point is that `updateAllowance` sets some states for the component as well, so it immediately becomes more layered to understand how all state updates are performed here. Ideally, all state updates were performed in this function and only in this function and other functions would be simply for side effects.

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -42,11 +175,22 @@ export default function ConfirmSwapModal({
...
+      onSwap: onConfirm,
+      allowance,
+      doesTradeDiffer: Boolean(doesTradeDiffer),
+    })
+
+  const showAcceptChanges = Boolean(
+    trade && doesTradeDiffer && confirmModalState !== ConfirmModalState.PENDING_CONFIRMATION
   )
 
+  const confirmed = useIsTransactionConfirmed(txHash)
```
grabbou, line 192: Will this keep running multiple times every render? Is this desired?

grabbou (changes requested):

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -359,7 +333,7 @@ export default function TransactionConfirmationModal({
           currencyToAdd={currencyToAdd}
         />
       ) : (
-        content()
+        reviewContent()
```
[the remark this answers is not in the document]
just-toby: agreed - i want to do a follow up refactor for these components

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -1,39 +1,172 @@
...
 import SwapModalHeader from './SwapModalHeader'
 
+export enum ConfirmModalState {
+  REVIEWING,
+  APPROVING_TOKEN,
+  PERMITTING,
+  PENDING_CONFIRMATION,
+}
+
+function isAllowing(confirmModalState: ConfirmModalState) {
```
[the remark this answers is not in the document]
just-toby: i can rename to something like `isInApprovePhase` ? it's just a shorthand for checking if the current state is one of two values. i also use the function more in stacked diffs

src/components/swap/ConfirmSwapModal.tsx
just-toby: this actually needs to be a function, because we don't want the steps to update when you're in the middle of the flow. if the flow starts with 3 steps, it shouldn't lose the first 2 step indicators once the `allowance` is set

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -1,39 +1,172 @@
...
+      // These two lines capture the state update that should trigger the signature request:
+      !allowance.needsPermit2Approval &&
+      previousPermitNeeded
+    ) {
+      updateAllowance()
+    }
+  }, [allowance, previousPermitNeeded, updateAllowance])
+
+  // Automatically triggers signing swap tx if allowance requirements are met
+  useEffect(() => {
```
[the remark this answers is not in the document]
just-toby: i agree that the useEffect hooks to trigger the next steps are not ideal, but we need to wait for the `allowance` value from the hook in the parent component to update before we trigger the swap.

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -42,11 +175,22 @@ export default function ConfirmSwapModal({
...
+    useConfirmModalState({
+      trade,
+      allowedSlippage,
+      onSwap: onConfirm,
+      allowance,
+      doesTradeDiffer: Boolean(doesTradeDiffer),
+    })
+
+  const showAcceptChanges = Boolean(
+    trade && doesTradeDiffer && confirmModalState !== ConfirmModalState.PENDING_CONFIRMATION
```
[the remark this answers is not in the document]
just-toby: `PENDING_CONFIRMATION` means the first trade was already sent, so we don't want to show a price update at this point. 

this condition was added because in the new flow, we still use show component after the trade is sent

src/components/swap/ConfirmSwapModal.tsx
just-toby: discussed offline, need these because they're deps of the effect hooks

src/components/TransactionConfirmationModal/AnimatedConfirmation.tsx
```diff
@@ -1,8 +1,8 @@
 import styled, { keyframes, useTheme } from 'styled-components/macro'
 
-const Wrapper = styled.div`
-  height: 90px;
-  width: 90px;
+const Wrapper = styled.div<{ size?: string }>`
+  height: ${({ size }) => size ?? '90px'};
+  width: ${({ size }) => size ?? '90px'};
```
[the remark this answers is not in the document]
just-toby: this worked and the height/width were overridden correctly

src/components/swap/ConfirmSwapModal.tsx
just-toby: yeah, i can move this logic into `startSwapFlow` and just call `startSwapFlow` from the effect hooks below

src/components/swap/ConfirmSwapModal.tsx
just-toby: i think this only updates when the `txHash` changes or when the result of `useAllTransactions` hooks changes. i wasn't sure what the improvement should be here, but i moved this hook into the child component where it's actually used

src/components/TransactionConfirmationModal/AnimatedConfirmation.tsx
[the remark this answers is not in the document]
grabbou: awesome, good to know!

src/components/swap/ConfirmSwapModal.tsx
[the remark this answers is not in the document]
grabbou: If the name would be `isInApprovalPhase`, I would expect the second argument to be `||` instead of `&&`. Reading in English: either is approving or is approved. So yeah, I guess I just need to understand the reason behind conditional!

src/components/swap/ConfirmSwapModal.tsx
grabbou: I mean you could always make it `useMemo` instead of a function - right? That way it would only run every time allowance changes, no need to call it.

grabbou: In other words, why not just make it:
```ts
const steps = useMemo(() => {
 // do the thing
}, [allowance])
```

just-toby: @grabbou we don't want it to run when the allowance changes though, only when the user starts the flow. 

when the allowance changes, we move from the approval phase to the final step, but we don't want the number of steps indicated at the bottom of the modal to change (it should still be two or three depending on what it was at the start)

src/components/swap/ConfirmSwapModal.tsx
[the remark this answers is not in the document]
just-toby: this specific conditional is checking if we need to move the `confirmModalState` past the "approving" phase, based on whether `allowance` has updated. so this is expected to be &&  - it means "we are in the approval phase, but it's just ended because the final approval we needed has finalized"

src/components/swap/ConfirmSwapModal.tsx
just-toby: discussed offline

src/components/swap/ConfirmSwapModal.tsx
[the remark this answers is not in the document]
just-toby: discussed offline and i added a comment explaining this here

src/components/swap/ConfirmSwapModal.tsx
tinaszheng: @just-toby you can accomplish that by setting a variable that only reads from the initial value of `allowance`

```
const initialAllowance = useRef()

useEffect(() => {
   if (initialAllowance) return 

  initialAllowance.current = allowance
}, [allowance])

// and then use initialAllowance only
```

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -1,39 +1,170 @@
...
+          await allowance.permit()
+        }
+      } catch (e) {
+        setConfirmModalState(ConfirmModalState.REVIEWING)
+        if (didUserReject(e)) {
+          return
+        }
+        console.error(e)
+        setApprovalError(
+          allowance.needsPermit2Approval ? PendingModalError.TOKEN_APPROVAL_ERROR : PendingModalError.PERMIT_ERROR
```
tinaszheng, line 104: it would be really great in the future to break up the `allowance` hook into two separate ones because token approval and permitting are technically two different things and there's no reason to bundle them together

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -1,39 +1,170 @@
...
+  }, [allowance, chainId, maximumAmountIn?.currency.address, maximumAmountIn?.currency.symbol, onSwap, trace])
+
+  const previousPermitNeeded = usePrevious(
+    allowance.state === AllowanceState.REQUIRED ? allowance.needsPermit2Approval : undefined
+  )
+  useEffect(() => {
+    if (
+      allowance.state === AllowanceState.REQUIRED &&
+      allowance.needsSignature &&
+      // These two lines capture the state update that should trigger the signature request:
```
tinaszheng, line 120: this comment is confusing 🤔 not sure what this means

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -1,39 +1,170 @@
...
+    // Automatically triggers the next phase if the local modal state still thinks we're in the approval phase,
+    // but the allowance has been set. This will automaticaly trigger the swap.
+    if (isInApprovalPhase(confirmModalState) && allowance.state === AllowanceState.ALLOWED) {
+      // Caveat: prevents swap if trade has updated mid approval flow.
+      if (doesTradeDiffer) {
+        setConfirmModalState(ConfirmModalState.REVIEWING)
+        return
+      } else {
+        startSwapFlow()
+      }
```
tinaszheng, line 138:
```suggestion
     startSwapFlow()
```

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -1,39 +1,170 @@
...
+    }
+  }, [allowance, confirmModalState, doesTradeDiffer, startSwapFlow])
+
+  const onCancel = useCallback(() => {
```
tinaszheng, line 142: does this need to go in a `useCallback`?

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -57,80 +197,91 @@ export default function ConfirmSwapModal({
...
+        SwapEventName.SWAP_PRICE_UPDATE_ACKNOWLEDGED,
+        formatSwapPriceUpdatedEventProperties(trade, priceUpdate, SwapPriceUpdateUserResponse.REJECTED)
+      )
+    }
     onDismiss()
-  }, [onDismiss, priceUpdate, trade])
+    setTimeout(() => {
+      // Reset local state after the modal dismiss animation finishes, to avoid UI flicker as it dismisses
+      onCancel()
+    }, 200)
```
tinaszheng, line 211: put this variable in a constant

src/hooks/usePermit2Allowance.ts
```diff
@@ -101,24 +105,64 @@ export default function usePermit2Allowance(amount?: CurrencyAmount<Token>, spen
...
+    if (shouldRequestSignature) {
+      await updatePermitAllowance()
+    }
+  }, [shouldRequestSignature, updatePermitAllowance])
+
   return useMemo(() => {
     if (token) {
       if (!tokenAllowance || !permitAllowance) {
         return { state: AllowanceState.LOADING }
       } else if (!(isPermitted || isSigned)) {
```
tinaszheng, line 125: reuse `shouldRequestSignature` here

tinaszheng (changes requested):
overall looks great, just a few suggestions!

src/components/swap/ConfirmSwapModal.tsx
just-toby: https://linear.app/uniswap/issue/WEB-2079/[permit2-follow-up]-split-up-the-allowance-hook

src/components/swap/ConfirmSwapModal.tsx
just-toby: nope

src/components/swap/ConfirmSwapModal.tsx
tinaszheng: bump on this comment ^

tinaszheng (approved):

cartcrom (approved):

zzmp (approved):

PR #6539, opened 2023-05-11
build: report test failures via Slack

zzmp:
## Description

Reports test failures via Slack

_Slack thread:_ https://uniswapteam.slack.com/archives/C02T729PMQE/p1683730705255649

### QA (ie manual testing)

- [ ] Opened this PR with a failing build and ensured the webhook posted to #web-build-alerts.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,6 +30,7 @@ jobs:
         uses: ./.github/actions/report
         with:
           name: Lint
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_TEST_REPORTER_WEBHOOK }}
   
   typecheck:
     runs-on: ubuntu-latest
@@ -47,6 +48,7 @@ jobs:
         uses: ./.github/actions/report
         with:
           name: Typecheck
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_TEST_REPORTER_WEBHOOK }}
 
   deps-tests:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


JFrankfurt:
I think `.github/actions/report/action.yml` could have a more descriptive name

.github/workflows/test.yml
```diff
@@ -39,13 +44,23 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup
       - run: yarn yarn-deduplicate --strategy=highest --list --fail
+      - if: failure() && github.ref_name == 'main'
+        uses: ./.github/actions/report
+        with:
+          name: Dependency checks
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_TEST_REPORTER_WEBHOOK }}
```
JFrankfurt, line 63: failure here would not cause runtime issues, so I don't think it warrants an alert

cmcewen: that's true, but i guess i would personally just like to know about any red x's on main - and my understanding is this just pings a slack channel, not triggers pagerduty right?

cmcewen (approved):

.github/workflows/test.yml
zzmp: It would _likely_ not cause runtime issues, but it might cause some version skew bugs, and it _would_ prevent new PRs from being merged, as `deps-tests` is required to pass per our branch protections.

JFrankfurt (approved):

zzmp:
> I think `.github/actions/report/action.yml` could have a more descriptive name

I left it vague in case we switched it to a different reporting mechanism but i think you're right.

PR #6540, opened 2023-05-11
fix: remove top-level Fragment

vm:
## Description
This should resolve these errors: https://uniswap-labs.sentry.io/issues/4174993328/?project=4504255148851200&query=release%3A59b757dda03c2772c73b710fe4690208a5c7b001&referrer=issue-stream&sort=freq&statsPeriod=7d&stream_index=19

I found the solution here: https://github.com/remix-run/react-router/issues/8834#issuecomment-1118083034

The claim is that a top-level fragment will have issues when the DOM node is attempted to be removed. In this case, the Fragment isn't even necessary.

### Reproducing the error
I haven't actually been able to reproduce this locally.

### QA (ie manual testing)
I tested making sure the Swap page loads and works when I navigate away.

### Automated testing
- [ ] Unit test
- [ ] Integration/E2E test

I don't think any new tests are necessary here beyond making sure existing tests pass.

### DIFF
--- a/src/pages/Swap/index.tsx
+++ b/src/pages/Swap/index.tsx
@@ -140,20 +140,18 @@ export default function SwapPage({ className }: { className?: string }) {
   const loadedUrlParams = useDefaultsFromURLSearch()
   return (
     <Trace page={InterfacePageName.SWAP_PAGE} shouldLogImpression>
-      <>
-        <PageWrapper>
-          <Swap
-            className={className}
-            chainId={connectedChainId}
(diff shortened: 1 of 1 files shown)


just-toby (approved):
nice

cmcewen (approved):
nice - though you haven't been able to repro/confirm this fix it sounds like?

vm:
Yeah... can't figure out how to make it happen locally. I think for this we'll know if it works if the errors go away in Sentry.

PR #6543, opened 2023-05-11
chore: Goodbye widget :(

tinaszheng:
## Description
This PR removes the Widget dependency as a clean-up after returning on the Widget removal feature flag.

### QA (ie manual testing)
- [ ] Check token details page is still loading the main Swap component

### Automated testing
updated some of the unit tests that reference the `is_widget` tag


package.json
```diff
@@ -169,7 +169,6 @@
     "@uniswap/v3-core": "1.0.0",
     "@uniswap/v3-periphery": "^1.1.1",
     "@uniswap/v3-sdk": "^3.9.0",
-    "@uniswap/widgets": "^2.49.0",
```
zzmp, line 172: `@uniswap/widgets` was also the only thing requiring our `resolutions` - can you remove that field from the package.json?

src/tracing/trace.test.ts
```diff
@@ -38,12 +38,11 @@ describe('trace', () => {
   })
 
   it('records transaction', async () => {
-    const metadata = { data: { a: 'a', b: 2 }, tags: { is_widget: true } }
+    const metadata = { data: { a: 'a', b: 2 } }
```
zzmp, line 41: This test just verified that tags worked - can you change it from `is_widget` to another value?

src/tracing/trace.test.ts
```diff
@@ -74,17 +73,6 @@ describe('trace', () => {
     })
   })
 
-  describe('setTraceTag', () => {
```
zzmp, line 77: This test should be retained - just use a different dummy value.

src/tracing/trace.test.ts
```diff
@@ -166,15 +154,14 @@ describe('trace', () => {
   describe('traceChild', () => {
     it('starts a span under a transaction', async () => {
       await trace('test', ({ traceChild }) => {
-        traceChild('child', () => Promise.resolve(), { data: { e: 'e' }, tags: { is_widget: true } })
+        traceChild('child', () => Promise.resolve(), { data: { e: 'e' } })
```
zzmp, line 157: This test should continue to verify that tags work.

src/tracing/trace.test.ts
zzmp: You'll have to update the type of `TraceTags` locally to get this to work :(

zzmp (changes requested):

zzmp (approved):

PR #6546, opened 2023-05-12
fix: filter error caused by missing meta tags

vm:
## Description
https://uniswap-labs.sentry.io/issues/3788384150/events/db7c75d3f05a409780cebf540c3c69ca/?end=2023-05-12T14%3A15%3A59&project=4504255148851200&query=release%3A59b757dda03c2772c73b710fe4690208a5c7b001&referrer=next-event&sort=freq&start=2023-05-11T12%3A08%3A00&stream_index=7

Filters out errors caused by missing meta tags. This is not actionable because those browsers should be checking if they're available anyways. All of these tags are technically optional.

### Automated testing
- [x] Unit test


grabbou:
Are these checks caused by the browser? Is there anything we can do to avoid them from happening in first place rather than ignoring?

vm:
We could add the tags, but they are completely optional and I think it's a browser bug that they don't check if they exist before checking the content field

just-toby (approved):

cmcewen (approved):

PR #6547, opened 2023-05-12
refactor: filterKnownErrors -> shouldRejectError

vm:
## Description
Small refactor to return a boolean of whether to reject an error, instead of returning null in the helper function when we want to drop.

Now there is a comment explaining why we use null to drop events in a single location.

### Automated testing
- [x] Unit test

### DIFF
--- a/src/tracing/errors.ts
+++ b/src/tracing/errors.ts
@@ -69,7 +69,7 @@ function getChunkResponseStatus(asset?: string): number | undefined {
   return resource?.responseStatus
 }
 
-function shouldRejectError(error: unknown) {
+function shouldRejectError(error: EventHint['originalException']) {
   if (error instanceof Error) {
     // ethers aggressively polls for block number, and it sometimes fails (whether spuriously or through rate-limiting).
     // If block number polling, it should not be considered an exception.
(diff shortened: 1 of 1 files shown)


src/tracing/errors.ts
```diff
@@ -9,9 +9,16 @@ declare global {
 }
 
 export function beforeSend(event: ErrorEvent, hint: EventHint) {
```
zzmp, line 11: This should retain the typing that was previously used for `filterKnownErrors`. I also think it should keep the comment, and that the new comment on L12 is unnecessary with the doc-comment restored.

zzmp (changes requested):

src/tracing/errors.ts
```diff
@@ -8,10 +8,19 @@ declare global {
   }
 }
 
-export function beforeSend(event: ErrorEvent, hint: EventHint) {
+/**
+ * Filters known (ignorable) errors out before sending them to Sentry. Also, adds tags to the event.
+ * Intended as a {@link ClientOptions.beforeSend} callback. Returning null filters the error from Sentry.
+ */
+export const beforeSend: Required<ClientOptions>['beforeSend'] = (event: ErrorEvent, hint: EventHint) => {
```
zzmp, line 15: This is an exceptional case where I prefer `const fn = () => {` over `function fn() {` because you can then apply the derived type, as you do :)

zzmp (approved):

PR #6553, opened 2023-05-12
fix: polyfill ResizeObserver

vm:
## Description
https://uniswap-labs.sentry.io/issues/4178534607/?project=4504255148851200&referrer=release-issue-stream

ResizeObserver is not polyfilled by default. This makes sure we do that for legacy browsers (like slightly older Safari mobile). I switched the polyfill library to what's recommended by https://github.com/ZeeCoder/use-resize-observer#transpilation--polyfilling

### Automated testing
- [x] Unit test
- [x] Integration/E2E test


src/polyfills.ts
```diff
@@ -1,7 +1,12 @@
 import 'polyfill-object.fromentries'
 
+import { ResizeObserver } from '@juggle/resize-observer'
```
zzmp, line 3: Why change our dependency instead of just importing the existing package in `src/polyfills`? What makes `@juggle/resize-observer` preferable to `resize-observer-polyfill`?

NB: `resize-observer-polyfill` has much more usage, although it is larger and staler.

vm: it just looked like it was more maintained tbh, but we can leave the old one if better

cmcewen: if you look at yarn.lock we already had `@juggle/resize-observer` somewhere - the net of this is just removing a dep

cmcewen (approved):

src/setupTests.ts
```diff
@@ -1,10 +1,10 @@
 import '@testing-library/jest-dom' // jest custom assertions
 import 'jest-styled-components' // adds style diffs to snapshot tests
 
+import { ResizeObserver } from '@juggle/resize-observer'
```
zzmp, line 4: Instead of this, can you just `import 'polyfills'` on L3?

zzmp (approved):
lgtm with a nit: I think it'd be better to just `import 'polyfills'` in `setupTests.ts`.

src/setupTests.ts
vm: gonna do in separate PR so I don't need another approval on this before merging

PR #6554, opened 2023-05-12
chore: remove Google Analytics

vm:
## Description
Removes Google Analytics as we no longer need to send analytics there.

### QA (ie manual testing)
- Make sure external links continue to work
- Check the content security policy with security

### Automated testing
- [x] Unit test (N/A)
- [ ] Integration/E2E test

This will be checked off once this PR is merged: https://github.com/Uniswap/interface/pull/6580


src/theme/components/index.tsx
```diff
@@ -166,25 +164,6 @@ export const UniTokenAnimated = styled.img`
...
-  // don't prevent default, don't redirect if it's a new tab
-  if (target === '_blank' || event.ctrlKey || event.metaKey) {
-    outboundLink({ label: anonymizedHref }, () => {
-      console.debug('Fired outbound link event', anonymizedHref)
-    })
-  } else {
-    event.preventDefault()
-    // send a ReactGA event and then trigger a location change
-    outboundLink({ label: anonymizedHref }, () => {
-      window.location.href = anonymizedHref
```
AndrewMohawk, line 183: Wont this prevent external link from opening?

public/index.html
```diff
@@ -19,9 +19,9 @@
     <meta
       http-equiv="Content-Security-Policy"
       <% if (process.env.REACT_APP_CSP_ALLOW_UNSAFE_EVAL) { %>
-        content="script-src 'self' https://www.google-analytics.com https://www.googletagmanager.com 'unsafe-inline' 'unsafe-eval'"
+        content="script-src 'self' https://www.googletagmanager.com 'unsafe-inline' 'unsafe-eval'"
```
zzmp, line 22: Why is `googletagmanager` still necessary? What is it used for?

public/index.html
```diff
@@ -37,8 +37,6 @@
     -->
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
 
-    <link rel="preconnect" href="https://www.google-analytics.com/" />
```
zzmp, line 40: We should probably preconnect to our existing analytics servers. Can you ticket that? No need to change anything in _this_ PR.

src/components/analytics/index.ts
```diff
@@ -1,94 +0,0 @@
...
-
-const storedClientId = window.localStorage.getItem(GOOGLE_ANALYTICS_CLIENT_ID_STORAGE_KEY)
-
-const googleAnalytics = new GoogleAnalyticsProvider()
-
-export function sendEvent(event: string | UaEventOptions, params?: any) {
-  return googleAnalytics.sendEvent(event, params)
-}
-
-export function outboundLink(
```
zzmp, line 21: Do we want to capture outboundLink's through Amplitude?

src/components/analytics/index.ts
```diff
@@ -1,94 +0,0 @@
...
-} else {
-  googleAnalytics.initialize('test', { gtagOptions: { debug_mode: true } })
-}
-
-const installed = Boolean(window.navigator.serviceWorker?.controller)
-const hit = Boolean((window as any).__isDocumentCached)
-const action = installed ? (hit ? 'Cache hit' : 'Cache miss') : 'Not installed'
-sendEvent({ category: 'Service Worker', action, nonInteraction: true })
-
-function reportWebVitals({ name, delta, id }: Metric) {
```
zzmp, line 61: Are we capturing web vitals through amplitude?

zzmp (approved):
I'm concerned that we're losing some analytics (outbound links and web vitals), and want to make sure we have a plan in place to recapture them via Amplitude if they're still wanted. Otherwise, lgtm!

src/components/analytics/index.ts
zzmp: Could you ticket this, linking this PR, and then reach out to product to see if they still want this as a metric?
In case they do, make sure this PR is linked to from the ticket, so we can recover the relevant code without re-implementing (eg anonymizeLink, outboundLink).

src/components/analytics/index.ts
zzmp: I think this is already captured by Amplitude. Could you just verify?

cmcewen (changes requested):
if you want to merge this more quickly - can you make `sendEvent` a noop so that we don't lose the analytics events? i think it will be easier just to instrument or delete if they are dupes vs redo all of this in the future

cmcewen (changes requested):
if you want to merge this more quickly - can you make `sendEvent` a noop so that we don't lose the analytics events? i think it will be easier just to instrument or delete if they are dupes vs redo all of this in the future

src/theme/components/index.tsx
```diff
@@ -166,25 +164,6 @@ export const UniTokenAnimated = styled.img`
...
 `
 
-function handleClickExternalLink(event: React.MouseEvent<HTMLAnchorElement>) {
-  const { target, href } = event.currentTarget
-
-  const anonymizedHref = anonymizeLink(href)
-
-  // don't prevent default, don't redirect if it's a new tab
-  if (target === '_blank' || event.ctrlKey || event.metaKey) {
-    outboundLink({ label: anonymizedHref }, () => {
```
blairmason, line 176: We should send this to Amplitude

src/components/analytics/index.ts
just-toby: yup confirmed

PR #6556, opened 2023-05-12
ci: Final CI fixes pass

lavalamp-:
## Description

This PR introduces the following fixes to the deployment CI pipeline:

- Truncates Slack merge messages to 3,000 characters (API limit)
- Emits Slack messages for (1) a build starting, (2) a deploy starting, and (3) a deploy completing
- Updates Cloudflare pages deployments to specify `main` branch

### DIFF
--- a/.github/workflows/2-deploy-to-staging.yml
+++ b/.github/workflows/2-deploy-to-staging.yml
@@ -39,6 +39,8 @@ jobs:
           projectName: interface-staging
           directory: build
           githubToken: ${{ secrets.GITHUB_TOKEN }}
+          # Cloudflare uses `main` as the default production branch, so we push using the `main` branch so that it can be aliased by a custom domain.
+          branch: main
       - name: Send Slack message that deploy has failed
         uses: slackapi/slack-github-action@007b2c3c751a190b6f0f040e47ed024deaa72844
         if: always() && steps.pages-deployment.outcome == 'failure'
--- a/.github/workflows/2-deploy-to-staging.yml
+++ b/.github/workflows/2-deploy-to-staging.yml
@@ -43,21 +43,11 @@ jobs:
           branch: main
       - name: Send Slack message that deploy has failed
         uses: slackapi/slack-github-action@007b2c3c751a190b6f0f040e47ed024deaa72844
-        if: always() && steps.pages-deployment.outcome == 'failure'
+        if: always()
         with:
           payload: |
             {
-              "text": "Staging Cloudflare pages deploy **FAILED** for: ${{ github.ref_name }}"
-            }
-        env:
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
-          SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
-      - name: Send Slack message that deploy has completed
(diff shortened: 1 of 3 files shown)


.github/workflows/2-deploy-to-staging.yml
```diff
@@ -13,13 +13,35 @@ jobs:
...
+        id: slack
+        uses: slackapi/slack-github-action@007b2c3c751a190b6f0f040e47ed024deaa72844
+        with:
+          payload: |
+            {
+              "text": "Starting staging Cloudflare pages deploy for branch: ${{ github.ref_name }}"
+            }
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+          SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
```
zzmp, line 44: Rather than send a slack message at start/deploy/deployed, I think it'd be more useful to send a slack message at start/success/fail.

I also worry this will introduce message fatigue, so an alternative would be just success/fail.

zzmp (changes requested):

zzmp (approved):

.github/workflows/4-deploy-to-prod.yml
```diff
@@ -78,6 +89,27 @@ jobs:
           githubToken: ${{ secrets.GITHUB_TOKEN }}
           branch: main
```
zzmp, line 90: We need to do this (the same thing, `branch: main`) in `2-deploy-to-staging.yml`, or else our staging deploy will not go out to the "production" staging domain.

zzmp (changes requested):

.github/workflows/4-deploy-to-prod.yml
zzmp: Also, could you comment both settings of `branch`, with

    # Cloudflare uses `main` as the default production branch, so we push using the `main` branch so that it can be aliased by a custom domain.

.github/workflows/2-deploy-to-staging.yml
```diff
@@ -21,13 +31,37 @@ jobs:
...
+      - name: Send Slack message that deploy has completed
+        uses: slackapi/slack-github-action@007b2c3c751a190b6f0f040e47ed024deaa72844
+        with:
+          payload: |
+            {
+              "text": "Staging Cloudflare pages deploy completed for: ${{ github.ref_name }}"
+            }
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+          SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
```
zzmp, line 64: What do you think of simplifying this to:

```suggestion
        if: always()
        with:
          payload: |
            {
              "text": "Staging deploy **${{ steps.pages-deployment.outcome }}** for: ${{ github.ref_name }}"
            }
        env:
          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
          SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
```

zzmp (approved):

grabbou (approved):

Issue #6557, opened 2023-05-12
MobileBottomBar too short on iPhone PWA


Jdecristi:
The MobileBottomBar is too short and overlaps with native iPhone bar

1. Go to  app.uniseap.org on safari
2. Click on share site
3. Click on "Add to Home Screen
4. Open App

### Expected Behavior
There should be a space below the MobileBottomBar items so the it doesn't overlap with native iPhone bar

### Screen Shots
<details>
<summary>Current</summary>
<br />

<img width="437" alt="Screenshot 2023-05-12 at 3 53 56 PM" src="https://github.com/Uniswap/interface/assets/89173284/8dec1494-5328-417f-9e9d-1b448a3fc59b">

</details>

<details>
<summary>Expected</summary>
<br />

<img width="437" alt="Screenshot 2023-05-12 at 3 53 30 PM" src="https://github.com/Uniswap/interface/assets/89173284/ca91c22c-34f5-4344-b154-8fe4bb4fcd92">

</details>

PR #6558, opened 2023-05-12
fix: Increase MobileBottomBar height to not overlap with iPhone bottom bar

Jdecristi:
## General Notes
Fixes: #6557 

Increased MobileBottomBar height to 70px

## Screen Shots
| Before  |  After PWA |  After Browser |
|:-------:|:-----------:|:--------------:|
| Screenshot 2023-05-12 at 3 53 56 PM | Screenshot 2023-05-12 at 3 53 30 PM  | Screenshot 2023-05-12 at 3 53 30 PM |

## Testing
1. Go to app.uniseap.org on safari
2. Click on share site
3. Click on "Add to Home Screen
4. Open App

#### Automated testing
- [x] Unit test N/A

### DIFF
--- a/src/theme/index.tsx
+++ b/src/theme/index.tsx
@@ -85,7 +85,7 @@ function getSettings(darkMode: boolean) {
     deprecated_mediaWidth: deprecated_mediaWidthTemplates,
 
     navHeight: 72,
-    mobileBottomBarHeight: 52,
+    mobileBottomBarHeight: window.matchMedia('(display-mode: standalone)').matches ? 76 : 52,
     maxWidth: MAX_CONTENT_WIDTH,
 
     // deprecated - please use hardcoded exported values instead of


src/theme/index.tsx
```diff
@@ -85,7 +85,7 @@ function getSettings(darkMode: boolean) {
     deprecated_mediaWidth: deprecated_mediaWidthTemplates,
 
     navHeight: 72,
-    mobileBottomBarHeight: 52,
+    mobileBottomBarHeight: 70,
```
JFrankfurt, line 88:
```suggestion
    mobileBottomBarHeight: 68,
```

`72` would also work here

It's not documented well, but we try to keep all px values to increments of `4`

JFrankfurt (changes requested):

src/theme/index.tsx
infredible: Jordan's right about us trying to use a 4px base. But this should be even taller @ `76` to have more spacing between the tab labels and the home screen bar.

infredible:
Thanks for catching this @Jdecristi . Other than my comment above, your fix looks good when the app is added to the home screen, but we don't want this height when the app is viewed directly in mobile safari. 

This is what my chat with GPT said:

> Yes, you can change the height of an element depending on whether it is being viewed in the Mobile Safari app or a PWA. Here's how you can achieve this using JavaScript:
> 
> Detecting the PWA environment: In your JavaScript code, you can check if the current environment is a PWA by checking the window.matchMedia function. PWAs usually have a specific media query that can be used for detection. Here's an example:
> ```
> const isPWA = window.matchMedia('(display-mode: standalone)').matches;
> ```
> The isPWA variable will be true if the current environment is a PWA and false otherwise.
> 
> Changing the height of the element: Once you have determined the environment, you can use JavaScript to change the height of the element accordingly. Here's an example:
> 
> ```
> const element = document.getElementById('your-element-id');
> if (isPWA) {
>   // Set height for PWA
>   element.style.height = '300px';
> } else {
>   // Set height for Mobile Safari
>   element.style.height = '200px';
> }
> ```
> In this example, your-element-id should be replaced with the actual ID of the element you want to modify. The if-else statement sets different heights for the element based on whether it is viewed in a PWA or Mobile Safari.
> 
> By combining these steps, you can conditionally change the height of an element depending on whether it is being viewed in the Mobile Safari app or a PWA.

Jdecristi:
@infredible @JFrankfurt I fixed the issues you brought up. I changed the height from 70px to 76px and added a ternary to check if the app is PWA or not

kristiehuang:
hey - looks like this got stamped but didn't get merged! Closing due to age but feel free to resolve merge conflicts & reopen.

Issue #6563, opened 2023-05-14
"Minimum output" should be "Maximum output" when trade is Exact Output


Jdecristi:
**Bug Description**
When making and exact output trade the "Minimum output" row is showing the maximum input value

**Steps to Reproduce**

1. Go to app.uniswap.org
2. Make a trade with an exact output
3. "Minimum output" row is showing the maximum input value

**Expected Behavior**
"Minimum Output" should be "Maximum Output" when trade is Exact Output

**Additional Context**
<p align="center">
<img width="500" alt="Screenshot " src="https://github.com/Uniswap/interface/assets/89173284/045ec7ef-e712-4d0d-b2d6-b3f02dfe6dda">
</p>

PR #6565, opened 2023-05-14
fix: "Minimum output" should be "Maximum output" when trade is Exact Output

Jdecristi:
Fixes: #6563 

## Description
Simple fix, just checking the trade.type same conditions as the value on line 84

## Screen capture

| Before  |  After  |
|:-------:|:------:|
| Screenshot | Screenshot |

### Reproducing the error

1. Go to app.uniswap.org
2. Make a trade with an exact output
3. "Minimum output" row is showing the maximum input value

### QA (ie manual testing)

- [ ] Make sure "Maximum input" when

### Automated testing
- [x] Unit test N/A

### DIFF
--- a/src/components/swap/AdvancedSwapDetails.tsx
+++ b/src/components/swap/AdvancedSwapDetails.tsx
@@ -75,7 +75,7 @@ export function AdvancedSwapDetails({ trade, allowedSlippage, syncing = false }:
             }
           >
             <ThemedText.BodySmall color="textSecondary">
-              <Trans>Minimum output</Trans>
+              {trade.tradeType === TradeType.EXACT_INPUT ? <Trans>Minimum output</Trans> : <Trans>Maximum input</Trans>}
             </ThemedText.BodySmall>
           </MouseoverTooltip>
         </RowFixed>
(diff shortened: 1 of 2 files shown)


src/components/swap/AdvancedSwapDetails.tsx
```diff
@@ -75,7 +75,7 @@ export function AdvancedSwapDetails({ trade, allowedSlippage, syncing = false }:
             }
           >
             <ThemedText.BodySmall color="textSecondary">
-              <Trans>Minimum output</Trans>
+              <Trans>{trade.tradeType === TradeType.EXACT_INPUT ? 'Minimum output' : 'Maximum input'}</Trans>
```
tinaszheng, line 78: so i think our Translation library doesn't work super well when there's a ternary inside - can you amend to do the switch outside of the `<Trans>`? otherwise lgtm and thanks for this fix :)

tinaszheng (changes requested):

src/components/swap/AdvancedSwapDetails.tsx
Jdecristi: Fixed!

tinaszheng (approved):

JFrankfurt (approved):

Jdecristi:
@tinaszheng @JFrankfurt I didn't realize there was a test that needed to be fixed. It's fixed now, but I wasn't sure if I needed to let you know so you could re-review them. 

Thanks

tinaszheng (approved):

JFrankfurt:
Thanks @Jdecristi! We really appreciate the fix!

PR #6575, opened 2023-05-15
feat: add l2 chain logo to modal

just-toby:
## Description

adds the Layer 2 chain logo to the top left of the modal

_JIRA ticket:_
_Slack thread:_
_Relevant docs:_ https://www.figma.com/file/ksWAw6EZiiMuRyMQQHicoH/Permit2-Revisions?type=design&node-id=531-39584&t=6hrgUNPMrWRoeRHZ-0

## Screen capture

| image | image      | 

### QA (ie manual testing)

- [x] manually verified the l2 icon appears

### Automated testing

- [x] Unit test N/A


vm (dismissed):

src/components/TransactionConfirmationModal/ConfirmationModalContent.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { render, screen } from 'test-utils/render'
+
+import { ConfirmationModalContent } from '.'
+
+describe('ConfirmationModalContent', () => {
+  it('should render the L2 icon for optimism', () => {
+    render(
+      <ConfirmationModalContent
+        title="title"
+        onDismiss={jest.fn()}
```
grabbou, line 11: nit: use `noop` from test utils

src/components/TransactionConfirmationModal/ConfirmationModalContent.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+  it('should not render a chain icon', () => {
+    render(
+      <ConfirmationModalContent
+        title="title"
+        onDismiss={jest.fn()}
+        topContent={() => <div>topContent</div>}
+        bottomContent={() => <div>bottomContent</div>}
+      />
+    )
+    expect(screen.queryByTestId('confirmation-modal-chain-icon')).toBeNull()
```
grabbou, line 29: nit: `not.toBeInTheDocument()`

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -173,16 +173,27 @@ export function ConfirmationModalContent({
...
   topContent,
+  chainId,
 }: {
   title: ReactNode
   onDismiss: () => void
   topContent: () => ReactNode
   bottomContent?: () => ReactNode | undefined
+  chainId?: SupportedL2ChainId | undefined
 }) {
+  const info = getChainInfo(chainId)
```
grabbou, line 184: Why not just get `chainId` here and check if its L2? I think encapsulating that logic within this component makes a bit more sense because:
- we can check whether it renders a badge for L2 or not (which is good)
- it is less confusing - undefined `chainId` means to me that we may not be connected to any chain and it's not clearly visible from the following code that we want to render for L2 only

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -173,16 +173,27 @@ export function ConfirmationModalContent({
   bottomContent,
   onDismiss,
   topContent,
+  chainId,
 }: {
   title: ReactNode
   onDismiss: () => void
   topContent: () => ReactNode
   bottomContent?: () => ReactNode | undefined
+  chainId?: SupportedL2ChainId | undefined
```
grabbou, line 182: `| undefined` is unnecessary if you already have `?`

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -173,16 +173,27 @@ export function ConfirmationModalContent({
...
 }) {
+  const info = getChainInfo(chainId)
   return (
     <Wrapper>
       <AutoColumn gap="sm">
         <Row>
+          {chainId && (
```
grabbou, line 189: You should check for `{info &&}` instead of `chainId`, and then, inside the component, do not write `{info?.label}` but just `{info.label}`. We don't want any undefined.

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -173,16 +173,27 @@ export function ConfirmationModalContent({
...
+            <Badge>
+              <RowFixed data-testid="confirmation-modal-chain-icon">
+                <StyledLogo src={info?.logoUrl} style={{ margin: '0 8px 0 0' }} />
```
grabbou, line 192: Is this right margin? What is it needed for? Also, prefer Styled Components if absolutely necessary.

src/components/TransactionConfirmationModal/index.tsx
grabbou: Okay, I see, this may not work because of the fact we need modalState to decide whether we want to show or not. In that case, I would recommend you stick to `headerContent` pattern (function to render modal elements) and we just repeat that.

I would add a TODO to Linear that we need this to be refactored similarly how you refactored in your last PR from array/dynamic rendering to static.

Ideally, there should be a simple if conditional / switch statement, three different modals for each stage and each one having all the elements explicitly defined. 

If you have capacity this week, you can pick this work up. If not, feel free to assign it to me and I will work on it this week.

grabbou (changes requested):

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -173,16 +173,19 @@ export function ConfirmationModalContent({
...
   onDismiss: () => void
   topContent: () => ReactNode
   bottomContent?: () => ReactNode | undefined
+  headerContent?: () => ReactNode
 }) {
   return (
     <Wrapper>
       <AutoColumn gap="sm">
         <Row>
+          {headerContent && headerContent()}
```
grabbou, line 188: nit: headerContent?.()

src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -173,16 +173,19 @@ export function ConfirmationModalContent({
   bottomContent,
   onDismiss,
   topContent,
+  headerContent,
 }: {
   title: ReactNode
   onDismiss: () => void
   topContent: () => ReactNode
   bottomContent?: () => ReactNode | undefined
```
grabbou, line 181: nit: we can also remove `| undefined` here for clarity

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -261,6 +266,21 @@ export default function ConfirmSwapModal({
     startSwapFlow,
   ])
 
+  const l2Badge = useCallback(() => {
```
grabbou, line 269: nit: In this particular context, since this is not used in any other hook, `l2Badge` does not need `useCallback`

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -261,6 +266,21 @@ export default function ConfirmSwapModal({
...
+      return (
+        <Badge>
+          <RowFixed data-testid="confirmation-modal-chain-icon" gap="sm">
+            <StyledLogo src={info.logoUrl} />
+            {info.label}
+          </RowFixed>
+        </Badge>
+      )
+    }
+    return undefined
```
grabbou, line 281: nit: this is not needed, `return undefined` is the default return value and I believe TypeScript will infer that automatically too

grabbou (approved):
LGTM. Some non-blocking nits, but it looks good

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -169,6 +173,7 @@ export default function ConfirmSwapModal({
   fiatValueInput: { data?: number; isLoading: boolean }
   fiatValueOutput: { data?: number; isLoading: boolean }
 }) {
+  const { chainId } = useWeb3React()
```
tinaszheng, line 176: is it accurate to use the wallet's chainId here? doesnt the swap flow have its own internal chainId in case there's a difference between the two?

tinaszheng (approved):

src/components/swap/ConfirmSwapModal.tsx
just-toby: yes, but the user shouldn't be able to reach this part of the flow with these 2 values being unsynced. i think i can switch to use the contextual chainId without an change in behavior though.

src/components/swap/ConfirmSwapModal.tsx
just-toby: this is needed to silence a linter complaint that "not all code paths return a value"

zzmp: `return` has type `void`, whereas `return undefined` has type `undefined`.

zzmp (approved):

PR #6576, opened 2023-05-15
feat: swap rejection error handling

just-toby:
## Description

fixes the handling of user rejection in the swap submission flow at the actual Swap step. if the user rejects the actual swap transaction, we should revert to the Review step

### QA (ie manual testing)

- [x] reject swap in wallet and verify that the modal reverts to the review step instead of showing an error

### Automated testing

- [x] Unit test N/A
- [x] Integration/E2E test


src/hooks/useUniversalRouter.ts
```diff
@@ -116,7 +116,10 @@ export function useUniversalRouterSwapCallback(
         // GasEstimationErrors are already traced when they are thrown.
         if (!(swapError instanceof GasEstimationError)) setTraceError(swapError)
 
-        throw new Error(swapErrorToUserReadableMessage(swapError))
+        const rejectError = new Error(swapErrorToUserReadableMessage(swapError))
+        // This allows us to distinguish between user rejections and other errors later too.
+        ;(rejectError as any).code = 4001
```
zzmp, line 121: Where is this used?

It would be cleaner to instead create a `class UserRejectionError extends Error` and then check for `error instanceof UserRejectionError`.

src/components/swap/SwapModalFooter.tsx
```diff
@@ -300,8 +298,6 @@ export default function SwapModalFooter({
               </ThemedText.HeadlineSmall>
             </ConfirmButton>
           </TraceEvent>
-
```
zzmp, line 303: Should this only be removed in the case of _user rejection_ error?

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -124,6 +127,13 @@ export default function ConfirmSwapModal({
     }
   }, [allowance, previousPermitNeeded])
 
+  const swapFailed = useMemo(() => Boolean(swapError) && !didUserReject(swapError), [swapError])
+  useEffect(() => {
+    if (swapError && !swapFailed && ConfirmModalState.PENDING_CONFIRMATION === confirmModalState) {
+      setConfirmModalState(ConfirmModalState.REVIEWING)
+    }
+  }, [confirmModalState, swapError, swapFailed])
```
zzmp, line 135: Can this be put somewhere where it is triggered imperatively instead of as part of a `useEffect`?

zzmp (changes requested):

src/components/swap/ConfirmSwapModal.tsx
just-toby: yes, updated

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -187,6 +187,14 @@ export default function ConfirmSwapModal({
     doesTradeDiffer: Boolean(doesTradeDiffer),
   })
 
+  const swapFailed = Boolean(swapError) && !didUserReject(swapError)
+  useEffect(() => {
+    // Reset the modal state if the user rejected the swap.
+    if (swapError && !swapFailed) {
+      onCancel()
+    }
+  }, [onCancel, swapError, swapFailed])
```
just-toby, line 196: we need to use `useEffect` here to reset the ConfirmModalState since the state is local and the rejection happens in `useUniversalRouterSwapCallback`. maybe a future improvement could be made here if we start using a Context to manage Swap state that all children can access equally?

src/utils/errors.ts
```diff
@@ -0,0 +1,6 @@
+export class UserRejectedRequestError extends Error {
+  constructor(message: string) {
+    super(message)
+    this.name = 'UserRejectedRequestError'
+  }
+}
```
JFrankfurt, line 6: Can you add some documentation explaining when this error should be thrown?

JFrankfurt (approved):

zzmp (approved):

zzmp (approved):

zzmp (approved):

PR #6578, opened 2023-05-15
fix: rm unused const

zzmp:
## Description

Removes unused `DataPagePaddingContainer`, which was triggering an ESLint failure.

### Automated testing

- [x] ESLint


JackShort (approved):

JackShort:
Is there any way to make this trigger an error and not a warning? Because it passes the github action lint request. But going back to the PR I see that it was raised as a warning

cbachmeier (changes requested):
we actually need this container and I believe it became unused as part of a merge conflict resolution, should be used as shown here: https://github.com/Uniswap/interface/pull/6579

PR #6580, opened 2023-05-16
feat: only pre-cache the document

zzmp:
## Description

Updates the ServiceWorker (SW) strategy to only precache the document (index.html).

This significantly decreases the time to activate the SW, making it (1) more likely to be used on a fast-refresh and (2) faster to run the e2e test on. It is acceptable to move the JS/CSS chunks from precache to on-demand,because if they fail to load due to version skew (ie because the document is outdated), the whole page should be reloaded - their individual failure is acceptable.

- Adds analytics for the SW to amplitude. These were previously only implemented for Google Analytics (GA).

- Updates the SW e2e test by waiting for events from amplitude to be sent, instead of GA. The test is also de-flaked because of the much smaller precache, which was taking long enough to load that the tests had been flaky.

- Installs the SW in a staging environment.

### Automated testing

- [x] Integration/E2E test


src/serviceWorker/utils.ts
```diff
@@ -17,13 +17,21 @@ export function isDevelopment() {
...
   return entries.reduce<GroupedEntries>(
     ({ onDemandEntries, precacheEntries }, entry) => {
       if (typeof entry === 'string' || entry.url.includes('/media/')) {
         return { precacheEntries, onDemandEntries: [...onDemandEntries, entry] }
-      } else {
+      } else if (entry.revision) {
+        // index.html should be the only non-media entry with a revision, as code chunks have a hashed URL.
         return { precacheEntries: [...precacheEntries, entry], onDemandEntries }
+      } else {
+        return { precacheEntries, onDemandEntries: [...onDemandEntries, entry] }
```
vm, line 34: you should cover this in the utils test since it changes the logic

src/serviceWorker/index.ts
```diff
@@ -25,8 +25,8 @@ registerRoute(
   new Route(
     ({ url }) => onDemandURLs.includes('.' + url.pathname),
     new CacheFirst({
-      cacheName: 'media',
-      plugins: [new ExpirationPlugin({ maxEntries: 16 })],
+      cacheName: 'on-demand',
```
vm, line 28: good idea. btw i found out this won't delete the cache automatically so maybe we figure that out now?

cypress/e2e/service-worker.test.ts
```diff
@@ -20,66 +20,59 @@ describe('Service Worker', () => {
...
-      // This is emitted after caching the entry file, which takes some time to load.
-      .wait('@NotInstalled', { timeout: 60000 })
-      .window()
-      .and((win) => {
-        expect(win.navigator.serviceWorker.controller?.state).to.equal('activated')
-      })
+    cy.wait('@ServiceWorker:Uninstalled')
+    cy.window().should(
+      'have.nested.property',
+      // The parent is checked instead of the AUT because it is on the same origin,
```
vm, line 48: what's an AUT?

zzmp: application under test - it's testing-specific terminology but it is used in the Cypress typings.

vm (approved):

cypress/e2e/service-worker.test.ts
```diff
@@ -30,19 +30,18 @@ describe('Service Worker', () => {
...
-      if (body.match(/"service_worker":"uninstalled"/)) {
-        req.alias = 'ServiceWorker:Uninstalled'
-      } else if (body.match(/"service_worker":"hit"/)) {
-        req.alias = 'ServiceWorker:Hit'
-      } else if (body.match(/"service_worker":"miss"/)) {
-        req.alias = 'ServiceWorker:Miss'
+      const serviceWorkerStatus = body.match(/"service_worker":"(\w+)"/)?.[1]
+      if (serviceWorkerStatus) {
+        const alias = `ServiceWorker:${serviceWorkerStatus}`
+        console.log(alias)
```
vm, line 36: rm

src/serviceWorker/index.ts
```diff
@@ -33,5 +37,5 @@ precacheAndRoute(precacheEntries) // precache cache
 // We only use the precache and runtime caches, so we delete the rest to avoid taking space.
 // Wait to do so until 'activate' in case activation fails.
 self.addEventListener('activate', () =>
-  deleteUnusedCaches(self.caches, { usedCaches: [cacheNames.runtime, cacheNames.precache] })
+  deleteUnusedCaches(self.caches, { usedCaches: [cacheNames.precache, onDemandCacheName] })
```
vm, line 40: should this be `preCache`?

vm (approved):

src/serviceWorker/index.ts
zzmp: no, this comes from `workbox-core`, and is the cache automatically used by `precacheAndRoute`.
https://developer.chrome.com/docs/workbox/reference/workbox-core/#property-cacheNames

src/serviceWorker/index.ts
```diff
@@ -21,14 +21,21 @@ registerRoute(new DocumentRoute())
...
 
+const onDemandCacheName = `${cacheNames.prefix}-on-demand-${cacheNames.suffix}`
 registerRoute(
   new Route(
     ({ url }) => onDemandURLs.includes('.' + url.pathname),
     new CacheFirst({
-      cacheName: 'media',
-      plugins: [new ExpirationPlugin({ maxEntries: 16 })],
+      cacheName: onDemandCacheName,
+      plugins: [new ExpirationPlugin({ maxEntries: 64 })],
```
cmcewen, line 30: how did you pick this number?

cmcewen (approved):

src/serviceWorker/index.ts
zzmp: I navigated to all tabs on the navbar, and then checked how many resources had been cached: 33.
I then rounded up to the nearest power of 2: 64.

PR #6582, opened 2023-05-16
feat(wallet-connect): add support for v2

JFrankfurt:
## Description
Adds support for Wallet Connect v2. 

https://docs.walletconnect.com/2.0/advanced/migration-from-v1.x/overview

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2048/wc-20-web-ui

### Before 
image

### After (Desktop) 
image

### After (Mobile) 
image    |

### QA (ie manual testing)

- [ ] test connecting a wallet app via mobile browser (deep link)
- [ ] test connecting a wallet app via desktop browser (qr code)
- [ ] confirm interactions listed in testing doc: https://www.notion.so/uniswaplabs/Uniswap-WalletConnect-v2-Test-Instructions-External-4b1176c29f9741cf8fcab77880b4878a?pvs=4

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test


src/components/NavBar/ChainSelector.tsx
```diff
@@ -29,6 +30,10 @@ const NETWORK_SELECTOR_CHAINS = [
   SupportedChainId.BNB,
 ]
 
+if (!isProductionEnv()) {
+  NETWORK_SELECTOR_CHAINS.push(SupportedChainId.GOERLI)
+}
```
JFrankfurt, line 35: if #6610 lands first we should swap this out for Sepolia

src/components/WalletModal/Option.tsx
```diff
@@ -17,26 +24,17 @@ const OptionCardLeft = styled.div`
...
   transition: ${({ theme }) => theme.transition.duration.fast};
-  opacity: ${({ disabled, selected }) => (disabled && !selected ? '0.5' : '1')};
-
-  &:hover {
-    cursor: ${({ disabled }) => !disabled && 'pointer'};
-    background-color: ${({ theme, disabled }) => !disabled && theme.hoverState};
-  }
-  &:focus {
-    background-color: ${({ theme, disabled }) => !disabled && theme.hoverState};
-  }
```
JFrankfurt, line 39: 2 things happening here:

1. I alphabetized some scattered properties
2. I removed the hover/focus rules and bumped them up to a parent element so the new dom structure would retain the existing behavior.

src/components/WalletModal/Option.tsx
```diff
@@ -62,38 +60,163 @@ const IconWrapper = styled.div`
...
+  padding: 0;
+  position: absolute;
+  right: 16px;
+  top: calc(50% - 12px);
+  width: 24px;
+
+  &:hover {
+    opacity: 0.6;
+  }
+`
```
JFrankfurt, line 94: this is an intentionally non-standard button approved by @infredible

src/components/WalletModal/Option.tsx
```diff
@@ -62,38 +60,163 @@ const IconWrapper = styled.div`
...
   const { activationState, tryActivation } = useActivationState()
-  const closeDrawer = useCloseAccountDrawer()
-  const activate = () => tryActivation(connection, closeDrawer)
+  const [WC2PromptOpen, setWC2PromptOpen] = useState(false)
+  const [accountDrawerOpen, toggleAccountDrawerOpen] = useAccountDrawer()
+  const activate = () => tryActivation(connection, toggleAccountDrawerOpen)
+
+  useEffect(() => {
+    if (!accountDrawerOpen) setWC2PromptOpen(false)
+  }, [accountDrawerOpen])
```
JFrankfurt, line 164: close the wc2 prompt when the account drawer parent is closed

src/components/WalletModal/Option.tsx
```diff
@@ -62,38 +60,163 @@ const IconWrapper = styled.div`
...
 
+  const handleClickConnectViaWCv2 = (e: MouseEvent<HTMLButtonElement>) => {
+    e.stopPropagation()
+    tryActivation(walletConnectV2Connection, () => {
+      setWC2PromptOpen(false)
+      toggleAccountDrawerOpen()
+    })
+  }
+  const handleClickOpenWCv2Tooltip = (e: MouseEvent<HTMLButtonElement>) => {
+    e.stopPropagation()
```
JFrankfurt, line 178: `stopPropagation` is necessary here to to prevent triggering parent handlers

src/connection/WalletConnectV2.ts
```diff
@@ -0,0 +1,62 @@
...
+          },
+        },
+      },
+      onError,
+    })
+  }
+
+  activate(chainId?: number) {
+    sendAnalyticsEvent(this.ANALYTICS_EVENT)
+    return super.activate(chainId)
```
JFrankfurt, line 60: this class will eventually replace the WalletConnect class (and therefore be the new underlying class for UniWallet)

src/connection/utils.ts
```diff
@@ -32,6 +33,7 @@ export enum ErrorCode {
 export function didUserReject(connection: Connection, error: any): boolean {
   return (
     error?.code === ErrorCode.USER_REJECTED_REQUEST ||
+    (connection.type === ConnectionType.WALLET_CONNECT_V2 && error?.toString?.() === ErrorCode.WC_V2_MODAL_CLOSED) ||
```
JFrankfurt, line 36: this catches the error thrown when a user declines to connect scan the wc modal

src/pages/Swap/index.tsx
```diff
@@ -27,13 +27,11 @@ import { useSwapCallback } from 'hooks/useSwapCallback'
...
-import { useCallback, useEffect, useMemo, useReducer, useState } from 'react'
-import { ReactNode } from 'react'
+import { ReactNode, useCallback, useEffect, useMemo, useReducer, useState } from 'react'
 import { ArrowDown } from 'react-feather'
 import { useNavigate } from 'react-router-dom'
 import { Text } from 'rebass'
-import { InterfaceTrade } from 'state/routing/types'
-import { TradeState } from 'state/routing/types'
+import { InterfaceTrade, TradeState } from 'state/routing/types'
 import styled, { useTheme } from 'styled-components/macro'
```
JFrankfurt, line 35: I did some work in this file that I ended up not including. I decided to leave these import consolidations since they're helpful on their own.

src/components/NavBar/ChainSelector.tsx
cmcewen: just approved!

src/components/NavBar/ChainSelector.tsx
```diff
@@ -29,12 +31,42 @@ const NETWORK_SELECTOR_CHAINS = [
...
 interface ChainSelectorProps {
   leftAlign?: boolean
 }
 
+// accounts is an array of strings in the format of "eip155:<chainId>:<address>"
+function getChainsFromEIP155Accounts(accounts?: string[]): SupportedChainId[] {
+  if (!accounts) return []
+  return accounts.map((account) => {
+    const splitAccount = account.split(':')
+    return parseInt(splitAccount[1])
```
cmcewen, line 47: i'm surprised TS lets you do this - console tells me `parseInt(undefined) = NaN`

src/components/NavBar/ChainSelector.tsx
```diff
@@ -29,12 +31,42 @@ const NETWORK_SELECTOR_CHAINS = [
...
+  const { connector } = useWeb3React()
+
+  const getConnection = useGetConnection()
+
+  const connectionType = getConnection(connector).type
+
+  switch (connectionType) {
+    case ConnectionType.WALLET_CONNECT_V2:
+      return getChainsFromEIP155Accounts((connector as WalletConnect).provider?.session?.namespaces.eip155.accounts)
+    case ConnectionType.UNIWALLET:
```
cmcewen, line 61: nit but i think this should be UNISWAP_WALLET

src/components/NavBar/ChainSelector.tsx
```diff
@@ -29,12 +31,42 @@ const NETWORK_SELECTOR_CHAINS = [
...
+// accounts is an array of strings in the format of "eip155:<chainId>:<address>"
+function getChainsFromEIP155Accounts(accounts?: string[]): SupportedChainId[] {
+  if (!accounts) return []
+  return accounts.map((account) => {
+    const splitAccount = account.split(':')
+    return parseInt(splitAccount[1])
+  })
+}
+
+function useWalletSupportsChain() {
```
cmcewen, line 51: this returns an array of supported chains so would vote `useWalletSupportedChains` - had to check the type below

src/components/NavBar/ChainSelector.tsx
```diff
@@ -74,13 +104,13 @@ export const ChainSelector = ({ leftAlign }: ChainSelectorProps) => {
   const dropdown = (
     <NavDropdown top="56" left={leftAlign ? '0' : 'auto'} right={leftAlign ? 'auto' : '0'} ref={modalRef}>
       <Column paddingX="8">
-        {NETWORK_SELECTOR_CHAINS.map((chainId: SupportedChainId) => (
+        {NETWORK_SELECTOR_CHAINS.map((selectorChain: SupportedChainId) => (
           <ChainSelectorRow
-            disabled={isUniWallet && !UniWalletSupportedChains.includes(chainId)}
+            disabled={!walletSupportsChain.includes(selectorChain)}
```
cmcewen, line 109: i would think prettier would do `!(walletSupportsChain.includes(selectorChain))` - or at least that is how it reads in my head

src/components/WalletModal/Option.tsx
cmcewen: thanks for this comment!

src/components/WalletModal/Option.tsx
```diff
@@ -62,38 +60,163 @@ const IconWrapper = styled.div`
...
+const WCv2PopoverContent = styled(ThemeButton)`
+  background: ${({ theme }) => theme.backgroundSurface};
+  border: 1px solid ${({ theme }) => theme.backgroundOutline};
+  border-radius: 12px;
+  cursor: pointer;
+  display: flex;
+  max-width: 240px;
+  padding: 16px;
+  position: absolute;
+  z-index: ${Z_INDEX.popover};
```
cmcewen, line 72: 👨‍🍳 💋

src/components/WalletModal/Option.tsx
```diff
@@ -62,38 +60,163 @@ const IconWrapper = styled.div`
...
+  cursor: pointer;
+  display: flex;
+  height: 24px;
+  justify-content: center;
+  margin: 0;
+  max-width: 48px;
+  padding: 0;
+  position: absolute;
+  right: 16px;
+  top: calc(50% - 12px);
```
cmcewen, line 88: does this 12px relate to any of the other properties? if so i like to math them together with a variable so changing one changes the other

src/components/WalletModal/Option.tsx
```diff
@@ -62,38 +60,163 @@ const IconWrapper = styled.div`
...
+  show: boolean
+  onClick: (e: MouseEvent<HTMLButtonElement>) => void
+  onClose: () => void
+}
+function PopupButtonContent({ connection, isDarkMode, show, children, onClick, onClose }: PopupButtonContentProps) {
+  const popoverElement = useRef<HTMLButtonElement>(null)
+  useOnClickOutside(popoverElement, onClose)
+  return (
+    <>
+      {children}
```
cmcewen, line 127: it's not clear to me why you need to pass children here - is it just to save a fragment? it looks like you could do this side by side in the parent too

src/components/WalletModal/Option.tsx
```diff
@@ -62,38 +60,163 @@ const IconWrapper = styled.div`
...
+      {show && (
+        <WCv2PopoverContent
+          onClick={onClick}
+          style={{ top: 52, right: 10 }}
```
cmcewen, line 131: do we still have a no inline styles practice?

src/components/WalletModal/Option.tsx
```diff
@@ -62,38 +60,163 @@ const IconWrapper = styled.div`
...
+          ref={popoverElement}
+          size={ButtonSize.small}
+          emphasis={ButtonEmphasis.medium}
+        >
+          <IconWrapper>
+            <img src={connection.getIcon?.(isDarkMode)} alt="Icon" style={{ height: '20px', width: '20px' }} />
```
cmcewen, line 137: same on inline styles - and should the alt match the connection name?

src/connection/WalletConnectV2.ts
```diff
@@ -0,0 +1,50 @@
+import { sendAnalyticsEvent } from '@uniswap/analytics'
+import { WalletConnect, WalletConnectConstructorArgs } from '@web3-react/walletconnect-v2'
+import { SupportedChainId } from 'constants/chains'
+import { Z_INDEX } from 'theme/zIndex'
+
+import { RPC_URLS } from '../constants/networks'
+
+// Avoid testing for the best URL by only passing a single URL per chain.
+// Otherwise, WC will not initialize until all URLs have been tested (see getBestUrl in web3-react).
```
cmcewen, line 9: was this a web3-react or wc decision? this seems not ideal to me (cc @grabbou)

src/components/NavBar/ChainSelector.tsx
JFrankfurt: good catch. This is technically safe is we assume WC implementations are stable, but the types as written shouldn't allow this imo.

src/components/Button/index.tsx
```diff
@@ -524,15 +525,17 @@ const BaseThemeButton = styled.button<BaseThemeButtonProps>`
...
   return (
-    <BaseThemeButton {...rest}>
+    <BaseThemeButton {...rest} ref={ref}>
       <ButtonOverlay />
       {children}
     </BaseThemeButton>
   )
-}
+})
+ThemeButton.displayName = 'ThemeButton'
```
natew, line 538: in general react is deprecating this and encouraging just `function ThemeButton` on the inner function, but idk if our eslint maybe doesnt like that

src/components/WalletModal/Option.tsx
```diff
@@ -62,38 +60,163 @@ const IconWrapper = styled.div`
...
+    background-color: ${({ theme, disabled }) => !disabled && theme.hoverState};
+  }
+  &:focus {
+    background-color: ${({ theme, disabled }) => !disabled && theme.hoverState};
+  }
+`
 
-export default function Option({ connection }: { connection: Connection }) {
+interface PopupButtonContentProps {
+  children: JSX.Element
```
natew, line 115: React.ReactNode considered best practice here i think

src/components/Button/index.tsx
JFrankfurt: Yeah I never use displayName, but eslint complains about this when ref forwarding.. 🤔

natew (approved):
Beyond @cmcewen comments and minor nits looks good and worked for me without a hitch

src/components/NavBar/ChainSelector.tsx
JFrankfurt: I like that, but prettier won't allow parens around single bools

src/components/WalletModal/Option.tsx
JFrankfurt: Here's what that looks like. I like the principle, but in this case it feels a bit cludgy to me. 
```
const TOGGLE_SIZE = 24
const WCv2PopoverToggle = styled.button`
//...
  height: ${TOGGLE_SIZE}px;
//...
  top: calc(50% - ${TOGGLE_SIZE / 2}px);
  width: ${TOGGLE_SIZE}px;
//...
`
```

src/components/WalletModal/Option.tsx
JFrankfurt: thanks for catching this--used it during debugging

src/components/NavBar/ChainSelector.tsx
cmcewen: boo prettier

cmcewen (approved):
woot woot!

cmcewen (approved):

src/connection/WalletConnectV2.ts
JFrankfurt: `Promise.race()` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race

?

PR #6584, opened 2023-05-16
feat: special case arb search

cartcrom:
## Description

Special cases the ARB token on Arbitrum to always display over the ARB token on Mainnet in search results when both are returned by the backend, regardless of the user's current chain.

Also cleans up / renames the cross token search code to make functionality more clear.

_[Linear ticket](https://linear.app/uniswap/issue/WEB-1949/special-case-dollararb-in-search-so-the-arbitrum-network-one-shows-up)_

## Screen capture

| image |  image      |

### QA (ie manual testing)

- [ ] Search for a string that should return ARB as a search result, ensure it has the Arbitrum logo rather than no logo

### DIFF
--- a/src/graphql/data/SearchTokens.ts
+++ b/src/graphql/data/SearchTokens.ts
@@ -1,6 +1,7 @@
-import { WRAPPED_NATIVE_CURRENCY } from 'constants/tokens'
+import { ARB, WRAPPED_NATIVE_CURRENCY } from 'constants/tokens'
 import gql from 'graphql-tag'
 import { useMemo } from 'react'
+import invariant from 'tiny-invariant'
 
 import { Chain, SearchTokensQuery, useSearchTokensQuery } from './__generated__/types-and-hooks'
 import { chainIdToBackendName } from './util'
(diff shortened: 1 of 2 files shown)


cbachmeier (commented):
I'd like to avoid 1 offs whenever possible. MATIC is cross chain, what's the logic we use for this token to show it's native network first and can we re-use it for ARB?
Screen Shot 2023-05-16 at 15 07 11

cartcrom:
> I'd like to avoid 1 offs whenever possible. MATIC is cross chain, what's the logic we use for this token to show it's native network first and can we re-use it for ARB? Screen Shot 2023-05-16 at 15 07 11

I completely agree about 1 offs, and suggested using huge volume differences as a metric to override the default behaviour of showing the token on the wallet's current chain, but product wants to keep the scope of this change to just ARB for now. We can't re-use the MATIC logic, because that logic relies on Polygon MATIC being a native token

cbachmeier (changes requested):

cbachmeier (approved):
Returns expected result, however not sure if second check is necessary

src/graphql/data/SearchTokens.ts
```diff
@@ -41,19 +42,30 @@ gql`
...
-function isMoreRevelantToken(current: SearchToken, existing: SearchToken | undefined, searchChain: Chain) {
-  if (!existing) return true
+/* Returns the more relevant cross-chain token based on native status and search chain */
+function dedupeCrosschainTokens(current: SearchToken, existing: SearchToken | undefined, searchChain: Chain) {
+  if (!existing) return current
+  invariant(current.project?.id === existing.project?.id, 'Cannot dedupe tokens within different tokenProjects')
+
+  // Special case: always prefer Arbitrum ARB over Mainnet ARB
+  if (current.address?.toLowerCase() === ARB_ADDRESS) return current
+  if (existing.address?.toLowerCase() === ARB_ADDRESS) return existing
```
cbachmeier, line 56: Are you sure both of these are necessary? Having just the first check, `if (current.address?.toLowerCase() === ARB_ADDRESS) return current`, with the lowercase casting works for mainnet and arbitrum networks
Screen Shot 2023-05-18 at 13 42 25

cartcrom: It depends on the order the backend returns the two ARB tokens in, if the backend started returning them in the opposite order, only having one check wouldn't work.

src/graphql/data/SearchTokens.ts
```diff
@@ -87,7 +99,7 @@ export function useSearchTokens(searchQuery: string, chainId: number) {
     data?.searchTokens?.forEach((token) => {
       if (token.project?.id) {
         const existing = selectionMap[token.project.id]
-        if (isMoreRevelantToken(token, existing, searchChain)) selectionMap[token.project.id] = token
```
JFrankfurt, line 90: it looks like there's a behavior change here that might not be intentional--in the old one there were cases where `selectionMap[token.project.id]` wouldn't be assigned a value here, but now every iteration assigns a value. Is that intended?

cartcrom: In cases where `isMoreRevelantToken` would've returned false, `dedupeCrosschainTokens` returns `existing`, so this is just a refactor for readability

JFrankfurt (approved):

JFrankfurt:
smol

PR #6585, opened 2023-05-16
feat: [DetailsV2] Show data page header when nft scrolled out of view

cbachmeier:
## Description

- Adds an InteractionObserver to check whether the NFT has been scrolled out of view
  - Combined logic from these two examples: https://mtm.dev/sticky-stuck-styles & https://codepen.io/hey-nick/pen/mLpmMV
- Using the observer show/hide the data page header
- Data page header is sticky with a backdrop blur effect

## Screen capture

scrollHeader

### Automated testing

- [x] Updated snapshots

### DIFF
--- a/src/nft/components/details/detailsV2/DataPage.test.tsx
+++ b/src/nft/components/details/detailsV2/DataPage.test.tsx
@@ -4,6 +4,6 @@ import { render } from 'test-utils/render'
 import { DataPage } from './DataPage'
 
 it('placeholder containers load', () => {
-  const { asFragment } = render(<DataPage asset={TEST_NFT_ASSET} />)
+  const { asFragment } = render(<DataPage asset={TEST_NFT_ASSET} showDataHeader={true} />)
   expect(asFragment()).toMatchSnapshot()
 })
--- a/src/nft/components/details/detailsV2/DataPage.test.tsx
+++ b/src/nft/components/details/detailsV2/DataPage.test.tsx
@@ -3,7 +3,12 @@ import { render } from 'test-utils/render'
 
 import { DataPage } from './DataPage'
 
-it('placeholder containers load', () => {
+it('data page loads with header showing', () => {
   const { asFragment } = render(<DataPage asset={TEST_NFT_ASSET} showDataHeader={true} />)
+  expect(asFragment()).toMatchSnapshot()
+})
+
+it('data page loads without header showing', () => {
+  const { asFragment } = render(<DataPage asset={TEST_NFT_ASSET} showDataHeader={false} />)
   expect(asFragment()).toMatchSnapshot()
 })
--- a/src/nft/components/details/detailsV2/DataPage.test.tsx
+++ b/src/nft/components/details/detailsV2/DataPage.test.tsx
@@ -8,6 +8,8 @@ it('data page loads with header showing', () => {
   expect(asFragment()).toMatchSnapshot()
 })
 
+// The header is hidden via opacity: 0 to maintain its spacing, so it still exists in the DOM
+// Therefore we can not check for its non-existence and instead rely on comparing the full generated snapshots
 it('data page loads without header showing', () => {
   const { asFragment } = render(<DataPage asset={TEST_NFT_ASSET} showDataHeader={false} />)
   expect(asFragment()).toMatchSnapshot()
--- a/src/nft/components/details/detailsV2/DataPage.tsx
+++ b/src/nft/components/details/detailsV2/DataPage.tsx
@@ -35,6 +35,20 @@ const DataPageContainer = styled(Column)`
   margin: 0 auto;
 `
 
+const HeaderContainer = styled.div<{ showDataHeader?: boolean }>`
+  position: sticky;
+  top: ${({ theme }) => `${theme.navHeight}px`};
+  padding-top: 16px;
+  backdrop-filter: blur(12px);
+  z-index: 1;
+  transition: ${({ theme }) => `opacity ${theme.transition.duration.fast}`};
+  opacity: ${({ showDataHeader }) => (showDataHeader ? '1' : '0')};
+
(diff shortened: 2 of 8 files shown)


src/nft/components/details/detailsV2/LandingPage.test.tsx
```diff
@@ -3,10 +3,26 @@ import { render } from 'test-utils/render'
...
+beforeEach(() => {
+  // IntersectionObserver isn't available in test environment
+  const mockIntersectionObserver = jest.fn()
+  mockIntersectionObserver.mockReturnValue({
+    observe: () => null,
+    unobserve: () => null,
+    disconnect: () => null,
+  })
+  window.IntersectionObserver = mockIntersectionObserver
+})
```
zzmp, line 15: Consider setting this up as a polyfill. Are you at all worried that some browsers will not support this?

See https://www.npmjs.com/package/intersection-observer.

src/nft/components/details/detailsV2/DataPage.test.tsx
```diff
@@ -3,7 +3,12 @@ import { render } from 'test-utils/render'
...
 import { DataPage } from './DataPage'
 
-it('placeholder containers load', () => {
-  const { asFragment } = render(<DataPage asset={TEST_NFT_ASSET} />)
+it('data page loads with header showing', () => {
+  const { asFragment } = render(<DataPage asset={TEST_NFT_ASSET} showDataHeader={true} />)
+  expect(asFragment()).toMatchSnapshot()
+})
+
+it('data page loads without header showing', () => {
```
zzmp, line 11: You're not actually asserting these expectations. You should include some sort of selector into the rendered component ensuring the header is/is not shown, _along_ with the snapshot.

The snapshot acts as a guard that nothing changes, but if you do change something, and it breaks the header, and a dev just blindly updates the snapshot, you won't catch the breakage.

zzmp (changes requested):

src/nft/components/details/detailsV2/LandingPage.test.tsx
cbachmeier: No this is strictly for tests. The IntersectionObserver isn't available in the jest test env. See here: https://stackoverflow.com/questions/44249985/testing-code-that-uses-an-intersectionobserver

Can I word this comment to be more clear?

src/nft/components/details/detailsV2/DataPage.test.tsx
cbachmeier: With the header we still need its spacing so to hide it we're setting `opacity: 0` thus the element is still in the DOM so we can't simply check if it's present/not present. Jest's method `.not.toBeVisible()` also doesn't work with opacity, instead it relies on `display: none`. Jest style checking only works for inline styles so `.toHaveStyle('opacity: 0')` never returns true since we are using styled components.

src/nft/components/details/detailsV2/LandingPage.tsx
```diff
@@ -102,12 +103,34 @@ const MediaContainer = styled.div`
...
 }
 
-export const LandingPage = ({ asset, collection }: LandingPageProps) => {
+export const LandingPage = ({ asset, collection, setShowDataHeader }: LandingPageProps) => {
+  const intersectionRef = useRef<HTMLDivElement>(null)
+
+  // Checks if the intersectionRef is in the viewport
+  // If it is not in the viewport, the data page header becomes visible
+  useEffect(() => {
+    const intersectionHandler = (entries: IntersectionObserverEntry[]) => {
```
JackShort, line 115: a potentially nicer way to do this instead of creating a new function and IntersectionHandler on every scroll event is that we can throw it into a ref to be initialized once like this: 

```
  const intersectionRef = useRef<HTMLDivElement>(null)
  const observableRef = useRef(new IntersectionObserver((entries) => {
      if (!entries[0].isIntersecting) {
        setShowDataHeader(true)
      } else {
        setShowDataHeader(false)
      }
  }))

  // Checks if the intersectionRef is in the viewport
  // If it is not in the viewport, the data page header becomes visible
  useEffect(() => {
    const cachedRef = intersectionRef.current
    const observer = observableRef.current
    if (cachedRef && observer) {
      observer.observe(cachedRef)
      return () => observer.unobserve(cachedRef)
    }
    return
  }, [intersectionRef, observableRef, setShowDataHeader])
```

JackShort (approved):

src/nft/components/details/detailsV2/DataPage.tsx
```diff
@@ -35,6 +35,20 @@ const DataPageContainer = styled(Column)`
   margin: 0 auto;
 `
 
+const HeaderContainer = styled.div<{ showDataHeader?: boolean }>`
+  position: sticky;
```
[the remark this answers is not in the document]
cbachmeier: Yeah, Phil confirmed it should be sticky

src/nft/components/details/detailsV2/DataPage.test.tsx
zzmp: This is definitely not ideal but sounds fine for this use case. You may want to leave a comment in the file explaining why you were unable to do more than just snapshot tests.

zzmp (approved):

src/nft/components/details/detailsV2/DataPage.test.tsx
cbachmeier: Added a comment to provide context

src/nft/components/details/detailsV2/LandingPage.test.tsx
cbachmeier: Thanks for the callout. IO is [very well supported](https://caniuse.com/intersectionobserver) by modern browsers so I'm not worried in this case. But if we do happen to find a large portion of our user base is using IE we can look to this solution.

PR #6588, opened 2023-05-16
fix: update nonce deduplication logic

cartcrom:
## Description

Updates nonce deduplication logic to not hide all transactions that are present on both backend and frontend cache. These transactions were previously hidden by recent changes meant to handle cancelled tx's.

## Screen capture

| image | image | 

### Reproducing the error

1. Make a swap on mainnet, wait for it to confirm, and view the tx in your mini portfolio
2. Refresh the page (to fetch remote tx's) and the tx should no longer be in your mini portfolio

### DIFF
--- a/src/components/AccountDrawer/MiniPortfolio/Activity/index.tsx
+++ b/src/components/AccountDrawer/MiniPortfolio/Activity/index.tsx
@@ -82,7 +82,23 @@ const ActivityGroupWrapper = styled(Column)`
   gap: 8px;
 `
 
-function combineActivities(localMap: ActivityMap = {}, remoteMap: ActivityMap = {}): Array<Activity> {
+/* Detects transactions from same account with the same nonce and different hash */
+function wasTxCancelled(localActivity: Activity, remoteMap: ActivityMap, account: string): boolean {
+  if (!localActivity.nonce) return false // handles locally cached tx's that were stored before we started tracking nonces
+  
+  return Object.values(remoteMap).some((remoteTx) => {
+    if (!remoteTx) return false
+  
+    // Cancellations are only possible when both nonce and tx.from are the same
+    if (remoteTx.nonce === localActivity.nonce && remoteTx.receipt?.from === account) {
+      // If the remote tx has a different hash than the local tx, the local tx was cancelled
+      return remoteTx.hash.toLowerCase() !== localActivity.hash.toLowerCase()
+    }
+    return false
+  })
+}
+
+function combineActivities(localMap: ActivityMap = {}, remoteMap: ActivityMap = {}, account: string): Array<Activity> {
   const txHashes = [...new Set([...Object.keys(localMap), ...Object.keys(remoteMap)])]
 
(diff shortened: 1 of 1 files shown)


just-toby (approved):

src/components/AccountDrawer/MiniPortfolio/Activity/index.tsx
```diff
@@ -82,24 +82,37 @@ const ActivityGroupWrapper = styled(Column)`
...
 `
 
-function combineActivities(localMap: ActivityMap = {}, remoteMap: ActivityMap = {}): Array<Activity> {
+/* Detects transactions from same account with the same nonce and different hash */
+function wasTxCancelled(localActivity: Activity, remoteMap: ActivityMap, account: string): boolean {
+  // handles locally cached tx's that were stored before we started tracking nonces
+  if (!localActivity.nonce || localActivity.status !== TransactionStatus.Pending) return false
+
+  return Object.values(remoteMap).some((remoteTx) => {
+    if (!remoteTx) return false
```
cbachmeier, line 91: nit: you can remove this line if your `if` statement includes null checks:
`if (remoteTx?.nonce === localActivity.nonce && remoteTx?.receipt?.from.toLowerCase() === account.toLowerCase()) {`

cbachmeier (approved):
Successfully tested and successful swap does not clear on refresh. The contract interaction activity above the swap in the screenshot below is the cancelled request. Are we expecting this to be hidden? [Link to etherscan for cancelled tx](https://etherscan.io/tx/0x471893cb3f9a52530cc56aecc03d629b8dfb12d9250ed66b19aba773397c1f4c)
Screen Shot 2023-05-16 at 16 42 42

src/components/AccountDrawer/MiniPortfolio/Activity/index.tsx
tinaszheng: I actually prefer carters version better, the breakdown is more clear 😅

tinaszheng (approved):

cartcrom:
> Successfully tested and successful swap does not clear on refresh. The contract interaction activity above the swap in the screenshot below is the cancelled request. Are we expecting this to be hidden? [Link to etherscan for cancelled tx](https://etherscan.io/tx/0x471893cb3f9a52530cc56aecc03d629b8dfb12d9250ed66b19aba773397c1f4c) Screen Shot 2023-05-16 at 16 42 42

I think we should display _something_, so at least the user can click on etherscan for more details. Next steps would be using the locally cancelled tx to show what was cancelled, if available, and then further steps would be to have better cancel data on the BE

src/components/AccountDrawer/MiniPortfolio/Activity/index.tsx
cbachmeier: It might be worth standardizing how we handle potential undefined values. I also noticed the if statement already checks for remoteTx.receipt being undefined instead of returning false earlier. We should try to use the same standard between the two if possible.

PR #6591, opened 2023-05-17
test(e2e): configure Cypress to post PR status comments

zzmp:
## Description

Adds `cypress/flake` as a first-class check in GitHub PRs. Once existing flakes are fixed, we can use it to block add'l flakes from merging in.

Configures Cypress to post a comment to PRs with test results and two additional checks: a group check (`cypress:e2e`), and a flakes check (`cypress/flake`). These checks can be used in branch protections to gate merging to `main`, and `cypress:e2e` can replace the more complex `cypress-tests` check currently in use. Also configures Cypress so that structured data (author, PR) are available in the cloud dashboard.

Upgrades Cypress to v12.12.0, which is necessary for the Cypress cloud dashboard to link back to the PR. `CYPRESS_PULL_REQUEST_*` env vars are not available until v12. This makes tests isolated, so each test must now begin with `cy.visit()`. This should have no other effect on our tests. If we're lucky, it'll make them run run faster ⏳ .

_Linear ticket:_ https://linear.app/uniswap/issue/INFRA-7/configure-github-cypress-integration
_Relevant docs:_ https://docs.cypress.io/guides/continuous-integration/introduction#Git-information

## Test plan

This PR acts as its own test, as it is adding functionality to the PR review process.
The PR checks run, which should generate the two additional checks (`cypress:chrome` and `cypress/flake`) and a PR comment.
Once it is merged to `main`, I'll also need to check that the checks run and are posted to the commit on main. Failure here will not cause breakage, as removing the old `cypress-tests` check is left to a follow-up PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -122,11 +122,6 @@ jobs:
         containers: [1, 2, 3, 4]
     steps:
       - uses: actions/checkout@v3
-
-      # Sets ownership, which is misconfigured when checking out a new container.
-      # See https://github.com/actions/runner/issues/2033#issuecomment-1204205989.
-      - run: chown -R $(id -u):$(id -g) $PWD
-
       - uses: ./.github/actions/setup
       - uses: actions/cache@v3
         id: cypress-cache
@@ -154,13 +149,15 @@ jobs:
         env:
           CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          # Additional information is required for Cypress to post status updates to the PR:
-          # - if the event is push, the title will be undefined and Cypress will get the commit message from git 
-          # - if the event is pull_request, then we set the commit message to the pull request title
-          COMMIT_INFO_MESSAGE: ${{ github.event.pull_request.title }}
-          COMMIT_INFO_SHA: ${{ github.event.pull_request.head.sha }}
+          COMMIT_INFO_BRANCH: ${{ github.event.pull_request.head.ref || github.ref_name }}
+          COMMIT_INFO_MESSAGE: ${{ github.event.pull_request.title || github.event.head_commit.message }}
+          COMMIT_INFO_EMAIL: ${{ github.event.sender.user.email || github.event.pusher.email }}
+          COMMIT_INFO_AUTHOR: ${{ github.event.sender.user.name || github.event.pusher.name }}
+          COMMIT_INFO_SHA: ${{ github.event.pull_request.head.sha || github.event.head_commit.sha }}
+          COMMIT_INFO_TIMESTAMP: ${{ github.event.pull_request.updated_at || github.event.head_commit.timestamp }}
           CYPRESS_PULL_REQUEST_ID: ${{ github.event.pull_request.number }}
           CYPRESS_PULL_REQUEST_URL: ${{ github.event.pull_request.html_url }}
+          DEBUG: commit-info,cypress:server:record
 
       - uses: codecov/codecov-action@v3
         with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -122,6 +122,11 @@ jobs:
         containers: [1, 2, 3, 4]
     steps:
       - uses: actions/checkout@v3
+
+      # Sets ownership, which is misconfigured when checking out a new container.
+      # See https://github.com/actions/runner/issues/2033#issuecomment-1204205989.
+      - run: chown -R $(id -u):$(id -g) $PWD
+
       - uses: ./.github/actions/setup
       - uses: actions/cache@v3
         id: cypress-cache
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -122,12 +122,17 @@ jobs:
...
+      # See https://github.com/actions/runner/issues/2033#issuecomment-1204205989.
+      - run: chown -R $(id -u):$(id -g) $PWD
+
       - uses: ./.github/actions/setup
       - uses: actions/cache@v3
         id: cypress-cache
         with:
           path: /root/.cache/Cypress
-          key: ${{ runner.os }}-cypress
+          key: ${{ runner.os }}-cypress-12-12-0
```
cmcewen, line 135: what does this suffix mean?

.github/workflows/test.yml
```diff
@@ -122,12 +122,17 @@ jobs:
         containers: [1, 2, 3, 4]
     steps:
       - uses: actions/checkout@v3
+
+      # Sets ownership, which is misconfigured when checking out a new container.
+      # See https://github.com/actions/runner/issues/2033#issuecomment-1204205989.
+      - run: chown -R $(id -u):$(id -g) $PWD
```
cmcewen, line 128: cc @AndrewMohawk @pwnslinger

cypress/e2e/swap/swap.test.ts
```diff
@@ -7,7 +7,7 @@ const UNI_MAINNET = UNI[SupportedChainId.MAINNET]
 
 describe('Swap', () => {
   describe('Swap on main page', () => {
-    before(() => {
+    beforeEach(() => {
```
cmcewen, line 10: why are these test changes needed?

cypress/e2e/wallet-dropdown.test.ts
```diff
@@ -1,96 +1,108 @@
...
 
-  it('should select a language when not connected', () => {
-    cy.get(getTestSelector('wallet-language-item')).contains('Deutsch').click({ force: true })
-    cy.get(getTestSelector('wallet-header')).should('contain', 'Sprache')
-    cy.get(getTestSelector('wallet-language-item')).contains('English').click({ force: true })
-    cy.get(getTestSelector('wallet-header')).should('contain', 'Language')
-    cy.get(getTestSelector('wallet-back')).click()
-  })
+  describe('system dark mode', () => {
+    function visitWithSystemDarkMode(darkMode: boolean) {
```
cmcewen, line 56: i feel like `visitSwap` is more accurate? and default value would be nice. it seems more likely to me that you would add other params about the page visit than you would change the page you're visiting

.github/workflows/test.yml
zzmp: This reflects the cypress version, 12.12.0.
It could be generated programmatically, something like `yarn list cypress --silent` or hashfiles of `node_modules/cypress`.

cypress/e2e/wallet-dropdown.test.ts
zzmp: This can be `visitSwapWithSystemDarkMode`, but the important bit is that it is mocking the system dark mode, so I'd rather leave it.

This is only used twice, once with each value of `darkMode`, so I think a default would only hinder readability.

.github/workflows/test.yml
zzmp: updated to use `**/node_modules/cypress/package.json`.

.github/workflows/test.yml
AndrewMohawk: Discussed in slack :)

.github/workflows/test.yml
```diff
@@ -140,14 +140,24 @@ jobs:
...
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COMMIT_INFO_BRANCH: ${{ github.event.pull_request.head.ref || github.ref_name }}
+          COMMIT_INFO_MESSAGE: ${{ github.event.pull_request.title || github.event.head_commit.message }}
+          COMMIT_INFO_EMAIL: ${{ github.event.sender.user.email || github.event.pusher.email }}
+          COMMIT_INFO_AUTHOR: ${{ github.event.sender.user.name || github.event.pusher.name }}
+          COMMIT_INFO_SHA: ${{ github.event.pull_request.head.sha || github.event.head_commit.sha }}
+          COMMIT_INFO_TIMESTAMP: ${{ github.event.pull_request.updated_at || github.event.head_commit.timestamp }}
+          CYPRESS_PULL_REQUEST_ID: ${{ github.event.pull_request.number }}
+          CYPRESS_PULL_REQUEST_URL: ${{ github.event.pull_request.html_url }}
+          DEBUG: commit-info,cypress:server:record
```
cmcewen, line 160: you want to keep this on merge?

cmcewen (approved):

.github/workflows/test.yml
zzmp: nope, this is only here for testing

zzmp: I'm still trying to get the author to populate. as soon as that's coming through in cypress, I'll remove this line.

JFrankfurt (approved):

PR #6594, opened 2023-05-18
feat: Use routing-api v2 behind feature flag

tinaszheng:
## Description
This adds the updated router URL for swap quotes behind a feature flag

image

## Screen capture
No UI changes 😉 

## Test plan
Use Vercel link - in the feature flags modal `Enable the Unified Routing API`


src/state/routing/slice.ts
```diff
@@ -21,55 +20,13 @@ export enum RouterPreference {
...
 const API_QUERY_PARAMS = {
   protocols: 'v2,v3,mixed',
 }
 const CLIENT_PARAMS = {
   protocols: [Protocol.V2, Protocol.V3, Protocol.MIXED],
 }
-// Price queries are tuned down to minimize the required RPCs to respond to them.
-// TODO(zzmp): This will be used after testing router caching.
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const PRICE_PARAMS = {
```
tinaszheng, line 50: deleting this because it's not used

src/state/routing/uraSlice.ts
```diff
@@ -0,0 +1,122 @@
...
+// routing API quote query params: https://github.com/Uniswap/routing-api/blob/main/lib/handlers/quote/schema/quote-schema.ts
+const CLASSIC_SWAP_QUERY_PARAMS = {
+  ...CLIENT_PARAMS,
+  routingType: 'CLASSIC',
+}
+
+export const uraRoutingApi = createApi({
+  reducerPath: 'ura-routing',
+  baseQuery: fetchBaseQuery({
+    baseUrl: 'https://4uemel8n4g.execute-api.us-east-2.amazonaws.com/prod',
```
tinaszheng, line 24: awaiting final URL for here

just-toby: is this something we usually put in the .env file?

tinaszheng: we can, we don't typically switch api URLs for staging though so dont know if that would make sense

src/state/routing/slice.ts
```diff
@@ -1,15 +1,14 @@
 import { createApi, fetchBaseQuery, FetchBaseQueryError } from '@reduxjs/toolkit/query/react'
 import { Protocol } from '@uniswap/router-sdk'
 import { TradeType } from '@uniswap/sdk-core'
-import { AlphaRouter, ChainId } from '@uniswap/smart-order-router'
-import { RPC_PROVIDERS } from 'constants/providers'
-import { getClientSideQuote, toSupportedChainId } from 'lib/hooks/routing/clientSideSmartOrderRouter'
+import { ChainId } from '@uniswap/smart-order-router'
```
cmcewen, line 4: random but do we export this from sdk-core as well? should we import from there?

src/state/routing/useRoutingAPITrade.ts
```diff
@@ -45,17 +47,34 @@ export function useRoutingAPITrade<TTradeType extends TradeType>(
...
+
   const {
-    isError,
-    data: tradeResult,
-    currentData: currentTradeResult,
-  } = useGetQuoteQuery(queryArgs ?? skipToken, {
+    isError: isLegacyAPIError,
+    data: legacyAPITradeResult,
+    currentData: currentLegacyAPITradeResult,
+  } = useGetQuoteQuery(!shouldUseRoutingApiV2 ? queryArgs ?? skipToken : skipToken, {
```
cmcewen, line 56: this one liner is a little hard for me - if you're not using v2 api, query args, unless they don't exist in which case skipToken. if you're using v2, skipToken. i think maybe `shouldUseRoutingApiV2 ? skipToken : queryArgs ?? skipToken` is more readable to me, though i also would be opposed to moving this to a seprate statement

src/state/routing/useRoutingAPITrade.ts
```diff
@@ -45,17 +47,34 @@ export function useRoutingAPITrade<TTradeType extends TradeType>(
...
     pollingInterval: routerPreference === INTERNAL_ROUTER_PREFERENCE_PRICE ? ms`1m` : AVERAGE_L1_BLOCK_TIME,
     // If latest quote from cache was fetched > 2m ago, instantly repoll for another instead of waiting for next poll period
     refetchOnMountOrArgChange: 2 * 60,
   })
 
+  const {
+    isError: isV2APIError,
+    data: v2TradeResult,
+    currentData: currentV2TradeResult,
+  } = useGetQuoteQueryV2(shouldUseRoutingApiV2 ? queryArgs ?? skipToken : skipToken, {
```
cmcewen, line 67: this is the same as above but reversed? maybe it's one block to assign `v1QueryArg` and `v2QueryArg` above?

src/state/routing/useRoutingAPITrade.ts
```diff
@@ -45,17 +47,34 @@ export function useRoutingAPITrade<TTradeType extends TradeType>(
...
+    // Price-fetching is informational and costly, so it's done less frequently.
+    pollingInterval: routerPreference === INTERNAL_ROUTER_PREFERENCE_PRICE ? ms`1m` : AVERAGE_L1_BLOCK_TIME,
```
cmcewen, line 69: these look repeated from above?

src/state/routing/utils.ts
```diff
@@ -17,6 +20,22 @@ import {
   V3PoolInRoute,
 } from './types'
 
+const routers = new Map<ChainId, AlphaRouter>()
+export function getRouter(chainId: ChainId): AlphaRouter {
```
cmcewen, line 24: this is a copy/paste for future reviewers

src/state/routing/v2Slice.ts
```diff
@@ -0,0 +1,122 @@
+import { createApi, fetchBaseQuery, FetchBaseQueryError } from '@reduxjs/toolkit/query/react'
```
cmcewen, line 1: this file looks pretty copy/paste - could you differentiate what is new vs not?

src/state/routing/slice.ts
tinaszheng: so interestingly this one should be from SOR specifically since these are chainIds that are accepted by the client side router - theres a chance that sdk-core will support a superset of these chainIds

src/state/routing/useRoutingAPITrade.ts
tinaszheng: yup, copy pasted so it's easier to rip out the first block later

src/state/routing/useRoutingAPITrade.ts
tinaszheng: would prefer to keep these separate for ease of deletion later

src/state/routing/v2Slice.ts
```diff
@@ -0,0 +1,120 @@
...
+                }
+              } catch {
+                throw response.error
+              }
+            }
+
+            const quoteData = response.data as QuoteDataV2
+            const tradeResult = transformRoutesToTrade(args, quoteData.quote)
+
+            return { data: tradeResult }
```
tinaszheng, line 95: these lines are new logic - note v2 quote requests is a POST instead of GET, and has slightly different response/error formats

src/state/routing/v2Slice.ts
```diff
@@ -0,0 +1,120 @@
...
+            )
+          }
+        }
+        try {
+          const router = getRouter(args.tokenInChainId)
+          const quoteResult = await getClientSideQuote(args, router, CLIENT_PARAMS)
+          if (quoteResult.state === QuoteState.SUCCESS) {
+            return { data: transformRoutesToTrade(args, quoteResult.data) }
+          } else {
+            return { data: quoteResult }
```
tinaszheng, line 108: pretty much same logic here as routing/slice.ts

src/state/routing/v2Slice.ts
```diff
@@ -0,0 +1,120 @@
...
+}
+
+export const routingApiV2 = createApi({
+  reducerPath: 'routingApiV2',
+  baseQuery: fetchBaseQuery({
+    baseUrl: 'https://api.uniswap.org/v2/',
+  }),
+  endpoints: (build) => ({
+    getQuote: build.query<TradeResult, GetQuoteArgs>({
+      async onQueryStarted(args: GetQuoteArgs, { queryFulfilled }) {
```
tinaszheng, line 28: this function is pure copy-paste, just updated the trace name to include "v2"

src/state/routing/v2Slice.ts
```diff
@@ -0,0 +1,120 @@
...
+import { getClientSideQuote } from 'lib/hooks/routing/clientSideSmartOrderRouter'
+import ms from 'ms.macro'
+import { trace } from 'tracing/trace'
+
+import { GetQuoteArgs, INTERNAL_ROUTER_PREFERENCE_PRICE, RouterPreference } from './slice'
+import { QuoteDataV2, QuoteState, TradeResult } from './types'
+import { getRouter, isExactInput, shouldUseAPIRouter, transformRoutesToTrade } from './utils'
+
+const CLIENT_PARAMS = {
+  protocols: [Protocol.V2, Protocol.V3, Protocol.MIXED],
```
tinaszheng, line 12: note the params array format is different for v2 than in v1

natew (approved):

src/state/routing/useRoutingAPITrade.ts
cmcewen: i think just adding a comment to stuff you expect to rip out is useful context - don't worry about it this time but just helpful in the future 😄

src/state/routing/v2Slice.ts
```diff
@@ -0,0 +1,120 @@
...
+
+export const routingApiV2 = createApi({
+  reducerPath: 'routingApiV2',
+  baseQuery: fetchBaseQuery({
+    baseUrl: 'https://api.uniswap.org/v2/',
+  }),
+  endpoints: (build) => ({
+    getQuote: build.query<TradeResult, GetQuoteArgs>({
+      async onQueryStarted(args: GetQuoteArgs, { queryFulfilled }) {
+        trace(
```
cmcewen, line 29: separate but this trace API could be improved imo

src/state/routing/v2Slice.ts
```diff
@@ -0,0 +1,120 @@
...
+              } else {
+                throw error
+              }
+            }
+          },
+          {
+            data: {
+              ...args,
+              isPrice: args.routerPreference === INTERNAL_ROUTER_PREFERENCE_PRICE,
+              isAutoRouter:
```
cmcewen, line 50: nit but seems same as `shouldUseAPIRouter`

cmcewen (approved):
good to merge to me. however i would request we prioritize adding a unit test that mocks the API request/response at some point in the future

src/state/routing/v2Slice.ts
tinaszheng: 🤔 yeah not sure what this data tag is for, i think in the future we could just track the routerPreference instead of these boolean tags

PR #6597, opened 2023-05-18
build: notify from notify/test

zzmp:
## Description

Runs test workflow from a protected environment if on `main`.

_Linear ticket:_ https://linear.app/uniswap/issue/INFRA-12/isolate-the-slack-webhook-url-to-an-environment

### Test Plan

The logic (`${{ github.ref_name == 'main' && 'notify/test' }}`) was tested by modifying it to read `${{ github.ref_name == '6597/merge' && 'notify/test' }}` and observing the correct environment selection: https://github.com/Uniswap/interface/actions/runs/5019599122.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,8 +17,9 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     environment:
-      name: ${{ github.ref_name == 'main' && 'notify/test' }}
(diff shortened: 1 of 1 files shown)


grabbou (approved):

AndrewMohawk (approved):
The expression ```name: ${{ github.ref_name == 'main' && 'notify/test' }}``` is a little difficult in terms of readability, but other options would mean a lot more restructuring. I think we can add it as a TODO for the future.

PR #6599, opened 2023-05-18
feat: [DetailsV2] instant buy

JackShort:
Implementing single purchase asset for nft details. Design has not figured out the way to handle price changes yet or assets that happen to be unavailable so right now we just return and don't execute a metamask transaction.

### Automated testing

- [x] Unit test


src/nft/components/details/detailsV2/BuyButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+    )
+  }
+
+  return (
+    <>
+      <StyledBuyButton disabled={isLoadingRoute} onClick={() => fetchAndPurchaseSingleAsset(asset)}>
+        {isLoadingRoute ? (
+          <>
+            <Trans>Fetching Route</Trans>
+            <Loader size="24px" stroke="white" />
```
cbachmeier, line 59: Is the loader supposed to be a different color than the text?
Screen Shot 2023-05-18 at 15 59 40

src/nft/components/details/detailsV2/BuyButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+            <Trans>Buy</Trans>
+            <Price>{formatNumber(price)} ETH</Price>
+          </>
+        )}
+      </StyledBuyButton>
+      {onDataPage && (
+        <MakeOfferButtonSmall>
+          <HandHoldingDollarIcon />
+        </MakeOfferButtonSmall>
+      )}
```
cbachmeier, line 72: I would separate the offer buttons from this component into their own file. When we do implement making offers, there will be complicated logic we'll want to keep separate from buying.

src/nft/components/details/detailsV2/BuyButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const { fetchAndPurchaseSingleAsset, isLoading: isLoadingRoute } = useFetchSingleAsset()
+  const price = asset.sellorders?.[0]?.price.value
+
+  if (!price) {
+    return (
+      <MakeOfferButtonLarge>
+        <Trans>Make an offer</Trans>
+      </MakeOfferButtonLarge>
+    )
+  }
```
cbachmeier, line 51: Same comment here about moving offer buttons outside of the BuyButton component

src/nft/components/details/detailsV2/BuyButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  return (
+    <>
+      <StyledBuyButton disabled={isLoadingRoute} onClick={() => fetchAndPurchaseSingleAsset(asset)}>
+        {isLoadingRoute ? (
```
cbachmeier, line 56: Should we have a state for when we're waiting for a user to approve in their wallet and then again when waiting for mainnet to confirm?
Screen Shot 2023-05-18 at 15 58 12

src/nft/hooks/useFetchAssets.ts
```diff
@@ -100,3 +106,48 @@ export function useFetchAssets(): () => Promise<void> {
...
+          setIsLoading(false)
+
+          if (!data.nftRoute || !data.nftRoute.route) {
+            return
+          }
+
+          const { route, routeResponse } = buildRouteResponse(data.nftRoute, false)
+          const { hasPriceAdjustment, updatedAssets } = compareAssetsWithTransactionRoute([asset], route)
+          const { unavailable } = filterUpdatedAssetsByState(updatedAssets)
+          const invalidData = hasPriceAdjustment || unavailable.length > 0
```
cbachmeier, line 139: Can you rebase off main so we can test with pooled assets?

src/nft/hooks/useFetchAssets.ts
```diff
@@ -100,3 +106,48 @@ export function useFetchAssets(): () => Promise<void> {
     tokenTradeInput,
   ])
 }
+
+export const useFetchSingleAsset = () => {
```
cbachmeier, line 110: A lot of this code is duplicated from `useFetchAssets`. Have you tried when a users hits the buy button to add it to the bag, keep the bag hidden, and then just call `useFetchAssets()`? This would pair well with the updated designs for handling price changes as then you'd just have to reveal the bag.

cbachmeier: The follow up would be if a user already has items in their bag. In that case you could add an optional `asset?` param to `useFetchAsset` which would override using the assets currently in the bag.

cbachmeier (changes requested):

src/nft/components/details/detailsV2/BuyButton.tsx
JackShort: the loading animation and stuff hasn't been designed yet this is more of a placeholder

src/nft/components/details/detailsV2/BuyButton.tsx
JackShort: these states haven't been designed yet

src/nft/components/details/detailsV2/BuyButton.tsx
JackShort: yaya I kinda stated it above going to do a pr which dives more into handling the button this is just like initial setup

src/nft/hooks/useFetchAssets.ts
cbachmeier: I still can't seem to buy pooled assets even without a price change. Tested with this [collection](https://interface-git-jack-singleassetbuy-uniswap.vercel.app/#/nfts/collection/0xec9c519d49856fd2f8133a0741b4dbe002ce211b)
pooledBuy

src/nft/hooks/useFetchAssets.ts
cbachmeier: This function would still need to touch bag state in the case of a price change since we'll need to display that change in the bag to the user to review. If we leverage `useFetchAssets` that logic is already handled for us. The only substantial change for the fetch query would be a conditional on `nftTrades: singleAsset ? buildNftTradeInput([singleAsset]) : buildNftTradeInputFromBagItems(itemsInBag),` the rest of the logic should be able to be left untouched.

cbachmeier (commented):
unable to 1 click buy pooled assets with no price change

cbachmeier (approved):
LGTM. Successfully 1 click purchased pooled asset

src/nft/components/details/detailsV2/BuyButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const { fetchAndPurchaseSingleAsset, isLoading: isLoadingRoute } = useFetchSingleAsset()
+  const price = asset.sellorders?.[0]?.price.value
+
+  if (!price) {
+    return <OfferButton />
+  }
+
+  return (
+    <>
+      <StyledBuyButton disabled={isLoadingRoute} onClick={() => fetchAndPurchaseSingleAsset(asset)}>
```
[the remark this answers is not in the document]
JackShort: I am a little confused by this comment it is using a callback - wondering if you could explain more as to what you would like me to do here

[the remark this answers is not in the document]
cmcewen: the onClick function takes a function, which you create new every render. you could use `useCallback` to keep the function referentially equal if you want, but since this is a simple component and unlikely to change, it's a bit of a wash

cmcewen (approved):

PR #6601, opened 2023-05-18
build: restrict promotions to their branches

zzmp:
## Description

Restricts promotion workflows to their branches. These flows will fail if not run from the correct branch, due to environment configurations.

This uses an undocumented setting, and needs to be tried (in a test repo) first.

_Linear ticket:_ https://linear.app/uniswap/issue/INFRA-20/3-should-be-auto-run-from-releasesstaging


lavalamp- (approved):
No idea if this will work but let's give it a go??

PR #6603, opened 2023-05-18
feat: expand settings by default when custom values are set

grabbou:
Each time Settings menu is open, deadline/slippage should be expanded when custom value was provided. Required moving the state outside of Expand component to make it a controlled component. 

https://linear.app/uniswap/issue/WEB-2072/add-expand-animation-to-settings-popup


src/components/Expand/index.tsx
```diff
@@ -22,22 +22,25 @@ export default function Expand({
   button,
   children,
   testId,
+  isOpen,
```
zzmp, line 25: You can name the prop `open` and still have a better naming internally:

```suggestion
  open: isOpen,
```

zzmp (approved):

src/components/Expand/index.tsx
```diff
@@ -22,22 +22,25 @@ export default function Expand({
...
+  onToggle,
 }: PropsWithChildren<{
   header: ReactElement
   button: ReactElement
   testId?: string
+  isOpen: boolean
+  onToggle: () => void
```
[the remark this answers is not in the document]
grabbou: Yeah, initially that was discussed, but after searching existing conventions in the codebase, I found that `isOpen`, `isExpanded` and `isX` are way more popular than `open` (just used in one or two components)

src/components/Expand/index.tsx
tinaszheng: isOpen feels more canonical  - booleans usually start with `is` or `should`

tinaszheng (approved):

PR #6604, opened 2023-05-18
fix: Use client side router only for price fetching

tinaszheng:
## Description
We were previously using client-side-only routing for price fetching, and switched to API, and the increase in pricing requests for illiquid tokens spiked our 4xx errors, so revert back to client-side routing for now

### DIFF
--- a/src/state/routing/slice.ts
+++ b/src/state/routing/slice.ts
@@ -126,11 +126,7 @@ export const routingApi = createApi({
         )
       },
       async queryFn(args, _api, _extraOptions, fetch) {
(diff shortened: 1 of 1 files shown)


zzmp (approved):
If I'm not mistaken, this is a partial revert of your recent refactor. Is that right, @tinaszheng?

zzmp (changes requested):
Could you include comments as to why this is necessary?

JFrankfurt (approved):

zzmp (approved):

PR #6606, opened 2023-05-19
feat: upgrade to node 18

cbachmeier:
## Description

- Updates node env to node 18
- Skipping 16 as 16's EOL is later this year where as Node 18 will be supported until [Apr 2025](https://endoflife.date/nodejs)
- Change fetchSchema to use `yarn` over `npx`
  - this was needed as upgrading node revealed that we had no version control on the npm package used by the compiler (ie Vercel) so we ran into a build error where Vercel was using an outdated version of npm for node 18.
  - We encourage yarn use in `engines` so this maintains consistency
- The node version set in `engines` [will overwrite the package setting in Verce](https://vercel.com/docs/concepts/functions/serverless-functions/runtimes/node-js)l, waiting until after this is deployed to prod to update the package setting. [[WEB-2094](https://linear.app/uniswap/issue/WEB-2094/configure-build-infrastructure-to-use-node-18)]

_JIRA ticket:_ https://linear.app/uniswap/issue/WEB-2090/upgrade-to-node-18

## Test plan

Leverages out existing test infrastructure. Additionally I will be setting up a bug bash session similar to our Webpack 5 upgrade when the build pipeline is finalized.

[WEB-2094]: https://uniswaplabs.atlassian.net/browse/WEB-2094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JFrankfurt:
remember to update the version setting in vercel as well 
image

cbachmeier:
> remember to update the version setting in vercel as well 
Thanks for pointing out. The node version specific in `engine` will overwrite the Vercel settings. I plan on updating the Vercel package settings once this change has been successfully released. I also have a node-18 project setup in Vercel with this setting for redundancy.

scripts/fetch-schema.js
```diff
@@ -10,7 +10,7 @@ const thegraphConfig = require('../graphql.thegraph.config')
 const exec = promisify(child_process.exec)
 
 function fetchSchema(url, outputFile) {
-  exec(`npx get-graphql-schema --h Origin=https://app.uniswap.org ${url}`)
+  exec(`get-graphql-schema --h Origin=https://app.uniswap.org ${url}`)
```
cbachmeier, line 13: When left in the use of `npx` causes an error as the machine's `npm` version is out of date. We currently do not specify npm version in our package.json `engines` to suggest use of `yarn`. Replacing `npx` with the following `yarn`, `yarn x`, `yarn dlx`, `yarn create`, all lead to syntax errors when trying to parse the retrieved schema.

.github/actions/setup/action.yml
```diff
@@ -8,7 +8,7 @@ runs:
 
     - uses: actions/setup-node@v3
       with:
-        node-version: 14
+        node-version: 18
```
zzmp, line 11: our release workflows can also be updated to omit the extra setup-node steps:
https://github.com/Uniswap/interface/blob/2415a1e3cd6ba6cf3bab43188251bca2b56e5e45/.github/workflows/4-deploy-to-prod.yml#L77-L80

https://github.com/Uniswap/interface/blob/2415a1e3cd6ba6cf3bab43188251bca2b56e5e45/.github/workflows/2-deploy-to-staging.yml#L30-L33

https://github.com/Uniswap/interface/blob/2415a1e3cd6ba6cf3bab43188251bca2b56e5e45/.github/workflows/release.yaml#L101-L104

scripts/fetch-schema.js
```diff
@@ -10,12 +10,14 @@ const thegraphConfig = require('../graphql.thegraph.config')
 const exec = promisify(child_process.exec)
 
 function fetchSchema(url, outputFile) {
-  exec(`npx get-graphql-schema --h Origin=https://app.uniswap.org ${url}`)
+  exec(`yarn get-graphql-schema --h Origin=https://app.uniswap.org ${url}`)
```
zzmp, line 13: This should use `yarn --silent` to avoid yarn logs, otherwise you'll get extra lines at the top _and_ bottom.

```suggestion
  exec(`yarn --silent get-graphql-schema --h Origin=https://app.uniswap.org ${url}`)
```

zzmp (changes requested):

.github/actions/setup/action.yml
cbachmeier: Do we have a good way to dryrun test that these deploys still work?

scripts/fetch-schema.js
cbachmeier: Beautiful, confirmed the schema was unchanged after fetching. Thanks for the solution.

.github/actions/setup/action.yml
zzmp: I think that a staging deploy could be considered a dry run.

package.json
```diff
@@ -261,7 +261,7 @@
   },
   "engines": {
     "npm": "please-use-yarn",
-    "node": "14",
+    "node": "18.x",
```
zzmp, line 264: nit: functionally equivalent to `18`, can we just use `18` to simplify?

```suggestion
    "node": "18",
```

zzmp (approved):

package.json
cbachmeier: Vercel specifically uses `18.x` notation which is why I stuck with it https://vercel.com/docs/concepts/functions/serverless-functions/runtimes/node-js

JFrankfurt (approved):

PR #6607, opened 2023-05-19
fix: use ephemeral props for styled component

JFrankfurt:
## Description
React was complaining about these

### Before
image

### After
image


tinaszheng (approved):
do b like that

zzmp (approved):

PR #6609, opened 2023-05-19
feat: improve yarn prepare scripts

JFrankfurt:
This makes independent prepare scripts run in parallel which speeds up the default `yarn` command--it also gives labeled logs which should make issues easier to debug.


scripts/prepare.js
```diff
@@ -0,0 +1,36 @@
+const { spawn } = require('child_process')
```
zzmp, line 1: Add `/* eslint-env node */` to the top to fix this lint :)

scripts/prepare.js
```diff
@@ -0,0 +1,36 @@
...
+        return reject(new Error(`${command} ${args} exited with code ${code}`))
+      }
+      console.log(`${command} ${args} completed successfully`)
+      resolve()
+    })
+  })
+}
+
+// Add setup commands to this array
+const commands = ['ajv', 'contracts', 'graphql', 'i18n']
```
zzmp, line 32: Move this to a constant at the top of the file to make configuration more obvious:

```
const SETUP_COMMANDS = ['ajv', 'contracts', graphql', i18n']
```

scripts/prepare.js
```diff
@@ -0,0 +1,36 @@
+const { spawn } = require('child_process')
+
+function runCommand(command, args) {
```
zzmp, line 3: These will always be yarn scripts, so you can simplify this by omitting `yarn`:

```suggestion
function runYarnScript(command) {
```

scripts/prepare.js
```diff
@@ -0,0 +1,36 @@
...
+  return new Promise((resolve, reject) => {
+    const process = spawn(command, args)
+
+    // Log when the process starts
+    console.log(`${command} ${args} initiated`)
```
zzmp, line 8: nit: I'd prefer these look like this: `[${command}] initiated`

scripts/prepare.js
```diff
@@ -0,0 +1,36 @@
...
+
+    // Uncomment to see output of each command
+    // process.stdout.on('data', (data) => {
+    //   console.log(`${command}: ${data}`)
+    // })
```
zzmp, line 13: Why don't we want to see output? Could you instead condition this on a `--verbose` flag?

scripts/prepare.js
```diff
@@ -0,0 +1,36 @@
...
+    process.stderr.on('data', (data) => {
+      console.error(`${command} ${args} error: ${data}`)
+    })
+
+    // Log when and how the command exited
+    process.on('close', (code) => {
+      if (code !== 0) {
+        return reject(new Error(`${command} ${args} exited with code ${code}`))
+      }
+      console.log(`${command} ${args} completed successfully`)
```
zzmp, line 25: I think yarn will itself log when a command is completed. How much of the yarn command log is duplicated here? Do you even need to log when a command is initiated?

scripts/prepare.js
```diff
@@ -0,0 +1,36 @@
...
+      resolve()
+    })
+  })
+}
+
+// Add setup commands to this array
+const commands = ['ajv', 'contracts', 'graphql', 'i18n']
+
+Promise.all(commands.map((command) => runCommand('yarn', [command])))
+  .then(() => console.log('All commands finished'))
```
zzmp, line 35: A timer here would be nice

scripts/prepare.js
JFrankfurt: I think the init log is useful to remind you what is executing under the hood.

Yarn doesn't log on its own when these child processes complete individually.

zzmp (changes requested):

scripts/prepare.js
```diff
@@ -0,0 +1,51 @@
...
+    // Uncomment to see output of each yarn
+    if (verbose) {
+      childProcess.stdout.on('data', (data) => {
+        console.log(`Output from yarn ${command}:\n${data}`)
+      })
+    }
+
+    // Log errors
+    childProcess.stderr.on('data', (data) => {
+      console.error(`Error during execution of yarn ${command}:\n${data}`)
```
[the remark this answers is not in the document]
JFrankfurt: vanilla terminals don't give this a styling distinction so I think the text designation is appropriate

scripts/prepare.js
```diff
@@ -0,0 +1,54 @@
...
+        data
+          .split('\n')
+          .map((line) => `[${command}] ${line}`)
+          .forEach(console.log)
+      })
+    }
+
+    // Log errors
+    childProcess.stderr.on('data', (data) => {
+      console.error(`Error during execution of [${command}]:\n${data}`)
```
zzmp, line 35: This still needs to be split and rejoined into lines like stdout.

scripts/prepare.js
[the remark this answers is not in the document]
zzmp: could you maybe do `[${command}] Error: ${line}`?

cmcewen (approved):

zzmp (approved):

PR #6617, opened 2023-05-22
feat: add animation to Settings menu

grabbou:
## Description

It adds back expand animation to settings menu by reusing already existing `AnimationDropdown` component.

Note:

In a follow-up PR, `AnimatedDropdown` and `Expand` components will be merged into a single component. I will replace a single direct use of `AnimatedDropdown` with `Expand` since APIs are the same. I am splitting that work into a separate PR to keep it more atomic. Until then, it may seem that they're very similar and their test coverage overlaps.

## Test plan

Go to preview link and toggle slippage/deadline in settings menu


tinaszheng (approved):

src/test-utils/matchers.ts
```diff
@@ -0,0 +1,32 @@
...
+// Overrides the Testing Library matcher to check for height when determining whether an element is visible.
+// We are doing this because:
+// - original `toBeVisible()` does not take `height` into account
+// https://github.com/testing-library/jest-dom/issues/450
+// - original `toBeVisible()` and `toHaveStyle()` does not work at all in some cases
+// https://github.com/testing-library/jest-dom/issues/209
+// - `getComputedStyles()` returns empty object, making it impossible to check for Styled Components styles
+// https://github.com/styled-components/styled-components/issues/3262
+// https://github.com/jsdom/jsdom/issues/2986
+// For the reasons above, this matcher only works for inline styles.
```
zzmp, line 17: This should be a doc-comment, not an implementation comment.
In other words, `/**` :P

zzmp (approved):
lgtm, but please update your comment to a doc-comment before merging.
Feel free to ping me for re-approval.

src/test-utils/matchers.ts
grabbou: Why should it be a doc comment? I didn't know `//` was an implementation comment. So far, I think we made a distinction that we're only using `/**` where we plan to use JSDoc. I don't think applies here?

grabbou:
I will ship this PR "as-is" to avoid pinging both Tina and you Zach for re-approval. Let's continue the discussion and based on the outcomes, I will send a follow-up PR. I feel like I should also follow-up with a PSA on Slack to make sure everyone's aware of that rule.

src/test-utils/matchers.ts
zzmp: It explains what a function is doing, so it's a doc-comment. In general, if you're commenting above a function, it's a doc-comment; if you're commenting in a function definition, it's an impl comment.

> So far, I think we made a distinction that /** was only to be used when JSDoc/TSDoc were present.

Using `/**` makes it JSDoc, even if you don't use specific JSDoc tags. With TypeScript, most JSDoc comments are very simple, just `/** Return xyz. */`, because the actual types are already documented through the language. If it's something you'd expect to be surfaced as documentation, explaining a function's usage, showing up through the IDE, it's a doc-comment and should utilize `/**`.

grabbou: You are right, I remember we made a distinction between function-level documents and multi-line comments inside the body (implementation). I will follow-up with an update.

PR #6618, opened 2023-05-22
chore: cleaning up buy button states

JackShort:
* better state handling for buy button, less confusing, clear in/out data
* also added tests for each possible state

### Automated testing

- [x] Unit test


src/nft/components/bag/ButtonStates.tsx
```diff
@@ -0,0 +1,148 @@
...
+  disabled: boolean
+  warningText: ReactNode | undefined
+  warningTextColor: string
+  helperText: ReactNode | undefined
+  helperTextColor: string
+  buttonColor: string
+  buttonTextColor: string
+}
+
+export function getBuyButtonStateData(
```
cbachmeier, line 37: Can you add unit tests for this fn?

cbachmeier, line 37: nit: If you rename this to `useBuyButtonStateData` to treat it as a hook, you can then use `const theme = useTheme` to not have to pass the theme object with each call.

src/nft/components/bag/BagFooter.tsx
```diff
@@ -377,105 +379,78 @@ export const BagFooter = ({ setModalIsOpen, eventProperties }: BagFooterProps) =
...
+
+    if (sufficientBalance === false) {
+      return getBuyButtonStateData(BuyButtonStates.INSUFFICIENT_BALANCE, theme)
+    }
+
+    if (bagStatus === BagStatus.WARNING) {
+      return getBuyButtonStateData(BuyButtonStates.ERROR, theme)
+    }
+
+    if (!connected) {
```
cbachmeier, line 396: you can consolidate this check and ln383 `if (connected && chainId !== SupportedChainId.MAINNET)`

src/nft/components/bag/BagFooter.tsx
```diff
@@ -377,105 +379,78 @@ export const BagFooter = ({ setModalIsOpen, eventProperties }: BagFooterProps) =
...
-    let disabled = true
-    let warningText = undefined
-    let warningTextColor = theme.accentWarning
-    let helperText = undefined
-    let helperTextColor = theme.textSecondary
-    let buttonColor = theme.accentAction
-    let buttonTextColor = theme.accentTextLightPrimary
-
+  } = useMemo((): BuyButtonState => {
     if (connected && chainId !== SupportedChainId.MAINNET) {
```
cbachmeier, line 383: Thoughts on these multiple return calls vs instantiating lets for the arguments for `getBuyButtonStateData`, updating based on if statements, and having 1 return fn call at the end?

src/nft/components/bag/BagFooter.tsx
```diff
@@ -377,105 +379,78 @@ export const BagFooter = ({ setModalIsOpen, eventProperties }: BagFooterProps) =
...
-      handleClick = () => switchChain(connector, SupportedChainId.MAINNET)
-      buttonText = <Trans>Switch networks</Trans>
```
cbachmeier, line 393: The new refactor doesn't check for `BuyButtonStates.NOT_SUPPORTED_CHAIN` ie I'm on Polygon and getting the 'Connect Wallet' text
Screen Shot 2023-05-23 at 09 45 19

cbachmeier (changes requested):

cbachmeier:
So much more readable 🙌

src/nft/components/bag/BagFooter.tsx
JackShort: the point of this refactor was to have clear data-in -> data-out, having a mutable object makes that less clear and its only/should going to get mutated once. Also this doesn't require you have to go down all of the branches to return

src/nft/components/bag/BagFooter.tsx
JackShort: slightly confused as to how I would do that

src/nft/components/bag/ButtonStates.tsx
JackShort: unit tests don't really make sense for this function, what would it be defending against. If a future dev adds to the `BuyButtonStates` enum and doesn't add an appropriate case eslint will throw an error and vice versa

src/nft/components/bag/BagFooter.tsx
```diff
@@ -274,18 +276,18 @@ const FiatValue = ({
   )
 }
 
-interface BagFooterProps {
-  setModalIsOpen: (open: boolean) => void
-  eventProperties: Record<string, unknown>
-}
-
 const PENDING_BAG_STATUSES = [
```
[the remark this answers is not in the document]
JackShort: this is a little different these are just the BagStatuses that are considered to be pending - they have overlap but they are for different purposes and BagStatus refers to a slightly different state than the buy button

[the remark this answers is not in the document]
JackShort: like we don't store the button states as an actual state but we do for the bag statuses, so in order to converge the state we would have to like start storing the button state and everything which is kinda overkill I think

src/nft/components/bag/ButtonStates.tsx
JackShort: as you can see adding a state throws an eslint error

Screen Shot 2023-05-23 at 1 08 04 PM

cbachmeier: You can pass in the various different `BuyButtonStates` and expect the corresponding `BuyButtonStateData`. This will help prevent future devs from unknowingly making changes to the `BuyButtonStateData` possible leading to a `BuyButton` regression

src/nft/components/bag/BagFooter.tsx
cbachmeier: My mistake, I misread `if (connected && chainId !== SupportedChainId.MAINNET)` as `if (!connected && chainId !== SupportedChainId.MAINNET)`

src/nft/components/bag/BagFooter.test.tsx
```diff
@@ -0,0 +1,415 @@
...
+      isLocked: false,
+      itemsInBag: [],
+    }) as ReturnType<typeof useBag>
+
+    renderBagFooter()
+    const buyButton = getBuyButton()
+
+    expect(buyButton.textContent).toBe('Proceed in wallet')
+    expect(buyButton).toBeDisabled()
+  })
```
cbachmeier, line 199: these two look identical

src/nft/components/bag/BagFooter.test.tsx
```diff
@@ -0,0 +1,415 @@
+import { BigNumber } from '@ethersproject/bignumber'
```
cbachmeier, line 1: very thorough, great job

cbachmeier (approved):

src/nft/components/bag/BagFooter.test.tsx
JackShort: mb - was supposed to be `CONFIRMING_IN_WALLET` for the bag status

cbachmeier (approved):

cmcewen (approved):

src/nft/components/bag/BagFooter.tsx
```diff
@@ -377,105 +358,78 @@ export const BagFooter = ({ setModalIsOpen, eventProperties }: BagFooterProps) =
...
-      buttonText,
-      buttonTextColor,
-      disabled,
-      warningText,
-      warningTextColor,
-      helperText,
-      helperTextColor,
-      handleClick,
-      buttonColor,
+    if (bagStatus === BagStatus.CONFIRM_QUOTE) {
```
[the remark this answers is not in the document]
JackShort: Yaya I think the next step after this is to do exactly that and have a step function that evaluates everything into a BagStatus

PR #6622, opened 2023-05-22
feat: show affordance in swap UI when we can't fetch usd quote

grabbou:
## Description

As per the PR title and following screenshots, show the tooltip and `-` instead of empty space when we can't fetch a USD quote.

Unfortunately, due to the fact that we're using hooks quite extensively and that `useUSDPrice` always returns an object, it required an additional boolean in order to differentiate between `undefined` (indicating no USD price due to the user not providing all the information yet - in this case, we don't show `-`) and `data` being undefined (indicating no USD price due to insufficient liquidity - show `-`).

I have also refactored the component while on it, to remove deprecated UI components (such as `DeprecatedBody`) and replaced whitespace with Figma-defined `sm` spacing.

## Screen capture

| Web_capture_23-3-2023_14658_app_uniswap_org | Screenshot 2023-05-23 at 00 03 59 |

## Tests

I did not add new tests, due to lack of existing tests for those components and the fact that the overall diff here is small. I will add a note to add an appropriate e2e scenario, where we can test for other things too.

### DIFF
--- a/src/components/CurrencyInputPanel/FiatValue.tsx
+++ b/src/components/CurrencyInputPanel/FiatValue.tsx
@@ -1,15 +1,13 @@
 import { Trans } from '@lingui/macro'
-// eslint-disable-next-line no-restricted-imports
-import { t } from '@lingui/macro'
 import { formatNumber, formatPriceImpact, NumberType } from '@uniswap/conedison/format'
 import { Percent } from '@uniswap/sdk-core'
+import Row from 'components/Row'
 import { LoadingBubble } from 'components/Tokens/loading'
 import { MouseoverTooltip } from 'components/Tooltip'
 import { useMemo } from 'react'
-import styled, { useTheme } from 'styled-components/macro'
-
-import { ThemedText } from '../../theme'
-import { warningSeverity } from '../../utils/prices'
+import styled from 'styled-components/macro'
+import { ThemedText } from 'theme'
+import { warningSeverity } from 'utils/prices'
 
 const FiatLoadingBubble = styled(LoadingBubble)`
   border-radius: 4px;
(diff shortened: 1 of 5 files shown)


tinaszheng (commented):
so `useRoutingAPITrade` returns a `state` variable that highlights this exact error case through `TradeState.NO_ROUTE_FOUND`. what do you think of bubbling up that error up to the `useUSDPrice` hook? that way we can depend on the error state to show the no liquidity UI, rather than relying on `undefined` or `null`. that way other generic errors (like if the routing-api is down) won't be bundled into the same insufficient liquidity message.

grabbou:
@tinaszheng honestly, I think this is good call, but I would rather do it as a separate PR, because changing the return type of `useUSDPrice` hook to be `price | undefined` requires some changes in other parts that seem rather dangerous to do now (more specifically, understanding the UI behavior when that price is undefined). 

An alternative to do in this PR would be to somehow get the `TradeState` via a separate hook if possible, but rather than that, I think that approach is good enough for now. What I don't like about this approach however is the timing issue and potential glitch when currency is updated (to exist) but `useUSDPrice` did not refresh yet (so it would be reported as `-`).

zzmp (approved):

tinaszheng (approved):

tinaszheng:
> Okay, I gave it another look and what I found out is that we would have to bubble the error up from `useStablecoinPrice` which requires changing its signature from `Price<Currency, Token> | undefined` to `[Price<Currency, Token> | undefined, Error | undefined]`.
> 
> On top of that, we would need to rewrite the `useUSDPrice` hook entirely also to error in case the following line:
> 
> ```
> const ethUSDPrice = data?.token?.project?.markets?.[0]?.price?.value
> ```
> 
> returns undefined.
> 
> That could technically give us an opportunity to handle API errors/timeouts on top of the aforementioned insufficient liquidity, but it's going to be a more involved process.

fuq it too complicated

PR #6541, opened 2023-05-11
feat: permit2 e2e tests

just-toby:
## Description

Update the permit2 e2e tests to handle the new flow.

## Test plan
tests pass

### Automated testing

- [x] <s>Unit test N/A</s>
- [x] Integration/E2E test

PR #6624, opened 2023-05-22
fix: prevent race condition for swap state

just-toby:
## Description

This bug happens when the `ConfirmSwapModal` is dismissed before  the `swapCallback`  promise resolves.

When the bug happens, the state is set like this:
1. user proceeds normally and opens the `ConfirmSwapModal` -> `showConfirm` is true (set on L783)
2. `handleSwap` is called and triggers the `swapCallback` (L426). `showConfirm` is still true at this point
3. `handleConfirmDismiss` is called and `showConfirm` is set to `false`
4. `swapCallback` resolves and sets `showConfirm` to an outdated value (L428) which re-opens the modal

### Reproducing the error

1. see the [test results](https://github.com/Uniswap/interface/pull/6541) for the `swap/errors.test.ts` test suite - it's flaky because this race condition causes the modal to stay visible when it shouldn't

### QA (ie manual testing)

- [x] this change totally fixes flakiness in certain swap e2e tests (tested in other branches)
- [x] manually verified that the ConfirmSwapModal still works in the new permit2 flow with these changes

### Automated testing

- [x] Unit test N/A
- [x] Integration/E2E test (existing test flakiness in other PRs is fixed by this change)

### DIFF
--- a/src/pages/Swap/index.tsx
+++ b/src/pages/Swap/index.tsx
@@ -417,10 +417,20 @@ export function Swap({
     if (stablecoinPriceImpact && !confirmPriceImpactWithoutFee(stablecoinPriceImpact)) {
       return
     }
-    setSwapState({ attemptingTxn: true, tradeToConfirm, showConfirm, swapErrorMessage: undefined, txHash: undefined })
+    setSwapState((currentState) => ({
+      ...currentState,
+      attemptingTxn: true,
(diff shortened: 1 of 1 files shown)


zzmp (approved):

vm (approved):

tinaszheng:
nice fix!

PR #6523, opened 2023-05-10
docs(e2e): cypress README

zzmp:
## Description

Documents cypress testing:
- [x] Running e2e tests
- [x] Writing e2e tests
- [x] Working with hardhat
- [x] Best practices

## Test plan
- [x] N/A

PR #6625, opened 2023-05-23
fix: only remove expired transactions from updater

zzmp:
## Description

Instead of removing every tx except the latest from pending tx's, this removes tx's from pending tx's if they are expired (if they have a deadline) or are more than 6h old. This fixes flakiness in e2e tests, because it makes the transaction popup notification work reliably.

This is a hodge-podge PR, because flakiness wasn't so simple as a small fix:
- Logs and aliases requests to hardhat, and waits for them in swap tests
  This prevents flakiness by adding extra timeout time in case the hardhat node is backed up
- Improves popupList - a nop
  This isn't a cause of flakes, but I thought it was at one point, so I improved the impl/test
- Fixes the retry backoff logic in the transaction updater, and removes transactions based on age, not nonce
  With this fixed, I was also able to re-enable popup assertions in swap tests
- Disables video compression when uploading failures to Cypress
  This should help us identify sources of failures/flakes in the future

## Test plan

- Updated e2e tests to be more comprehensive, now that flaky behavior is fixed.

I'll continue to re-run this a few times to look for new flakes it may cause.
Some flakes are expected (there are still hardhat timeouts), and are caught early enough that they don't overly prolong the tests, so they can be ignored.


src/pages/Swap/index.tsx
```diff
@@ -410,53 +410,59 @@ export function Swap({
     allowance.state === AllowanceState.ALLOWED ? allowance.permitSignature : undefined
   )
 
-  const handleSwap = useCallback(() => {
+  const handleSwap = useMemo(() => {
```
zzmp, line 413: This is changed to `useMemo` so that we don't have a nop callback when it's not ready to render, but we just have `null`, and can disable the button based on that value.

cypress/e2e/swap/errors.test.ts
```diff
@@ -13,12 +13,16 @@ describe('Swap errors', () => {
...
-      // Attempt to swap.
-      cy.get('#swap-currency-output .token-amount-input').clear().type('1').should('have.value', '1')
+      // Enter amount to swap
+      cy.get('#swap-currency-output .token-amount-input').type('1').should('have.value', '1')
       cy.get('#swap-currency-input .token-amount-input').should('not.have.value', '')
+
+      // Submit transaction
       cy.get('#swap-button').click()
       cy.get('#confirm-swap-or-send').click()
+      cy.wait('@eth_estimateGas')
```
JFrankfurt, line 23: tell me about this

cypress.config.ts
```diff
@@ -9,6 +9,7 @@ export default defineConfig({
   chromeWebSecurity: false,
   experimentalMemoryManagement: true, // better memory management, see https://github.com/cypress-io/cypress/pull/25462
   retries: { runMode: 2 },
+  videoCompression: false,
```
JFrankfurt, line 12: what are the tradeoffs here? lower cpu utilization vs bandwidth cost? (are we charged for bandwidth?)

cypress/e2e/swap/errors.test.ts
```diff
@@ -29,32 +33,44 @@ describe('Swap errors', () => {
...
-      cy.hardhat().then((hardhat) => hardhat.mine(1, /* 10 minutes */ 1000 * 60 * 10)) // mines past the deadline
-      cy.get(getTestSelector('web3-status-connected')).should('not.contain', 'Pending')
-
-      // TODO(WEB-2085): Fix this test - transaction popups are flakey.
-      // cy.get(getTestSelector('transaction-popup')).contains('Swap failed')
+      cy.hardhat().then(async (hardhat) => {
+        // Remove the transaction from the mempool, so that it doesn't fail but it is past the deadline.
+        // This should result in it being removed from pending transactions, without a failure notificiation.
+        const transactions = await hardhat.send('eth_pendingTransactions', [])
+        console.log(transactions)
```
JFrankfurt, line 63: remove console.log

cypress/e2e/swap/errors.test.ts
```diff
@@ -74,41 +90,32 @@ describe('Swap errors', () => {
...
-        .type(AMOUNT_TO_SWAP.toString())
-        .should('have.value', AMOUNT_TO_SWAP.toString())
-      cy.get('#swap-currency-output .token-amount-input').should('not.have.value', '')
-      cy.get('#swap-button').click()
-      cy.get('#confirm-swap-or-send').click()
-      cy.get(getTestSelector('dismiss-tx-confirmation')).click()
-
-      // The pending transaction indicator should reflect the state.
+      // Submit 2 transactions
+      for (let i = 0; i < 2; i++) {
```
JFrankfurt, line 97: 😱 you're breaking your own rules mate!

imo this is an exception where it's prob ok

cypress/e2e/swap/swap.test.ts
```diff
@@ -37,33 +37,40 @@ describe('Swap', () => {
       cy.get('#swap-currency-input .token-amount-input').should('have.value', '')
     })
 
-    it('swaps ETH for USDC', () => {
+    it.only('swaps ETH for USDC', () => {
```
JFrankfurt, line 40:
```suggestion
    it('swaps ETH for USDC', () => {
```

src/lib/hooks/transactions/retry.ts
```diff
@@ -6,21 +6,15 @@ function waitRandom(min: number, max: number): Promise<void> {
...
- */
-class CancelledError extends Error {
-  public isCancelledError = true as const
-  constructor() {
-    super('Cancelled')
-  }
+/** Thrown if the function is canceled before resolving. */
+export class CanceledError extends Error {
+  name = 'CanceledError'
+  message = 'Retryable was canceled'
```
JFrankfurt, line 12: this diff seems fine but would have preferred it go in standalone

JFrankfurt: seems unrelated afaict

JFrankfurt (approved):

cypress/e2e/swap/errors.test.ts
zzmp: Calls to the wallet are now aliased, using `@${METHOD_CALLED}`, so this waits for the wallet to call `eth_estimateGas`.
Adding this throughout the `swap/` tests makes the tests more comprehensive, but also gives hardhat more time to resolve these requests, because each time we call `cy.wait` the test is given a 5s request timeout and 30s response timeout.

src/lib/hooks/transactions/retry.ts
zzmp: it's related in that this PR touches the consumer of the retry logic, but that's a stretch. but it feels harder to extricate, because the consumer is coupled to the implementation, so the new PR would necessarily have merge conflicts once this one went in.

cypress.config.ts
zzmp: afaict cypress does not charge for storage, so the tradeoff is in github actions. I have no transparency into github actions, but I'd strongly prefer better videos of failures so it's easier to debug them.

cypress/e2e/swap/wrap.test.ts
```diff
@@ -12,39 +12,36 @@ describe('Swap wrap', () => {
   })
 
   it('ETH to wETH is same value (wrapped swaps have no price impact)', () => {
-    cy.get('#swap-currency-input .token-amount-input').clear().type('0.01').should('have.value', '0.01')
+    cy.get('#swap-currency-input .token-amount-input').type('0.01').should('have.value', '0.01')
```
cmcewen, line 15: is this due to the cypress upgrade autoreloading?

src/hooks/useSwapCallback.tsx
```diff
@@ -30,33 +30,32 @@ export function useSwapCallback(
 
   const callback = useMemo(() => {
     if (!trade || !swapCallback) return null
+    const info: TransactionInfo =
+      trade.tradeType === TradeType.EXACT_INPUT
```
cmcewen, line 34: can you DRY this block a bit?

cypress/e2e/swap/errors.test.ts
zzmp: The click should resolve immediately, and `wait` detects the request sent out. It's the response that hangs.
This should be fixed by caching hardhat archive calls, done in #6650.

cypress.config.ts
zzmp: this only impacts the final step, uploading tests, so it only affects runtime if there are flakes or failures.

cypress/e2e/swap/errors.test.ts
zzmp: rules are in flux, but will be documented in a README: #6523.

cmcewen (approved):

PR #6627, opened 2023-05-23
style: navbar ui updated

Shubham-Rasal:
## Description

Modified the margin of the navIcon which when hovered was very close to the search bar.

## Screen capture

 Before       

interface-before-web
<br>
After (Desktop)
<br>
interface-after-web 

 

### Automated testing

- [x] Unit test
- [x] Integration/E2E test


src/components/NavBar/NavIcon.css.ts
```diff
@@ -15,6 +15,7 @@ export const navIcon = style([
     cursor: 'pointer',
     borderRadius: '8',
     transition: '250',
+    margin: '4',
```
cbachmeier, line 18: `margin-right` would be better for this purpose. Having the top-margin un-aligns the menu button
Screen Shot 2023-05-23 at 10 26 05

cbachmeier (changes requested):

src/components/NavBar/NavIcon.css.ts
Shubham-Rasal: On it

src/components/NavBar/NavIcon.css.ts
```diff
@@ -15,6 +15,7 @@ export const navIcon = style([
     cursor: 'pointer',
     borderRadius: '8',
     transition: '250',
+    marginRight: '4',
```
cbachmeier, line 18: This changes the style of the generic `NavIcon` component which is also shared with the NFT Shopping bag icon as well as the overflow menu. Updating the NFT Shopping bag is unintended. But you can move this style change one level up to the flex `Box` that wraps this `NavIcon` in `MenuDropdown.tsx`

cbachmeier (changes requested):

src/components/NavBar/NavIcon.css.ts
Shubham-Rasal: That should work , will try it.

Shubham-Rasal: Added the margin to the box.

cbachmeier (approved):
LGTM 👍

Shubham-Rasal:
Review please @tinaszheng

PR #6628, opened 2023-05-23
fix: Simplify event logging for SWAP_QUOTE_RECEIVED

tinaszheng:
## Description
The previous version of the hook was infinite looping when `trade` is always defined, so simplify logic to just keep track of `trade` changes. Also removes the time-based logging since we have stats on the backend now

## Screen capture
No UI changes

### Reproducing the error
Throw in a console.log right before `sendAnalyticsEvent` and see it firing like 1 million times lol

### DIFF
--- a/src/pages/Swap/index.tsx
+++ b/src/pages/Swap/index.tsx
@@ -177,8 +177,6 @@ export function Swap({
   disableTokenInputs?: boolean
 }) {
   const { account, chainId: connectedChainId, connector } = useWeb3React()
-  const [newSwapQuoteNeedsLogging, setNewSwapQuoteNeedsLogging] = useState(true)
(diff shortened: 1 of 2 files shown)


src/pages/Swap/index.tsx
```diff
@@ -526,42 +524,16 @@ export function Swap({
...
-    routeIsSyncing,
-    routeIsLoading,
-    fetchingSwapQuoteStartTime,
-    trade,
-    setSwapQuoteReceivedDate,
-    trace,
-  ])
+    if (!trade || prevTrade === trade) return // no new swap quote to log
+
+    setSwapQuoteReceivedDate(new Date())
```
tinaszheng, line 531: this property is still used downstream in child components, so not getting rid of it for now

lynnshaoyu (approved):

PR #6629, opened 2023-05-23
fix: allow unsupported chain in pwat

JackShort:
this logic was removed in #6127 which caused the bag not to fail properly when switching to an unsupported chain

### Automated testing

- [x] Unit test


tinaszheng (approved):

cbachmeier (approved):
LGTM

PR #6631, opened 2023-05-23
fix: increase useBestTrade debounce time

vm:
## Description
Increases the debounce time on fetching quotes from 200ms to 600ms. This reduces the number of requests we make to the routing-api to get quotes, which reduces our vendor usage.

## Screen capture
| 
Screen Shot 2023-05-23 at 2 00 53 PM
 | 
Screen Shot 2023-05-23 at 2 01 04 PM
      |    |

### Reproducing the error
1. Type in values on the swap input in quick succession
2. See the number of requests in the network tab while filtered to 'quote'

### Automated testing
- [ ] Unit test
- [ ] Integration/E2E test

N/A I think but I can add something if this is the right direction

### DIFF
--- a/src/hooks/useBestTrade.ts
+++ b/src/hooks/useBestTrade.ts
@@ -31,7 +31,7 @@ export function useBestTrade(
 
   const [debouncedAmount, debouncedOtherCurrency] = useDebounce(
     useMemo(() => [amountSpecified, otherCurrency], [amountSpecified, otherCurrency]),
-    200
+    600
   )
 
   const isAWrapTransaction = useMemo(() => {


src/hooks/useBestTrade.ts
```diff
@@ -31,7 +31,7 @@ export function useBestTrade(
 
   const [debouncedAmount, debouncedOtherCurrency] = useDebounce(
     useMemo(() => [amountSpecified, otherCurrency], [amountSpecified, otherCurrency]),
-    200
+    600
```
tinaszheng, line 34: can you move this to a constant

vm: at the top of the file or in a constants file?

src/hooks/useBestTrade.ts
```diff
@@ -11,6 +11,9 @@ import { useClientSideV3Trade } from './useClientSideV3Trade'
 import useDebounce from './useDebounce'
 import useIsWindowVisible from './useIsWindowVisible'
 
+// Adds a debounce to the trade request to prevent excessive requests between keystrokes.
```
zzmp, line 14: nit: "// The debounce added to..."

zzmp (approved):
Is there a plan to improve analytics around this? How will we know that it "worked"?

vm:
Without a way to A/B test this, really the only thing I can imagine is to see if the chart goes down in Infura. If we want to be more specific, we could do this under a feature flag and enable for a specific hour?

Beyond that, we'd need to set up tracing for JsonRpcProvider.send and also log the requests somewhere. I think we're kind of far off from that happening.

just-toby (approved):

vm:
Depending on convo with Andy we may add a feature flag on this, but would rather get this merged incase that change can't happen by tomorrow.

PR #6633, opened 2023-05-23
feat: dismiss the landing page when the account drawer is opened

JFrankfurt:
## Description

Dismisses the landing page when the account drawer is opened

_JIRA ticket:_ https://linear.app/uniswap/issue/WEB-2088/dismiss-landing-page-when-connection-drawer-opens

### Before
image

### After
image

### DIFF
--- a/cypress/e2e/link.test.ts
+++ b/cypress/e2e/link.test.ts
@@ -2,7 +2,7 @@
 describe('Link', () => {
   it('should update route', () => {
     cy.viewport(2000, 1600)
-    cy.visit('/')
+    cy.visit('/swap')
     cy.contains('Pool').click()
     cy.get('[data-cy="join-pool-button"]').should('exist')
   })


src/components/AccountDrawer/index.tsx
```diff
@@ -18,7 +18,7 @@ const DRAWER_MARGIN = '8px'
 const DRAWER_OFFSET = '10px'
 const DRAWER_TOP_MARGIN_MOBILE_WEB = '72px'
 
-const accountDrawerOpenAtom = atom(false)
+export const accountDrawerOpenAtom = atom(false)
```
zzmp, line 21: I'd rather this not be exported, because it's safer to wrap it in helpers. In this case, you can use `const [accountDrawerOpen] = useAccountDrawer()` for the same functionality.

zzmp (changes requested):
Can you pass this by @infredible? The transition is very abrupt.

src/components/AccountDrawer/index.tsx
infredible: This would be better if the landing page content faded out. Opacity change easeInOut at the same speed of the drawer animation. I think it might be 250ms

just-toby (approved):

src/pages/Landing/index.tsx
```diff
@@ -18,6 +19,7 @@ import { useAppSelector } from 'state/hooks'
 import styled, { css } from 'styled-components/macro'
 import { BREAKPOINTS } from 'theme'
 import { useIsDarkMode } from 'theme/components/ThemeToggle'
+import { TRANSITION_DURATIONS } from 'theme/styles'
 import { Z_INDEX } from 'theme/zIndex'
 
 const PageContainer = styled.div<{ isDarkMode: boolean }>`
```
cbachmeier, line 25: Background removal seems unrelated but if it was meant to be removed, looks like `isDarkMode` is no longer used by this component

cbachmeier (approved):
LGTM

src/components/NavBar/index.tsx
```diff
@@ -91,6 +92,18 @@ const Navbar = ({ blur }: { blur: boolean }) => {
   const sellPageState = useProfilePageState((state) => state.state)
   const navigate = useNavigate()
 
+  const [accountDrawerOpen, toggleAccountDrawer] = useAccountDrawer()
+
+  const handleUniIconClick = useCallback(() => {
```
just-toby, line 97: we could remove `useCallback` here right?

just-toby (approved):

PR #6638, opened 2023-05-24
build: set null environment for PRs

zzmp:
## Description

Uses a "ternary" statement to set a null environment for the test.yml workflow.

This is necessary to avoid "temporarily deployed to false" commit comments on the PR, as described in https://github.com/orgs/community/discussions/36919#discussioncomment-4110859.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     environment:
-      name: ${{ github.ref_name == 'main' && 'notify/test' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -17,7 +17,7 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     environment:
-      name: ${{ github.ref_name == 'main' && 'notify/test' }}
+      name: ${{ github.ref_name == 'main' && 'notify/test' || null }}
     steps:
       - uses: actions/checkout@v3
```
Joci1980, line 22: -

PR #6639, opened 2023-05-24
build: use repository slack secret

zzmp:
## Description

Uses the repo secret for the slack test reporter webhook, instead of using an environment.

This is necessary to avoid "temporarily deployed to false" commit comments on the PR, as described in https://github.com/orgs/community/discussions/36919#discussioncomment-4110859, which are caused by using an environment for a workflow.


AndrewMohawk (approved):
LGTM to improve dev-x

JFrankfurt (approved):

PR #6640, opened 2023-05-24
fix: no-undefined-or in object field types

just-toby:
## Description

adds a new eslint rule to replace object fields with types like this:

`val: Type | undefined`

with types like this:

`val?: Type`

### QA (ie manual testing)

I manually inspected every correction to ensure that it caught all edge cases and the auto-fixer generated well formatted results

### DIFF
--- a/eslint_rules/no-undefined-or.js
+++ b/eslint_rules/no-undefined-or.js
@@ -1,0 +1,44 @@
+/* eslint-env node */
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Enforce the use of optional object fields',
+      category: 'Best Practices',
+      recommended: true,
+    },
+    fixable: 'code',
+    schema: [],
+  },
+  create(context) {
+    return {
+      'TSPropertySignature > TSTypeAnnotation > TSUnionType': (node) => {
+        const types = node.types
+        const hasUndefined = types.some((typeNode) => typeNode.type === 'TSUndefinedKeyword')
+
+        if (hasUndefined) {
+          const typesWithoutUndefined = types.filter((typeNode) => typeNode.type !== 'TSUndefinedKeyword')
+
+          // If there is more than one type left after removing 'undefined',
+          // join them together with ' | ' to create a new union type.
+          const newTypeSource =
+            typesWithoutUndefined.length > 1
+              ? typesWithoutUndefined.map((typeNode) => context.getSourceCode().getText(typeNode)).join(' | ')
+              : context.getSourceCode().getText(typesWithoutUndefined[0])
+
+          context.report({
+            node,
+            message: `Prefer optional properties to "Type | undefined".`,
+            fix(fixer) {
+              const propertySignature = node.parent.parent
+              const isAlreadyOptional = propertySignature.optional
+              const newTypeAnnotation = isAlreadyOptional ? `: ${newTypeSource}` : `?: ${newTypeSource}`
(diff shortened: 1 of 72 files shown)


vm:
I'm open minded here, but I think `arg?` and `arg: T | undefined` are different because the latter enforces that an argument was passed in

JackShort:
> I'm open minded here, but I think `arg?` and `arg: T | undefined` are different because the latter enforces that an argument was passed in

I agree with this it allows you to be verbose in what the input should be into a function. Also I feel like `arg?` is used mostly for flag like data being taken in - ie the function is being used across components and one component might want a slightly different output from the function. While `arg: T | undefined` is used for when an input value might actually be `undefined` when passed in and you are kinda creating a type for that.

tinaszheng:
> I'm open minded here, but I think `arg?` and `arg: T | undefined` are different because the latter enforces that an argument was passed in
> 
> A rule to make sure that `arg?: T | undefined` can't happen is certainly useful. less sold on the full rule but don't have a strong opinion

i believe you're talking about function arguments? this rule applies to object types only

tinaszheng (approved):
<3

eslint_rules/no-undefined-or.js
```diff
@@ -0,0 +1,44 @@
...
+          // If there is more than one type left after removing 'undefined',
+          // join them together with ' | ' to create a new union type.
+          const newTypeSource =
+            typesWithoutUndefined.length > 1
+              ? typesWithoutUndefined.map((typeNode) => context.getSourceCode().getText(typeNode)).join(' | ')
+              : context.getSourceCode().getText(typesWithoutUndefined[0])
+
+          context.report({
+            node,
+            message: `Prefer optional properties to "Type | undefined".`,
```
grabbou, line 32: Great work @just-toby, quick question - can we ship this as a part of Uniswap/eslint-preset? I have access to that project, so can release a new version. Would add it to the `node` config so it's applied everywhere.

grabbou:
My honest question and the reason behind this rule (actually) is whether we are actually aware of the difference between `?` and `| undefined` on a daily basis and whether it actually matters (over creating confusion). I have asked around while doing PR reviews in the past and I don't think it was that clear. Most of the time, it was assumed that both variants were interchangeable. With that in mind, I think it may make more sense to keep it more consistent.

I think that `| undefined` was initially used in our codebase as a pure styling preference, not a constraint. I agree that there's difference between both, it's just that the fact we're using `| undefined` without it being actually required for the code to work is confusing.

JackShort:
> @vig @JackShort My honest question and the reason behind this rule (actually) is whether we are actually aware of the difference between `?` and `| undefined` on a daily basis and whether it actually matters (over creating confusion). I have asked around while doing PR reviews in the past and I don't think it was that clear. Most of the time, it was assumed that both variants were interchangeable. With that in mind, I think it may make more sense to keep it more consistent.
> 
> I think that `| undefined` was initially used in our codebase as a pure styling preference, not a constraint. I agree that there's difference between both, it's just that the fact we're using `| undefined` without it being actually required for the code to work is confusing.

Sorry, somehow I missed the word `object` in the title of this pr and didn't realize it was only for object types - that's completely different and that's my bad. I agree with dropping it because the inconsistency does hurt my brain

zzmp (approved):
@grabbou: I'd love for this to be part of `Uniswap/eslint-config`, but I think a good first step is merging it into `Uniswap/interface`. `interface` already builds on `eslint-config`, so we should be migrating upstream regularly anyways.

PR #6651, opened 2023-05-25
build: disable scheduled releases

zzmp:
## Description

Disables the scheduled release (M-Th), in favor of our new release pipeline.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,7 +1,5 @@
 name: Release
 on:
-  schedule:
-    - cron: '0 16 * * 1-4' # every day 16:00 UTC Monday-Thursday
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -1,7 +1,5 @@
 name: Release
 on:
```
cbachmeier, line 2: Will the adding of a cron job to `Push staging -> prod` on Wednesdays be a separate task?

tinaszheng (approved):

.github/workflows/release.yaml
zzmp: It won't be done as a cron. The build sheriff will kick it off manually every week.

PR #6654, opened 2023-05-25
feat: log swap button click

just-toby:
## Description

logs an event to amplitude when the top-level swap button is clicked. previously we only logged the clicks on the "confirm swap" button in the subsequent modal

## Screen capture
image

### QA (ie manual testing)

- [x] clicked the button, viewed the log

### DIFF
--- a/src/pages/Swap/index.tsx
+++ b/src/pages/Swap/index.tsx
@@ -6,6 +6,7 @@ import {
   InterfaceEventName,
   InterfacePageName,
   InterfaceSectionName,
+  SharedEventName,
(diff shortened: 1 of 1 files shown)


adjkant (approved):

zzmp (approved):

just-toby:
> why TraceEvent instead of a sendAnalytics call in the click handler?

i think they both work, but TraceEvent gets the analytics context automatically, while sometimes calls to `sendAnalyticsEvent` need manual merging of the context into the object.

PR #6656, opened 2023-05-26
test(e2e): use aliases for permit2 test

zzmp:
## Description

Adds aliasing to Permit2 tests.

- Expands scope of initial test by turning off automining and asserting on intermediate states
- Eliminates redundant test


cypress/e2e/permit2.test.ts
```diff
@@ -5,212 +5,187 @@ import { getTestSelector } from '../utils'
...
+    // Verify transaction
+    cy.contains('Success')
+    cy.get(getTestSelector('popups')).contains('Swapped')
   })
 
   it('swaps after completing full permit2 approval process', () => {
-    cy.hardhat().then(({ provider }) => {
-      cy.spy(provider, 'send').as('permitApprovalSpy')
-    })
+    cy.hardhat().then((hardhat) => hardhat.setAutomine(false))
```
zzmp, line 64: This appears to be flaky - sometimes the subsequent commands will get in before the command to toggle automine. See https://cloud.cypress.io/projects/yp82ef/runs/11453/overview/bcb08413-dfde-4d9d-8ddd-5878a4e3d021?roarHideRunsWithDiffGroupsAndTags=1.

zzmp: I've tried to refactor it to what I _hope_ will be less flaky, but I'll also investigate how cypress-hardhat works to see if it's misconfigured.

cypress/e2e/permit2.test.ts
```diff
@@ -5,212 +5,193 @@ import { getTestSelector } from '../utils'
...
+      cy.hardhat().then((hardhat) => hardhat.mine())
+      cy.get(getTestSelector('popups')).contains('Approved')
+      expectTokenAllowanceForPermit2ToBeMax()
 
-    cy.then(() => {
-      const approvalTime = Date.now()
+      // Verify permit2 approval
+      cy.contains('Allow DAI to be used for swapping')
+      cy.wait('@eth_signTypedData_v4')
+      cy.contains('Confirm Swap').click()
```
just-toby, line 66: this will only appear if the quote price changed beyond a certain threshold

just-toby: it actually never appeared in any test runs i've seen so far. i know we're slowing down the test but i'm not sure we should assume it will appear

just-toby (changes requested):

cypress/e2e/permit2.test.ts
zzmp: This will appear if there was _not_ a price update, I think.
See https://github.com/Uniswap/interface/blob/e2dd78fd0e05e91c5eb286f754404bbd81e80905/src/components/swap/SwapModalFooter.tsx#L156-L204

zzmp: we're both wrong :(
this is the final review screen, which happens to share the same text

just-toby (dismissed):

cypress/e2e/permit2.test.ts
```diff
@@ -5,212 +5,192 @@ import { getTestSelector } from '../utils'
...
   const OUTPUT_TOKEN = USDC_MAINNET
-  const TEST_BALANCE_INCREMENT = 0.01
 
   beforeEach(() => {
     // Sets up a swap between INPUT_TOKEN and OUTPUT_TOKEN.
     cy.visit(`/swap/?inputCurrency=${INPUT_TOKEN.address}&outputCurrency=${OUTPUT_TOKEN.address}`, {
       ethereum: 'hardhat',
     })
-    cy.get('#swap-currency-input .token-amount-input').type(TEST_BALANCE_INCREMENT.toString())
+    cy.get('#swap-currency-input .token-amount-input').type('0.01')
```
cmcewen, line 25: is this a stylistic choice?

cypress/e2e/permit2.test.ts
```diff
@@ -5,212 +5,192 @@ import { getTestSelector } from '../utils'
...
   })
 
   /** Asserts permit2 has a max approval for spend of the input token on-chain. */
   function expectTokenAllowanceForPermit2ToBeMax() {
     // check token approval
-    return cy
-      .hardhat()
+    cy.hardhat()
```
cmcewen, line 31: does prettier not run on this files?

cmcewen (approved):

cypress/e2e/permit2.test.ts
zzmp: I think this _was_ done by prettier, as a result of the `return` being removed.

cypress/e2e/permit2.test.ts
zzmp: yes, I think using explicit literals makes it easier to tell what your test is trying to do.
I should add this to the https://github.com/Uniswap/interface/blob/main/cypress/README.md.

cmcewen (approved):

just-toby (approved):

PR #6662, opened 2023-05-27
test(e2e): switching network

zzmp:
## Description

Adds integration tests for chain switching.

Fixes a bug where slow chain switching might show a disconnected state:
- Creates a new `state.wallets.switchingChain` state which is set during a chain switch
- Disables the account icon and displays a message on the Swap bar during a chain switch
- Attempts to reconnect to the chain after a failed chain switch

As part of the new chain switching intermediate state:
- Refactored `switchChain` to a hook, so it could appropriately `useAppDispatch`
- Refactored status icons to take account, as the account is unavailable during a switch and must be passed outside of web3-react

### Before

https://github.com/Uniswap/interface/assets/5403956/a46fad35-89df-4052-a5ce-a50610244213

### After
| paste_after  | image |

## Test plan

You can view the new screen by promoting a switch in the interface while connected. Before approving the switch in your wallet, view the interface to see the changes. 

### Reproducing the error
On MetaMask, switch the chain repeatedly. Some intermediate screens will show a lost connection.

### Automated testing

- [x] Unit test
- [x] Integration/E2E test


zzmp:
@SocketSecurity ignore cypress-hardhat@2.4.0

JFrankfurt (changes requested):
note the flake

cypress/e2e/wallet-connection/switch-network.test.ts
```diff
@@ -0,0 +1,124 @@
+import { createDeferredPromise } from '../../../src/test-utils/promise'
+import { getTestSelector } from '../../utils'
+
+function getChainSelector(activeChain: string) {
+  cy.get(getTestSelector('chain-selector-logo')).invoke('attr', 'alt').should('eq', activeChain)
```
JFrankfurt, line 5: A `should` assertion feels weird to me in a `get` function, but I can see a use for it, given cypress's "wait until the subject exists" behavior.

JFrankfurt: maybe `waitForConnection` ?

cypress/e2e/wallet-connection/switch-network.test.ts
```diff
@@ -0,0 +1,124 @@
...
+  })
+
+  it('should not disconnect while switching', () => {
+    const promise = createDeferredPromise()
+
+    cy.hardhat().then((hardhat) => {
+      // Reject network switch with CHAIN_NOT_ADDED
+      const switchChainStub = cy.stub(hardhat.provider, 'send').log(false).as('switch')
+      switchChainStub.withArgs('wallet_switchEthereumChain').returns(promise)
+      switchChainStub.callThrough() // allows other calls to return non-stubbed values
```
JFrankfurt, line 86:
```suggestion
      const sendStub = cy.stub(hardhat.provider, 'send').log(false).as('switch')
      sendStub.withArgs('wallet_switchEthereumChain').returns(promise)
      sendStub.callThrough() // allows other calls to return non-stubbed values
```

cypress/e2e/wallet-connection/switch-network.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    cy.hardhat().then((hardhat) => {
+      // https://docs.metamask.io/guide/rpc-api.html#unrestricted-methods
+      const CHAIN_NOT_ADDED = { code: 4902 } // missing message in useSelectChain
+
+      // Reject network switch with CHAIN_NOT_ADDED
+      const switchChainStub = cy.stub(hardhat.provider, 'send').log(false).as('switch')
+      let added = false
+      switchChainStub
+        .withArgs('wallet_switchEthereumChain')
+        .callsFake(() => (added ? Promise.resolve(null) : Promise.reject(CHAIN_NOT_ADDED)))
```
JFrankfurt, line 54: I think these stubs should be named after the method being stubbed, not the `withArgs` provided to the stub

zzmp:
> note the flake

I think it's due to load on the machine, with two separate forked chain servers running, but I am not sure. I only see it in the one spot (waiting for a hardhat method), and I've never seen it actually fail, so I think it is necessary for now :/

zzmp:
These tests are still a little flaky, and I think it has to do with # of pending network requests from the browser. in a network switch scenario the eth_chainId's are not cached, and so they stack up and prevent the switch request from going through at all.

I'm going to try to fix this in a fast-follow to avoid adding complexity to this (already large) PR, but I'll avoid merging this until both are ready.

zzmp:
> These tests are still a little flaky, and I think it has to do with # of pending network requests from the browser. in a network switch scenario the eth_chainId's are not cached, and so they stack up and prevent the switch request from going through at all.
> 
> I'm going to try to fix this in a fast-follow to avoid adding complexity to this (already large) PR, but I'll avoid merging this until both are ready.

I believe this will be fixed in https://github.com/Uniswap/cypress-hardhat/pull/20/files. Once that is merged I'll update this PR.
This PR should be considered ready for review, though. The only thing that will be changing is a dependency.

src/components/Identicon/index.tsx
```diff
@@ -18,8 +17,7 @@ const StyledAvatar = styled.img`
   border-radius: inherit;
 `
 
-export default function Identicon({ size }: { size?: number }) {
-  const { account } = useWeb3React()
+export default function Identicon({ account, size }: { account: string; size?: number }) {
```
JFrankfurt, line 20: why?

zzmp: We must pass the account in order for the memoized account to be available when rendering during a chain switch, because technically the account is not connected during that switch, so `useWeb3React()` will not return one.

See https://github.com/Uniswap/interface/pull/6662/files#diff-72207ac383212f07b07e35204f4ee12dcee52d563af7ed7f4e3c669a6544a57dR137, where the account is memoized using `useLast` in case a chain switch occurs.

src/components/Web3Status/index.tsx
```diff
@@ -205,11 +210,13 @@ function Web3StatusInner() {
...
-      </Portal>
-    </PrefetchBalancesWrapper>
+    <div data-testid="web3-status">
+      <PrefetchBalancesWrapper>
+        <Web3StatusInner />
+        <Portal>
+          <PortfolioDrawer />
+        </Portal>
+      </PrefetchBalancesWrapper>
+    </div>
```
JFrankfurt, line 220: this reads like you're adding the div just so it can be used by a test to select the data-testid--can you find a way to test w/o modifying the app like this?

src/hooks/useSwitchChain.ts
```diff
@@ -0,0 +1,76 @@
...
+        } catch (error) {
+          // In activating a new chain, the connector passes through a deactivated state.
+          // If we fail to switch chains, it may remain in this state, and no longer be usable.
+          // We defensively re-activate the connector to ensure the user does not notice any change.
+          try {
+            await connector.activate()
+          } catch (error) {
+            console.error('Failed to re-activate connector', error)
+          }
+          throw error
```
JFrankfurt, line 68: this reads to me like the error will throw even if line 64 activates successfully--is that intended?

JFrankfurt (approved):

src/hooks/useSwitchChain.ts
zzmp: This is intentional - we are resetting the state, but not moving error handling into this function. The error (of not being able to switch) still must be handled at the callsite.

tinaszheng (approved):

PR #6664, opened 2023-05-30
fix: polygon bridge link leading to error page

LunrEclipse:
## Description

Fixes the Polygon token bridge link on the Swap page when connected to the Polygon Network. Originally, the link took users to an Error 404 page before redirecting them to the proper bridging page.

Now the link takes users directly to the proper bridging page without navigating to the Error 404 page. This was done by changing the `bridge` link in `chainInfo.ts` from "https://wallet.polygon.technology/login" to "https://wallet.polygon.technology"

_Linear ticket:_ WEB-2101

### Reproducing the error

1. Go to the Swap page and change network to Polygon
2. Click the "Polygon Token Bridge" banner

### QA (ie manual testing)

- [x] Link redirects users to proper bridge page without error

### Automated testing

- [x] Unit test N/A

Screenshot 2023-05-30 at 9 57 47 AM

Screenshot 2023-05-30 at 9 57 56 AM

### DIFF
--- a/src/constants/chainInfo.ts
+++ b/src/constants/chainInfo.ts
@@ -148,7 +148,7 @@ const CHAIN_INFO: ChainInfoMap = {
   [SupportedChainId.POLYGON]: {
     networkType: NetworkType.L1,
     blockWaitMsBeforeWarning: ms`10m`,
-    bridge: 'https://wallet.polygon.technology',
+    bridge: 'https://wallet.polygon.technology/polygon/bridge',
     docs: 'https://polygon.io/',
     explorer: 'https://polygonscan.com/',
     infoLink: 'https://info.uniswap.org/#/polygon/',
--- a/src/constants/chainInfo.ts
+++ b/src/constants/chainInfo.ts
@@ -148,7 +148,7 @@ const CHAIN_INFO: ChainInfoMap = {
   [SupportedChainId.POLYGON]: {
     networkType: NetworkType.L1,
     blockWaitMsBeforeWarning: ms`10m`,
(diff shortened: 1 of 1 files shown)


src/constants/chainInfo.ts
```diff
@@ -148,7 +148,7 @@ const CHAIN_INFO: ChainInfoMap = {
   [SupportedChainId.POLYGON]: {
     networkType: NetworkType.L1,
     blockWaitMsBeforeWarning: ms`10m`,
-    bridge: 'https://wallet.polygon.technology/login',
```
cbachmeier, line 151: Thanks for catching this! We should probably link directly to the POS Bridge here instead of the zkEVM to avoid confusion until that's supported. Here is the direct link to the POS Bridge: https://wallet.polygon.technology/polygon/bridge

cbachmeier (changes requested):

src/constants/chainInfo.ts
LunrEclipse: For some reason that link still redirects me to the zkEVM bridge, however, I just updated the link in case that is just an issue on my end.

cbachmeier (approved):
LGTM. You'll need 2 reviewers, can you also add @Uniswap/web-reviewers for that second review?

tinaszheng (approved):

src/constants/chainInfo.ts
cmcewen: yeah i'm also getting redirected here

cmcewen (approved):

PR #6666, opened 2023-05-30
feat: increased debounce swap quote rate

vm:
## Description
Allows us to run a production test of the debounce rate to see if it decreases RPC usage. This is based on seeing positive results from https://github.com/Uniswap/interface/pull/6631 and wanting more data to determine the proper UX/cost tradeoff.

### QA (ie manual testing)
- [x] Update inputs on swap
- [x] Make sure that quotes show up if typing quickly

### Automated testing
- [ ] Unit test
- [ ] Integration/E2E test

N/A


src/featureFlags/index.tsx
```diff
@@ -1,7 +1,18 @@
 import { atomWithStorage, useAtomValue, useUpdateAtom } from 'jotai/utils'
 import { createContext, ReactNode, useCallback, useContext } from 'react'
 import { useGate } from 'statsig-react'
-export { FeatureFlag } from './flags/featureFlags'
+
```
vm, line 4: I moved this here because a separate file seemed unnecessary

cmcewen (approved):

src/hooks/useBestTrade.ts
```diff
@@ -14,6 +15,9 @@ import useIsWindowVisible from './useIsWindowVisible'
 // Prevents excessive quote requests between keystrokes.
 const DEBOUNCE_TIME = 350
 
+// Temporary until we remove the feature flag.
```
just-toby, line 18: if there is a task tracking this experiment, it may help to make this a TODO

`TODO (WEB-1234): clean up this experiment when finished`

src/hooks/useBestTrade.ts
```diff
@@ -34,7 +38,7 @@ export function useBestTrade(
 
   const [debouncedAmount, debouncedOtherCurrency] = useDebounce(
     useMemo(() => [amountSpecified, otherCurrency], [amountSpecified, otherCurrency]),
-    DEBOUNCE_TIME
+    useDebounceSwapQuoteFlag() === DebounceSwapQuoteVariant.Enabled ? DEBOUNCE_TIME_INCREASED : DEBOUNCE_TIME
```
just-toby, line 41: i wish there was a way to control the actual integer value here from statsig, but from what i can tell it's boolean only.

just-toby (approved):

src/hooks/useBestTrade.ts
vm: we could do it via an experiment but i didnt really wanna mess with that atm

src/hooks/useBestTrade.ts
vm: i think it's fine without bc typescript will complain if i remove the flag but leave the variable

PR #6670, opened 2023-05-31
fix: token bridge pointer bug

LunrEclipse:
## Description

Fixes the pointer bug when selecting the token bridge banner on the Swap page. Before the banner would not be selected if the pointer came from the left side. Now the banner is selected from all locations on the banner.

Accomplished by adding some CSS styling to the `LinkOutToBridge` tag in the `NetworkAlert` component. First set `z-index` to 1 due to some weird overlap between the `LinkOutToBridge` tag and the `ContentWrapper` tag. Also removed the `margin-right` as it was causing the link to not cover the entirety of the banner's area. 

_Linear ticket:_ Web-2102

### Before

https://github.com/Uniswap/interface/assets/35351983/24d40a41-be51-4c84-bb32-9911afeca78b

### After

https://github.com/Uniswap/interface/assets/35351983/eb7242d3-3f17-4d19-bb26-3b74bba42743

### Reproducing the error

1. Go to the Swap page and change network to Polygon
2. Try and hover over the banner from the left side

### QA (ie manual testing)

- [x] Test to make sure banner selected on all chains and that it redirects properly

### Automated testing

- [x] Unit test N/A

### DIFF
--- a/src/components/NetworkAlert/NetworkAlert.tsx
+++ b/src/components/NetworkAlert/NetworkAlert.tsx
@@ -7,6 +7,7 @@ import styled from 'styled-components/macro'
 import { ExternalLink, HideSmall } from 'theme'
 import { colors } from 'theme/colors'
 import { useDarkModeManager } from 'theme/components/ThemeToggle'
+import { Z_INDEX } from 'theme/zIndex'
 
 import { AutoRow } from '../Row'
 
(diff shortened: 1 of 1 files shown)


Mulligani777 (approved):

src/components/NetworkAlert/NetworkAlert.tsx
```diff
@@ -122,9 +122,9 @@ const LinkOutToBridge = styled(ExternalLink)`
   font-size: 16px;
   justify-content: space-between;
   padding: 6px 8px;
-  margin-right: 12px;
   text-decoration: none !important;
   width: 100%;
+  z-index: 1;
```
cmcewen, line 127: nit: could you use the enum from the `zIndex.ts` file?

LunrEclipse: Fixed!

vm (approved):

cmcewen (approved):

PR #6671, opened 2023-05-31
fix: compilation error on local build

JackShort:
Local build is running into compilation errors because of this import
Screen Shot 2023-05-31 at 3 18 36 PM


just-toby (approved):
this isn't causing an error on my machine, and it didn't affect CI/CD right? why is this necessary?

JackShort:
> this isn't causing an error on my machine, and it didn't affect CI/CD right? why is this necessary?

closing because this error seems to only affect my machine

PR #6680, opened 2023-06-01
feat: Add Circle native Arbitrum USDC to common base tokens in token selector

tinaszheng:
This PR adds native USDC on arbitrum behind a feature flag that should be flipped on June 8th, the launch date.

Note: We should remove the legacy logic once native USDC is fully launched :D 

Enabled state:
image

Note we don't have the token logo yet because it's not available as part of any token lists. But as soon as any of our token list providers (including our own) adds, it will be available :D

### DIFF
--- a/src/constants/routing.ts
+++ b/src/constants/routing.ts
@@ -4,6 +4,7 @@ import { SupportedChainId } from 'constants/chains'
 
 import {
   AMPL,
+  BRIDGED_USDC_ARBITRUM,
   BTC_BSC,
   BUSD_BSC,
   CAKE_BSC,
@@ -149,7 +150,7 @@ export const COMMON_BASES: ChainCurrencyList = {
   [SupportedChainId.ARBITRUM_ONE]: [
     nativeOnChain(SupportedChainId.ARBITRUM_ONE),
     DAI_ARBITRUM_ONE,
-    USDC_ARBITRUM,
+    BRIDGED_USDC_ARBITRUM,
     USDT_ARBITRUM_ONE,
     WBTC_ARBITRUM_ONE,
     WRAPPED_NATIVE_CURRENCY[SupportedChainId.ARBITRUM_ONE] as Token,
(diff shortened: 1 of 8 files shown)


src/constants/tokens.ts
```diff
@@ -39,13 +39,20 @@ const USDC_OPTIMISM_GOERLI = new Token(
...
   6,
   'USDC',
   'USD//C'
 )
+export const USDC_ARBITRUM = new Token(
+  SupportedChainId.ARBITRUM_ONE,
+  '0xaf88d065e77c8cC2239327C5EDb3A432268e5831',
+  6,
+  'USDC',
+  'USD//C'
```
tinaszheng, line 54: not sure about the token name here - pls double check for me what this is

src/constants/routing.ts
```diff
@@ -206,6 +207,19 @@ export const COMMON_BASES: ChainCurrencyList = {
...
+export const COMMON_BASES_V2: ChainCurrencyList = {
+  ...COMMON_BASES,
+  [SupportedChainId.ARBITRUM_ONE]: [
+    nativeOnChain(SupportedChainId.ARBITRUM_ONE),
+    DAI_ARBITRUM_ONE,
+    USDC_ARBITRUM,
+    USDT_ARBITRUM_ONE,
+    WBTC_ARBITRUM_ONE,
+    WRAPPED_NATIVE_CURRENCY[SupportedChainId.ARBITRUM_ONE] as Token,
+  ],
```
just-toby, line 220:
```suggestion
  [SupportedChainId.ARBITRUM_ONE]: [
    ...COMMON_BASES[SupportedChainId.ARBITRUM_ONE],
    USDC_ARBITRUM,
  ],
```

just-toby (approved):
we're only adding it to common bases, and not replacing the previous USDC everywhere? e.g. in USD price checking ?

src/constants/routing.ts
```diff
@@ -149,7 +150,7 @@ export const COMMON_BASES: ChainCurrencyList = {
   [SupportedChainId.ARBITRUM_ONE]: [
     nativeOnChain(SupportedChainId.ARBITRUM_ONE),
     DAI_ARBITRUM_ONE,
-    USDC_ARBITRUM,
+    BRIDGED_USDC_ARBITRUM,
```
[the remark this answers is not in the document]
cbachmeier: something like `useGatedArbUsdc` which depending on the flag will return the relevant usdc address

src/constants/routing.ts
tinaszheng: that would include the old USDC too

tinaszheng:
@just-toby we do price checking in ETH!

src/constants/routing.ts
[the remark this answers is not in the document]
tinaszheng: we'll delete one of them after launch so keeping like this for ease of migration later

cbachmeier (approved):

cmcewen (approved):

PR #6682, opened 2023-06-01
feat: MP closes on scroll down

LunrEclipse:
## Description

Adds features to have Mini Portfolio close when users on mobile scroll down.

Adds an event listener to `AccountDrawer` to see if a user swipes down and if so, will trigger `toggleWalletDrawer` to close the Mini Portfolio.

_Linear ticket:_ WEB-1801

### Before

https://github.com/Uniswap/interface/assets/35351983/b6f9e657-dd5a-4db7-b185-79768ba570b3

### After

https://github.com/Uniswap/interface/assets/35351983/a8f647fe-21d1-4328-bdaa-a6ef03fb7d68

### QA (ie manual testing)

- [x] Manually testing on mobile (or web simulator) to make sure that swipe downs properly close the drawer

### Automated testing

- [x] Unit test N/A


infredible:
Tested this on my iphone and I'm seeing some issues. The sensitivity seems to be a bit off which causes some jittery behavior. If the app is registering my gesture to pull down the sheet, the sheet should track the movement of my finger.

In the second video, you can also see that sometimes the content in the sheet will get stuck in a scrolled state.
https://github.com/Uniswap/interface/assets/3887562/9b284595-cf0b-4cd4-9ba7-393beae8adf6
https://github.com/Uniswap/interface/assets/3887562/f5ac5fae-b90f-4b0b-b960-87504af22ddd

src/components/AccountDrawer/index.tsx
```diff
@@ -181,21 +183,34 @@ function AccountDrawer() {
     }
   }, [walletDrawerOpen, toggleWalletDrawer])
 
-  // close on escape keypress
```
zzmp, line 184: Was this behavior removed intentionally, or should it be put back for non-mobile users?

zzmp (changes requested):
Looks good! Can you just add comments within the `useGesture` so it's more apparent what the logic there is doing?

src/components/AccountDrawer/index.tsx
LunrEclipse: so when I was making this change there was the exact same code block twice.
Screenshot 2023-06-02 at 12 45 25 PM
so I just removed one of the code blocks which is what that is

zzmp (approved):

cartcrom (approved):

PR #6687, opened 2023-06-02
fix: show '-' instead of .00 when we don't have price data on a token

JFrankfurt:
## Description

Prior to this change, tokens which had not traded recently would have a $0.00 price instead of reporting a lack of data.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-1977/on-token-explore-when-price-is-not-found-we-should-show-a-dash-instead

### Before
| image| image |

### After
| image| image|

### Reproducing the error

1. go to [the tokens page](https://app.uniswap.org/#/tokens/ethereum) and search for a token without recent trade data
2. observe $0 price (incorrect)

### QA (ie manual testing)

- [ ] attempt to reproduce the bug and observe '-' instead of '$0.00'

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test

### DIFF
--- a/src/components/Tokens/TokenTable/TokenRow.tsx
+++ b/src/components/Tokens/TokenTable/TokenRow.tsx
@@ -453,6 +453,11 @@ export const LoadedRow = forwardRef((props: LoadedRowProps, ref: ForwardedRef<HT
     search_token_address_input: filterString,
   }
 
+  const price =
+    token.market?.price?.value !== undefined
+      ? formatUSDPrice(token.market?.price?.value, NumberType.FiatTokenStats)
+      : '-'
+
   // TODO: currency logo sizing mobile (32px) vs. desktop (24px)
   return (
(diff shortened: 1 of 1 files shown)


src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -453,6 +453,11 @@ export const LoadedRow = forwardRef((props: LoadedRowProps, ref: ForwardedRef<HT
     search_token_address_input: filterString,
   }
 
+  const price =
+    token.market?.price?.value !== undefined
+      ? formatUSDPrice(token.market?.price?.value, NumberType.FiatTokenStats)
```
tinaszheng, line 458: doesnt conedison already handle this for you? https://github.com/Uniswap/conedison/blob/main/src/format.ts#L366

JFrankfurt: sure looks like it should--I'll look into why it wasn't doing that

PR #6691, opened 2023-06-02
fix: add '-' instead of '$0.00' when the api does not return a price for a tokens page entry

JFrankfurt:
this replaces #6687

https://linear.app/uniswap/issue/WEB-1977/on-token-explore-when-price-is-not-found-we-should-show-a-dash-instead

image
image


src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -404,7 +404,7 @@ export function LoadingRow(props: { first?: boolean; last?: boolean }) {
     <TokenRow
       header={false}
       listNumber={<SmallLoadingBubble />}
-      loading
+      $loading
       tokenInfo={
```
JFrankfurt, line 408: react was complaining about this prop

tinaszheng (approved):

src/components/Tokens/TokenTable/TokenRow.tsx
```diff
@@ -453,6 +453,9 @@ export const LoadedRow = forwardRef((props: LoadedRowProps, ref: ForwardedRef<HT
     search_token_address_input: filterString,
   }
 
+  // A simple 0 price indicates the price is not currently available from the api
+  const price = token.market?.price?.value === 0 ? '-' : formatUSDPrice(token.market?.price?.value)
```
just-toby, line 457: do you want to show '-' if it's undefined as well?

JFrankfurt: formatUSDPrice will do that already, 0 is the only special case we care about here

PR #6695, opened 2023-06-05
fix: stop resetting modal when we interrupt quote polling

just-toby:
## Description

When the trade state is INVALID, returns the stale trade so consumers of this state don't unmount temporarily if they use `trade !== undefined` as a visibility check.

_Slack thread:_ https://uniswapteam.slack.com/archives/C04D07TQBT4/p1685682395746999

### Before

https://github.com/Uniswap/interface/assets/66155195/13bd2ee2-d6b1-4fd0-9ea3-421ec20d67db

### After

https://github.com/Uniswap/interface/assets/66155195/2e462428-8ba2-407f-8d58-1ef188e0919c

### Reproducing the error

1. load a trade, change the modal state (e.g. by clicking swap and progressing to the next step), then switch tabs
2. note that the modal has reset its internal state

### QA (ie manual testing)

- [x] followed the steps above to ensure that the modal doesn't reset its state

### Automated testing

- [x] Unit test n/a
- [x] Integration/E2E test (added)


tinaszheng (approved):
would be cool to see which step of the modal the user has left off on too so we don't prompt them for 2 signatures 🤔

just-toby:
> would be cool to see which step of the modal the user has left off on too so we don't prompt them for 2 signatures 🤔

@tinaszheng with this fix, we won't prompt twice anymore!

cypress/e2e/permit2.test.ts
```diff
@@ -81,6 +81,31 @@ describe('Permit2', () => {
     })
   })
 
+  it('retains modal state when the window loses focus', () => {
```
zzmp, line 84: Any chance you could rebase this over https://github.com/Uniswap/interface/pull/6656 to avoid the conflict?

zzmp (approved):
lgtm, but I'd prefer if this could use aliasing as in https://github.com/Uniswap/interface/pull/6656

cypress/e2e/permit2.test.ts
```diff
@@ -106,6 +106,28 @@ describe('Permit2', () => {
     cy.get(getTestSelector('popups')).contains('Swapped')
   })
 
+  it('retains modal state when the window loses focus', () => {
+    initiateSwap()
+
+    cy.window().trigger('blur')
+
+    // Verify token approval
+    cy.contains('Enable spending DAI on Uniswap')
```
zzmp, line 115: Your wallet will auto-approve this, so I'm afraid that checking for it may be flaky, because it may disappear before cypress sees it. Can you put this under the "(with intermediate screens)" describe block and stubbing and/or turning off auto-mining to guarantee that this is visible and is not flaky.

zzmp (changes requested):
Would this test have failed before your changes?

zzmp (changes requested):
Would this test have failed before your changes?

tinaszheng (approved):

PR #6700, opened 2023-06-05
fix: type error in cypress infra

just-toby:
## Description

fixes a type error in `cypress/support/commands.ts` causing the FeatureFlags to fail

### DIFF
--- a/cypress/tsconfig.json
+++ b/cypress/tsconfig.json
@@ -7,7 +7,8 @@
     "strict": true,
     "target": "ES5",
     "tsBuildInfoFile": "../node_modules/.cache/tsbuildinfo/cypress", // avoid clobbering the build tsbuildinfo
-    "types": ["cypress", "node"]
+    "types": ["cypress", "node"],
+    "jsx": "react"
   },
   "exclude": ["node_modules"],
   "include": ["**/*.ts"],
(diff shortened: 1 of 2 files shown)


cypress/tsconfig.json
```diff
@@ -7,7 +7,8 @@
     "strict": true,
     "target": "ES5",
     "tsBuildInfoFile": "../node_modules/.cache/tsbuildinfo/cypress", // avoid clobbering the build tsbuildinfo
-    "types": ["cypress", "node"]
+    "types": ["cypress", "node"],
+    "jsx": "react"
```
vm, line 11: is this necessary?

vm (approved):

cypress/tsconfig.json
zzmp: what does this do? why wasn't it happening before?

just-toby: @zzmp @vm no tests were actually using the featureFlags option before (i'm adding one in a follow up diff)

afaict this is needed because we moved the `enum FeatureFlag` into `index.tsx` (from `featureFlags.ts`) and this is just a flag required to support .tsx files

zzmp (approved):

PR #6701, opened 2023-06-05
fix: use shortenAddress for mini portfolio addresses

natew:
## Description

Uses shortenAddress to truncate for mini portfolio addresses.

Ended up organizing some utils, and even though `shortendAddressIfValid` is mostly the same as `shortenAddress` I've left them both as we have testing set up around the error throwing and I want to avoid too much scope creep here.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2071/truncate-wallet-addresses-in-mini-portfolio
_Slack thread:_
_Relevant docs:_

### Before

CleanShot 2023-06-05 at 10 47 38@2x

### After

CleanShot 2023-06-05 at 10 58 04@2x

### Reproducing the error

1. Send money from a wallet
2. Check the Activity pane in the wallet sidebar to see shortened address

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test


src/nft/utils/address.ts
```diff
@@ -1,7 +1,7 @@
 import { isAddress } from '@ethersproject/address'
 
 /**
- * Shortens an Ethereum address by N characters
+ * Shortens an Ethereum address by N characters (search keywords: .slice .truncate)
```
tinaszheng, line 4: can this be moved to a general util file? maybe a new file called `src/utils/addresses.ts` since this isnt unique to just the NFT flow

src/components/AccountDrawer/MiniPortfolio/Activity/ActivityRow.tsx
```diff
@@ -52,7 +53,7 @@ export function ActivityRow({
         descriptor={
           <ActivityRowDescriptor color="textSecondary">
             {descriptor}
-            {ENSName ?? otherAccount}
+            {ENSName ?? shortenAddress(otherAccount || '', 4, 4)}
```
tinaszheng, line 56:
```suggestion
            {ENSName ?? shortenAddress(otherAccount ?? '', 4, 4)}
```

src/nft/utils/address.ts
natew: we actually also already have this one 🙈 https://github.com/Uniswap/interface/blob/f8dd0f7349c21319baf56fa78605fe8de5bc23ad/src/utils/index.ts

should I look into unifying?

natew: oh seeing thats a "0x" prefix only

natew: nope they are the same basically but one errors, i'll check if i can swap it out safely

src/utils/addresses.ts
```diff
@@ -0,0 +1,35 @@
+import { getAddress } from '@ethersproject/address'
```
natew, line 1: this is unchanged from existing besides naming update to `shortenAddressIfValid` to avoid collision.

src/utils/getContract.ts
```diff
@@ -0,0 +1,25 @@
+import { AddressZero } from '@ethersproject/constants'
```
natew, line 1: moved this into its own utils file since it was using `isAddress` and we split that into `addresses.ts` to avoid potential dep cycle.

src/utils/addresses.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+/**
+ * Shortens an Ethereum address by N characters (search keywords: .slice .truncate)
+ * @param address blockchain address
+ * @param charsStart amount of character to shorten (from both ends / in the beginning)
+ * @param charsEnd amount of characters to shorten in the end
+ * @returns formatted string
+ */
+export function shortenAddress(address: string, charsStart = 4, charsEnd?: number): string {
```
tinaszheng, line 21: lol semantically this one feels more like `shortenAddressIfValid` than the other one - since the other one errors, it assumes the address has to be valid 🤔

natew: hm my intuition says the `if` sort of indicates it fails, i was thinking `UnlessInvalid` but realize thats the same thing...

sometimes i've seen `shortenAddressSafe` used for the one that always returns, but that's a bit icky too.

`shortenAddressStrict` perhaps for the one that throws?

natew: went with Strict for now i kind of like it

tinaszheng (approved):

src/utils/getContract.ts
```diff
@@ -0,0 +1,25 @@
...
+import { Contract } from '@ethersproject/contracts'
+import type { JsonRpcProvider, JsonRpcSigner } from '@ethersproject/providers'
+
+import { isAddress } from './addresses'
+
+// account is optional
```
cbachmeier, line 7: nit: all of these `// account is optional` comments aren't helpful, can you replace them with better descriptions?

src/utils/index.test.ts
```diff
@@ -23,21 +23,21 @@ describe('utils', () => {
     })
   })
 
-  describe('#shortenAddress', () => {
+  describe('#shortenAddressStrict', () => {
```
cbachmeier, line 26: nit: add some tests for `shortenAddress` as well

src/utils/index.test.ts
```diff
@@ -1,4 +1,4 @@
-import { isAddress, shortenAddress } from '.'
+import { isAddress, shortenAddressStrict } from './addresses'
```
cbachmeier, line 1: this test file should be renamed `addresses.test.ts` as it only tests functions from that file

src/utils/index.ts
```diff
@@ -1,47 +1,3 @@
...
-// account is optional
-export function getContract(address: string, ABI: any, provider: JsonRpcProvider, account?: string): Contract {
-  if (!isAddress(address) || address === AddressZero) {
-    throw Error(`Invalid 'address' parameter '${address}'.`)
-  }
-
-  return new Contract(address, ABI, getProviderOrSigner(provider, account) as any)
-}
-
 export function escapeRegExp(string: string): string {
```
cbachmeier, line 1: while we're organizing and this is the last fn in `src/utils/index.ts` you should move this to a relevant file or create its own and then make this index file a barrel file as we do elsewhere. This will simplify a lot of imports.

src/utils/getContract.ts
cbachmeier, line 1: nit: I would rename this file `contract.ts` as it has more helper methods than `getContract`

cbachmeier: It would also be very beneficial to add some tests while we're here

src/components/AccountDrawer/MiniPortfolio/Activity/ActivityRow.tsx
```diff
@@ -52,7 +53,7 @@ export function ActivityRow({
         descriptor={
           <ActivityRowDescriptor color="textSecondary">
             {descriptor}
-            {ENSName ?? otherAccount}
+            {ENSName ?? shortenAddress(otherAccount ?? '')}
```
cbachmeier, line 56: nit: instead of passing in an empty string if otherAcount is undefined, you can check beforehand ie:
`{otherAccount && (ENSName ?? shortenAddress(otherAccount))}`

(you'll know ENSName is undefined if otherAccount is also undefined)

src/utils/getContract.ts
natew: tried to think of something but i think its already expressed in the type system so just removing

src/utils/addresses.ts
```diff
@@ -0,0 +1,35 @@
...
+  return `${address.substring(0, charsStart + 2)}...${address.substring(42 - (charsEnd || charsStart))}`
+}
+
+// shorten the checksummed version of the input address to have 0x + 4 characters at start and end
+export function shortenAddressStrict(address: string, chars = 4): string {
+  const parsed = isAddress(address)
+  if (!parsed) {
+    throw Error(`Invalid 'address' parameter '${address}'.`)
+  }
+  return `${parsed.substring(0, chars + 2)}...${parsed.substring(42 - chars)}`
```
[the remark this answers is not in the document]
natew: i wanted to, but the `charsEnd || charsStart` part is different and afraid subtly break things, granted probably this is worth doing but its used in a bunch of areas

src/components/AccountDrawer/MiniPortfolio/Activity/ActivityRow.tsx
natew: this one less sure of, reads nicer i think as is and shouldn't cost any performance really

src/utils/getContract.ts
natew: testing this is pretty intense, feels like a bit of creep but i can add to backlog

natew: (given the scope of this ticket is unrelated, smaller in scope, and those contracts seem to have a lot of parts involved with them)

yarn.lock
```diff
@@ -7460,7 +7460,7 @@ babel-plugin-macros@^3.0.1, babel-plugin-macros@^3.1.0:
 
 babel-plugin-named-asset-import@^0.3.8:
   version "0.3.8"
-  resolved "https://registry.yarnpkg.com/babel-plugin-named-asset-import/-/babel-plugin-named-asset-import-0.3.8.tgz#6b7fa43c59229685368683c28bc9734f24524cc2"
```
JFrankfurt, line 7463: should revert this diff

src/utils/noop.ts
```diff
@@ -1,4 +1,4 @@
 /** No-op function. Returns `null` to satisfy most React typings. */
-export default function noop() {
+export function noop() {
```
JFrankfurt, line 2: I don't think this change is necessary

JFrankfurt (changes requested):
It's a kinda daunting to see a PR titled "fix: use shortenAddress for mini portfolio addresses" and then need to recognize and review convention changes.

I think things in this PR like the newline additions between declarations, converting default exports to named exports, and moving utils, should preferably each be submitted individually, and most of them should be enforced via linting instead of executed ad-hoc. As done here it's impossible for a reviewer to know if all or only some of the suggested convention changes possible in the codebase have been actually been implemented.

natew:
Yea it was requested for me to change to a barrel export on utils which ended up being lots of code change. In the future I’ll just send those as add on PRs.

src/utils/index.ts
```diff
@@ -1,47 +0,0 @@
-import { getAddress } from '@ethersproject/address'
```
cbachmeier, line 1: This isn't a barrel file, look at `src/nft/queries/index.ts` for reference where the index file is a conglomerate of the various exports in the folder. This will allow all imports from this folder to just be from `utils` such as 
`import { isIOS } from 'utils'`

cbachmeier (changes requested):
As @JFrankfurt mentioned it would be more accurate to move ALL file organization to its own PR and this PR can be strictly for use of `shorternAddress` and it's added test cases

src/utils/index.ts
natew: I did, because there were issues caused by it unfortunately. I fixed 5-6 circular dependencies but there’s one more that’s actually non trivial to unroll, so I bailed out and reverted to non barrel.

yarn.lock
natew: any idea why this happens? i notice our yarn.lock has a mix already, i'll send a pr making them all match.

PR #6712, opened 2023-06-06
fix: web app without NFTs

LunrEclipse:
## Description

Removed NFT tabs and displays from web app when nftRoute is disabled (such as for coinbase and Apple apps). 

Added the `shouldDisableNFTRoutesAtom` flag to `SearchBar.tsx`, `SearchBarDropdown.tsx`, and `Landing/index.tsk` to hide NFT related content when the flag is set to true. 

_Linear ticket:_ WEB-1856
_Slack thread:_
_Relevant docs:_

### Before

https://github.com/Uniswap/interface/assets/35351983/4d2869ef-5a7f-4a70-8478-0ac654fb9d8a

### After

https://github.com/Uniswap/interface/assets/35351983/66770344-eaba-42ff-99b2-08982390e25f

### QA (ie manual testing)

- [x] Going through web app to make sure that word NFT doesn't appear anywhere and that NFT tabs are hidden

### Automated testing

- [x] Unit test N/A


src/components/NavBar/SearchBar.tsx
```diff
@@ -102,7 +105,11 @@ export const SearchBar = () => {
     ...trace,
   }
   const placeholderText = useMemo(() => {
-    return isMobileOrTablet ? t`Search` : t`Search tokens and NFT collections`
+    return isMobileOrTablet
+      ? t`Search`
+      : shouldDisableNFTRoutes
+      ? t`Search tokens`
+      : t`Search tokens and NFT collections`
```
zzmp, line 112: These should not be translated - this is fixed separately in https://github.com/Uniswap/interface/pull/6720

src/pages/Landing/index.tsx
```diff
@@ -384,15 +384,27 @@ export default function Landing() {
...
-              <CardGrid cols={2} ref={cardsRef}>
-                {MAIN_CARDS.map(({ darkBackgroundImgSrc, lightBackgroundImgSrc, ...card }) => (
-                  <Card
-                    {...card}
-                    backgroundImgSrc={isDarkMode ? darkBackgroundImgSrc : lightBackgroundImgSrc}
-                    key={card.title}
-                  />
-                ))}
-              </CardGrid>
+              {shouldDisableNFTRoutes ? (
```
zzmp, line 387: To avoid duplication, how would you feel about doing this logic in an earlier useMemo, something like

```
const cards = useMemo(() => MAIN_CARDS.filter((card) => shouldDisableNFTRoutes ? !card.to.startsWith('/nft') : true), [shouldDisableNFTRoutes])
```

And then you can set the `cols={cards.length}`.

zzmp (changes requested):

src/components/NavBar/SearchBar.tsx
LunrEclipse: Screenshot 2023-06-08 at 10 32 15 AM

Got it. Should I just ignore the second task then?

src/pages/Landing/index.tsx
LunrEclipse: That makes sense! Just updated

src/components/NavBar/SearchBar.tsx
zzmp: You should update it, but you should first merge in main, where the translation issue is already fixed.

zzmp (approved):
You should add back in the search bar language modification - I can re-approve quickly after that :)

zzmp:
These would also be excellent candidates for unit tests.

src/pages/Landing/index.tsx
```diff
@@ -323,7 +322,12 @@ export default function Landing() {
     }
   }, [navigate, selectedWallet, queryParams.intro, accountDrawerOpen])
 
-  const shouldDisableNFTRoutes = useAtomValue(shouldDisableNFTRoutesAtom)
+  const shouldDisableNFTRoutes = useDisableNFTRoutes()
+
+  const cards = useMemo(
+    () => MAIN_CARDS.filter((card) => (shouldDisableNFTRoutes ? !card.to.startsWith('/nft') : true)),
+    [shouldDisableNFTRoutes]
+  )
```
JFrankfurt, line 330:
```suggestion
  const cards = useMemo(
    () => MAIN_CARDS.filter((card) => !(shouldDisableNFTRoutes && card.to.startsWith('/nft'))),
    [shouldDisableNFTRoutes]
  )
```

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -311,7 +313,7 @@ export const SearchBarDropdown = ({
                 isLoading={!trendingTokenData}
               />
             )}
-            {!isTokenPage && (
+            {!isTokenPage && !shouldDisableNFTRoutes && (
```
JFrankfurt, line 316: nit: try to avoid compound boolean statements as render conditions

src/components/NavBar/SearchBar.tsx
```diff
@@ -102,8 +104,8 @@ export const SearchBar = () => {
     ...trace,
   }
   const placeholderText = useMemo(() => {
-    return isMobileOrTablet ? `Search` : `Search tokens and NFT collections`
-  }, [isMobileOrTablet])
+    return isMobileOrTablet ? `Search` : shouldDisableNFTRoutes ? `Search tokens` : `Search tokens and NFT collections`
+  }, [isMobileOrTablet, shouldDisableNFTRoutes])
```
JFrankfurt, line 108:
```suggestion
  let result = "Search";
  if (!isMobileOrTablet) {
    result += " tokens";
    if (!shouldDisableNFTRoutes) {
      result += " and NFT collections";
    }
  }
  return result;
}, [isMobileOrTablet, shouldDisableNFTRoutes]);

```

JFrankfurt (approved):

src/hooks/useDisableNFTRoutes.ts
```diff
@@ -0,0 +1,6 @@
+import { useAtomValue } from 'jotai/utils'
+import { shouldDisableNFTRoutesAtom } from 'state/application/atoms'
+
+export function useDisableNFTRoutes() {
+  return useAtomValue(shouldDisableNFTRoutesAtom)
+}
```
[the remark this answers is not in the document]
LunrEclipse: That makes sense, but I changed this to allow for better testing with jest as it was cleaner to just mock `useDisableNFTRoutes()` rather than the entirety of `useAtom`.

src/components/NavBar/SearchBarDropdown.test.tsx
```diff
@@ -0,0 +1,32 @@
+import { useDisableNFTRoutes } from 'hooks/useDisableNFTRoutes'
+import { mocked } from 'test-utils/mocked'
+import { render } from 'test-utils/render'
+
+import { SearchBarDropdown } from './SearchBarDropdown'
+
+jest.mock('hooks/useDisableNFTRoutes')
+
+const SearchBarDropdownProps = {
+  toggleOpen: () => {},
```
zzmp, line 10:
```suggestion
  toggleOpen: () => void 0,
```

src/components/NavBar/SearchBarDropdown.tsx
zzmp: this is fine as-is, because both are booleans, but there is an edge case where you may accidentally render a non-boolean falsey value. To be safe, you could make it a simple boolean:

```suggestion
            {Boolean(!isTokenPage && !shouldDisableNFTRoutes) && (
```

src/components/NavBar/SearchBar.tsx
```diff
@@ -102,8 +104,16 @@ export const SearchBar = () => {
     ...trace,
   }
   const placeholderText = useMemo(() => {
```
zzmp, line 106: This will not generate translations, because translations must be statically known. (This is a bug that existed before.)

In order to generate translations, you can't build a string dynamically, but must declare it statically. You also can't dynamically build the `Trans` component. Instead, you'll want your `useMemo` to return a tagged string, like

    t`Search`

Each case will need to return a full string, something like

    const placeholderText = useMemo(() => {
      if (isMobileOrTablet) {
        return t`Search`
      } else {
        if (shouldDisableNFTRoutes) {
          return t`Search tokens`
        } else {
          return t`Search tokens and NFT collections`
        }
      }
    }, [isMobileOrTablet, shouldDisableNFTRoutes])

Lastly, you'll want to put the translated string into the `<Box>` directly, without using a `<Trans>` wrapper.

You can test if this works by extracting translations locally - using `yarn i18n:extract` - and then looking for the extracted strings in the generated `src/locales/en-US.po` file.

zzmp (changes requested):

src/components/NavBar/SearchBar.tsx
LunrEclipse: Screenshot 2023-07-06 at 11 16 11 AM

Is this what I'm looking for in `src/locales/en-US.po`?

zzmp (approved):
Run `yarn lint --fix` locally, and then this should be good to go!

zzmp (approved):

PR #6713, opened 2023-06-06
chore: removing error state for rejecting chain switch

JackShort:
## Description
* removed error alert from web3 status
* if user rejected switch do not throw error/toast

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2211/updates-to-network-switch-error

## Screen capture

https://github.com/Uniswap/interface/assets/9094792/d30fc674-e334-4b2e-a20c-c2db58778f42


src/components/Web3Status/index.tsx
```diff
@@ -174,15 +153,6 @@ function Web3StatusInner() {
...
     return null
-  } else if (error) {
-    return (
-      <Web3StatusError onClick={handleWalletDropdownClick}>
-        <NetworkIcon />
-        <Text>
-          <Trans>Error</Trans>
-        </Text>
-      </Web3StatusError>
-    )
```
JFrankfurt, line 185: Is there any more generic error (not user rejection) that we do want to preserve this error for? (I'm thinking like the wallet genuinely failed to switch...?)

JFrankfurt (approved):

src/components/Web3Status/index.tsx
JackShort: yeah in the linear ticket it actually said to remove that for this and just show a toast

src/hooks/useSelectChain.ts
```diff
@@ -16,12 +17,17 @@ export default function useSelectChain() {
...
         dispatch(updateConnectionError({ connectionType, error: undefined }))
         await switchChain(connector, targetChain)
       } catch (error) {
+        if (didUserReject(connection, error)) {
+          return
+        }
+
         console.error('Failed to switch networks', error)
 
         dispatch(updateConnectionError({ connectionType, error: error.message }))
```
zzmp, line 33: What happens with this connection error? Is it still used?

src/hooks/useSelectChain.ts
```diff
@@ -16,12 +17,17 @@ export default function useSelectChain() {
...
 
-      const connectionType = getConnection(connector).type
+      const connection = getConnection(connector)
+      const connectionType = connection.type
 
       try {
         dispatch(updateConnectionError({ connectionType, error: undefined }))
         await switchChain(connector, targetChain)
       } catch (error) {
+        if (didUserReject(connection, error)) {
```
zzmp, line 27: If the user did not reject, but the chain change failed, shouldn't we still show the error state?

zzmp (changes requested):
FYI I have a [pending PR](https://github.com/Uniswap/interface/pull/6662) to add these flows to integration tests. It is pending updates to the underlying test framework (cypress-hardhat), but I can update those tests with whatever changes you make here :)

src/hooks/useSelectChain.ts
JackShort: Yeah it still goes through if there is a connection error the toast still should pop up

src/hooks/useSelectChain.ts
JackShort: A toast should pop up which is what happens here

src/hooks/useSelectChain.ts
JackShort: This gets picked up by our sentry logger that's the reason I kept it around

zzmp: If this is only here for Sentry, can you just explicitly call Sentry to report the exception? No need to go the roundabout way and add complexity.

zzmp (changes requested):
Could you report the sentry error directly, and remove the unused connection error state? Or is there some reason it should still be used (like other infra that depends on it)?

JackShort:
> Could you report the sentry error directly, and remove the unused connection error state? Or is there some reason it should still be used (like other infra that depends on it)?

Actually it was being store as part of the state in sentry -  so I removed it since we literally don't use it anywhere now and its logging an error thats being picked up so I don't really see any point in it being part of the sentry state

zzmp (approved):

PR #6714, opened 2023-06-06
refactor: convert getConnection to non-hook

cartcrom:
## Description

Converts `useGetConnection` to a non-hook `getConnection` function


src/components/NavBar/ChainSelector.tsx
```diff
@@ -53,9 +53,7 @@ function getChainsFromEIP155Accounts(accounts?: string[]): SupportedChainId[] {
 function useWalletSupportedChains() {
   const { connector } = useWeb3React()
 
-  const getConnection = useGetConnection()
-
-  const connectionType = getConnection(connector).type
+  const connectionType = getConnection(connector)?.type
```
zzmp, line 56: Is the optional operator necessary?

yarn.lock
```diff
@@ -6157,28 +6157,6 @@
     "@walletconnect/types" "^1.8.0"
```
zzmp, line 6157: This is fine, but should ideally be done in a separate PR.

zzmp (changes requested):

yarn.lock
cartcrom: this wasn't intentional, I think it just happened from running yarn?

src/components/NavBar/ChainSelector.tsx
cartcrom: You're right it's not necessary, I'm not sure why vscode was complaining about it originally

zzmp (approved):

PR #6719, opened 2023-06-07
fix: swap flashing the wrong value after the input is cleared.

LunrEclipse:
## Description

Fixes swap flashing the wrong value after the input is cleared.

Added a useState to the `swap/hooks.tsx` file to check to see if the state of the trade before the current one was `INVALID` and if so, sets the cached trade to `undefined` until a new quote is fetched.

_Linear ticket:_ WEB-1930

### Before

https://github.com/Uniswap/interface/assets/35351983/24015931-5315-4667-9629-588f14cd8949

### After

https://github.com/Uniswap/interface/assets/35351983/68d5f7ca-b3b2-41cf-9c69-9a495e3e40d6

### Reproducing the error

1. Select two random assets on any network
2. Input an amount and wait for the quote to fetch
3. Clear the input
4. Enter another amount
5. When fetching the quote, the output amount should be the previous amount instead of 0

### QA (ie manual testing)

- [x] Test input bar when changed to another value (should have the previous amount in the output bar) and from 0 (should have 0 in the output bar)

### Automated testing

- [x] Unit test N/A


src/state/swap/hooks.tsx
```diff
@@ -119,6 +120,18 @@ export function useDerivedSwapInfo(
     (isExactIn ? outputCurrency : inputCurrency) ?? undefined
   )
 
+  useEffect(() => {
```
natew, line 123: Can avoid a useEffect here and just do the conditional inline, instead of:

```tsx
useEffect(() => {
    if (trade.state === TradeState.INVALID) {
      setPreviouslyInvalid(true)
    } else if (trade.state !== TradeState.LOADING) {
      setPreviouslyInvalid(false)
    }
  }, [trade.state])
```

can do something like:

```tsx
const nextPreviouslyInvalid = (() => {
    if (trade.state === TradeState.INVALID) {
      return true
    } else if (trade.state !== TradeState.LOADING) {
      return false
    }
})()
if (typeof nextPreviouslyInvalid === 'boolean' && nextPreviouslyInvalid !== previouslyInvalid) {
  setPreviouslyInvalid(nextPreviouslyInvalid)
}
```

This will immediately bail from the current render and re-start, rather than rendering twice.

src/state/swap/hooks.tsx
```diff
@@ -119,6 +120,18 @@ export function useDerivedSwapInfo(
...
+    if (trade.state === TradeState.INVALID) {
+      setPreviouslyInvalid(true)
+    } else if (trade.state !== TradeState.LOADING) {
+      setPreviouslyInvalid(false)
+    }
+  }, [trade.state])
+
+  if (trade.state == TradeState.LOADING && previouslyInvalid) {
+    trade.trade = undefined
```
natew, line 132: The mutation here feels wrong to me especially with the useMemo using trade below, I wonder if we shouldn't put this logic into `useBestTrade` behind a new option like `onlyWhenValid`? Otherwise, and either way, I think you'd want to do something like `let trade = useBestTrade` and then here do `trade = { ...trade, trade: undefined }`

natew (commented):
Not that it's changed here but the two useMemo at the bottom of this hook look like they could be combined into one, not 100% sure I'd need to dig in a bit more but useMemo is generally a bit expensive and if the values mostly change together they can memo together. Also just wonder if no memo at all is possible, but maybe the Swap pane is perf sensitive, I'd want to profile and see if it makes a difference.

LunrEclipse:
> 

Yeah as you mentioned it was a design choice to only show the previous estimate once the entire input bar has been fully cleared and processed.

src/state/swap/hooks.tsx
LunrEclipse: Fixed! Out of curiosity is there a reason we want to avoid using useEffects?

src/state/swap/hooks.tsx
LunrEclipse: Fixed! Thanks for pointing out

src/state/swap/hooks.tsx
vm: I'm also unclear about this. Feels more complicated to use a closure here?

natew: Closure isn't necessary it just imo is a way to group the value into one variable, certainly could be done any number of ways.

The reason for avoiding effect is just because it's not necessary, and effects have a cost because they run after the render, and then re-run another render. Whereas if you setState inside the render, you actual bail out of the render immediately and restart it, so you avoid finishing that render + you avoid a whole new render.

Great reading here https://react.dev/learn/you-might-not-need-an-effect

natew (approved):

src/state/swap/hooks.tsx
vm: i think this is clearer
```ts
let nextPreviouslyInvalid

if (trade.state === TradeState.INVALID) {
  nextPreviouslyInvalid = true
} else if (trade.state !== TradeState.LOADING) {
  nextPreviouslyInvalid = false
}
```

vm (approved):

PR #6729, opened 2023-06-08
feat: refactor ConfirmSwapModal for clarity

just-toby:
## Description

Refactors the state management code in the ConfirmSwapModal: 
1. instead of having the same function (`startSwapFlow`) handle every step, we explicitly call the function with the step we want to enact under certain conditions.
2. remove `prepareSwapFlow` from the top level hook result, and instead call it internally within `startSwapFlow`, which just generates the steps and triggers the first step.

### QA (ie manual testing)

- [x] manually verify that the flow behaves exactly the same as before

### Automated testing

- [x] Unit test (existing tests pass)
- [x] Integration/E2E test (existing tests pass)

### DIFF
--- a/src/components/swap/ConfirmSwapModal.tsx
+++ b/src/components/swap/ConfirmSwapModal.tsx
@@ -174,14 +174,6 @@ function useConfirmModalState({
     }
   }, [allowance, performStep, previousSetupApprovalNeeded])
 
-  const previousResetPending = usePrevious(allowance.state === AllowanceState.REQUIRED && allowance.isResetPending)
-  useEffect(() => {
-    const currentResetPending = allowance.state === AllowanceState.REQUIRED && allowance.isResetPending
-    if (previousResetPending && !currentResetPending) {
-      performStep(ConfirmModalState.APPROVING_TOKEN)
-    }
-  }, [allowance, performStep, previousResetPending])
-
   useEffect(() => {
     // Automatically triggers the next phase if the local modal state still thinks we're in the approval phase,
     // but the allowance has been set. This will automaticaly trigger the swap.
--- a/src/components/swap/ConfirmSwapModal.tsx
+++ b/src/components/swap/ConfirmSwapModal.tsx
@@ -101,7 +101,7 @@ function useConfirmModalState({
         case ConfirmModalState.APPROVING_TOKEN:
           setConfirmModalState(ConfirmModalState.APPROVING_TOKEN)
           invariant(allowance.state === AllowanceState.REQUIRED, 'Allowance should be required')
-          await allowance
+          allowance
             .approve()
             .then(() => {
               sendAnalyticsEvent(InterfaceEventName.APPROVE_TOKEN_TXN_SUBMITTED, {
@@ -116,7 +116,7 @@ function useConfirmModalState({
         case ConfirmModalState.PERMITTING:
           setConfirmModalState(ConfirmModalState.PERMITTING)
           invariant(allowance.state === AllowanceState.REQUIRED, 'Allowance should be required')
-          await allowance.permit().catch((e) => catchUserReject(e, PendingModalError.TOKEN_APPROVAL_ERROR))
+          allowance.permit().catch((e) => catchUserReject(e, PendingModalError.TOKEN_APPROVAL_ERROR))
           break
         case ConfirmModalState.PENDING_CONFIRMATION:
           setConfirmModalState(ConfirmModalState.PENDING_CONFIRMATION)
--- a/src/components/swap/ConfirmSwapModal.tsx
+++ b/src/components/swap/ConfirmSwapModal.tsx
@@ -20,6 +20,7 @@ import { useCallback, useEffect, useState } from 'react'
 import { InterfaceTrade } from 'state/routing/types'
 import styled from 'styled-components/macro'
 import { ThemedText } from 'theme'
+import invariant from 'tiny-invariant'
 import { isL2ChainId } from 'utils/chains'
 import { formatSwapPriceUpdatedEventProperties } from 'utils/loggingFormatters'
 import { didUserReject } from 'utils/swapErrorToUserReadableMessage'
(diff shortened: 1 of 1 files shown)


src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -80,48 +81,85 @@ function useConfirmModalState({
...
+                token_symbol: maximumAmountIn?.currency.symbol,
+                token_address: maximumAmountIn?.currency.address,
+                ...trace,
+              })
+            },
+            function onRealError() {
+              setApprovalError(PendingModalError.TOKEN_APPROVAL_ERROR)
+            }
+          )
+          break
```
cartcrom, line 124: I think this would potentially work better w/ the updated implementation of catchUserReject above:
```suggestion
          allowance.approve()
            .then(sendAnalytics)
            .catch((e) => catchUserReject(e, PendingModalError.TOKEN_APPROVAL_ERROR))
          break
```

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -80,48 +81,85 @@ function useConfirmModalState({
...
+      await action()
+    } catch (e) {
+      setConfirmModalState(ConfirmModalState.REVIEWING)
+      if (didUserReject(e)) {
+        return
+      }
+      console.error(e)
+      onRealError()
+    }
+  }
```
cartcrom, line 102: see new suggested usage in other comment

```suggestion
  const catchUserReject = async (e: any, errorType: PendingModalError) => {
    setConfirmModalState(ConfirmModalState.REVIEWING)

    if (didUserReject(e)) return

    console.error(e)
    setApprovalError(errorType)
  }
```

src/components/swap/ConfirmSwapModal.tsx
cartcrom: (and do similar for the steps)

cartcrom (changes requested):

src/components/swap/ConfirmSwapModal.tsx
just-toby: i like it

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -80,48 +81,66 @@ function useConfirmModalState({
...
-  const startSwapFlow = useCallback(async () => {
-    setApprovalError(undefined)
-    if (allowance.state === AllowanceState.REQUIRED) {
-      // Starts the approval process, by triggering either the Token Approval or the Permit signature.
-      try {
-        if (allowance.needsSetupApproval) {
+  const catchUserReject = async (e: any, errorType: PendingModalError) => {
+    setConfirmModalState(ConfirmModalState.REVIEWING)
+    if (didUserReject(e)) return
+    console.error(e)
```
JFrankfurt, line 94: confirming console.error and not .warn is what you want here

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -80,48 +81,66 @@ function useConfirmModalState({
...
-          await allowance.approve()
-          sendAnalyticsEvent(InterfaceEventName.APPROVE_TOKEN_TXN_SUBMITTED, {
-            chain_id: chainId,
-            token_symbol: maximumAmountIn?.currency.symbol,
-            token_address: maximumAmountIn?.currency.address,
-            ...trace,
-          })
-        } else {
+          invariant(allowance.state === AllowanceState.REQUIRED, 'Allowance should be required')
+          await allowance
```
cartcrom, line 104: do we need the awaits here?

cartcrom (approved):

src/components/swap/ConfirmSwapModal.tsx
cartcrom: I disagree with leaving something that isn't necessary

just-toby: you're right, we don't need it

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -80,48 +81,66 @@ function useConfirmModalState({
...
+            .approve()
+            .then(() => {
+              sendAnalyticsEvent(InterfaceEventName.APPROVE_TOKEN_TXN_SUBMITTED, {
+                chain_id: chainId,
+                token_symbol: maximumAmountIn?.currency.symbol,
+                token_address: maximumAmountIn?.currency.address,
+                ...trace,
+              })
+            })
+            .catch((e) => catchUserReject(e, PendingModalError.TOKEN_APPROVAL_ERROR))
```
tinaszheng, line 114: instead of catching in every individual callback here, can you catch the user rejection error at the top level? (around the switch)

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -80,48 +81,66 @@ function useConfirmModalState({
...
   const trace = useTrace()
   const maximumAmountIn = useMaxAmountIn(trade, allowedSlippage)
 
-  const startSwapFlow = useCallback(async () => {
-    setApprovalError(undefined)
-    if (allowance.state === AllowanceState.REQUIRED) {
-      // Starts the approval process, by triggering either the Token Approval or the Permit signature.
-      try {
-        if (allowance.needsSetupApproval) {
+  const catchUserReject = async (e: any, errorType: PendingModalError) => {
```
tinaszheng, line 91: does there have to be a separate `PendingModalError` type based on step - can you just use the same `ConfirmModalState` variable and deduce based on that

src/components/swap/ConfirmSwapModal.tsx
just-toby: i think this is better because we throw a different error for each case

src/components/swap/ConfirmSwapModal.tsx
just-toby: yes, but on the other hand these two components don't need to be so tightly coupled. maybe we want to use the error UI in a different modal in the future ?

tinaszheng (approved):

PR #6732, opened 2023-06-08
feat: add tracking params and go straight to app store for iOS for the landing page wallet CTA

natew:
## Description

Adds tracking + go to app store for iOS.

_Linear ticket:_
_Slack thread:_
_Relevant docs:_

### Before

### After

### Reproducing the error

1. 

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test


src/pages/Landing/index.tsx
```diff
@@ -384,7 +386,14 @@ export default function Landing() {
                 <LearnMoreArrow />
               </LearnMoreContainer>
 
-              <DownloadWalletLink href="https://wallet.uniswap.org/">
+              <DownloadWalletLink
+                href={getDownloadAppLink()}
+                onClick={(e) => {
+                  // prevent and run command to send analytics events, but keep href for browsers to show right link
+                  e.preventDefault()
+                  openDownloadApp(undefined, downloadLinkOptions)
```
natew, line 394: didn't want to add analytics event but will do that now...

natew: pending here, we can follow on with it https://github.com/Uniswap/analytics-events/pull/47

src/utils/openDownloadApp.ts
```diff
@@ -0,0 +1,42 @@
...
+const defaultDownloadAppOptions = {
+  appStoreParams: `?pt=123625782&ct=In-App-Banners&mt=8`,
+  analyticsEventName: SharedEventName.ELEMENT_CLICKED,
+}
+
+export function openDownloadApp(
+  element?: InterfaceElementName,
+  options: OpenDownloadAppOptions = defaultDownloadAppOptions
+) {
+  sendAnalyticsEvent(options.analyticsEventName ?? SharedEventName.ELEMENT_CLICKED, { element })
```
adjkant, line 23: Will this just log an element clicked event with no properties if there's no element? Should we only log when there's an element?

src/pages/Landing/index.tsx
adjkant: What is this event capturing vs the other existing one / the options?

src/utils/openDownloadApp.ts
natew: good question, i defer to @tinaszheng

natew:
@adjkant re:

> What is this event capturing vs the other existing one / the options?

In this case it's just capturing the download vs the wallet landing page, but we need to land analytics tags to pass there instead of `undefined`, but this link doesn't have analytics otherwise so I think it does add a unique event.

src/utils/openDownloadApp.ts
adjkant: I think it will in this case - once the analytics PR merges we should make sure this doesn't allow an undefined analytics event here / doesn't log any event in that case.

src/utils/openDownloadApp.ts
```diff
@@ -0,0 +1,42 @@
...
+import { isIOS } from 'utils/userAgent'
+
+const APP_STORE_LINK = 'https://apps.apple.com/app/apple-store/id6443944476'
+const MICROSITE_LINK = 'https://wallet.uniswap.org/'
+
+type OpenDownloadAppOptions = {
+  appStoreParams?: string
+  microSiteParams?: string
+  analyticsEventName?: SharedEventName
+}
```
adjkant, line 12: IMO this is a case where abstracting the functionality confuses more than it helps. Opening the microsite is very different than opening the app store - can we do two minimal param functions for each of these that log their respective event automatically, no param needed beyond the interface element?

adjkant (changes requested):

src/utils/openDownloadApp.ts
natew: @adjkant i think these are coupled somewhat how they are used though, as we basically use the same link to open one or the other in a few different areas, but in one area we need to customize the params. if i split them out then i'd have to add logic to the call-site in multiple places that is duplicated. open to some third alternative or something though

natew: @adjkant did a pass over this and i think its a bit cleaner.

natew: added the new event as well

src/components/Banner/UniswapWalletBanner.tsx
```diff
@@ -127,14 +127,14 @@ export default function UniswapWalletBanner() {
...
-            <BannerButton backgroundColor="black" onClick={openWalletMicrosite}>
+            <BannerButton backgroundColor="black" onClick={() => openWalletMicrosite()}>
               <ThemedText.LabelSmall color="white">
                 <Trans>Learn more</Trans>
               </ThemedText.LabelSmall>
             </BannerButton>
           </>
         ) : (
-          <BannerButton backgroundColor="white" width="125px" onClick={openWalletMicrosite}>
+          <BannerButton backgroundColor="white" width="125px" onClick={() => openWalletMicrosite()}>
```
adjkant, line 137: any reason not to pass directly as before?

src/utils/openDownloadApp.ts
```diff
@@ -0,0 +1,69 @@
...
+    onClick(e) {
+      e.preventDefault()
+      openDownloadApp(options)
+    },
+  }
+}
+
+type AnalyticsLinkOptions = {
+  element?: InterfaceElementName
+  params?: string
```
adjkant, line 56: naming nit: `urlParamString`?

src/pages/Landing/index.tsx
```diff
@@ -384,7 +385,12 @@ export default function Landing() {
                 <LearnMoreArrow />
               </LearnMoreContainer>
 
-              <DownloadWalletLink href="https://wallet.uniswap.org/">
+              <DownloadWalletLink
+                {...getDownloadAppLinkProps({
+                  microSiteParams: `utm_source=home_page&utm_medium=webapp&utm_campaign=wallet_microsite&utm_id=1`,
+                  appStoreParams: `ct=Uniswap-Home-Page&mt=8`,
```
adjkant, line 391: Make these strings constants still so it's clear what they represent / the case?

adjkant (approved):
Looks good on the analytics logging part! Some small nits but otherwise good to go

src/utils/openDownloadApp.ts
adjkant: Looks good on this part!

adjkant:
need to fix callsites as well, see typescript issues

src/pages/Landing/index.tsx
adjkant: reopening this one, can we move this object into `src/utils/openDownloadApp.ts` as `landingPageParams`?

adjkant (approved):
one nit, otherwise looks good!

src/utils/openDownloadApp.ts
```diff
@@ -0,0 +1,69 @@
...
+ * but with JS on it will send the analytics event before navigating to the href.
+ *
+ * I've added a helper `getDownloadAppLinkProps` that unifies this behavior into one thing.
+ */
+
+export function openDownloadApp(options: OpenDownloadAppOptions = defaultDownloadAppOptions) {
+  if (isIOS) {
+    openAppStore({ element: options?.element, urlParamString: options.appStoreParams })
+  } else {
+    openWalletMicrosite({ element: options?.element, urlParamString: options.appStoreParams })
```
zzmp, line 34:
```suggestion
    openWalletMicrosite({ element: options?.element, urlParamString: options?.microSiteParams })
```

src/utils/openDownloadApp.ts
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * This way with JS disabled and when hovering the <a /> you see and nav to the full href properly,
+ * but with JS on it will send the analytics event before navigating to the href.
+ *
+ * I've added a helper `getDownloadAppLinkProps` that unifies this behavior into one thing.
+ */
+
+export function openDownloadApp(options: OpenDownloadAppOptions = defaultDownloadAppOptions) {
+  if (isIOS) {
+    openAppStore({ element: options?.element, urlParamString: options.appStoreParams })
```
zzmp, line 32:
```suggestion
    openAppStore({ element: options?.element, urlParamString: options?.appStoreParams })
```

src/utils/openDownloadApp.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+type AnalyticsLinkOptions = {
+  element?: InterfaceElementName
+  urlParamString?: string
+}
+
+const openAppStore = (options?: AnalyticsLinkOptions) => {
+  sendAnalyticsEvent(InterfaceEventName.UNISWAP_WALLET_APP_DOWNLOAD_OPENED, { element: options?.element })
+  window.open(linkWithParams(APP_STORE_LINK, options?.urlParamString), /* target = */ 'uniswap_wallet_appstore')
```
zzmp, line 61: What does `target` do here (and below)?

src/pages/Landing/__snapshots__/index.test.tsx.snap
```diff
@@ -4512,7 +4512,7 @@ exports[`disable nft on landing page renders nft information and card 1`] = `
       </div>
       <a
         class="9 c58"
-        href="https://wallet.uniswap.org/"
+        href="https://wallet.uniswap.org/?utm_source=home_page&utm_medium=webapp&utm_campaign=wallet_microsite&utm_id=1"
```
zzmp, line 4515: Can you make sure both the iOS and microsite are tested in a snapshot, by mocking the isIOS hook?

src/components/AccountDrawer/DownloadButton.tsx
```diff
@@ -62,7 +44,9 @@ export function DownloadButton({
   const onButtonClick = useCallback(() => {
     // handles any actions required by the parent, i.e. cancelling wallet connection attempt or dismissing an ad
     onClick?.()
-    openDownloadApp(element)
+    openDownloadApp({
+      element,
```
zzmp, line 48: nit: if you remove the comma the linter will allow it as a one-liner

zzmp (changes requested):
I think one of the params is incorrectly passed

src/utils/openDownloadApp.ts
zzmp: Shouldn't this be `options?.microSiteParams`?

zzmp (changes requested):

zzmp (approved):

PR #6733, opened 2023-06-08
fix: insufficient liquidity state flashes

JackShort:
## Description
* the insufficient liquidity state is flashing when you first input a number to the swap widget
* this is due to the fact that we have a debounce of a rather significant amount of time on it so the the return from the routing api hook is `undefined` so it seems like we have an invalid trade state
* the solve was to add a boolean which checks if we are currently in a debounced state by checking if any of the debounced items are undefined when they are defined when passed into the hook and to update the loading state appropriately

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-1929/insufficient-liquidity-state-flashes-when-getting-a-quote
_Slack thread:_
_Relevant docs:_

### Before

https://github.com/Uniswap/interface/assets/9094792/89be5d62-9029-4aad-aaf9-45363eab5182

### After

https://github.com/Uniswap/interface/assets/9094792/148e0c82-1050-41ac-96c0-c0a341b96971


src/hooks/useBestTrade.ts
```diff
@@ -40,6 +40,7 @@ export function useBestTrade(
     useMemo(() => [amountSpecified, otherCurrency], [amountSpecified, otherCurrency]),
     useDebounceSwapQuoteFlag() === DebounceSwapQuoteVariant.Enabled ? DEBOUNCE_TIME_INCREASED : DEBOUNCE_TIME
   )
+  const inDebounce = (!debouncedAmount || !debouncedOtherCurrency) && Boolean(amountSpecified) && Boolean(otherCurrency)
```
cbachmeier, line 43: nit: I think `const inDebounce =
    (!debouncedAmount && Boolean(amountSpecified)) || (!debouncedOtherCurrency && Boolean(otherCurrency))` is slightly more readable to correlate that there has been an input but is currently being debounced

cbachmeier (approved):

src/hooks/useBestTrade.ts
```diff
@@ -40,6 +40,7 @@ export function useBestTrade(
     useMemo(() => [amountSpecified, otherCurrency], [amountSpecified, otherCurrency]),
     useDebounceSwapQuoteFlag() === DebounceSwapQuoteVariant.Enabled ? DEBOUNCE_TIME_INCREASED : DEBOUNCE_TIME
```
cbachmeier, line 41: this is unrelated to your change but can we also move `useDebounceSwapQuoteFlag() === DebounceSwapQuoteVariant.Enabled` to its own `const` to avoid calling a hook inside of a hook?

JackShort: will do thats weird

src/hooks/useBestTrade.ts
zzmp: another nit: I'd prefer `inDebounce` be placed directly above `isLoading`, as that's where it is used.

zzmp (approved):

PR #6735, opened 2023-06-09
fix: using search hotkey enter navigates to the wrong result

natew:
## Description

fix SearchBarDropdown selecting invalid result on enter after initial search when recent searches are filled

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-1290/using-search-hotkey-enter-navigates-to-the-wrong-result
_Slack thread:_
_Relevant docs:_

### Reproducing the error

1. make sure you have a recent search in nav bar search
2. press “ / “ to open search
3. search something for other than your top most recent search
4. a match for your query will be highlighted, press enter
5. the app will navigate to your most recent search rather than the previously highlighted item

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test


src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -130,24 +134,68 @@ interface SearchBarDropdownProps {
...
+
+export const SearchBarDropdown = (props: SearchBarDropdownProps) => {
+  const { isLoading } = props
+  const { chainId } = useWeb3React()
+  const showBNBComingSoonBadge = chainId === SupportedChainId.BNB && !isLoading
+
+  return (
+    <Column overflow="hidden" className={clsx(styles.searchBarDropdownNft, styles.searchBarScrollable)}>
+      <Box opacity={isLoading ? '0.3' : '1'} transition="125">
+        <SuspenseFallbackToPreviousContents>
```
natew, line 159: 💅

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -130,24 +134,68 @@ interface SearchBarDropdownProps {
   isLoading: boolean
 }
 
-export const SearchBarDropdown = ({
-  toggleOpen,
-  tokens,
-  collections,
+const useSearchTrendingTokens = () => {
```
natew, line 137: originally i refactored this differently so the `SearchBarSuggestion` and this hook being split out are just remnants of that.. but not terrible

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -130,24 +134,68 @@ interface SearchBarDropdownProps {
...
+              <Trans>Coming soon: search and explore tokens on BNB Chain</Trans>
+            </ThemedText.BodySmall>
+          </BNBComingSoonBadge>
+        )}
+      </Box>
+    </Column>
+  )
+}
+
+const SearchBarDropdownContents = ({
```
natew, line 177: below here its just the same logic pulled into a component rather than a useEffect, with some minor changes

natew (commented):
Assuming we want to add cypress test here? I'll add that..

cypress/e2e/universal-search.test.ts
```diff
@@ -1,13 +1,17 @@
 import { getTestSelector } from '../utils'
 
 describe('Universal search bar', () => {
-  before(() => {
-    cy.visit('/')
+  function openSearch() {
     cy.get('[data-cy="magnifying-icon"]')
```
zzmp, line 5: This should be updated. If we expect search to be accessible via `/`, I'd prefer it be accessed it that way. [Data attributes should generally be considered brittle and only used when necessary](https://github.com/Uniswap/interface/blob/main/cypress/README.md#prefer-selecting-elements-using-on-screen-text-over-data-testid-attributes).

cypress/e2e/universal-search.test.ts
```diff
@@ -1,13 +1,17 @@
...
       .parent()
       .then(($navIcon) => {
         $navIcon.click()
       })
```
zzmp, line 9: In general, you can just invoke `click` on the yielded subject: `.parent().click()`.

cypress/e2e/universal-search.test.ts
```diff
@@ -35,6 +39,36 @@ describe('Universal search bar', () => {
     cy.contains('UNI is the governance token for Uniswap').should('exist')
   })
 
+  it('should go to the right result when recent results are shown and you search something new', () => {
```
zzmp, line 42:
```suggestion
  it('should go to the selected result when recent results are shown', () => {
```

cypress/e2e/universal-search.test.ts
```diff
@@ -35,6 +39,36 @@ describe('Universal search bar', () => {
...
+    // Search for uni token by name.
+    cy.get('[data-cy="search-bar-input"]').last().clear().type('uni')
```
zzmp, line 44:
```suggestion
    cy.get('[data-cy="search-bar-input"]').last().type('uni')
```

cypress/e2e/universal-search.test.ts
```diff
@@ -35,6 +39,36 @@ describe('Universal search bar', () => {
...
+    cy.get('[data-cy="searchbar-token-row-UNI"]')
+      .should('contain.text', 'Uniswap')
+      .and('contain.text', 'UNI')
+      .and('contain.text', '$')
+      .and('contain.text', '%')
```
zzmp, line 49: Verifying the row exists should be sufficient (here and with ETH).

```suggestion
    cy.get('[data-cy="searchbar-token-row-UNI"]')
```

zzmp (commented):
just some cypress nits

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -212,162 +258,114 @@ export const SearchBarDropdown = ({
...
-      </Box>
-    </Column>
+    <>
+      {hasInput ? (
+        // Empty or Up to 8 combined tokens and nfts
+        <Column gap="20">
+          {showCollectionsFirst ? (
+            <>
+              {collectionSearchResults}
+              {tokenSearchResults}
```
just-toby, line 307: these are conditionally rendered, so should we just inline the jsx? there may be a perf hit from putting them in variables before we know they're needed

src/components/Suspense/SuspenseFallbackToPreviousContents.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { Suspense, useEffect, useRef } from 'react'
+
+export const SuspenseFallbackToPreviousContents = (props: { children: React.ReactNode }) => {
```
just-toby, line 3: can you add a comment explaining when / how this should be used

just-toby (approved):

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -14,13 +14,15 @@ import { Box } from 'nft/components/Box'
 import { Column, Row } from 'nft/components/Flex'
 import { subheadSmall } from 'nft/css/common.css'
 import { GenieCollection, TrendingCollection } from 'nft/types'
-import { ReactNode, useEffect, useMemo, useState } from 'react'
+import React, { useEffect, useMemo, useState } from 'react'
```
zzmp, line 17: In general, try to only import what you're using. In this case, you shouldn't be importing `React`, as you'll only ever use its properties.

This doesn't matter so much for React because it's only distributed as CommonJS (cjs), but for ESModules (esm) importing the global will unnecessarily increase our bundle size, so it's a good habit to follow.

cypress/e2e/universal-search.test.ts
```diff
@@ -35,6 +35,32 @@ describe('Universal search bar', () => {
...
+    cy.get('[data-cy="search-bar-input"]').last().type('eth')
+
+    // Validate ETH result now exists.
+    cy.get('[data-cy="searchbar-token-row-ETH"]')
+
+    // Hit enter
+    cy.get('[data-cy="search-bar-input"]').last().type('{enter}')
+
+    // Validate we went to ethereum address
+    cy.url().should('contain', 'tokens/ethereum/NATIVE')
```
zzmp, line 61: The search bar element should be static (I think), so you can grab it once and re-use it from a closure:

```
    cy.get('[data-cy="search-bar-input"]').last().then((searchBar) => {
      // Search for uni token by name.
      cy.wrap(searchBar).type('uni')
      ...
    })
```

Also, avoid new instances of `data-cy`, using `data-testid` (and `getTestSelector`) instead. No changes necessary for this (unless you want to), just something to keep in mind.

src/components/NavBar/SearchBarDropdown.tsx
natew: i think actually the perf hit is only in an extra createElement() call with an object, at least i think react only does the sub-tree stuff based on what's actually returned not at the call-site

cypress/e2e/universal-search.test.ts
natew: good points, addressed the first but have a lot of unblocking work to get to so i'll remember to avoid data-cy in future pr's!

src/hooks/useSuspendWhen.tsx
```diff
@@ -0,0 +1,15 @@
+import { useState } from 'react'
+
+export const useSuspendWhen = (shouldSuspend = false) => {
```
zzmp, line 3: I think this belongs in `Suspense/Suspend.tsx`, as it is only used there.

src/components/Suspense/SuspenseFallbackToPreviousContents.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { Suspense, useEffect, useRef } from 'react'
+
+/**
+ * This is useful for keeping the "last rendered" components on-screen while any suspense
+ * is triggered below this component.
+ *
+ * It stores a reference to the current children, and then returns them as the fallback.
+ */
+export const SuspenseFallbackToPreviousContents = (props: { children: React.ReactNode }) => {
```
zzmp, line 9: Would you consider `SuspenseWithLastRenderAsFallback`?

src/components/Suspense/SuspenseFallbackToPreviousContents.tsx
```diff
@@ -0,0 +1,17 @@
...
+  const lastContents = useRef(props.children)
+
+  useEffect(() => {
+    lastContents.current = props.children
+  }, [props.children])
```
zzmp, line 14: I think this duplicates `usePrevious`. Can you use that instead?

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -130,24 +134,68 @@ interface SearchBarDropdownProps {
...
+  const { chainId } = useWeb3React()
+  const showBNBComingSoonBadge = chainId === SupportedChainId.BNB && !isLoading
+
+  return (
+    <Column overflow="hidden" className={clsx(styles.searchBarDropdownNft, styles.searchBarScrollable)}>
+      <Box opacity={isLoading ? '0.3' : '1'} transition="125">
+        <SuspenseFallbackToPreviousContents>
+          <Suspend when={isLoading}>
+            <SearchBarDropdownContents {...props} />
+          </Suspend>
```
zzmp, line 162: Why can't these be unified into a single element? This seems like it is avoiding using the imperative pattern of `useTransition` by setting up a declarative pattern instead. Could you have a composed element that does this all for us (eg `<SuspendFallbackToPreviousContents when={isLoading}>`)? The composed element should also include docs that it allows you to use suspense declaratively instead of imperatively, as through `useTransition`.

zzmp (changes requested):

cypress/e2e/universal-search.test.ts
```diff
@@ -16,14 +25,40 @@ describe('Universal search bar', () => {
...
+    cy.get('[data-cy="searchbar-token-row-ETH"]')
+
+    // Hit enter
+    getSearchBar().type('{enter}')
+
+    // Validate we went to ethereum address
+    cy.url().should('contain', 'tokens/ethereum/NATIVE')
+  })
+
   it.skip('should show recent tokens and popular tokens with empty search term', () => {
```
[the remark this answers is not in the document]
natew: no idea 🤷

src/hooks/useSuspendWhen.tsx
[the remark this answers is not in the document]
natew: then when matches the `<Suspend when={}>`, which I've always like when for some reason, maybe just because if is a reserved word and sometimes it causes issues when you want to use a variable and then pass it to an object without having to rename it, but i can change both to if in this case

src/components/Suspense/SuspenseFallbackToPreviousContents.tsx
natew: i sort of like Previous overlap with usePrevious now that we made that change but don't mind much either way

natew: split the diff and used your structure with previous instead of last

src/components/NavBar/SearchBarDropdown.tsx
natew: you could use `SuspendFallbackToPreviousContents` with any component that throws a suspense, and then Suspend is just one component that does so. in fact maybe we should rename Suspened to be more explicit: `SuspendConditionally`, because they are sort of two separate utilities and can both be used independently usefully.

natew: tried that out

natew: @zzmp made some changes here, lmk if this looks good

src/components/Suspense/SuspendConditionally.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import { useState } from 'react'
+
+export const SuspendConditionally = (props: { if: boolean; children: React.ReactNode }) => {
+  useSuspendIf(props.if)
+  return <>{props.children}</>
```
zzmp, line 6: Can't this just be returned as-is?

```suggestion
  return props.children
```

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -222,158 +246,117 @@ export const SearchBarDropdown = ({
...
-            <ChainLogo src={logoUri} />
-            <ThemedText.BodySmall color="textSecondary" fontSize="14px" fontWeight="400" lineHeight="20px">
-              <ComingSoonText chainId={chainId} />
-            </ThemedText.BodySmall>
-          </ChainComingSoonBadge>
-        )}
-      </Box>
-    </Column>
+    <>
+      {hasInput ? (
```
zzmp, line 291: I don't think you need this fragment: `return hasInput ? (`...

zzmp (changes requested):
I think you unnecessarily use fragments (`<></>`), otherwise this looks great! Very usable new components - thank you!

src/components/Suspense/SuspendConditionally.tsx
zzmp: type system complained.

zzmp (approved):

PR #6737, opened 2023-06-09
fix: <a> cannot appear as a descendant of <a>

sarobolket:
## Description
Fix `<a> cannot appear as a descendant of <a>` in the landing page with the prop `switchLocaleOff`.


JFrankfurt (changes requested):
I think it might be a bit cleaner to make the swap page component only render the `a` tag here if it's on the `/swap` page

sarobolket:
@JFrankfurt is it clearer now?

JFrankfurt (approved):

PR #6738, opened 2023-06-09
fix: Each child in a list should have a unique "key" prop

sarobolket:
## Description
Fix the warning `Each child in a list should have a unique "key" prop`.
In the Table component there is an error because `headerGroup` don't have an id.


src/nft/components/explore/Table.tsx
```diff
@@ -148,8 +148,8 @@ export function Table<D extends Record<string, unknown>>({
     <GlowEffect>
       <table {...getTableProps()} className={styles.table}>
         <thead className={styles.thead}>
-          {headerGroups.map((headerGroup) => (
-            <tr {...headerGroup.getHeaderGroupProps()} key={headerGroup.id}>
+          {headerGroups.map((headerGroup, index) => (
+            <tr {...headerGroup.getHeaderGroupProps()} key={index}>
```
tinaszheng, line 152: for arrays like `headerGroups` that can dynamically change in size - don't use `index` as a key because it's not guaranteed that the same instance will have the same index

can you investigate why `headerGroup.id` is not working instead?

tinaszheng (changes requested):

sarobolket:
@tinaszheng Are there any problems with this PR?

src/nft/components/explore/Table.tsx
```diff
@@ -149,7 +149,8 @@ export function Table<D extends Record<string, unknown>>({
       <table {...getTableProps()} className={styles.table}>
         <thead className={styles.thead}>
           {headerGroups.map((headerGroup) => (
-            <tr {...headerGroup.getHeaderGroupProps()} key={headerGroup.id}>
+            // eslint-disable-next-line react/jsx-key
+            <tr {...headerGroup.getHeaderGroupProps()}>
```
cbachmeier, line 153: the table row should be given a valid key instead of removing and disabling the error

src/nft/components/explore/Table.tsx
```diff
@@ -246,7 +247,8 @@ function LoadingTable({ headerGroups, visibleColumns, ...props }: LoadingTablePr
       <table {...props} className={styles.table}>
         <thead className={styles.thead}>
           {headerGroups.map((headerGroup) => (
-            <tr {...headerGroup.getHeaderGroupProps()} key={headerGroup.id}>
+            // eslint-disable-next-line react/jsx-key
+            <tr {...headerGroup.getHeaderGroupProps()}>
```
cbachmeier, line 251: the table row should be given a valid key instead of removing and disabling the error

src/nft/components/explore/CarouselCard.tsx
```diff
@@ -291,7 +291,7 @@ const DEFAULT_TABLE_ELEMENTS = 12
 const LoadingTable = () => {
   return (
     <>
-      {[...Array(DEFAULT_TABLE_ELEMENTS)].map((index) => (
+      {Array.from({ length: DEFAULT_TABLE_ELEMENTS }, (_, i) => i + 1).map((_, index) => (
```
cbachmeier, line 294: Was CarouselCard also getting a unique key warning? why was this added?

src/nft/components/explore/Table.tsx
```diff
@@ -280,7 +282,7 @@ function LoadingTable({ headerGroups, visibleColumns, ...props }: LoadingTablePr
           ))}
         </thead>
         <tbody {...props}>
-          {[...Array(DEFAULT_TRENDING_TABLE_QUERY_AMOUNT)].map((_, index) => (
+          {Array.from({ length: DEFAULT_TRENDING_TABLE_QUERY_AMOUNT }, (_, i) => i + 1).map((_, index) => (
```
cbachmeier, line 285: what advantage does this syntax have over the current?

src/nft/components/explore/Table.tsx
sarobolket: The getHeaderGroupProps function return a key prop

src/nft/components/explore/CarouselCard.tsx
sarobolket: Yes, CarouselCard also getting a unique key warning.
The `[...Array(DEFAULT_TABLE_ELEMENTS)]` code return an array of undefined, instead `Array.from({ length: DEFAULT_TABLE_ELEMENTS }, (_, i) => i + 1)` return an array of number element

src/nft/components/explore/Table.tsx
sarobolket: you are right, there are no advantages, We can use `[...Array(DEFAULT_TRENDING_TABLE_QUERY_AMOUNT)].map((_, index)`

mattbspector:

PR #6739, opened 2023-06-09
fix: prevent scrolling with settings modal open

LunrEclipse:
## Description

Fixes bug where screen can continue to be scrolled even when the settings/share modal are open.

Added a `useEffect` to `ShareButton.tsx` and `Settings/index.tsx` so that the document's body is fixed and unscrollable when either modal is open on mobile.

Follow this [tutorial](https://css-tricks.com/prevent-page-scrolling-when-a-modal-is-open/).

_Linear ticket:_ WEB-1473

### Before

https://github.com/Uniswap/interface/assets/35351983/ff07ab1b-fd28-4765-86be-502fbc587cb9

https://github.com/Uniswap/interface/assets/35351983/512635ea-ec7f-4701-a571-6764e1e0bc9c

### After

https://github.com/Uniswap/interface/assets/35351983/a3d088ae-79b3-45de-b5a2-cb4722e3b0ca

https://github.com/Uniswap/interface/assets/35351983/523915b7-760b-4750-979d-575f193a0a86

### Reproducing the error

1. Go to swap, token, or nft screen on a mobile device.
2. Click the settings or share button
3. Try swiping on the background

### QA (ie manual testing)

- [x] Go to swap screen, click settings modal, make sure the screen isn't scrollable
- [x] Go to token/nft details screen, click the share button, make sure the screen isn't scrollable
- [x] Make sure normal functions on web

### Automated testing

- [x] Unit test N/A

### DIFF
--- a/src/components/Settings/index.tsx
+++ b/src/components/Settings/index.tsx
@@ -5,11 +5,12 @@ import { AutoColumn } from 'components/Column'
 import { L2_CHAIN_IDS } from 'constants/chains'
 import { useOnClickOutside } from 'hooks/useOnClickOutside'
 import { isSupportedChainId } from 'lib/hooks/routing/clientSideSmartOrderRouter'
-import { useRef } from 'react'
+import { useEffect, useRef } from 'react'
 import { useModalIsOpen, useToggleSettingsMenu } from 'state/application/hooks'
 import { ApplicationModal } from 'state/application/reducer'
 import styled from 'styled-components/macro'
 import { Divider } from 'theme'
+import { isMobile } from 'utils/userAgent'
 
 import MaxSlippageSettings from './MaxSlippageSettings'
 import MenuButton from './MenuButton'
@@ -50,6 +51,19 @@ export default function SettingsTab({ autoSlippage, chainId }: { autoSlippage: P
 
   const toggleMenu = useToggleSettingsMenu()
   useOnClickOutside(node, isOpen ? toggleMenu : undefined)
+
+  useEffect(() => {
+    if (isOpen && isMobile) {
+      document.body.style.position = 'fixed'
+      document.body.style.top = `-${window.scrollY}px`
+      document.body.style.width = '100%'
+    } else if (isMobile) {
+      const scrollY = document.body.style.top
+      document.body.style.position = ''
+      document.body.style.top = ''
+      window.scrollTo(0, parseInt(scrollY || '0') * -1)
+    }
+  }, [isOpen])
 
   const isSupportedChain = isSupportedChainId(chainId)
 
--- a/src/components/Tokens/TokenDetails/ShareButton.tsx
+++ b/src/components/Tokens/TokenDetails/ShareButton.tsx
@@ -3,7 +3,7 @@ import { Currency } from '@uniswap/sdk-core'
 import { NATIVE_CHAIN_ID } from 'constants/tokens'
 import { chainIdToBackendName } from 'graphql/data/util'
 import { useOnClickOutside } from 'hooks/useOnClickOutside'
-import { useRef } from 'react'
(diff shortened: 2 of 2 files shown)


src/components/Settings/index.tsx
```diff
@@ -51,6 +52,19 @@ export default function SettingsTab({ autoSlippage, chainId }: { autoSlippage: P
   const toggleMenu = useToggleSettingsMenu()
   useOnClickOutside(node, isOpen ? toggleMenu : undefined)
 
+  useEffect(() => {
```
zzmp, line 55: This is the same effect across both files - can you consolidate into a single hook, `useDisableScrolling(disabled: boolean)`, in `src/hooks/useDisableScrolling.ts`?

zzmp (changes requested):
It doesn't look like you're backing out the scrolling disabling after the useEffect runs - should that be done in a returned callback function?

Would it be safer to disable by capturing the scroll or drag event, instead of by modifying the page? Modifying the page seems a lot scarier.

LunrEclipse:
> It doesn't look like you're backing out the scrolling disabling after the useEffect runs - should that be done in a returned callback function?
> 
> Would it be safer to disable by capturing the scroll or drag event, instead of by modifying the page? Modifying the page seems a lot scarier.

I found a much easier approach where we just set the `overflow` tag of the page to either `hidden` or `auto` depending on if the modal is open or not. In the previous approach I would be concerned with page modification, however, with this new approach I don't think I'm as concerned since the whole purpose of the `overflow` tag is to dictate scroll behavior. Let me know what you think though in terms of page modifications.

src/hooks/useDisableScrolling.ts
```diff
@@ -0,0 +1,12 @@
+import { useEffect } from 'react'
+import { isMobile } from 'utils/userAgent'
+
+export default function useDisableScrolling(open: boolean | undefined | null) {
```
just-toby, line 4: maybe rename this input to "disable" since `open` isn't related to anything internal to this function

just-toby: also this file seems like a great candidate for a unit test - see the `renderHook` function in test-utils and examples of its usage

LunrEclipse: created tests!

just-toby (approved):

src/hooks/useDisableScrolling.ts
```diff
@@ -0,0 +1,12 @@
+import { useEffect } from 'react'
+import { isMobile } from 'utils/userAgent'
+
+export default function useDisableScrolling(disable: boolean | undefined | null) {
+  useEffect(() => {
+    if (disable && isMobile) {
+      document.body.style.overflow = 'hidden'
+    } else if (isMobile) {
+      document.body.style.overflow = 'auto'
+    }
```
zzmp, line 10: Simplify the mobile check by making the whole thing a nop:

```suggestion
    if (!isMobile) return
    document.body.style.overflow = disable ? 'hidden' : 'auto'
```

src/hooks/useDisableScrolling.ts
```diff
@@ -0,0 +1,12 @@
+import { useEffect } from 'react'
+import { isMobile } from 'utils/userAgent'
+
+export default function useDisableScrolling(disable: boolean | undefined | null) {
```
zzmp, line 4: Add a JSDoc, something like:

```suggestion
/** Disables scrolling on mobile when `true` is passed. */
export default function useDisableScrolling(disable: boolean | undefined | null) {
```

zzmp (changes requested):

zzmp (approved):

zzmp (approved):

PR #6742, opened 2023-06-09
fix: Catch routing-api v2 no liquidity error case

tinaszheng:
routing-api v2 has a new error format where the `error.detail` is `No quotes available`, so catch this case for displaying insufficient liquidity.

image

### DIFF
--- a/src/state/routing/v2Slice.ts
+++ b/src/state/routing/v2Slice.ts
@@ -80,7 +80,10 @@ export const routingApiV2 = createApi({
                 // cast as any here because we do a runtime check on it being an object before indexing into .errorCode
                 const errorData = response.error.data as any
                 // NO_ROUTE should be treated as a valid response to prevent retries.
-                if (typeof errorData === 'object' && errorData?.errorCode === 'NO_ROUTE') {
+                if (
+                  typeof errorData === 'object' &&
+                  (errorData?.errorCode === 'NO_ROUTE' || errorData?.detail === 'No quotes available')
+                ) {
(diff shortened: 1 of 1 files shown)


marktoda (approved):

just-toby (approved):

src/state/routing/v2Slice.ts
```diff
@@ -80,7 +80,10 @@ export const routingApiV2 = createApi({
                 // cast as any here because we do a runtime check on it being an object before indexing into .errorCode
                 const errorData = response.error.data as any
                 // NO_ROUTE should be treated as a valid response to prevent retries.
-                if (typeof errorData === 'object' && errorData?.errorCode === 'NO_ROUTE') {
+                if (
+                  typeof errorData === 'object' &&
```
zzmp, line 84: Do we want a unit test to capture these error cases?

src/state/routing/v2Slice.ts
```diff
@@ -80,7 +80,10 @@ export const routingApiV2 = createApi({
...
+                  (errorData?.errorCode === 'NO_ROUTE' || errorData?.detail === 'No quotes available')
```
zzmp, line 85: This should be in a const exported by the routing-api package. If that's not available, can you make it a follow-up with the routing-api team?

zzmp (approved):

src/state/routing/v2Slice.ts
tinaszheng: do you mean SOR? There is no routing-api package

PR #6749, opened 2023-06-12
fix: update web3-react

JFrankfurt:
## Description

This updates @web3-react/walletconnect-v2 to incorporate upstream fixes to the walletconnect modal/provider.

### DIFF
--- a/package.json
+++ b/package.json
@@ -196,7 +196,7 @@
     "@web3-react/types": "^8.2.0",
     "@web3-react/url": "^8.2.0",
     "@web3-react/walletconnect": "^8.2.0",
-    "@web3-react/walletconnect-v2": "8.3.2",
+    "@web3-react/walletconnect-v2": "8.3.3",
     "ajv": "^8.11.0",
     "ajv-formats": "^2.1.1",
     "array.prototype.flat": "^1.2.4",
--- a/yarn.lock
+++ b/yarn.lock
@@ -6157,10 +6157,10 @@
     "@walletconnect/types" "^1.8.0"
     "@walletconnect/utils" "^1.8.0"
 
-"@walletconnect/core@2.7.8":
-  version "2.7.8"
-  resolved "https://registry.yarnpkg.com/@walletconnect/core/-/core-2.7.8.tgz#e75329379cc61dc124c85619998a65eecabe4f53"
-  integrity sha512-Ptp1Jo9hv5mtrQMF/iC/RF/KHmYfO79DBLj77AV4PnJ5z6J0MRYepPKXKFEirOXR4OKCT5qCrPOiRtGvtNI+sg==
+"@walletconnect/core@2.8.0":
+  version "2.8.0"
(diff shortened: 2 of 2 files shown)


src/nft/components/details/detailsV2/__snapshots__/DataPageTable.test.tsx.snap
```diff
@@ -464,7 +464,7 @@ exports[`data page listings table content loads with a given asset 1`] = `
             <div
               class="c10 css-18hn7mq"
             >
-              5 months
+              4 months
```
natew, line 467: weird i had this too, and then in a later test it wanted 5 again, seems indeterministic?

natew (approved):

src/nft/components/details/detailsV2/__snapshots__/DataPageTable.test.tsx.snap
JFrankfurt: I let @cbachmeier know about this and I think he merged a fix for it yesterday. It's deterministic, but based on an offset from a fixed timestamp. He updated the test to account for this.

src/connection/WalletConnectV2.ts
```diff
@@ -33,11 +33,21 @@ export class WalletConnectV2Popup extends WalletConnect {
...
+          mobileWallets: undefined,
+          privacyPolicyUrl: undefined,
+          termsOfServiceUrl: undefined,
           themeMode: darkmode ? 'dark' : 'light',
           themeVariables: {
             '--w3m-font-family': '"Inter custom", sans-serif',
             '--w3m-z-index': Z_INDEX.modal.toString(),
           },
+          tokenImages: undefined,
+          walletImages: undefined,
```
JFrankfurt, line 50: The new upstream wc package requires explicit enumeration of all options 🤦

zzmp (approved):

src/connection/WalletConnectV2.ts
JFrankfurt: Here's the [upstream type](https://github.com/WalletConnect/walletconnect-monorepo/blob/5159f7767e14f503bea4e1718b0cdf3e0931a916/providers/ethereum-provider/src/EthereumProvider.ts#L14)
image

I'm wondering now if we could update[ this type ](https://github.com/Uniswap/web3-react/blob/main/packages/walletconnect-v2/src/index.ts#L11-L24)in `@web3-react/walletconnect-v2` to something like the following so we can get the correct optional typing. 
```
/**
 * Options to configure the WalletConnect provider.
 * For the full list of options, see {@link https://docs.walletconnect.com/2.0/javascript/providers/ethereum#initialization WalletConnect documentation}.
 */
export type WalletConnectOptions = Omit<Parameters<typeof WalletConnectProvider.init>[0], 'rpcMap'> & {
  /**
   * Map of chainIds to rpc url(s). If multiple urls are provided, the first one that responds
   * within a given timeout will be used. Note that multiple urls are not supported by WalletConnect by default.
   * That's why we extend its options with our own `rpcMap` (@see getBestUrlMap).
   */
  rpcMap?: { [chainId: number]: string | string[] }
  /** @deprecated Use `rpcMap` instead. */
  rpc?: { [chainId: number]: string | string[] }
}
```

PR #6752, opened 2023-06-12
feat: detect more injections

cartcrom:
## Description

Adds more injection detection via new `getInjection` utility. This adds:
* Phantom
* Trust Wallet
* Brave Browser
* Ledger Connect
* Rabby

Adds more wallet icon image files, moves them to a new subfolder, and makes them all square. Border radius is now handled by css.

## Screen capture

https://github.com/Uniswap/interface/assets/39385577/1ff60788-14ff-4c88-873e-edb2f7c51119

## Test plan
* Try a variety of injected wallets, confirm the expected icon/name appears
* connect to the app using mobile wallet browsers (but some mobile wallet browsers e.g. brave & phantom cannot be tested properly yet as they will not inject into interface on vercel/local)

### Automated testing
* new unit tests


src/assets/wallets/uniswap-wallet-icon.png
zzmp, line 1: Why isn't this one an svg like the rest? Because it's bundled in, and not served statically, using a png over an svg has real bundle size implications.

src/connection/utils.ts
```diff
@@ -1,14 +1,56 @@
...
+    if (detectedWallet) return detectedWallet
+  }
+
+  // Phantom sets it's flag in a different part of the window object
+  if (window.phantom?.ethereum?.isPhantom) return { name: 'Phantom', icon: PHANTOM_ICON }
+
+  // Check for MetaMask last, as other injectors will also set this flag, i.e. Brave browser and Phantom wallet
+  if (window.ethereum?.isMetaMask) return { name: 'MetaMask', icon: METAMASK_ICON }
+
+  // Prompt metamask installation when there is no injection present or the only injection detected is coinbase
```
zzmp, line 39: Clarify that we are not attacking coinbase lol

```suggestion
  // Prompt metamask installation when there is no injection present or the only injection detected is coinbase,
  // because coinbase already has its own entry in our UI
```

zzmp (approved):
still would like uni icon as an svg but not a blocker. we'll probably do an audit of bundle size at some point.

src/assets/wallets/uniswap-wallet-icon.png
cartcrom: The wallet icon unfortunately does not export well from figma, or there was something up a gradient in the icon that was making it blurry on certain browsers, so we opted to use a png here instead.

src/connection/utils.ts
```diff
@@ -1,14 +1,56 @@
...
+}
+
+/**
+ * Checks the window object for the presence of a known injectors and returns the most relevant injector name and icon.
+ * Returns a default metamask installation object if no wallet is detected.
+ *
+ * @param isDarkMode - optional parameter to determine which color mode of the
+ */
+export function getInjection(isDarkMode?: boolean): { name: string; icon: string } {
+  for (const key in window.ethereum) {
```
tinaszheng, line 27: wouldnt it be slightly more efficient to loop through the keys in `InjectedWalletTable` rather than window.ethereum

src/connection/utils.ts
```diff
@@ -1,14 +1,56 @@
...
+    const detectedWallet = InjectedWalletTable[key as keyof Window['ethereum']]
+
+    if (detectedWallet) return detectedWallet
+  }
+
+  // Phantom sets it's flag in a different part of the window object
```
tinaszheng, line 33:
```suggestion
  // Phantom sets its flag in a different part of the window object
```

tinaszheng (approved):

src/connection/utils.ts
```diff
@@ -1,14 +1,56 @@
...
+  if (window.phantom?.ethereum?.isPhantom) return { name: 'Phantom', icon: PHANTOM_ICON }
+
+  // Check for MetaMask last, as other injectors will also set this flag, i.e. Brave browser and Phantom wallet
```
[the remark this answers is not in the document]
cartcrom: ty ty

src/connection/utils.ts
```diff
@@ -1,14 +1,56 @@
...
+ */
+export function getInjection(isDarkMode?: boolean): { name: string; icon: string } {
+  for (const key in window.ethereum) {
+    const detectedWallet = InjectedWalletTable[key as keyof Window['ethereum']]
+
+    if (detectedWallet) return detectedWallet
+  }
+
+  // Phantom sets it's flag in a different part of the window object
+  if (window.phantom?.ethereum?.isPhantom) return { name: 'Phantom', icon: PHANTOM_ICON }
```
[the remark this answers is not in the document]
cartcrom: ikr

src/connection/utils.ts
cartcrom: so tru

src/connection/utils.ts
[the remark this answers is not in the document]
cartcrom: whats messed up is that there is a window.ethereum.isPhantom variable, but it only is initialized after the wallet is connected... T_T

tinaszheng (approved):

PR #6756, opened 2023-06-13
feat: move to swc-minify for better errors and faster builds

natew:
## Description

I'm debugging why minification is breaking here: https://github.com/Uniswap/interface/pull/6701

And using swc made the error way better.

### Before

CleanShot 2023-06-13 at 07 35 18@2x

### After

CleanShot 2023-06-13 at 07 34 02@2x

## Test plan

We'd need a bug bash to make sure this doesn't break anything, and I can do some analysis on bundle size difference if we want to go forward with this.

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test


zzmp:
We have upcoming work slated to migrate completely from babel to swc ([design doc](https://www.notion.so/uniswaplabs/Optimizing-interface-build-performance-29adef9205ae44928bac0e04ed50e382#c467cae8358f494d93a0a4a208bba32d)), and AFAIK we haven't seen issues with the minifier before; so I think this work should be deferred unless you run into similar problems again and - even if you do - timeboxxed if you do work on it. I don't know that it's worth spending a lot of time for a piecemeal migration when we have the entire migration already planned for the near future.

That said, I love that you've got this working with swc already <3. Want to help with the entire swc migration?

PR #6759, opened 2023-06-13
fix: currency select modal when chain is not supported

JackShort:
* fixes a bug where if you select the token select modal when connected to an unsupported network it will throw and error

### DIFF
--- a/src/nft/components/bag/BagFooter.tsx
+++ b/src/nft/components/bag/BagFooter.tsx
@@ -462,23 +462,27 @@ export const BagFooter = ({ setModalIsOpen, eventProperties }: BagFooterProps) =
         <FooterHeader gap="xs">
           <CurrencyRow>
             <Column gap="xs">
-              <ThemedText.SubHeaderSmall>
-                <Trans>Pay with</Trans>
-              </ThemedText.SubHeaderSmall>
-              <CurrencyInput
-                onClick={() => {
-                  if (!bagIsLocked && chainAllowed) {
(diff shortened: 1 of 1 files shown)


src/nft/components/bag/BagFooter.tsx
```diff
@@ -454,6 +454,7 @@ export const BagFooter = ({ setModalIsOpen, eventProperties }: BagFooterProps) =
     using_erc20: !!inputCurrency,
     ...eventProperties,
   }
+  const chainAllowed = isSupportedChain(chainId)
 
   return (
     <FooterContainer>
```
cbachmeier, line 460: I think it'd be better to hide the whole section so you don't get this view
Screen Shot 2023-06-13 at 15 41 26

cbachmeier (changes requested):

cbachmeier (approved):

src/nft/components/bag/BagFooter.tsx
```diff
@@ -454,30 +454,35 @@ export const BagFooter = ({ setModalIsOpen, eventProperties }: BagFooterProps) =
     using_erc20: !!inputCurrency,
     ...eventProperties,
   }
+  const chainAllowed = isSupportedChain(chainId)
```
just-toby, line 457: this variable is used once, you could just use `isSupportedChain(chainId)` on L465 instead

just-toby (approved):

just-toby (approved):

PR #6761, opened 2023-06-14
Feat/user prop router pref

just-toby:
## Description

logs the user's router preference as a custom user property to amplitude

_Linear ticket:_ WEB-2290

### QA (ie manual testing)

- [x] turned `debug` mode on for amplitude and verified that the effect hook runs only when the router preference is changed, and on initial page load.

### DIFF
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -9,6 +9,7 @@ import { useBag } from 'nft/hooks/useBag'
 import { lazy, Suspense, useEffect, useMemo, useState } from 'react'
 import { Navigate, Route, Routes, useLocation, useSearchParams } from 'react-router-dom'
 import { shouldDisableNFTRoutesAtom } from 'state/application/atoms'
+import { useRouterPreference } from 'state/user/hooks'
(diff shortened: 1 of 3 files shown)


src/pages/App.tsx
```diff
@@ -160,6 +162,10 @@ export default function App() {
     user.set(CustomUserProperties.DARK_MODE, isDarkMode)
   }, [isDarkMode])
 
+  useEffect(() => {
+    user.set(CustomUserProperties.ROUTER_PREFERENCE, routerPreference)
+  }, [routerPreference])
```
zzmp, line 167: `user` is outside of the React lifecycle - could this be set outside of a hook, in the reducer when the router preference is actually set? That would simplify and avoid coupling it to rendering.

zzmp (changes requested):

src/pages/App.tsx
zzmp: given the other user settings set here, I think it's fine to leave. This does set them all in the same file, so it's easier to track.

JFrankfurt (approved):

PR #6765, opened 2023-06-14
fix: remove sentry logging for swap failures

just-toby:
reverts https://github.com/Uniswap/interface/pull/6698/files

this logging is too noisy and causing false positive alerts in sentry + pagerduty. we want this mostly for analytics, so it will be better to use amplitude instead


vm (commented):
Can you add the context for why we're reverting? For future ref

vm (approved):

tinaszheng (approved):

PR #6766, opened 2023-06-14
fix: remove sentry logging for swap failures

just-toby:
reverts https://github.com/Uniswap/interface/pull/6698/files from the staging branch

this logging is too noisy and causing false positive alerts in sentry + pagerduty. we want this mostly for analytics, so it will be better to use amplitude instead

### DIFF
--- a/src/pages/Swap/index.tsx
+++ b/src/pages/Swap/index.tsx
@@ -1,5 +1,4 @@
 import { Trans } from '@lingui/macro'
-import * as Sentry from '@sentry/react'
 import { sendAnalyticsEvent, Trace, TraceEvent, useTrace } from '@uniswap/analytics'
 import {
   BrowserEvent,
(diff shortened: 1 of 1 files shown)


cbachmeier (approved):

cbachmeier:
staging equivalent to #6765

zzmp (changes requested):
Don't we still want to capture this data?
If we are removing this without re-adding another mechanism, can you track it in a linear ticket and mark it with a TODO?

just-toby:
> Don't we still want to capture this data? If we are removing this without re-adding another mechanism, can you track it in a linear ticket and mark it with a TODO?

@zzmp yes we do, but it's just simpler and faster to revert for the release. i don't want to try to rush in the amplitude logging into this release, it can go in the next one

zzmp (approved):

PR #6767, opened 2023-06-14
fix: handle both UNIWALLET and UNISWAP_WALLET in LocalStorage

cbachmeier:
## Description

 - resolves the bug found in prod where the app crashed for users previously connected via Uniswap because they had `UNIWALLET` as the selectedWallet in localStorage and it was updated to `UNISWAP_WALLET`. This resolves by autocasting to to the latter. Verified locally and by using vercel deploys.

### DIFF
--- a/src/state/user/reducer.ts
+++ b/src/state/user/reducer.ts
@@ -131,6 +131,12 @@ const userSlice = createSlice({
     // for all existing users with a previous version of the state in their localStorage.
     // In order to avoid this, we need to set a default value for each new property manually during hydration.
     builder.addCase(updateVersion, (state) => {
+      // If `selectedWallet` is ConnectionType.UNI_WALLET (deprecated) switch it to ConnectionType.UNISWAP_WALLET
+      // @ts-ignore
+      if (state.selectedWallet === 'UNIWALLET') {
+        state.selectedWallet = ConnectionType.UNISWAP_WALLET
+      }
+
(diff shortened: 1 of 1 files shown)


vm (approved):

src/state/user/reducer.ts
```diff
@@ -131,6 +131,12 @@ const userSlice = createSlice({
     // for all existing users with a previous version of the state in their localStorage.
     // In order to avoid this, we need to set a default value for each new property manually during hydration.
     builder.addCase(updateVersion, (state) => {
+      // If `selectedWallet` is ConnectionType.UNI_WALLET (deprecated) switch it to ConnectionType.UNISWAP_WALLET
+      // @ts-ignore
```
vm, line 135: why the ts-ignore?

JackShort (approved):
lgtm

src/state/user/reducer.ts
cbachmeier: React never expects `state.selectedWallet` to equal `UNIWALLET` so it throws an error

zzmp: because it is not a typed value of state.selectedWallet anymore.

This is good, but I'd like to see a more proper, segmented out migration codepath in the future. Can that be ticketed?

zzmp (approved):

src/state/user/reducer.ts
cbachmeier: cc: @JFrankfurt

PR #6768, opened 2023-06-14
refactor: add deprecated value back to enum

cbachmeier:
adds deprecated value back to enum and remove ts-ignore

Screen Shot 2023-06-14 at 14 29 06
Screen Shot 2023-06-14 at 14 29 00

### DIFF
--- a/src/connection/types.ts
+++ b/src/connection/types.ts
@@ -3,6 +3,10 @@ import { Connector } from '@web3-react/types'
 
 export enum ConnectionType {
   UNISWAP_WALLET = 'UNISWAP_WALLET',
+  /**
+   * @deprecated
+   */
+  UNIWALLET = 'UNIWALLET',
   INJECTED = 'INJECTED',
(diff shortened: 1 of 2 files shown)


src/connection/types.ts
```diff
@@ -3,6 +3,10 @@ import { Connector } from '@web3-react/types'
 
 export enum ConnectionType {
   UNISWAP_WALLET = 'UNISWAP_WALLET',
+  /**
+   * @deprecated
+   */
```
zzmp, line 8:
```suggestion
  /** @deprecated - Use {@link UNISWAP_WALLET} instead. */
```

I think `@link` is right but it may be a different decorator.

zzmp (approved):
you should be able to fix and not re-request because the base branch is not `main`, so lgtm

src/connection/types.ts
cbachmeier: links back to enum value as expected

PR #6769, opened 2023-06-14
feat: log minimum_output_after_slippage to amplitude

just-toby:
## Description

adds minimum output after slippage to these two events: swap signed, swap quote received

_Linear ticket:_ WEB-2292

### After

image

image

### QA (ie manual testing)

- [x] checked logs in debug mode

### DIFF
--- a/src/components/swap/SwapDetailsDropdown.tsx
+++ b/src/components/swap/SwapDetailsDropdown.tsx
@@ -1,5 +1,5 @@
 import { Trans } from '@lingui/macro'
-import { TraceEvent } from '@uniswap/analytics'
+import { TraceEvent, useTrace } from '@uniswap/analytics'
 import { BrowserEvent, InterfaceElementName, SwapEventName } from '@uniswap/analytics-events'
 import { Percent } from '@uniswap/sdk-core'
 import { useWeb3React } from '@web3-react/core'
@@ -8,6 +8,7 @@ import Column from 'components/Column'
 import { LoadingOpacityContainer } from 'components/Loader/styled'
 import { RowBetween, RowFixed } from 'components/Row'
 import { SUPPORTED_GAS_ESTIMATE_CHAIN_IDS } from 'constants/chains'
+import { formatEventPropertiesForTrade } from 'lib/utils/analytics'
 import { useState } from 'react'
 import { ChevronDown } from 'react-feather'
 import { InterfaceTrade } from 'state/routing/types'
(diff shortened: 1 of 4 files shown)


src/components/swap/SwapDetailsDropdown.tsx
```diff
@@ -102,13 +103,18 @@ export default function SwapDetailsDropdown({ trade, syncing, loading, allowedSl
...
   return (
     <Wrapper>
       <TraceEvent
         events={[BrowserEvent.onClick]}
         name={SwapEventName.SWAP_DETAILS_EXPANDED}
         element={InterfaceElementName.SWAP_DETAILS_DROPDOWN}
+        properties={{
+          ...(trade ? formatEventPropertiesForTrade(trade, allowedSlippage) : {}),
+          ...trace,
+        }}
```
JFrankfurt, line 117: this will create a new object and re-render `TraceEvent` repeatedly, right?

just-toby: @JFrankfurt i don't think it was causing much more rendering than the memoized version in my testing, but I memoized just to be safe

JFrankfurt (approved):

src/lib/utils/analytics.ts
```diff
@@ -76,5 +69,6 @@ export const formatSwapQuoteReceivedEventProperties = (
         : undefined,
     token_in_amount: formatToDecimal(trade.inputAmount, trade.inputAmount.currency.decimals),
     token_out_amount: formatToDecimal(trade.outputAmount, trade.outputAmount.currency.decimals),
+    minimum_output_after_slippage: trade.minimumAmountOut(allowedSlippage).toSignificant(6),
```
tinaszheng, line 72: maybe log the actual `allowedSlippage` here too? since there are BigNumber operations comparing outputAmount to minimumAmountOut may result in loss of precision

src/components/swap/SwapDetailsDropdown.tsx
tinaszheng: @JFrankfurt dont think rerenders work that way - creating a new object here will change the prop but rerenders only happen when there are downstream state changes

src/components/swap/SwapDetailsDropdown.tsx
```diff
@@ -102,13 +103,23 @@ export default function SwapDetailsDropdown({ trade, syncing, loading, allowedSl
...
   const [showDetails, setShowDetails] = useState(false)
+  const trace = useTrace()
+
+  const analyticsProps = useMemo(
+    () => ({
+      ...(trade ? formatEventPropertiesForTrade(trade, allowedSlippage) : {}),
+      ...trace,
+    }),
+    [allowedSlippage, trace, trade]
+  )
```
tinaszheng, line 114: dont need to memoize this - the events array is already a new object on every render so this doesnt add anything perf wise

tinaszheng (approved):

PR #6773, opened 2023-06-15
fix: log the error message instead of [object Object]

JFrankfurt:
## Description

We are currently debugging a permit-related bug in walletconnect v2 and this catch block is logging `[object Object]` instead of the `error.message` value.

_Linear:_ https://linear.app/uniswap/issue/WEB-2368/improve-permit2-error-logging

image

image

### DIFF
--- a/src/hooks/usePermitAllowance.ts
+++ b/src/hooks/usePermitAllowance.ts
@@ -8,8 +8,7 @@ import { useContract } from 'hooks/useContract'
 import { useSingleCallResult } from 'lib/hooks/multicall'
 import ms from 'ms.macro'
 import { useCallback, useEffect, useMemo, useState } from 'react'
-import { UserRejectedRequestError } from 'utils/errors'
-import { didUserReject } from 'utils/swapErrorToUserReadableMessage'
+import { throwReadableError } from 'utils/errors'
 
 const PERMIT_EXPIRATION = ms`30d`
 const PERMIT_SIG_EXPIRATION = ms`30m`
@@ -83,10 +82,7 @@ export function useUpdatePermitAllowance(
       return
     } catch (e: unknown) {
       const symbol = token?.symbol ?? 'Token'
-      if (didUserReject(e)) {
-        throw new UserRejectedRequestError(`${symbol} permit allowance failed: User rejected signature`)
-      }
-      throw new Error(`${symbol} permit allowance failed: ${e instanceof Error ? e.message : e}`)
+      throwReadableError(`${symbol} permit allowance failed:`, e)
     }
   }, [account, chainId, nonce, onPermitSignature, provider, spender, token])
 }
--- a/src/hooks/usePermitAllowance.ts
+++ b/src/hooks/usePermitAllowance.ts
@@ -8,7 +8,8 @@ import { useContract } from 'hooks/useContract'
 import { useSingleCallResult } from 'lib/hooks/multicall'
 import ms from 'ms.macro'
 import { useCallback, useEffect, useMemo, useState } from 'react'
-import { throwReadableError } from 'utils/errors'
+import { toReadableError, UserRejectedRequestError } from 'utils/errors'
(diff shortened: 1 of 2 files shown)


src/hooks/usePermitAllowance.ts
```diff
@@ -81,12 +81,12 @@ export function useUpdatePermitAllowance(
...
       onPermitSignature?.({ ...permit, signature })
       return
-    } catch (e: unknown) {
+    } catch (e) {
       const symbol = token?.symbol ?? 'Token'
       if (didUserReject(e)) {
         throw new UserRejectedRequestError(`${symbol} permit allowance failed: User rejected signature`)
       }
-      throw new Error(`${symbol} permit allowance failed: ${e instanceof Error ? e.message : e}`)
+      throw new Error(`${symbol} permit allowance failed: ${e instanceof Error || e.message ? e.message : e}`)
```
tinaszheng, line 89: you might be better off checking `typeof e === 'object' ? e.message ?? e.reason ?? 'unknown' : e`

tinaszheng: you could keep the `e: unknown` typing above

zzmp: should this just be a util? I feel like this pattern occurs a lot, and we often have this same bug - of logging an object when we don't mean to.

JFrankfurt: i like that

src/hooks/usePermitAllowance.ts
```diff
@@ -81,12 +81,14 @@ export function useUpdatePermitAllowance(
       const signature = await signTypedData(provider.getSigner(account), domain, types, values)
       onPermitSignature?.({ ...permit, signature })
       return
-    } catch (e: unknown) {
+    } catch (e) {
```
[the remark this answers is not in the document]
JFrankfurt: unfortunately, all else equal we cannot
image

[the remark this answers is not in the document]
JFrankfurt:
```
      if (typeof e === 'object' && e !== null) {
        const error = e as Error & { reason?: string }
        throw new Error(`${symbol} permit allowance failed: ${error.message ?? error.reason ?? 'unknown'}`)
      }
```

could do this 🤔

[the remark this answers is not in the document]
JFrankfurt: @tinaszheng lmk what you think of this latest commit

zzmp:
Can you make this a broader util?

src/utils/errors.ts
```diff
@@ -6,3 +8,14 @@ export class UserRejectedRequestError extends Error {
     this.name = 'UserRejectedRequestError'
   }
 }
+
+export function throwReadableError(errorText: string, error: unknown) {
+  if (didUserReject(error)) {
```
zzmp, line 13: This should be independent of didUserReject to avoid confusing callsites. There are also cases (like calls to the routing api) where we want to throw the parsed error without checking for user rejection. What about just toReadableError, so the return value can be thrown?

zzmp (changes requested):

src/utils/errors.ts
```diff
@@ -6,3 +6,11 @@ export class UserRejectedRequestError extends Error {
     this.name = 'UserRejectedRequestError'
   }
 }
+
+export function toReadableError(errorText: string, error: unknown) {
```
zzmp, line 10: nit: `message`, not `errorText`

src/utils/errors.ts
```diff
@@ -6,3 +6,11 @@ export class UserRejectedRequestError extends Error {
...
+  if (typeof error === 'object' && error !== null) {
+    const e = error as Error & { reason?: string }
+    return new Error(`${errorText} 👺 ${e.message ?? e.reason ?? 'unknown'}`)
```
zzmp, line 13: nit: use `:` instead of `👺`

zzmp (approved):

PR #6782, opened 2023-06-15
fix: clear dependent variable when trade inputs become invalid

just-toby:
## Description

the UX of the swap component depends on the trade becoming undefined to clear the *dependent* variable (i.e. the output amount, if the INPUT is the independent variable)

so this PR is needed to properly clear the output when the user clears the input.

### Before

https://github.com/Uniswap/interface/assets/66155195/934b5d88-3020-4a39-811a-cc5f66c3cd6c

### After

https://github.com/Uniswap/interface/assets/66155195/d3ac1a2c-8efd-4623-9467-da37824634fd

### Automated testing

- [x] Unit test
- [x] Integration/E2E test

### DIFF
--- a/src/hooks/useBestTrade.ts
+++ b/src/hooks/useBestTrade.ts
@@ -59,7 +59,7 @@ export function useBestTrade(
     debouncedAmount,
     debouncedOtherCurrency,
     routerPreference,
-    autoRouterSupported && shouldGetTrade
+    !(autoRouterSupported && shouldGetTrade) // skip fetching
   )
 
   const inDebounce =
--- a/src/hooks/useBestTrade.ts
+++ b/src/hooks/useBestTrade.ts
@@ -59,7 +59,7 @@ export function useBestTrade(
     debouncedAmount,
     debouncedOtherCurrency,
     routerPreference,
-    shouldGetTrade
+    autoRouterSupported && shouldGetTrade
   )
 
   const inDebounce =
--- a/src/hooks/useBestTrade.ts
+++ b/src/hooks/useBestTrade.ts
@@ -56,7 +56,7 @@ export function useBestTrade(
   const [routerPreference] = useRouterPreference()
   const routingAPITrade = useRoutingAPITrade(
     tradeType,
-    debouncedAmount,
+    amountSpecified ? debouncedAmount : undefined,
     debouncedOtherCurrency,
     routerPreference,
     !(autoRouterSupported && shouldGetTrade) // skip fetching
--- a/src/hooks/useBestTrade.ts
+++ b/src/hooks/useBestTrade.ts
@@ -56,9 +56,10 @@ export function useBestTrade(
   const [routerPreference] = useRouterPreference()
   const routingAPITrade = useRoutingAPITrade(
     tradeType,
-    autoRouterSupported && shouldGetTrade ? debouncedAmount : undefined,
+    debouncedAmount,
     debouncedOtherCurrency,
-    routerPreference
+    routerPreference,
+    shouldGetTrade
   )
 
   const inDebounce =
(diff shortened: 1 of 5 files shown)


src/hooks/useBestTrade.ts
```diff
@@ -56,9 +56,10 @@ export function useBestTrade(
   const [routerPreference] = useRouterPreference()
   const routingAPITrade = useRoutingAPITrade(
     tradeType,
-    autoRouterSupported && shouldGetTrade ? debouncedAmount : undefined,
+    debouncedAmount,
```
just-toby, line 59: don't pass an undefined `amountSpecified` to `useRoutingAPITrade` to force the polling interrupt - let's just have a separate variable below to control that

src/state/routing/useRoutingAPITrade.ts
```diff
@@ -41,7 +42,7 @@ export function useRoutingAPITrade<TTradeType extends TradeType>(
   const queryArgs = useRoutingAPIArguments({
     tokenIn: currencyIn,
     tokenOut: currencyOut,
-    amount: amountSpecified,
+    amount: shouldGetTrade ? amountSpecified : undefined,
```
just-toby, line 45: we still want to interrupt polling when `shouldGetTrade` is false, so we've moved this into this hook.

src/state/routing/useRoutingAPITrade.ts
```diff
@@ -77,8 +78,11 @@ export function useRoutingAPITrade<TTradeType extends TradeType>(
   const isCurrent = currentTradeResult === tradeResult
 
   return useMemo(() => {
-    if (!amountSpecified || isError || !queryArgs) {
-      return { state: TradeState.INVALID, trade: tradeResult?.trade }
+    if (!shouldGetTrade && amountSpecified) {
+      // If we don't want to fetch new trades, but have valid inputs, return the stale trade.
```
just-toby, line 82: `shouldGetTrade` means we aren't polling new trades, but if the input is valid we still want to return the stale trade. for the UI to display.

JackShort (approved):

src/state/routing/useRoutingAPITrade.ts
```diff
@@ -25,7 +25,8 @@ export function useRoutingAPITrade<TTradeType extends TradeType>(
   tradeType: TTradeType,
   amountSpecified: CurrencyAmount<Currency> | undefined,
   otherCurrency: Currency | undefined,
-  routerPreference: RouterPreference | typeof INTERNAL_ROUTER_PREFERENCE_PRICE
+  routerPreference: RouterPreference | typeof INTERNAL_ROUTER_PREFERENCE_PRICE,
+  shouldGetTrade = true
```
tinaszheng, line 29: i think more common practice to do the inverse - `skip` to hooks

src/state/routing/useRoutingAPITrade.ts
```diff
@@ -77,8 +78,11 @@ export function useRoutingAPITrade<TTradeType extends TradeType>(
   const isCurrent = currentTradeResult === tradeResult
 
   return useMemo(() => {
-    if (!amountSpecified || isError || !queryArgs) {
-      return { state: TradeState.INVALID, trade: tradeResult?.trade }
+    if (!shouldGetTrade && amountSpecified) {
+      // If we don't want to fetch new trades, but have valid inputs, return the stale trade.
+      return { state: TradeState.STALE, trade: tradeResult?.trade }
```
tinaszheng, line 83: where is this state used? dont see it anywhere else

tinaszheng (dismissed):

src/state/routing/useRoutingAPITrade.ts
just-toby: @tinaszheng it wasn't used, but i thought it might be useful to differentiate between INVALID and STALE since this is technically not invalid

src/hooks/useBestTrade.ts
zzmp: I think if `amountSpecified` is undefined it should be passed immediately, otherwise you have a weird delay in the UI from debouncing.

```suggestion
    amountSpecified ? debouncedAmount : undefined,
```

Another way to specify this would be in the definition of `debouncedAmount`.

zzmp (changes requested):

zzmp (approved):

PR #6786, opened 2023-06-16
feat(testnets): add toggle for testnets

JFrankfurt:
## Description

Adds a settings toggle in all environments to allow access to testnets.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2244/show-testnets-in-dropdown-selector

### Before
image
image

### After

 image
image

### QA (ie manual testing)

- [ ] toggle testnet access from the account drawer settings menu
- [ ] observer Sepolia appearing/disappearing

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test

### DIFF
--- a/src/components/AccountDrawer/SettingsMenu.tsx
+++ b/src/components/AccountDrawer/SettingsMenu.tsx
@@ -11,6 +11,7 @@ import ThemeToggle from 'theme/components/ThemeToggle'
 import { GitVersionRow } from './GitVersionRow'
 import { SlideOutMenu } from './SlideOutMenu'
 import { SmallBalanceToggle } from './SmallBalanceToggle'
+import { TestnetsToggle } from './TestnetsToggle'
 
 const InternalLinkMenuItem = styled(Link)`
   ${ClickableStyle}
@@ -48,9 +49,11 @@ const ThemeToggleContainer = styled.div`
   margin: 0 0 6px;
 `
 
-const BalanceToggleContainer = styled.div`
-  padding: 16px 0;
-  margin-bottom: 26px;
+const ToggleContainer = styled.div`
+  padding: 16px 0 0 0;
+  &:last-child {
+    margin-bottom: 24px;
+  }
 `
 
 export default function SettingsMenu({ onClose }: { onClose: () => void }) {
@@ -61,12 +64,17 @@ export default function SettingsMenu({ onClose }: { onClose: () => void }) {
       <SectionTitle>
         <Trans>Preferences</Trans>
       </SectionTitle>
-      <ThemeToggleContainer>
-        <ThemeToggle />
-      </ThemeToggleContainer>
-      <BalanceToggleContainer>
-        <SmallBalanceToggle />
-      </BalanceToggleContainer>
+      <div>
+        <ThemeToggleContainer>
+          <ThemeToggle />
+        </ThemeToggleContainer>
+        <ToggleContainer>
+          <SmallBalanceToggle />
+        </ToggleContainer>
+        <ToggleContainer>
+          <TestnetsToggle />
+        </ToggleContainer>
+      </div>
 
       <SectionTitle data-testid="wallet-header">
         <Trans>Language</Trans>
--- a/src/components/AccountDrawer/SettingsMenu.tsx
+++ b/src/components/AccountDrawer/SettingsMenu.tsx
@@ -45,15 +45,11 @@ const SectionTitle = styled(ThemedText.SubHeader)`
   padding-bottom: 24px;
 `
 
-const ThemeToggleContainer = styled.div`
-  margin: 0 0 6px;
-`
-
-const ToggleContainer = styled.div`
-  padding: 16px 0 0 0;
-  &:last-child {
-    margin-bottom: 24px;
-  }
+const ToggleWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: 16px;
+  margin-bottom: 24px;
 `
 
 export default function SettingsMenu({ onClose }: { onClose: () => void }) {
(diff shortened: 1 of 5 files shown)


vm:
Product nit: I find the wording to be confusing. Maybe we should be using `Show small balances` and `Show testnets` and flip the booleans?

src/components/NavBar/ChainSelector.tsx
```diff
@@ -72,6 +70,12 @@ export const ChainSelector = ({ leftAlign }: ChainSelectorProps) => {
 
   const theme = useTheme()
 
+  const [chains, setChains] = useState(NETWORK_SELECTOR_CHAINS)
+  const showTestnets = useAtomValue(showTestnetsAtom)
+  useEffect(() => {
+    setChains(NETWORK_SELECTOR_CHAINS.filter((chain) => !(chain === SupportedChainId.SEPOLIA && !showTestnets)))
+  }, [showTestnets])
```
just-toby, line 77: skip the filter step if `showTestnets` is true?

```suggestion
  useEffect(() => {
    if (!showTestnets) {
          setChains(NETWORK_SELECTOR_CHAINS.filter((chain) => !(chain === SupportedChainId.SEPOLIA)))
    }
  }, [showTestnets])
```

src/components/NavBar/ChainSelector.tsx
```diff
@@ -72,6 +70,12 @@ export const ChainSelector = ({ leftAlign }: ChainSelectorProps) => {
 
   const theme = useTheme()
 
+  const [chains, setChains] = useState(NETWORK_SELECTOR_CHAINS)
```
just-toby, line 73: this could be a `useMemo` instead although i'm not sure it brings any benefits. `useMemo` may be simpler to parse at first glance since we don't have to go find all the `setChains` callsites (in this case only 1 but reader doesn't know that)

src/components/NavBar/ChainSelector.tsx
just-toby: also this may be a needless optimization for a small array, but we don't really need a filter step at all. we could have 2 arrays (prodnets and testnets) and just do something like this:

`const chains = showTestnets ? NETWORK_SELECTOR_CHAINS : [... PRODUCTION_CHAINS , ...TESTNET_CHAINS]`

just-toby (approved):
no blockers just suggestions

src/components/AccountDrawer/SettingsMenu.tsx
```diff
@@ -48,9 +49,11 @@ const ThemeToggleContainer = styled.div`
   margin: 0 0 6px;
 `
 
-const BalanceToggleContainer = styled.div`
-  padding: 16px 0;
-  margin-bottom: 26px;
+const ToggleContainer = styled.div`
+  padding: 16px 0 0 0;
```
tinaszheng, line 53: feels like this is used as a gap between the different toggle options right? maybe put a `gap` property on the parent div instead, and have `margin-bottom` be on the parent only

src/components/NavBar/ChainSelector.tsx
```diff
@@ -72,6 +70,9 @@ export const ChainSelector = ({ leftAlign }: ChainSelectorProps) => {
 
   const theme = useTheme()
 
+  const showTestnets = useAtomValue(showTestnetsAtom)
+  const chains = NETWORK_SELECTOR_CHAINS.filter((chain) => !(chain === SupportedChainId.SEPOLIA && !showTestnets))
```
tinaszheng, line 74: y u special case sepolia?

tinaszheng (changes requested):

src/components/NavBar/ChainSelector.tsx
cbachmeier, line 74: For future testnet addition can you add a `TESTNETS` array in `src/constants/chains.ts` which contains all testnets and then here you can just check if chain is in that array

src/components/NavBar/ChainSelector.tsx
```diff
@@ -29,12 +30,9 @@ const NETWORK_SELECTOR_CHAINS = [
   SupportedChainId.ARBITRUM_ONE,
   SupportedChainId.CELO,
   SupportedChainId.BNB,
+  SupportedChainId.SEPOLIA,
```
cbachmeier, line 33: We support other testnets such as Polygon Mumbai, do we want to add it as well?

cbachmeier (changes requested):

cbachmeier:
More on design but this is taking up a lot of space on mobile
Screen Shot 2023-06-16 at 11 35 07

cbachmeier (approved):
LGTM but I would also implement Tina's suggestion about padding

src/components/NavBar/ChainSelector.tsx
```diff
@@ -72,6 +62,9 @@ export const ChainSelector = ({ leftAlign }: ChainSelectorProps) => {
 
   const theme = useTheme()
 
+  const showTestnets = useAtomValue(showTestnetsAtom)
+  const chains = NETWORK_SELECTOR_CHAINS.filter((chain) => !(TESTNET_SET.has(chain) && !showTestnets))
```
tinaszheng, line 66: sry one more nit - this is kinda confusing to me, wdyt 

`const chains = showTestnets ? NETWORK_SELECTOR_CHAINS : NETWORK_SELECTOR_CHAINS.filter(chain => !TESTNET_SET.has(chain))`

JFrankfurt:
```suggestion
  const chains = showTestnets ? NETWORK_SELECTOR_CHAINS : NETWORK_SELECTOR_CHAINS.filter(chain => !TESTNET_SET.has(chain))
```

tinaszheng (approved):

JackShort (approved):

src/constants/chains.ts
```diff
@@ -87,6 +87,16 @@ export const TESTNET_CHAIN_IDS = [
   SupportedChainId.OPTIMISM_GOERLI,
 ] as const
 
+export const TESTNET_SET = new Set([
```
tinaszheng, line 90: wait why add a new one here with more testnets instead of using existing `TESTNET_CHAIN_IDS`? seems like these are out of sync now?

PR #6788, opened 2023-06-16
chore: matching swap input number formatting to mobile

JackShort:
* went through the swap flow to update all number formatting to use coned and it was made to match the same formatting we use in the swap flow on mobile

## Before
Screen Shot 2023-06-14 at 3 58 55 PM

## After
Screen Shot 2023-06-14 at 3 59 15 PM


src/pages/Swap/index.tsx
```diff
@@ -359,7 +359,7 @@ export function Swap({
       [independentField]: typedValue,
       [dependentField]: showWrap
         ? parsedAmounts[independentField]?.toExact() ?? ''
-        : formatTransactionAmount(currencyAmountToPreciseFloat(parsedAmounts[dependentField])),
+        : formatCurrencyAmount(parsedAmounts[dependentField], NumberType.SwapTradeAmount, ''),
```
cbachmeier, line 362: nit: is that last arg of the empty string needed?

cbachmeier (approved):
LGTM

src/pages/Swap/index.tsx
JackShort: yeah

zzmp (approved):

PR #6789, opened 2023-06-16
feat: log swap failures to amplitude

just-toby:
## Description

logs swap failures to amplitude for trend tracking.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2307/log-swap-failures-to-amplitude

## Screen capture

image

### Reproducing the error

1. enabled debug mode locally and verified the `tradeToConfirm` is properly included in the 

### QA (ie manual testing)

- [x] manually verified the event logs at the appropriate time

### DIFF
--- a/src/pages/Swap/index.tsx
+++ b/src/pages/Swap/index.tsx
@@ -441,6 +441,11 @@ export function Swap({
         })
       })
       .catch((error) => {
+        if (!didUserReject(error)) {
+          sendAnalyticsEvent(SwapEventName.SWAP_ERROR, {
+            confirmedTrade: tradeToConfirm,
+          })
+        }
         setSwapState((currentState) => ({
           ...currentState,
           swapError: error,
@@ -455,6 +460,7 @@ export function Swap({
     account,
     trade?.inputAmount?.currency?.symbol,
     trade?.outputAmount?.currency?.symbol,
+    tradeToConfirm,
   ])
 
   // errors


src/pages/Swap/index.tsx
```diff
@@ -441,6 +441,11 @@ export function Swap({
         })
       })
       .catch((error) => {
+        if (!didUserReject(error)) {
+          sendAnalyticsEvent(SwapEventName.SWAP_ERROR, {
+            confirmedTrade: tradeToConfirm,
```
vm, line 446: are you sure that sending a full object will be okay here? if yes then all good

vm: there should be a way to test sending into the amplitude dev environment if you want to try that. i worry you won't be able to filter properly this way

natew (approved):

src/pages/Swap/index.tsx
just-toby: https://analytics.amplitude.com/uniswap/project/396672/search/amplitude_id%3D525267312650?sessionHandle=YjFTPAt_YjFTPgP_HpMXyAK_Bg2A_19308d93-cd11-4971-961e-fc55a0f758c9&eventId=e2ba7448-8ff4-4fdd-9ba8-9c1f3c739a85

yes it seems fine! checked with blair too

zzmp (approved):

PR #6792, opened 2023-06-16
chore: rm committed t9n

zzmp:
## Description

Removes committed translations, and prevents their re-committal by adding them to the `.gitignore`.
These are now added through the release pipeline, which uses `git add -f` to override the `.gitignore`.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2279/generate-translations-as-part-of-release-pipeline


JackShort (approved):

cypress/e2e/wallet-dropdown.test.ts
```diff
@@ -21,13 +21,17 @@ describe('Wallet Dropdown', () => {
...
-  function itShouldChangeTheLanguage() {
-    it('should select a language', () => {
-      cy.get(getTestSelector('wallet-language-item')).contains('Deutsch').click({ force: true })
-      cy.get(getTestSelector('wallet-header')).should('contain', 'Sprache')
+  function itChangesLocale() {
+    it('should change locale', () => {
+      cy.contains('Uniswap available in: English').should('not.exist')
+
+      cy.get(getTestSelector('wallet-language-item')).contains('Afrikaans').click({ force: true })
+      cy.location('hash').should('match', /\?lng=af-ZA$/)
```
cbachmeier, line 29: Can we add a test to see if the translations actually loaded? The presence of the query param and switch to english button do not necessarily guarantee that.

cypress/e2e/wallet-dropdown.test.ts
```diff
@@ -1,8 +1,8 @@
 import { getTestSelector } from '../utils'
```
cbachmeier, line 1: non-blocking nit: these test improvements aren't related to removing and ignoring the translation files and would preferably be in their own PR

cbachmeier (changes requested):

cypress/e2e/wallet-dropdown.test.ts
zzmp: We cannot, because the translations are no longer available on main.

That's why I've changed these tests in this PR - so that we continue to test without needing to rely on actual translations. Translations are now not available until the staging branch.

cbachmeier: This seems like a regression, can we add a way to verify translation mounting as part of the release to staging process?

zzmp: Yes, this does not verify that translations were downloaded nor that they mounted. That can no longer be done as a CI check, but it _can_ be rewritten as a test after cutting the staging branch. I'll move it there.

zzmp: I propose adding a new `test-t9n.yml` workflow that will run on push to `releases/staging` and will run a cypress test to verify that translations have downloaded and mount, and will post to #web-build-alerts if it fails but not block anything. This will allow us to retain t9n as part of the release pipeline _with test coverage_ while still minimizing build time for PRs (by eliminating t9n from PR builds).

wdyt? If you're in favor, I'd like to do that as a fast-follow PR.

cbachmeier: What do we see as the process for handling when those tests fail? Since it's not blocking release there won't be an immediate need to fix before releasing to prod and I could see this failing test just becoming noise for the build sheriff.

Can we have the test also run on pushes `releases/prod` and it be blocking there? This way if it errors on the way to staging, Sheriff can deploy a hotfix to staging and main while still allowing the staging deploy to be created for testing/

zzmp: I don't think it should be blocking anywhere, otherwise we risk bricking the release pipeline over a flaky test. I've spec'ed this a little more in https://www.notion.so/uniswaplabs/Include-web-translations-in-the-release-pipeline-baa996cba77c49db95c88306f0cb64c1?pvs=4#7c502a0bb28d40b882e1360ccfcaf247.

cbachmeier: I'm fine approving this atm but do want to continue the conversation of requiring test for prod in general. These would be a separate PR however

cbachmeier (approved):

PR #6794, opened 2023-06-16
chore: rm pseudo locale

zzmp:
## Description

Removes the unused pseudo-locale.

### DIFF
--- a/lingui.config.ts
+++ b/lingui.config.ts
@@ -115,13 +115,11 @@ const linguiConfig = {
     'vi-VN',
     'zh-CN',
     'zh-TW',
-    'pseudo',
   ],
   orderBy: 'messageId',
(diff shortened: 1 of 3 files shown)


cbachmeier (commented):
looked into what psuedo is used for and[ it seems minorly useful](https://lingui.dev/guides/pseudolocalization), are we sure we have no further interest in keeping it? Is it presence causing any issues?

zzmp:
It's presence adds complexity, and as far as I know it is unused. I added it about a year ago but it was never really used. 

I think it's simple to remove it and re-add it if it's desired, but I don't think we focus on our translations enough to make it worthwhile.

cbachmeier (approved):

JackShort (approved):

PR #6797, opened 2023-06-17
feat: wrangler infra and test environment

LunrEclipse:
## Description

Created testing infrastructure to prepare for Cloudflare Functions for dynamic metadata.

Added `wrangler` dependency and a script called `yarn test:wrangler` to manually setup a wrangler proxy on the localhost and test Cloudflare functions manually.

Added `jest.config.json` separate of the normal CRA jest config to for testing specifically on Cloudflare Function tests. These specific jest configs are only applied to tests within the `functions` folder and can be ran with the script `yarn test:cloud`. Also created `global-setup.js` and `global-teardown.js` using `jest-dev-server` to automatically setup `wrangler` proxy server before tests are ran to see how the server responds to requests. 

_Linear ticket:_ WEB-2354

### QA (ie manual testing)

- [x] Run the normal `yarn test` script to make sure existing tests are not impacted.
lude different devices and screen sizes that may be affected, and how you've tested them. -->

### Automated testing

- [x] Unit test - Added a basic `nft.test.ts` to show that the test is indeed ran and is successful


functions/global-setup.js
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable */
+const { setup: setupDevServer } = require('jest-dev-server')
+
+module.exports = async function globalSetup() {
+  globalThis.servers = await setupDevServer({
+    command: `yarn start:wrangler`,
+  })
+  await new Promise((r) => setTimeout(r, 30000))
```
natew, line 8: this is just waiting 30s before starting tests? could we do a `wait-port` type thing instead?

LunrEclipse: I originally tried doing a `port-ready`, but for some reason it didn't seem to work with `wrangler` so I just defaulted to a wait which worked.

LunrEclipse: Will look into potentially other libraries

LunrEclipse: `wait-port` worked so jut pushed that into the newest commit!

natew (approved):

natew:
also dep-test breaking just heads up

LunrEclipse:
> just curious why the pre-release version?

Yeah, so I was having an issue with the types which I found a patch for, but I don't think it has been included yet in Wrangler's latest release. 

https://github.com/cloudflare/workers-sdk/pull/3048

functions/global-setup.js
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable */
```
zzmp, line 1: Instead of disabling eslint, can you get this to work through typescript?
You'll probably need a custom tsconfig for the `functions/` folder - see `cypress/tsconfig.json` for an example.

functions/global-setup.js
```diff
@@ -0,0 +1,8 @@
...
+const { setup: setupDevServer } = require('jest-dev-server')
```
zzmp, line 2: This should use import so that it has typings.

functions/nft.test.ts
```diff
@@ -0,0 +1,15 @@
...
+const waitPort = require('wait-port')
+
+const params = {
+  port: 3001,
+  host: 'localhost',
+}
+
+beforeAll(async () => {
+  await waitPort(params)
+}, 60000)
```
zzmp, line 11: Instead of using `waitPort`, you can directly specify this in global-setup. Use the `port` option of `jest-dev-server`. This is a better option because it will delay _all_ tests, instead of just this test suite.

package.json
```diff
@@ -19,13 +19,16 @@
     "i18n": "yarn i18n:extract --clean && yarn i18n:compile",
     "prepare": "concurrently \"npm:ajv\" \"npm:contracts\" \"npm:graphql\" \"npm:i18n\"",
     "start": "craco start",
+    "start:wrangler": "NODE_OPTIONS=--dns-result-order=ipv4first npx wrangler pages dev --proxy=3001 --port=3000 -- yarn start:3001",
+    "start:3001": "PORT=3001 craco start",
```
zzmp, line 23:
```suggestion
    "start:wrangler": "NODE_OPTIONS=--dns-result-order=ipv4first PORT=3001 npx wrangler pages dev --proxy=3001 --port=3000 -- yarn start",
```

zzmp (changes requested):

functions/global-teardown.ts
```diff
@@ -0,0 +1,5 @@
+import { teardown } from 'jest-dev-server'
+
+module.exports = async function globalTeardown() {
+  await teardown(await globalThis.servers)
```
zzmp, line 4:
```suggestion
  await teardown(globalThis.servers)
```

functions/global.d.ts
```diff
@@ -0,0 +1,6 @@
+import { setup } from 'jest-dev-server'
+
+declare global {
+  // eslint-disable-next-line no-var
+  var servers: ReturnType<typeof setup>
```
zzmp, line 5:
```suggestion
  var servers: Awaited<ReturnType<typeof setup>>
```

functions/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "compilerOptions": {
+    "esModuleInterop": true,
+    "incremental": true,
+    "lib": ["DOM", "DOM.Iterable", "ESNext"],
+    "noEmit": true,
+    "strict": true,
+    "target": "ES6",
+    "tsBuildInfoFile": "../node_modules/.cache/tsbuildinfo/jest", // avoid clobbering the build tsbuildinfo
```
zzmp, line 9:
```suggestion
    "tsBuildInfoFile": "../node_modules/.cache/tsbuildinfo/functions", // avoid clobbering the build tsbuildinfo
```

functions/tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+    "types": ["jest", "node"],
+    "jsx": "react",
+    "moduleResolution": "NodeNext",
```
zzmp, line 12:
```suggestion
    "jsx": "react"
```

functions/tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+  },
+  "exclude": ["node_modules"],
+  "include": ["**/*.ts", "global.d.ts"],
```
zzmp, line 15: `global.d.ts` should already be matched by `**/*.ts`.

```suggestion
  "include": ["**/*.ts"],
```

package.json
```diff
@@ -19,13 +19,15 @@
     "i18n": "yarn i18n:extract --clean && yarn i18n:compile",
     "prepare": "concurrently \"npm:ajv\" \"npm:contracts\" \"npm:graphql\" \"npm:i18n\"",
     "start": "craco start",
+    "start:wrangler": "NODE_OPTIONS=--dns-result-order=ipv4first PORT=3001 npx wrangler pages dev --proxy=3001 --port=3000 -- yarn start",
```
zzmp, line 22: I recommend matching your start and test commands: `start:cloud`/`test:cloud` or `start:wrangler`/`test:wrangler`.

```suggestion
    "start:cloud": "NODE_OPTIONS=--dns-result-order=ipv4first PORT=3001 npx wrangler pages dev --proxy=3001 --port=3000 -- yarn start",
```

package.json
```diff
@@ -212,6 +217,8 @@
     "graphql-request": "^3.4.0",
     "immer": "^9.0.6",
     "inter-ui": "^3.13.1",
+    "jest": "^29.6.1",
```
zzmp, line 220: Can we use the same version of jest that is already installed, to avoid adding a new dependency?

package.json
```diff
@@ -247,6 +254,7 @@
     "statsig-react": "^1.22.0",
     "styled-components": "^5.3.5",
     "tiny-invariant": "^1.2.0",
+    "ts-jest": "^29.1.1",
```
zzmp, line 257: Do we need to use ts-jest, or can we use babel-jest to avoid adding a new dependency?

zzmp (changes requested):

package.json
```diff
@@ -131,6 +133,8 @@
   },
   "dependencies": {
     "@apollo/client": "^3.7.2",
+    "@babel/preset-env": "^7.22.7",
```
zzmp, line 136: These should all be added to `devDependencies`, not `dependencies`, because they are only needed as part of development.

functions/tsconfig.json
LunrEclipse: I only added this since for some reason without the setting I was getting an error with imports and package resolution. 

Screenshot 2023-07-07 at 10 36 56 AM

Is there a better way to resolve this issue?

zzmp: Weird. I think you should leave as is, then. We can always review/polish later.

zzmp (approved):

PR #6800, opened 2023-06-19
feat: sort supported chains to top of the chain selector list

JFrankfurt:
## Description

Adds a sort to the chain selector so unsupported options show up at the bottom. 

### Before
| paste_before | image|

### After
| paste_after  | image|

### Reproducing the error

1. 

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test


src/components/NavBar/ChainSelector.tsx
```diff
@@ -69,9 +69,13 @@ export const ChainSelector = ({ leftAlign }: ChainSelectorProps) => {
...
 
   const showTestnets = useAtomValue(showTestnetsAtom)
-  const chains = showTestnets
-    ? NETWORK_SELECTOR_CHAINS
-    : NETWORK_SELECTOR_CHAINS.filter((chain) => !TESTNET_CHAIN_IDS.has(chain))
+  const walletSupportsChain = useWalletSupportedChains()
+
+  const chains = useMemo(() => {
+    return (
+      showTestnets ? NETWORK_SELECTOR_CHAINS : NETWORK_SELECTOR_CHAINS.filter((chain) => !TESTNET_CHAIN_IDS.has(chain))
```
zzmp, line 76: nit: stick the ternary inside the filter function so it's easier to follow, ie 'return showTestnets ? true : TESTNET_CHAIN_IDS.has(chainId)'

zzmp (approved):
Just to verify: are unsupported chains also grayed out?

Some design questions:
- why are we showing unsupported chains at all? Should it be an option to just hide them?
- should there be a horizontal rule between supported and unsupported chains?

src/components/NavBar/ChainSelector.tsx
zzmp: Could you also add unit tests around this logic?

just-toby (approved):

Issue #6801, opened 2023-06-20
WalletConnectV2 expects all accounts to be under `eip155` namespace.


antondalgren:
**Bug Description**
> TypeError: Cannot read properties of undefined (reading 'accounts')
    at jt (https://interface-git-add-wc-v2-uniswap.vercel.app/static/js/main.e1825944.js:1:788270)
    at kt (https://interface-git-add-wc-v2-uniswap.vercel.app/static/js/main.e1825944.js:1:789476)
    at Sa (https://interface-git-add-wc-v2-uniswap.vercel.app/static/js/5917.3aead96b.js:2:913192)
    at Ss (https://interface-git-add-wc-v2-uniswap.vercel.app/static/js/5917.3aead96b.js:2:970305)
    at _c (https://interface-git-add-wc-v2-uniswap.vercel.app/static/js/5917.3aead96b.js:2:959411)
    at gc (https://interface-git-add-wc-v2-uniswap.vercel.app/static/js/5917.3aead96b.js:2:959339)
    at mc (https://interface-git-add-wc-v2-uniswap.vercel.app/static/js/5917.3aead96b.js:2:959202)
    at ac (https://interface-git-add-wc-v2-uniswap.vercel.app/static/js/5917.3aead96b.js:2:955982)
    at cc (https://interface-git-add-wc-v2-uniswap.vercel.app/static/js/5917.3aead96b.js:2:956373)
    at Zi (https://interface-git-add-wc-v2-uniswap.vercel.app/static/js/5917.3aead96b.js:2:896913)
    
`return getChainsFromEIP155Accounts((connector as WalletConnect).provider?.session?.namespaces.eip155.accounts)`

**Steps to Reproduce**
Use a wallet that returns the session namespaces on the format:
```
{
  "eip155:1": {
      "accounts": ["eip155:1:address"],
      ....
   },
   "eip155:137": {
      "accounts": ["eip155:137:address"],
      ....
   },
}
```

**Expected Behavior**
It should be able to work with these kind of responses as it is also a valid response according to the spec. It allows wallet developers to be more reserved in what methods and events they want the dapp to have access to for a specific chain.

PR #6802, opened 2023-06-20
fix: allow more session namespace keys

JFrankfurt:
## Description

Sessions have a number of possible ways to communicate what chains are valid. Wallets may choose to use any of them, so we need to query each. 

1. namespace.chains (optional array of string chainIds)
2. key encoded chain i.e., `{ "eip155:1": Namespace }`
3. account encoded chain i.e., `{ "eip155": { accounts: ['eip155:1:0xdeadbeef'] } }`

**WalletConnect Session object type**
```
export declare namespace SessionTypes {
    type Expiry = number;
    interface BaseNamespace {
        chains?: string[];
        accounts: string[];
        methods: string[];
        events: string[];
    }
    type Namespace = BaseNamespace;
    type Namespaces = Record<string, Namespace>;
    interface Struct {
        topic: string;
        pairingTopic: string;
        relay: RelayerTypes.ProtocolOptions;
        expiry: Expiry;
        acknowledged: boolean;
        controller: string;
        namespaces: Namespaces;
        requiredNamespaces: ProposalTypes.RequiredNamespaces;
        optionalNamespaces: ProposalTypes.OptionalNamespaces;
        sessionProperties?: ProposalTypes.SessionProperties;
        self: {
            publicKey: string;
            metadata: SignClientTypes.Metadata;
        };
        peer: {
            publicKey: string;
            metadata: SignClientTypes.Metadata;
        };
    }
}
```

issue: #6801 

### Reproducing the error

1. 

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test

### DIFF
--- a/src/utils/getSupportedChainIdFromWalletConnectSession.ts
+++ b/src/utils/getSupportedChainIdFromWalletConnectSession.ts
@@ -1,0 +1,35 @@
+import type { SessionTypes } from '@walletconnect/types'
+import { SupportedChainId } from 'constants/chains'
+
+// Helper function to extract chainId from string in format 'eip155:{chainId}'
+function getChainIdFromFormattedString(item: string): number | null {
+  const splitItem = item.split(':')
+  return splitItem.length > 1 && !isNaN(Number(splitItem[1])) ? Number(splitItem[1]) : null
+}
+
+export function getSupportedChainIdFromWalletConnectSession(session?: SessionTypes.Struct): SupportedChainId[] {
+  if (!session?.namespaces) return []
+
+  const eip155Keys = Object.keys(session.namespaces)
+  const namespaces = Object.values(session.namespaces)
+
+  // Collect all arrays into one for unified processing
+  const allItems = [
+    ...eip155Keys,
+    ...namespaces.flatMap((namespace) => namespace.chains),
+    ...namespaces.flatMap((namespace) => namespace.accounts),
+  ]
+
+  // Process all items to extract chainIds
+  const allChainIds = allItems
+    .map((item) => {
+      if (typeof item === 'string') {
+        return getChainIdFromFormattedString(item)
+      }
+      // Check if the item is a number
+      return isNaN(Number(item)) ? null : Number(item)
+    })
+    .filter((item) => item !== null) // Filter out any null values
+
+  return Array.from(new Set(allChainIds)) as SupportedChainId[]
+}
--- a/package.json
+++ b/package.json
@@ -103,6 +103,7 @@
     "@vanilla-extract/babel-plugin": "^1.1.7",
     "@vanilla-extract/jest-transform": "^1.1.1",
     "@vanilla-extract/webpack-plugin": "^2.1.11",
+    "@walletconnect/types": "^2.8.1",
     "babel-plugin-istanbul": "^6.1.1",
     "buffer": "^6.0.3",
     "concurrently": "^8.0.1",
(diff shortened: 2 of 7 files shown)


src/utils/getSupportedChainIdFromWalletConnectSession.ts
```diff
@@ -0,0 +1,35 @@
+import type { SessionTypes } from '@walletconnect/types'
+import { SupportedChainId } from 'constants/chains'
+
+// Helper function to extract chainId from string in format 'eip155:{chainId}'
+function getChainIdFromFormattedString(item: string): number | null {
+  const splitItem = item.split(':')
```
spenciefy, line 6: could consider enforcing `eip155:` format instead of just splitting at `:`

spenciefy:
```
 const splitItem = chain.startsWith('eip155:') ? chain.split(':') : []
return splitItem.length > 1 && !isNaN(Number(splitItem[1])) ? Number(splitItem[1]) : null
```

src/utils/getSupportedChainIdFromWalletConnectSession.ts
```diff
@@ -0,0 +1,35 @@
...
+  return splitItem.length > 1 && !isNaN(Number(splitItem[1])) ? Number(splitItem[1]) : null
+}
+
+export function getSupportedChainIdFromWalletConnectSession(session?: SessionTypes.Struct): SupportedChainId[] {
```
spenciefy, line 10: plural chainIds?
 
```suggestion
export function getSupportedChainIdsFromWalletConnectSession(session?: SessionTypes.Struct): SupportedChainId[] {
```

src/utils/getSupportedChainIdFromWalletConnectSession.ts
```diff
@@ -0,0 +1,35 @@
...
+  if (!session?.namespaces) return []
+
+  const eip155Keys = Object.keys(session.namespaces)
+  const namespaces = Object.values(session.namespaces)
+
+  // Collect all arrays into one for unified processing
+  const allItems = [
```
spenciefy, line 17: technically wouldn't a "bad" namespace request like this with mismatched chains and accounts pass this function and return [10, 137, 1, 42161]?

not a huge deal since that should be an issue on the wallet side 

```
 const namespaces = {
      eip155: {
        chains: ['10', '137'],
        accounts: ['eip155:1:0xc0ffee254729296a45a3885639AC7E10F9d54979', 'eip155:42161:address'],
        methods: [],
        events: [],
      },
    }
```

spenciefy (approved):

src/utils/getSupportedChainIdFromWalletConnectSession.ts
JFrankfurt: Yeah it would pass, but I don't really want to go down the route of enforcing the spec client side 🙈

JFrankfurt:
@antondalgren does this sufficiently address your bug?

antondalgren:
> @antondalgren does this sufficiently address your bug?

Thanks for addressing this issue swiftly. @JFrankfurt 

Wanted to let you know, in case you didn't know, that the WC team has build nice parsing utils around namespaces. https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/utils/src/namespaces.ts 

I just want a test case to cover this format, as this is how we return the namespace. I believe the code already covers it, but doesnt hurt to add one extra testcase. 
```
    const namespaces = {
      'eip155:1': {
        accounts: ['eip155:1:0xc0ffee254729296a45a3885639AC7E10F9d54979'],
        methods: [],
        events: [],
      },
      'eip155:137': {
        accounts: ['eip155:137:0xdeadbeef'],
        methods: [],
        events: [],
      },
    }
```

natew (approved):
not really sure what's going on here tbh but on a functional level the code looks fine and tests are there, so if anyone else with context approves then lgtm

just-toby (approved):

PR #6804, opened 2023-06-20
fix: 0 showing as collected fee amount

JackShort:
## Description
* fixes bug where 0s are showed for the amount when collecting LP position fee

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-1906/fee-collection-from-pool-appearing-as-0-in-activity-feed

### Before
Screen Shot 2023-06-20 at 4 09 55 PM

### After
Screen Shot 2023-06-20 at 4 08 25 PM


zzmp (changes requested):
Can you construct a unit test with mocked network responses (unit tests should not rely on network)?

JackShort:
> Can you construct a unit test with mocked network responses (unit tests should not rely on network)?

we actually have this in `parseLocal` and tests and it tests all of the functions that

src/pages/Pool/PositionPage.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+        CurrencyAmount.fromRawAmount(USDC_MAINNET, '1224156977'),
+        CurrencyAmount.fromRawAmount(WETH9[1], '500807669662847869'),
+      ]
+    })
+
+    render(<PositionPage />)
+
+    const collectFeesButton = screen.queryByTestId('collect-fees-button') as HTMLButtonElement
+    expect(collectFeesButton).toBeInTheDocument()
+    expect(screen.getByText('Collect fees')).toBeInTheDocument()
```
zzmp, line 74: You mock it so it renders actual numbers, but you should also assert on that: expect(...).toContain(actualNumbers)...

zzmp (changes requested):
sorry last nit to get the tests up to prevent regressing

zzmp (approved):

vm (approved):

PR #6808, opened 2023-06-21
fix: remove blue dot next to buy

LunrEclipse:
## Description

Removed blue dot next to buy for users that have completed the buy flow. 

_Linear ticket:_ WEB-2314
_Slack thread:_
_Relevant docs:_

### Before

Screenshot 2023-06-21 at 2 50 43 PM

### After

Screenshot 2023-06-21 at 2 50 57 PM

### DIFF
--- a/src/components/swap/SwapBuyFiatButton.tsx
+++ b/src/components/swap/SwapBuyFiatButton.tsx
@@ -6,7 +6,6 @@ import { useAccountDrawer } from 'components/AccountDrawer'
 import { ButtonText } from 'components/Button'
 import { MouseoverTooltip } from 'components/Tooltip'
 import { useCallback, useEffect, useState } from 'react'
-import { useBuyFiatFlowCompleted } from 'state/user/hooks'
 import styled from 'styled-components/macro'
 import { ExternalLink } from 'theme'
 
@@ -39,7 +38,6 @@ const StyledTextButton = styled(ButtonText)`
 export default function SwapBuyFiatButton() {
   const { account } = useWeb3React()
   const openFiatOnRampModal = useOpenModal(ApplicationModal.FIAT_ONRAMP)
-  const [, setBuyFiatFlowCompleted] = useBuyFiatFlowCompleted()
   const [checkFiatRegionAvailability, setCheckFiatRegionAvailability] = useState(false)
   const {
     available: fiatOnrampAvailable,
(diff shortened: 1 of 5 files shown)


src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -135,7 +128,6 @@ export default function SwapBuyFiatButton() {
       >
         <StyledTextButton onClick={handleBuyCrypto} disabled={buyCryptoButtonDisabled} data-testid="buy-fiat-button">
           <Trans>Buy</Trans>
-          {!buyFiatFlowCompleted && <Dot data-testid="buy-fiat-flow-incomplete-indicator" />}
```
vm, line 138: are you able to remove the hook now? useBuyDiatFlowCompleted?

unclear if it's used elsewhere

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -46,7 +39,7 @@ const StyledTextButton = styled(ButtonText)`
 export default function SwapBuyFiatButton() {
   const { account } = useWeb3React()
   const openFiatOnRampModal = useOpenModal(ApplicationModal.FIAT_ONRAMP)
-  const [buyFiatFlowCompleted, setBuyFiatFlowCompleted] = useBuyFiatFlowCompleted()
+  const [, setBuyFiatFlowCompleted] = useBuyFiatFlowCompleted()
```
tinaszheng, line 42: you can remove this too

tinaszheng (changes requested):

src/components/swap/SwapBuyFiatButton.tsx
LunrEclipse: Believe so, just checked and the hook isn't used anywhere else!

src/state/user/hooks.tsx
```diff
@@ -65,18 +64,6 @@ export function useUserLocaleManager(): [SupportedLocale | null, (newLocale: Sup
   return [locale, setLocale]
 }
 
-export function useBuyFiatFlowCompleted(): [boolean | undefined, (buyFiatFlowCompleted: boolean) => void] {
-  const dispatch = useAppDispatch()
-  const buyFiatFlowCompleted = useAppSelector((state) => state.user.buyFiatFlowCompleted)
-  const setBuyFiatFlowCompleted = useCallback(
-    (buyFiatFlowCompleted: boolean) => {
-      dispatch(updateUserBuyFiatFlowCompleted(buyFiatFlowCompleted))
```
vm, line 73: you can delete these actions now also

LunrEclipse: which actions?

tinaszheng (changes requested):
so looks like you also want to remove the field `buyFiatFlowCompleted` from the user store, which is persisted through localStorage

make sure you also write a migration [here](https://github.com/Uniswap/interface/blob/main/src/state/user/reducer.ts#L172) to delete the property if it exists

vm (approved):

tinaszheng (approved):
🙌 
image

PR #6814, opened 2023-06-21
build: simplify workflows

zzmp:
## Description

Simplifies workflows:
- Removes overly verbose naming
- Removes `cypress-tests`, which is replaced by `cypress:e2e`
- Removes `yarn prepare`, which duplicates `./.github/actions/setup`


JFrankfurt (approved):

.github/workflows/2-deploy-to-staging.yml
```diff
@@ -10,8 +10,7 @@ jobs:
     environment:
       name: deploy/staging
     steps:
-      - name: Send Slack message that deploy is starting
```
cbachmeier, line 13: QQ: What's the rationale for deleting the naming? Seems helpful if a new eng needs to pick up any future workflow work

.github/workflows/notify-slack-on-release-merge.yml
```diff
@@ -1,4 +1,4 @@
-name: Slack notifications for releases/* merges
+name: Slack on pushes to releases/*
```
cbachmeier, line 1: nit: I actually think `notification` is a helpful clarification here

.github/workflows/test.yml
```diff
@@ -10,8 +10,6 @@ on:
     branches:
       - main
   pull_request:
-  # manual trigger
-  workflow_dispatch:
```
cbachmeier, line 14: qq: will this prevent our ability to manually re-run failed tests?

.github/workflows/test.yml
```diff
@@ -175,12 +173,3 @@ jobs:
         with:
           name: Cypress tests
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_TEST_REPORTER_WEBHOOK }}
-
-  # Included as a single job to check for cypress-test-matrix success, as a matrix cannot be checked.
-  cypress-tests:
```
cbachmeier, line 180: qq: Could we still enforce all cypress tests passing before a PR is merged without this check?

cbachmeier (approved):
All nits or qs but I personally find the step naming helpful especially as someone who doesn't work with the actions frequently

.github/workflows/test.yml
zzmp: no, it will just prevent running tests on arbitrary commits outside of PRs, as through https://github.com/Uniswap/interface/actions/workflows/test.yml.

.github/workflows/test.yml
zzmp: yes, we are now enforcing this through `cypress:e2e`, instead of through `cypress-tests`.

cbachmeier (approved):

PR #6824, opened 2023-06-22
build: get/set test status per commit

zzmp:
## Description

Annotates the commit on which a check is run, so that future workflows can refer to that commit's test status.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2365/write-a-test-to-maintain-coverage-of-downloadingmounting-translations
_Relevant docs:_


.github/workflows/test.yml
```diff
@@ -173,3 +173,44 @@ jobs:
...
+              repo: context.repo.repo,
+              sha: '${{ github.event.pull_request.head.sha || github.event.head_commit.sha }}',
+              state: 'pending',
+              context: 'Test',
+              description: 'Running tests…',
+              target_url: 'https://github.com/Uniswap/interface/actions/runs/${{ github.run_id }}',
+            })
+
+  post:
+    if: always()
```
natew, line 194: whats up with this pattern? if we can add notes in yaml maybe worth noting why

natew (approved):

vm (approved):

PR #6825, opened 2023-06-22
build: improve workflow caching

zzmp:
## Description

Improve check times for PRs by optimizing caching:
- Only cache based on yarn.lock, instead of descending into children lockfiles. Resolving the glob took ~10s.
- Only save the cache when running on `main`. Saving the cache takes ~10-20s, and all feature branches fork from `main`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,8 +22,8 @@ jobs:
       - uses: ./.github/actions/cache-on-main
         with:
           path: node_modules/.cache
-          key: ${{ runner.os }}-eslint-${{ hashFiles('**/yarn.lock') }}-${{ github.run_id }}
-          restore-keys: ${{ runner.os }}-eslint-${{ hashFiles('**/yarn.lock') }}-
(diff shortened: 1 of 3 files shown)


JFrankfurt (approved):

.github/workflows/test.yml
```diff
@@ -19,12 +19,11 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup
-      - uses: actions/cache@v3
-        id: eslint-cache
+      - uses: ./.github/actions/cache-on-main
         with:
           path: node_modules/.cache
-          key: ${{ runner.os }}-eslint-${{ hashFiles('**/yarn.lock') }}-${{ github.run_id }}
-          restore-keys: ${{ runner.os }}-eslint-${{ hashFiles('**/yarn.lock') }}-
```
natew, line 27: do we need eslint hashed here

natew (approved):

.github/workflows/test.yml
zzmp: eslint needs to be in the name of the key, as this is an eslint specific hash

PR #6827, opened 2023-06-22
test(e2e): use electron

zzmp:
## Description

Uses electron to run e2e cypress tests.
electron is bundled with cypress, so this avoids us having to use a non-default container, and saves ~30s in cypress test setup.


just-toby (approved):

natew (approved):
pending cypress passing lgtm

PR #6830, opened 2023-06-22
feat: redux migration

just-toby:
## Description

introduces a version management system for the redux state - tests for the type of the App State, migrations which run when versions don't match, and an initial migration to move away from the old library.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2224/add-redux-migrations-to-the-interface
_Relevant docs:_ https://www.notion.so/Redux-migrations-in-interface-30aca899626a4ac4af7b070f92ce976d

### Before Migration
image

### After Migration (updated 8/14 to show IndexedDB)

image

image

### QA (ie manual testing)

- [x] manually tested the initial migration away from `redux-localstorage-simple` and ensured my old state was copied over

### Automated testing

- [x] Unit tests - added


src/state/migrations.test.ts
```diff
@@ -0,0 +1,154 @@
...
+  })
+
+  it('clears legacy redux_localstorage_simple values during the initial migration', async () => {
+    localStorage.setItem('redux_localstorage_simple_transactions', JSON.stringify({ test: 'transactions' }))
+    localStorage.setItem('redux_localstorage_simple_user', JSON.stringify({ test: 'user' }))
+    localStorage.setItem('redux_localstorage_simple_lists', JSON.stringify({ test: 'lists' }))
+
+    persistStore(store)
+    // wait for the migration to complete
+    await new Promise((resolve) => setTimeout(resolve, 0))
```
natew, line 106: they don't give us an await persistStore i guess

natew (approved):
overall looks good

JFrankfurt (approved):
looks clean and I like the tests :)

src/state/migrations.test.ts
```diff
@@ -0,0 +1,154 @@
...
+        lists: initialListsState,
+        _persist: { version: -1 },
+      })
+    )
+
+    persistStore(store)
+    // wait for the migration to complete
+    await new Promise((resolve) => setTimeout(resolve, 0))
+
+    const state = store.getState()
```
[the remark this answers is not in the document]
just-toby: yeah, i spent alot of time trying to verify the persisted data but i couldn't get localStorage working in jest . i need to mock it somehow because redux-persist doesn't seem to write to jest's localStorage by default.

in the meantime, i've verified the persisted data with manual testing (see screenshots in PR description)

src/state/lists/reducer.ts
```diff
@@ -111,30 +110,4 @@ export default createReducer(initialState, (builder) =>
         pendingUpdate: null,
       }
     })
-    .addCase(updateVersion, (state) => {
```
just-toby, line 114: @vm is this code intended to run every session? (that's my understanding of how `updateVersion` works)

i want to move it to the initial migration to version 0 of the new system - see `migrations.ts` - and i'm wondering if lists will break if it doesn't run on every session.

src/state/migrations.test.ts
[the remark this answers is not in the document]
zzmp: jest runs in node, so you'd need a mock regardless. you can use a polyfill if you can find one on npm, if that would increase your confidence.

just-toby:
this is pretty much ready to go, but we probably want to wait for #6923 to be ready and merge these together

src/state/index.ts
```diff
@@ -1,35 +1,40 @@
...
+    middleware: (getDefaultMiddleware) =>
+      getDefaultMiddleware({
+        thunk: true,
+        serializableCheck: {
+          // meta.arg and meta.baseQueryMeta are defaults. payload.trade is a nonserializable return value, but that's ok
+          // because we are not adding it into any persisted store that requires serialization (e.g. localStorage)
+          ignoredActionPaths: ['meta.arg', 'meta.baseQueryMeta', 'payload.trade'],
+          ignoredPaths: [routingApi.reducerPath],
+          ignoredActions: [
+            // ignore the redux-persist actions
```
zzmp, line 23: Why are we ignoring these actions?

cypress/support/commands.ts
```diff
@@ -54,13 +54,16 @@ Cypress.Commands.overwrite(
           onBeforeLoad(win) {
             options?.onBeforeLoad?.(win)
 
-            // We want to test from a clean state, so we clear the local storage (which clears redux).
-            win.localStorage.clear()
+            // We want to test from a clean state, so we clear the persisted state.
+            win.indexedDB.deleteDatabase('redux')
```
zzmp, line 58: This may be done automatically by cypress - you'd need to check the docs.

src/state/reducer.ts
```diff
@@ -13,18 +17,49 @@ import transactions from './transactions/reducer'
...
-  application,
+const persistedReducers = {
   user,
   transactions,
   signatures,
+  lists,
+}
+
+const appReducer = combineReducers({
+  ...persistedReducers,
```
zzmp, line 28: This is error-prone because a duplicated property will overwrite it. Could you instead move `...persistedReducers` to the _bottom_ of the object definition?

src/state/reducer.ts
```diff
@@ -13,18 +17,49 @@ import transactions from './transactions/reducer'
...
-  lists,
   logs,
   [routingApi.reducerPath]: routingApi.reducer,
+})
+
+export type AppState = ReturnType<typeof appReducer>
+
+function createStorage() {
+  const db = localForage.createInstance({
+    name: 'redux',
```
zzmp, line 44: Can this be more specific, in case some other web app we publish on the same domain (uniswap.org) ends up also using redux?

```suggestion
    name: 'redux-interface',
```

src/state/reducer.ts
```diff
@@ -13,18 +17,49 @@ import transactions from './transactions/reducer'
...
+  })
+  return {
+    getItem: db.getItem,
+    setItem: db.setItem,
+    removeItem: db.removeItem,
+  }
```
zzmp, line 50: Why do you need to break this out?
```suggestion
  return db
```

src/state/legacyMigrations.ts
```diff
@@ -0,0 +1,88 @@
+import { DEFAULT_DEADLINE_FROM_NOW } from 'constants/misc'
```
zzmp, line 1: Can you rename this file `src/state/migrations/legacy.ts`?

src/state/legacyMigrations.ts
```diff
@@ -0,0 +1,88 @@
...
+
```
zzmp, line 2: What happened to the tests for these migrations?

src/state/migrations.test.ts
```diff
@@ -0,0 +1,134 @@
...
+  it('clears legacy redux_localstorage_simple values during the initial migration', async () => {
+    localStorage.setItem(
+      'redux_localstorage_simple_transactions',
+      JSON.stringify({ 1: { test: { info: 'transactions' } } })
+    )
+    localStorage.setItem('redux_localstorage_simple_user', JSON.stringify({ test: 'user' }))
+    localStorage.setItem('redux_localstorage_simple_lists', JSON.stringify({ test: 'lists' }))
+    localStorage.setItem('redux_localstorage_simple_signatures', JSON.stringify({ test: 'signatures' }))
+
+    persistStore(store)
```
zzmp, line 77: Instead of the arbitrary await below this, capture the `const persistor = persistStore(store)` object and use that API.

src/state/migrations.ts
```diff
@@ -0,0 +1,69 @@
...
+  const defaultMigrate = createMigrate(migrations, options)
+
+  return (state: PersistedState, currentVersion: number) => {
+    if (state === undefined) {
+      // If no state exists, run the migration for version 1
+      return Promise.resolve(migrations[0](undefined))
+    }
+
+    // Otherwise, use the default migration process
+    return defaultMigrate(state, currentVersion)
```
zzmp, line 67: This will run the `0` migration, so the legacy migration should not be included in `0`. `0` should be a no-op as an example for future migrations.

src/state/migrations.ts
```diff
@@ -0,0 +1,69 @@
...
+// We use a custom migration function for the initial state, because redux-persist
+// skips migration if there is no initial state in localStorage, but we want to migrate
+// previous persisted state from redux-localstorage-simple.
+export function customCreateMigrate(migrations: MigrationManifest, options: MigrationConfig): PersistMigrate {
+  const defaultMigrate = createMigrate(migrations, options)
+
+  return (state: PersistedState, currentVersion: number) => {
+    if (state === undefined) {
+      // If no state exists, run the migration for version 1
+      return Promise.resolve(migrations[0](undefined))
```
zzmp, line 63: This should run the legacy migration, _followed by_ `defaultMigrate`, so it should not return early.

src/state/migrations.ts
```diff
@@ -0,0 +1,69 @@
...
+ * These run once per state re-hydration when a version mismatch is detected.
+ * Keep them as lightweight as possible.
+ *
+ * Migration functions should not assume that any value exists in localStorage previously,
+ * because a user may be visiting the site for the first time or have cleared their localStorage.
+ */
+
+// The target version number is the key
+export const migrations: MigrationManifest = {
+  0: () => {
```
zzmp, line 27: This is a legacy migration, not a v0 migration, because it checks localStorage which need only be done when migrating. Can you not put it in the `MigrationManifest`, and instead leave it outside as `legacy`?

src/state/migrations.ts
```diff
@@ -0,0 +1,69 @@
...
+    const oldTransactions = localStorage.getItem('redux_localstorage_simple_transactions')
```
zzmp, line 28: Can you move the legacy migration code into `src/state/migrations/legacy.ts`.
Can you move the `0` (a no-op) migration into `src/state/migrations/0.ts`.
This will prevent this file from growing over time, while still keeping migrations organized.

src/state/migrations.ts
```diff
@@ -0,0 +1,69 @@
...
+      transactions: legacyTransactionMigrations(newTransactions),
+      lists: newLists,
+      signatures: newSignatures,
+      _persist: { version: 0, rehydrated: true },
+    }
+
+    localStorage.removeItem('redux_localstorage_simple_transactions')
+    localStorage.removeItem('redux_localstorage_simple_user')
+    localStorage.removeItem('redux_localstorage_simple_lists')
+    localStorage.removeItem('redux_localstorage_simple_signatures')
```
zzmp, line 49: I think you should predicate removing these items until after the state is flushed to disk, or you risk a user opening the site, quickly closing it, and losing all of their state.

zzmp (changes requested):
I think you need a `PersistGate` (see https://www.npmjs.com/package/redux-persist) or something similar to make sure the state is loaded before the first render.

You may not want a `PersistGate`, to avoid a slow pageload experience, but it doesn't seem to pick up my previously connected wallet when I reload the page.

src/state/reducer.ts
zzmp: Alternatively (and possibly cleaner), continue to use `redux` as the db name, but change the `key` from `'root'` to `'interface'`.

zzmp:
Can you also add an e2e test that ensures that saved state is being reflected in pageload?
This can be as small as setting selected wallet to INJECTED and ensuring that there _is_ a connected wallet on pageload. This test passes for you right now because you are setting it in localStorage, but the main codepath here will have it set in indexedDB, and I believe that would make the e2e test fail as-written.

cypress/support/commands.ts
just-toby: looks like it doesn't: https://github.com/cypress-io/cypress/issues/1208

src/state/index.ts
just-toby: https://redux-toolkit.js.org/usage/usage-guide#use-with-redux-persist

this is the instructions for how to integrate redux-persist with redux toolkit

cypress/support/commands.ts
```diff
@@ -54,10 +54,10 @@ Cypress.Commands.overwrite(
...
 
-            // Set initial user state.
+            // Since the IndexedDB state is clear, the first migration will run and pick up this legacy state from localStorage.
             win.localStorage.setItem(
               'redux_localstorage_simple_user', // storage key for the user reducer using 'redux-localstorage-simple'
               JSON.stringify({
```
zzmp, line 63: This should be setting indexedDB state, not localStorage state, as it currently acts as a migration test for the legacy state instead of an e2e test for the current state.

src/index.tsx
```diff
@@ -57,25 +58,27 @@ const Router = isBrowserRouterEnabled() ? BrowserRouter : HashRouter
...
-                      <App />
-                    </ThemeProvider>
-                  </BlockNumberProvider>
-                </ApolloProvider>
-              </Web3Provider>
-            </LanguageProvider>
-          </Router>
-        </QueryClientProvider>
-      </FeatureFlagsProvider>
+      <PersistGate loading={null} persistor={persistor}>
```
zzmp, line 61: This delays the first render until the state has been read, which is not what we want, because then the wallet connection attempt occurs, and overall it just adds jank to the initial render.

Would redux-persist still work without PersistGate? It's ok to render initially with empty state, so long as it re-renders after first load.

(Sorry - I know I initially suggested this, but it adds a lot of time to the initial render so I don't think it can go in.)

src/state/index.ts
```diff
@@ -23,18 +23,19 @@ export function createDefaultStore() {
             // ignore the redux-persist actions
             'persist/PERSIST',
             'persist/REHYDRATE',
+            'persist/PURGE',
+            'persist/FLUSH',
```
zzmp, line 27: Can these be globby, eg `'persist/*'`?

src/state/migrations.ts
```diff
@@ -24,43 +14,20 @@ function tryParseOldState<T>(value: string | null, fallback: T): T {
...
   const defaultMigrate = createMigrate(migrations, options)
 
-  return (state: PersistedState, currentVersion: number) => {
+  return async (state: PersistedState, currentVersion: number) => {
     if (state === undefined) {
-      // If no state exists, run the migration for version 1
-      return Promise.resolve(migrations[0](undefined))
+      // If no state exists, run the legacy migration to set initial state
+      const initialState = await legacyLocalStorageMigration()
+      return defaultMigrate(initialState, currentVersion)
```
zzmp, line 30: This can be simplified with a fallthrough:

```suggestion
      // If no state exists, run the legacy migration to set initial state
      state = await legacyLocalStorageMigration()
```

src/state/migrations/0.ts
```diff
@@ -0,0 +1,5 @@
+import { PersistedState } from 'redux-persist'
+
+export const migration0 = (state: PersistedState) => {
```
zzmp, line 3: doc-comment explaining why this exists (as a proof of concept?)

src/state/reducer.ts
```diff
@@ -35,29 +35,25 @@ const appReducer = combineReducers({
...
   version: 0, // see migrations.ts for more details about this version
-  storage: createStorage(),
+  storage: localForage.createInstance({
+    name: 'redux',
+  }),
   migrate: customCreateMigrate(migrations, { debug: false }),
   whitelist: Object.keys(persistedReducers),
   throttle: 1000, // ms
   serialize: false,
+  // The typescript definitions are wrong - we need this to be false for unserialized storage to work.
```
zzmp, line 53:
```suggestion
  // The typescript definitions are wrong - we need this to be false for unserialized storage to work, and we need unserialized storage for inspectable db entries for debugging.
```

cypress/support/commands.ts
```diff
@@ -59,10 +59,11 @@ Cypress.Commands.overwrite(
...
             win.localStorage.setItem(
-              'redux_localstorage_simple_user',
+              'redux_localstorage_simple_user', // storage key for the user reducer using 'redux-localstorage-simple'
               JSON.stringify({
+                hideUniswapWalletBanner: true,
                 ...CONNECTED_WALLET_USER_STATE,
-                ...options?.userState,
+                ...(options?.userState ?? {}),
               })
             )
```
zzmp, line 68: This should use an up-to-date entry, instead of using a legacy entry and expecting it to migrate. Otherwise it's not actually testing the most common codepath.

zzmp (changes requested):

cypress/support/commands.ts
```diff
@@ -54,18 +54,31 @@ Cypress.Commands.overwrite(
...
+              store.put(
+                {
+                  user: {
+                    ...initialState,
+                    hideUniswapWalletBanner: true,
+                    ...CONNECTED_WALLET_USER_STATE,
+                    ...(options?.userState ?? {}),
+                  },
+                },
+                'persist:interface'
```
zzmp, line 74: This should be pulled out as a key from `../../src/state/user/reducer` if it can be.

cypress/support/commands.ts
```diff
@@ -54,18 +54,31 @@ Cypress.Commands.overwrite(
...
+              )
+            }
+
+            dbRequest.onupgradeneeded = function () {
+              const db = dbRequest.result
+              db.createObjectStore('keyvaluepairs')
+            }
```
zzmp, line 81: Add a comment explaining what this does. Consider moving it to a helper function, something like

    resetIndexedDB(state: unknown): Promise<void>

I'm not sure if you need to wait for it to be set or not.

zzmp (approved):
awesome

zzmp (approved):

PR #6833, opened 2023-06-23
fix: chrome tie sort resolution

JFrankfurt:
## Description

Chrome handles ties differently than other browsers which caused the chain sorting to appear reversed. This replaces the sort with two arrays to solve the issue, and in anticipation of labeling unsupported networks as a set.

### Before
Chrome
Screen Shot 2023-06-23 at 10 44 33

### After
Chrome
Screen Shot 2023-06-23 at 10 44 50

### Reproducing the error

1. Go to https://interface-git-main-uniswap.vercel.app/#/swap on Chrome and open Chain Switcher


cbachmeier (approved):
I prefer this alternative

cbachmeier:
nice description

just-toby (approved):

PR #6776, opened 2023-06-15
feat: Add Avalanche support

cbachmeier:
## Description

_Linear ticket:_
_Slack thread:_
_Relevant docs:_

### Before
|    Mobile    |   Desktop    |
| ------------ | ------------ |
| paste_before | paste_before |

### After
|    Mobile    |   Desktop   |
| ------------ | ----------- |
| paste_after  | paste_after |

### Reproducing the error

1. 

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test

PR #6757, opened 2023-06-13
feat: upgrade sdk-core to 3.2.6 and add AVAX

cbachmeier:
## Description

- Updates sdk-core to 3.2.6
- `src/constants/addresses.ts` has been deleted and x-chain addresses will instead be pulled from `sdk-core`
- The `SupportedChainId` in `src/constants/chains.ts` has been replaced with `ChainId` from `sdk-core`
  - There is a slight functional difference as `ChainId` also includes a few chains the BE partially supports but the UX and Protocol do not
  - To handle this difference there is an array of fully supported `ChainId`s in `skd-core` called `SUPPORTED_CHAINS` and a type `SupportedChainsType` representing them
  - in `src/constants/chains.ts` two helper fns were moved from `src/lib/hooks/routing/clientSideSmartOrderRouter.ts`: `isSupportedChain` and `asSupportedChain`
    - **WHERE THESE NEW FUNCTIONS ARE USED IS THE MAIN LOGIC TO HELP VERIFY**
      - For example a chain partially supported by BE is [Gnosis](https://chainlist.org/chain/100). You should still get the unsupported chain behaviour when connected to a wallet on this chain.
      - You should also get the normal supported chain experience on all the uxs listed in the switcher, some screenshots of my testing in Polygon are below.
      - All other changes are 1:1 such as `SupportedChainId` from `constants/chains` -> `ChainId` from `sdk-core`
      - See the full list of `ChainIds` and `SUPPORTED_CHAINS` in [sdk-core](https://github.com/Uniswap/sdk-core/blob/main/src/chains.ts#L1)
      - Files I would specifically call out would be:
        - [src/components/AccountDrawer/MiniPortfolio/Pools/hooks.ts](https://github.com/Uniswap/interface/pull/6757/files#diff-0f1a20741499f6adddd2843621a4f9bc4ee86146e7923ab84c89d25262b0d297)
        - [src/components/WalletModal/index.tsx](https://github.com/Uniswap/interface/pull/6757/files#diff-8e2184422130a605f9acef279acdeb7fa8b14a3832f04bb0abf451403bdc50ac)
        - [src/constants/chains.ts](https://github.com/Uniswap/interface/pull/6757/files#diff-397f5cd42e5c043e670c0b1ccacb46c9c824d534c1077681707ef9fe260aee18)
          - where the helper fns are defined
[...]

PR #6834, opened 2023-06-23
fix: handle BE avax support

cbachmeier:
## Description

BE has added AVAX as a ChainId and now shows avax tokens in the miniportfolio. We'll need an interim state before #6776 is fully merged where this no longer causes builds to fail. **Avalanche swapping, LPing, wrapping, etc will still not be supported as a result of this PR.** This is the minimal amount of content needed to prevent the build from crashing. Yes this will cause merge conflicts with my branch which I will have to resolve. Thank you for asking.

_Linear ticket:_
_Slack thread:_
_Relevant docs:_

## Screen capture
AVAX still not supported
Screen Shot 2023-06-23 at 10 52 45
Screen Shot 2023-06-23 at 10 52 51

### DIFF
--- a/src/constants/chains.test.ts
+++ b/src/constants/chains.test.ts
@@ -2,6 +2,7 @@ import { SupportedChainId as SdkSupportedChainId } from '@uniswap/sdk-core'
 
 import { ALL_SUPPORTED_CHAIN_IDS, SupportedChainId } from './chains'
 
+// These tests will be deprecated as part of the moving of ChainId to sdk-core
 describe('ChainIds', () => {
   describe('SupportedChainId', () => {
     it('derives from sdk-core', () => {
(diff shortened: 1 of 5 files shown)


src/constants/chains.test.ts
```diff
@@ -2,14 +2,15 @@ import { SupportedChainId as SdkSupportedChainId } from '@uniswap/sdk-core'
 
 import { ALL_SUPPORTED_CHAIN_IDS, SupportedChainId } from './chains'
 
+// These tests will be deprecated as part of the moving of ChainId to sdk-core
```
just-toby, line 5: make this a todo?

just-toby (approved):

src/constants/chains.test.ts
cbachmeier: Already handled in open PR https://github.com/Uniswap/interface/pull/6757

zzmp (approved):
Just confirming - this will have no UX change visible to the user?

tinaszheng (changes requested):
dont think this is the right way to do support for gql changes - adding AVAX to SupportedChainId means that there are certain parts of the app that will try to render and pass in AVAX to Chain lookups - if our current version of universal-router-sdk doesn't support it yet, for example, this will cause things to crash

from your vercel link, using avax:

image

cbachmeier:
I'm in agreement with @tinaszheng as this was also not future proof, I've opened https://github.com/Uniswap/interface/pull/6836 instead which is!

PR #6835, opened 2023-06-23
fix: loosen chain types

zzmp:
## Description

Loosen chain types until AVAX support is added.

### DIFF
--- a/src/graphql/data/util.tsx
+++ b/src/graphql/data/util.tsx
@@ -120,7 +120,7 @@ export function validateUrlChainParam(chainName: string | undefined) {
 }
 
 // TODO(cartcrom): refactor into safer lookup & replace usage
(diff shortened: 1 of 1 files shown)


tinaszheng (approved):

src/graphql/data/util.tsx
```diff
@@ -133,7 +133,9 @@ export const CHAIN_NAME_TO_CHAIN_ID: { [key in Chain]: SupportedChainId } = {
 }
 
 export function fromGraphQLChain(chain: Chain): SupportedChainId {
-  return CHAIN_NAME_TO_CHAIN_ID[chain]
+  const chainId = CHAIN_NAME_TO_CHAIN_ID[chain]
+  if (!chainId) throw new Error(`Unknown chain: ${chain}`)
```
cbachmeier, line 137: we don't want to error in this case as on unsupported chains we show Mainnet data

PR #6836, opened 2023-06-23
fix: alternate solution to future proof chain ads

cbachmeier:
## Description

Alternate solution to #6834 
This will actually future proof future GQL chain additions without causing our builds to crash. It does make the return type optional so I've add some null checks and updated the helper fn with a default to mainnet which has been our standard UX behaviour prior. 

This will cause even worse merge conflicts for my own AVAX PR 🥲

_Linear ticket:_
_Slack thread:_
_Relevant docs:_

## Screen capture

No crashes on Avalanche
Screen Shot 2023-06-23 at 12 18 41
Celo Tokens
Screen Shot 2023-06-26 at 12 40 00
No Avax tokens or activity
Screen Shot 2023-06-26 at 12 40 56
Screen Shot 2023-06-26 at 12 41 06


src/graphql/data/util.tsx
```diff
@@ -120,7 +120,7 @@ export function validateUrlChainParam(chainName: string | undefined) {
 }
 
 // TODO(cartcrom): refactor into safer lookup & replace usage
-export const CHAIN_NAME_TO_CHAIN_ID: { [key in Chain]: SupportedChainId } = {
+export const CHAIN_NAME_TO_CHAIN_ID: { [key in Chain]?: SupportedChainId } = {
```
cbachmeier, line 123: with the addition of `?` the keys don't have to be exhaustive which makes this future proof for more chain additions. the cost is that the return type is now optional

src/graphql/data/util.tsx
```diff
@@ -133,7 +133,7 @@ export const CHAIN_NAME_TO_CHAIN_ID: { [key in Chain]: SupportedChainId } = {
 }
 
 export function fromGraphQLChain(chain: Chain): SupportedChainId {
-  return CHAIN_NAME_TO_CHAIN_ID[chain]
+  return CHAIN_NAME_TO_CHAIN_ID[chain] ?? SupportedChainId.MAINNET
```
tinaszheng, line 136: hmm the backend here is still gonna return avax tokens - theyre gonna show up as mainnet instead?

tinaszheng (changes requested):

src/graphql/data/util.tsx
cbachmeier: the avax tokens won't show up until we modify our queries to add them, at which point we can add them to `CHAIN_NAME_TO_CHAIN_ID` as well

cbachmeier: defaulting to main is the current behaviour for our gql chain behaviour including portfolio https://github.com/Uniswap/interface/blob/main/src/components/AccountDrawer/MiniPortfolio/PortfolioLogo.tsx#L88

tinaszheng (commented):
is there a solution in which we have something like a `GQL_SUPPORTED_CHAIN_IDS` list be a list of SupportedChainId, pass that as a variable to all of our graphql calls (so no hard-coded variables that need to be updated at the query level), and when we parse the return data from the gql queries, we filter out tokens/entities not belonging in `GQL_SUPPORTED_CHAIN_IDS`? 

we'd be able to remove using `Chain` as a type in non query call-sites (and use `SupportedChainId` everywhere instead), and also couple the logic for what gql chains we should query vs filter out in one place

cbachmeier:
@tinaszheng I've refactored our portfolio query to use the `GQL_SUPPORTED_CHAINS` already present in the `util` file. This has the added benefit of showing Celo tokens in the portfolio! This is the only query where we currently support passing in a chains filter so as discussed I added a UX filter to only show support chain activity in the miniport

src/graphql/data/util.tsx
```diff
@@ -133,10 +133,17 @@ export const CHAIN_NAME_TO_CHAIN_ID: { [key in Chain]: SupportedChainId } = {
...
 }
 
-export const BACKEND_CHAIN_NAMES: Chain[] = [Chain.Ethereum, Chain.Polygon, Chain.Optimism, Chain.Arbitrum, Chain.Celo]
+export const GQL_SUPPORTED_CHAINS: Chain[] = [
```
tinaszheng, line 139: so my comment was more trying to make this `SupportedChainId[]` (or a separate `SupportedGqlChain` type that is just the gql name, instead of using the GraphQL type `Chain`, that way we can filter the data returned by gql and then they can be either SupportedChainId or SupportedGqlChain and you wouldn't have to loosen the type for `CHAIN_NAME_TO_CHAIN_ID`

lmk this makes sense - happy to draft a PR to show an example

cbachmeier: Are you thinking something like?:
```
const UX_SUPPORTED_CHAINS = [
  Chain.Ethereum,
  Chain.EthereumGoerli,
  Chain.EthereumSepolia,
  Chain.Polygon,
  Chain.Celo,
  Chain.Optimism,
  Chain.Arbitrum,
  Chain.UnknownChain,
  Chain.Bnb,
] as const
type UxSupportedChainsType = typeof UX_SUPPORTED_CHAINS[number]

export const CHAIN_NAME_TO_CHAIN_ID: { [key in UxSupportedChainsType]: SupportedChainId } = { ... }

cbachmeier: Otherwise would appreciate a snippet to know more of what you're thinking

cbachmeier:
Pursuing the solution in #6878

PR #6837, opened 2023-06-23
fix: divide by zero error on pool page

JackShort:
* divide by zero error was being caused by inverting the price when the quote currency was zero / did not exist it seems like

### DIFF
--- a/src/pages/AddLiquidityV2/PoolPriceBar.tsx
+++ b/src/pages/AddLiquidityV2/PoolPriceBar.tsx
@@ -21,7 +21,9 @@ export function PoolPriceBar({
   price?: Price<Currency, Currency>
 }) {
   const theme = useTheme()
-  const canInvertPrice = Boolean(price?.baseCurrency && price.quoteCurrency && !price.baseCurrency.equals(price.quoteCurrency))
+  const canInvertPrice = Boolean(
+    price && price.baseCurrency && price.quoteCurrency && !price.baseCurrency.equals(price.quoteCurrency)
+  )
   const invertedPrice = canInvertPrice ? price?.invert()?.toSignificant(6) : undefined
 
   return (
--- a/src/pages/AddLiquidityV2/PoolPriceBar.tsx
+++ b/src/pages/AddLiquidityV2/PoolPriceBar.tsx
@@ -21,6 +21,9 @@ export function PoolPriceBar({
   price?: Price<Currency, Currency>
 }) {
   const theme = useTheme()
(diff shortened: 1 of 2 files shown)


src/pages/AddLiquidityV2/PoolPriceBar.tsx
```diff
@@ -21,6 +21,9 @@ export function PoolPriceBar({
   price?: Price<Currency, Currency>
 }) {
   const theme = useTheme()
+  const canInvertPrice = Boolean(price?.baseCurrency && price.quoteCurrency && !price.baseCurrency.equals(price.quoteCurrency))
```
vm, line 24: some of these price accessors have ? and some don't. do we need them?

src/pages/MigrateV2/MigrateV2Pair.tsx
```diff
@@ -46,7 +46,7 @@ import useIsArgentWallet from '../../hooks/useIsArgentWallet'
 import { useTotalSupply } from '../../hooks/useTotalSupply'
 import { useTokenBalance } from '../../state/connection/hooks'
 import { TransactionType } from '../../state/transactions/types'
-import { BackArrowLink, ExternalLink, ThemedText } from '../../theme'
+import { BackArrow, ExternalLink, ThemedText } from '../../theme'
```
JackShort, line 49: ignore this right now trying to remove these changes

src/pages/AddLiquidityV2/PoolPriceBar.tsx
JackShort: the first one makes sense because if there is no price it will return false and won't evaluate the next condition

src/pages/MigrateV2/MigrateV2Pair.tsx
vm, line 49: is this part of the change relevant?

JackShort: lmao see my comment

src/pages/AddLiquidityV2/PoolPriceBar.tsx
vm: i think it'd be better as `price && price.baseCurrency && price.quoteCurrency && !price.baseCurrency.equals(price.quoteCurrency)`

JackShort: done

vm (approved):

just-toby (approved):

PR #6839, opened 2023-06-24
fix: filter more CSP errors

vm:
## Description
Removes more CSP errors that we shouldn't be alerted on.

https://uniswap-labs.sentry.io/issues/4251643566/?project=4504255148851200&query=is%3Aunresolved&referrer=issue-stream&sort=freq&statsPeriod=24h&stream_index=4

https://uniswap-labs.sentry.io/issues/3789706876/?project=4504255148851200&query=is%3Aunresolved&referrer=issue-stream&sort=freq&statsPeriod=24h&stream_index=21

### Automated testing
- [x] Unit test
- [ ] Integration/E2E test

### DIFF
--- a/src/tracing/errors.ts
+++ b/src/tracing/errors.ts
@@ -69,9 +69,9 @@ function shouldRejectError(error: EventHint['originalException']) {
     // Content security policy 'unsafe-eval' errors can be filtered out because there are expected failures.
     // For example, if a user runs an eval statement in console this error would still get thrown.
     // TODO(WEB-2348): We should extend this to filter out any type of CSP error.
-    if (error.message.match(/'unsafe-eval'.*content security policy/i)) {
-      return true
-    }
+    if (error.message.match(/'unsafe-eval'.*content security policy/i)) return true
+    if (error.message.match(/Blocked a frame with origin "*" from accessing a cross-origin frame./)) return true
+    if (error.message.match(/NotAllowedError: Write permission denied./)) return true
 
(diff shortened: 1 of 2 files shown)


zzmp (approved):

src/tracing/errors.ts
```diff
@@ -69,9 +69,9 @@ function shouldRejectError(error: EventHint['originalException']) {
     // Content security policy 'unsafe-eval' errors can be filtered out because there are expected failures.
     // For example, if a user runs an eval statement in console this error would still get thrown.
     // TODO(WEB-2348): We should extend this to filter out any type of CSP error.
-    if (error.message.match(/'unsafe-eval'.*content security policy/i)) {
-      return true
-    }
+    if (error.message.match(/'unsafe-eval'.*content security policy/i)) return true
+    if (error.message.match(/Blocked a frame with origin ".*" from accessing a cross-origin frame./)) return true
+    if (error.message.match(/NotAllowedError: Write permission denied./)) return true
```
cbachmeier, line 74: Can we create a global array of error messages to filter out and then just check if current message is present?

cbachmeier (changes requested):

src/tracing/errors.ts
vm: Talked in DM but think we should handle that in a separate PR, I can make a task for it.

cbachmeier (approved):

PR #6840, opened 2023-06-24
fix: filter user reject errors temporarily

vm:
## Description
I've changed my mind on this a few times. First I added this filter https://github.com/Uniswap/interface/pull/6330 because it was high volume. Then I removed the filter here https://github.com/Uniswap/interface/pull/6533 because I felt this was something we should be handling. I then ignored it in Sentry directly so we didn't get alerted.

I do have a WIP to fix this here but haven't had time to finish it. https://github.com/Uniswap/interface/pull/6348

The problem is we now have an alert which checks for crash free session rate, and unfortunately ignored errors still count against that metric. So I'd like to temporarily pause sending these until we merge the other PR.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2398/catch-user-rejection-errors

### Automated testing
- [x] Unit test
- [ ] Integration/E2E test

### DIFF
--- a/src/tracing/errors.test.ts
+++ b/src/tracing/errors.test.ts
@@ -54,9 +54,9 @@ describe('beforeSend', () => {
     expect(beforeSend(ERROR, { originalException })).toBe(ERROR)
   })
 
-  it('propagates user rejected request errors', () => {
+  it('filters user rejected request errors', () => {
     const originalException = new Error('user rejected transaction')
-    expect(beforeSend(ERROR, { originalException })).toBe(ERROR)
+    expect(beforeSend(ERROR, { originalException })).toBeNull()
   })
 
   it('filters block number polling errors', () => {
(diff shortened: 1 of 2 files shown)


zzmp (approved):

src/tracing/errors.test.ts
```diff
@@ -54,9 +54,9 @@ describe('beforeSend', () => {
     expect(beforeSend(ERROR, { originalException })).toBe(ERROR)
   })
 
-  it('propagates user rejected request errors', () => {
+  it('filters user rejected request errors', () => {
```
cbachmeier, line 57: nit: add the TODO here as well so implementer also adds test to support graceful handling

vm: the TODO will basically cause them to change the test right, since the test will fail?

PR #6845, opened 2023-06-26
feat: gate v1 retirement

JFrankfurt:
This gates between two experiences:

1. Uniswap Wallet on wcv1, WalletConnect on v1 with v2 option via ... menu
2. Uniswap Wallet on wcv2, WalletConnect on v2


src/connection/WalletConnectV2.ts
```diff
@@ -61,3 +61,37 @@ export class WalletConnectV2 extends WalletConnect {
     return super.activate(chainId)
   }
 }
+
+// Custom class for Uniswap Wallet specific functionality
+export class UniwalletConnect extends WalletConnectV2 {
```
JackShort, line 66: since this is called UniswapWalletV2 everywhere else should you just add v2 to the naming here

JackShort (approved):

src/connection/index.ts
```diff
@@ -107,6 +107,19 @@ export const uniwalletConnectConnection: Connection = {
   isNew: true,
```
zzmp, line 107: Will `initializeConnector<UniwalletConnect>` fail to initialize once V1 shuts down? If this throws a synchronous error it will crash the application, I think. (If it is synchronous, I don't think it will matter.)

src/components/WalletModal/Option.tsx
```diff
@@ -183,21 +180,21 @@ export default function Option({ connection }: OptionProps) {
...
+  const handleClickOpenPopover = (e: MouseEvent<HTMLButtonElement>) => {
     e.stopPropagation()
-    setWC1PromptOpen(true)
+    setPopoverOpen(true)
   }
 
   const isWalletConnect =
     connection.type === ConnectionType.WALLET_CONNECT || connection.type === ConnectionType.WALLET_CONNECT_V2
-  const showExtraMenuToggle = isWalletConnect && !isCurrentOptionPending
+  const showExtraMenuToggle = isWalletConnect && !isCurrentOptionPending && !walletConnectV2AsDefault
```
zzmp, line 197: This is the crux of it, right? If `!walletConnectV2AsDefault` then we shouldn't show V1 as an option, even?

zzmp (changes requested):
Does this re-purpose the existing flag, which ramps v1->v2 as default, to instead eliminate v1 as an option once v2 is the default? Do we want to do that - will it be a poorer user experience if we need to rollback?

src/components/WalletModal/Option.tsx
vm: isn't that the case?

src/featureFlags/flags/walletConnectPopover.ts
```diff
@@ -0,0 +1,5 @@
+import { BaseVariant, FeatureFlag, useBaseFlag } from '../index'
+
+export function useWalletConnectFallback(): BaseVariant {
+  return useBaseFlag(FeatureFlag.walletConnectFallback)
+}
```
zzmp, line 5:
```suggestion
export function useWalletConnectFallbackFlag(): BaseVariant {
  return useBaseFlag(FeatureFlag.walletConnectFallback)
}

export function useWalletConnectFallback(): boolean {
  return useWalletConnectFallbackFlag() === BaseVariant.Enabled
}
```

src/components/WalletModal/Option.tsx
```diff
@@ -169,35 +172,41 @@ interface OptionProps {
 }
 export default function Option({ connection }: OptionProps) {
   const { activationState, tryActivation } = useActivationState()
-  const [WC1PromptOpen, setWC1PromptOpen] = useState(false)
+  const [WC2PromptOpen, setPopoverOpen] = useState(false)
```
zzmp, line 175:
```suggestion
  const [WCPopoverOpen, setWCPopoverOpen] = useState(false)
```

zzmp (changes requested):

vm (approved):

PR #6848, opened 2023-06-27
chore: removed an unused label property.

2-NOW:
## Description

summary:
Fixed an issue with `label="[object Object]"` in SwapCurrencyInputPanel.

details:
As label prop, `<Tran / >` jsx was passed as a styled div and the label of the div element of the deployed service was checked for the value `[object Object]`.

Removed the label prop because it was found to be an unused prop and the SwapCurrencyInputPanel component was also found to be not referenced on any page other than swap.

Modified both the Input and Output field components.

https://github.com/Uniswap/interface/blob/1c4a383a4906f058c3c255e89f4a93dee576d7a3/src/pages/Swap/index.tsx#L584-L587

https://github.com/Uniswap/interface/blob/1c4a383a4906f058c3c255e89f4a93dee576d7a3/src/components/CurrencyInputPanel/SwapCurrencyInputPanel.tsx#L245

https://github.com/Uniswap/interface/blob/1c4a383a4906f058c3c255e89f4a93dee576d7a3/src/components/CurrencyInputPanel/SwapCurrencyInputPanel.tsx#L28

### DIFF
--- a/src/pages/Swap/index.tsx
+++ b/src/pages/Swap/index.tsx
@@ -582,9 +582,6 @@ export function Swap({
         <SwapSection>
           <Trace section={InterfaceSectionName.CURRENCY_INPUT_PANEL}>
             <SwapCurrencyInputPanel
-              label={
-                independentField === Field.OUTPUT && !showWrap ? <Trans>From (at most)</Trans> : <Trans>From</Trans>
-              }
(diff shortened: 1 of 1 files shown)


gnewfield:
The `label` prop has been updated. Thank you for your contribution!

PR #6849, opened 2023-06-27
fix: combine userState with default injected wallet in cypress setup

just-toby:
## Description

this allows you to set custom `userState` in a test in combination with the default injected hardhat provider being automatically connected. 

### Reproducing the error

1.  write a test with both the `hardhat` option and some custom initial `userState` and note that the wallet isn't automatically connected at the start of the test

### QA (ie manual testing)

- [x] verified the wallet connects automatically with this change

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -58,11 +58,7 @@ Cypress.Commands.overwrite(
             // Set initial user state.
             win.localStorage.setItem(
               'redux_localstorage_simple_user', // storage key for the user reducer using 'redux-localstorage-simple'
-              JSON.stringify(
-                options?.userState
-                  ? { ...CONNECTED_WALLET_USER_STATE, ...options.userState }
-                  : CONNECTED_WALLET_USER_STATE
-              )
+              JSON.stringify({ ...CONNECTED_WALLET_USER_STATE, ...(options?.userState ?? {}) })
             )
 
             // Set feature flags, if configured.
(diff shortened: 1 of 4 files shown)


zzmp (changes requested):
Two tests rely on a disconnected state. I'm not sure why they are still passing with your changes:

https://github.com/Uniswap/interface/blob/main/cypress/e2e/landing.test.ts#L6
https://github.com/Uniswap/interface/blob/main/cypress/e2e/wallet-connection/connect.test.ts#L30

zzmp (approved):

cypress/support/commands.ts
```diff
@@ -58,7 +58,11 @@ Cypress.Commands.overwrite(
             // Set initial user state.
             win.localStorage.setItem(
               'redux_localstorage_simple_user', // storage key for the user reducer using 'redux-localstorage-simple'
-              JSON.stringify(options?.userState ?? CONNECTED_WALLET_USER_STATE)
+              JSON.stringify(
+                options?.userState
```
vm, line 62: could do `{ ...CONNECTED_WALLET_USER_STATE, ...(options?.userState ?? {}) }`

cypress/utils/user-state.ts
```diff
@@ -1,3 +1,5 @@
 import { UserState } from '../../src/state/user/reducer'
 
 export const CONNECTED_WALLET_USER_STATE: Partial<UserState> = { selectedWallet: 'INJECTED' }
+
+export const UNCONNECTED_WALLET_USER_STATE: Partial<UserState> = { selectedWallet: undefined }
```
vm, line 5: DISCONNECTED?

vm (approved):

vm (approved):

PR #6852, opened 2023-06-28
fix: allow wallets w/o mainnet to connect to WCv2

zzmp:
## Description

Passes the currently active chain as the required chain when first initializing WalletConnect v2, so that those wallets which do not support mainnet (like a gnosis safe created on an L2) will still be able to connect.

_Slack thread:_ https://uniswapteam.slack.com/archives/C03JTMYT96E/p1687948306256509

### Reproducing the error

Create a gnosis safe on an L2 and try to connect it via WCv2

### QA (ie manual testing)

- Ensure Uniswap wallet can still connect
- Ensure gnosis safe issue is fixed
- Ensure another wallet can still connect via WCv2

### DIFF
--- a/cypress/e2e/universal-search.test.ts
+++ b/cypress/e2e/universal-search.test.ts
@@ -1,13 +1,7 @@
-import { getTestSelector } from '../utils'
-
 describe('Universal search bar', () => {
-  before(() => {
+  beforeEach(() => {
     cy.visit('/')
-    cy.get('[data-cy="magnifying-icon"]')
-      .parent()
-      .then(($navIcon) => {
-        $navIcon.click()
-      })
+    cy.get('[data-cy="magnifying-icon"]').parent().eq(1).click()
   })
 
   it('should yield clickable result for regular token or nft collection search term', () => {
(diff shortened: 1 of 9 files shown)


vm (approved):

src/components/WalletModal/Option.test.tsx
```diff
@@ -16,25 +16,25 @@ beforeEach(() => {
...
   getName: () => 'Mock Connection 1',
+  activate: jest.fn(),
   connector: {
-    activate: jest.fn(),
     deactivate: jest.fn(),
   } as unknown as Connector,
   getIcon: () => UNIWALLET_ICON,
   type: ConnectionType.UNISWAP_WALLET,
-} as unknown as Connection
+} as unknown as Connection & { connector: Connector }
```
vm, line 27: can you explain this change?

zzmp: In order to mock, we cannot use an optional operator, so we need to assert that this type is not undefined. Rather than use an `assert` I just define the type in it's declaration.

vm (approved):

vm (approved):

cartcrom (approved):

vm (approved):

vm (approved):

PR #6855, opened 2023-06-28
chore: remove WalletConnect feature flags

vm:
Removes all WalletConnect related flags now that we are at 100% rollout for v2.

To test this we have some existing tests but otherwise will test:
- Connecting to Uniswap Wallet
- Connecting to other WalletConnect wallet on Desktop
- Connecting to other WalletConnect wallet on Mobile


JFrankfurt (approved):

src/components/WalletModal/Option.tsx
```diff
@@ -120,90 +76,18 @@ const Wrapper = styled.div<{ disabled: boolean }>`
...
-}
-
 interface OptionProps {
   connection: Connection
 }
 export default function Option({ connection }: OptionProps) {
   const { activationState, tryActivation } = useActivationState()
-  const [wCPopoverOpen, setWCPopoverOpen] = useState(false)
-  const [accountDrawerOpen, toggleAccountDrawerOpen] = useAccountDrawer()
+  const [, toggleAccountDrawerOpen] = useAccountDrawer()
```
zzmp, line 84: Use `useToggleAccountDrawer` from the same file to simplify this a smidge.

src/components/WalletModal/index.tsx
```diff
@@ -46,11 +45,7 @@ export default function WalletModal({ openSettings }: { openSettings: () => void
 
   const { activationState } = useActivationState()
 
-  const walletConnectV2AsDefault = useWalletConnectV2AsDefault()
-  const hiddenWalletConnectTypes = [
-    walletConnectV2AsDefault ? ConnectionType.WALLET_CONNECT : ConnectionType.WALLET_CONNECT_V2,
-    walletConnectV2AsDefault ? ConnectionType.UNISWAP_WALLET : ConnectionType.UNISWAP_WALLET_V2,
-  ]
+  const hiddenWalletConnectTypes = [ConnectionType.WALLET_CONNECT, ConnectionType.UNISWAP_WALLET]
```
zzmp, line 48: I think this logic is no longer necessary, and can be removed.

zzmp (changes requested):

src/components/WalletModal/index.tsx
vm: I do need this because we haven't deleted the old versions yet from code. I have another PR which handles that.

zzmp (approved):

src/connection/activate.ts
```diff
@@ -32,7 +32,7 @@ function useTryActivation() {
   const currentPage = getCurrentPageFromLocation(pathname)
 
   return useCallback(
-    async (connection: Connection, onSuccess: () => void, chainId?: SupportedChainId) => {
+    async (connection: Connection, onSuccess?: () => void, chainId?: SupportedChainId) => {
```
vm, line 35: I needed to change this to get the test to pass. If we're not happy with this we can try to fix the test instead?

zzmp: which test?

zzmp (changes requested):
I don't understand why `onSuccess` needs to be optional. Can that be fixed here?

zzmp (approved):

src/components/WalletModal/Option.test.tsx
```diff
@@ -13,7 +13,7 @@ jest.mock('components/AccountDrawer')
 
 beforeEach(() => {
   jest.spyOn(console, 'debug').mockReturnValue()
-  mocked(useAccountDrawer).mockReturnValue([true, mockToggleDrawer])
+  mocked(useToggleAccountDrawer).mockReturnValue(mockToggleDrawer)
```
zzmp, line 16: I think you should add an add'l assertion in the test that this is actually called.
That it was missing is why you felt comfortable changing `onSuccess` to optional, before.

zzmp (changes requested):

src/components/WalletModal/Option.test.tsx
```diff
@@ -64,7 +64,9 @@ describe('Wallet Option', () => {
     expect(option2).toBeEnabled()
     expect(option2).toHaveProperty('selected', false)
 
-    await act(() => option1.click())
+    expect(mockConnection1.connector.activate).toHaveBeenCalledTimes(0)
```
zzmp, line 67:
```suggestion
```

zzmp (approved):

PR #6856, opened 2023-06-28
chore: remove walletConnectV2 flag

vm:
Removes the defaulting logic as we should always show WalletConnect v2.

### DIFF
--- a/src/components/AccountDrawer/UniwalletModal.tsx
+++ b/src/components/AccountDrawer/UniwalletModal.tsx
@@ -1,17 +1,14 @@
 import { Trans } from '@lingui/macro'
 import { sendAnalyticsEvent } from '@uniswap/analytics'
 import { InterfaceElementName } from '@uniswap/analytics-events'
-import { WalletConnect } from '@web3-react/walletconnect'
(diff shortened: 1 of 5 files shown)


JFrankfurt (approved):

src/components/AccountDrawer/UniwalletModal.tsx
```diff
@@ -51,21 +48,12 @@ export default function UniwalletModal() {
...
-      connectorV1.events.addListener(UniwalletConnect.UNI_URI_AVAILABLE, (uri: string) => {
-        uri && setUri(uri)
-      })
-    }
-  }, [walletConnectV2AsDefault])
+    const connectorV2 = uniwalletWCV2ConnectConnection.connector as WalletConnectv2
+    connectorV2.events.addListener(UniwalletConnectV2.UNI_URI_AVAILABLE, (uri: string) => {
+      uri && setUri(uri)
+    })
+  }, [])
```
cbachmeier, line 56: nit: do we still want this wrapped in a useEffect if there's no depedencies?

vm: i think this is async so it has to be in an effect?

PR #6878, opened 2023-07-03
fix: Handle new and unsupported chains passed in from GQL BE

cbachmeier:
## Description

- Adds subset type of `Chain` that only includes currently supported chains on the UX: `InterfaceGqlChain` 
  - **This will allow the BE to add a new Chain without causing our build to fail**
- Helper function that converts `Chain` to `SupportedChainId` or returns undefined if an unsupported chain is passed in
  - Added tests for these fns
- If the ux receives data from GQL from an unsupported chain, log an error in Sentry, and escalate to BE
  - Note this is different than a user connecting to the UX from an unsupported chain. This refers to data we received from our GQL BE was for a chain we currently do not support. This is not expected behaviour and BE team should be notified.
  - There's also a BE ticket to add Chain filtering to other queries like we do for Token Balances https://linear.app/uniswap/issue/DAT-55/add-chain-filter-to-miniportfolio-transactions-query
- Use the `GQL_MAINNET_CHAINS` array to dynamically fetch token balances
  - This fixes a bug where celo tokens were not being shown in portfolio

### Before
Screenshot 2023-07-03 at 5 25 49 PM

### After
Screenshot 2023-07-03 at 5 48 55 PM

### QA (ie manual testing)

- [x] Connect to the UX from a supported chain and valid data loads in Explore and MiniPort
- [x] Connect to the UX from an unsupported chain and valid data loads in Explore and MiniPort
- [x] Added an unsupported chain to `types-and-hooks` and build did not crash


JFrankfurt (commented):
Could you take a look at adding a test case for mismatched backend chains?

src/graphql/data/util.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+      expect(fromGraphQLChain(chain)).not.toBe(undefined)
+    }
+  })
+
+  it('should return undefined for unsupported chains', () => {
+    expect(fromGraphQLChain(Chain.UnknownChain)).toBe(undefined)
+
+    for (const chain of Object.values(Chain)) {
+      if (isSupportedGQLChain(chain)) continue
+      expect(fromGraphQLChain(chain)).toBe(undefined)
```
JFrankfurt, line 21: https://github.com/Uniswap/interface/blob/main/cypress/README.md#avoid-branching-logic

These tests seem solid, but I think this recommendation applies here. It would be simpler to just provide two blocks, one with a chain that is supported, and a second with the unknown chain.

cbachmeier: the reason i added the loop was to allow the test to auto update as new supported chains are added

If we have a strong aversion to the branching logic I can make it more static with the tradeoff that when adding a new supported chain implementing engineer should add it to this test

src/graphql/data/util.tsx
```diff
@@ -119,21 +134,23 @@ export function validateUrlChainParam(chainName: string | undefined) {
...
   [Chain.Bnb]: SupportedChainId.BNB,
 }
 
-export function fromGraphQLChain(chain: Chain): SupportedChainId {
-  return CHAIN_NAME_TO_CHAIN_ID[chain]
+export function isSupportedGQLChain(chain: Chain): chain is InterfaceGqlChain {
+  return (UX_SUPPORTED_GQL_CHAINS as ReadonlyArray<Chain>).includes(chain)
+}
+
+export function fromGraphQLChain(chain: Chain): SupportedChainId | undefined {
```
just-toby, line 152: `fromGraphQLChain` - this doesn't describe what you're getting back. maybe `chainIdFromGraphQLChain` ?

src/components/AccountDrawer/PrefetchBalancesWrapper.tsx
```diff
@@ -44,7 +45,7 @@ export default function PrefetchBalancesWrapper({ children }: PropsWithChildren)
   const [hasUnfetchedBalances, setHasUnfetchedBalances] = useState(true)
   const fetchBalances = useCallback(() => {
     if (account) {
-      prefetchPortfolioBalances({ variables: { ownerAddress: account } })
+      prefetchPortfolioBalances({ variables: { ownerAddress: account, chains: GQL_MAINNET_CHAINS.concat() } })
```
just-toby, line 48: why `concat()` with no second array as an argument? if you're trying to make a shallow copy here i think there's more canonical ways - `.slice()` would be my go to

src/components/AccountDrawer/MiniPortfolio/Tokens/index.tsx
```diff
@@ -31,7 +31,7 @@ export default function Tokens({ account }: { account: string }) {
   const [showHiddenTokens, setShowHiddenTokens] = useState(false)
 
   const { data } = usePortfolioBalancesQuery({
-    variables: { ownerAddress: account },
+    variables: { ownerAddress: account, chains: GQL_MAINNET_CHAINS.concat() },
```
just-toby, line 34: see comment about `concat()` below

src/components/AccountDrawer/PrefetchBalancesWrapper.tsx
cbachmeier: it was intended to be a mutable copy of the Readonly array. Based on the these discussion posts concat and slice seem interchangeable, I'm fine with making our standard `slice`
- https://stackoverflow.com/questions/58981365/converting-a-readonlyarrayany-to-a-usual-mutable-array
- https://stackoverflow.com/questions/17803586/javascript-array-copying-concat-vs-slice-which-one-is-better

src/graphql/data/util.tsx
cbachmeier: I went with `supportedChainIdFromGQLChain`

just-toby (approved):

src/components/AccountDrawer/AuthenticatedHeader.tsx
```diff
@@ -227,7 +228,7 @@ export default function AuthenticatedHeader({ account, openSettings }: { account
   const closeFiatOnrampUnavailableTooltip = useCallback(() => setShow(false), [setShow])
 
   const { data: portfolioBalances } = usePortfolioBalancesQuery({
-    variables: { ownerAddress: account ?? '' },
+    variables: { ownerAddress: account ?? '', chains: GQL_MAINNET_CHAINS.slice() },
```
tinaszheng, line 231: this will create a new array instance on every render - is that what you want? why not just pass in `GQL_MAINNET_CHAINS` as is?

src/components/NavBar/RecentlySearchedAssets.ts
```diff
@@ -86,7 +86,15 @@ export function useRecentlySearchedAssets() {
     shortenedHistory.forEach((asset) => {
       if (asset.address === 'NATIVE') {
         // Handles special case where wMATIC data needs to be used for MATIC
-        const native = nativeOnChain(CHAIN_NAME_TO_CHAIN_ID[asset.chain] ?? SupportedChainId.MAINNET)
+        const chain = supportedChainIdFromGQLChain(asset.chain)
+        if (!chain) {
+          Sentry.withScope((scope) => {
```
tinaszheng, line 91: maybe create a helper function that does all the Sentry stuff for you across the app? otherwise feels a bit repetitive

```
function assertDefined(chain, { extras, errorMessage }) {
  Sentry.withScope((scope) => {
    Object.entries(extra).map(([k, v]) => {
      scope.setExtra(k, v)
    })
    Sentry.captureException(new Error(errorMessage))
  })
}
```

src/graphql/data/util.tsx
```diff
@@ -119,21 +134,23 @@ export function validateUrlChainParam(chainName: string | undefined) {
...
   [Chain.Bnb]: SupportedChainId.BNB,
 }
 
-export function fromGraphQLChain(chain: Chain): SupportedChainId {
-  return CHAIN_NAME_TO_CHAIN_ID[chain]
+export function isSupportedGQLChain(chain: Chain): chain is InterfaceGqlChain {
+  return (UX_SUPPORTED_GQL_CHAINS as ReadonlyArray<Chain>).includes(chain)
+}
+
+export function supportedChainIdFromGQLChain(chain: Chain): SupportedChainId | undefined {
```
tinaszheng, line 152: you can add an extra typing here to specify that if the user passes in an `InterfaceGqlChain` then the return type will definitely be defined

```
export function supportedChainIdFromGQLChain(chain: InterfaceGqlChain): SupportedChainId
export function supportedChainIdFromGQLChain(chain: Chain): SupportedChainId | undefined {
  return isSupportedGQLChain(chain) ? CHAIN_NAME_TO_CHAIN_ID[chain] : undefined
}
```

src/components/Tokens/TokenTable/NetworkFilter.tsx
```diff
@@ -143,7 +144,7 @@ export default function NetworkFilter() {
       {open && (
         <MenuTimeFlyout>
           {BACKEND_CHAIN_NAMES.map((network) => {
-            const chainInfo = getChainInfo(CHAIN_NAME_TO_CHAIN_ID[network])
+            const chainInfo = getChainInfo(supportedChainIdFromGQLChain(network))
```
tinaszheng, line 147: so ideally `BACKEND_CHAIN_NAMES` is typed as `InterfaceGqlChain[]` and if you do the typescript function overloading thing mentioned in another comment for `supportedChainIdFromGQLChain`, then you `chainInfo` will always be defined and you won't have to do the null check on 148 and 172-173

tinaszheng (changes requested):

src/components/AccountDrawer/MiniPortfolio/Tokens/index.tsx
```diff
@@ -103,6 +103,7 @@ function TokenRow({ token, quantity, denominatedValue, tokenProjectMarket }: Tok
   }, [navigate, token, toggleWalletDrawer])
 
   const currency = gqlToCurrency(token)
+  if (!currency) return null
```
tinaszheng, line 106: sentry exception here too?

src/components/AccountDrawer/AuthenticatedHeader.tsx
tinaszheng: so looks like our codegen doesn't mark these as `readonly` which is more correct - i made a separate PR that helps out with that: https://github.com/Uniswap/interface/pull/6889

cbachmeier: GQL_MAINNET_CHAINS is readonly, the query needs to have a mutable copy of the array passed in

tinaszheng: the better solution here is to mark the variables accept readonly as opposed to making a copy on every render

cbachmeier: `usePortfolioBalancesQuery` is autogenerated by Apollo

cbachmeier: What we could do is in utils have a readonly array used for typing and make a copy used as a mutable array ie:

```
const GQL_MAINNET_CHAINS_READONLY = [
  Chain.Ethereum,
  Chain.Polygon,
  Chain.Celo,
  Chain.Optimism,
  Chain.Arbitrum,
  Chain.Bnb,
] as const

export const GQL_MAINNET_CHAINS_MUTABLE = GQL_MAINNET_CHAINS_READONLY.slice()

src/components/NavBar/RecentlySearchedAssets.ts
cbachmeier: made some slight modifications wrt naming and removed chain as an arg as it was redundant with the extras

cbachmeier:
Failing test is not caused by this PR but because of another token with name ETH making it into the top 100. Fix in separate PR https://github.com/Uniswap/interface/pull/6890

src/components/AccountDrawer/AuthenticatedHeader.tsx
cbachmeier: @tinaszheng  Thank you for making the apollo change. I've resolved the merge conflicts and merged that change into my branch

tinaszheng (approved):
LETS GO

PR #6883, opened 2023-07-05
feat: use path-based router with hash fallback for IPFS

grabbou:
## Description

As per the design doc, using the path-based router for Cloudflare deployments (and environments that support catch-all index.html) with hash fallback for IPFS.

This in theory should not require any further work, as most linking is handled internally by `<Link />` and other primitives from React Router that abstract this away for us already.

### DIFF
--- a/.github/workflows/4-deploy-to-prod.yml
+++ b/.github/workflows/4-deploy-to-prod.yml
@@ -23,7 +23,10 @@ jobs:
 
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup
+      
       - run: yarn build
+        env:
+          IS_IPFS_BUILD: 1
 
       - name: Bump and tag
         id: github-tag-action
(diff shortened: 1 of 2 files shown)


.github/workflows/4-deploy-to-prod.yml
```diff
@@ -23,7 +23,10 @@ jobs:
 
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup
+      
       - run: yarn build
+        env:
+          IS_IPFS_BUILD: 1
```
zzmp, line 29: The same build is deployed to both IPFS and CF. In order to differentiate, you'd need to either:
- make two separate builds
- use a runtime check on an env var set in CF
- check the origin (ie `isAppUniswapOrg() || isAppUniswapStagingOrg()`)

zzmp (changes requested):

.github/workflows/4-deploy-to-prod.yml
grabbou: Yeah, sorry, for some reason, not a full diff got shipped. I initially settled on 
> make two separate builds
So, re-build in the latter stage with a flag set. Considered that a bit easier to maintain than two separate build pipelines.

PR #6885, opened 2023-07-05
build: use webpack-retry-chunk-load-plugin

vm:
Instead of handling retries at the client level in application code by manually wrapping all `import` function usage, we now enable it at the transpilation level.

By using this webpack plugin, we ensure that all code automatically gets retry logic added. We get the added benefit that any npm packages we use also get retry logic added, which was an issue for us with WalletConnect's bundles not being loaded properly.

Implements retry with backoff for 3 attempts. Also reduces the starting retry wait from 1s to .5s.

See that this retries for a blocked URL of a chunk with WalletConnect modal content in it.

Screen Shot 2023-07-06 at 3 25 49 PM

Chunk: https://app.uniswap.org/static/js/4458.0e527482.chunk.js?cache-bust=true&retry-attempt=1
Location of code split: https://github.com/WalletConnect/modal/blob/main/projects/modal/src/client.ts#L21


craco.config.cjs
```diff
@@ -93,6 +94,10 @@ module.exports = {
       // See https://vanilla-extract.style/documentation/integrations/webpack/#identifiers for docs.
       // See https://github.com/vanilla-extract-css/vanilla-extract/issues/771#issuecomment-1249524366.
       new VanillaExtractPlugin({ identifiers: 'short' }),
+      new RetryChunkLoadPlugin({
+        retryDelay: 2000,
```
zzmp, line 98: This should be a function to add exponential backoff.

craco.config.cjs
```diff
@@ -93,6 +94,10 @@ module.exports = {
       // See https://vanilla-extract.style/documentation/integrations/webpack/#identifiers for docs.
       // See https://github.com/vanilla-extract-css/vanilla-extract/issues/771#issuecomment-1249524366.
       new VanillaExtractPlugin({ identifiers: 'short' }),
+      new RetryChunkLoadPlugin({
```
zzmp, line 97: I think you'll want to include `cacheBust`.

craco.config.cjs
vm: Not possible in their repo, so I'll have to fork if we want that.

yarn.lock
```diff
@@ -16093,10 +16093,10 @@ prettier-linter-helpers@^1.0.0:
   dependencies:
     fast-diff "^1.1.2"
 
-prettier@^2.1.2, prettier@^2.7.1, prettier@^2.8.0:
-  version "2.8.1"
-  resolved "https://registry.yarnpkg.com/prettier/-/prettier-2.8.1.tgz#4e1fd11c34e2421bc1da9aea9bd8127cd0a35efc"
-  integrity sha512-lqGoSJBQNJidqCHE80vqZJHWHRFoNYsSpP9AjFhlhi9ODCJA541svILes/+/1GM3VaL/abZi7cpFzOpdR9UPKg==
+prettier@^2.1.2, prettier@^2.6.2, prettier@^2.7.1, prettier@^2.8.0:
+  version "2.8.8"
```
zzmp, line 16097: Prettier bumping to 2.8.8 is the reason for all of the "unrelated" lint changes. Can you back this out?
You can still update prettier, but it should be split to a separate PR.

craco.config.cjs
```diff
@@ -93,6 +94,12 @@ module.exports = {
...
+        retryDelay: `function(retryAttempt) {
+          return 2 ** (retryAttempt - 1) * 1000;
```
zzmp, line 99: This first retry will not occur for 1 second. Is this the intended behavior?
I think this is the same as the old behavior - just double-checking.

zzmp (changes requested):
Do you need to add query parameters when retrying? Is it possible (with this plugin) to omit these query parameters?

craco.config.cjs
vm: yeah this is same as existing behavior, but i'm open to changing this around if we want. could do 500ms to start?

vm:
I don't think we need the query params, let me see if I can remove them

yarn.lock
vm: I'll make a separate PR 👍

vm: i'll rebase once the other pr is merged

zzmp (commented):
pls add back in the cache buster

craco.config.cjs
```diff
@@ -93,6 +94,16 @@ module.exports = {
       // See https://vanilla-extract.style/documentation/integrations/webpack/#identifiers for docs.
       // See https://github.com/vanilla-extract-css/vanilla-extract/issues/771#issuecomment-1249524366.
       new VanillaExtractPlugin({ identifiers: 'short' }),
+      new RetryChunkLoadPlugin({
+        cacheBust: `function() {
+          return 'cache-bust=' + Date.now();
+        }`,
+        // Retries with backoff (500ms, 1000ms, 2000ms).
```
zzmp, line 101:
```suggestion
        // Retries with exponential backoff (500ms, 1000ms, 2000ms).
```

craco.config.cjs
```diff
@@ -93,6 +94,16 @@ module.exports = {
       // See https://vanilla-extract.style/documentation/integrations/webpack/#identifiers for docs.
       // See https://github.com/vanilla-extract-css/vanilla-extract/issues/771#issuecomment-1249524366.
       new VanillaExtractPlugin({ identifiers: 'short' }),
+      new RetryChunkLoadPlugin({
+        cacheBust: `function() {
+          return 'cache-bust=' + Date.now();
```
zzmp, line 99:
```suggestion
          return Date.now();
```

zzmp (changes requested):

craco.config.cjs
vm: you're fine with the URL being `app.uniswap.org/chunk.3.js?253523525`?

zzmp (approved):

craco.config.cjs
```diff
@@ -93,6 +94,16 @@ module.exports = {
       // See https://vanilla-extract.style/documentation/integrations/webpack/#identifiers for docs.
       // See https://github.com/vanilla-extract-css/vanilla-extract/issues/771#issuecomment-1249524366.
       new VanillaExtractPlugin({ identifiers: 'short' }),
+      new RetryChunkLoadPlugin({
+        cacheBust: `function() {
```
just-toby, line 98: FYI you actually can just define these functions in Javascript above, and call `.toString()` on them

function cacheBust() {
  return 'cache-bust=' + Date.now();
}

function retryDelay(retryAttempt) {
  return 2 ** (retryAttempt - 1) * 500;
}

// Use the functions and convert to strings
new RetryChunkLoadPlugin({
  cacheBust: cacheBust.toString(),
  retryDelay: retryDelay.toString(),
  maxRetries: 3,
});

just-toby (approved):

PR #6888, opened 2023-07-06
chore: remove Google Analytics from CSP

vm:
## Description
Disables Google Analytics by changing the provider to not send any events. Also removes Google Analytics and Google Tag Manager from the CSP.

Once we migrate the events to Amplitude, we can merge this PR: https://github.com/Uniswap/interface/pull/6554

### QA (ie manual testing)
- [x] Web vitals are not being sent to GA
- [x] Init is not happening
- [x] Events are not being sent 

### Automated testing
- [ ] Unit test
- [ ] Integration/E2E test
N/A

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -19,9 +19,9 @@
     <meta
       http-equiv="Content-Security-Policy"
       <% if (process.env.REACT_APP_CSP_ALLOW_UNSAFE_EVAL) { %>
-        content="script-src 'self' https://www.google-analytics.com https://www.googletagmanager.com 'unsafe-inline' 'unsafe-eval'"
+        content="script-src 'self' 'unsafe-inline' 'unsafe-eval'"
       <% } else { %>
-        content="script-src 'self' https://www.google-analytics.com https://www.googletagmanager.com 'unsafe-inline'"
+        content="script-src 'self' 'unsafe-inline'"
       <% } %>
     />
 
@@ -36,8 +36,6 @@
       See https://developers.google.com/web/fundamentals/web-app-manifest/
     -->
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
-
-    <link rel="preconnect" href="https://www.google-analytics.com/" />
 
     <link rel="preload" href="%PUBLIC_URL%/fonts/Inter-roman.var.woff2" as="font" type="font/woff2" crossorigin />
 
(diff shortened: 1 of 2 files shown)


public/index.html
```diff
@@ -19,9 +19,9 @@
     <meta
       http-equiv="Content-Security-Policy"
       <% if (process.env.REACT_APP_CSP_ALLOW_UNSAFE_EVAL) { %>
-        content="script-src 'self' https://www.google-analytics.com https://www.googletagmanager.com 'unsafe-inline' 'unsafe-eval'"
+        content="script-src 'self' 'unsafe-inline' 'unsafe-eval'"
```
pwnslinger, line 22: Where in the code do we need to have `unsafe-eval`?

vm: I actually do not know yet, will need to look into it separately 🤔

zzmp: We do not, that is why this is under a conditional. `unsafe-eval` is used by istanbul to provide coverage reporting, so e2e test builds are built with `REACT_APP_CSP_ALLOW_UNSAFE_EVAL=true` so that istanbul can function.

src/components/analytics/GoogleAnalyticsProvider.tsx
```diff
@@ -1,12 +1,13 @@
-import ReactGA from 'react-ga4'
+/* eslint-disable no-unused-vars */
```
zzmp, line 1: Why are you disabling lint?

src/components/analytics/GoogleAnalyticsProvider.tsx
```diff
@@ -1,12 +1,13 @@
...
+
 import { GaOptions, InitOptions, UaEventOptions } from 'react-ga4/types/ga4'
 
 /**
  * Google Analytics Provider containing all methods used throughout app to log events to Google Analytics.
  */
 export default class GoogleAnalyticsProvider {
   public sendEvent(event: string | UaEventOptions, params?: any) {
-    ReactGA.event(event, params)
+    // ReactGA.event(event, params)
```
zzmp, line 10: Why are you commenting these out, instead of just removing them from the codebase?

zzmp (changes requested):
Why isn't this being grouped into the PR that removes Google Analytics? Done separately, this is incomplete and forces us to disable lint rules and include commented out code.

src/components/analytics/GoogleAnalyticsProvider.tsx
zzmp: Add a TODO above this with the PR that will remove this file.

zzmp (approved):

vm:
We are doing this so we can rm GA from the CSP, without removing all the event definitions until we move those into Amplitude. I added a TODO in the code with the eslint disabled.

cartcrom (approved):

Rezacb0413 (approved):

PR #6893, opened 2023-07-06
build: invalidate typecheck cache for dep changes

zzmp:
## Description

Invalidates the typecheck build cache (for testing) if dependencies change.

_Slack thread:_ https://uniswapteam.slack.com/archives/C047U65H422/p1688662674473499

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,8 +38,9 @@ jobs:
       - uses: ./.github/actions/cache-on-main
         with:
           path: node_modules/.cache
-          key: ${{ runner.os }}-tsc-${{ github.run_id }}
-          restore-keys: ${{ runner.os }}-tsc-
+          # Types must be regenerated if any typings change: dependencies or the GraphQL schema.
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -38,8 +38,9 @@ jobs:
       - uses: ./.github/actions/cache-on-main
         with:
           path: node_modules/.cache
-          key: ${{ runner.os }}-tsc-${{ github.run_id }}
-          restore-keys: ${{ runner.os }}-tsc-
+          # Types must be regenerated if any typings change: dependencies or the GraphQL schema.
+          key: ${{ runner.os }}-tsc-${{ hashFiles('yarn.lock', 'src/graphql/**/schema.graphql') }}-${{ github.run_id }}
```
cbachmeier, line 42: do we need to check in the `schema.graphql` files for this to run?

https://github.com/Uniswap/interface/blob/main/.gitignore#L11

vm (approved):

.github/workflows/test.yml
vm: i just figured this got created during the build process

zzmp: no, it will exist because it has been downloaded as part of `./.github/actions/setup` on L37.

cbachmeier (approved):

PR #6896, opened 2023-07-06
feat: add user id to sentry

vm:
## Description
To reduce alerts caused by a single user (or a few users), we now set a user id stored in localStorage between sessions. It uses UUIDv4, which is a package we already had installed.

### QA (ie manual testing)
- [ ] Send an error to Sentry after initializing user id in that session
- [ ] Send an error to Sentry where the user id was recovered from local storage

### Automated testing
- [ ] Unit test
- [ ] Integration/E2E test
N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -249,7 +249,7 @@
     "tiny-invariant": "^1.2.0",
     "ua-parser-js": "^1.0.35",
     "use-resize-observer": "^9.1.0",
-    "uuid": "^8.3.2",
+    "uuid": "^9.0.0",
     "video-extensions": "^1.2.0",
     "wcag-contrast": "^3.0.0",
     "web-vitals": "^2.1.0",
--- a/package.json
+++ b/package.json
@@ -249,7 +249,7 @@
     "tiny-invariant": "^1.2.0",
     "ua-parser-js": "^1.0.35",
     "use-resize-observer": "^9.1.0",
-    "uuid": "^9.0.0",
+    "uuid": "^8.3.2",
     "video-extensions": "^1.2.0",
     "wcag-contrast": "^3.0.0",
     "web-vitals": "^2.1.0",
--- a/src/tracing/index.ts
+++ b/src/tracing/index.ts
@@ -40,6 +40,11 @@ Sentry.init({
   beforeSend,
 })
 
+const SENTRY_USER_ID_KEY = 'sentry-user-id'
+let sentryUserId = localStorage.getItem(SENTRY_USER_ID_KEY)
+if (!sentryUserId) {
+  localStorage.setItem(SENTRY_USER_ID_KEY, (sentryUserId = uuidv4()))
+}
 Sentry.setUser({ id: sentryUserId })
 
 initializeAnalytics(AMPLITUDE_DUMMY_KEY, OriginApplication.INTERFACE, {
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -249,7 +249,7 @@
     "tiny-invariant": "^1.2.0",
     "ua-parser-js": "^1.0.35",
     "use-resize-observer": "^9.1.0",
-    "uuid": "^8.3.2",
+    "uuid": "^9.0.0",
```
vm, line 252: i might not need to upgrade this actually, i thought we didn't have the package at all

src/tracing/index.ts
```diff
@@ -6,6 +6,7 @@ import { initializeAnalytics, OriginApplication } from '@uniswap/analytics'
 import { SharedEventName } from '@uniswap/analytics-events'
 import { isSentryEnabled } from 'utils/env'
 import { getEnvName, isProductionEnv } from 'utils/env'
+import { v4 as v4uuid } from 'uuid'
```
zzmp, line 9: For consistency with the rest of the codebase:

```suggestion
import { v4 as uuidv4 } from 'uuid'
```

src/tracing/index.ts
```diff
@@ -16,6 +17,14 @@ window.GIT_COMMIT_HASH = process.env.REACT_APP_GIT_COMMIT_HASH
...
 export const STATSIG_DUMMY_KEY = 'client-0000000000000000000000000000000000000000000'
 
+const SENTRY_USER_ID_KEY = 'sentry-user-id'
+
+let sentryUserId = localStorage.getItem(SENTRY_USER_ID_KEY)
+if (!sentryUserId) {
+  sentryUserId = v4uuid()
+  localStorage.setItem(SENTRY_USER_ID_KEY, sentryUserId)
+}
+
```
zzmp, line 27: This is unused before Sentry.init, so rather than split it from its usage, move it below the Sentry.init to be colocated with Sentry.setUser.

src/tracing/index.ts
```diff
@@ -31,6 +40,8 @@ Sentry.init({
   beforeSend,
 })
 
+Sentry.setUser({ id: sentryUserId })
```
zzmp, line 43:
```suggestion
const SENTRY_USER_ID_KEY = 'sentry-user-id'
let sentryUserId = localStorage.getItem(SENTRY_USER_ID_KEY)
if (!sentryUserId) {
  localStorage.setItem(SENTRY_USER_ID_KEY, (sentryUserId = uuidv4()))
}
Sentry.setUser({ id: sentryUserId })
```

zzmp (changes requested):

zzmp (approved):
I was too nitpicky here - this lgtm. Changes are nits.

src/tracing/index.ts
zzmp, line 27:
```suggestion
```

zzmp (approved):

src/tracing/index.ts
```diff
@@ -6,12 +6,16 @@ import { initializeAnalytics, OriginApplication } from '@uniswap/analytics'
...
 import { getEnvName, isProductionEnv } from 'utils/env'
+import { v4 as uuidv4 } from 'uuid'
 
 import { beforeSend } from './errors'
 
 // Dump some metadata into the window to allow client verification.
 window.GIT_COMMIT_HASH = process.env.REACT_APP_GIT_COMMIT_HASH
 
+// This is used to identify the user in Sentry.
+const SENTRY_USER_ID_KEY = 'sentry-user-id'
```
JFrankfurt, line 17: define this below with its usage

JFrankfurt (approved):

src/tracing/index.ts
vm: left above because it's a constant, like the other stuff up there

PR #6899, opened 2023-07-07
test(staging): add staging-only t9n test

zzmp:
## Description

Includes a new test suite which is only run on the staging branch. This allows us to test that translations - which are only available on staging - correctly load and mount.

## Test plan
I've cherry-picked the translations (from staging) and disabled the "only run on the staging branch" bit, to prove that the tests will in fact pass. This is visible through CI: [GitHub Job](https://github.com/Uniswap/interface/actions/runs/5481725797) and [Cypress Cloud Run](https://cloud.cypress.io/projects/yp82ef/runs/12471/overview?roarHideRunsWithDiffGroupsAndTags=1).
I've then reverted both of these commits, so that I avoid commiting translations or running these tests on `main`.


.github/workflows/test.yml
```diff
@@ -144,6 +144,7 @@ jobs:
           wait-on: 'http://localhost:3000'
           browser: electron
           group: e2e
+          spec: ${{ github.ref_name == 'releases/staging' && 'cypress/{e2e,staging}/**/*.spec.ts' || 'cypress/e2e/**/*.spec.ts' }}
```
zzmp, line 147: Use 'test' instead of 'spec'

cypress.config.ts
```diff
@@ -28,6 +28,6 @@ export default defineConfig({
       return config
     },
     baseUrl: 'http://localhost:3000',
-    specPattern: 'cypress/e2e/**/*.{js,jsx,ts,tsx}',
+    specPattern: 'cypress/{e2e,staging}/**/*.ts',
```
zzmp, line 31: Add 'test' to the suffix.

zzmp (commented):
Whoops - used 'spec' instead of 'test'.

JackShort (approved):

.github/workflows/test.yml
```diff
@@ -144,6 +144,7 @@ jobs:
           wait-on: 'http://localhost:3000'
           browser: electron
           group: e2e
+          spec: ${{ github.ref_name == 'releases/staging' && 'cypress/{e2e,staging}/**/*.test.ts' || 'cypress/e2e/**/*.test.ts' }}
```
cbachmeier, line 147: do you need this second `'cypress/e2e/**/*.test.ts'` if the regex `cypress/{e2e,staging}/**/*.test.ts` already includes it?

cbachmeier (approved):
Cela me semble bon

.github/workflows/test.yml
zzmp: I don't but I err on the side of specificity

PR #6901, opened 2023-07-07
feat: cloudflare worker to inject meta tags

LunrEclipse:
## Description

Created Cloudflare Workers to dynamically inject metadata into NFTs, NFT Collections, and Tokens. Created routed functions `/asset/[[index]].ts`, `/tokens/[[index]]/ts`, and `collection/[index]/ts` that utilize GraphQL to query the appropriate information and inject them into the `<head>` using Cloudflare's `HTMLRewriter. 

Unit testing can be found in `collection.test.ts`, `nft.test.ts`, and `token.test.ts`. 

### QA (ie manual testing)

- [x] Manually start Cloudflare test environment using `yarn start:cloud` and navigate to the corresponding routes to check if metadata properly injected. 

### Automated testing

- [x] Unit test `yarn test:cloud`

### DIFF
--- a/functions/components/metaTagInjector.ts
+++ b/functions/components/metaTagInjector.ts
@@ -5,7 +5,7 @@ type MetaTagInjectorInput = {
 }
 
 /**
- * Wrapper class for Cloudflare's HTMLRewriter {@link https://developers.cloudflare.com/workers/runtime-apis/html-rewriter}
+ * Listener class for Cloudflare's HTMLRewriter {@link https://developers.cloudflare.com/workers/runtime-apis/html-rewriter}
  * to inject meta tags into the <head> of an HTML document.
  */
 export class MetaTagInjector {
--- a/functions/utils/getAsset.ts
+++ b/functions/utils/getAsset.ts
@@ -1,6 +1,19 @@
 import { AssetDocument } from '../../src/graphql/data/__generated__/types-and-hooks'
 import client from '../client'
 
+function formatTitleName(name: string, collectionName: string, tokenId: string) {
+  if (name) {
+    return name
+  }
+  if (collectionName && tokenId) {
+    return collectionName + ' #' + tokenId
+  }
+  if (tokenId) {
+    return 'Asset #' + tokenId
+  }
+  return 'View NFT on Uniswap'
+}
+
 export default async function getAsset(collectionAddress: string, tokenId: string, url: string) {
   const { data } = await client.query({
     query: AssetDocument,
@@ -15,12 +28,9 @@ export default async function getAsset(collectionAddress: string, tokenId: strin
   if (!asset) {
     return undefined
   }
+  const title = formatTitleName(asset.name, asset.collection?.name, asset.tokenId)
   const formattedAsset = {
-    title: asset.name
-      ? asset.name
-      : asset.collection?.name
-      ? asset.collection.name + ' #' + asset.tokenId
-      : 'Asset #' + asset.tokenId,
+    title,
     image: asset.image?.url,
     url,
   }
--- a/functions/utils/getToken.ts
+++ b/functions/utils/getToken.ts
@@ -1,6 +1,16 @@
 import { TokenDocument } from '../../src/graphql/data/__generated__/types-and-hooks'
 import client from '../client'
 
+function formatTitleName(symbol: string, name: string) {
+  if (symbol) {
+    return 'Get ' + symbol + ' on Uniswap'
+  }
+  if (name) {
+    return 'Get ' + name + ' on Uniswap'
+  }
+  return 'View Token on Uniswap'
+}
+
 export default async function getToken(networkName: string, tokenAddress: string, url: string) {
   const { data } = await client.query({
     query: TokenDocument,
@@ -13,8 +23,9 @@ export default async function getToken(networkName: string, tokenAddress: string
   if (!asset) {
     return undefined
   }
+  const title = formatTitleName(asset.symbol, asset.name)
   const formattedAsset = {
-    title: 'Get ' + asset.symbol + ' on Uniswap',
+    title,
     image: asset.project?.logoUrl,
     url,
   }


functions/components/assetInjector.ts
```diff
@@ -0,0 +1,54 @@
...
+  image: any
+  uniswapUrl: any
+}
+
+export class MetaTagInjector {
+  private input: MetaTagInjectorInput
+
+  constructor(input: MetaTagInjectorInput) {
+    this.input = input
+  }
```
zzmp, line 12:
```suggestion
  constructor(private input: MetaTagInjectorInput) {}
```

functions/components/assetInjector.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  element(element) {
+    //Open Graph Tags
+    element.append(`<meta property="og:title" content = "${this.input.name}"/>`, {
```
zzmp, line 16: Here and below, remove the spacing around the content assignment: `content="`

functions/components/assetInjector.ts
```diff
@@ -0,0 +1,54 @@
+type MetaTagInjectorInput = {
+  name: any
+  image: any
+  uniswapUrl: any
```
zzmp, line 4: This can just be `url` to make it less confusing.

functions/components/collectionInjector.ts
```diff
@@ -0,0 +1,55 @@
+type MetaTagInjectorInput = {
+  name: any
+  image: any
+  uniswapUrl: any
+  isVerified: any
```
zzmp, line 5: Is this used?

functions/components/tokenInjector.ts
```diff
@@ -0,0 +1,56 @@
+type MetaTagInjectorInput = {
+  name: any
+  network: any
```
zzmp, line 3: Is this used?

functions/nfts/asset/[[index]].ts
```diff
@@ -0,0 +1,19 @@
...
+import getAsset from '../../utils/getAsset'
+
+export const onRequest: PagesFunction = async ({ params, request, next }) => {
+  try {
+    const { index } = params
+    const collectionAddress = String(index[0])
+    const tokenId = String(index[1])
+    const data = await getAsset(collectionAddress, tokenId, request.url)
+    if (!data) {
+      return await next()
```
Comment (zzmp) [inline functions/nfts/asset/[[index]].ts:12]:
```suggestion
      return next()
```

functions/nfts/asset/[[index]].ts
```diff
@@ -0,0 +1,19 @@
...
+    }
+    return new HTMLRewriter().on('head', new MetaTagInjector(data)).transform(await next())
```
Comment (zzmp) [inline functions/nfts/asset/[[index]].ts:14]: You can parallelize this, but you'll need to save a reference to `res` to return in case of rejections

```
const resPromise = next()
try {
  const [asset, res] = await Promise.all([assetPromise, resPromise])
  ...
} catch (e) {
  return resPromise
}
```

functions/utils/getApolloClient.ts
```diff
@@ -0,0 +1,21 @@
...
+  return new ApolloClient({
+    connectToDevTools: true,
+    uri: GRAPHQL_ENDPOINT,
+    headers: {
+      'Content-Type': 'application/json',
+      Origin: 'https://app.uniswap.org',
+      'User-Agent':
+        'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.5735.110 Safari/537.36',
+    },
+    cache: new InMemoryCache(),
```
zzmp, line 14: Do you know how caching will work within the worker? I assume it'd be hard to cache across requests.
Will endpoints with the same parameters just be cached for us automatically?

zzmp (changes requested):
It looks like every injector is injecting the same tags with different values. Instead of having three different injectors, which means more code to test, could you write one injector that takes the tag values as input, and then have the handlers convert asset metadata to tag value inputs? This could hopefully reduce your code by a lot.

functions/collection.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      `<meta property="og:image:alt" content = "https://app.uniswap.org/images/512x512_App_Icon.png"/>`
+    )
+    expect(body).toContain(`<meta property="twitter:card" content = "summary_large_image"/>`)
+    expect(body).toContain(`<meta property="twitter:title" content = "${nft.collectionName} on Uniswap"/>`)
+    expect(body).toContain(`<meta property="twitter:image" content = "${nft.image}"/>`)
+    expect(body).toContain(
+      `<meta property="twitter:image:alt" content = "https://app.uniswap.org/images/512x512_App_Icon.png"/>`
+    )
+  }
+}, 50000)
```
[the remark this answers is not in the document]
LunrEclipse: Yeah, so on the initial run, yarn takes extra time to compile and setup the proxy, so the 50 second timeout is needed otherwise the tests will just fail as they wait for the server to get started

functions/collection.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  for (const nft of collections) {
+    const url = 'http://127.0.0.1:3000/nfts/collection/' + nft.address
+    const req = new Request(url)
+    const res = await fetch(req)
+    const body = await res.text()
+    expect(body).toMatchSnapshot()
+    expect(body).toContain(nft.collectionName + ' on Uniswap')
+    expect(body).toContain(nft.image)
+    expect(body).toContain(url)
+    expect(body).toContain(`<meta property="og:title" content = "${nft.collectionName} on Uniswap"/>`)
```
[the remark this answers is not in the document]
LunrEclipse: Good question! I will remove

functions/collection.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    expect(body).toContain(`<meta property="twitter:title" content = "${nft.collectionName} on Uniswap"/>`)
+    expect(body).toContain(`<meta property="twitter:image" content = "${nft.image}"/>`)
+    expect(body).toContain(
+      `<meta property="twitter:image:alt" content = "https://app.uniswap.org/images/512x512_App_Icon.png"/>`
+    )
+  }
+}, 50000)
+
+test('should not inject metadata for invalid calls', async () => {
+  const baseReq = new Request('http://127.0.0.1:3000/nfts/collection/0xed5af388653567af2f388e6224dc7c4b3241c545')
```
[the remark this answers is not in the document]
LunrEclipse: The collection address is invalid (i.e. I don't think the contract exists)

functions/collection.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  const baseRes = await fetch(baseReq)
+  const baseBody = await baseRes.text()
+  expect(baseBody).toMatchSnapshot()
```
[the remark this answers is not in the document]
LunrEclipse: Not really

functions/nfts/asset/[[index]].ts
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable import/no-unused-modules */
+import { MetaTagInjector } from '../../components/metaTagInjector'
+import getAsset from '../../utils/getAsset'
+
+export const onRequest: PagesFunction = async ({ params, request, next }) => {
+  const { index } = params
+  const collectionAddress = String(index[0])
+  const tokenId = String(index[1])
```
Comment (zzmp) [inline functions/nfts/asset/[[index]].ts:8]: Why do you use `String`? This is pretty non-standard - can you instead do something like `index[0]?.toString()`? Or is it guaranteed to be a string?

Is another alternative to change the filepath of this file to capture the collectionAddress and tokenId?

functions/nfts/asset/[[index]].ts
```diff
@@ -0,0 +1,20 @@
...
+  const assetPromise = getAsset(collectionAddress, tokenId, request.url)
+  const resPromise = next()
+  try {
+    const [data, res] = await Promise.all([assetPromise, resPromise])
```
Comment (zzmp) [inline functions/nfts/asset/[[index]].ts:12]: Call it `asset` to match the `assetPromise`.
```suggestion
    const [asset, res] = await Promise.all([assetPromise, resPromise])
```

functions/nfts/asset/[[index]].ts
```diff
@@ -0,0 +1,20 @@
...
+    if (!data) {
+      return resPromise
+    }
+    return new HTMLRewriter().on('head', new MetaTagInjector(data)).transform(res)
+  } catch (e) {
+    return resPromise
```
Comment (zzmp) [inline functions/nfts/asset/[[index]].ts:18]: Do we want to `console.error(e)`? Do you know where it will be logged?

functions/token.test.ts
```diff
@@ -0,0 +1,92 @@
...
+
+test.each(tokens)(
+  'should inject metadata for valid tokens',
+  async (token) => {
+    const url = 'http://127.0.0.1:3000/tokens/' + token.network + '/' + token.address
+    const body = await fetch(new Request(url)).then((res) => res.text())
+    expect(body).toMatchSnapshot()
+    expect(body).toContain('Get ' + token.symbol + ' on Uniswap')
+    expect(body).toContain(token.image)
+    expect(body).toContain(url)
```
zzmp, line 46:
```suggestion
```

functions/tokens/[[index]].ts
```diff
@@ -0,0 +1,24 @@
...
+import getToken from '../utils/getToken'
+
+export const onRequest: PagesFunction = async ({ params, request, next }) => {
+  const { index } = params
+  const networkName = String(index[0]).toUpperCase()
+  let tokenAddress = String(index[1])
+  tokenAddress =
+    tokenAddress !== 'undefined' && tokenAddress === 'NATIVE'
+      ? '0x0000000000000000000000000000000000000000'
+      : tokenAddress
```
Comment (zzmp) [inline functions/tokens/[[index]].ts:12]: Instead of using a var (vs const), pull this logic into a `function toTokenAddress` at the top of the file. This has the benefits of (1) being const and (2) being more self-documenting in code.

functions/utils/getApolloClient.ts
```diff
@@ -0,0 +1,21 @@
...
+export const getApolloClient = () => {
+  return new ApolloClient({
+    connectToDevTools: true,
+    uri: GRAPHQL_ENDPOINT,
+    headers: {
+      'Content-Type': 'application/json',
+      Origin: 'https://app.uniswap.org',
+      'User-Agent':
+        'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.5735.110 Safari/537.36',
+    },
```
zzmp, line 13: Is there a reason for this specific user-agent?

functions/utils/getApolloClient.ts
```diff
@@ -0,0 +1,21 @@
+import { ApolloClient, InMemoryCache } from '@apollo/client'
+const GRAPHQL_ENDPOINT = 'https://api.uniswap.org/v1/graphql'
+
+export const getApolloClient = () => {
+  return new ApolloClient({
+    connectToDevTools: true,
+    uri: GRAPHQL_ENDPOINT,
+    headers: {
+      'Content-Type': 'application/json',
+      Origin: 'https://app.uniswap.org',
```
zzmp, line 10: Do you _need_ an origin?

functions/utils/getApolloClient.ts
```diff
@@ -0,0 +1,21 @@
...
+      'User-Agent':
+        'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.5735.110 Safari/537.36',
+    },
+    cache: new InMemoryCache(),
+    defaultOptions: {
+      watchQuery: {
+        fetchPolicy: 'cache-and-network',
```
zzmp, line 17: Why not `cache-first`? [docs](https://www.apollographql.com/docs/react/data/queries/#supported-fetch-policies)

functions/utils/getApolloClient.ts
```diff
@@ -0,0 +1,21 @@
+import { ApolloClient, InMemoryCache } from '@apollo/client'
+const GRAPHQL_ENDPOINT = 'https://api.uniswap.org/v1/graphql'
+
+export const getApolloClient = () => {
```
zzmp, line 4: You can do this in a follow-up, but you should leave a TODO for yourself:
This instantiates a new client for every call, so the in-memory cache will never even be used. You need to store this as a global var. You can probably just `export default new ApolloClient...`, instead of having a `getApolloClient` method.

functions/utils/getAsset.ts
```diff
@@ -0,0 +1,25 @@
...
+        tokenIds: [tokenId],
+      },
+    },
+  })
+  const asset = data?.nftAssets?.edges[0]?.node
+  if (!asset) {
+    return undefined
+  }
+  const formattedAsset = {
+    title: asset.name ? asset.name : asset.collection?.name + ' #' + asset.tokenId,
```
zzmp, line 20: you should check for `asset.collection?.name` before using it, otherwise you'll get a leading space if it's not there.

functions/utils/getAsset.ts
```diff
@@ -0,0 +1,25 @@
...
+    image: asset.image?.url,
```
zzmp, line 21: These images may not be available. If it is not available, do we still want to inject meta tags for images?

zzmp (changes requested):

functions/collection.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    expect(body).toContain(`<meta property="twitter:title" content="${collection.collectionName} on Uniswap"/>`)
+    expect(body).toContain(`<meta property="twitter:image" content="${collection.image}"/>`)
+    expect(body).toContain(
+      `<meta property="twitter:image:alt" content="https://app.uniswap.org/images/512x512_App_Icon.png"/>`
+    )
+  },
+  50000
+)
+
+const invalidCollections = [
```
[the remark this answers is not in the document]
LunrEclipse: Screenshot 2023-07-13 at 4 33 50 PM

I tried that, but I kept getting redeclaration issues

functions/utils/getApolloClient.ts
LunrEclipse: `cache-first` would make sense, but also if we're using cloudflare's cache wouldn't `network-only` be better?

functions/utils/getApolloClient.ts
LunrEclipse: yeah otherwise it gets autofiltered by backend

functions/utils/getAsset.ts
LunrEclipse: I think it should be fine. Worst case we just have the default Uniswap image as the fallback image url.

functions/components/metaTagInjector.ts
```diff
@@ -7,44 +7,27 @@ type MetaTagInjectorInput = {
 export class MetaTagInjector {
```
zzmp, line 7: The `MetaTagInjectorInput` properties should not use `any`, it should require `string`s.

functions/components/metaTagInjector.ts
```diff
@@ -0,0 +1,50 @@
+type MetaTagInjectorInput = {
+  title: any
+  image: any
+  url: any
+}
+
+export class MetaTagInjector {
```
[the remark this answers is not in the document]
zzmp: nit: The class should also have a doc-comment.

functions/utils/getApolloClient.ts
zzmp: this refers to the in-memory cache (as you've set it), so I think you want both.

functions/utils/getAsset.ts
zzmp: but you shouldn't render the tag at all in that case, right?

zzmp (changes requested):
there's a few unresolved comments left - can you look at those, and then this is virtually good to go.

functions/utils/getApolloClient.ts
LunrEclipse: yeah your right. i guess we can always change it in the cache pr

functions/components/metaTagInjector.ts
```diff
@@ -17,17 +21,19 @@ export class MetaTagInjector {
...
-    this.append(element, 'og:image', this.input.image)
-    this.append(element, 'og:image:width', '1200')
-    this.append(element, 'og:image:height', '630')
+    if (this.input.image) {
+      this.append(element, 'og:image', this.input.image)
+      this.append(element, 'og:image:width', '1200')
+      this.append(element, 'og:image:height', '630')
+      this.append(element, 'og:image:alt', this.input.title)
+      this.append(element, 'twitter:image', this.input.image)
+    }
```
zzmp, line 30: Put this with the twitter tags - just do the same `if (this.input.image) {` check down there.

functions/utils/getApolloClient.ts
```diff
@@ -0,0 +1,20 @@
+import { ApolloClient, InMemoryCache } from '@apollo/client'
+const GRAPHQL_ENDPOINT = 'https://api.uniswap.org/v1/graphql'
+
+//TODO: Figure out how to make ApolloClient global variable
+export default new ApolloClient({
```
zzmp, line 5: this filename is no longer correct. Can it be moved to `functions/client.ts`?

functions/utils/getApolloClient.ts
```diff
@@ -0,0 +1,20 @@
...
+  connectToDevTools: true,
+  uri: GRAPHQL_ENDPOINT,
+  headers: {
+    'Content-Type': 'application/json',
+    Origin: 'https://app.uniswap.org',
+    'User-Agent':
```
zzmp, line 11: Why is the User-Agent set to this value?

functions/tokens/[[index]].ts
```diff
@@ -0,0 +1,27 @@
...
+  const { index } = params
+  const networkName = index[0]?.toString().toUpperCase()
+  let tokenAddress = index[1]?.toString()
+  if (!tokenAddress) {
+    return next()
+  }
+  tokenAddress =
+    tokenAddress !== 'undefined' && tokenAddress === 'NATIVE'
+      ? '0x0000000000000000000000000000000000000000'
+      : tokenAddress
```
Comment (zzmp) [inline functions/tokens/[[index]].ts:15]: Can you move this to a function at top of file to avoid using `let`?

zzmp (changes requested):

functions/utils/getApolloClient.ts
LunrEclipse: No particular reason LOL. According to Andrew I just needed a user-agent, so I used the one he sent me and have been using it ever since

functions/components/metaTagInjector.ts
```diff
@@ -0,0 +1,41 @@
...
+    this.append(element, 'og:type', 'website')
+    this.append(element, 'og:url', this.input.url)
+
+    //Twitter Tags
+    this.append(element, 'twitter:card', 'summary_large_image')
+    this.append(element, 'twitter:title', this.input.title)
+    if (this.input.image) {
+      this.append(element, 'twitter:image', this.input.image)
+    }
+    this.append(element, 'twitter:image:alt', this.input.title)
```
zzmp, line 39:
```suggestion
      this.append(element, 'twitter:image:alt', this.input.title)
    }
```

zzmp (changes requested):
last nit

zzmp (approved):

functions/components/metaTagInjector.ts
```diff
@@ -0,0 +1,41 @@
+type MetaTagInjectorInput = {
+  title: string
+  image?: string
+  url: string
+}
+
+/**
+ * Wrapper class for Cloudflare's HTMLRewriter {@link https://developers.cloudflare.com/workers/runtime-apis/html-rewriter}
```
just-toby, line 8: i would rephrase this as it's not really wrapping HTMLRewriter. maybe say it's a "listener" class for HTMLRewriter events. or a "subscriber"

functions/tokens/[[index]].ts
```diff
@@ -0,0 +1,27 @@
...
+
+const convertTokenAddress = (tokenAddress: string) => {
+  return tokenAddress && tokenAddress === 'NATIVE' ? '0x0000000000000000000000000000000000000000' : tokenAddress
+}
+
+export const onRequest: PagesFunction = async ({ params, request, next }) => {
+  const { index } = params
+  const networkName = index[0]?.toString().toUpperCase()
+  const tokenAddress = convertTokenAddress(index[1]?.toString())
+  if (!tokenAddress) {
```
Comment (just-toby) [inline functions/tokens/[[index]].ts:13]: what if `networkName` is not defined?

functions/utils/getAsset.ts
```diff
@@ -0,0 +1,28 @@
...
+  const asset = data?.nftAssets?.edges[0]?.node
+  if (!asset) {
+    return undefined
+  }
+  const formattedAsset = {
+    title: asset.name
+      ? asset.name
+      : asset.collection?.name
+      ? asset.collection.name + ' #' + asset.tokenId
+      : 'Asset #' + asset.tokenId,
```
just-toby, line 23: this is a little hard to read. maybe use the "nullish coalescing" operator (i.e. `??`) to cut down on visual branching, or just make a helper function

functions/client.ts
```diff
@@ -0,0 +1,20 @@
+import { ApolloClient, InMemoryCache } from '@apollo/client'
+const GRAPHQL_ENDPOINT = 'https://api.uniswap.org/v1/graphql'
+
+//TODO: Figure out how to make ApolloClient global variable
```
just-toby, line 4: what does this mean? are you just trying to share a single instance of the client across files, or something else?

functions/utils/getAsset.ts
just-toby: also, maybe we should have a fallback title here if none of the information is found

functions/utils/getCollection.ts
```diff
@@ -0,0 +1,21 @@
...
+  const { data } = await client.query({
+    query: CollectionDocument,
+    variables: {
+      addresses: collectionAddress,
+    },
+  })
+  const collection = data?.nftCollections?.edges[0]?.node
+  if (!collection || !collection.name) {
+    return undefined
+  }
```
just-toby, line 14: this behaves differently than the asset one right? should we return undefined in asset if we can't find a name / info ?

functions/utils/getToken.ts
```diff
@@ -0,0 +1,22 @@
...
+      chain: networkName,
+      address: tokenAddress,
+    },
+  })
+  const asset = data?.token
+  if (!asset) {
+    return undefined
+  }
+  const formattedAsset = {
+    title: 'Get ' + asset.symbol + ' on Uniswap',
```
just-toby, line 17: is symbol always defined?

just-toby (changes requested):
RC for some discussion, overall looks good though

functions/tokens/[[index]].ts
Comment (LunrEclipse) [reply functions/tokens/[[index]].ts:13]: I don't believe this function ever gets called if `networkName` is undefined.

functions/utils/getCollection.ts
LunrEclipse: From what I've looked into, a lot of assets don't actually have names which is why I had to create the weird branching naming structure. For collections we can do this since for some reason the GraphQL indexes every single valid address as an `NFTContract` even if that is not the case, so the only way we can differentiate whether or not a collection is in fact an `NFTContract` is through the name.

functions/utils/getToken.ts
LunrEclipse: It should be as symbol is part of the constructor for ERC20 contracts, but then again someone can always deployed a modified version so I guess not always.

just-toby (approved):

PR #6905, opened 2023-07-10
feat: add feature flag for avax

cbachmeier:
## Description

- Add feature flag to gate Avalanche launch

_Slack thread:_ https://uniswapteam.slack.com/archives/C052ARNFZGU/p1689010169325719

## Screen capture
avaxGate

### DIFF
--- a/src/components/NavBar/SearchBarDropdown.tsx
+++ b/src/components/NavBar/SearchBarDropdown.tsx
@@ -6,7 +6,7 @@ import { useWeb3React } from '@web3-react/core'
 import clsx from 'clsx'
 import Badge from 'components/Badge'
 import { getChainInfo } from 'constants/chainInfo'
-import { filterChainsForAvalanche } from 'featureFlags/flags/avalanche'
+import { useFilterChainsForAvalanche } from 'featureFlags/flags/avalanche'
 import { HistoryDuration, SafetyLevel } from 'graphql/data/__generated__/types-and-hooks'
 import { useTrendingCollections } from 'graphql/data/nft/TrendingCollections'
 import { SearchToken } from 'graphql/data/SearchTokens'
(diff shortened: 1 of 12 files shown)


zzmp:
FYI merging this will immediately deploy to prod

src/featureFlags/flags/avalanche.ts
```diff
@@ -0,0 +1,15 @@
...
+  return useBaseFlag(FeatureFlag.avalanche)
+}
+
+export function useIsAvalancheEnabled(): boolean {
+  return useAvalancheFlag() === BaseVariant.Enabled
+}
+
+export function filterChainsForAvalanche(chains: ChainId[]): ChainId[] {
+  return chains.filter((chain) => chain !== ChainId.AVALANCHE || useIsAvalancheEnabled())
+}
```
cartcrom, line 15: I'm confused on how this function is able to use a hook... is useIsAvalancheEnabled not actually a react hook?

wouldn't this need to be something like
```suggestion
export function useFilterChainsForAvalanche(chains: ChainId[]): ChainId[] {
  const isAvalancheEnabled = useIsAvalancheEnabled()
  return chains.filter((chain) => chain !== ChainId.AVALANCHE || isAvalancheEnabled)
}
```

but clearly it's compiling so I must be missing something here?

cbachmeier: I've updated the fn to follow hook naming conventions

cartcrom (approved):

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -356,8 +357,8 @@ export const SearchBarDropdown = ({
     trendingCollectionsAreLoading,
     shouldDisableNFTRoutes,
   ])
-
-  const showChainComingSoonBadge = chainId && BACKEND_NOT_YET_SUPPORTED_CHAIN_IDS.includes(chainId) && !isLoading
+  const gatedUnsupportedChains = useFilterChainsForAvalanche([...BACKEND_NOT_YET_SUPPORTED_CHAIN_IDS])
```
vm, line 360: nit: don't think you need to spread this?

vm (approved):

JFrankfurt (approved):

cbachmeier:
For anyone using this as a future reference for adding a new chain feature flag, the gate logic in `src/theme/components/RadialGradientByChainUpdater.ts` to turn on the background gradient worked locally but did not work in prod. A new solution will need to be determined.

PR #6922, opened 2023-07-12
fix: swap divide by zero

JackShort:
* divide by zero error caused by not wrapping inverting price in a try catch
* added regression test

### DIFF
--- a/src/components/swap/TradePrice.tsx
+++ b/src/components/swap/TradePrice.tsx
@@ -33,7 +33,14 @@ export default function TradePrice({ price }: TradePriceProps) {
   const { baseCurrency, quoteCurrency } = price
   const { data: usdPrice } = useUSDPrice(tryParseCurrencyAmount('1', showInverted ? baseCurrency : quoteCurrency))
 
-  const formattedPrice = formatPrice(showInverted ? price : price.invert(), NumberType.TokenTx)
+  let formattedPrice: string
+  try {
+    formattedPrice = showInverted
+      ? formatPrice(price, NumberType.TokenTx)
+      : formatPrice(price.invert(), NumberType.TokenTx)
+  } catch (error) {
+    formattedPrice = '0'
+  }
 
   const label = showInverted ? `${price.quoteCurrency?.symbol}` : `${price.baseCurrency?.symbol} `
   const labelInverted = showInverted ? `${price.baseCurrency?.symbol} ` : `${price.quoteCurrency?.symbol}`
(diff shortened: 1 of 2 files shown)


src/components/swap/TradePrice.tsx
```diff
@@ -33,7 +33,14 @@ export default function TradePrice({ price }: TradePriceProps) {
   const { baseCurrency, quoteCurrency } = price
   const { data: usdPrice } = useUSDPrice(tryParseCurrencyAmount('1', showInverted ? baseCurrency : quoteCurrency))
 
-  const formattedPrice = formatPrice(showInverted ? price : price.invert(), NumberType.TokenTx)
+  let formattedPrice: string
+  try {
+    formattedPrice = showInverted
+      ? formatPrice(price, NumberType.TokenTx)
+      : formatPrice(price.invert(), NumberType.TokenTx)
```
cbachmeier, line 40: nit: the old syntax would cut down on duplicated code
```suggestion
    formattedPrice = formatPrice(showInverted ? price : price.invert(), NumberType.TokenTx)
```

src/components/swap/TradePrice.tsx
```diff
@@ -33,7 +33,14 @@ export default function TradePrice({ price }: TradePriceProps) {
   const { baseCurrency, quoteCurrency } = price
   const { data: usdPrice } = useUSDPrice(tryParseCurrencyAmount('1', showInverted ? baseCurrency : quoteCurrency))
 
-  const formattedPrice = formatPrice(showInverted ? price : price.invert(), NumberType.TokenTx)
+  let formattedPrice: string
+  try {
+    formattedPrice = showInverted
+      ? formatPrice(price, NumberType.TokenTx)
```
cbachmeier, line 39: With this happening across mobile and web, and it being an expected occurrence, can you add a ticket and comment to update `formatPrice` in `conedison` to return '0' and log the divide by 0 instead of throwing?

cbachmeier (approved):

src/components/swap/TradePrice.tsx
JackShort: the problem isn't the `formatPrice` its actually `price.invert()` that is throwing

src/components/swap/TradePrice.tsx
```diff
@@ -33,7 +33,14 @@ export default function TradePrice({ price }: TradePriceProps) {
   const { baseCurrency, quoteCurrency } = price
   const { data: usdPrice } = useUSDPrice(tryParseCurrencyAmount('1', showInverted ? baseCurrency : quoteCurrency))
 
-  const formattedPrice = formatPrice(showInverted ? price : price.invert(), NumberType.TokenTx)
+  let formattedPrice: string
```
zzmp, line 36: nit: if you put this in a helper function you can continue to use a `const` (which is generally less error-prone):

```
const formattedPrice = useMemo(() => {
  try {
    return formatPrice(showInverted ? price : price.invert(), NumberType.TokenTx)
  } catch {
    return '0'
  }
}, [price, showInverted])
```

zzmp (approved):

ASISBusiness (approved):

zzmp (approved):

PR #6923, opened 2023-07-12
feat: use indexeddb for redux-persist storage

vm:
new localstorage state:
Screen Shot 2023-07-12 at 5 59 33 PM

new indexeddb state:
Screen Shot 2023-07-12 at 5 59 42 PM
Screen Shot 2023-07-12 at 5 59 51 PM

old localstorage state:
Screen Shot 2023-07-12 at 6 00 23 PM
old indexeddb state:
Screen Shot 2023-07-12 at 6 00 32 PM


zzmp (changes requested):
You'll need to update a few more references to localStorage:
.github/workflows/4-deploy-to-prod.yaml
cypress/support/commands.ts
src/hooks/useAccountRiskCheck.ts
src/state/global/action.ts
src/state/lists/reducer.ts
src/state/user/reducer.ts
src/tracing/index.ts

Some of these you may want to continue using localStorage for.

You'll also need a migration strategy for users with data already in their localStorage.

vm:
All of those references are okay to keep.

Good point on the migration path...

vm:
Handled here: https://github.com/Uniswap/interface/pull/6830/files#diff-7b5ee586f7bece02448f36cca1d4317fd47f6c1fe8b83d80128d80ca6b40d410R83

src/state/reducer.ts
```diff
@@ -37,10 +37,21 @@ const appReducer = combineReducers({
 
 export type AppState = ReturnType<typeof appReducer>
 
+function createStorage(dbName: string) {
+  const db = localForage.createInstance({
+    name: dbName,
+  })
+  return {
+    getItem: db.getItem,
```
zzmp, line 45: This will not get old state, because it will look in indexedDB for it instead of localStorage. You'll need to write your own migration to pull it out of localStorage and into localForage.

zzmp (changes requested):

src/state/reducer.ts
just-toby: @vm the initial migration can be done in `migrations.ts` - currently i'm copying the old localStorage `lists` into the new naming scheme but leaving it in localStorage. you can modify that to remove it from localStorage entirely and instead write it to the indexDB

src/state/reducer.ts
```diff
@@ -37,10 +37,21 @@ const appReducer = combineReducers({
...
+    setItem: db.setItem,
+    removeItem: db.removeItem,
+  }
+}
+
 const persistConfig = {
   key: 'root',
   version: 0, // see migrations.ts for more details about this version
-  storage,
+  storage: createStorage('redux'),
```
just-toby, line 54: wait, do we want to move everything out of localStorage? i thought we were just moving lists?

if we are switching everything, we need to modify the initial migration since it assumes we're copying localStorage => localStorage

just-toby (changes requested):

PR #6924, opened 2023-07-12
fix: duck type error checking

zzmp:
## Description

Use duck-typing to identify filtered errors, even if they are just thrown objects.

_Targeted Sentry issue:_ https://uniswap-labs.sentry.io/issues/3788454678/?environment=production&project=4504255148851200&query=bookmarks%3Ame&referrer=issue-stream&statsPeriod=24h&stream_index=4

PR #6925, opened 2023-07-12
fix: rename non error messages in Sentry

vm:
## Description
This will help us group the errors in Sentry based on the message, while keeping the other information about the stack trace and serialized information intact.

Otherwise, we end up with grouped errors like this: https://uniswap-labs.sentry.io/issues/3788318733/?query=is%3Aunresolved&referrer=issue-stream&stream_index=12

### Automated testing
- [x] Unit test
- [ ] Integration/E2E test


src/tracing/errors.test.ts
```diff
@@ -45,6 +45,20 @@ describe('beforeSend', () => {
     })
   })
 
+  describe('renameNonErrorMessage', () => {
+    test('should rename the error message', () => {
+      const event = {} as ErrorEvent
+
```
zzmp, line 51:
```suggestion
```

src/tracing/errors.ts
```diff
@@ -18,11 +18,23 @@ export const beforeSend: Required<ClientOptions>['beforeSend'] = (event: ErrorEv
     return null
   }
 
+  renameNonErrorMessages(event, hint)
   updateRequestUrl(event)
 
   return event
 }
 
+// If the error is not an instance of Error, it may be an object with a message property.
```
zzmp, line 27:
```suggestion
// Some libraries throw Error-like objects instead of Errors.
```

src/tracing/errors.ts
```diff
@@ -18,11 +18,23 @@ export const beforeSend: Required<ClientOptions>['beforeSend'] = (event: ErrorEv
...
+// In those cases, we rename the message property so that they're grouped properly in Sentry.
+function renameNonErrorMessages(event: ErrorEvent, hint: EventHint) {
```
zzmp, line 29:
```suggestion
function renameNonErrorEvents(event: ErrorEvent, hint: EventHint) {
```

src/tracing/errors.ts
```diff
@@ -18,11 +18,23 @@ export const beforeSend: Required<ClientOptions>['beforeSend'] = (event: ErrorEv
...
+  if (!(hint.originalException instanceof Error)) {
+    const message = (hint.originalException as any)?.message
+    if (message) {
+      event.message = message
+    }
```
zzmp, line 34: Stack this on https://github.com/Uniswap/interface/pull/6924 and use `isErrorLike`:
```suggestion
    const error = hint.originalException
    if (isErrorLike(error)) {
      event.message = error.message
    }
```

What if `!isErrorLike(error)`? Do we want to do something like:
```
} else {
  event.message = error?.toString().substring(0, 40)
}
```

zzmp (changes requested):

zzmp (approved):

PR #6926, opened 2023-07-13
fix: chain query parameter doesn't block you from switching chains 

cbachmeier:
## Description

Previously when directed to the app with a chain query parameter, when a user switched chains manually after pageload, they'd be prompted to switch back to the query parameter chain
- If a user changes network via their wallet, the chain query parameter will be cleared
- If a user switches chains via the in-app network switcher, the chain query param will be updated to the new chain
  - Param does not change if the chain switch fails (ie user rejection)

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2439/chain-query-param-does-not-clear-when-swapping-chains

### Before
oldChainSwitcher

### After
Change via Network Switcher
changeViaSwitcher

Change via MM
changeViaWallet

### Reproducing the error

1. Load https://app.uniswap.org/#/swap?chain=optimism
2. Switch to OP
3. Attempt to switch to another chain
4. Be re-prompted to switch back to OP

### QA (ie manual testing)

- [x] Switch chain via connected wallet
- [x] Switch chain with in app network switcher without a connected wallet
- [x] Switch chain with in app network switcher with a connected wallet

### Automated testing

- [x] Added test to check for query param added by network switcher
- [x] Added test to switch chains after loading via query param


cypress/e2e/wallet-connection/switch-network.test.ts
```diff
@@ -127,4 +128,17 @@ describe('network switching from URL param', () => {
...
+    cy.get(getTestSelector('web3-status-connected'))
+    cy.wait('@wallet_switchEthereumChain')
+    waitsForActiveChain('Polygon')
+
+    // switching to another chain clears query param
+    switchChain('Ethereum')
+    cy.wait('@wallet_switchEthereumChain')
+    waitsForActiveChain('Ethereum')
+    cy.url().should('not.contain', 'chain=polygon')
+  })
```
JFrankfurt, line 143: good test

JFrankfurt (changes requested):
> If a user switches chains via the in-app network switcher, the chain query param will be updated to the new chain 


Can you add a test for this case?

cbachmeier:
wrt: https://github.com/Uniswap/interface/pull/6926#pullrequestreview-1527423674
@JFrankfurt I'm actually not sure. @zzmp do you know if there's a way in cypress we can mock a connected wallet and have it switch chains?

cbachmeier:
> > If a user switches chains via the in-app network switcher, the chain query param will be updated to the new chain
> 
> Can you add a test for this case?

Added the check for that here https://github.com/Uniswap/interface/pull/6926/files#diff-2bf73383b4cf2cbd5626034f1c548f73c75ee0d833f3898238748633e93f1f93R114

src/hooks/useSelectChain.ts
```diff
@@ -21,13 +24,14 @@ export default function useSelectChain() {
 
       try {
         await switchChain(connector, targetChain)
+        if (isSupportedChain(targetChain)) setSearchParams({ chain: CHAIN_IDS_TO_NAMES[targetChain] })
```
zzmp, line 27: Should this retain other search params?

```suggestion
        if (isSupportedChain(targetChain)) {
          setSearchParams((searchParams) => ({
            ...searchParams,
            chain: CHAIN_IDS_TO_NAMES[targetChain]
          }))
        }
```

src/hooks/useSyncChainQuery.ts
```diff
@@ -31,8 +32,19 @@ export default function useSyncChainQuery() {
   const [searchParams, setSearchParams] = useSearchParams()
 
   useEffect(() => {
-    if (isActive && urlChainId && chainId !== urlChainId) {
+    // Change a user's chain on pageload of the connected chainId does not match the query param chain
```
zzmp, line 35: If this is only meant to run on pageload, should we set it as a mount-only effect (by setting the deps to an empty array)?

src/hooks/useSyncChainQuery.ts
```diff
@@ -31,8 +32,19 @@ export default function useSyncChainQuery() {
...
+    if (isActive && urlChainId && chainIdRef.current === chainId && chainId !== urlChainId) {
       selectChain(urlChainId)
     }
-  }, [urlChainId, selectChain, searchParams, setSearchParams, isActive, chainId])
+    // If a user has a connected wallet and has manually changed their chain, delete the chain query param
+    else if (account && urlChainId && chainIdRef.current !== chainId && chainId !== urlChainId) {
+      searchParams.delete('chain')
```
zzmp, line 41: Why do we delete the chain param here, but set it through useSelectChain?

zzmp (changes requested):

JFrankfurt (approved):

src/hooks/useSyncChainQuery.ts
cbachmeier: useSelectChain is used when a user selects a chain from the in-app Network Switcher and sets the corresponding query parameter

This case is for when a user switches chains using their wallet so we want to clear the existing query parameter if it exists

src/hooks/useSyncChainQuery.ts
cbachmeier: We added deps because connected wallets load on a delay and would overwrite this setting with their chain and user wouldn't be prompted to switch

src/hooks/useSyncChainQuery.ts
zzmp: why do we want to clear it? why do we have two separate ux's here?

cbachmeier: To clear the now out of date original query parameter

zzmp (approved):

src/hooks/useSyncChainQuery.ts
```diff
@@ -37,9 +37,14 @@ export default function useSyncChainQuery() {
       selectChain(urlChainId)
     }
     // If a user has a connected wallet and has manually changed their chain, delete the old chain query param
-    else if (account && urlChainId && chainIdRef.current !== chainId && chainId !== urlChainId) {
-      searchParams.delete('chain')
-      setSearchParams(searchParams)
+    else if (account && chainIdRef.current !== chainId && chainId !== urlChainId) {
```
zzmp, line 40: `else if` should be `} else if` - not broken across two lines.

cbachmeier: I placed this a line below intentionally because I had the comment explaining above it. It reads more clearly imo than having `} else if` on the same line and having the comment be inside of the `if` block

zzmp: I completely agree with you but thought that lint disallowed this. interesting!

src/hooks/useSyncChainQuery.ts
```diff
@@ -21,8 +21,18 @@ function getParsedChainId(parsedQs?: ParsedQs) {
...
+  const chainIdRef = useRef(chainId)
+  const accountRef = useRef(account)
+
+  useMemo(() => {
+    // Update chainIdRef when the account is retrieved from Web3React
+    if (account && account !== accountRef.current) {
+      chainIdRef.current = chainId
+      accountRef.current = account
+    }
+  }, [account, chainId])
```
zzmp, line 35: `useMemo` runs in a different stage than `useEffect`, and should not be used for side effects. Can this be outside of a `useMemo`?

zzmp (changes requested):

src/hooks/useSyncChainQuery.ts
cbachmeier: updated to useeffect

zzmp (changes requested):
When I switch to celo in the switcher, I can no longer switch back to ethereum.

cbachmeier:
> When I switch to celo in the switcher, I can no longer switch back to ethereum.

@zzmp Good find. This was happening when a wallet wasn't connected and a user tried to change back to mainnet after having already changed to another chain. Trying to swap to another chain other than mainnet worked.

The fix was to update `chainIdRef.current` when `isActive` was true instead of `account` to work for users without a connected wallet

src/hooks/useSelectChain.ts
```diff
@@ -21,6 +24,10 @@ export default function useSelectChain() {
 
       try {
         await switchChain(connector, targetChain)
+        if (isSupportedChain(targetChain)) {
```
zzmp, line 27: Instead of checking `isSupportedChain`, I think that L20 could restrict usage to just `async (targetChain: SupportedChainsType) => {`. wdyt?

src/hooks/useSyncChainQuery.ts
```diff
@@ -31,8 +41,23 @@ export default function useSyncChainQuery() {
...
       selectChain(urlChainId)
     }
-  }, [urlChainId, selectChain, searchParams, setSearchParams, isActive, chainId])
+    // If a user has a connected wallet and has manually changed their chain, delete the old chain query param
+    else if (account && chainIdRef.current !== chainId && chainId !== urlChainId) {
+      if (isSupportedChain(chainId)) {
+        searchParams.set('chain', CHAIN_IDS_TO_NAMES[chainId])
+      } else {
+        searchParams.delete('chain')
+      }
```
zzmp, line 54: This logic does not match the comment:

```suggestion
      searchParams.delete('chain')
```

src/hooks/useSyncChainQuery.ts
```diff
@@ -21,8 +21,18 @@ function getParsedChainId(parsedQs?: ParsedQs) {
 }
 
 export default function useSyncChainQuery() {
-  const { chainId, isActive } = useWeb3React()
+  const { chainId, isActive, account } = useWeb3React()
   const parsedQs = useParsedQueryString()
+  const chainIdRef = useRef(chainId)
+  const accountRef = useRef(account)
```
zzmp, line 27: Why is `accountRef` necessary? Why can't you just use `account` as a dep?

src/hooks/useSyncChainQuery.ts
```diff
@@ -21,8 +21,18 @@ function getParsedChainId(parsedQs?: ParsedQs) {
...
+
+  useEffect(() => {
+    // Update chainIdRef when the account is retrieved from Web3React
+    if (account && account !== accountRef.current) {
+      chainIdRef.current = chainId
```
zzmp, line 32: Why can't this be either outside of a `useEffect` or grouped in with the _other_ `useEffect`?

src/hooks/useSyncChainQuery.ts
```diff
@@ -31,8 +41,23 @@ export default function useSyncChainQuery() {
...
+      } else {
+        searchParams.delete('chain')
+      }
+      setSearchParams(searchParams)
+    }
+
+    // If a user has a connected wallet and the chainId matches the query param chain, update the chainIdRef
+    if (isActive && chainId === urlChainId) {
+      chainIdRef.current = urlChainId
+    }
```
zzmp, line 61: Does this _need_ to be in a `useEffect`?

zzmp (changes requested):
The `useSyncChainQuery` seems overly complicated. Does it _need_ to use refs? I feel like this will be harder to maintain and may be more error prone, but if it's the only way to do it then that's also fine.

src/hooks/useSelectChain.ts
cbachmeier: I believe we'd still want the behaviour `await switchChain(connector, targetChain)` to occur on unsupported chains, just not adding them as a query parameter

src/hooks/useSyncChainQuery.ts
cbachmeier: We need to monitor changes to account and chainId for this change but only these values which is a subset of the larger useEffect. This keeps this function from unnecessarily running if there's a state change to one of the other dependencies in the larger useEffect

src/hooks/useSyncChainQuery.ts
cbachmeier: Because account is originally `undefined` when pulled in from useWeb3React and later given a value if there is a connected wallet. We want to update the chainIdRef only when the account is first given that value so we need the context of if `account` has diverged from `accountRef`

src/hooks/useSyncChainQuery.ts
cbachmeier: Comment should be updated in stead, thanks for the callout

src/hooks/useSyncChainQuery.ts
cbachmeier: Yes, we don't want this triggering outside of one of the dependencies changing, which can cascade to a user's chain being accidentally changed

src/hooks/useSyncChainQuery.ts
```diff
@@ -31,8 +41,23 @@ export default function useSyncChainQuery() {
   const [searchParams, setSearchParams] = useSearchParams()
 
   useEffect(() => {
-    if (isActive && urlChainId && chainId !== urlChainId) {
+    // Change a user's chain on pageload if the connected chainId does not match the query param chain
+    if (isActive && urlChainId && chainIdRef.current === chainId && chainId !== urlChainId) {
```
zzmp, line 45: Isn't `isActive` the same as `account`? If so, could you just return early in this useEffect to simplify a bit: `if (!account) return`?

zzmp (changes requested):

src/hooks/useSyncChainQuery.ts
```diff
@@ -31,8 +41,23 @@ export default function useSyncChainQuery() {
...
+      } else {
+        searchParams.delete('chain')
+      }
+      setSearchParams(searchParams)
+    }
+
+    // If a user has a connected wallet and the chainId matches the query param chain, update the chainIdRef
+    if (isActive && chainId === urlChainId) {
+      chainIdRef.current = urlChainId
+    }
```
zzmp, line 61: Can this be moved into the `else` clause above? Other than that, lgtm.

zzmp (changes requested):

zzmp (approved):

src/hooks/useSyncChainQuery.ts
cbachmeier: No `isActive` can be true if `account` is undefined

cbachmeier: That was actually the issue with the Celo bug you pointed out earlier, I had been using account as a check but it doesn't work when no wallet is connected so I switched to isActive

cbachmeier: Commit for that fix for reference https://github.com/Uniswap/interface/pull/6926/files/b26d1d52477a08e6ba662a0be543274c55fc91d9..db314e2da203ea8a040e2a5852d0f54f6e269651

src/hooks/useSyncChainQuery.ts
cbachmeier: It can't because the `else if` around the `else` clause checks for the web3react chainId being different than the chainId from the url query parameter (`chainId !== urlChainId`) whereas this condition checks if they're equal.

I can make this an `else if` as well however which I've done in the latest commit.

zzmp (approved):

PR #6929, opened 2023-07-13
feat: caches GraphQL queries for Cloudflare workers

LunrEclipse:
## Description

Uses Cloudflare's Caching API to cache GraphQL requests for Meta Tag Injection.

### Automated testing

- [x] Unit test N/A


functions/nfts/asset/[[index]].ts
```diff
@@ -9,11 +9,20 @@ export const onRequest: PagesFunction = async ({ params, request, next }) => {
...
+      const data = JSON.parse(await response.text())
+      return new HTMLRewriter().on('head', new MetaTagInjector(data)).transform(await resPromise)
+    } else {
+      const [data, res] = await Promise.all([assetPromise, resPromise])
+      if (!data) {
+        return resPromise
+      }
+      const response = new Response(JSON.stringify(data))
+      await cache.put(request.url, response)
+      return new HTMLRewriter().on('head', new MetaTagInjector(data)).transform(res)
```
Comment (zzmp) [inline functions/nfts/asset/[[index]].ts:24]: You should cache the response _with_ the meta tags already injected to avoid having to redo the work.

functions/nfts/asset/[[index]].ts
```diff
@@ -9,11 +9,20 @@ export const onRequest: PagesFunction = async ({ params, request, next }) => {
   const assetPromise = getAsset(collectionAddress, tokenId, request.url)
   const resPromise = next()
   try {
-    const [data, res] = await Promise.all([assetPromise, resPromise])
-    if (!data) {
-      return resPromise
+    const cache = caches.default
+    const response = await cache.match(request.url)
```
Comment (zzmp) [inline functions/nfts/asset/[[index]].ts:13]: Will the cache clear when we release a new version? Will we need to always get a response from cloudflare? should we only be caching the metadata?

zzmp (changes requested):

functions/nfts/asset/[[index]].ts
Comment (LunrEclipse) [reply functions/nfts/asset/[[index]].ts:13]: I don't believe the cache is cleared upon a new release.

I assume we do always want to get a response from cloudflare as it ideally should be faster to just get the response from the cloudflare cache and return it rather than building our own. The only situation this would not be ideal for is if its a first load, but I can't think of a way to determine that without eliciting a response from cloudflare.

Because the cache doesn't clear upon new version release, I do think it is best if we only cache the metadata as if we cache the entire response, we might serve an older version.

functions/cache.test.ts
```diff
@@ -0,0 +1,51 @@
+import { mocked } from '../src/test-utils/mocked'
```
zzmp, line 1: This is a collection-specific test. You can cover all of caching with one test file, but then you'd need to cache through a utility function that you could test separately.

just-toby (approved):
i don't see anything wrong with this, but i don't understand how `caches` is accessible at the top level. is this just a feature of node?

LunrEclipse:
> i don't see anything wrong with this, but i don't understand how `caches` is accessible at the top level. is this just a feature of node?

No Cloudflare has a Runtime Cache API which we hope to use

https://developers.cloudflare.com/workers/runtime-apis/cache/

functions/token.test.ts
```diff
@@ -19,12 +19,6 @@ const tokens = [
     image:
       'https://raw.githubusercontent.com/Uniswap/assets/master/blockchains/ethereum/assets/0x7D1AfA7B718fb893dB30A3aBc0Cfc608AaCfeBB0/logo.png',
   },
-  {
```
zzmp, line 22: Should this change be in a separate PR?

functions/utils/useCache.ts
```diff
@@ -0,0 +1,15 @@
+export async function getCache(request: string, name: string) {
```
zzmp, line 1: there's no hook here, so you should rename this file to `cache.ts`

functions/cache.test.ts
```diff
@@ -0,0 +1,30 @@
+import CacheMock from 'browser-cache-mock'
```
zzmp, line 1: Can this be moved to be colocated with the file it tests: functions/utils/cache.test.ts?

functions/tokens/[[index]].ts
```diff
@@ -16,11 +17,17 @@ export const onRequest: PagesFunction = async ({ params, request, next }) => {
   const tokenPromise = getToken(networkName, tokenAddress, request.url)
```
Comment (zzmp) [inline functions/tokens/[[index]].ts:17]: You shouldn't be getting the token unless you have a cache miss (similarly in other files). Something like:

```
const cachedToken = await getCache(request.url, 'tokens-cache')
if (!cachedToken) {
  const token = await getToken(networkName, tokenAddress, request.url)
  if (!token) return res
  
  putCache(new Response(JSON.stringify(token)), request.url, 'tokens-cache')
  return new HTMLRewriter().on('head', new MetaTagInjector(data)).transform(await res)
}
```

functions/utils/useCache.ts
```diff
@@ -0,0 +1,15 @@
...
+  const cache = await caches.open(name)
+  const response = await cache.match(request)
+  if (response) {
+    const data = JSON.parse(await response.text())
+    return data
+  }
+  return undefined
+}
+
+export async function putCache(response: Response, request: string, name: string) {
```
zzmp, line 11: Does the cached object _need_ to be a `Response`, because you're not actually dealing with `Response` objects.

functions/utils/useCache.ts
```diff
@@ -0,0 +1,15 @@
+export async function getCache(request: string, name: string) {
+  const cache = await caches.open(name)
+  const response = await cache.match(request)
+  if (response) {
+    const data = JSON.parse(await response.text())
+    return data
+  }
+  return undefined
```
zzmp, line 8:
```suggestion
  if (!response) return undefined
  return JSON.parse(await response.text())
```

functions/utils/useCache.ts
zzmp, line 1: These would be easier to test and to read if you exposed a `Cache` singleton instead, with `get` and `put` methods.

functions/utils/useCache.ts
```diff
@@ -0,0 +1,15 @@
...
+}
+
+export async function putCache(response: Response, request: string, name: string) {
+  response.headers.set('Cache-Control', 'max-age=3600')
```
zzmp, line 12: Why this max-age?

zzmp (changes requested):

functions/token.test.ts
LunrEclipse: Yeah I already merged this change into `main` but I updated it on this branch to stop the failure and then just forgot to merge `main` back in. Should be good now!

functions/utils/useCache.ts
LunrEclipse: I believe so

Screenshot 2023-08-03 at 10 51 51 AM

functions/utils/useCache.ts
LunrEclipse: Beautiful!

functions/utils/useCache.ts
LunrEclipse: Like just returning `await caches.open(name)` and then allowing the functions to call the `get` and `put` methods?

functions/utils/useCache.ts
LunrEclipse: It was very arbitrarily chosen LOL

functions/tokens/[[index]].ts
```diff
@@ -18,22 +18,18 @@ const convertTokenAddress = (tokenAddress: string, networkName: string) => {
...
+    const { index } = params
+    const networkName = index[0]?.toString().toUpperCase()
+    const tokenString = index[1]?.toString()
+    if (!tokenString) {
+      return res
     }
-    return new HTMLRewriter().on('head', new MetaTagInjector(data)).transform(res)
+    const tokenAddress = convertTokenAddress(tokenString, networkName)
+    const graphCall = getToken(networkName, tokenAddress, request.url)
+    return getRequest(res, request.url, 'tokens-cache', graphCall)
```
Comment (zzmp) [inline functions/tokens/[[index]].ts:31]: eg

```suggestion
    return getRequest(res, request.url, 'tokens-cache', () => getToken(networkName, tokenAddress, request.url))
```

functions/utils/cache.test.ts
```diff
@@ -0,0 +1,30 @@
...
+
+beforeAll(() => {
+  const globalAny: any = global
+  globalAny.caches = {
+    open: async () => cacheMock,
+    ...cacheMock,
+  }
+})
+
+test('Should use cache properly', async () => {
```
zzmp, line 15: Have a separate test for `put` and `match` so it's easy to tell which if one of them is broken.

functions/utils/getRequest.ts
```diff
@@ -0,0 +1,33 @@
+import { MetaTagInjector } from '../components/metaTagInjector'
```
zzmp, line 1: Can you write tests for this file to ensure that things are called iff they should be?
When you think of your test cases, strive forfull code coverage: try to write a test for each branch of your conditionals.
To check that things are called iff they should be, use [.toHaveBeenCalled](https://jestjs.io/docs/expect#tohavebeencalled), and also install [jest-extended](https://www.npmjs.com/package/jest-extended) and use the [mock matchers](https://jest-extended.jestcommunity.dev/docs/matchers/mock/) (like .toHaveBeenCalledAfter).

iff == if and only if

zzmp (changes requested):

functions/utils/cache.ts
```diff
@@ -0,0 +1,25 @@
+class Cache {
+  async match(request: string) {
```
zzmp, line 2: You could make this its own type:

```
interface Data {
  title: string
  image: string
  url: string
}
```

```suggestion
  async match(request: string): { title: string; image: string; url: string } {
```

functions/utils/cache.ts
```diff
@@ -0,0 +1,25 @@
...
+    const cache = await caches.open('function-cache')
+    const response = await cache.match(request)
+    if (!response) return undefined
+    return JSON.parse(await response.text())
```
zzmp, line 6: Can you validate that `JSON.parse(await response.text())` matches the `Data` type, and throw if it does not?

functions/utils/getRequest.ts
```diff
@@ -0,0 +1,31 @@
...
+    | {
+        title: string
+        image: string
+        url: string
+      }
+    | undefined
+  >
+) {
+  try {
+    const cacheResponse = await Cache.match(url)
```
zzmp, line 17:
```suggestion
    const cachedData = await Cache.match(url)
```

to match `data` and make it obvious that they are the same type.

functions/utils/getRequest.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  await getRequest(Promise.resolve(new Response()), url, getData)
+  expect(Cache.match).toHaveBeenCalledWith(url)
+  expect(getData).toHaveBeenCalled()
+  expect(Cache.match).toHaveBeenCalledBefore(getData)
+  expect(Cache.put).toHaveBeenCalledAfter(getData)
+})
+
+test('getData should not be called when request is cached', async () => {
+  const url = 'https://example.com'
+  ;(Cache.match as jest.Mock).mockResolvedValueOnce({
```
zzmp, line 28:
```suggestion
  jest.mocked(Cache.match).mockResolvedValueOnce({
```

or if `jest.mocked` is not available (because we may be on an old version of jest) use the `mocked` (or copy it over) from our test-utils.

functions/utils/cache.test.ts
```diff
@@ -0,0 +1,42 @@
...
+beforeAll(() => {
+  const globalAny: any = global
+  globalAny.caches = {
+    open: async () => cacheMock,
+    ...cacheMock,
+  }
+})
+
+test('Should put cache properly', async () => {
+  //wait for the server to start
```
zzmp, line 16: These tests should not be server-dependent. You can instantiate your own `Response` objects inline, without any need for a server.

functions/utils/cache.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  expect(response).toBeUndefined()
+  const data = {
+    title: 'test',
+    image: 'testImage',
+    url: 'testUrl',
+  }
+  await Cache.put(data, 'https://example.com')
+})
+
+test('Should match cache properly', async () => {
```
zzmp, line 30: These tests should be isolated - you should reset the cache between them.

For the put test, put data in and then test that it is present.
For the match test, seed data and then test that it comes out.
You can mock match/put for the appropriate tests.

zzmp (changes requested):

functions/utils/cache.ts
```diff
@@ -0,0 +1,26 @@
...
+  url: string
+}
+
+class Cache {
+  async match(request: string): Promise<Data | undefined> {
+    const cache = await caches.open('function-cache')
+    const response = await cache.match(request)
+    if (!response) return undefined
+    const data: Data = JSON.parse(await response.text())
+    if (!data.title || !data.image || !data.url) throw new Error('Invalid data')
```
zzmp, line 13: Should we be throwing or returning undefined so that it can be re-fetched?

functions/utils/cache.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  globalAny.caches = {
+    open: async () => cacheMock,
+    ...cacheMock,
+  }
+})
+
+test('Should put cache properly', async () => {
+  jest.spyOn(cacheMock, 'put')
+  const response = await Cache.match('https://example.com')
+  expect(response).toBeUndefined()
```
zzmp, line 25:
```suggestion
```

functions/utils/cache.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  await Cache.put(data, 'https://example.com')
+  expect(cacheMock.put).toHaveBeenCalledWith('https://example.com', expect.anything())
```
zzmp, line 27: check that it's being called with a response that matches data.
you can do this by getting the actual arg and then parsing it: cacheMock.put.calls...

functions/utils/cache.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  const response2 = await Cache.match('https://example.com')
+  expect(response2).toStrictEqual(data)
```
zzmp, line 30:
```suggestion
  await expect(Cache.match('https://example.com').resolves.toStrictEqual(data)
```

functions/utils/cache.ts
zzmp: This could also be a test case, to ensure that it continue to do what we expect (either throw or return undefined).

zzmp (changes requested):

functions/utils/cache.ts
LunrEclipse: returning undefined makes more sense yeah!

functions/utils/cache.test.ts
LunrEclipse: what a 1 liner!

functions/utils/cache.test.ts
```diff
@@ -21,17 +21,24 @@ beforeAll(() => {
 
 test('Should put cache properly', async () => {
   jest.spyOn(cacheMock, 'put')
-  const response = await Cache.match('https://example.com')
-  expect(response).toBeUndefined()
+  jest.spyOn(cacheMock, 'match')
```
zzmp, line 24:
```suggestion
```

functions/utils/cache.ts
```diff
@@ -6,19 +6,19 @@ interface Data {
 
 class Cache {
   async match(request: string): Promise<Data | undefined> {
-    const cache = await caches.open('function-cache')
+    const cache = await caches.open('functions-cache')
```
zzmp, line 9: Pull the cache name into a static variable:

```
const CACHE_NAME = 'functions-cache' as const
```

zzmp (approved):
last nits then this is 🌋

zzmp (approved):

PR #6930, opened 2023-07-13
feat: parallelize cache

LunrEclipse:
## Description

Parallelize cache within cloud functions.

### Automated testing

- [x] Unit test N/A


functions/nfts/asset/[[index]].ts
```diff
@@ -6,22 +6,22 @@ export const onRequest: PagesFunction = async ({ params, request, next }) => {
...
   const assetPromise = getAsset(collectionAddress, tokenId, request.url)
   const resPromise = next()
+  const cachePromise = cache.match(request.url)
   try {
-    const cache = caches.default
-    const response = await cache.match(request.url)
-    if (response) {
-      const data = JSON.parse(await response.text())
-      return new HTMLRewriter().on('head', new MetaTagInjector(data)).transform(await resPromise)
+    const [graphData, cacheResponse, res] = await Promise.all([assetPromise, cachePromise, resPromise])
```
Comment (zzmp) [inline functions/nfts/asset/[[index]].ts:14]: This will always wait for `graphData` to resolve, even if it is already cached.

Instead, you should only call `graphData` if it is not yet cached. You could construct a helper utility (that you could use across all of these files) that - given a url and getter, tries the cache and then the network:

```
async function matchThenFetch<T>(url: string, fetcher: () => T) {
  const match = await caches.default.match(url)
  return match ? match : fetcher()
}
```

zzmp (changes requested):
You should include a unit test to make sure you are not fetching a cached asset.

functions/nfts/asset/[[index]].ts
Comment (LunrEclipse) [reply functions/nfts/asset/[[index]].ts:14]: Could I also just take it out of the parallelization?

zzmp (changes requested):
I think this is pretty coupled to the cache. Would you be ok including this in the cache PR and closing this one?

PR #6949, opened 2023-07-14
test: add signatureToActivity undefined tests

natew:
nitty tests


src/components/AccountDrawer/MiniPortfolio/Activity/parseLocal.test.ts
```diff
@@ -492,4 +494,37 @@ describe('parseLocalActivity', () => {
       from: mockAccount2,
     })
   })
+
+  it('Signature to activity - returns undefined if not UniswapX order', () => {
```
tinaszheng, line 498: ooh you need to delete this case since we got rid of the placeholder type

tinaszheng (approved):

PR #6950, opened 2023-07-14
fix: function defined twice

tinaszheng:


src/state/signatures/hooks.ts
```diff
@@ -59,10 +59,6 @@ export function useAddOrder() {
   )
 }
 
-function isPendingOrder(signature: SignatureDetails): signature is UniswapXOrderDetails {
```
tinaszheng, line 62: defined on line 74

cbachmeier (approved):

PR #6952, opened 2023-07-14
fix: pwat stack overflow

JackShort:
* the nft pwat was crashing and it was due to a stack overflow error that seemed to be coming from here


cbachmeier (approved):
Verified fix by selecting to pay with alternate currency and page does not crash
Screenshot 2023-07-14 at 2 14 12 PM

just-toby (approved):

cbachmeier:
Original error message: `Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.`

tinaszheng (approved):

PR #6953, opened 2023-07-14
chore: some last nits

zzmp:
- use codebase-standard base type
- add test for uniswapx test hook
- simplify useAllSignatures usages


tinaszheng (approved):
🙌

PR #6956, opened 2023-07-15
fix: display token images on first pageload

zzmp:
## Description

Fixes token selector images such that they are always visible on first pageload.

Token images were not being displayed because tokens with (or with unknown) safety warning statuses only display their symbols, and the `TokenSafetyLookupTable` was being initialized on pageload (triggered by the initial WETH token) before the token list was loaded. This meant that - for new users - on first load, the safety status was unknown, and the image was not displayed.
I've fixed this by updating the `TokenSafetyLookupTable` when a token list is successfully fetched, so that it remains up-to-date. This means that if tokens are displayed in the Token Selector the `TokenSafetyLookupTable` should always be initialized, because the Token Selector is populated from a token list.

For cases where the token list fails to load, I've added special-casing for common bases, so that they will always display images.

For initial load, before the token list has loaded, I've added special-casing for native (wrapped) tokens, so that the initial screen always displays the ETH image. Other tokens may still flash the symbol on pageload if the page is first navigated to using query parameters to select a token.

Lastly, for cases when a token image is incorrectly specified, I've updated the `backupImg` to be included in the first render.

As a whole, these changes should ensure that token images that _should_ show, _do_ show.

NB: This has likely been a long-standing bug, present since the `TokenSafetyLookupTable` was introduced to the codebase.

_Slack thread:_ https://uniswapteam.slack.com/archives/C03NZGRTS3A/p1689445894451209

### Reproducing the error

1. Clear localStorage: `localStorage.clear()`
2. Reload the page
- NB: This is equivalent to opening the page in an incognito session
- Observe that ETH has no image
4. Open the Token Selector
- Observe that no tokens have an image

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test


src/constants/tokenSafety.tsx
```diff
@@ -3,8 +3,9 @@ import { TokenStandard } from 'graphql/data/__generated__/types-and-hooks'
 import { SearchToken } from 'graphql/data/SearchTokens'
 
 import { ZERO_ADDRESS } from './misc'
+import { COMMON_BASES } from './routing'
 import { NATIVE_CHAIN_ID } from './tokens'
-import WarningCache, { TOKEN_LIST_TYPES } from './TokenSafetyLookupTable'
+import TokenSafetyLookupTable, { TOKEN_LIST_TYPES } from './TokenSafetyLookupTable'
```
cartcrom, line 8: Importing the default w/ same name as the class is confusing to me, makes it look like you're accessing static methods of the class rather calling methods of an instance of it. Exporting as non-default here probably is the better option

src/constants/TokenSafetyLookupTable.ts
```diff
@@ -14,40 +14,39 @@ export enum TOKEN_LIST_TYPES {
 }
 
 class TokenSafetyLookupTable {
-  dict: { [key: string]: TOKEN_LIST_TYPES } | null = null
+  dict: { [key: string]: TOKEN_LIST_TYPES } = {}
 
-  createMap() {
-    const dict: { [key: string]: TOKEN_LIST_TYPES } = {}
+  constructor() {
+    this.update()
```
cartcrom, line 20: I originally received feedback that I _shouldn't_ do it this way because of pageload time. Is the perf okay here?

zzmp: I backed this out because it was causing a lot of tests to fail.
Practically, this happens at pageload anyway, because a token logo is nearly always displayed.

src/constants/tokenSafety.tsx
zzmp: The class is never actually exported, only a singleton, so the filename is misleading.

zzmp: Updated to `tokenLogoLookup`, `tokenSafetyLookup`.

tinaszheng (approved):

src/hooks/useAssetLogoSource.ts
```diff
@@ -61,25 +66,25 @@ export default function useAssetLogoSource(
   isNative?: boolean,
   backupImg?: string | null
 ): [string | undefined, () => void] {
-  const hasWarning = Boolean(address && checkWarning(address))
+  const showLogo = Boolean((address && checkWarning(address, chainId) === null) || isNative)
```
just-toby, line 69: should we add a check here to see if the token is one of the common bases besides native?

just-toby (approved):
lgtm

src/hooks/useAssetLogoSource.ts
zzmp: I added that logic to checkWarning so the common bases should always have their image displayed.

PR #6959, opened 2023-07-17
fix: log opt in impression

just-toby:
`shouldLogImpression={isVisible}` doesn't work because the Trace component doesn't log an impression when the prop changes.

this simplifies the logic by making `shouldLogImpression` always true, so an impression logs every time the UI renders

### DIFF
--- a/src/pages/Swap/UniswapXOptIn.tsx
+++ b/src/pages/Swap/UniswapXOptIn.tsx
@@ -111,40 +111,38 @@ const OptInContents = ({
 
   const containerRef = useRef<HTMLDivElement>()
 
-  const wrapTrace = (children: JSX.Element) => {
+  if (isSmall) {
     return (
       <Trace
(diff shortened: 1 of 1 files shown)


tinaszheng (approved):

tinaszheng:
@zzmp can we hot-fix this one onto prod?

src/pages/Swap/UniswapXOptIn.tsx
```diff
@@ -111,40 +111,38 @@ const OptInContents = ({
 
   const containerRef = useRef<HTMLDivElement>()
 
-  const wrapTrace = (children: JSX.Element) => {
+  if (isSmall) {
```
zzmp, line 114: I think you want `<Trace>` to enclose everything, and put the conditional within it, so that you don't double-log if a user modifies their screen size.

zzmp (changes requested):

PR #6962, opened 2023-07-17
fix: Silence linter warnings for GA

cbachmeier:
## Description

- the eslint disable command needed to be updated to properly silence and not spam PRs with `Unchanged files with check annotations`
- you're welcome


zzmp (approved):
I take it back I thought at first this was a global silence but now I see it is specific to GA.

natew (approved):

PR #6966, opened 2023-07-17
fix: format currency amounts as strings in logging

just-toby:
## Description

fixes the format of the values for `token_in_amount_max` and `token_out_amount_min` so that they don't create new properties in the aplitude backend. they should be formatted as strings when sent to amplitude.

## Screen capture

image

### DIFF
--- a/src/lib/utils/analytics.ts
+++ b/src/lib/utils/analytics.ts
@@ -93,7 +93,7 @@ export const formatSwapQuoteReceivedEventProperties = (
     swap_quote_block_number: isClassicTrade(trade) ? trade.blockNumber : undefined,
     swap_quote_received_timestamp: swapQuoteReceivedDate.getTime(),
     allowed_slippage_basis_points: formatPercentInBasisPointsNumber(allowedSlippage),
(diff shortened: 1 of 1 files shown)


LunrEclipse (approved):

src/lib/utils/analytics.ts
```diff
@@ -93,7 +93,7 @@ export const formatSwapQuoteReceivedEventProperties = (
     swap_quote_block_number: isClassicTrade(trade) ? trade.blockNumber : undefined,
     swap_quote_received_timestamp: swapQuoteReceivedDate.getTime(),
     allowed_slippage_basis_points: formatPercentInBasisPointsNumber(allowedSlippage),
-    token_in_amount_max: trade.maximumAmountIn(allowedSlippage),
-    token_out_amount_min: trade.minimumAmountOut(allowedSlippage),
+    token_in_amount_max: trade.maximumAmountIn(allowedSlippage).toFixed(2),
```
zzmp, line 96: Why not toExact?

just-toby: yeah that works too

zzmp (approved):

PR #6981, opened 2023-07-19
fix: display ens name/avatar on other chains

LunrEclipse:
## Description

Fixed issue where ENS information would not display on any non-Mainnet chain.

The reason for this stemmed from the fact that we were using the wallet's provider to gather this information. As a result, when the user switched chains, we would of course be using a provider/RPC for another chain in hopes of getting information from an Ethereum contract resulting in failure. In order to bypass this, we can use our own RPC to fetch this information when the user is on a non-eth chain.

Created contract variable called `useEthOnlyContract` that uses the user's provider, unless we are on another chain, in which case we use our own RPC Provider. From there, also created a permanent `multicall.Updater` that always runs on mainnet to fetch this information and cache it for the first time. Utilizes `NEVER_RELOAD` listener option to ensure that information isn't refetched and wasting our own RPC calls. 

_Linear ticket:_ WEB-1643

### Before

Screenshot 2023-07-19 at 2 40 39 PM

### After
Screenshot 2023-07-19 at 2 39 01 PM

### QA (ie manual testing)

- [x] Testing normal functions to ensure cadence of other data collection is not affected 

### Automated testing

- [x] Unit test N/A


src/lib/hooks/useBlockNumber.tsx
```diff
@@ -63,6 +71,12 @@ export function BlockNumberProvider({ children }: { children: ReactNode }) {
           console.error(`Failed to get block number for chainId ${activeChainId}`, error)
         })
 
+      if (activeChainId === ChainId.MAINNET) {
+        setEthBlockNumber(block)
+      } else {
+        RPC_PROVIDERS[ChainId.MAINNET].getBlockNumber().then(setEthBlockNumber)
```
natew, line 77: any reason not to use async await here?

src/lib/state/multicall.tsx
```diff
@@ -33,20 +33,38 @@ function getBlocksPerFetchForChainId(chainId: number | undefined): number {
...
-      listenerOptions={listenerOptions}
-    />
+    <>
+      <multicall.Updater
+        chainId={ChainId.MAINNET}
+        latestBlockNumber={latestEthBlockNumber}
+        contract={ethContract}
+        listenerOptions={ethListener}
+      />
+      {chainId !== ChainId.MAINNET && (
```
natew, line 60: should this render both multicalls? intuitively we're just on one chain at a time so i'd think its a ternary. 

also in general i think its good to do Multicall caps style for any react components

natew (approved):

src/lib/state/multicall.tsx
LunrEclipse: yeah pretty much if we're on another chain (i.e. Polygon), we want a multicall for polygon to do the normal fetches as well as one for Mainnet to fetch ENS information.

Also wdym by caps style 😅

src/lib/hooks/useBlockNumber.tsx
LunrEclipse: We could but then we need to make the entire useEffect function async. Also I was just following the structure used above on line 65

src/lib/hooks/multicall.ts
```diff
@@ -22,6 +23,11 @@ export function useSingleCallResult(...args: SkipFirstTwoParams<typeof multicall
   return multicall.hooks.useSingleCallResult(chainId, latestBlock, ...args)
 }
 
+export function useEthSingleCallResult(...args: SkipFirstTwoParams<typeof multicall.hooks.useSingleCallResult>) {
+  const { latestEthBlock } = useCallContext()
```
zzmp, line 27: You don't need to call `useCallContext`, you can just call `useEthBlockNumber` - adding that to `useCallContext` is overkill and subscribes to the block updates unnecessarily.

src/lib/hooks/useBlockNumber.tsx
```diff
@@ -7,6 +9,7 @@ const BlockNumberContext = createContext<
   | {
       value?: number
       fastForward(block: number): void
+      ethValue?: number
```
zzmp, line 12: `mainnetValue`

src/lib/hooks/useBlockNumber.tsx
```diff
@@ -28,9 +31,14 @@ export function useFastForwardBlockNumber(): (block: number) => void {
...
 }
 
+export function useEthBlockNumber(): number | undefined {
+  return useBlockNumberContext().ethValue
+}
+
 export function BlockNumberProvider({ children }: { children: ReactNode }) {
   const { chainId: activeChainId, provider } = useWeb3React()
   const [{ chainId, block }, setChainBlock] = useState<{ chainId?: number; block?: number }>({ chainId: activeChainId })
+  const [ethBlockNumber, setEthBlockNumber] = useState<number | undefined>(undefined)
```
zzmp, line 41: `mainnetBlockNumber`

src/lib/state/multicall.tsx
```diff
@@ -33,20 +33,38 @@ function getBlocksPerFetchForChainId(chainId: number | undefined): number {
...
+  const latestEthBlockNumber = useEthBlockNumber()
   const contract = useInterfaceMulticall()
+  const ethContract = useEthOnlyInterfaceMulticall()
   const listenerOptions: ListenerOptions = useMemo(
     () => ({
       blocksPerFetch: getBlocksPerFetchForChainId(chainId),
     }),
     [chainId]
   )
+  const ethListener: ListenerOptions = useMemo(
```
zzmp, line 45: `mainnetListener`

src/lib/state/multicall.tsx
zzmp: `multicall` is lower-cased because it comes from `createMulticall` - the component is still upper-cased (`multicall.Updater`) - is that what you mean?

src/lib/hooks/useBlockNumber.tsx
zzmp, line 77: I don't think you need to invoke this at all when `activeChainId !== ChainId.MAINNET`.
wdyt about just removing this line?

src/lib/hooks/useBlockNumber.tsx
```diff
@@ -63,6 +71,12 @@ export function BlockNumberProvider({ children }: { children: ReactNode }) {
           console.error(`Failed to get block number for chainId ${activeChainId}`, error)
         })
 
+      if (activeChainId === ChainId.MAINNET) {
+        setEthBlockNumber(block)
```
zzmp, line 75: This has no guard against backslipping. Could you instead incorporate this into `onBlock`, something like

```
if (chainBlock.chainId === ChainId.MAINNET) {
  ...
}
```

src/lib/hooks/useBlockNumber.tsx
zzmp, line 41: I think you can incorporate this into the existing state, and that it shouldn't be separated. Can you augment the chainBlock state to include `mainnetBlock`?

zzmp (changes requested):
Are there any existing tests for this?
If not, can you add an integration test for ENS on Polygon?

src/lib/hooks/useBlockNumber.tsx
LunrEclipse: i think the `else` is still required if we load initially onto another chain. I just tested it and it works fine if we connect to `mainnet` first and then switch chains, but if we for example connect to `Polygon` first then it doesn't load the ENS

src/lib/hooks/useBlockNumber.tsx
```diff
@@ -62,7 +84,11 @@ export function BlockNumberProvider({ children }: { children: ReactNode }) {
         .catch((error) => {
           console.error(`Failed to get block number for chainId ${activeChainId}`, error)
         })
-
+      if (mainnetBlock === undefined) {
```
zzmp, line 87: I like that you are only updating this if it is undefined. Can you precede this line with a comment explaining why this is only necessary when it is `undefined`?

src/lib/hooks/useBlockNumber.tsx
```diff
@@ -81,8 +107,9 @@ export function BlockNumberProvider({ children }: { children: ReactNode }) {
           setChainBlock({ chainId: activeChainId, block: update })
         }
       },
+      mainnetValue: chainId === ChainId.MAINNET ? block : mainnetBlock,
```
zzmp, line 110: Why wouldn't this be up-to-date?

```suggestion
      mainnetValue: mainnetBlock,
```

zzmp (changes requested):
last bit of cleanup in blocknumber

src/lib/hooks/useBlockNumber.tsx
```diff
@@ -28,16 +31,26 @@ export function useFastForwardBlockNumber(): (block: number) => void {
...
   const onBlock = useCallback(
     (block: number) => {
       setChainBlock((chainBlock) => {
         if (chainBlock.chainId === activeChainId) {
           if (!chainBlock.block || chainBlock.block < block) {
-            return { chainId: activeChainId, block }
+            return { chainId: activeChainId, block, mainnetBlock: chainBlock.mainnetBlock }
           }
         }
         return chainBlock
```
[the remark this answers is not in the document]
LunrEclipse: yup!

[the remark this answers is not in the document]
LunrEclipse: i just had to update line 71-75 to keep the mainnetBlock upon chain swap instead of clearing state

cypress/e2e/mini-portfolio/accounts.test.ts
```diff
@@ -76,4 +76,37 @@ describe('Mini Portfolio account drawer', () => {
         })
     })
   })
+
+  it('fetches ENS name', () => {
+    cy.hardhat().then((hardhat) => {
+      const account = hardhat.wallets[0].address
```
zzmp, line 82: lint:

```suggestion
```

zzmp (approved):
🎆

zzmp (changes requested):
I think you still need to request the chainBlock for `MAINNET` on load in `useBlockNumber` so that ENS will function if the page is loaded on an alternate chain.

zzmp (approved):

src/lib/hooks/useBlockNumber.tsx
```diff
@@ -99,9 +97,7 @@ export function BlockNumberProvider({ children }: { children: ReactNode }) {
         .then((block) => {
           onChainBlock(ChainId.MAINNET, block)
         })
-        .catch((error) => {
-          console.error(`Failed to get block number for chainId ${ChainId.MAINNET}`, error)
-        })
+        .catch()
```
zzmp, line 100: retain this console.error - it shouldn't be blocking tests, as it was there before.

src/lib/hooks/useBlockNumber.tsx
```diff
@@ -77,9 +77,7 @@ export function BlockNumberProvider({ children }: { children: ReactNode }) {
         .then((block) => {
           if (!stale) onChainBlock(activeChainId, block)
         })
-        .catch((error) => {
-          console.error(`Failed to get block number for chainId ${activeChainId}`, error)
-        })
+        .catch()
```
zzmp, line 80: I think you may need something in the catch block for it to work (although I'm unsure):

```suggestion
        .catch(() => undefined)
```

zzmp (changes requested):

src/lib/hooks/useBlockNumber.tsx
```diff
@@ -71,7 +88,18 @@ export function BlockNumberProvider({ children }: { children: ReactNode }) {
...
+  }, [activeChainId, provider, windowVisible, onChainBlock])
+
+  useEffect(() => {
+    if (mainnetBlock === undefined) {
+      RPC_PROVIDERS[ChainId.MAINNET]
+        .getBlockNumber()
+        .then((block) => {
+          onChainBlock(ChainId.MAINNET, block)
+        })
+        .catch()
```
zzmp, line 100: You something in the catch block for it to work:

```suggestion
        // swallow errors - it's ok if this fails, as we'll try again if we activate mainnet
        .catch(() => undefined)
```

zzmp (changes requested):

zzmp (approved):

src/lib/hooks/useBlockNumber.tsx
```diff
@@ -28,41 +31,57 @@ export function useFastForwardBlockNumber(): (block: number) => void {
...
+      )
 
       provider
         .getBlockNumber()
         .then((block) => {
-          if (!stale) onBlock(block)
+          if (!stale) onChainBlock(activeChainId, block)
         })
-        .catch((error) => {
+        .catch(() => {
```
zzmp, line 80:
```suggestion
        .catch((error) => {
```

zzmp (approved):

src/lib/state/multicall.tsx
natew: Ah ok seems fine, I tend to like just keeping things uppercase when used as components but its a personal perference / nit, could de-structure it above but not worth the effort

natew (approved):

PR #6982, opened 2023-07-19
feat: use router depending on the origin

grabbou:
Ultimately we want to use BrowserRouter and path-based routing for Uniswap. Unfortunately, IPFS does not support catch-all, which means all paths different than `/` would not hit our `index.html` and would result in a 404 error as such.

This PR brings in a runtime workaround where router is chosen depending on the origin.

This change is transparent, since all routing happens via `<Link />` abstraction that deals with URLs. 

## Where do we turn this on

Everywhere but on IPFS. We treat IPFS deployment target as every target where `isProductionEnv` is true, but it does not match own origin.

## Places where hashes were left

NFT codebase (generating tweet URIs) still relies on `#` because it would make too much unnecessary effort to refactor to hooks and handle potentially undefined values. I also have some predictions that logic for generating tweet URIs is flawed, will submit a follow-up Linear ticket to track that.

Updating that path as a part of this PR would mean fixing those issues and doing testing too. This PR on its own is serious enough that I would like to keep it separate.

> It is okay to still leave references to hashes and some old URLs because we are redirecting anyway. To do so, you can test that by going to `/#/any-url` DIRECTLY from the browser URL and noticing that redirect works.

## Nested paths 

As per https://github.com/facebook/create-react-app/issues/527#issuecomment-244793467, package.json was updated by removing `"homepage": "."`.

I believe [it was incorrectly set from the very beginning](https://github.com/Uniswap/interface/commit/d4911d1054403e6dac20185a1498b25059915bdd) - it is not recommended to configure it unless your app is expected to be running at different paths at the same time, e.g.:
- /
- /app
and both `/` as well as `/app` would have the `index.html` and `assets` available. Configuring it to "." simply instructs CRA to build asset URLs relative to index.html.

In our case, we don't want this, because:
- /tokens/ethereum

Is now going to point to the root. Without this update, all we would see is blank page, as the browser would look for assets under "/tokens/ethereum/static" path. This is obviously not going to work for client-side-routing.

Note this is happening on main at the time of writing and is not related to the work in this PR. We still want this setting to be removed, even if we would switch back to Hash Router one day.  

The reason is that, right now, in production (app.uniswap.org), all requests are routed to the root index.html anyway, regardless of the path:
- `/` - works
- `/some/random/path/#` - loads index.html with inline CSS, but has client-side error and renders nothing

Screenshot 2023-07-31 at 13 05 17

With this fixed applied, but without turning BrowserRouter (path-based router) on, `/some/random/path/#` would still work and potentially let us redirect to `/#` (subject to a follow-up PR).

> To reproduce, navigate to any nested URL, such as /tokens/ethereum and RELOAD the app. Reloading w/o this change would result in an empty background.


src/pages/App.tsx
```diff
@@ -209,7 +217,16 @@ export default function App() {
             <Suspense fallback={<Loader />}>
               {isLoaded ? (
                 <Routes>
-                  <Route path="/" element={<Landing />} />
+                  <Route
+                    path="/"
+                    element={
+                      // if react-router-dom matches "/" with window.location.hash defined, it means that we're
+                      // using BrowserRouter and can safely redirect to a path route
+                      <RedirectHashToPath>
```
zzmp, line 225: Why is this redirect necessary?

grabbou: If there's an old URL somewhere out there e.g. on Twitter with a hashtag, e.g. `uniswap.org/#/swap`, this would match the home page when BrowserRouter is used. Redirect will make sure that it is redirected to `uniswap.org/swap` before rendering home page.

zzmp (approved):

src/utils/env.ts
```diff
@@ -23,6 +23,14 @@ export function isAppUniswapStagingOrg({ hostname }: { hostname: string }): bool
   return hostname === 'app.uniswap-staging.org'
 }
 
+export function isBrowserRouterEnabled(): boolean {
+  return (
+    isAppUniswapOrg(window.location) ||
+    isAppUniswapStagingOrg(window.location) ||
+    window.location.hostname === 'localhost'
```
cbachmeier, line 30: I think we have an `isDev` node env hook to replace checking for localhost

zzmp (changes requested):
You can remove the hash handling in src/tracing, and in cypress.visit in the cypress tests.
some other instances (I just searched for `/#`)

zzmp (changes requested):
Our assets are loaded from a relative path. They must be loaded from an absolute path, or they will error with more than one path segment.

src/utils/env.ts
grabbou: Theoretically yes, but what if we push it to some external location in DEV mode (for whatever reason) and it breaks? Kind of an edge case, but just thinking out loud.

To be honest, I was thinking to invert this. Rather than checking for Uniswap/hostname, localhost and Vercel (clearly a lot of environments), check if it's IPFS instead. That would require just one check.

grabbou:
Pushed some updates as discussed, mainly removing remaining # and replacing with `useHref`. Required refactoring of a few methods into hooks.

Next up is addressing the issue raised by @zzmp which is resources not loading for nested routes.

src/utils/env.ts
grabbou: Okay, I think I will replace with `isDev` instead, b/c all IPFS builds won't have that flag anyway. Will also add Vercel to the list.

src/nft/components/collection/Activity.tsx
```diff
@@ -138,9 +137,11 @@ export const Activity = ({ contractAddress, rarityVerified, collectionName, chai
                 (event, i) =>
                   event.eventType && (
                     <Box
```
cbachmeier, line 139: we'll need to update the styling to get rid of the link underlines introduced by this change https://interface-git-router-new-uniswap.vercel.app/nfts/collection/0xbc4ca0eda7647a8ab7c2061c2e118a18a936f13d/activity
Screenshot 2023-08-01 at 11 15 46 AM

cbachmeier: compared to staging https://app.corn-staging.com/#/nfts/collection/0xbc4ca0eda7647a8ab7c2061c2e118a18a936f13d/activity
Screenshot 2023-08-01 at 11 17 08 AM

src/nft/components/card/containers.tsx
```diff
@@ -95,7 +95,7 @@ const ActionButton = ({
     <StyledActionButton
       selected={isSelected}
       isDisabled={isDisabled}
-      onClick={(e: React.MouseEvent) => (isDisabled ? undefined : clickActionButton(e))}
+      onClick={(e) => (isDisabled ? undefined : clickActionButton(e))}
```
cbachmeier, line 98: why was this typing removed?

cbachmeier (changes requested):
Found a couple more locations:
- README
  - ln72-75
- `src/nft/utils/asset.tsx`
  - ln92 & ln125
-  `src/pages/Landing/__snapshots__/index.test.tsx.snap`
  - I think you just need to update this snapshot test

src/nft/components/collection/Activity.tsx
```diff
@@ -138,9 +137,11 @@ export const Activity = ({ contractAddress, rarityVerified, collectionName, chai
...
-                      as="a"
+                      as={Link}
```
zzmp, line 140: Why is this changed to `Link` from `a`?

src/components/FiatOnrampModal/index.tsx
```diff
@@ -98,7 +101,7 @@ export default function FiatOnrampModal() {
           theme: isDarkMode ? 'dark' : 'light',
           colorCode: theme.accentAction,
           defaultCurrencyCode: 'eth',
-          redirectUrl: 'https://app.uniswap.org/#/swap',
+          redirectUrl: swapUrl,
```
zzmp, line 104: swapUrl needs to be added to the deps array. Why can't the path be used without the full url?

zzmp (changes requested):

src/nft/components/card/containers.tsx
grabbou: It is unnecessary - the type is inferred from the event handler

src/nft/components/collection/Activity.tsx
grabbou: Good catch. I'll take a look at that.

src/nft/components/collection/Activity.tsx
grabbou: Using `<Link />` lets us always refer to a path (`/`) and will use hashes or paths depending on deployment. Without it, and by relying on `<a />`, we would have to handle that case ourselves.

grabbou: In principle, we should always rely on `<link />` where possible.

src/utils/env.ts
```diff
@@ -23,6 +23,21 @@ export function isAppUniswapStagingOrg({ hostname }: { hostname: string }): bool
   return hostname === 'app.uniswap-staging.org'
 }
 
+export function isBrowserRouterEnabled(): boolean {
+  if (isProductionEnv()) {
+    if (
+      isAppUniswapOrg(window.location) ||
+      isAppUniswapStagingOrg(window.location) ||
+      // Cypress tests
+      window.location.hostname === 'localhost'
```
zzmp, line 32: Move this to its own (not exported) function:

```
function isLocalhost({ hostname }: { hostname: string }): boolean {
  return hostname === 'localhost'
}
```

grabbou:
> Found a couple more locations:
> 
> * README
>   
>   * ln72-75
> * `src/nft/utils/asset.tsx`
>   
>   * ln92 & ln125
> * `src/pages/Landing/__snapshots__/index.test.tsx.snap`
> * I think you just need to update this snapshot test

@cbachmeier handled everything but `utils/assets.tsx` - check the "places where hashes were left" in the main message for details.

src/components/FiatOnrampModal/index.tsx
grabbou: Added, although it's not required (it won't change throughout the lifetime of the app)

grabbou:
@cbachmeier thanks for the catch! Fixed!
Screenshot 2023-08-02 at 15 37 39

src/nft/components/collection/Activity.tsx
grabbou: Addressed in a follow-up comment

zzmp (approved):

grabbou:
Follow-up issue: https://linear.app/uniswap/issue/WEB-2606/update-nft-twitter-embed-links-to-use-paths-instead-of-hashes

cbachmeier (approved):

zzmp (approved):

zzmp (changes requested):
This lgtm and I love it but I'm "Request changes"ing so it doesn't get merged before this week's staging cut. We'd like it to go in next week to reduce risk.

I'll switch back to "Approve" once staging is cut.

grabbou:
Please @zzmp merge at your earliest convenience.

zzmp (approved):

PR #6983, opened 2023-07-19
feat: add opt out analytics

JackShort:
* Adds toggle to opt out of analytics
* adds wrappers around analytics functions to check atom value if analytics should be sent
Screen Shot 2023-07-19 at 4 47 00 PM

_I am sorry for the number of files this touches - but it if makes you feel better it sucked for me too_


src/components/AccountDrawer/SmallBalanceToggle.tsx
```diff
@@ -1,30 +1,21 @@
-import { Trans } from '@lingui/macro'
-import Row from 'components/Row'
-import Toggle from 'components/Toggle'
+import { t } from '@lingui/macro'
```
zzmp, line 1: nit: If you're trying to get it in faster, this refactor could be a separate PR.

src/components/analytics/TraceAnalytics.tsx
```diff
@@ -0,0 +1,35 @@
+import { sendAnalyticsEvent, Trace, TraceEvent } from '@uniswap/analytics'
+import { atomWithStorage, useAtomValue } from 'jotai/utils'
+import { memo } from 'react'
+
+const analyticsOptOutAtomName = 'optOutOfAnalytics'
+export const analyticsOptOutToggle = atomWithStorage<boolean>(analyticsOptOutAtomName, true)
+
+export const TraceAnalytics = memo((props: React.ComponentProps<typeof Trace>) => {
```
zzmp, line 8: This should be called `AnalyticsTrace`, not `TraceAnalytics`, following the filepath.

src/components/analytics/TraceAnalytics.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const allowAnalytics = useAtomValue(analyticsOptOutToggle)
+
+  return <Trace {...props} shouldLogImpression={allowAnalytics} />
+})
+
+TraceAnalytics.displayName = 'TraceAnalytics'
+
+export const TraceAnalyticsEvent = memo((props: React.ComponentProps<typeof TraceEvent>) => {
```
zzmp, line 16: This should be called `AnalyticsTraceEvent`, not `TraceAnalyticsEvent`, following the filepath. Alternatively, it could be called `AnalyticsEvent`.

src/components/analytics/TraceAnalytics.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const allowAnalytics = useAtomValue(analyticsOptOutToggle)
+
+  return <TraceEvent {...props} shouldLogImpression={allowAnalytics} />
+})
+
+TraceAnalyticsEvent.displayName = 'TraceAnalyticsEvent'
+
+export const sendOptOutAnalyticsEvent: typeof sendAnalyticsEvent = (event, properties) => {
```
zzmp, line 24: the "OptOut" is an implementation detail to the consumers of this file, so it should be omitted. Export this as `sendAnalyticsEvent`.

src/components/analytics/TraceAnalytics.tsx
```diff
@@ -0,0 +1,35 @@
+import { sendAnalyticsEvent, Trace, TraceEvent } from '@uniswap/analytics'
+import { atomWithStorage, useAtomValue } from 'jotai/utils'
+import { memo } from 'react'
+
+const analyticsOptOutAtomName = 'optOutOfAnalytics'
+export const analyticsOptOutToggle = atomWithStorage<boolean>(analyticsOptOutAtomName, true)
```
zzmp, line 6: Have we used other atoms with storage, or is this a new pattern? If we're using localStorage, should this potentially go through redux instead, to avoid writing to localStorage from multiple libraries?

src/components/analytics/TraceAnalytics.tsx
```diff
@@ -0,0 +1,35 @@
...
+export const TraceAnalyticsEvent = memo((props: React.ComponentProps<typeof TraceEvent>) => {
+  const allowAnalytics = useAtomValue(analyticsOptOutToggle)
+
+  return <TraceEvent {...props} shouldLogImpression={allowAnalytics} />
+})
+
+TraceAnalyticsEvent.displayName = 'TraceAnalyticsEvent'
+
+export const sendOptOutAnalyticsEvent: typeof sendAnalyticsEvent = (event, properties) => {
+  const value = localStorage.getItem(analyticsOptOutAtomName)
```
zzmp, line 25: Instead of referencing localStorage every time, could this be stored in-memory? Does it matter - ie, is accessing localStorage slow?

zzmp (changes requested):
In order to prevent a regression, can you re-export `useTrace` from `components/analytics/TraceAnalytics`, and make a lint rule forbidding import of `@uniswap/analytics` except in `components/analytics/TraceAnalytics`?

I would also move the file from `components/analytics/TraceAnalytics` to `analytics`, as it is not just components, and it is very tightly scoped.

src/components/AccountDrawer/SmallBalanceToggle.tsx
```diff
@@ -1,30 +1,21 @@
...
-            updateHideSmallBalances(!hideSmallBalances)
-          }}
-        />
-      </Row>
-    </Row>
+    <SettingsToggle
+      title={t`Hide small balances`}
+      isActive={hideSmallBalances}
+      toggle={() => {
+        updateHideSmallBalances(!hideSmallBalances)
```
zzmp, line 17:
```suggestion
        updateHideSmallBalances((value) => !value)
```

src/components/AccountDrawer/SmallBalanceToggle.tsx
```diff
@@ -1,30 +1,21 @@
...
+      }}
```
zzmp, line 18:
```suggestion
      toggle={() => void updateHideSmallBalances((value) => !value)}
```

src/components/AccountDrawer/TestnetsToggle.tsx
```diff
@@ -1,31 +1,22 @@
...
-        />
-      </Row>
-    </Row>
+    <SettingsToggle
+      title={t`Show testnets`}
+      dataid="testnets-toggle"
+      isActive={showTestnets}
+      toggle={() => {
+        updateShowTestnets(!showTestnets)
+      }}
```
zzmp, line 19:
```suggestion
      toggle={() => void updateShowTestnets((value) => !value)}
```

src/components/AccountDrawer/AnalyticsToggle.tsx
```diff
@@ -0,0 +1,20 @@
...
+  const [allowAnalytics, updateAllowAnalytics] = useAtom(analyticsOptOutToggle)
+
+  return (
+    <SettingsToggle
+      title={t`Allow analytics`}
+      description={t`We use anonymized data to enhance your experience with Uniswap Labs products.`}
+      isActive={allowAnalytics}
+      toggle={() => {
+        updateAllowAnalytics(!allowAnalytics)
+      }}
```
zzmp, line 17: This may cause race conditions if you double-click it. Instead, negate the current value - as opposed to the value in closure:

```suggestion
      toggle={() => voidupdateAllowAnalytics((value) => !value)}
```

src/components/analytics/TraceAnalytics.tsx
JackShort: its like a 10 thousandth of a millisecond to access local storage, so def not slow. The way to access this in-memory would be to use the useAtomValue hook in the component that calls this function then passing the value down, but that kinda defeats the purpose of this function

src/analytics/index.tsx
```diff
@@ -0,0 +1,41 @@
+import {
+  sendAnalyticsEvent as sendAnalyticsTraceEvent,
+  Trace as AnalyticsTrace,
+  TraceEvent as AnalyticsEvent,
+} from '@uniswap/analytics'
+import { atomWithStorage, useAtomValue } from 'jotai/utils'
+import { memo } from 'react'
+
+const analyticsOptOutAtomName = 'optOutOfAnalytics'
+export const analyticsOptOutToggle = atomWithStorage<boolean>(analyticsOptOutAtomName, true)
```
zzmp, line 10: This reads like its negation. If you keep this name, you should invert the default value and all usages. Otherwise, something like 'allow_analytics'.

```suggestion
const analyticsOptOutKey = 'analytics_opt_out'
export const analyticsOptOutAtom = atomWithStorage<boolean>(analyticsOptOutKey, false)
```

src/analytics/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const Trace = memo((props: React.ComponentProps<typeof AnalyticsTrace>) => {
+  const allowAnalytics = useAtomValue(analyticsOptOutToggle)
+
+  return <AnalyticsTrace {...props} shouldLogImpression={allowAnalytics} />
```
zzmp, line 15:
```suggestion
  const analyticsOptOut = useAtomValue(analyticsOptOutAtom)
  const shouldLogImpression = analyticsOptOut ? false : props.shouldLogImpression
  return <AnalyticsTrace {...props} shouldLogImpression={shouldLogImpression} />
```

src/analytics/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+})
+
+Trace.displayName = 'Trace'
+
+export const TraceEvent = memo((props: React.ComponentProps<typeof AnalyticsEvent>) => {
+  const allowAnalytics = useAtomValue(analyticsOptOutToggle)
```
zzmp, line 21: See above.

src/analytics/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const sendAnalyticsEvent: typeof sendAnalyticsTraceEvent = (event, properties) => {
+  const value = localStorage.getItem(analyticsOptOutAtomName)
+  let canSendAnalytics = true
+
+  if (typeof value === 'string' && JSON.parse(value) === false) {
+    canSendAnalytics = false
+  }
+
+  if (canSendAnalytics) {
```
zzmp, line 36: This can still fail. Could you use a try/catch?

```suggestion
  let analyticsOptOut = false
  try {
    analyticsOptOut = JSON.parse(localStorage.getItem(analyticsOptOutKey)
  } catch { }

  if (!analyticsOptOut) {
```

src/analytics/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+    sendAnalyticsTraceEvent(event, properties)
+  }
+}
+
+export { getDeviceId, initializeAnalytics, OriginApplication, user, useTrace } from '@uniswap/analytics'
```
zzmp, line 41: nit: I prefer re-exports directly under imports.

src/components/Web3Provider/index.test.tsx
```diff
@@ -15,7 +15,10 @@ import { mocked } from 'test-utils/mocked'
 
 import Web3Provider from '.'
 
-jest.mock('@uniswap/analytics', () => ({
+jest.mock('analytics', () => ({
+  user: { set: jest.fn(), postInsert: jest.fn() },
+}))
```
zzmp, line 20:
```suggestion
```

zzmp (changes requested):

src/analytics/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+TraceEvent.displayName = 'TraceEvent'
+
+export const sendAnalyticsEvent: typeof sendAnalyticsTraceEvent = (event, properties) => {
+  let allowAnalytics = true
+
+  try {
+    const value = localStorage.getItem(allowAnalyticsAtomKey)
+
+    if (typeof value === 'string') {
```
zzmp, line 38: You no longer need this check because you are in a try catch block.

zzmp (approved):
One last nit around checking typeof but lgtm

src/analytics/index.tsx
JackShort: it actually won't let me call `JSON.parse` without checking because it could be `null`

cbachmeier (approved):
LGTM. Is there a way we can verify that analytic events are still being sent?

src/analytics/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+      allowAnalytics = JSON.parse(value)
+    }
+    // eslint-disable-next-line no-empty
+  } catch {}
+
```
cbachmeier, line 43: You could just console.error that value was null to remove this eslint ignore

JackShort: The problem is that will log it into sentry and idk if we want this to be logged to sentry because it's not an actual error

JackShort:
> LGTM. Is there a way we can verify that analytic events are still being sent?

@cbachmeier actually one of the service worker tests checks if analytics is logged on first load

PR #6984, opened 2023-07-19
fix: token img loading state

just-toby:
## Description

adds a background to token logos while the images are loading. once they load, fade the image in and the background out.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-1928/fix-token-selector-loading-states

### Before

https://www.loom.com/share/639ce971313047778eec00aea3fd82b4

### After

https://github.com/Uniswap/interface/assets/66155195/c96cdde6-961f-4693-9d7a-6bff99fe7291

### Reproducing the error

1.  open chrome devtools, open the "network" tab, and click "disable cache" and set "throttling" to "slow 3g"

image

open the token selector and notice how the token images load

### QA (ie manual testing)

- [x] with the above settings configured, open the token selector and note that the images load with the new UX


src/components/Logo/AssetLogo.tsx
```diff
@@ -77,11 +82,22 @@ export default function AssetLogo({
...
   const L2Icon = getChainInfo(chainId)?.circleLogoUrl
+  const [imgLoaded, setImgLoaded] = useState(false)
 
   return (
     <LogoContainer style={style}>
       {src ? (
-        <LogoImage {...imageProps} src={src} onError={nextSrc} useBG={true} />
+        <LogoImageWrapper size={size} useBG={!imgLoaded}>
+          <LogoImage
+            {...imageProps}
```
zzmp, line 92: nit: I think `imageProps` makes this more confusing - can you remove it and just inline the props?

src/components/Logo/AssetLogo.tsx
```diff
@@ -77,11 +82,22 @@ export default function AssetLogo({
...
+            src={src}
+            onLoad={() => {
+              setImgLoaded(true)
+            }}
```
zzmp, line 96: nit: one-liner
```suggestion
            onLoad={() => void setImgLoaded(true)}
```

src/components/Logo/AssetLogo.tsx
```diff
@@ -77,11 +82,22 @@ export default function AssetLogo({
 
   const [src, nextSrc] = useTokenLogoSource(address, chainId, isNative, backupImg)
   const L2Icon = getChainInfo(chainId)?.circleLogoUrl
+  const [imgLoaded, setImgLoaded] = useState(false)
```
zzmp, line 85: nit: I'd prefer this state match the props to which it is passed. Could you rename `imgLoading` and `useBg` to both be `imgLoaded` so there is less renaming of this boolean?

zzmp (approved):
lgtm but a few nits that I'd prefer be resolved before merging.
Also, `LogoImage` shouldn't be used outside of this file: could you update `PortfolioLogo` to use `AssetLogo` instead of `LogoImage` as a follow-up?

JackShort (approved):

JackShort (approved):

JackShort (approved):

PR #6988, opened 2023-07-21
feat: improve logic around hidden section of mini-portfolio balances spam tokens

natew:
Spam tokens are now hidden by default.

This is only for the first part of the ticket, but it's p0 and the rest takes a bit longer so will do in parts.

## Description

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-1940/[mp]-update-how-we-handle-what-goes-in-hidden-token-section-of-mini
_Slack thread:_
_Relevant docs:_

### Reproducing the error

1. Swap for a spam token so it's in your portfolio (DOG works for me right now), must be over $1 to test because under $1 already hides
2. It should show under Hidden

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test


cbachmeier (changes requested):

src/components/AccountDrawer/MiniPortfolio/Tokens/index.tsx
```diff
@@ -33,27 +41,35 @@ export default function Tokens({ account }: { account: string }) {
...
-      ? []
-      : data?.portfolios?.[0].tokenBalances?.filter(
-          (tokenBalance) => !meetsThreshold(tokenBalance, hideSmallBalances)
-        ) ?? []
-  }, [data?.portfolios, hideSmallBalances])
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const tokenBalances = data?.portfolios?.[0].tokenBalances ?? []
+  type TokenBalance = (typeof tokenBalances)[0]
+
+  const [visibleTokens, hiddenTokens] = useMemo(() => {
```
cbachmeier, line 48: Can we add some tests for spam/non-spam/<$1/>$1/no value tokens in the MiniPort?

src/components/AccountDrawer/MiniPortfolio/Tokens/index.tsx
```diff
@@ -23,7 +23,15 @@ import PortfolioRow, { PortfolioSkeleton, PortfolioTabWrapper } from '../Portfol
...
 
 function meetsThreshold(tokenBalance: TokenBalance, hideSmallBalances: boolean) {
-  return !hideSmallBalances || (tokenBalance.denominatedValue?.value ?? 0) > HIDE_SMALL_USD_BALANCES_THRESHOLD
+  const value = tokenBalance.denominatedValue?.value
+
+  // we no longer hide tokens with no price
+  // https://linear.app/uniswap/issue/WEB-1940/[mp]-update-how-we-handle-what-goes-in-hidden-token-section-of-mini
+  if (typeof value !== 'number') {
+    return true
+  }
```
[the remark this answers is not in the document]
natew: asked here but no answer quite yet

https://uniswapteam.slack.com/archives/C047U65H422/p1691432486229279

src/components/AccountDrawer/MiniPortfolio/Tokens/index.tsx
```diff
@@ -33,27 +67,18 @@ export default function Tokens({ account }: { account: string }) {
...
+    () => splitHiddenTokens(tokenBalances ?? [], { hideSmallBalances }),
+    [hideSmallBalances, tokenBalances]
+  )
 
   if (!data) {
     return <PortfolioSkeleton />
   }
 
-  if (data?.portfolios?.[0].tokenBalances?.length === 0) {
+  if (tokenBalances.length === 0) {
```
cbachmeier, line 81:
```suggestion
  if (tokenBalances?.length === 0) {
```

src/components/AccountDrawer/MiniPortfolio/Tokens/splitHiddenTokens.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+    },
+  },
+  // valid
+  {
+    id: 'valid',
+    ownerAddress: '',
+    __typename: 'TokenBalance',
+    denominatedValue: {
+      id: '',
+      value: 100,
```
cbachmeier, line 51: can we also add a test for a non-spam token with no value?

src/components/AccountDrawer/MiniPortfolio/Tokens/index.tsx
```diff
@@ -22,8 +22,42 @@ import PortfolioRow, { PortfolioSkeleton, PortfolioTabWrapper } from '../Portfol
...
-function meetsThreshold(tokenBalance: TokenBalance, hideSmallBalances: boolean) {
-  return !hideSmallBalances || (tokenBalance.denominatedValue?.value ?? 0) > HIDE_SMALL_USD_BALANCES_THRESHOLD
+function meetsThreshold(tokenBalance: TokenBalance) {
+  const value = tokenBalance.denominatedValue?.value ?? 0
+  return value > HIDE_SMALL_USD_BALANCES_THRESHOLD
+}
+
+// exporting for use in tests only
+// eslint-disable-next-line import/no-unused-modules
+export function splitHiddenTokens(
```
cbachmeier, line 32: we generally should try to avoid exporting only for tests. An alternative would be to do a cypress test for the `MiniPortfolio/Token` component with mock input. See `src/components/AccountDrawer/MiniPortfolio/PortfolioLogo.test.tsx` for an example

natew: @cbachmeier i think starting from scratch i'd do that and mock the `useCachedPortfolioBalancesQuery` hook, but rewriting it now feels like it ends up testing the same basic thing but with a decent amount of added effort. sound fair?

i did refactor the `splitHiddenTokens` into its own file and moved it to utils which actually feels right given we likely want this function to be shared anywhere we want to hide tokens, which could be (is?) in more places than here. lmk if that sounds alright.

cbachmeier (approved):

just-toby (approved):

just-toby (approved):

PR #6990, opened 2023-07-21
fix: display already-loaded imgs with no transition

zzmp:
## Description

Fix image loading so that already-loaded images do not fade in, but are immediately displayed.
This fixes the transition to /swap from the landing page. It also improves Token Selector transitions.

_before_

https://github.com/Uniswap/interface/assets/5403956/19619584-610c-4d46-8f27-4020f787aec8

### DIFF
--- a/src/components/Logo/AssetLogo.tsx
+++ b/src/components/Logo/AssetLogo.tsx
@@ -76,7 +76,11 @@ export default function AssetLogo({
 }: AssetLogoProps) {
   const [src, nextSrc] = useTokenLogoSource(address, chainId, isNative, backupImg)
   const L2Icon = getChainInfo(chainId)?.circleLogoUrl
-  const [imgLoaded, setImgLoaded] = useState(false)
+  const [imgLoaded, setImgLoaded] = useState(() => {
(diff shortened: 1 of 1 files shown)


just-toby (approved):

src/components/Logo/AssetLogo.tsx
```diff
@@ -76,7 +76,11 @@ export default function AssetLogo({
 }: AssetLogoProps) {
   const [src, nextSrc] = useTokenLogoSource(address, chainId, isNative, backupImg)
   const L2Icon = getChainInfo(chainId)?.circleLogoUrl
-  const [imgLoaded, setImgLoaded] = useState(false)
+  const [imgLoaded, setImgLoaded] = useState(() => {
```
cbachmeier, line 79: does this mean for every asset logo we'll basically be generating two images, one to be displayed and one to check if it has loaded? Does this impact page performance?

zzmp: yes. I don't think so, because although the image will exist in JS, it will not be rendered to the DOM.

cbachmeier (approved):

PR #6993, opened 2023-07-23
fix: use redirect for landing

zzmp:
## Description

Fixes the page flash when loading the landing page. This was caused by hiding the landing chrome using state, which was initially set to hide the state. Instead, "hiding the landing chrome" should be done by redirecting to the swap route.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2438/im-often-seeing-the-landing-page-flash-when-loading-a-fresh-tab-or

---

**Reviewers:**
Hide white space when reviewing this change;
the bulk of this change is de-indentation from removing the conditional gating from the `showContent` state.

### DIFF
--- a/src/pages/Landing/index.tsx
+++ b/src/pages/Landing/index.tsx
@@ -10,8 +10,9 @@ import { BaseButton } from 'components/Button'
 import { AppleLogo } from 'components/Logo/AppleLogo'
 import { useDisableNFTRoutes } from 'hooks/useDisableNFTRoutes'
 import Swap from 'pages/Swap'
+import { RedirectPathToSwapOnly } from 'pages/Swap/redirects'
 import { parse } from 'qs'
-import { useEffect, useMemo, useRef, useState } from 'react'
+import { useEffect, useMemo, useRef } from 'react'
 import { ArrowDownCircle } from 'react-feather'
 import { useLocation, useNavigate } from 'react-router-dom'
 import { Link as NativeLink } from 'react-router-dom'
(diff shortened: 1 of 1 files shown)


src/pages/Landing/index.tsx
```diff
@@ -300,35 +301,29 @@ const Link = styled(NativeLink)`
...
   useEffect(() => {
-    if ((queryParams.intro || !selectedWallet) && !accountDrawerOpen) {
-      setShowContent(true)
-    } else {
-      setShowContent(false)
+    if (accountDrawerOpen) {
       setTimeout(() => {
         navigate('/swap')
-      }, TRANSITION_DURATIONS.medium)
+      }, TRANSITION_DURATIONS.fast)
```
zzmp, line 318: This is intentionally changed to match the account drawer animation speed.

src/pages/Landing/index.tsx
```diff
@@ -300,35 +301,28 @@ const Link = styled(NativeLink)`
...
-  }, [navigate, selectedWallet, queryParams.intro, accountDrawerOpen])
+  }, [accountDrawerOpen, navigate])
 
-  const shouldDisableNFTRoutes = useDisableNFTRoutes()
-
-  const cards = useMemo(
-    () => MAIN_CARDS.filter((card) => !(shouldDisableNFTRoutes && card.to.startsWith('/nft'))),
-    [shouldDisableNFTRoutes]
-  )
+  const queryParams = parse(useLocation().search, { ignoreQueryPrefix: true })
```
JFrankfurt, line 322: nit: I find inline hook usage a little harder to parse at first glance and prefer the prior 2-line setup

src/pages/Landing/index.tsx
```diff
@@ -300,35 +301,28 @@ const Link = styled(NativeLink)`
...
+  )
 
   const [accountDrawerOpen] = useAccountDrawer()
+  const navigate = useNavigate()
   useEffect(() => {
-    if ((queryParams.intro || !selectedWallet) && !accountDrawerOpen) {
-      setShowContent(true)
-    } else {
-      setShowContent(false)
+    if (accountDrawerOpen) {
```
JFrankfurt, line 315: I expected:
`if ((!queryParams.intro && selectedWallet) || accountDrawerOpen)`

zzmp: This is intentionally split. Handling both as navigations (ie `navigate` or `Redirect`) allows us to avoid having any state, and keeps the landing page "purer" by not having it mimic the /swap page.

`!queryParams.intro && selectedWallet` can be redirected immediately. They are state that exist on pageload, and so can be handled immediately to avoid even rendering the landing page.

`accountDrawerOpen` is dependent on user action after pageload, so it needs to be handled separately, in a `useEffect`. It also needs to be delayed so that the account drawer _can_ open before navigating, to avoid jank.

JFrankfurt (approved):

natew (approved):

PR #6994, opened 2023-07-24
fix: fixing price range filter passing as number

JackShort:
* app crashes when url is loaded with price range because it is passing in a number instead of a string causing it to crash

https://linear.app/uniswap/issue/WEB-2523/[safari]-blank-page-and-then-get-something-went-wrong-error-when-set

### DIFF
--- a/src/nft/components/collection/CollectionNfts.tsx
+++ b/src/nft/components/collection/CollectionNfts.tsx
@@ -1,10 +1,10 @@
 import { BigNumber } from '@ethersproject/bignumber'
+import { parseEther } from '@ethersproject/units'
 import { BrowserEvent, InterfaceElementName, NFTEventName } from '@uniswap/analytics-events'
 import { useWeb3React } from '@web3-react/core'
 import { TraceEvent } from 'analytics'
 import clsx from 'clsx'
 import { OpacityHoverState } from 'components/Common'
-import { parseEther } from 'ethers/lib/utils'
 import { NftAssetTraitInput, NftMarketplace, NftStandard } from 'graphql/data/__generated__/types-and-hooks'
 import { ASSET_PAGE_SIZE, AssetFetcherParams, useNftAssets } from 'graphql/data/nft/Asset'
 import useDebounce from 'hooks/useDebounce'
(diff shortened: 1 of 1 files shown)


cbachmeier (approved):

just-toby (approved):
can we add a test that would catch this crash?

zzmp (changes requested):
This should be caught by typechecking. Why is `maxPrice` _able_ to be anything other than a string? I think that's the root cause, and what should be fixed. This seems like a band-aid that may not prevent recurrence in the future.

> can we add a test that would catch this crash?

typechecking _should_ do this for us, as `parseEther` is typed to only accept a `string`.

zzmp:
Can you also link the ticket or include a repro in the description, so that we can repro the original issue?

JackShort:
> This should be caught by typechecking. Why is `maxPrice` _able_ to be anything other than a string? I think that's the root cause, and what should be fixed. This seems like a band-aid that may not prevent recurrence in the future.
> 
> > can we add a test that would catch this crash?
> 
> typechecking _should_ do this for us, as `parseEther` is typed to only accept a `string`.

Yeah i agree this is a bandaid solution that is because the underlying problem is a lot bigger - it has to do with the way the nft page encodes and decodes the url params. So this is definitely something we have to look at more, but the actual solution involves changing a lot of things involving the urlParams and how we handle them. And is a little more tedious and I just wanted to get this fix in so it we can patch staging and it doesn't crash prod

zzmp (approved):
That's fine, as long as you can ticket it so it doesn't get lost. Someone should audit the query param parsing to see why it is casting params to the wrong types.

JackShort:
> That's fine, as long as you can ticket it so it doesn't get lost. Someone should audit the query param parsing to see why it is casting params to the wrong types.

yeah already pinged adrian about it. I am aware why this is happening there is a spot where we query for url params and we have convert to numbers set as true and then we set the state of collectionFilters based of a variable that is typed as any which overrides it at runtime to a number. but in order to change that we have to change a bunch of the way we handle urlParams

PR #6996, opened 2023-07-24
fix: use deferred value to avoid suspense issues with inputting text

natew:
use deferred value to avoid suspense issues with inputting text during supsended render causing errors

## Description

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2525/[web][ipados][search]-deleting-any-search-causes-an-error-in-the-page
_Slack thread:_
_Relevant docs:_

### Before

If you hit backspace on a search it breaks search pretty often

### After

No break

### Reproducing the error

1. 

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test

### DIFF
--- a/src/components/Suspense/SuspenseWithPreviousRenderAsFallback.tsx
+++ b/src/components/Suspense/SuspenseWithPreviousRenderAsFallback.tsx
@@ -1,5 +1,4 @@
-import usePrevious from 'hooks/usePrevious'
-import React, { Suspense } from 'react'
+import React, { Suspense, useDeferredValue } from 'react'
 
 /**
  * This is useful for keeping the "last rendered" components on-screen while any suspense
@@ -9,6 +8,7 @@ import React, { Suspense } from 'react'
  */
 
 export const SuspenseWithPreviousRenderAsFallback = (props: { children: React.ReactNode }) => {
-  const previousContents = usePrevious(props.children)
-  return <Suspense fallback={previousContents ?? null}>{props.children}</Suspense>
+  const previousChildren = useDeferredValue(props.children)
+
+  return <Suspense fallback={previousChildren ?? null}>{props.children}</Suspense>
 }


src/components/Suspense/SuspenseWithPreviousRenderAsFallback.tsx
```diff
@@ -10,5 +10,7 @@ import React, { Suspense } from 'react'
 
 export const SuspenseWithPreviousRenderAsFallback = (props: { children: React.ReactNode }) => {
   const previousContents = usePrevious(props.children)
-  return <Suspense fallback={previousContents ?? null}>{props.children}</Suspense>
+  const previousDeferred = useDeferredValue(previousContents)
```
natew, line 13: evidently this works? https://stackoverflow.com/a/76173194

JackShort: do we need `usePrevious` and `useDeferredValue` or would this work with just `useDeferredValue`?

cbachmeier: Confirmed that just using `useDeferredValue` gives the desired search behaviour without crashing on iPad

cbachmeier:
```suggestion
  const previousDeferred = useDeferredValue(props.children)
```

cbachmeier (changes requested):

src/components/Suspense/SuspenseWithPreviousRenderAsFallback.tsx
natew: we may need previous, this area also fixes another issue we had in search, let me make sure tests pass and it works still without

natew: confirm it works this way!

natew:
added flake fixes as requested by @zzmp it's not beautiful but from what i can see in cypress its just cypress being flaky for whatever reason

cypress/e2e/universal-search.test.ts
```diff
@@ -25,31 +25,42 @@ describe('Universal search bar', () => {
     cy.location('hash').should('equal', '#/tokens/ethereum/0x1f9840a85d5af5bf1d1762f925bdaddc4201f984')
   })
 
-  it('should go to the selected result when recent results are shown', () => {
-    // Search for uni token by name.
-    getSearchBar().type('uni')
-    cy.get('[data-cy="searchbar-token-row-UNI"]')
+  it(
```
cbachmeier, line 28: these changes should be in their own PR and not coupled to this hotfix

cbachmeier (changes requested):

cypress/e2e/universal-search.test.ts
natew: it's failing this test though, this tests the same element we're fixing

natew: more context: i added this test when i implemented the SuspenseFallback stuff, basically this test was made for the same code were fixing here

cbachmeier: These are two different fixes.
1. Fixes the suspense issue causing a page crash
2. Fixes a flaky test relating to the searchbar that is currently failing on main
These should be their own PRs. The PR that fixes the flaky test by adding retries should not have the rider clause of altering use Suspense behaviour or vice versa because they touch the same component. They're two separate changes.

cbachmeier (changes requested):

cypress/e2e/universal-search.test.ts
natew: The flaky test was actually added to cover the thing that's erroring. I'll split it out, but they are one and the same (test is testing the area that breaks, was added to cover the thing that broke). I do think adding yet another test to test backspace would be good here as well.

cbachmeier (approved):

zzmp (approved):
I was not able to reproduce the error, but this looks good nonetheless.

PR #6998, opened 2023-07-24
feat: update infura fork block height for cypress tests to be dynamic

cbachmeier:
## Description

- Update UR SDK
- Import `UNIVERSAL_ROUTER_CREATION_BLOCK` to get dynamically get block height per chain that UR was deployed to

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2205/import-a-minimum-block-height-in-hardhatconfigjs-from-universal-router

### DIFF
--- a/hardhat.config.js
+++ b/hardhat.config.js
@@ -4,12 +4,6 @@ import { UNIVERSAL_ROUTER_CREATION_BLOCK } from '@uniswap/universal-router-sdk'
 /* eslint-env node */
 require('dotenv').config()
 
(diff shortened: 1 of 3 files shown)


just-toby (approved):
why are we doing this?

zzmp (approved):

zzmp:
> why are we doing this?

We had some test failures due to the block height not being set appropriately and it was difficult to debug. This guarantees that the block height is up-to-date with the router version, to prevent the hard-to-debug failures from recurring. This took <1hr to implement, whereas debugging took half a day.

PR #6999, opened 2023-07-24
fix: remove duplicate hook for permit2 approvals

just-toby:
## Description

`usePermit2Approval` was a wrapper hook around `usePermit2Allowance` used for NFT flows but didn't add much extra logic. to reduce confusion i'm proposing we remove it and use the main hook for the NFT flow.

this is part of a little effort i'm doing to refactor the permit2 hook to be less confusing, which will come in a  follow up diff

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2079/[permit2-follow-up]-split-up-the-allowance-hook

### QA (ie manual testing)

1. ran unit tests and e2e tests
2. worked w/ NFT team to verify the core flows still work


src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -161,17 +161,7 @@ function useConfirmModalState({
         case ConfirmModalState.APPROVING_TOKEN:
           setConfirmModalState(ConfirmModalState.APPROVING_TOKEN)
           invariant(allowance.state === AllowanceState.REQUIRED, 'Allowance should be required')
-          allowance
-            .approve()
-            .then(() => {
-              sendAnalyticsEvent(InterfaceEventName.APPROVE_TOKEN_TXN_SUBMITTED, {
```
just-toby, line 167: moved this logging to be internal to the permit2 hooks

JackShort (approved):
tested and was able to buy an NFT through pwat

src/nft/components/bag/BagFooter.tsx
```diff
@@ -401,14 +412,15 @@ export const BagFooter = ({ setModalIsOpen, eventProperties }: BagFooterProps) =
...
-    if (allowance.state === AllowanceState.REQUIRED || loadingAllowance) {
-      const handleClick = () => updateAllowance()
+    const allowanceRequired = allowance.state === AllowanceState.REQUIRED
+    if (allowanceRequired || loadingAllowance) {
+      const handleClick = () => allowanceRequired && allowance.approveAndPermit()
 
       if (loadingAllowance) {
         return getBuyButtonStateData(BuyButtonStates.LOADING_ALLOWANCE, theme, handleClick)
-      } else if (isAllowancePending) {
+      } else if (allowanceRequired && allowance.isApprovalPending) {
```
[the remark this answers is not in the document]
just-toby: we actually do to make the type checker happy

src/nft/components/bag/BagFooter.test.tsx
```diff
@@ -5,13 +5,12 @@ import { useWeb3React } from '@web3-react/core'
 import { nativeOnChain } from 'constants/tokens'
```
cbachmeier, line 5: With the deletion of `src/nft/hooks/usePermit2Approval.test.ts` we do lose the test case that the correct UR address is being used. Can you add that back here?

src/nft/components/bag/BagFooter.tsx
```diff
@@ -262,6 +262,17 @@ const PENDING_BAG_STATUSES = [
   BagStatus.PROCESSING_TRANSACTION,
 ]
 
+function getURAddress(chainId?: number, nftURAddress?: string) {
```
cbachmeier, line 265: can you move this util function to `useNftUniversalRouterAddress`

src/nft/components/bag/BagFooter.tsx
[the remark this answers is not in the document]
cbachmeier: The linter doesn't seem to parse that since you already checked `allowanceRequired || loadingAllowance` and confirmed that `loadingAllowance` is false, this must mean that `allowanceRequired` is true. 

Looking over the code @JackShort it doesn't seem like there's that much overlap between allowanceRequired and loadingAllowance, notably the `onClick` doesn't affect the button when it's in the `LOADING_ALLOWANCE` state. Could we split the code as follows? 

```
    const allowanceRequired = allowance.state === AllowanceState.REQUIRED
    if (loadingAllowance) {
      return getBuyButtonStateData(BuyButtonStates.LOADING_ALLOWANCE, theme, handleClick)
    }

    if (allowanceRequired) {
      const handleClick = () => allowance.approveAndPermit()

      if (allowance.isApprovalPending) {
        return getBuyButtonStateData(BuyButtonStates.IN_WALLET_ALLOWANCE_APPROVAL, theme, handleClick)
      } else if (allowance.isApprovalLoading) {
        return getBuyButtonStateData(BuyButtonStates.PROCESSING_APPROVAL, theme, handleClick)
      }

      return getBuyButtonStateData(BuyButtonStates.REQUIRE_APPROVAL, theme, handleClick)
    }

cbachmeier (changes requested):

cbachmeier (approved):

PR #7000, opened 2023-07-24
feat: vertical LP

JFrankfurt:
_note: this PR may be superseded by an upcoming one from @gnewfield_

https://github.com/Uniswap/interface/pull/7092

## Description

This converts the default LP experience to be vertical instead of horizontal.

image

### Before
| image | image |

### After
| image  | image |

### QA (ie manual testing)
Behavior should remain unchanged.

- [ ] open a new LP position

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test


src/components/LiquidityChartRangeInput/index.tsx
```diff
@@ -46,7 +46,8 @@ const ZOOM_LEVELS: Record<FeeAmount, ZoomLevels> = {
 
 const ChartWrapper = styled.div`
   position: relative;
-
+  width: 100%;
+  max-height: 150px;
```
JFrankfurt, line 50: @callil confirming 150 (not 148 or 152)

gnewfield:
I'm seeing a bug on pool creation at new fee tiers, where the range selector is not being shown. The range selector should be shown below the new price input. Attaching before and after screenshots for comparison. This is probably just due to the reordering, which is causing a mismatch between the components and the state variables that are used for conditional rendering.

|    Before    |   After    |
|Screenshot 2023-07-31 at 5 32 10 PM |Screenshot 2023-07-31 at 5 34 54 PM |

src/components/CurrencyInputPanel/index.tsx
```diff
@@ -226,110 +212,99 @@ export default function CurrencyInputPanel({
...
-                    >
-                      <StyledBalanceMax onClick={onMax}>
-                        <Trans>MAX</Trans>
-                      </StyledBalanceMax>
-                    </TraceEvent>
-                  ) : null}
-                </RowFixed>
-              ) : (
-                <span />
+      {!locked && (
```
just-toby, line 215: can we do an early return `if (locked) return null` and un-indent this JSX for readability?

just-toby: also, are we intentionally removing the `locked` UI that was shown before? see deleted Line 234.

src/pages/AddLiquidity/index.tsx
```diff
@@ -600,7 +601,7 @@ function AddLiquidity() {
           )}
           pendingText={pendingText}
         />
-        <PageWrapper wide={!hasExistingPosition}>
+        <BodyWrapper style={{ padding: hasExistingPosition ? '10px' : 0, maxWidth: '640px' }}>
```
just-toby, line 604: no inline style, pass a prop to the styled component

just-toby (changes requested):

cartcrom (approved):
We should def look into Gray's bug comment, but assuming eddie's comments are fixed LGTM

just-toby (approved):

just-toby (approved):

cartcrom (approved):

gnewfield (approved):

PR #7001, opened 2023-07-24
fix: use click instead of enter in search test

just-toby:
## Description

this test is now flaking out because sometimes cypress's `type({enter})` doesn't actually trigger the navigation to the selected token. i'm hoping this will help the flakiness.


zzmp (approved):

natew:
when i ran this test a bunch of times i saw it fail only 4 times out of 10 or so, but each time it was on trying to find the element rather than the enter, but maybe i was reading it wrong? i don't mind this fix if it works

PR #7012, opened 2023-07-26
fix: improve v2 network support

JFrankfurt:


src/constants/chains.ts
```diff
@@ -41,16 +41,9 @@ export const SUPPORTED_GAS_ESTIMATE_CHAIN_IDS = [
...
  */
-export const UNSUPPORTED_V2POOL_CHAIN_IDS = [
-  ChainId.POLYGON,
-  ChainId.OPTIMISM,
-  ChainId.ARBITRUM_ONE,
-  ChainId.BNB,
-  ChainId.ARBITRUM_GOERLI,
-  ChainId.AVALANCHE,
-] as const
+export const SUPPORTED_V2POOL_CHAIN_IDS = [ChainId.MAINNET, ChainId.GOERLI] as const
```
tinaszheng, line 46: sepolia works too!

tinaszheng (approved):
thoughts on also amending `useV2RouterContract` to return undefined if the chainId is not supported?

src/pages/Pool/index.tsx
```diff
@@ -225,7 +226,6 @@ export default function Pool() {
   }
 
   const showConnectAWallet = Boolean(!account)
-  const showV2Features = Boolean(V2_FACTORY_ADDRESSES[chainId])
```
zzmp, line 228: My only issue here is that you're adding a new file to the V2 checking codepath, which complicates adding add'l chains. However, I don't think we plan on adding add'l chains to V2, so this seems fine.

zzmp (approved):

vm (approved):
just approving after main was merged

PR #7015, opened 2023-07-26
fix: improve v2 network support

just-toby:
staging hotfix copy of #7012 

tested by navigating to each of the following URLs on unsupported chains and verifying the unsupported UI appears:

- /pool/v2
- /pools/v2
- /pools/v2/find
- /add/v2
- /add/v2/{currencyIdA}/{currencyIdB}
- /remove/v2/{currencyIdA}/{currencyIdB}
- /migrate/v2/{address}


src/pages/Pool/index.tsx
```diff
@@ -225,7 +226,6 @@ export default function Pool() {
   }
 
   const showConnectAWallet = Boolean(!account)
-  const showV2Features = Boolean(V2_FACTORY_ADDRESSES[chainId])
```
zzmp, line 228: My only issue here is that you're adding a new file to the V2 checking codepath, which complicates adding add'l chains. However, I don't think we plan on adding add'l chains to V2, so this seems fine.

zzmp (approved):

JFrankfurt (approved):

PR #7016, opened 2023-07-26
fix: uniswap wallet deeplinking

cartcrom:
Updates the custom uniswap wallet connector to use a regular deeplink (uniswap://wc) rather than a universal link (uniswap.org/wc...), since the universal link stopped functioning correctly. This makes the uniswap modal's link consistent with the link the proprietary walletconnect connector uses.

Also updates the uniswap wallet modal to not display on iOS, correcting visual lag that would appear upon the the wallet navigating back to interface after successfully walletconnecting.

In addition, we now set `window.location.href` rather than using `window.open`, which is the more correct way to deeplink from mobile browsers (window.open may be hindered by popup blockers or security measures, whereas window.location.href enables the operating system to handle the link appropriately, automatically launching the corresponding app if installed)

Safari (app Installed)

https://github.com/Uniswap/interface/assets/39385577/b9e63bc5-3998-491c-bf4d-1261bff2ab3d

Chrome (app installed)

https://github.com/Uniswap/interface/assets/39385577/6d32728b-f6ef-410e-985d-bcd2ea7e9a8f

Safari (app not installed)

https://github.com/Uniswap/interface/assets/39385577/5bac612b-b275-4c11-9c0c-bc3978d95e2a

Chrome (app not installed)

https://github.com/Uniswap/interface/assets/39385577/fe5f2634-723a-4c0f-b44c-9bafc9af32e0

### DIFF
--- a/src/connection/WalletConnectV2.ts
+++ b/src/connection/WalletConnectV2.ts
@@ -83,10 +83,7 @@ export class UniwalletConnect extends WalletConnectV2 {
 
       // Opens deeplink to Uniswap Wallet if on iOS
       if (isIOS) {
-        const newTab = window.open(`https://uniswap.org/app/wc?uri=${encodeURIComponent(uri)}`)
-
-        // Fixes blank tab opening on mobile Chrome
-        newTab?.close()
+        window.location.href = `uniswap://wc?uri=${encodeURIComponent(uri)}`
       }
     })
   }
--- a/src/components/AccountDrawer/UniwalletModal.tsx
+++ b/src/components/AccountDrawer/UniwalletModal.tsx
@@ -13,6 +13,7 @@ import { QRCodeSVG } from 'qrcode.react'
 import { useEffect, useState } from 'react'
 import styled, { useTheme } from 'styled-components/macro'
 import { CloseIcon, ThemedText } from 'theme'
+import { isIOS } from 'utils/userAgent'
 
 import uniPng from '../../assets/images/uniwallet_modal_icon.png'
 import { DownloadButton } from './DownloadButton'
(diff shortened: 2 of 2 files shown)


src/connection/WalletConnectV2.ts
```diff
@@ -83,10 +83,7 @@ export class UniwalletConnect extends WalletConnectV2 {
 
       // Opens deeplink to Uniswap Wallet if on iOS
       if (isIOS) {
-        const newTab = window.open(`https://uniswap.org/app/wc?uri=${encodeURIComponent(uri)}`)
-
-        // Fixes blank tab opening on mobile Chrome
-        newTab?.close()
+        window.location.href = `uniswap://wc?uri=${encodeURIComponent(uri)}`
```
just-toby, line 86: are you sure we want to change the location in the current tab? if the user is trying to connect the wallet for an interaction on this page, they probably expect to be able to return to this tab and continue the desired flow

just-toby (changes requested):

just-toby (approved):

src/connection/WalletConnectV2.ts
zzmp: I'm concerned about this too.

zzmp (changes requested):

src/connection/WalletConnectV2.ts
just-toby: discussed offline and it seems like window.open is broken on ios, so we need to use this method. i asked for updates to the PR description so we can validate this is working as intended.

apparently window.location.href doesn't replace the tab contents IF the URL is a deeplink

zzmp: @cartcrom could you document this as a comment inline?

zzmp (changes requested):

zzmp (approved):
documentation is needed to PR this back to `main`, but this is fine for staging :)

PR #7020, opened 2023-07-26
fix: change style of mobile pool buttons and menu

gnewfield:
## Description
[WEB-2473](https://linear.app/uniswap/issue/WEB-2473/pool-page-on-mweb-new-position-and-more-buttons-should-be-50percent) identifies a couple of issues with the style of the buttons on the mobile web "Pools" page. As requested, this PR:
* Switches the order of the buttons so that (from left to right) "More" appears first, followed by "New Position"
* Changes the sizing of the buttons so that they share the space equally
* Switches the alignment of the fly-out menu from right to left, now that the "More" button is on the left side of the page

In order to include this last change, it was necessary to fix an issue with the `Menu` component, specifically the `MenuFlyout` implementation of alignment props and styling. The old implementation worked for larger screen sizes but was being permanently overriden to right-alignment for medium sizes and smaller (i.e. mobile). Now the alignment will be correct for all screen sizes.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2473/pool-page-on-mweb-new-position-and-more-buttons-should-be-50percent

### Before
| Old Mobile | Old Desktop |

### After
| New Mobile  | New Desktop |

### Reproducing the error
1. Visit the Pools [page](https://app.uniswap.org/#/pools) with a mobile screen size (720 pixels or smaller)
2. Observe that the ordering of the buttons (from left to right) is "New Position," followed by "More"
3. Observe that the size of the "New Position" button is significantly larger than the "More" button
4. Observe that the fly-out menu is right-aligned with the "More" button, even though it is left-aligned in the [code](https://github.com/Uniswap/interface/blob/4d084d0055c2ab0f6ec7d60acd43f8ee50d756b7/src/pages/Pool/index.tsx#L276)

### QA (ie manual testing)
- [x] Visit the preview link and see that the above observations have been corrected
- [x] This PR changes the style of the Menu component, which is used throughout the app, so report any unintended changes to other menus that may have been relying on the broken behavior

#### Devices
Previously, all devices with screen size of 960px or smaller had menu fly-outs that were permanently right-aligned. This PR honors the `flyoutAlignment` props that are passed by parent components. Since right is actually the default alignment if no props are passed, any components that didn't care to set the alignment will remain unchanged. Any components that have the alignment set to left (but weren't) will now actually be aligned to the left.

### Automated testing

- [x] Unit test
- [x] Integration/E2E test

### DIFF
--- a/src/pages/Pool/index.tsx
+++ b/src/pages/Pool/index.tsx
@@ -62,15 +62,13 @@ const ButtonRow = styled(RowFixed)`
     width: 100%;
     flex-direction: row;
     justify-content: space-between;
-    flex-direction: row-reverse;
   `};
 `
 const PoolMenu = styled(Menu)`
   margin-left: 0;
   ${({ theme }) => theme.deprecated_mediaWidth.deprecated_upToSmall`
     flex: 1 1 auto;
-    width: 49%;
-    right: 0px;
+    width: 50%;
   `};
 
   a {
@@ -129,7 +127,7 @@ const ResponsiveButtonPrimary = styled(ButtonPrimary)`
   width: fit-content;
   ${({ theme }) => theme.deprecated_mediaWidth.deprecated_upToSmall`
     flex: 1 1 auto;
-    width: 100%;
+    width: 50%;
   `};
 `
 
--- a/src/pages/Pool/index.tsx
+++ b/src/pages/Pool/index.tsx
@@ -29,47 +29,41 @@ const PageWrapper = styled(AutoColumn)`
   max-width: 870px;
   width: 100%;
 
-  ${({ theme }) => theme.deprecated_mediaWidth.deprecated_upToMedium`
+  @media (max-width: ${({ theme }) => `${theme.breakpoint.md}px`}) {
     max-width: 800px;
-  `};
-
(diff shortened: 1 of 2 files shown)


zzmp (approved):

src/pages/Pool/index.tsx
```diff
@@ -62,15 +62,13 @@ const ButtonRow = styled(RowFixed)`
     width: 100%;
     flex-direction: row;
     justify-content: space-between;
-    flex-direction: row-reverse;
   `};
 `
 const PoolMenu = styled(Menu)`
   margin-left: 0;
   ${({ theme }) => theme.deprecated_mediaWidth.deprecated_upToSmall`
```
just-toby, line 69: it would be nice to move off the deprecated breakpoint while you're updating this, if you have time. 

we should at least be tracking this work in some linear tickets, i don't think we are right now D:

gnewfield: Agreed! I've pushed a new commit which replaces the deprecated media queries and text components in this file. Will be good to incrementally update these usages as we work across the codebase

just-toby (approved):

PR #7022, opened 2023-07-26
build: fix promotion sha

zzmp:
## Description

Updates the sha in the promotion check to be derived correctly from the context.
This enables checking tests before deploying to prod.


zzmp:
I think there should be an override for this, in case something goes flaky after getting to staging, so that we don't further delay a release. wdyt?

cbachmeier (approved):

just-toby (approved):

PR #7023, opened 2023-07-26
feat: bump uniswapx-sdk

tinaszheng:
bumps SDK version

this also adds a feature flag, `forceUniswapXOn`, that forces the routing-api's backend feature flag to be on so we can test uniswapx on staging before the backend fully turns it back on.

### DIFF
--- a/src/featureFlags/index.tsx
+++ b/src/featureFlags/index.tsx
@@ -11,9 +11,10 @@ export enum FeatureFlag {
   fiatOnRampButtonOnSwap = 'fiat_on_ramp_button_on_swap_page',
   detailsV2 = 'details_v2',
   debounceSwapQuote = 'debounce_swap_quote',
-  uniswapXEnabled = 'uniswapx_enabled',
+  uniswapXEnabled = 'uniswapx_enabled', // enables sending dutch_limit config to routing-api
   uniswapXSyntheticQuote = 'uniswapx_synthetic_quote',
   routingAPIPrice = 'routing_api_price',
+  forceUniswapXOn = 'uniswapx_force_on', // forces routing-api's feature flag for uniswapx to turn on as well
 }
 
 interface FeatureFlagsContextType {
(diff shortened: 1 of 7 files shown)


src/featureFlags/index.tsx
```diff
@@ -11,9 +11,10 @@ export enum FeatureFlag {
   fiatOnRampButtonOnSwap = 'fiat_on_ramp_button_on_swap_page',
   detailsV2 = 'details_v2',
   debounceSwapQuote = 'debounce_swap_quote',
-  uniswapXEnabled = 'uniswapx_enabled',
+  uniswapXEnabled = 'uniswapx_enabled', // enables sending dutch_limit config to routing-api
   uniswapXSyntheticQuote = 'uniswapx_synthetic_quote',
   routingAPIPrice = 'routing_api_price',
+  forceUniswapXOn = 'uniswapx_force_on', // forces routing-api's feature flag for uniswapx to turn on as well
```
zzmp, line 17:
```suggestion
  forceUniswapXOn = 'uniswapx_force_on', // explicitly passes `useUniswapX=true` to the routing API
```

src/state/routing/slice.ts
```diff
@@ -143,6 +152,8 @@ export const routingApi = createApi({
               tokenOut: tokenOutAddress,
               amount,
               type,
+              // if forceUniswapXOn is not ON, then use the backend's default value
+              useUniswapX: forceUniswapXOn || undefined,
```
zzmp, line 156: nit: I think this would be safer (for future maintenance) if this were pulled into a separate typed variable:

```
const useUniswapX: true | undefined = forceUniswapXOn || undefined
```

Then it's `tsc` checking the logic instead of a human.

zzmp (approved):
lgtm but nits

src/featureFlags/index.tsx
tinaszheng: eh i think the former is more descriptive for behavior

src/state/routing/slice.ts
tinaszheng: not terribly concerned about future maintenance as we'll rip this out as soon as its deployed :D. but good to keep in mind for next time!

just-toby (approved):

PR #7025, opened 2023-07-26
fix: correct prices on TDP chart

just-toby:
## Description

3 changes here in this PR now:
- fixes the boolean check bug so that $0.00 will appear instead of "price unknown". sometimes we want to show $0.00 as the current price
- updates our logic for showing "price unavailable" - we should show this when there's actually no price data showing up on the charts - i.e. when `prices` is empty
- updates the logic for the default/initial display price - we should use the most recent non-zero value 

_Slack thread:_ https://uniswapteam.slack.com/archives/C04D07TQBT4/p1690404981637159

### Before
image

### After

image

### Reproducing the error

1. navigate to https://app.uniswap.org/#/tokens/optimism/0x3c8b650257cfb5f272f799f5e2b4e65093a11a05

### QA (ie manual testing)

- [x] navigate to same TDP on the preview link and verify that the chart and price data looks correct


cartcrom (changes requested):
These changes cause some issues with the charts. The token I'm testing with is here: https://app.uniswap.org/#/tokens/ethereum/0xb17548c7b510427baac4e267bea62e800b247173

price unavailable text (preview vs prod):
image

The data here is incorrect, it's showing way less than a year in the "1Y slot" (preview vs prod):
Screenshot 2023-07-26 at 6 13 25 PM

just-toby:
@cartcrom thanks for checking, i wasn't really sure what's going on between `prices` and `originalPrices` . i'm updating this PR to _only_ have the boolean bug fix which we definitely want - we can follow up separately to figure out why the most recent price is $0.00 for some tokens when it shouldn't be.

zzmp (approved):
lgtm but please update the PR description to match the updated changes

cartcrom (changes requested):
we're still missing the price unavailable ui here

image

cartcrom (changes requested):
I still think this chart is wrong. If our BE isn't able to get us a current price, we shouldn't just display the last price we had.

image

vs prod
image
also note that prod is showing the past year of data, while this pr's chart is showing april to april

cartcrom:
I disagree with using the most recent non-zero value for the  for the initial display price... if the is no hover input on the chart, it should display the current price of the token. And if we can't provide a current price, we should convey that to the user, which we currently do. I think we should check w/ design if there is actually anything that needs to be done here?

PR #7026, opened 2023-07-26
fix: update token tests for cloudflare functions

LunrEclipse:
## Description

Fixes failing token test by updating expected image url link.


zzmp (approved):

just-toby (approved):
it's not great that this test can break from backend changes. can we at least make it non-blocking?

PR #7028, opened 2023-07-27
feat: don't fetch UniswapX quotes if user has disabled UniswapX

cartcrom:
## Description

only requests classic quotes if the user has disabled UniswapX in swap settings

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2559/dont-fetch-with-dutch-order-config-if-user-has-opted-out-of-uniswapx


tinaszheng (approved):
💯

just-toby (approved):

PR #7034, opened 2023-07-27
test: deflake and clean universal search

zzmp:
## Description

Reverts #7002 and refactors the tests for the universal search bar. These now pass reliably locally.

### DIFF
--- a/cypress/e2e/universal-search.test.ts
+++ b/cypress/e2e/universal-search.test.ts
@@ -1,3 +1,5 @@
+import { getTestSelector } from '../utils'
+
 describe('Universal search bar', () => {
   function openSearch() {
     // can't just type "/" because on mobile it doesn't respond to that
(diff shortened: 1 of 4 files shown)


tinaszheng (approved):

cypress/e2e/universal-search.test.ts
```diff
@@ -6,86 +8,46 @@ describe('Universal search bar', () => {
...
-
-    // Most popular 3 tokens should be shown.
-    cy.get('[data-cy="searchbar-dropdown"]')
-      .contains('[data-cy="searchbar-dropdown"]', 'Popular tokens')
-      .find('[data-cy^="searchbar-token-row"]')
-      .its('length')
-      .should('be.eq', 3)
   })
 
-  it.skip('should show blocked badge when blocked token is searched for', () => {
```
JFrankfurt, line 86: this seems like an important test we're dropping

natew (approved):
is the fix basically that data-testid is more reliable than data-cy? the test itself looks largely the same unless i'm missing something so just wondering

zzmp:
> is the fix basically that data-testid is more reliable than data-cy? the test itself looks largely the same unless i'm missing something so just wondering

honestly I don't know. It's possible this is still flakey, but I did not see flakiness when I tested locally.
I switched to data-testid to standardize, not because of reliability.

PR #7038, opened 2023-07-27
chore: align @reduxjs/toolkit dep versions with wallet

natew:
dep align - no change to functionality, or breaking changes in changelog, did a brief test, but we should test a variety of pages that use redux heavily


cbachmeier (changes requested):
Can you rebase? It seems like there is a lot of unrelated changes in this PR

natew:
this is some stupid github bug i've seen happen now a few times, locally it shows the right diff

natew:
oh maybe it wasnt that bug? but i have confirmed theres a github bug where branches show a bunch of changes? but now i've merged the latest and it fixed which usually with that bug it gets stuck until you push to a new branch and here it didnt, so should be good

cbachmeier (approved):

zzmp (approved):

PR #7041, opened 2023-07-27
chore: align @graphql-codegen dep versions with wallet

natew:
dep align with wallet repo for future monorepo merge, nothing should change in terms of tests/functionality.


yarn.lock
```diff
@@ -2596,6 +2608,15 @@
     "@graphql-tools/utils" "^8.8.0"
     tslib "~2.4.0"
 
+"@graphql-codegen/schema-ast@^3.0.1":
```
cbachmeier, line 2611: we're importing 2 copies of `@graphql-codegen/schema-ast` can this be condensed?

cbachmeier (changes requested):

zzmp:
I don't think we should use `resolutions` for exactly your reasons. It's ok to have multiple versions, as long as we try not to _ship_ multiple versions. A lot of this duplicity is because the build pipelines that we use use pinned versions that require that we download multiple versions, but our actual shipped product usually only uses one version (as enforced by tsc).

zzmp (approved):
@natew I'm not sure why but your installation used npm, whereas the lockfile uses yarnpkg as a package manager url. Could you try to fix that configuration on your own system? otherwise it causes add'l churn in lockfile updates.

I pushed a commit which reverts to the yarnpkg urls.

cbachmeier (approved):

PR #7044, opened 2023-07-28
fix: reverted swap state

just-toby:
## Description

updates ConfirmSwapModal to actually wait for the transaction status to be reported, not just for the TransactionResponse (which may not indicate a successful transaction).

this new logic now corresponds with the Mini Portfolio and notification Popups, so they should all update at the same time.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2577/when-tx-fails-ui-is-showing-success
_Slack thread:_ https://uniswapteam.slack.com/archives/C04D07TQBT4/p1690491370411699

### Before

image

if swap is reverted on chain, we recognize that as a successful swap in the ConfirmSwapModal

### After

successful swap - updates at the same time the popup appears

https://github.com/Uniswap/interface/assets/66155195/d0e349e9-256a-4a75-a939-ff61bbd207e2

### Reproducing the error

1. lower slippage tolerance so that the swap is reverted on chain

### QA (ie manual testing)

- [x] manually tested swaps both successful and failures to validate the expected behavior

### Automated testing

- [x] Unit test n/a
- [x] Integration/E2E test

### DIFF
--- a/cypress/e2e/swap/errors.test.ts
+++ b/cypress/e2e/swap/errors.test.ts
@@ -64,17 +64,15 @@ describe('Swap errors', () => {
 
   it('slippage failure', () => {
     cy.visit(`/swap?inputCurrency=${USDC_MAINNET.address}&outputCurrency=${DAI.address}`, { ethereum: 'hardhat' })
-    cy.hardhat({ automine: false }).then((hardhat) =>
-      hardhat.fund(hardhat.wallet, CurrencyAmount.fromRawAmount(USDC_MAINNET, 500e6))
-    )
-    cy.hardhat().then((hardhat) => hardhat.mine())
-    cy.hardhat().then(({ approval, wallet }) =>
-      Promise.all([
-        approval.setTokenAllowanceForPermit2({ owner: wallet, token: USDC_MAINNET }),
-        approval.setPermit2Allowance({ owner: wallet, token: USDC_MAINNET }),
+    cy.hardhat({ automine: false }).then(async (hardhat) => {
+      await hardhat.fund(hardhat.wallet, CurrencyAmount.fromRawAmount(USDC_MAINNET, 500e6))
+      await hardhat.mine()
+      await Promise.all([
+        hardhat.approval.setTokenAllowanceForPermit2({ owner: hardhat.wallet, token: USDC_MAINNET }),
+        hardhat.approval.setPermit2Allowance({ owner: hardhat.wallet, token: USDC_MAINNET }),
       ])
-    )
-    cy.hardhat().then((hardhat) => hardhat.mine())
+      await hardhat.mine()
+    })
 
     getBalance(DAI).then((initialBalance) => {
       // Gas estimation fails for this transaction (that would normally fail), so we stub it.
--- a/cypress/e2e/swap/errors.test.ts
+++ b/cypress/e2e/swap/errors.test.ts
@@ -1,11 +1,9 @@
 import { BigNumber } from '@ethersproject/bignumber'
-import { ChainId } from '@uniswap/sdk-core'
+import { CurrencyAmount } from '@uniswap/sdk-core'
 
 import { DEFAULT_DEADLINE_FROM_NOW } from '../../../src/constants/misc'
-import { UNI, USDC_MAINNET } from '../../../src/constants/tokens'
+import { DAI, USDC_MAINNET } from '../../../src/constants/tokens'
 import { getBalance, getTestSelector } from '../../utils'
-
-const UNI_MAINNET = UNI[ChainId.MAINNET]
 
 describe('Swap errors', () => {
   it('wallet rejection', () => {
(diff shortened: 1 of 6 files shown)


cypress/e2e/swap/errors.test.ts
```diff
@@ -90,18 +99,25 @@ describe('Swap errors', () => {
...
+        }
       }
       cy.get(getTestSelector('web3-status-connected')).should('contain', '2 Pending')
 
       // Mine transactions
       cy.hardhat().then((hardhat) => hardhat.mine())
       cy.wait('@eth_getTransactionReceipt')
 
-      // Verify transaction did not occur
+      cy.contains('Swap failed')
```
just-toby, line 112: this is testing the UX change from the other files

src/components/swap/PendingModalContent/index.tsx
```diff
@@ -223,14 +226,17 @@ export function PendingModalContent({
...
 
-  const classicSwapConfirmed = useIsTransactionConfirmed(
-    swapResult?.type === TradeFillType.Classic ? swapResult.response.hash : undefined
-  )
+  const transaction = useTransaction(swapResult?.type === TradeFillType.Classic ? swapResult.response.hash : undefined)
+  const tokens = useAllTokensMultichain()
+
+  const swapActivity = transaction && chainId ? transactionToActivity(transaction, chainId, tokens) : undefined
+
+  const classicSwapConfirmed = swapActivity?.status === TransactionStatus.Confirmed
```
just-toby, line 234: this change ensures we update to the success state at the same time as other places in the app

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -297,7 +300,16 @@ export default function ConfirmSwapModal({
...
+  const transaction = useTransaction(swapResult?.type === TradeFillType.Classic ? swapResult.response.hash : undefined)
+  const tokens = useAllTokensMultichain()
+
+  const swapActivity = transaction && chainId ? transactionToActivity(transaction, chainId, tokens) : undefined
+
+  // Swap was reverted onchain.
+  const swapReverted = swapActivity?.status === TransactionStatus.Failed
+  // Swap failed locally and was not broadcast to the blockchain.
+  const localSwapFailure = Boolean(swapError) && !didUserReject(swapError)
+  const swapFailed = localSwapFailure || swapReverted
```
just-toby, line 312: this change ensures we show the "swap failed" UI

cypress/e2e/swap/errors.test.ts
```diff
@@ -64,10 +62,21 @@ describe('Swap errors', () => {
...
+      hardhat.fund(hardhat.wallet, CurrencyAmount.fromRawAmount(USDC_MAINNET, 500e6))
+    )
+    cy.hardhat().then((hardhat) => hardhat.mine())
+    cy.hardhat().then(({ approval, wallet }) =>
+      Promise.all([
+        approval.setTokenAllowanceForPermit2({ owner: wallet, token: USDC_MAINNET }),
+        approval.setPermit2Allowance({ owner: wallet, token: USDC_MAINNET }),
+      ])
+    )
+    cy.hardhat().then((hardhat) => hardhat.mine())
```
zzmp, line 77: You don't need to keep re-calling hardhat, you can just continue to use promises or chain.
```suggestion
    cy.hardhat({ automine: false }).then(async (hardhat) => {
      await hardhat.fund(hardhat.wallet, CurrencyAmount.fromRawAmount(USDC_MAINNET, 500e6))
      await hardhat.mine()
      await Promise.all([
        approval.setTokenAllowanceForPermit2({ owner: wallet, token: USDC_MAINNET }),
        approval.setPermit2Allowance({ owner: wallet, token: USDC_MAINNET }),
      ])
      await hardhat.mine()
    )
```

cypress/e2e/swap/errors.test.ts
```diff
@@ -90,18 +99,25 @@ describe('Swap errors', () => {
         cy.contains('Confirm swap').click()
         cy.wait('@eth_sendRawTransaction').wait('@eth_getTransactionReceipt')
         cy.contains('Swap submitted')
-        cy.get(getTestSelector('confirmation-close-icon')).click()
+        if (i === 0) {
+          cy.get(getTestSelector('confirmation-close-icon')).click()
+        }
```
zzmp, line 104: Is this branching necessary?

zzmp (approved):
lgtm, just a change to the cy.hardhat usage to avoid repeated calls.

cypress/e2e/swap/errors.test.ts
just-toby: yes we only want to dismiss the first modal. i could take the code out of a loop to avoid the branching

zzmp (approved):

src/state/transactions/hooks.tsx
```diff
@@ -89,6 +94,15 @@ export function useIsTransactionConfirmed(transactionHash?: string): boolean {
   return Boolean(transactions[transactionHash].receipt)
 }
 
+export function useSwapTransactionStatus(swapResult: SwapResult | undefined): TransactionStatus | undefined {
+  const { chainId } = useWeb3React()
+  const transaction = useTransaction(swapResult?.type === TradeFillType.Classic ? swapResult.response.hash : undefined)
+  const tokens = useAllTokensMultichain()
+
+  const swapActivity = transaction && chainId ? transactionToActivity(transaction, chainId, tokens) : undefined
+  return swapActivity?.status
```
cartcrom, line 103: I think converting to the Activity type is overkill here. I suggest pulling the following  code from mini portfolio into a helper fn:

```
const status = !details.receipt
      ? TransactionStatus.Pending
      : details.receipt.status === 1 || details.receipt?.status === undefined
      ? TransactionStatus.Confirmed
      : TransactionStatus.Failed
```

so you can just do this:
```suggestion
  return getTransactionStatus(transaction)
```

cartcrom (changes requested):

src/state/transactions/hooks.tsx
just-toby: sure

cartcrom (approved):

PR #7045, opened 2023-07-28
feat: add spore colors and font and apply new styles

callil:
## Description
Draft - opening to track

### Before

### After


public/fonts/Basel-Grotesk-Essential-Variable.woff
natew, line 1: So this is a pretty huge font file, and if we do subsetting we could save a ton of size (maybe get it down to 20-30kb per weight * 2 = 60kb).

The only thing is then we'd be subset to mostly english characters, not sure how complete these variable fonts are generally and not sure how our i18n is set up, I see we call u18n.load() but then we'd have to subset for different locales which is quite a task.

So all in all I guess this is the right option.

We have one more problem which is that the variable version of this font doesn't do the nice `$` like the non-variable does for some reason.

src/components/AccountDrawer/AuthenticatedHeader.tsx
```diff
@@ -301,54 +305,57 @@ export default function AuthenticatedHeader({ account, openSettings }: { account
             <LoadingBubble height="16px" width="100px" margin="4px 0 20px 0" />
           </Column>
         )}
-        {!shouldDisableNFTRoutes && (
+        <RowBetween gap="sm">
```
natew, line 308: @callil this diff scares me did, did you add RowBetween or maybe theres a merge issue?

JFrankfurt:
@callil can we close this in favor of #7118 ?

natew:
yep

PR #7046, opened 2023-07-28
test(cypress): clean up types/assumptions/infura

zzmp:
## Description

This accidentally grew into a general cypress cleanup PR:
- Adds typechecking for cypress to the CI. While there _was_ a tsconfig, typechecking was never actually run.
  - As a corollary, extends the tsconfig from the main tsconfig.
  - As a corollary, moves `RouterPreference` to `state/routing/types`.
- Removes the `{ ethereum: 'hardhat' | 'goerli' }` parameter from `cy.visit` and removes the old mock injected wallet.
- Disables infura for cypress tests (by returning 404's), making it easier to reason about e2e tests.
- Disables video recording for cypress tests (retains screenshotting on failures) because it is documented to slow down the tests (_a lot_): tests require 2 CPUs, recording requires +1 CPU, GH Actions only have 2 CPUs.

With these changes:
- Tests run in under 5m!
- Tests are typechecked!
- There's no more goerli nor infura, only hardhat!

### Automated testing

- [ ] Unit test
- [X] Integration/E2E test


just-toby (approved):

cypress/tsconfig.json
```diff
@@ -1,18 +1,13 @@
 {
+  "extends": "../tsconfig.json",
   "compilerOptions": {
-    "esModuleInterop": true,
+    "composite": false,
     "incremental": true,
-    "lib": ["DOM", "DOM.Iterable", "ESNext"],
-    "noEmit": true,
-    "strict": true,
```
natew, line 7: yay no strict for tests!

natew (approved):

cypress/tsconfig.json
zzmp: Erm i think it is strict, just inherited. 

Tbh I'm not sure what this does, but if you're against it you should pr it to explicitly not be strict with a comment so it doesn't backslide.

PR #7051, opened 2023-07-31
feat: time-to-swap metric

just-toby:
## Description

implement a "time to swap" metric which we log for the first successful swap per session

We're resurrecting an unused event name for this, `SWAP_TRANSACTION_COMPLETED`, because it can be used more generically and brings web to parity w/ mobile (discussed w/ blair on this). 

using the `performance` [web api](https://developer.mozilla.org/en-US/docs/Web/API) we track the time since initial page load until the app identifies the success of a swap. this requires the app to remain open until the on-chain confirmation happens and is recognized.

as a follow up we'll instrument intermediate steps in the flow and include those timings in this logging.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2533/implement-a-time-to-swap-metric
_Relevant docs:_ https://www.notion.so/uniswaplabs/Time-to-Swap-TTS-Metric-556a912a6fb34199902e9f273fdaaf0b

### QA (ie manual testing)

- [x] manually verified that the `tts` property exists on `SWAP_TRANSACTION_COMPLETED` using amplitude's debug mode and inspecting the console.log .... and made sure that it only appeared on the first swap in a session.

### Automated testing

- [x] Integration/E2E test


src/state/transactions/updater.tsx
```diff
@@ -54,6 +57,24 @@ export default function Updater() {
...
 
+      const ttsMarks = performance.getEntriesByName('tts-mark', 'mark')
+      let elapsedTime: number | undefined
+      // Only log this metric for the first swap of a session.
+      if (ttsMarks.length === 0) {
+        elapsedTime = performance.now() / 1000 // time-to-swap in seconds
+        performance.mark('tts-mark')
+      }
+      if (elapsedTime && elapsedTime > 60 * 5) {
+        // 5 minute maximum
```
just-toby, line 68: open to suggestions on the timeout here

tinaszheng, line 68: think this would all be cleaner in a helper function that just returns `undefined` if there is > 1 entry 

and a bit confused on the 5 min metric here - is this meant to filter out swaps that just take foreverrr to get completed? i feel like we should still log those and just let data filter out according to whatever threshold they want?

just-toby: true, we can just filter on the backend!

cypress/support/commands.ts
```diff
@@ -66,3 +86,28 @@ Cypress.Commands.overwrite(
       )
   }
 )
+
+Cypress.Commands.add('waitForAmplitudeEvent', (eventName, options) => {
+  const { alias = '@analytics', timeout = 20000, shouldTimeout = false } = options ?? {}
+  const startTime = new Date().getTime()
```
JackShort, line 92: nit: is there a reason to use `Date` instead of `performance.now()` here?

JackShort (approved):

src/state/transactions/updater.tsx
```diff
@@ -24,7 +26,19 @@ export function toSerializableReceipt(receipt: TransactionReceipt): Serializable
...
+  const ttsMarks = performance.getEntriesByName('tts-mark', 'mark')
+  let elapsedTime: number | undefined
+  // Only log this metric for the first swap of a session.
+  if (ttsMarks.length === 0) {
+    elapsedTime = performance.now() / 1000 // time-to-swap in seconds
+    performance.mark('tts-mark')
+  }
+  return elapsedTime
+}
+
```
tinaszheng, line 39:
```suggestion
function getElapsedTime(): number | undefined {
  const ttsMarks = performance.getEntriesByName('tts-mark', 'mark')
  if (ttsMarks.length > 0) return
  
  performance.mark('tts-mark')
  return performance.now() / 1000 // time-to-swap in seconds
}

```

cypress/support/commands.ts
just-toby: i don't think it makes sense to use `performance.now()` here. the timeout is for the cypress command itself, not time since page load

src/state/transactions/updater.tsx
```diff
@@ -24,7 +26,20 @@ export function toSerializableReceipt(receipt: TransactionReceipt): Serializable
   }
 }
 
+function getElapsedTime(): number | undefined {
+  // Only log this metric for the first swap of a session.
+  const ttsMarks = performance.getEntriesByName('tts-mark', 'mark')
+
+  if (ttsMarks.length > 0) {
```
zzmp, line 33: This is not reliable, as the performance buffer will empty over time. Instead, you should use a `once` pattern, where you have a local variable that you set so the logic is only executed once. See https://developer.mozilla.org/en-US/docs/Web/API/Performance/setResourceTimingBufferSize.

src/state/transactions/updater.tsx
```diff
@@ -24,7 +26,20 @@ export function toSerializableReceipt(receipt: TransactionReceipt): Serializable
...
+    return
+  }
+
+  performance.mark('tts-mark')
```
zzmp, line 37:
```suggestion
  performance.mark('time-to-swap')
```

src/state/transactions/updater.tsx
```diff
@@ -24,7 +26,20 @@ export function toSerializableReceipt(receipt: TransactionReceipt): Serializable
...
+  return performance.now() / 1000 // time-to-swap in seconds
```
zzmp, line 38: You're re-deriving the timing already captured in the mark. Instead, use the value from the mark returned by `performance.mark`.

src/state/transactions/updater.tsx
```diff
@@ -54,6 +69,12 @@ export default function Updater() {
         })
       )
 
+      sendAnalyticsEvent(SwapEventName.SWAP_TRANSACTION_COMPLETED, {
+        tts: getElapsedTime(),
```
zzmp, line 73: why not call this time_to_swap to be more verbose?

src/state/transactions/updater.tsx
```diff
@@ -24,7 +26,20 @@ export function toSerializableReceipt(receipt: TransactionReceipt): Serializable
   }
 }
 
+function getElapsedTime(): number | undefined {
+  // Only log this metric for the first swap of a session.
```
zzmp, line 30: This should be a doc-comment on the function, so it is obvious that the function is only meant to return elapsed time from the first time it is called.

zzmp (changes requested):

cypress/support/commands.ts
```diff
@@ -66,3 +86,28 @@ Cypress.Commands.overwrite(
...
+      if (eventFound) {
+        return cy.wrap(eventFound)
+      } else if (new Date().getTime() - startTime > timeout) {
+        if (shouldTimeout) {
+          return cy.log(`Event ${eventName} not found within the specified timeout, as expected.`)
+        } else {
+          throw new Error(`Event ${eventName} not found within the specified timeout`)
+        }
+      } else {
+        return checkRequest()
```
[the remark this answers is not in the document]
just-toby: no, amplitude can send a bunch of network requests and we don't know which one will contain the event we're waiting for, or how many will occur before our event is sent.

an alternative approach would be to maintain a history of all the events sent during the test, do you think that's better?

cypress/e2e/swap/timeToSwap.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    cy.contains('Swap submitted')
+    cy.get(getTestSelector('confirmation-close-icon')).click()
+    cy.contains('Swap submitted').should('not.exist')
+    cy.get(getTestSelector('web3-status-connected')).should('contain', '1 Pending')
+
+    // Mine transaction
+    cy.hardhat().then((hardhat) => hardhat.mine(1))
+    cy.wait('@eth_getTransactionReceipt')
+    cy.get(getTestSelector('popups')).contains('Swapped')
+    cy.waitForAmplitudeEvent(SwapEventName.SWAP_TRANSACTION_COMPLETED, { shouldTimeout: true })
```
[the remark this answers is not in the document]
just-toby: depends on the implementation we take here - see other comment. 

this is just a way to verify that the event wasn't sent

cypress/support/commands.ts
```diff
@@ -12,6 +12,26 @@ declare global {
...
     }
+    interface Chainable<Subject> {
+      /**
+       * Wait for a specific event in a series of network requests. If the event is found, the subject will be the event.
+       *
+       * @param {string} eventName - The type of the event to search for.
+       * @param {object} [options] - Options to configure the event search.
+       *  {string} alias - The alias of the intercepted network request.
+       *  {number} [timeout=20000] - The maximum amount of time (in ms) to wait for the event.
+       *  {boolean} [shouldTimeout = false] - Whether the event should not be found within the timeout.
```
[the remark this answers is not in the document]
just-toby: removed

cypress/support/commands.ts
```diff
@@ -66,3 +86,28 @@ Cypress.Commands.overwrite(
...
+      const eventFound = events.find((event: any) => event.event_type === eventName)
+
+      if (eventFound) {
+        return cy.wrap(eventFound)
+      } else if (new Date().getTime() - startTime > timeout) {
+        if (shouldTimeout) {
+          return cy.log(`Event ${eventName} not found within the specified timeout, as expected.`)
+        } else {
+          throw new Error(`Event ${eventName} not found within the specified timeout`)
+        }
```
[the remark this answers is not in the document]
just-toby: removed `shouldTimeout`

cypress/support/commands.ts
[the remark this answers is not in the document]
just-toby: this is irrelevant now, resolving

src/state/transactions/updater.tsx
```diff
@@ -24,9 +26,22 @@ export function toSerializableReceipt(receipt: TransactionReceipt): Serializable
   }
 }
 
+/**
+ * Returns the time elapsed between page load and now,
+ * if the time-to-swap mark doesn't already exist.
+ *
+ * We only log the time-to-swap metric for the first swap of a session.
```
zzmp, line 33: This is no longer relevant to this function.
```suggestion
```

src/state/transactions/updater.tsx
```diff
@@ -24,9 +26,22 @@ export function toSerializableReceipt(receipt: TransactionReceipt): Serializable
...
+function getElapsedTime(): number {
+  const timeToSwap = performance.mark('time-to-swap')
+  return timeToSwap.startTime
+}
+
 export default function Updater() {
+  const analyticsContext = useTrace()
   const { chainId } = useWeb3React()
   const addPopup = useAddPopup()
+  const [timeToSwap, setTimeToSwap] = useState<number | undefined>(undefined)
```
zzmp, line 44: This doesn't need to be a part of the React lifecycle. In other words, you can move this to a file-level variable, `let hasReportedTimeToSwap = false`.

cypress/e2e/swap/timeToSwap.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    cy.hardhat()
+
+    // First swap in the session:
+    // Enter amount to swap
+    cy.get('#swap-currency-output .token-amount-input').type('1').should('have.value', '1')
+    cy.get('#swap-currency-input .token-amount-input').should('not.have.value', '')
+
+    // Submit transaction
+    cy.get('#swap-button').click()
+    cy.contains('Review swap')
```
zzmp, line 18:
```suggestion
```

cypress/e2e/swap/timeToSwap.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    cy.contains('Confirm swap').click()
+    cy.wait('@eth_estimateGas').wait('@eth_sendRawTransaction').wait('@eth_getTransactionReceipt')
```
zzmp, line 20:
```suggestion
```

cypress/e2e/swap/timeToSwap.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    })
+
+    // Second swap in the session:
+    // Enter amount to swap
+    cy.get('#swap-currency-output .token-amount-input').type('1').should('have.value', '1')
+    cy.get('#swap-currency-input .token-amount-input').should('not.have.value', '')
+
+    // Submit transaction
+    cy.get('#swap-button').click()
+    cy.contains('Review swap')
```
zzmp, line 39:
```suggestion
```

cypress/e2e/swap/timeToSwap.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    cy.contains('Confirm swap').click()
+    cy.wait('@eth_estimateGas').wait('@eth_sendRawTransaction').wait('@eth_getTransactionReceipt')
```
zzmp, line 41:
```suggestion
```

cypress/support/commands.ts
```diff
@@ -12,6 +12,16 @@ declare global {
     interface ApplicationWindow {
       ethereum: Eip1193Bridge
     }
+    interface Chainable<Subject> {
+      /**
+       * Wait for a specific event to be sent to amplitude. If the event is found, the subject will be the event.
+       *
+       * @param {string} eventName - The type of the event to search for e.g. SwapEventName.SWAP_TRANSACTION_COMPLETED
+       * @param {number} [timeout=20000] - The maximum amount of time (in ms) to wait for the event.
```
zzmp, line 20: instead of documenting the timeout, make it a default value in the definition.

cypress/support/commands.ts
```diff
@@ -12,6 +12,16 @@ declare global {
...
+       * @returns {Chainable<Subject>}
+       */
+      waitForAmplitudeEvent(eventName: string, timeout?: number): Chainable<Subject>
```
zzmp, line 23:
```suggestion
      waitForAmplitudeEvent(eventName: string, timeout = 20000): Chainable<Subject>
```

cypress/support/commands.ts
```diff
@@ -66,3 +76,23 @@ Cypress.Commands.overwrite(
       )
   }
 )
+
+Cypress.Commands.add('waitForAmplitudeEvent', (eventName, timeout = 5000 /* 5s */) => {
```
zzmp, line 80: This timeout default does not match the typing/documentation.

zzmp (changes requested):
down to nits, but the functionality looks 🔥

zzmp (approved):

JackShort (approved):

PR #7052, opened 2023-07-31
fix: init pool behavior

gnewfield:
## Description

Improve new pool creation experience by:
1. Disabling increment / decrement buttons when min and max prices are undefined (i.e. when buttons have no effect)
2. Adding "Full Range" button, which allows the user to spread liquidity evenly across the entire curve in a V2 style

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2444/disable-plus-and-minus-buttons-when-starting-a-liquidity-pool

### Before
| Mobile Before | Desktop Before |

### After
| Mobile After  | Desktop After |

### QA (ie manual testing)
1. Visit the Add Liquidity page (V3)
2. Select a token that has at least one uninitialized fee tier
3. With this change applied, the increment / decrement buttons should be disabled until the user inputs starting values for the min / max prices
4. The "Full Range" button should be present and enabled; when clicked, it populates the min price input with zero and the max price input with infinity


just-toby (approved):
can you verify the button is disabled is the `add-liquidity` cypress test? (i haven't worked on this page or that test much, so just checking if it's a reasonable test to add at this moment)

src/components/RangeSelector/index.tsx
```diff
@@ -49,8 +49,8 @@ export default function RangeSelector({
           width="48%"
           decrement={isSorted ? getDecrementLower : getIncrementUpper}
           increment={isSorted ? getIncrementLower : getDecrementUpper}
-          decrementDisabled={ticksAtLimit[isSorted ? Bound.LOWER : Bound.UPPER]}
-          incrementDisabled={ticksAtLimit[isSorted ? Bound.LOWER : Bound.UPPER]}
+          decrementDisabled={leftPrice === undefined || ticksAtLimit[isSorted ? Bound.LOWER : Bound.UPPER]}
+          incrementDisabled={leftPrice === undefined || ticksAtLimit[isSorted ? Bound.LOWER : Bound.UPPER]}
```
natew, line 53: any reason for this? should not be necessary it seems from the types

natew: nm me

natew (approved):

gnewfield:
> can you verify the button is disabled is the `add-liquidity` cypress test? (i haven't worked on this page or that test much, so just checking if it's a reasonable test to add at this moment)

Great suggestion. I've added cypress tests for the increment / decrement buttons, as well as full range behavior.

cypress/e2e/add-liquidity.test.ts
```diff
@@ -52,4 +52,51 @@ describe('Add Liquidity', () => {
       cy.get('#add-liquidity-selected-fee .selected-fee-percentage').should('contain.text', '40% select')
     })
   })
+
+  it('disables increment and decrement until initial prices are inputted', () => {
+    // ETH / BITCOIN pool (0.05% tier not created)
```
just-toby, line 57: will this test start failing if that tier is created on mainnet?

just-toby (approved):
thanks for adding tests. just make sure the test can't suddenly start failing if the onchain state changes on mainnet IRL

cypress/e2e/add-liquidity.test.ts
gnewfield: You're correct that it will start failing if the pool becomes created on mainnet. I'm expanding and hardening the test suite for the entire LP flow in a follow-up commit. Will address this.

PR #7057, opened 2023-08-01
fix: wrap fiat value if too large

LunrEclipse:
## Description

Wraps `fiatValue` to scientific notion if over 10 trillion to prevent issue where text would be to large and push balance/max outside of the boxes. 

_Linear ticket:_ WEB-2424

### Before

image (2)

### After

Screenshot 2023-08-01 at 11 11 33 AM
Screenshot 2023-08-01 at 11 11 26 AM
Screenshot 2023-08-01 at 11 15 18 AM

### Automated testing

- [x] Unit test N/A

### DIFF
--- a/src/components/CurrencyInputPanel/FiatValue.tsx
+++ b/src/components/CurrencyInputPanel/FiatValue.tsx
@@ -7,6 +7,7 @@ import { MouseoverTooltip } from 'components/Tooltip'
 import { useMemo } from 'react'
 import styled from 'styled-components/macro'
 import { ThemedText } from 'theme'
+import { formatTransactionAmount } from 'utils/formatNumbers'
 import { warningSeverity } from 'utils/prices'
 
 const FiatLoadingBubble = styled(LoadingBubble)`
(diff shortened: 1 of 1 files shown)


tinaszheng (changes requested):
can you just add a new rule here instead? https://github.com/Uniswap/conedison/blob/main/src/format.ts#L181

src/components/CurrencyInputPanel/FiatValue.tsx
```diff
@@ -35,11 +36,19 @@ export function FiatValue({
     return <FiatLoadingBubble />
   }
 
+  let formattedValue = formatNumber(fiatValue.data, NumberType.FiatTokenPrice)
+  if (formattedValue.length > 18) {
+    //Removes the dollar sign at the start, commas, and decimal places
+    formattedValue = formattedValue.slice(1).replaceAll(',', '').slice(0, -4)
+    formattedValue = formatTransactionAmount(Number(formattedValue))
+    formattedValue = '$' + formattedValue + 'T'
```
cbachmeier, line 44: Bringing in @JackShort as this logic may not work for internationalization of currencies

src/components/CurrencyInputPanel/FiatValue.tsx
```diff
@@ -35,11 +36,19 @@ export function FiatValue({
     return <FiatLoadingBubble />
   }
 
+  let formattedValue = formatNumber(fiatValue.data, NumberType.FiatTokenPrice)
+  if (formattedValue.length > 18) {
+    //Removes the dollar sign at the start, commas, and decimal places
+    formattedValue = formattedValue.slice(1).replaceAll(',', '').slice(0, -4)
```
cbachmeier, line 42: several countries use `,` instead of a decimal point

cbachmeier (changes requested):

LunrEclipse:
> can you just add a new rule here instead? https://github.com/Uniswap/conedison/blob/main/src/format.ts#L181

Yeah that would make more sense LOL
https://github.com/Uniswap/conedison/pull/24

src/components/CurrencyInputPanel/FiatValue.tsx
LunrEclipse: Would localization issues be bypassed if I just updated this in `conedison` like @tinaszheng mentioned?

tinaszheng: yeah would be easier to let the `conedison` formatter take care of it since it allows us to pass in a `currency` type into the formatter

PR #7060, opened 2023-08-01
fix: Router label copy

just-toby:
## Description

fixes a bug in the `RouterLabel` component where we checked an unused field to determine if the routing was client-side. now we correctly use the `QuoteMethod` field. 

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2597/fix-routerlabel-copy-for-local-routing

### Before

image

### After

image

### Reproducing the error

1. load a quote with client side routing forced and notice it still says "API"

### QA (ie manual testing)

- [x] repeat above steps, notice it says "client" instead of "API"

### Automated testing

- [x] Unit test

### DIFF
--- a/src/components/RouterLabel/index.tsx
+++ b/src/components/RouterLabel/index.tsx
@@ -1,4 +1,4 @@
-import { InterfaceTrade } from 'state/routing/types'
+import { InterfaceTrade, QuoteMethod } from 'state/routing/types'
 import { isUniswapXTrade } from 'state/routing/utils'
 import { ThemedText } from 'theme'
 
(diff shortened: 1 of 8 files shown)


src/components/RouterLabel/index.tsx
gnewfield, line 1: Is `isUniswapXTrade` function working? When I select UniswapX as my preferred router, I'm still seeing Uniswap API in the label.
Screenshot 2023-08-03 at 4 56 46 PM

cartcrom: You'll only get a UniswapX quote if it's better than a classic trade. You can enable synthetic quotes in the feature flag settings to get them more often

tinaszheng (approved):
💯

cartcrom (approved):
🔥

src/components/RouterLabel/index.tsx
gnewfield: Nice thanks for clarifying!

PR #7063, opened 2023-08-02
fix: bring back flake retries for flaky test (only flakes ~33% of time)

natew:
We've been struggling with this one but may as well give it some breathing room until we invest more into figuring it out (maybe upgrade cypress and pray?)


zzmp (approved):

LunrEclipse (approved):
lgtm! only question is if this test is still flakey or it fixed itself?

PR #7069, opened 2023-08-02
fix: make sure MenuItem in navbar doesnt wrap text and overflow is cutoff if super long

natew:
## Description

Fix wrapping in navbar

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-825/[nav]-tabs-do-not-text-wrap
_Slack thread:_
_Relevant docs:_

### Before

CleanShot 2023-08-02 at 10 03 46@2x

### After

CleanShot 2023-08-02 at 10 04 03@2x

I made it handle super super long ones, though that shouldn't ever appear, if we had some accidental translation that was like 7 words I think it's better to make it cut off than just push the entire navbar out of screen, it's pretty generous:

CleanShot 2023-08-02 at 10 07 11@2x

### Reproducing the error

1. 

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test

### DIFF
--- a/src/components/NavBar/index.tsx
+++ b/src/components/NavBar/index.tsx
@@ -47,10 +47,18 @@ const MenuItem = ({ href, dataTestId, id, isActive, children }: MenuItemProps) =
       style={{ textDecoration: 'none' }}
       data-testid={dataTestId}
     >
-      {children}
+      <MenuItemText>{children}</MenuItemText>
     </NavLink>
   )
 }
+
+const MenuItemText = styled.span`
+  display: inline;
+  overflow: hidden;
+  max-width: 250px;
+  text-overflow: ellipsis;
+  white-space: pre;
+`
 
 export const PageTabs = () => {
   const { pathname } = useLocation()


just-toby (approved):

cbachmeier (changes requested):
This comes up frequently and needs more design input. The issue is with a breakpoints and internationalization, if we just don't wrap you get into states where menu items are cutoff. See the below screenshot where, when in Spanish, on mobile (397px), the overflow menu is now cutoff
Screenshot 2023-08-03 at 10 14 04 AM

src/components/NavBar/index.tsx
```diff
@@ -47,11 +47,19 @@ const MenuItem = ({ href, dataTestId, id, isActive, children }: MenuItemProps) =
       style={{ textDecoration: 'none' }}
       data-testid={dataTestId}
     >
-      {children}
+      <MenuItemText>{children}</MenuItemText>
     </NavLink>
   )
 }
 
+const MenuItemText = styled.span`
```
cbachmeier, line 55: This would also be a good time refactor the `NavLink` to not use vanilla extract and to handle text wrapping

natew:
@cbachmeier ah good catch on the mobile need, perhaps that could be a scrollable horizontal input, i'll ping design

natew:
@callil i think design needs some input before this can go forward

callil:
We should use the "overflow" behavior like we do for breakpoints on the nav... so in this case the nav should be aware of how long the string lengths are and the screen width. If they will overflow, put those menu items in the triple dot overflow menu.

natew:
Closing and added to backlog to revisit https://linear.app/uniswap/issue/WEB-2945/navbar-overflowing-text-should-automatically-move-into-the-menu

PR #7070, opened 2023-08-02
feat: fallback metadata injection

LunrEclipse:
## Description

Created a function in root directory to take any page that doesn't have custom metadata and injects a default with a title and image. 

_Linear ticket:_ WEB-2604

### Automated testing

- [x] Unit test - Tested with `yarn cloud:tests` and added a test suite `default.test.ts`


functions/collection.test.ts
```diff
@@ -38,9 +38,7 @@ test.each(collections)('should inject metadata for valid collections', async (co
 
 const invalidCollections = [
   'http://127.0.0.1:3000/nfts/collection/0xed5af388653567af2f388e6224dc7c4b3241c545',
-  'http://127.0.0.1:3000/nfts/collection/0xed5af388653567af2f388e6224dc7c4b3241c545/10',
```
zzmp, line 41: Why are you removing these? It seems unrelated.

functions/tsconfig.json
```diff
@@ -15,5 +15,5 @@
     "baseUrl": "functions"
   },
   "exclude": ["../node_modules", "../src"],
-  "include": ["**/*.ts"],
+  "include": ["**/*.ts", ".ts"],
```
zzmp, line 18: What does this change do?

zzmp (changes requested):

functions/collection.test.ts
LunrEclipse: those tests actually get caught by the fallback so it caused them to start failing

functions/[[index]].ts
```diff
@@ -0,0 +1,16 @@
...
+  const data = {
+    title: 'Uniswap Interface',
+    image: imageUri,
+    url: request.url,
+  }
+  try {
+    return new HTMLRewriter().on('head', new MetaTagInjector(data)).transform(await next())
+  } catch (e) {
+    return next()
+  }
```
Comment (zzmp) [inline functions/[[index]].ts:15]:
```suggestion
  const res = next()
  try {
    return new HTMLRewriter().on('head', new MetaTagInjector(data)).transform(await res)
  } catch (e) {
    return res
  }
```

functions/[[index]].ts
```diff
@@ -0,0 +1,16 @@
+import { MetaTagInjector } from './components/metaTagInjector'
+
+// eslint-disable-next-line import/no-unused-modules
```
Comment (zzmp) [inline functions/[[index]].ts:3]: Do you need this comment?

zzmp (changes requested):

functions/[[index]].ts
Comment (LunrEclipse) [reply functions/[[index]].ts:3]: yeye its just like the other functions

natew (approved):
functions are cool

zzmp (approved):

PR #7072, opened 2023-08-02
feat: bottom sheet for mobile swap settings

JackShort:
## Description

* added a bottom sheet popup menu for swap settings on mobile

_Linear ticket:_  https://linear.app/uniswap/issue/WEB-2520/settings-dialog-content-overflow-on-small-width-w-long-translations

### Before

Screen Shot 2023-08-02 at 4 48 54 PM

### After

Screen Shot 2023-08-02 at 4 48 28 PM


just-toby (changes requested):
add a cypress test that sets the viewport size and verifies that the correct view appears

just-toby (approved):

gnewfield:
I'm seeing a strange effect when the Slippage panel is toggled open. The Transaction deadline panel disappears for an instant (gets pushed down below the fold?), and then reappears when the panel is done opening. It's hard to notice, but I made a recording of the screen and confirmed. This is the case in prod as well, so not an issue with the bottom sheet but with the settings menu. I've made a backlog item for this in [WEB-2642](https://linear.app/uniswap/issue/WEB-2642/toggling-defect-in-the-settings-panel). Otherwise, changes look good.

gnewfield (approved):

PR #7074, opened 2023-08-02
fix: failing token cloud function tests

LunrEclipse:
Fixes failing native token cloud function tests by removing the conversion from `NATIVE` to zero address to instead `NATIVE` to `undefined` or a predefined address for specific chains.

### DIFF
--- a/functions/tokens/[[index]].ts
+++ b/functions/tokens/[[index]].ts
@@ -1,13 +1,14 @@
 /* eslint-disable import/no-unused-modules */
+import { Chain } from '../../src/graphql/data/__generated__/types-and-hooks'
 import { MetaTagInjector } from '../components/metaTagInjector'
 import getToken from '../utils/getToken'
 
 const convertTokenAddress = (tokenAddress: string, networkName: string) => {
   if (tokenAddress === 'NATIVE') {
     switch (networkName) {
-      case 'CELO':
+      case Chain.Celo:
         return '0x471EcE3750Da237f93B8E339c536989b8978a438'
-      case 'POLYGON':
+      case Chain.Polygon:
         return '0x0000000000000000000000000000000000001010'
       default:
         return undefined
--- a/functions/tokens/[[index]].ts
+++ b/functions/tokens/[[index]].ts
@@ -2,17 +2,28 @@
 import { MetaTagInjector } from '../components/metaTagInjector'
 import getToken from '../utils/getToken'
 
-const convertTokenAddress = (tokenAddress: string) => {
-  return tokenAddress && tokenAddress === 'NATIVE' ? '0x0000000000000000000000000000000000000000' : tokenAddress
+const convertTokenAddress = (tokenAddress: string, networkName: string) => {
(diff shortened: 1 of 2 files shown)


functions/tokens/[[index]].ts
```diff
@@ -2,17 +2,28 @@
...
 
-const convertTokenAddress = (tokenAddress: string) => {
-  return tokenAddress && tokenAddress === 'NATIVE' ? '0x0000000000000000000000000000000000000000' : tokenAddress
+const convertTokenAddress = (tokenAddress: string, networkName: string) => {
+  if (tokenAddress === 'NATIVE') {
+    switch (networkName) {
+      case 'CELO':
+        return '0x471EcE3750Da237f93B8E339c536989b8978a438'
+      case 'POLYGON':
+        return '0x0000000000000000000000000000000000001010'
```
Comment (zzmp) [inline functions/tokens/[[index]].ts:11]: Can you check with the backend team that this is comprehensive? What about their testnets?

zzmp (changes requested):

functions/tokens/[[index]].ts
Comment (cbachmeier) [reply functions/tokens/[[index]].ts:11]: We also have a map of wrapped native currencies `WRAPPED_NATIVE_CURRENCY` in `src/constants/tokens.ts` we could leverage

Comment (LunrEclipse) [reply functions/tokens/[[index]].ts:11]: @cbachmeier yeah i took a look at that, however, we don't necessarily want the `WRAPPED_NATIVE_CURRENCY`. For `Celo` its fine since they're the same, however, `Polygon` has a wrapped native currency `wMatic` and the native currency `Matic` and `WRAPPED_NATIVE_CURRENCY` would return the address for `wMatic` instead of `Matic`. I don't know if this is an edge case or not though.

functions/tokens/[[index]].ts
```diff
@@ -2,17 +2,28 @@
 import { MetaTagInjector } from '../components/metaTagInjector'
 import getToken from '../utils/getToken'
 
-const convertTokenAddress = (tokenAddress: string) => {
-  return tokenAddress && tokenAddress === 'NATIVE' ? '0x0000000000000000000000000000000000000000' : tokenAddress
+const convertTokenAddress = (tokenAddress: string, networkName: string) => {
+  if (tokenAddress === 'NATIVE') {
+    switch (networkName) {
+      case 'CELO':
```
Comment (just-toby) [inline functions/tokens/[[index]].ts:8]: can you use `Chain.CELO` etc for these (the enum is in the gQL codegen)

just-toby (changes requested):
use the enum already defined for chain names

functions/tokens/[[index]].ts
Comment (LunrEclipse) [reply functions/tokens/[[index]].ts:11]: I was looking at `getNativeTokenDBAddress` in `src/constants/nativeTokens.ts` but saw that only `Celo` and `Polygon` use a different native token which is why I just opted to hard code those 2 chains

just-toby (approved):

zzmp (approved):

PR #7075, opened 2023-08-03
chore: e2e insufficient liquidity test

JackShort:
* adds e2e test for insufficient liquidity - downside is it takes a long time for our API to return an insufficient liquidity route

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2096/e2e-test-for-insufficient-liquidity-in-swap-flow


cypress/e2e/swap/errors.test.ts
```diff
@@ -116,4 +116,12 @@ describe('Swap errors', () => {
       getBalance(DAI).should('be.closeTo', initialBalance + 200, 1)
     })
   })
+
+  it('insufficient liquidity', () => {
+    cy.visit(`/swap?inputCurrency=${USDC_MAINNET.address}&outputCurrency=${DAI.address}`)
+    // This is assuming there is not > 100 Trillion Dai in a pool
+    cy.get('#swap-currency-output .token-amount-input').type('100000000000000').should('have.value', '100000000000000')
+    cy.contains('Insufficient liquidity for this trade.')
```
zzmp, line 124: If it's known that our API takes a long time to resolve this, you should explicitly add a timeout here to account for that. Otherwise this test will be flaky.

If you do extend the timeout, leave a comment explaining why:

```suggestion
    // The API takes X seconds to return a 404, so we extend the timeout.
    cy.contains('Insufficient liquidity for this trade.', { timeout: 48000 ))
```

zzmp (changes requested):

cypress/e2e/swap/errors.test.ts
```diff
@@ -116,4 +116,21 @@ describe('Swap errors', () => {
...
+    cy.visit(`/swap?inputCurrency=${USDC_MAINNET.address}&outputCurrency=${DAI.address}`)
+    // The API response is too variable so stubbing a 404.
+    cy.intercept('POST', 'https://api.uniswap.org/v2/quote', (req) => {
+      req.reply({
+        statusCode: 404,
+        fixture: 'insufficientLiquidity.json',
+        headers: {
+          'access-control-allow-origin': '*',
+        },
+      })
```
zzmp, line 130: I think this can be simplified to not use an anonymous function, see https://docs.cypress.io/api/commands/intercept#With-a-fixture.

zzmp: also, try moving the stub to before the `visit`. That way if it runs a quote on load it will still be intercepted.

JackShort:
For some reason intercepting didn't work so changed back to increasing the timeout

cypress/e2e/swap/errors.test.ts
```diff
@@ -116,4 +116,12 @@ describe('Swap errors', () => {
       getBalance(DAI).should('be.closeTo', initialBalance + 200, 1)
     })
   })
+
+  it('insufficient liquidity', () => {
+    cy.visit(`/swap?inputCurrency=${USDC_MAINNET.address}&outputCurrency=${DAI.address}`)
+    cy.get('#swap-currency-output .token-amount-input').type('100000000000000').should('have.value', '100000000000000') // 100 trillion
+    // The API takes a long and variable amount of time to return a 404, so setting the timeout to 60 seconds.
+    cy.contains('Insufficient liquidity for this trade.', { timeout: 60000 })
```
zzmp, line 124: this will potentially add a minute to our CI time - want to pair on getting `intercept` to work?

zzmp (changes requested):

JackShort:
it now seems to be working correctly after merging with main - unsure what happens but it seems to be intercepting now

cypress/fixtures/insufficientLiquidity.json
```diff
@@ -0,0 +1,5 @@
+{
```
just-toby, line 1: format this?

cypress/e2e/swap/insufficientLiquidity.test.ts
```diff
@@ -0,0 +1,18 @@
...
+  beforeEach(() => {
+    cy.intercept('POST', 'https://api.uniswap.org/v2/quote', {
+      statusCode: 404,
+      fixture: 'insufficientLiquidity.json',
+    })
+  })
+
+  it('insufficient liquidity', () => {
+    cy.visit(`/swap?inputCurrency=${USDC_MAINNET.address}&outputCurrency=${DAI.address}`)
+    // The API response is too variable so stubbing a 404.
```
just-toby, line 13: move this comment up to the place where it's happening?

just-toby (approved):

cypress/e2e/swap/insufficientLiquidity.test.ts
```diff
@@ -0,0 +1,18 @@
+import { DAI, USDC_MAINNET } from '../../../src/constants/tokens'
+
+describe('Swap errors (no liquidity)', () => {
```
zzmp, line 3: Shouldn't this be in cypress/e2e/swap/errors.test.ts?

zzmp (changes requested):
lgtm but I think it belongs in an existing file

PR #7084, opened 2023-08-04
build: simplify/fix test caching

zzmp:
## Description

Removes overly complex cache-on-main, which was implemented as an optimization to speed up CI actions.
Correctly caches hardhat calls, by fanning-in the hardhat cache directory using upload/download-artifact before caching it.


natew (approved):
ship it

cbachmeier (approved):

PR #7088, opened 2023-08-04
fix: revert #6928

just-toby:
reverts #6928 to resolve https://uniswaplabs.pagerduty.com/incidents/Q332GT7AO8MQ6J


zzmp (changes requested):
I think we can do a better job fixing this before the next staging cut, so I don't think we should merge a straight revert.

just-toby:
closing to wait for a better fix

PR #7090, opened 2023-08-05
fix: keep a referentially stable connectors list

zzmp:
## Description

Integrates the WCv2 connector with the React lifecycle, by:
- making all of our web3-react connections referentially stable (`getConnections()`->`connections`)
- proxying the `connector` and `hooks`, so they always refer to the active connection
- triggering React renders for new active connections with `useSyncExternalStore`

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2445/connecting-wallet-should-not-clear-inputs-in-swap
_Slack thread:_ https://uniswapteam.slack.com/archives/C047U65H422/p1691185006320819
_Relevant docs:_ [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy), [useSyncExternalStore](https://react.dev/reference/react/useSyncExternalStore)

### Reproducing the error

1. Connect with WCv2 - crash!
- Also note that we want to not clear inputs when switching wallet connection state

### QA (ie manual testing)

- Fill out a swap
- Connect a wallet
- Disconnect the wallet
- Connect another wallet
- Make sure to use WCv2 at some point there

### Automated testing

- [X] Unit test
- [X] Integration/E2E test

Updated existing unit tests. This is already covered by our integration tests, in so far as we can test wallet connections :/.


gnewfield (commented):
LGTM, manually verified as well

gnewfield (approved):

PR #7094, opened 2023-08-07
feat: add chainId to SWAP_TRANSACTION_COMPLETED

just-toby:
## Description

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2535/as-a-followup-segment-the-steps-in-the-swap-flow-so-we-can-see-the

## Screen capture

chainId addition:
image

### QA (ie manual testing)

- [x] manually verified that the fields are in the event as expected, see screenshot above

### Automated testing

- [x] Unit test N/A
- [x] Integration/E2E test - added


src/tracing/SignedTransactionTimestampRegistry.ts
```diff
@@ -0,0 +1,35 @@
...
+      // If a mark already exists for this hash, remove it before creating a new one
+      performance.clearMarks(markName)
+    }
+    performance.mark(markName)
+  }
+
+  public get(hash: string): number | undefined {
+    const markName = this.getMarkNameForHash(hash)
+    const marks = performance.getEntriesByName(markName, 'mark')
+    return marks?.length > 0 ? marks[0].startTime : undefined
```
tinaszheng, line 28: think you can just shorthand this with `marks[0]?.startTime`

src/hooks/useUniswapXSwapCallback.ts
```diff
@@ -89,6 +89,7 @@ export function useUniswapXSwapCallback({
             fiatValues,
           }),
           ...analyticsContext,
+          time_to_sign: performance.mark('time-to-sign-uniswapx').startTime,
```
tinaszheng, line 92: why not just use the difference between Date.now() before sending the request on line 69 and after

src/state/transactions/updater.tsx
```diff
@@ -75,6 +76,8 @@ export default function Updater() {
       sendAnalyticsEvent(SwapEventName.SWAP_TRANSACTION_COMPLETED, {
         // if timeToSwap was already set, we already logged this session
         time_to_swap: hasReportedTimeToSwap ? undefined : elapsedTime,
+        time_to_sign: hasReportedTimeToSwap ? undefined : SignedTransactionTimestampRegistry.getInstance().get(hash),
```
tinaszheng, line 79: i dot think this is right, this measures the amount of time it takes post sign -> tx is successful, whereas your time to sign should just be pre-signature -> post-signature

tinaszheng (changes requested):

tinaszheng (approved):

cartcrom (approved):

PR #7095, opened 2023-08-07
feat: add amplitude logging for gas estimate failures

cbachmeier:
## Description

- We've noticed an increase in swap failures so we're adding an analytics event if the `estimateGas` call fails for the data team to use for offline analysis. 
- This event adds data for swaps that fail before making it on chain. 
- Confirmed data points with @blairmason. 
- Added the new swap event to shared package [here](https://github.com/Uniswap/analytics-events/pull/49) and updated depedency

_Linear ticket:_ https://www.notion.so/uniswaplabs/Swap-fee-on-transfer-failures-aebe8b9206cf41fe9b8b8f8b26613a43
_Slack thread:_ https://uniswapteam.slack.com/archives/C05JRUHDD4L/p1691431753896049


gnewfield (approved):

src/hooks/useUniversalRouter.ts
```diff
@@ -80,6 +80,10 @@ export function useUniversalRouterSwapCallback(
         } catch (gasError) {
           setTraceStatus('failed_precondition')
           setTraceError(gasError)
+          sendAnalyticsEvent(SwapEventName.SWAP_ESTIMATE_GAS_CALL_FAILED, {
+            ...formatCommonPropertiesForTrade(trade, options.slippageTolerance),
```
tinaszheng, line 84: We'll need the tx request object here to see why it failed - can you include that too?

tinaszheng (changes requested):

src/hooks/useUniversalRouter.ts
tinaszheng: also the `gasError` object if possible

tinaszheng (approved):

PR #7098, opened 2023-08-07
fix: check chainId in findCancelTx

just-toby:
## Description

We are marking transactions as cancelled if we detect the same nonce but a different hash - even if that other transaction is on a different chain.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2621/screen-bugs-out-when-making-base-tx-while-mini-port-is-open

### Before

https://github.com/Uniswap/interface/assets/66155195/7efa4deb-f9dd-4712-936f-302345fec0e6

### After

https://github.com/Uniswap/interface/assets/66155195/996670de-1adb-49aa-8ba3-a33d85070cd8

### Reproducing the error

1. make a swap on base with the MP activity tab open

### QA (ie manual testing)

- [x] make a swap on base with the MP activity tab open, ensure it's correct. and that previously "cancelled" txs are not cancelled

### DIFF
--- a/src/components/AccountDrawer/MiniPortfolio/Activity/hooks.ts
+++ b/src/components/AccountDrawer/MiniPortfolio/Activity/hooks.ts
@@ -19,7 +19,8 @@ function findCancelTx(localActivity: Activity, remoteMap: ActivityMap, account: 
     if (
       remoteTx.nonce === localActivity.nonce &&
       remoteTx.from.toLowerCase() === account.toLowerCase() &&
-      remoteTx.hash.toLowerCase() !== localActivity.hash.toLowerCase()
+      remoteTx.hash.toLowerCase() !== localActivity.hash.toLowerCase() &&
+      remoteTx.chainId === localActivity.chainId
     ) {
       return remoteTx.hash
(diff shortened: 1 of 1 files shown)


zzmp (changes requested):
Tests

src/components/AccountDrawer/MiniPortfolio/Activity/hooks.ts
```diff
@@ -37,6 +38,10 @@ function combineActivities(localMap: ActivityMap = {}, remoteMap: ActivityMap =
     const remoteActivity = (remoteMap?.[hash] ?? {}) as Activity
 
     if (localActivity.cancelled) {
+      if (localActivity.chainId !== remoteActivity.chainId) {
```
zzmp, line 41: This needs a comment to explain, as you're guarding against past incorrectly cancelled tx's manifesting as hash collisions, not actual hash collisions.

zzmp (changes requested):

zzmp (approved):

cartcrom (approved):

PR #7102, opened 2023-08-08
fix: add back correct fellback setter

JFrankfurt:
https://uniswapteam.slack.com/archives/C02T729PMQE/p1691507534359689?thread_ts=1691444333.641919&cid=C02T729PMQE

looks like this line got dropped in the merge


tinaszheng (approved):
💯 ty

cartcrom (approved):

PR #7110, opened 2023-08-08
test: e2e UniswapX orders

cartcrom:
## Description

adds e2e tests for uniswapx orders, including eth-input orders. tests different order status responses' effects on the UI

### DIFF
--- a/cypress/e2e/swap/uniswapx.test.ts
+++ b/cypress/e2e/swap/uniswapx.test.ts
@@ -8,21 +8,14 @@ const QuoteWhereUniswapXIsBetter = 'uniswapx/quote1.json'
 const QuoteWithEthInput = 'uniswapx/quote2.json'
 
 const OrderSubmissionEndpoint = 'https://api.uniswap.org/v2/order'
-const OrderResponse = 'uniswapx/orderResponse.json'
 
 const OrderStatusEndpoint =
   'https://api.uniswap.org/v2/orders?swapper=0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266&orderHashes=0xa9dd6f05ad6d6c79bee654c31ede4d0d2392862711be0f3bc4a9124af24a6a19'
-const OpenStatusResponse = 'uniswapx/openStatusResponse.json'
-const FilledStatusResponse = 'uniswapx/filledStatusResponse.json'
-const ExpiredStatusResponse = 'uniswapx/expiredStatusResponse.json'
-const TransactionReceipt = 'uniswapx/fillTransactionReceipt.json'
-
-const InsufficientFundsStatusResponse = 'uniswapx/insufficientFundsStatusResponse.json'
 
 /** Stubs the provider to return a tx receipt corresponding to the mock filled uniswapx order's txHash */
 function stubSwapTxReceipt() {
   cy.hardhat().then((hardhat) => {
-    cy.fixture(TransactionReceipt).then((mockTxReceipt) => {
+    cy.fixture('uniswapx/fillTransactionReceipt.json').then((mockTxReceipt) => {
       const getTransactionReceiptStub = cy.stub(hardhat.provider, 'getTransactionReceipt').log(false)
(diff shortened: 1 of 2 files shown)


src/pages/Swap/index.tsx
```diff
@@ -558,7 +558,7 @@ export function Swap({
         showCancel={true}
       />
       <SwapHeader trade={trade} autoSlippage={autoSlippage} chainId={chainId} />
-      {trade && showConfirm && (
+      {trade && showConfirm && allowance.state !== AllowanceState.LOADING && (
```
cartcrom, line 561: @just-toby since the ui gets a response so quickly from the mocked quote endpoint, confirm swap was sometimes getting clicked before allowances had been loaded in, skipping the approval step of the flow. I added in this check to prevent that, just want to check w/ you that this seems like the right place to do it

just-toby: does `allowance` ever reload after the modal is already open? if we're refreshing the allowance and it's `LOADING` during the flow maybe the modal will auto-close...

just-toby: just tested and it didn't happen

just-toby (approved):

cypress/e2e/swap/uniswapx.test.ts
```diff
@@ -1,32 +1,55 @@
...
+
+const OrderSubmissionEndpoint = 'https://api.uniswap.org/v2/order'
+const OrderResponse = 'uniswapx/orderResponse.json'
+
+const OrderStatusEndpoint =
+  'https://api.uniswap.org/v2/orders?swapper=0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266&orderHashes=0xa9dd6f05ad6d6c79bee654c31ede4d0d2392862711be0f3bc4a9124af24a6a19'
+const OpenStatusResponse = 'uniswapx/openStatusResponse.json'
+const FilledStatusResponse = 'uniswapx/filledStatusResponse.json'
+const ExpiredStatusResponse = 'uniswapx/expiredStatusResponse.json'
+const TransactionReceipt = 'uniswapx/fillTransactionReceipt.json'
```
zzmp, line 18: nit: I think you could inline all of these fixture paths. they're already self-documented from the filenames.

cypress/e2e/swap/uniswapx.test.ts
```diff
@@ -1,32 +1,55 @@
...
+  })
 }
 
 describe('UniswapX Toggle', () => {
   beforeEach(() => {
     cy.intercept(QuoteEndpoint, { fixture: QuoteWhereUniswapXIsBetter })
     cy.visit(`/swap/?inputCurrency=${USDC_MAINNET.address}&outputCurrency=${DAI.address}`)
+
+    // Hide banner that partially covers UniswapX mustache
+    cy.get(getTestSelector('uniswap-wallet-banner')).click()
```
zzmp, line 39: This should be hidden by setting the localStorage bit that says it's been seen already, not by clicking on it. Otherwise any other changes to it will cause this test to fail.

I don't think the banner is tested in any e2e tests, so it would be fine to make this a global thing in setup.

cypress/e2e/swap/uniswapx.test.ts
```diff
@@ -50,7 +70,166 @@ describe('UniswapX Toggle', () => {
...
+  beforeEach(() => {
+    cy.intercept(QuoteEndpoint, { fixture: QuoteWhereUniswapXIsBetter })
+    cy.intercept(OrderSubmissionEndpoint, { fixture: OrderResponse })
+    cy.intercept(OrderStatusEndpoint, { fixture: OpenStatusResponse })
+
+    stubSwapTxReceipt()
+
+    cy.hardhat().then(async (hardhat) => {
+      await hardhat.fund(hardhat.wallet, CurrencyAmount.fromRawAmount(USDC_MAINNET, 3e8))
+    })
```
zzmp, line 90: nit: I think you can return the promise directly to avoid async-await.

zzmp (approved):
some nits but looks great

PR #7121, opened 2023-08-09
fix: Remove double client side quoting

tinaszheng:
## Description
Since we added client-side routing to the query in [routing/slice](https://github.com/Uniswap/interface/blob/main/src/state/routing/slice.ts#L180), we don't need another catch-all in `useBestTrade` to do client-side routing too.

Because I removed the client side routing fallback, the `useBestTrade` hook is now just a debounce wrapper, so I renamed it to `useDebouncedTrade`

## Screen capture
No UI changes

## Test plan
manually tested with local routing on/off for regular tokens and insufficient liquidity tokens


src/hooks/useDebouncedTrade.test.ts
```diff
@@ -0,0 +1,72 @@
+import { renderHook } from '@testing-library/react'
```
tinaszheng, line 1: this file is the almost the same minus the client side hook tests and the filename

just-toby (approved):

LunrEclipse (approved):

src/constants/routing.ts
```diff
@@ -72,66 +59,6 @@ const WRAPPED_NATIVE_CURRENCIES_ONLY: ChainTokenList = Object.fromEntries(
...
-    DAI_POLYGON,
-    USDC_POLYGON,
-    USDT_POLYGON,
-    WETH_POLYGON,
-  ],
-  [ChainId.BNB]: [
-    ...WRAPPED_NATIVE_CURRENCIES_ONLY[ChainId.BNB],
-    DAI_BSC,
-    USDC_BSC,
-    USDT_BSC,
```
JFrankfurt, line 97: so these were not being used post-uniswapx merge?

JFrankfurt: I was looking through the smart-order-router package for these and I don't see them. I'm a bit concerned that the quality of client-side generated routes has declined.

It's fine to remove this code in this PR since it's dead rn, but I think we should investigate the internals on smart-order-router and ensure this is present, or that it can be surfaced to consumers as a config.

PR #6902, opened 2023-07-07
feat: dynamically generated images for rich link previews

LunrEclipse:
## Description

Created Cloudflare Workers to dynamically images for NFTs, NFT Collections, and Tokens. Created routed api calls in the functions folder under `functions/api/image`. Specifically, created three apis: `/asset/[[index]].ts`, `/tokens/[[index]]/ts`, and `collection/[index]/ts` that utilize GraphQL to query the appropriate information and generate images using Vercel's OG Image Generation library. 

Unit testing can be found in `collection.test.ts`, `nft.test.ts`, and `token.test.ts`. 

### QA (ie manual testing)

- [x] Manually start Cloudflare test environment using `yarn start:cloud` and navigate to the corresponding routes to check if metadata properly injected. Example: `https://127.0.0.1:3000/api/image/nfts/collection/0xed5af388653567af2f388e6224dc7c4b3241c544`

### Automated testing

- [x] Unit test `yarn test:cloud`

PR #7128, opened 2023-08-09
fix: downgrade wrangler

LunrEclipse:
Downgrades wrangler to more stable version and removes a flakey test


zzmp (changes requested):
Can you only make changes to `test.yml`? The other changes may or may not work to deflake, but will definitely create merge conflicts with #6902.

LunrEclipse:
> Can you only make changes to `test.yml`? The other changes may or may not work to deflake, but will definitely create merge conflicts with #6902.

I could but I don't think the merge conflicts will be that bad. #6902 already uses the same downgraded version of wrangler and the changes to `default.test.ts` aren't touched in the dynamic image pr

zzmp:
> > Can you only make changes to `test.yml`? The other changes may or may not work to deflake, but will definitely create merge conflicts with #6902.
> 
> I could but I don't think the merge conflicts will be that bad. #6902 already uses the same downgraded version of wrangler and the changes to `default.test.ts` aren't touched in the dynamic image pr

You're going to have merge conflicts with the yarn.lock.

LunrEclipse:
> > > Can you only make changes to `test.yml`? The other changes may or may not work to deflake, but will definitely create merge conflicts with #6902.
> > 
> > 
> > I could but I don't think the merge conflicts will be that bad. #6902 already uses the same downgraded version of wrangler and the changes to `default.test.ts` aren't touched in the dynamic image pr
> 
> You're going to have merge conflicts with the yarn.lock.

zzmp (approved):

package.json
```diff
@@ -142,7 +142,8 @@
     "typescript": "^4.9.4",
     "webpack": "^5.88.2",
     "webpack-retry-chunk-load-plugin": "^3.1.1",
-    "wrangler": "^3.4.0",
+    "//": "downgraded-because-vercel-og-incompatible",
```
natew, line 145: can you test if we can have multple "//" or if we need to add some sort of "// wrangler" here to avoid conflicts

natew (approved):

PR #7139, opened 2023-08-10
fix: make swc not break WalletConnect by targeting a newer version of es

natew:
fixes breaking wallet connect / uniswap wallet connect on main

had to leave it in package.json because otherwise craco fails.

to repro: it was erroring if you hit connect => uniswap wallet and not allowing connection, this should fix that

### DIFF
--- a/.swcrc
+++ b/.swcrc
@@ -1,5 +1,9 @@
 {
   "$schema": "https://json.schemastore.org/swcrc",
+  // has to duplicate from package.json, see swc issue: https://swc.rs/docs/configuration/compilation#env
+  "env": {
+    "targets": "> 0.5%, not dead"
+  },
   "jsc": {
     "keepClassNames": true,
     "experimental": {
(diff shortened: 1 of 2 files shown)


.swcrc
```diff
@@ -1,5 +1,13 @@
 {
   "$schema": "https://json.schemastore.org/swcrc",
+  // has to duplicate from package.json, see swc issue: https://swc.rs/docs/configuration/compilation#env
+  "env": {
+    "targets": {
+      "chrome": "112",
```
zzmp, line 6: could you use the percentage format and match what we have in our package.json?
also our browserslist could probably use updating.

zzmp (changes requested):

zzmp (approved):
Do we care about excluding op_mini?

natew:
got browser list form working but had to use 0.5 (0.25 still breaks) but that seems reasonable.

zzmp (approved):

cbachmeier (approved):

PR #7141, opened 2023-08-11
feat: custom colors for rich link previews

LunrEclipse:
Imports predefined colors from widgets to use for rich link previews.

Also updates `launchTimeout` due to increased flakiness caused by jest timing out before the server sets up properly.

Screenshot 2023-08-11 at 11 17 01 AM


tinaszheng (approved):

JFrankfurt (approved):

zzmp (commented):
Why are they hardcoded by image url instead of chain and token address?

LunrEclipse:
> Why are they hardcoded by image url instead of chain and token address?

I think it makes more sense to be by image. For example, WBTC could have the same image across multiple chains and different token addresses so rather than having for example 5 copies for WBTC we could just filter by image.

PR #7142, opened 2023-08-11
fix: remove description from custom rich link previews

LunrEclipse:
Removes default description from `index.html` and adds it as a custom input into `metaTagInjector.ts` so a description won't show on rich link previews such as nfts, nft collections, and tokens.

Also adds static assets (font and network logos) to `public` folder to migrate dependencies from other sites to our own in the future.


JackShort (approved):

zzmp (changes requested):
assets do not need to be a part of this PR - can you remove them?

zzmp (approved):

PR #7143, opened 2023-08-11
feat: swap quote latency logging

just-toby:
## Description

log `quote_latency_milliseconds` in the `SWAP_QUOTE_RECEIVED` event - we're measuring the latency of the quote request using the `performance` web API in `routing/slice`

at the same time, i'm removing `swapQuoteReceivedDate` because it's not needed anymore

## Screen capture

image

### QA (ie manual testing)

1. verified the property was included in the event using debug mode

### Automated testing

- [x] Unit test n/a
- [x] Integration/E2E test


just-toby:
> @just-toby, for my understanding - this changeset will let us see a breakdown of quote performance by calls to the Routing API for classic quotes, X quotes, and direct calls to the client-side SOR?

@andysmith415 yes, in amplitude we can group by `method` and `routing` to see that info. 

`method` is either `QuoteMethod.ROUTING_API` or `QuoteMethod.CLIENT_SIDE` , and routing is either `CLASSIC` or `UNISWAPX`

JFrankfurt:
There's an open task to do this, but to specifically differentiate between the quote sources: https://linear.app/uniswap/issue/WEB-2644/measure-interface-routing-latency

JFrankfurt (approved):

src/lib/utils/analytics.ts
```diff
@@ -86,14 +81,14 @@ function getQuoteMethod(trade: InterfaceTrade) {
 export const formatSwapQuoteReceivedEventProperties = (
   trade: InterfaceTrade,
   allowedSlippage: Percent,
-  swapQuoteReceivedDate: Date
+  swapQuoteLatency: number | undefined
```
tinaszheng, line 84: i think for these `number` metrics its always helpful to suffix it by the unit used, like `swapQuoteLatencyMs`

src/state/routing/slice.ts
```diff
@@ -114,6 +119,7 @@ export const routingApi = createApi({
       },
       async queryFn(args, _api, _extraOptions, fetch) {
         let fellBack = false
+        const quoteStartMark = performance.mark(`quote-fetch-start-${Date.now()}`)
```
tinaszheng, line 122: so do we want to measure the time between user typing input -> user seeing quote? im guessing the interface-side measurement of latency should include things like debounce, and in that case the quote-fetch-start should really start on user input change and not the routing/slice fetch

because theoretically just the API fetch metrics can be measured on the backend side right 🤔

just-toby: @tinaszheng this ask specifically is for the API fetch latencies - see [this message](https://uniswapteam.slack.com/archives/C02T729PMQE/p1691681536496569)

i'm separately working on logging for those client side metrics that you mentioned

src/hooks/useDebouncedTrade.ts
```diff
@@ -37,6 +38,7 @@ export function useDebouncedTrade(
 ): {
   state: TradeState
   trade?: ClassicTrade
+  swapQuoteLatency?: number
```
tinaszheng, line 41: change all instances to match the `ms` convention?

tinaszheng (approved):

PR #7144, opened 2023-08-11
feat: change rich link previews to Basel

LunrEclipse:
Updates Rich Link Previews to use Basel instead of Inter.

TODO: Actually change `getFont` function to import Basel font instead of Inter.


just-toby:
@LunrEclipse hey

PR #7148, opened 2023-08-11
fix: remove env from swcrc

cbachmeier:
## Description

- tests are failing blocking promotion with error `jsc.env and jsc.target cannot be used together`
- we recently added env to swcrc to fix uniwallet connection in #7139
- I've tested locally removing the env and still able to connect to uniwallet and wallet connect

_Slack thread:_ https://uniswapteam.slack.com/archives/C052ARNFZGU

## Screen capture
Screenshot 2023-08-11 at 12 51 21 PM

### QA (ie manual testing)

- [x] connect via uniwallet
- [x] connect via WC


cbachmeier:
Uniwallet broken on deploy :/

PR #7150, opened 2023-08-11
fix: fail UR transaction if chain IDs don't match

just-toby:
this may prevent a mysterious potential bug causing users to send funds to the mainnet UR address on base

from the ethers docs

transactionRequest.chainId ⇒ number | Promise< number >
The chain ID this transaction is authorized on, as specified by [EIP-155](https://eips.ethereum.org/EIPS/eip-155).
If the chain ID is 0 will disable EIP-155 and the transaction will be valid on any network. This can be dangerous and care should be taken, since it allows transactions to be replayed on networks that were possibly not intended. Intentionally-replayable transactions are also disabled by default on recent versions of Geth and require configuration to enable.

### DIFF
--- a/src/hooks/useUniversalRouter.ts
+++ b/src/hooks/useUniversalRouter.ts
@@ -58,6 +58,8 @@ export function useUniversalRouterSwapCallback(
         if (!chainId) throw new Error('missing chainId')
         if (!provider) throw new Error('missing provider')
         if (!trade) throw new Error('missing trade')
+        const connectedChainId = await provider.getSigner().getChainId()
+        if (chainId !== connectedChainId) throw new Error('signer chainId does not match')
 
         setTraceData('slippageTolerance', options.slippageTolerance.toFixed(2))
         const { calldata: data, value } = SwapRouter.swapERC20CallParameters(trade, {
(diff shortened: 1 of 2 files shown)


gnewfield (approved):

tinaszheng (approved):

zzmp:
This should include logging so we know when it is occuring.

src/hooks/useUniversalRouter.ts
```diff
@@ -58,6 +58,8 @@ export function useUniversalRouterSwapCallback(
         if (!chainId) throw new Error('missing chainId')
         if (!provider) throw new Error('missing provider')
         if (!trade) throw new Error('missing trade')
+        const connectedChainId = await provider.getSigner().getChainId()
+        if (chainId !== connectedChainId) throw new Error('signer chainId does not match')
```
zzmp, line 62: Will this bubble up to Sentry?

zzmp (approved):

JackShort (approved):
YEET

src/hooks/useUniversalRouter.ts
just-toby: no currently we aren't logging general swap failures to sentry

PR #7152, opened 2023-08-14
fix: add distance check from white for token rich link previews

LunrEclipse:
Fixes issue of Token Rich Link Previews background color being too white from color extraction by adding an additional check at the end to check distance from [255, 255, 255] and dimming accordingly.

Before:
0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2-1

After:
0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2

### DIFF
--- a/functions/utils/getColor.ts
+++ b/functions/utils/getColor.ts
@@ -67,9 +67,9 @@ function getAverageColor(arrayBuffer: Uint8Array, type: string, checkDistance: b
     const distance = Math.sqrt(Math.pow(r - 255, 2) + Math.pow(g - 255, 2) + Math.pow(b - 255, 2))
 
     if (distance < 50) {
-      r -= 20
-      g -= 20
-      b -= 20
+      r = Math.max(0, r - 29)
+      g = Math.max(0, g - 29)
+      b = Math.max(0, b - 29)
     }
   }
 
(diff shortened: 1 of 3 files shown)


functions/utils/getColor.ts
```diff
@@ -63,5 +63,15 @@ function getAverageColor(arrayBuffer: Uint8Array, type?: string) {
...
 
+  if (checkDistance) {
+    const distance = Math.sqrt(Math.pow(r - 255, 2) + Math.pow(g - 255, 2) + Math.pow(b - 255, 2))
+
+    if (distance < 50) {
+      r -= 20
+      g -= 20
+      b -= 20
+    }
+  }
```
zzmp, line 74: Should this be a while loop to enforce that distance ends up >=50?
Should there be a check that r,g,b > 0?

```suggestion
  if (checkDistance) {
    let distance = Math.sqrt((r - 255)**2, (g - 255)**2, (b - 255)**2)
    while (distance < 50) {
      r = Math.max(r - 20, 0)
      g = Math.max(g - 20, 0)
      b = Math.max(b - 20, 0)
      distance = Math.sqrt((r - 255)**2, (g - 255)**2, (b - 255)**2)
    }
  }
```

zzmp (changes requested):

zzmp (approved):

JFrankfurt:
The after image is still super low contrast--can we add something like readableColor to this?

https://polished.js.org/docs/#readablecolor

LunrEclipse:
> The after image is still super low contrast--can we add something like readableColor to this?
> 
> https://polished.js.org/docs/#readablecolor

After some tweaking this is what it currently looks like. Is the contrast still too low?
Screenshot 2023-08-17 at 11 37 49 AM

JFrankfurt:
I prefer the second option, but I think it would be best to still draw the color from the token logo and use `readableColor` to render black/white text based on contrast.

JFrankfurt (approved):

JFrankfurt (approved):

PR #7153, opened 2023-08-14
fix: add assets to public folder

LunrEclipse:
Added Rich Link Preview dependent assets such as font and network logos to public folder


zzmp (approved):

public/images/logos/Arbitrum_Logo.png
JFrankfurt, line 1: We already have `src\assets\svg\arbitrum_logo.svg`, can that work for you here instead?

LunrEclipse: I have not figured out how to bundle svgs with cloudflare functions so pngs are preferred

LunrEclipse: also the cloudflare functions need the files to be bundled with the functions folder or in the public folder to access

JFrankfurt: Can you look into moving those files to the public folder and bundling svgs?

JFrankfurt (approved):

PR #7155, opened 2023-08-14
fix: token margin in searchbar

JackShort:
* fixes the token icon margin in the searchbar

### Before
Screen Shot 2023-08-14 at 1 08 01 PM

### After
Screen Shot 2023-08-14 at 1 07 25 PM


tinaszheng (approved):

LunrEclipse (approved):

cbachmeier (approved):
We also need to get this into staging before releasing to prod

PR #7156, opened 2023-08-14
fix: remove flakey tests from cloud functions and overall deflake

LunrEclipse:
Removed flakey tests from cloud functions and enabled setting to have jest retry tests 3 times upon fails. Increased setup time in `globalSetup` due to tests failing before the server has been setup properly. Also reduced `maxWorkers` to 1 to ensure that Jest does not overtake all resources and shut down proxy.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -225,7 +225,8 @@ jobs:
           path: node_modules/.cache
           key: ${{ runner.os }}-cloud-jest-${{ github.run_id }}
           restore-keys: ${{ runner.os }}-cloud-jest-
-      - run: yarn test:cloud --coverage --maxWorkers=80%
+      # Only use 1 worker, so the other can be used for the proxy server under test.
+      - run: yarn test:cloud --coverage --maxWorkers=1
         
   pre:
     if: ${{ github.ref_name == 'main' || github.ref_name == 'releases/staging' }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -225,7 +225,7 @@ jobs:
           path: node_modules/.cache
           key: ${{ runner.os }}-cloud-jest-${{ github.run_id }}
           restore-keys: ${{ runner.os }}-cloud-jest-
-      - run: yarn test:cloud --coverage --maxWorkers=100%
+      - run: yarn test:cloud --coverage --maxWorkers=80%
         
   pre:
     if: ${{ github.ref_name == 'main' || github.ref_name == 'releases/staging' }}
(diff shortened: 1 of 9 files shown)


.github/workflows/test.yml
```diff
@@ -225,7 +225,7 @@ jobs:
           path: node_modules/.cache
           key: ${{ runner.os }}-cloud-jest-${{ github.run_id }}
           restore-keys: ${{ runner.os }}-cloud-jest-
-      - run: yarn test:cloud --coverage --maxWorkers=100%
+      - run: yarn test:cloud --coverage --maxWorkers=80%
```
zzmp, line 228: because the other jest invocation uses 100%, you should justify this with a comment.
also, instead of 80%, and knowing that it competes with the server, can you be explicit and use `=1` or just omit it (which is equivalent to ["the number of cores available on your machine minus one" (ie `=1`)](https://jestjs.io/docs/cli#--maxworkersnumstring)?

```suggestion
      # Only use 1 worker, so the other can be used for the proxy server under test.
      - run: yarn test:cloud --coverage
```

functions/global-setup.ts
```diff
@@ -4,7 +4,7 @@ module.exports = async function globalSetup() {
   globalThis.servers = await setup({
     command: `yarn start:cloud`,
     port: 3000,
-    launchTimeout: 80000,
+    launchTimeout: 120000,
```
zzmp, line 7: Add a comment justifying this:

```suggestion
    launchTimeout: 120000, // takes ~2m on CI
```

functions/global-teardown.ts
```diff
@@ -1,5 +1,6 @@
 import { teardown } from 'jest-dev-server'
 
 module.exports = async function globalTeardown() {
+  await new Promise((resolve) => setTimeout(resolve, 5000))
```
zzmp, line 4: Why is this timeout necessary? Shouldn't all of the tests be done by this point?

zzmp (changes requested):

functions/global-teardown.ts
LunrEclipse: From what I've deduced, the 2 remaining sources of flakiness are:
1. the server just starting improperly which I haven't been able to figure out why this happens (only happens 1/10 runs)
2. The socket closing before tests finish.

I was troubleshooting the 2nd cause and thought that might help it, but I'm pretty sure it was because of jest competing with the proxy. Will remove and test!

zzmp (approved):

natew (approved):

PR #7161, opened 2023-08-14
fix: add back the disabled Base entry in the tokens page selector

JFrankfurt:
Base was removed from the "unsupported" chains list because most backend functionality supports it, but we still don't have the necessary data to support the tokens page. This PR adds back the disabled ux in the dropdown. 

image

### DIFF
--- a/src/components/NavBar/SearchBarDropdown.tsx
+++ b/src/components/NavBar/SearchBarDropdown.tsx
@@ -10,7 +10,6 @@ import { HistoryDuration, SafetyLevel } from 'graphql/data/__generated__/types-a
 import { useTrendingCollections } from 'graphql/data/nft/TrendingCollections'
 import { SearchToken } from 'graphql/data/SearchTokens'
 import useTrendingTokens from 'graphql/data/TrendingTokens'
-import { BACKEND_NOT_YET_SUPPORTED_CHAIN_IDS } from 'graphql/data/util'
 import { useDisableNFTRoutes } from 'hooks/useDisableNFTRoutes'
 import { useIsNftPage } from 'hooks/useIsNftPage'
 import { Box } from 'nft/components/Box'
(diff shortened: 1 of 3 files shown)


src/components/Tokens/TokenTable/NetworkFilter.tsx
```diff
@@ -1,5 +1,7 @@
+import { ChainId } from '@uniswap/sdk-core'
 import Badge from 'components/Badge'
 import { getChainInfo } from 'constants/chainInfo'
+import { Chain } from 'graphql/data/__generated__/types-and-hooks'
 import {
   BACKEND_NOT_YET_SUPPORTED_CHAIN_IDS,
```
cbachmeier, line 6: It should be added to this array. `BACKEND_NOT_YET_SUPPORTED_CHAIN_IDS` is only used to filter out Chains which don't have Token Explore data. We could also make that more clear by renaming

cbachmeier (changes requested):

src/components/Tokens/TokenTable/NetworkFilter.tsx
JFrankfurt: BACKEND_NOT_YET_SUPPORTED_CHAIN_IDS is also used to render the "search support coming soon" message in the search bar. 



image

src/graphql/data/util.tsx
```diff
@@ -130,14 +130,23 @@ const URL_CHAIN_PARAM_TO_BACKEND: { [key: string]: InterfaceGqlChain } = {
   avalanche: Chain.Avalanche,
 }
 
+const SUPPORTED_CHAINS = new Set([
+  Chain.Arbitrum,
+  Chain.Avalanche,
+  Chain.Base,
```
cbachmeier, line 136: these three are not supported by Token Explore

src/components/NavBar/SearchBarDropdown.tsx
```diff
@@ -137,7 +136,8 @@ interface SearchBarDropdownProps {
 export const SearchBarDropdown = (props: SearchBarDropdownProps) => {
   const { isLoading } = props
   const { chainId } = useWeb3React()
-  const showChainComingSoonBadge = chainId && BACKEND_NOT_YET_SUPPORTED_CHAIN_IDS.includes(chainId) && !isLoading
+  const COMING_SOON_CHAINS = new Set([ChainId.BNB, ChainId.AVALANCHE])
```
cbachmeier, line 139: BASE should also show the coming soon badge

cbachmeier (changes requested):
The `COMING_SOON_CHAINS` in the SearchBar are the same as `UNSUPPORTED_CHAINS` in the Network Filter and are the same as the deleted `BACKEND_NOT_YET_SUPPORTED_CHAIN_IDS`. I think you should keep the types central in `src/graphql/data/util.tsx`

see the comments in this thread https://uniswapteam.slack.com/archives/C047U65H422/p1692036300292369

src/components/NavBar/SearchBarDropdown.tsx
JFrankfurt: Base is supported

image

PR #7168, opened 2023-08-15
fix: broken spacing in tokens fixture

natew:
i had accidently remove whitespace from the token names.

this is a no-change basically, besides reverting the tokens names that i removed space from. eg DaiStablecoin should be Dai Stablecoin again.


cartcrom (approved):
ty!

zzmp (approved):

cbachmeier (approved):

PR #7171, opened 2023-08-16
feat: dedupe on commit

JFrankfurt:
## Description
Adds an automatic hook to run `yarn yarn-deduplicate --strategy=highest` whenever yarn.lock changes


package.json
```diff
@@ -33,6 +33,16 @@
...
     "postinstall": "yarn patch-package"
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "yarn.lock": [
+      "yarn yarn-deduplicate --strategy=highest"
```
zzmp, line 43: why not just run `yarn deduplicate`, and use our existing script? This seems more prone to skew.

package.json
```diff
@@ -33,6 +33,16 @@
     "deduplicate": "yarn-deduplicate --strategy=highest",
     "postinstall": "yarn patch-package"
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
```
zzmp, line 38: won't this cost some time?

zzmp (changes requested):

package.json
JFrankfurt: yeah, but less than 1 second

package.json
JFrankfurt: did not know this existed, thanks

zzmp (approved):

JackShort (approved):

PR #7172, opened 2023-08-16
feat: add description for rich link previews

LunrEclipse:
Updates meta tags for NFT assets and collections to include information regarding the NFT collection. Also added description tags to tokens including price and ticket.


functions/utils/types.ts
```diff
@@ -0,0 +1,162 @@
+import { gql } from '@apollo/client'
```
zzmp, line 1: Why is this necessary? Is it because you're making a new query, but the types are generated from src? Does that mean you were just re-using queries from src?

You should generate your own query here if possible. If not possible, you should commit a query into src (with comments as to why it lives there) and then pull it from there so you get auto-magic typings.

wdyt?

zzmp (changes requested):

functions/utils/types.ts
LunrEclipse: yeah i was just re-using queries from src... 

I think its probably best to generate my own queries since like 90% of the information from src aren't needed so we can strip out a lot and really simply it. should i do that in this pr or a new one?

zzmp: if you _can_ do that in this PR it would be great - if you've got time :)

PR #7173, opened 2023-08-16
feat: more swap metric logging

just-toby:
## Description

logs a new event when the user first has a meaningful interaction with the swap component - `FIRST_SWAP_ACTION`. this will be used to measure the elapsed time between page load and this interaction.

also, adds a new property to the `SWAP_TRANSACTION_COMPLETED` event, `time_to_swap_since_first_input` - this is the elapsed time between the above new event, and the existing TTS metric moment.

introduces `SwapEventTimestampTracker` as a singleton utility for managing the session timestamps of the swap funnel events .

_Relevant docs:_ https://www.notion.so/uniswaplabs/Swap-Metrics-Logging-448423923e6f45a5a5fca7a894113ece

## Screen capture

image

image

### QA (ie manual testing)

- [x] manually verified that the events and properties are logged in debug mode

### Automated testing

- [x] Unit test
- [x] Integration/E2E test


src/pages/Swap/index.tsx
```diff
@@ -327,15 +328,17 @@ export function Swap({
 
   const handleTypeInput = useCallback(
     (value: string) => {
+      maybeLogFirstSwapAction(trace)
       onUserInput(Field.INPUT, value)
```
gnewfield, line 332: <nit> maybe best to switch the order, so there's no chance that `onUserInput` is delayed by logging

src/pages/Swap/index.tsx
gnewfield, line 1: Should we also add an event on reversing the inputs? This would be a strange thing to do first but possible

src/tracing/SwapEventTimestampTracker.ts
gnewfield, line 1: This is nice compared to the old updater

gnewfield (approved):
LGTM

src/tracing/SwapEventTimestampTracker.ts
```diff
@@ -0,0 +1,51 @@
...
+  FIRST_QUOTE_FETCH_STARTED = 'FIRST_QUOTE_FETCH_STARTED',
+  FIRST_SWAP_SIGNATURE_REQUESTED = 'FIRST_SWAP_SIGNATURE_REQUESTED',
+  FIRST_SWAP_SIGNATURE_COMPLETED = 'FIRST_SWAP_SIGNATURE_COMPLETED',
+  FIRST_SWAP = 'FIRST_SWAP',
+}
+
+export class SwapEventTimestampTracker {
+  private static _instance: SwapEventTimestampTracker
+  private constructor() {
+    // Private constructor to prevent direct construction calls with the `new` operator.
```
cbachmeier, line 14: qq: why can't the `new` operator be used? so there's only 1?

src/tracing/swapFlowLoggers.ts
```diff
@@ -0,0 +1,33 @@
...
+      ? undefined
+      : tracker.getElapsedTime(SwapEventType.FIRST_SWAP, SwapEventType.FIRST_SWAP_ACTION),
+    hash,
+    chainId,
+    ...analyticsContext,
+  })
+}
+
+// We only log the time-to-first-swap-input metric for the first swap input of a session.
+export function maybeLogFirstSwapAction(analyticsContext: any) {
```
cbachmeier, line 25: dont use `any` I believe the type is `ITraceContext`

src/tracing/swapFlowLoggers.ts
```diff
@@ -0,0 +1,33 @@
+import { SwapEventName } from '@uniswap/analytics-events'
+import { sendAnalyticsEvent } from 'analytics'
+
+import { SwapEventTimestampTracker, SwapEventType } from './SwapEventTimestampTracker'
+
+const tracker = SwapEventTimestampTracker.getInstance()
+
+export function logSwapSuccess(hash: string, chainId: number, analyticsContext: any) {
```
cbachmeier, line 8: don't use `any` I believe the type is `ITraceContext`

src/tracing/SwapEventTimestampTracker.ts
```diff
@@ -0,0 +1,51 @@
+import { calculateElapsedTimeWithPerformanceMark } from './utils'
+
+export enum SwapEventType {
+  FIRST_SWAP_ACTION = 'FIRST_SWAP_ACTION',
+  FIRST_QUOTE_FETCH_STARTED = 'FIRST_QUOTE_FETCH_STARTED',
+  FIRST_SWAP_SIGNATURE_REQUESTED = 'FIRST_SWAP_SIGNATURE_REQUESTED',
+  FIRST_SWAP_SIGNATURE_COMPLETED = 'FIRST_SWAP_SIGNATURE_COMPLETED',
```
cbachmeier, line 7: qq: these 3 events aren't used other than for tests, are they to be implemented later?

src/tracing/utils.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Returns the time elapsed between page load and now.
```
cbachmeier, line 2: nit: add jsdoc for input and what `markName` refers to

src/tracing/swapFlowLoggers.ts
```diff
@@ -0,0 +1,33 @@
...
+  const hasSetSwapSuccess = tracker.hasTimestamp(SwapEventType.FIRST_SWAP)
+  const elapsedTime = tracker.setElapsedTime(SwapEventType.FIRST_SWAP)
+  sendAnalyticsEvent(SwapEventName.SWAP_TRANSACTION_COMPLETED, {
+    // We only log the time-to-swap metric for the first swap of a session,
+    // so if it was previously set we log undefined here.
+    time_to_swap: hasSetSwapSuccess ? undefined : elapsedTime,
+    time_to_swap_since_first_input: hasSetSwapSuccess
+      ? undefined
```
cbachmeier, line 16: qq: should we do something other than `undefined` for second+ swaps? This way we know when parsing logs that it's not because the variable was just never set? Maybe something like `-1`

src/tracing/SwapEventTimestampTracker.ts
```diff
@@ -0,0 +1,51 @@
...
+  FIRST_SWAP = 'FIRST_SWAP',
+}
+
+export class SwapEventTimestampTracker {
```
cbachmeier, line 11: Can you add the creation of this class and its purpose to the PR description

src/tracing/SwapEventTimestampTracker.ts
```diff
@@ -0,0 +1,51 @@
+import { calculateElapsedTimeWithPerformanceMark } from './utils'
+
+export enum SwapEventType {
+  FIRST_SWAP_ACTION = 'FIRST_SWAP_ACTION',
+  FIRST_QUOTE_FETCH_STARTED = 'FIRST_QUOTE_FETCH_STARTED',
+  FIRST_SWAP_SIGNATURE_REQUESTED = 'FIRST_SWAP_SIGNATURE_REQUESTED',
+  FIRST_SWAP_SIGNATURE_COMPLETED = 'FIRST_SWAP_SIGNATURE_COMPLETED',
+  FIRST_SWAP = 'FIRST_SWAP',
```
cbachmeier, line 8: nit: `FIRST_SWAP_SUCCESS` is more direct and would match the helper fn `logSwapSuccess`

src/pages/Swap/index.tsx
```diff
@@ -328,14 +329,16 @@ export function Swap({
   const handleTypeInput = useCallback(
     (value: string) => {
       onUserInput(Field.INPUT, value)
+      maybeLogFirstSwapAction(trace)
```
cbachmeier, line 332: what's the total list of first swap actions? tapping input, output, max, & switch arrows? Could it be added to `src/tracing/SwapEventTimestampTracker.ts` for reference?

src/tracing/SwapEventTimestampTracker.test.ts
```diff
@@ -0,0 +1,51 @@
+import { SwapEventTimestampTracker, SwapEventType } from './SwapEventTimestampTracker'
+
+// mock calculateElapsedTimeWithPerformanceMark in ./utils.ts:
```
cbachmeier, line 3: nit: self commenting code

cbachmeier (changes requested):
couple qs and nits, only thing blocking is the use of `any` when we should have the type available

src/tracing/SwapEventTimestampTracker.ts
just-toby: yes, i added them all at once to the names package but upcoming PRs will start using them in interface

src/tracing/swapFlowLoggers.ts
just-toby: i get that but logging a placeholder number here seems more dangerous to cause an incorrect analysis in amplitude, if we forget to filter and they are quietly included in some aggregations

PR #7174, opened 2023-08-16
feat: swap quote event

just-toby:
## Description

`logSwapQuoteRequest` to log each initiated quote request - including a `time_to_first_quote_request` property for the first one.

_Relevant docs:_ https://www.notion.so/uniswaplabs/Swap-Metrics-Logging-448423923e6f45a5a5fca7a894113ece

## Screen capture

image

### QA (ie manual testing)

- [x] manually verified that the durations are correct in debug mode

### Automated testing

- [x] Unit test n/a
- [x] Integration/E2E test


src/tracing/swapFlowLoggers.ts
```diff
@@ -31,3 +31,16 @@ export function maybeLogFirstSwapAction(analyticsContext: any) {
...
+  const elapsedTime = tracker.setElapsedTime(SwapEventType.FIRST_QUOTE_FETCH_STARTED)
+  sendAnalyticsEvent(SwapEventName.SWAP_QUOTE_FETCH, {
+    chainId,
+    // We only log the time_to_first_quote_request metric for the first quote request of a session.
+    time_to_first_quote_request: hasSetSwapQuote ? undefined : elapsedTime,
+    time_to_first_quote_request_since_first_input: hasSetSwapQuote
+      ? undefined
+      : tracker.getElapsedTime(SwapEventType.FIRST_QUOTE_FETCH_STARTED, SwapEventType.FIRST_SWAP_ACTION),
+  })
+}
```
JFrankfurt, line 46: unit testing this might be a better fit than e2e?

(also makes that  ❌  in the checks go to ✅ )

JFrankfurt (approved):

cypress/e2e/swap/swapFlowLogging.test.ts
```diff
@@ -13,6 +13,23 @@ describe('time-to-swap logging', () => {
...
+      cy.wrap(event.event_properties.time_to_first_swap_action).should('be.a', 'number')
+      cy.wrap(event.event_properties.time_to_first_swap_action).should('be.gte', 0)
+    })
+
+    // Verify Swap Quote
+    cy.waitForAmplitudeEvent(SwapEventName.SWAP_QUOTE_FETCH).then((event: any) => {
+      cy.wrap(event.event_properties).should('have.property', 'time_to_first_quote_request')
+      cy.wrap(event.event_properties.time_to_first_quote_request).should('be.a', 'number')
+      cy.wrap(event.event_properties.time_to_first_quote_request).should('be.gte', 0)
+      cy.wrap(event.event_properties).should('have.property', 'time_to_first_quote_request_since_first_input')
```
cbachmeier, line 28: nit: can we also add to the second swap in the session test to make sure this event property is undefined

cbachmeier (dismissed):

cypress/e2e/swap/swapFlowLogging.test.ts
```diff
@@ -3,8 +3,8 @@ import { SwapEventName } from '@uniswap/analytics-events'
 import { USDC_MAINNET } from '../../../src/constants/tokens'
 import { getTestSelector } from '../../utils'
 
-describe('time-to-swap logging', () => {
-  it('completes two swaps and verifies the TTS logging for the first', () => {
+describe('swap flow logging', () => {
+  it('completes two swaps and verifies the TTS logging for the first, plus all intermediate steps along the way', () => {
```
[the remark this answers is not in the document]
just-toby: i tried splitting it up into different test cases but it didnt work - and we don't want to reset the page for each test case because it'll be slower

cypress/e2e/swap/swapFlowLogging.test.ts
just-toby: done

src/tracing/swapFlowLoggers.ts
just-toby: yeah, you're right. i'll add unit tests for these in a follow up PR

src/tracing/swapFlowLoggers.ts
```diff
@@ -32,3 +33,25 @@ export function maybeLogFirstSwapAction(analyticsContext: ITraceContext) {
     })
   }
 }
+
+export function logSwapQuoteRequest(
+  chainId: number,
+  routerPreference: RouterPreference | typeof INTERNAL_ROUTER_PREFERENCE_PRICE
+) {
+  let performanceMetrics = {}
+  if (routerPreference !== INTERNAL_ROUTER_PREFERENCE_PRICE) {
```
just-toby, line 42: added since last commit - we don't want to log the performance metrics for a random price quote

JFrankfurt (approved):

src/tracing/swapFlowLoggers.ts
```diff
@@ -32,3 +33,25 @@ export function maybeLogFirstSwapAction(analyticsContext: ITraceContext) {
...
+    const elapsedTime = tracker.setElapsedTime(SwapEventType.FIRST_QUOTE_FETCH_STARTED)
+    performanceMetrics = {
+      // We only log the time_to_first_quote_request metric for the first quote request of a session.
+      time_to_first_quote_request: hasSetSwapQuote ? undefined : elapsedTime,
+      time_to_first_quote_request_since_first_input: hasSetSwapQuote
+        ? undefined
+        : tracker.getElapsedTime(SwapEventType.FIRST_QUOTE_FETCH_STARTED, SwapEventType.FIRST_SWAP_ACTION),
+    }
+  }
+  sendAnalyticsEvent(SwapEventName.SWAP_QUOTE_FETCH, {
```
cbachmeier, line 53: do we want to track `SWAP_QUOTE_FETCH` sent for every price fetch with just chainID and an empty obj? If not you could just have the sendAnalyticsEvent inside the `if` statement since it's a void returning fn

just-toby: yeah the idea is to send it for every quote request initiated. price requests also trigger the corresponding quote_received event so i think we should keep this.

i mostly just don't want to log time_to_first_quote for price requests because it's not really part of the UX flow that we care about

cbachmeier (approved):

PR #7175, opened 2023-08-16
fix: make WrongChainError more specific

just-toby:
use a more specific error for this case, so that we have the information in the sentry trace


src/hooks/useUniversalRouter.ts
```diff
@@ -36,6 +36,12 @@ class ModifiedSwapError extends Error {
   }
 }
 
+class WrongChainError extends Error {
+  constructor() {
+    super(t`Your wallet is connected to the wrong network.`)
+  }
+}
```
zzmp, line 43: This should be in `utils/errors` so it can be re-used, eg for #7177

zzmp (approved):

gnewfield (approved):

PR #7177, opened 2023-08-16
feat: check chain ID before all transactions

just-toby:
## Description

similar to #7150 - add a sanity check before every call to `provider.sendTransaction` that ensures the chainIds match


zzmp (changes requested):
this should be stacked on top of #7175

zzmp (approved):

gnewfield (approved):

PR #7182, opened 2023-08-17
feat: add real paths to app site association

zzmp:
## Description

Adds real (not hash-based) paths to app site association, as described in https://developer.apple.com/documentation/xcode/supporting-associated-domains. This should enable deep-linking into the Uniswap Mobile Wallet when using path-based routing.

_This is a hotfix of #7179._

_Slack thread:_ https://uniswapteam.slack.com/archives/C02T729PMQE/p1692222996293519?thread_ts=1692219376.681979&cid=C02T729PMQE

### DIFF
--- a/public/apple-app-site-association
+++ b/public/apple-app-site-association
@@ -22,6 +22,22 @@
           {
             "#": "/address/*",
             "comment": "Wallet address"
+          },
+          {
+            "/": "/nfts/asset/*",
+            "comment": "NFT Item"
+          },
+          {
+            "/": "/nfts/collection/*",
+            "comment": "NFT Collection"
+          },
+          {
+            "/": "/tokens/*",
+            "comment": "Token address"
+          },
+          {
+            "/": "/address/*",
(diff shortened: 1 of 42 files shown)


JackShort (approved):

LunrEclipse (approved):

adjkant (approved):

public/apple-app-site-association
```diff
@@ -22,6 +22,22 @@
           {
             "#": "/address/*",
             "comment": "Wallet address"
+          },
+          {
+            "/": "/nfts/asset/*",
+            "comment": "NFT Item"
```
cbachmeier, line 28: should we worry about the overlap with the existing component? Do we need to delete the old?
```
{
   "#": "/nfts/asset/*",
   "comment": "NFT Item"
},

zzmp: we want the overlap, as it will facilitate updating the mobile app for now.

tinaszheng (approved):

PR #7184, opened 2023-08-17
fix: unconnected state quotes

just-toby:
#6830  uncovered some incorrect assumptions in our codebase - when there is no wallet connected, we want alot of the UX to default to ETH Mainnet so the user can still lookup quotes.

## Description

_Linear ticket:_
_Slack thread:_
_Relevant docs:_

### Before

### After

### Reproducing the error

1. 

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test

### DIFF
--- a/cypress/e2e/swap/unconnected.test.ts
+++ b/cypress/e2e/swap/unconnected.test.ts
@@ -1,0 +1,26 @@
+import { SwapEventName } from '@uniswap/analytics-events'
+import { USDC_MAINNET } from 'constants/tokens'
+
+import { getTestSelector } from '../../utils'
+
+describe('Swap inputs with no wallet connected', () => {
+  it('can input and load a quote with no wallet connected', () => {
+    cy.visit(`/swap?inputCurrency=ETH&outputCurrency=${USDC_MAINNET.address}`)
+    cy.get(getTestSelector('web3-status-connected')).click()
+    // click twice, first time to show confirmation, second to confirm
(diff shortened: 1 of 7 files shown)


cypress/e2e/swap/unconnected.test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+describe('Swap inputs with no wallet connected', () => {
+  it('can input and load a quote with no wallet connected', () => {
+    cy.visit(`/swap?inputCurrency=ETH&outputCurrency=${USDC_MAINNET.address}`)
+    cy.get(getTestSelector('web3-status-connected')).click()
+    // click twice, first time to show confirmation, second to confirm
+    cy.get(getTestSelector('wallet-disconnect')).click()
+    cy.get(getTestSelector('wallet-disconnect')).should('contain', 'Disconnect')
+    cy.get(getTestSelector('wallet-disconnect')).click()
+    cy.get(getTestSelector('close-account-drawer')).click()
```
zzmp, line 14: Instead of disconnecting, could you pass a user state that is explicitly disconnected to `cy.visit`?

zzmp (changes requested):
I'm confused as to why `useWeb3React()` is not returning a chainId - it seems like the network connector (the default) is maybe not connecting? I think _that_ is what needs to be investigated/fixed.

PR #7186, opened 2023-08-17
fix: make line height bigger for token preview

LunrEclipse:
Increased line height for token preview with long names and slices symbol for some tokens to prevent symbol for crossing to 2 lines. 

Before:
0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0-1
0xaa6e8127831c9de45ae56bb1b0d4d4da6e5665bd-1

After:
0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0

0xaa6e8127831c9de45ae56bb1b0d4d4da6e5665bd

### DIFF
--- a/functions/api/image/tokens/[[index]].tsx
+++ b/functions/api/image/tokens/[[index]].tsx
@@ -145,9 +145,13 @@ export const onRequest: PagesFunction = async ({ params, request }) => {
                     fontSize: '168px',
                     lineHeight: '133px',
                     marginLeft: '-13px',
+                    overflow: 'hidden',
+                    textOverflow: 'ellipsis',
+                    whiteSpace: 'nowrap',
+                    width: '100%',
                   }}
                 >
-                  {data.symbol.slice(0, 6)}
+                  {data.symbol}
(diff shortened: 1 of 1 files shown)


functions/api/image/tokens/[[index]].tsx
```diff
@@ -147,7 +147,7 @@ export const onRequest: PagesFunction = async ({ params, request }) => {
                     marginLeft: '-13px',
                   }}
                 >
-                  {data.symbol}
+                  {data.symbol.slice(0, 6)}
```
Comment (JackShort) [inline functions/api/image/tokens/[[index]].tsx:150]: to solve the problem of the symbol not crossing two lines can't you just set the overflow here? to like overflow none, and text-wrap to ellipsis or will that not work for some reason?

Comment (LunrEclipse) [reply functions/api/image/tokens/[[index]].tsx]: i did not think of that LOL

Comment (JackShort) [reply functions/api/image/tokens/[[index]].tsx]: LOL

JackShort (approved):

zzmp (approved):

PR #7188, opened 2023-08-18
feat: use cache while debouncing quotes

zzmp:
## Description

Returns cache entries (using rtk's `useQueryState`) when debouncing quotes, to provide _instant_ quotes for already-loaded data.

_Relevant docs:_ https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#usequerystate

## Screen capture

|    Before    |   After    |
| <video src="https://github.com/Uniswap/interface/assets/5403956/a12013f7-c87f-460e-804d-0f52859e4fe2" > | <video src="https://github.com/Uniswap/interface/assets/5403956/e3c5791e-4c39-49ca-8303-7639d9f26f42" > |

### DIFF
--- a/src/hooks/useDebouncedTrade.ts
+++ b/src/hooks/useDebouncedTrade.ts
@@ -70,28 +70,27 @@ export function useDebouncedTrade(
     [amountSpecified, otherCurrency]
   )
   const debouncedSwapQuoteFlagEnabled = useDebounceSwapQuoteFlag() === DebounceSwapQuoteVariant.Enabled
-  const debouncedInputs = useDebounce(inputs, debouncedSwapQuoteFlagEnabled ? DEBOUNCE_TIME_INCREASED : DEBOUNCE_TIME)
-  const isDebouncing = debouncedInputs !== inputs
-  const [debouncedAmount, debouncedOtherCurrency] = debouncedInputs
+  const isDebouncing =
+    useDebounce(inputs, debouncedSwapQuoteFlagEnabled ? DEBOUNCE_TIME_INCREASED : DEBOUNCE_TIME) !== inputs
 
   const isWrap = useMemo(() => {
-    if (!chainId || !amountSpecified || !debouncedOtherCurrency) return false
+    if (!chainId || !amountSpecified || !otherCurrency) return false
     const weth = WRAPPED_NATIVE_CURRENCY[chainId]
     return (
-      (amountSpecified.currency.isNative && weth?.equals(debouncedOtherCurrency)) ||
-      (debouncedOtherCurrency.isNative && weth?.equals(amountSpecified.currency))
+      (amountSpecified.currency.isNative && weth?.equals(otherCurrency)) ||
+      (otherCurrency.isNative && weth?.equals(amountSpecified.currency))
     )
-  }, [amountSpecified, chainId, debouncedOtherCurrency])
+  }, [amountSpecified, chainId, otherCurrency])
 
   const shouldGetTrade = !isWrap && isWindowVisible
 
   const [routerPreference] = useRouterPreference()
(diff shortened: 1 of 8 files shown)


src/state/routing/useRoutingAPITrade.ts
```diff
@@ -78,55 +78,46 @@ export function useRoutingAPITrade<TTradeType extends TradeType>(
...
     // Price-fetching is informational and costly, so it's done less frequently.
     pollingInterval: routerPreference === INTERNAL_ROUTER_PREFERENCE_PRICE ? ms(`1m`) : AVERAGE_L1_BLOCK_TIME,
     // If latest quote from cache was fetched > 2m ago, instantly repoll for another instead of waiting for next poll period
     refetchOnMountOrArgChange: 2 * 60,
   })
-  const isCurrent = currentTradeResult === tradeResult
+  const isFetching = currentData !== tradeResult || !currentData
 
   return useMemo(() => {
-    if (skipFetch && amountSpecified) {
```
cbachmeier, line 100: q: checking that we do in fact want to no longer return the stale trade when skipping fetch but have valid input?

src/hooks/useDebouncedTrade.ts
```diff
@@ -70,39 +70,27 @@ export function useDebouncedTrade(
...
-      (amountSpecified.currency.isNative && weth?.equals(debouncedOtherCurrency)) ||
-      (debouncedOtherCurrency.isNative && weth?.equals(amountSpecified.currency))
+      (amountSpecified.currency.isNative && weth?.equals(otherCurrency)) ||
+      (otherCurrency.isNative && weth?.equals(amountSpecified.currency))
     )
-  }, [amountSpecified, chainId, debouncedOtherCurrency])
+  }, [amountSpecified, chainId, otherCurrency])
 
-  const shouldGetTrade = !isWrap && isWindowVisible
+  const skipFetch = isDebouncing || !autoRouterSupported || !isWindowVisible || isWrap
```
cbachmeier, line 85: q: why do we want to skip the fetch if autorouter isn't supported? do we know in that case that no valid route would be returned?

src/state/swap/hooks.tsx
```diff
@@ -116,33 +115,14 @@ export function useDerivedSwapInfo(state: SwapState, chainId: ChainId | undefine
...
-  let trade = useDebouncedTrade(
+  const trade = useDebouncedTrade(
     isExactIn ? TradeType.EXACT_INPUT : TradeType.EXACT_OUTPUT,
     parsedAmount,
     (isExactIn ? outputCurrency : inputCurrency) ?? undefined,
     undefined,
     account
   )
 
-  const nextPreviouslyInvalid = (() => {
```
cbachmeier, line 127: q: why do we no longer need this previouslyInvalid check?

cbachmeier (approved):
LGTM just had some Qs

src/state/swap/hooks.tsx
zzmp: this was adding rendering latency, so i dropped it. i did not understand what it added.

src/hooks/useDebouncedTrade.ts
zzmp: !autoRouterSupported was moved up from L94

src/state/swap/hooks.tsx
zzmp: not latency, jank. state doesn't render until one frame later, and this uses `useState`.

src/state/routing/useRoutingAPITrade.ts
zzmp: fixed

cbachmeier (approved):

PR #7191, opened 2023-08-18
chore: only show last trade for same pair

zzmp:
## Description

Clears the dimmed loading trade state if the input and output tokens do not match the old (dimmed) trade state.

## Screen capture

|    before    |   after    |


src/state/routing/useRoutingAPITrade.ts
```diff
@@ -87,19 +87,24 @@ export function useRoutingAPITrade<TTradeType extends TradeType>(
...
 
   return useMemo(() => {
     if (amountSpecified && queryArgs === skipToken) {
       return {
         state: TradeState.STALE,
-        trade: tradeResult?.trade,
+        trade: isProgressive ? tradeResult?.trade : undefined,
         swapQuoteLatency: tradeResult?.latencyMs,
       }
-    } else if (!amountSpecified || isError || queryArgs === skipToken) {
```
cbachmeier, line 102: qq: no longer need !amountSpecified?

cbachmeier (approved):
works well, only 1 Q
image

src/state/routing/useRoutingAPITrade.ts
JackShort, line 102: just so my understanding is correct - the check for `!amountSpecified` is redundant because if that is true than `queryArgs === skipToken` correct?

JackShort (approved):

zzmp:
This appears to have broken UniswapX tests so I'll have to give it another go 🦐 🍤

zzmp:
this is still relevant - i will refactor - pls do not close

PR #7195, opened 2023-08-21
feat: workaround for FOT swaps hotfix

cartcrom:
## Description

Updates the swap details dropdown and review modal to display tax percentages for hardcoded FOT tokens. (Tax rates will be fetched dynamically as a next step for FOT work)

Appends tax rates to slippage amount before encoding the swap to allow FOT swaps to work.

Adds feature flag that gates tax accounting logic, incase we need to revert these changes

## Screen capture

|    Before    |   After    |
| image | image |

|    Before    |   After   |
| image  | image |

### Reproducing the error

1. attempt to swap a FOT token like BULLET 0x8ef32a03784c8fd63bbf027251b9620865bd54b6
2. see gas estimation error


src/state/routing/types.ts
```diff
@@ -187,6 +193,21 @@ export class ClassicTrade extends Trade<Currency, Currency, TradeType> {
     this.requestId = requestId
     this.quoteMethod = quoteMethod
     this.approveInfo = approveInfo
+    this.inputTax = inputTax
+    this.outputTax = outputTax
+  }
+
+  public get totalTaxRate(): Percent {
+    return this.inputTax.add(new Fraction(ONE).subtract(this.inputTax).multiply(this.outputTax))
```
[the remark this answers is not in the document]
cartcrom: I'll add a comment, this is to handle for swapping from FOT to FOT. The output tax is only applied to what's left of the amountIn after the input tax has already been applied

So the total tax rate is `input_tax + ((1 - input_tax) * output_tax)` rather than `input_tax + output_tax`

src/components/swap/SwapModalFooter.tsx
```diff
@@ -209,3 +217,36 @@ export default function SwapModalFooter({
     </>
   )
 }
+
+function TokenTaxLineItem({ trade, type }: { trade: ClassicTrade; type: 'input' | 'output' }) {
```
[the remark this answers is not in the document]
cartcrom: I think there should be a higher-level refactor to share components between the swap details in the dropdown & the review modal details, whereas sharing a component for just these tax line items would be wonky since they have different styling

src/components/swap/AdvancedSwapDetails.tsx
```diff
@@ -176,3 +181,32 @@ export function AdvancedSwapDetails({ trade, allowedSlippage, syncing = false }:
...
+  let currency, percentage
+  if (type === 'input' && !trade.inputTax.equalTo(ZERO)) {
+    currency = trade.inputAmount.currency
+    percentage = trade.inputTax
+  } else if (type === 'output' && !trade.outputTax.equalTo(ZERO)) {
+    currency = trade.outputAmount.currency
+    percentage = trade.outputTax
+  } else {
+    return null
+  }
```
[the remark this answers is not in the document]
cartcrom: ooo i realized my if statements were overly complicated, was able to simplify without a helper ty

src/state/routing/types.ts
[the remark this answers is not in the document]
tinaszheng: hmm so i believe this assumes X% decrease in amount in means a X% decrease in amount out, which isnt right, but i also think this case is rare enough where we don't have to be ultra precise about it

i'd vote for doing the simpler but looser input_tax + output_tax until we can actually amend the `amountIn` calculated by the sdk, what do u think?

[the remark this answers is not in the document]
cartcrom: sounds good to me!

src/state/routing/types.ts
```diff
@@ -187,6 +194,23 @@ export class ClassicTrade extends Trade<Currency, Currency, TradeType> {
...
+    this.inputTax = inputTax
+    this.outputTax = outputTax
+  }
+
+  public get totalTaxRate(): Percent {
+    return this.inputTax.add(this.outputTax)
+  }
+
+  public get postTaxOutputAmount() {
+    return this.outputAmount.multiply(new Fraction(ONE).subtract(this.totalTaxRate))
```
tinaszheng, line 206: can you also add a comment about why we're doing the semi-jank total tax rate 

something like:
- ideally we should amend the `inputAmount` based on the input tax but its a bit tricky because V2Trade reconstructs the total `inputAmount` based on the swap routes
- TODO: Amend V2Trade objects in the v2-sdk to have a separate field for post-input tax routes

tinaszheng (approved):
🔥 🔥 🔥

src/constants/tax.ts
```diff
@@ -0,0 +1,75 @@
+import { ChainId, Currency, Percent } from '@uniswap/sdk-core'
+
+import { ZERO_PERCENT } from './misc'
+
+interface TokenTaxMetadata {
+  buyFee?: Percent
+  sellFee?: Percent
```
zzmp, line 7: It's weird to me that this is called `buyFee`/`sellFee` internally (to the file), but exported as `getInputTax`/`getOutputTax`.

src/constants/tax.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+const CHAIN_TOKEN_TAX_MAP: { [chainId in number]?: { [address in string]?: TokenTaxMetadata } } = {
+  [ChainId.MAINNET]: {
+    // BULLET
+    '0x8ef32a03784c8fd63bbf027251b9620865bd54b6': {
+      buyFee: new Percent(5, 100), // 5%
```
zzmp, line 14: Considering the duplication, it may be easier to understand if you just set some constants, eg

```
const ONE_PERCENT = new Percent(1, 100)
```

src/components/swap/AdvancedSwapDetails.tsx
```diff
@@ -176,3 +181,26 @@ export function AdvancedSwapDetails({ trade, allowedSlippage, syncing = false }:
...
+
+  if (percentage.equalTo(ZERO_PERCENT)) return null
+
+  return (
+    <RowBetween>
+      <MouseoverTooltip
+        text={
+          <Trans>
+            Some tokens take a fee when they are bought or sold, which is set by the token issuer. Uniswap does not
+            receive any share of these fees.
```
zzmp, line 197: Confusing wording. Can it just say "any of these fees", and omit "share"? Or even just "Uniswap does not receive these fees."?

src/components/swap/AdvancedSwapDetails.tsx
```diff
@@ -176,3 +181,26 @@ export function AdvancedSwapDetails({ trade, allowedSlippage, syncing = false }:
...
+          </Trans>
+        }
+      >
+        <ThemedText.BodySmall color="textSecondary">{`${currency.symbol} fee`}</ThemedText.BodySmall>
```
zzmp, line 201: Confusing (and *non-standard* way of expressing value - the currency symbol should be adjacent to the number, *not* separated by the word "fee".

src/components/swap/SwapModalFooter.tsx
```diff
@@ -41,9 +42,10 @@ const ConfirmButton = styled(ButtonError)`
   margin-top: 10px;
 `
 
-const DetailRowValue = styled(ThemedText.BodySmall)`
+const DetailRowValue = styled(ThemedText.BodySmall)<{ color?: string }>`
   text-align: right;
   overflow-wrap: break-word;
+  ${({ color }) => color && `color: ${color};`};
```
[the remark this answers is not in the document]
zzmp: You were passing a string into this, which could be interpolated on theme. Can you do that again, to avoid the use of a new hook (`useWarningColor`)? It'd look something like:

```
...<{ color?: keyof Theme }>`
  ...
  ${({ theme }) => color && `color: ${theme[color]}`};
```

src/components/swap/SwapModalFooter.tsx
```diff
@@ -209,3 +217,30 @@ export default function SwapModalFooter({
...
+  if (percentage.equalTo(ZERO_PERCENT)) return null
+
+  return (
+    <ThemedText.BodySmall>
+      <Row align="flex-start" justify="space-between" gap="sm">
+        <MouseoverTooltip
+          text={
+            <Trans>
+              Some tokens take a fee when they are bought or sold, which is set by the token issuer. Uniswap does not
+              receive any share of these fees.
```
zzmp, line 236: See language nits in other comment - both on this sentence, and on the presentation of the fee.

zzmp (changes requested):

src/components/swap/AdvancedSwapDetails.tsx
tinaszheng: this is the label for the token fee though

cartcrom: @zzmp this is a percentage line item, not a currency amount:
image

zzmp (approved):

tinaszheng (approved):

tinaszheng (approved):

PR #7197, opened 2023-08-21
feat: track connected wallet provider calls to be replaced with BE/infura

cbachmeier:
## Description

- Adds logging for calls to a user's connect wallet provider that will be replaced with calls to our Be/infura in the new multichain UX as noted in [this doc](https://www.notion.so/uniswaplabs/MultiChain-Web-UX-127bdec19777423abf12b7b7ad98a60f)
- This includes:
  - pool positions
  - multicall token address searches in the token selector
- Dependent on the addition of the new analytics event https://github.com/Uniswap/analytics-events/pull/52

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2723/add-amplitude-tracking-for-current-wallet-provider-calls
_Slack thread:_ https://uniswapteam.slack.com/archives/C047U65H422/p1692638017065729

### DIFF
--- a/src/hooks/useContract.ts
+++ b/src/hooks/useContract.ts
@@ -30,7 +30,7 @@ import WETH_ABI from 'abis/weth.json'
 import { sendAnalyticsEvent } from 'analytics'
 import { RPC_PROVIDERS } from 'constants/providers'
 import { WRAPPED_NATIVE_CURRENCY } from 'constants/tokens'
-import { useMemo } from 'react'
+import { useEffect, useMemo } from 'react'
 import { NonfungiblePositionManager, TickLens, UniswapInterfaceMulticall } from 'types/v3'
 import { V3Migrator } from 'types/v3/V3Migrator'
 import { getContract } from 'utils'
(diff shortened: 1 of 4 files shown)


src/lib/hooks/useCurrency.ts
```diff
@@ -46,6 +48,11 @@ export function useTokenFromActiveNetwork(tokenAddress: string | undefined): Tok
   const symbol = useSingleCallResult(tokenContract, 'symbol', undefined, NEVER_RELOAD)
   const symbolBytes32 = useSingleCallResult(tokenContractBytes32, 'symbol', undefined, NEVER_RELOAD)
   const decimals = useSingleCallResult(tokenContract, 'decimals', undefined, NEVER_RELOAD)
+  sendAnalyticsEvent(InterfaceEventName.WALLET_PROVIDER_USED, {
```
tinaszheng, line 51: so this one's a bit tricky i think, basically this event doesn't really measure how many times the `provider` is actually used. so `useSingleCallResult` uses rtk-query under the hood and the results are cached per block number. and here the `NEVER_RELOAD` flag means that the results are cached for the rest of the session. so it doesn't necessarily mean that every time this hook runs, the provider is used

im not super sure how to fix tbh, maybe an alternative would be to toss this in a `useEffect` where we estimate that the provider is only used every time a set of variable changes

tinaszheng (dismissed):

src/lib/hooks/useCurrency.ts
cbachmeier: I've moved the call up 1 level and wrapped it in a `useEffect` based on if the token retrieved changes and have the event only firing once when querying for a token
ex:  `0xf9c5ce0c5fa29b487b7329bb92d8e473c386f82b`
Screenshot 2023-08-23 at 2 35 35 PM

There's still the chance if the user continually queries the same token in the same session that the number might be slightly inflated but I think a marginal overestimate is acceptable

src/hooks/useContract.ts
```diff
@@ -145,11 +147,21 @@ export function useMainnetInterfaceMulticall() {
...
     withSignerIfPossible
   )
+  useEffect(() => {
+    if (contract && withSignerIfPossible !== false && account) {
+      sendAnalyticsEvent(InterfaceEventName.WALLET_PROVIDER_USED, {
+        source: 'useV3NFTPositionManagerContract',
+        contract,
+      })
+    }
+  }, [account, contract, withSignerIfPossible])
```
JFrankfurt, line 163:
```suggestion
  useEffect(() => {
    if (contract && account) {
      sendAnalyticsEvent(InterfaceEventName.WALLET_PROVIDER_USED, {
        source: 'useV3NFTPositionManagerContract',
        contract,
      })
    }
  }, [account, contract])
```

cbachmeier: when `withSignerIfPossible` is undefined in `useContract` it will be defaulted to `true` so we want to capture that case, that's why the check is explicitly only ignoring if `withSignerIfPossible === false`

JFrankfurt: why does that case matter here for the analytics?

JFrankfurt (approved):

PR #7201, opened 2023-08-22
fix: make popups appear above drawer and near top conditionally only …

natew:
…when drawer is open

## Description

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2599/transaction-modals-should-appear-above-the-mini-portfolio
_Slack thread:_
_Relevant docs:_

### Before

CleanShot 2023-08-21 at 14 22 10@2x

### After

sorry i accidentally made the change on the spore branch which is why it looks a lot nicer :P, but i cherry picked it here.

CleanShot 2023-08-21 at 14 17 11@2x

I made it adjust upwards when drawer is open since the default was down a bit:

https://github.com/Uniswap/interface/assets/12100/6978e770-fb9a-4090-859d-23d894c92845

### Reproducing the error

1. Trigger any popup, or hardcode `activePopups` in Popups/index to:

```
const activePopups = [
    { content: { type: PopupType.FailedSwitchNetwork, failedSwitchNetwork: ChainId.ARBITRUM_GOERLI }, key: '10', removeAfterMs: null }
  ] as const
```

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test


src/components/Popups/index.tsx
```diff
@@ -37,24 +39,36 @@ const StopOverflowQuery = `@media screen and (min-width: ${MEDIA_WIDTHS.deprecat
...
 }px)`
 
-const FixedPopupColumn = styled(AutoColumn)<{ extraPadding: boolean; xlPadding: boolean }>`
+type TopDistanceModifiers = {
+  bannerVisible: boolean
+  drawerOpen: boolean
+}
+
+const getTopDistance = ({ drawerOpen, bannerVisible }: TopDistanceModifiers) => {
+  return 64 + (drawerOpen ? -50 : 0) + (bannerVisible ? 8 : 0)
```
just-toby, line 48: any chance we can reference variable here and share it with the styled component in question? if possible it would be nice so we don't have to update in 2 places in the future

src/components/Popups/index.tsx
```diff
@@ -37,24 +39,36 @@ const StopOverflowQuery = `@media screen and (min-width: ${MEDIA_WIDTHS.deprecat
   MEDIA_WIDTHS.deprecated_upToMedium + 500
 }px)`
 
-const FixedPopupColumn = styled(AutoColumn)<{ extraPadding: boolean; xlPadding: boolean }>`
+type TopDistanceModifiers = {
+  bannerVisible: boolean
+  drawerOpen: boolean
+}
+
+const getTopDistance = ({ drawerOpen, bannerVisible }: TopDistanceModifiers) => {
```
just-toby, line 47: `getTopDistancePx` ?

just-toby (approved):

src/components/Popups/index.tsx
```diff
@@ -14,6 +15,8 @@ const MobilePopupWrapper = styled.div`
   max-width: 100%;
   margin: 0 auto;
   display: none;
+  padding-left: 20px;
+  padding-right: 20px;
```
natew, line 19: noticed the mobile toasts were touching the edge, fixed it:

CleanShot 2023-09-05 at 09 54 54@2x

src/components/Popups/index.tsx
```diff
@@ -33,28 +36,26 @@ const MobilePopupInner = styled.div`
...
+  z-index: ${Z_INDEX.modal};
+  transition: top ease-in-out 500ms;
 
   ${({ theme }) => theme.deprecated_mediaWidth.deprecated_upToSmall`
     display: none;
   `};
-
-  ${StopOverflowQuery} {
-    top: ${({ extraPadding, xlPadding }) => (xlPadding ? '72px' : extraPadding ? '72px' : '64px')};
-  }
```
natew, line 54: from what i see we moved the toast down an extra 8px when isNotOnMainnet was true, but i think we changed this functionality to no longer have a top bar banner at all? instead we use a bottom thing, so i'm removing this altogether..

src/state/logging.ts
```diff
@@ -43,7 +43,6 @@ export const sentryEnhancer = Sentry.createReduxEnhancer({
         userSlippageToleranceHasBeenMigratedToAuto: user.userSlippageToleranceHasBeenMigratedToAuto,
         userDeadline: user.userDeadline,
         timestamp: user.timestamp,
-        URLWarningVisible: user.URLWarningVisible,
```
natew, line 46: no other logic around this existed anymore so removed all this

just-toby (approved):

cartcrom (approved):

src/components/Popups/index.tsx
```diff
@@ -33,40 +33,31 @@ const MobilePopupInner = styled.div`
...
+}>`
   position: fixed;
-  top: ${({ extraPadding }) => (extraPadding ? '72px' : '64px')};
+  top: ${({ drawerOpen }) => `${64 + (drawerOpen ? -50 : 0)}px`};
   right: 1rem;
   max-width: 348px !important;
   width: 100%;
-  z-index: 3;
+  z-index: ${Z_INDEX.modal};
+  transition: top ease-in-out 500ms;
```
zzmp, line 45: This should use the themed transition tokens.

src/components/Popups/index.tsx
```diff
@@ -33,40 +33,31 @@ const MobilePopupInner = styled.div`
...
-  MEDIA_WIDTHS.deprecated_upToMedium + 500
-}px)`
-
-const FixedPopupColumn = styled(AutoColumn)<{ extraPadding: boolean; xlPadding: boolean }>`
+const FixedPopupColumn = styled(AutoColumn)<{
+  drawerOpen: boolean
+}>`
   position: fixed;
-  top: ${({ extraPadding }) => (extraPadding ? '72px' : '64px')};
+  top: ${({ drawerOpen }) => `${64 + (drawerOpen ? -50 : 0)}px`};
```
zzmp, line 40: Can you leave a comment explaining what this is doing?
Why is it better to move this up vertically rather than just use z-index to put it over the drawer?

zzmp (changes requested):

src/components/Popups/index.tsx
natew: hard to add a comment in the code there due to jsx tag, but it does use z-index to go over drawer, it just also shift itself up or down vertically so that when drawer is closed it goes below the header, whereas when its open, it shifts up. visually looks much better.

zzmp (approved):

PR #7205, opened 2023-08-22
feat: new app provider w/ fallback behavior

JFrankfurt:
No UX change.

This should make it possible to fall back from a failing rpc url to one that is more stable. 

To test:
- enable fallback_provider feature flag
image
- view mainnet
- do not connect a wallet
- observe successful rpc calls
- block infura (or quiknode) in dev tools
- observe rpc requests routed to unblocked rpc

You may also pull this down locally and update the default update interval to a very small value (1000 ms) and use devtools network blocker to alternatively disable quicknode/infura.


src/constants/extendedStaticJsonRpcProvider.ts
```diff
@@ -0,0 +1,53 @@
+import { deepCopy } from '@ethersproject/properties'
```
zzmp, line 1: This isn't a constant, it's a class. Can you move all of this to a new folder that deals with rpc, maybe `rpc/StaticJsonRpcProvider.ts`? (should start with a capital, as its a class)

src/constants/appRpcProvider.ts
```diff
@@ -0,0 +1,156 @@
+import { Network } from '@ethersproject/networks'
```
zzmp, line 1: rename file (to `rpc/AppRpcProvider.ts`)

src/constants/extendedStaticJsonRpcProvider.ts
```diff
@@ -0,0 +1,53 @@
...
+// This is the only file which should instantiate new Providers.
+// eslint-disable-next-line @typescript-eslint/no-restricted-imports
+import { StaticJsonRpcProvider } from '@ethersproject/providers'
+import { isPlain } from '@reduxjs/toolkit'
+
+import { AVERAGE_L1_BLOCK_TIME } from './chainInfo'
+import { CHAIN_IDS_TO_NAMES, SupportedInterfaceChain } from './chains'
+
+export class AppStaticJsonRpcProvider extends StaticJsonRpcProvider {
```
zzmp, line 10: Why not add this functionality into `AppRpcProvider`?

src/state/routing/gas.ts
```diff
@@ -26,7 +26,8 @@ export async function getApproveInfo(
   // If any of these arguments aren't provided, then we cannot generate approval cost info
   if (!account || !usdCostPerGas) return { needsApprove: false }
 
-  const provider = RPC_PROVIDERS[currency.chainId as SupportedInterfaceChain]
+  const appProvider = RPC_PROVIDERS[currency.chainId as SupportedInterfaceChain]
+  const provider = appProvider
```
zzmp, line 30:
```suggestion
  const provider = RPC_PROVIDERS[currency.chainId as SupportedInterfaceChain]
```

zzmp (dismissed):

src/constants/extendedStaticJsonRpcProvider.ts
JFrankfurt: I thought about that, but figured the simpler solution here is to let any valid provider be passed into the performance management layer. If we rolled this into the new AppRpcProvider we'd need to figure out a way to extend the argument providers' underlying behavior, and that feels messy to me.

src/constants/appRpcProvider.ts
```diff
@@ -0,0 +1,156 @@
...
+import { JsonRpcProvider, Provider } from '@ethersproject/providers'
+
+function now() {
+  return new Date().getTime()
+}
+
+function checkNetworks(networks: Array<Network>): Network | null {
+  let result: Network | undefined = undefined
```
cbachmeier, line 9: nit: This would let you remove the `?? null` in the return at the bottom
```suggestion
  let result: Network | null = null
```

cbachmeier (commented):
I'm on the deploy and blocked calls to `https://mainnet.infura.io` however I'm not seeing the calls rerouted to quicknode
Screenshot 2023-09-12 at 9 23 08 AM

JFrankfurt:
@cbachmeier I updated the vercel env variables and it's working for me now
image

cbachmeier (approved):

just-toby (commented):
how much usage do we expect to see to the quicknode account?

JFrankfurt:
@just-toby net RPC usage increase of about 0.5%

JFrankfurt:
@just-toby confirming here our out-of-band conversation and conclusion that current rpc pricing arrangement look good to go w/ this setup.

src/rpc/AppRpcProvider.ts
```diff
@@ -0,0 +1,156 @@
...
+    config.performance.lastEvaluated = now()
+  }
+
+  static sortProviders(providerEvaluations: FallbackProviderEvaluation[]) {
+    return providerEvaluations.sort((a, b) => {
+      const scoreA = a.performance.latency * (a.performance.failureRate + 1)
+      const scoreB = b.performance.latency * (b.performance.failureRate + 1)
+      return scoreA - scoreB
+    })
+  }
```
JFrankfurt, line 155: **example behavior**
Provider Name | Latency (ms) | Failure Rate | Score
-- | -- | -- | --
Provider B | 20 | 0 | 20
Provider A | 50 | 2 | 150
Provider E | 80 | 1 | 160
Provider C | 100 | 1 | 200
Provider D | 40 | 4 | 200

src/rpc/StaticJsonRpcProvider.ts
```diff
@@ -0,0 +1,53 @@
+import { deepCopy } from '@ethersproject/properties'
```
JFrankfurt, line 1: this file is unchanged from its prior implementation

src/constants/networks.ts
```diff
@@ -122,7 +126,11 @@ export const FALLBACK_URLS = {
...
 export const RPC_URLS = {
-  [ChainId.MAINNET]: [`https://mainnet.infura.io/v3/${INFURA_KEY}`, ...FALLBACK_URLS[ChainId.MAINNET]],
+  [ChainId.MAINNET]: [
+    `https://mainnet.infura.io/v3/${INFURA_KEY}`,
+    QUICKNODE_MAINNET_RPC_URL,
+    ...FALLBACK_URLS[ChainId.MAINNET],
+  ],
   [ChainId.GOERLI]: [`https://goerli.infura.io/v3/${INFURA_KEY}`, ...FALLBACK_URLS[ChainId.GOERLI]],
   [ChainId.SEPOLIA]: [`https://sepolia.infura.io/v3/${INFURA_KEY}`, ...FALLBACK_URLS[ChainId.SEPOLIA]],
   [ChainId.OPTIMISM]: [`https://optimism-mainnet.infura.io/v3/${INFURA_KEY}`, ...FALLBACK_URLS[ChainId.OPTIMISM]],
```
JFrankfurt, line 136: we could also add alternative urls for other networks

src/state/routing/gas.ts
```diff
@@ -26,7 +26,7 @@ export async function getApproveInfo(
   // If any of these arguments aren't provided, then we cannot generate approval cost info
   if (!account || !usdCostPerGas) return { needsApprove: false }
 
-  const provider = RPC_PROVIDERS[currency.chainId as SupportedInterfaceChain]
+  const provider = DEPRECATED_RPC_PROVIDERS[currency.chainId as SupportedInterfaceChain]
```
just-toby, line 29: why aren't we switching between these and the non-deprecated based on the feature flag here?

src/lib/hooks/routing/clientSideSmartOrderRouter.ts
```diff
@@ -16,7 +16,7 @@ export function getRouter(chainId: ChainId): AlphaRouter {
 
   const supportedChainId = asSupportedChain(chainId)
   if (supportedChainId) {
-    const provider = RPC_PROVIDERS[supportedChainId]
+    const provider = DEPRECATED_RPC_PROVIDERS[supportedChainId]
```
just-toby, line 19: why aren't we switching between these and the non-deprecated based on the feature flag here?

src/constants/providers.ts
```diff
@@ -1,75 +1,91 @@
...
-  [ChainId.POLYGON_MUMBAI]: new AppJsonRpcProvider(ChainId.POLYGON_MUMBAI),
-  [ChainId.CELO]: new AppJsonRpcProvider(ChainId.CELO),
-  [ChainId.CELO_ALFAJORES]: new AppJsonRpcProvider(ChainId.CELO_ALFAJORES),
-  [ChainId.BNB]: new AppJsonRpcProvider(ChainId.BNB),
-  [ChainId.AVALANCHE]: new AppJsonRpcProvider(ChainId.AVALANCHE),
-  [ChainId.BASE]: new AppJsonRpcProvider(ChainId.BASE),
+export const RPC_PROVIDERS: { [key in SupportedInterfaceChain]: AppRpcProvider } = {
+  [ChainId.MAINNET]: new AppRpcProvider([
+    new AppStaticJsonRpcProvider(ChainId.MAINNET, RPC_URLS[ChainId.MAINNET][0]),
+    new AppStaticJsonRpcProvider(ChainId.MAINNET, RPC_URLS[ChainId.MAINNET][1]),
```
just-toby, line 14: nit: should we do 
`RPC_URLS[ChainId.MAINNET].map(provider => new AppStaticJsonRpcProvider(ChainId.MAINNET, provider)`

so we can just add/remove options from the RPC_URLS arrays later?

or is there some reason we only want 2 providers here?

src/rpc/AppRpcProvider.ts
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * AppRpcProvider is an extension of the JsonRpcProvider class from the ethers.js library.
+ * This provider balances requests among multiple JSON-RPC endpoints.
+ */
+export default class AppRpcProvider extends JsonRpcProvider {
+  readonly providerEvaluations: ReadonlyArray<FallbackProviderEvaluation>
+  readonly evaluationInterval: number
+
+  constructor(providers: JsonRpcProvider[], evaluationInterval = 30000) {
```
just-toby, line 57: nit: `evaluationIntervalMs` ?

src/rpc/AppRpcProvider.ts
```diff
@@ -0,0 +1,156 @@
...
+    if (providers.length === 0) throw new Error('providers array empty')
+    providers.forEach((provider, i) => {
+      if (!Provider.isProvider(provider)) throw new Error(`invalid provider ${i}`)
```
just-toby, line 60: do we want to throw here without catching it ?  or do we expect to catch the error at the callsite?
we could also filter the list to just the valid providers.

src/rpc/AppRpcProvider.ts
```diff
@@ -0,0 +1,156 @@
...
+    })
+    const agreedUponNetwork = checkNetworks(providers.map((p) => p.network))
+    if (!agreedUponNetwork) throw new Error('networks mismatch')
```
just-toby, line 63: same question here

just-toby (approved):

src/featureFlags/flags/fallbackProvider.ts
```diff
@@ -0,0 +1,9 @@
+import { BaseVariant, FeatureFlag, useBaseFlag } from '../index'
+
+export function useFallbackProviderFlag(): BaseVariant {
+  return useBaseFlag(FeatureFlag.fallbackProvider)
+}
+
+export function useFallbackProviderFlagEnabled(): boolean {
```
zzmp, line 7:
```suggestion
export function useFallbackProvider(): boolean {
```

zzmp (changes requested):
Some missing deps (see lint) and echoing @just-toby's concerns.

src/constants/providers.ts
zzmp: ⬆️ this seems like a more sustainable pattern.

src/rpc/AppRpcProvider.ts
```diff
@@ -0,0 +1,156 @@
...
+   * params - The parameters for the JSON-RPC method.
+   * Returns a Promise that resolves with the result of the JSON-RPC method.
+   * Throws an error if all providers fail to perform the operation.
+   */
+  async perform(method: string, params: { [name: string]: any }): Promise<any> {
+    // Periodically evaluate all providers
+    const currentTime = now()
+    this.providerEvaluations.forEach((providerEval) => {
+      if (currentTime - providerEval.performance.lastEvaluated >= this.evaluationInterval) {
+        this.evaluateProvider(providerEval)
```
zzmp, line 93: This is async? Should be commented that this will not affect the current perform call.

src/rpc/AppRpcProvider.ts
```diff
@@ -0,0 +1,156 @@
...
+      }
+    })
+
+    // Sort providers by performance score (latency and failure rate)
+    const sortedEvaluations = AppRpcProvider.sortProviders(this.providerEvaluations.slice())
```
zzmp, line 98: Sort as part of evaluateProvider every time an evaluation comes in, so that you're not re-sorting every time you call `perform`.

src/rpc/AppRpcProvider.ts
```diff
@@ -0,0 +1,156 @@
...
+        })
+      )
+
+      // Any success is good enough
+      for (let i = 0; i < results.length; i++) {
+        if (typeof results[i] === 'string') return results[i]
+      }
+
+      // They were all an error; pick the first error
+      throw results[0]
```
zzmp, line 117: There should never be a `sendTransaction` sent to a backend, as these are read-only backends.

src/rpc/AppRpcProvider.ts
```diff
@@ -0,0 +1,156 @@
...
+    } else {
+      for (const { provider, performance } of sortedEvaluations) {
+        try {
+          return provider.perform(method, params)
```
zzmp, line 121: This is async - you need to await to catch the failure:

```suggestion
          return await provider.perform(method, params)
```

src/rpc/AppRpcProvider.ts
```diff
@@ -0,0 +1,156 @@
...
+        } catch (error) {
+          performance.failureRate++ // Increment failure rate
```
zzmp, line 123: Log the error (`console.warn`).

src/rpc/AppRpcProvider.ts
zzmp: Do you want to also reset the failure rate on the successful request?
Alternatively, should the failure rate be a rate, ie failures per X requests, in which case you also want to track successes?

src/rpc/AppRpcProvider.ts
zzmp: Isn't it possible for something which fails quickly to rank before something which succeeds slowly?

src/rpc/AppRpcProvider.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  test('handles sendTransaction', async () => {
+    const hash = '0x123'
+    mockProvider1.sendTransaction.mockResolvedValue({ hash } as TransactionResponse)
+    const provider = new AppRpcProvider([mockProvider1])
+
+    const result = await provider.perform('sendTransaction', { signedTransaction: '0xabc' })
+    expect(result).toBe(hash)
+  })
+
```
zzmp, line 60: Add a test for sends which are not sendTransaction.
Add a test for sends which fail, as they currently will not fall back, because they are not caught with an `await`.

zzmp (changes requested):

src/rpc/AppRpcProvider.ts
JFrankfurt: yes, but its failure rate would climb rapidly and all failed requests would be retried automatically at lower ranked venues immediately, so the ux effect is minimal imo. I'm fine with making tweaks to the algo if you have any in mind.

src/state/routing/gas.ts
JFrankfurt: because getApproveInfo is not a hook, and the current feature gate api is hook-only. Since this call site will be called very rarely compared to others, I chose to use the deprecated version instead and make the change to the new one when we remove the gate. I can see a reasonable argument for doing the opposite, though!

src/constants/providers.ts
JFrankfurt: great idea!

src/rpc/AppRpcProvider.ts
JFrankfurt: https://github.com/ethers-io/ethers.js/blob/master/packages/providers/src.ts/fallback-provider.ts

For our use case here this is probably not necessary, but I approached the task with somewhat of a mind towards publishing this later.

src/rpc/AppRpcProvider.ts
```diff
@@ -0,0 +1,156 @@
...
+    if (providers.length === 0) throw new Error('providers array empty')
+    providers.forEach((provider, i) => {
+      if (!Provider.isProvider(provider)) throw new Error(`invalid provider ${i}`)
+    })
+    const agreedUponNetwork = checkNetworks(providers.map((p) => p.network))
+    if (!agreedUponNetwork) throw new Error('networks mismatch')
+
+    super(providers[0].connection)
+    this.providerEvaluations = providers.map((provider) => {
+      return Object.freeze({
```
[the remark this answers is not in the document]
JFrankfurt: https://github.com/ethers-io/ethers.js/blob/master/packages/providers/src.ts/fallback-provider.ts#L406

Another case of working from the upstream example here.

src/rpc/AppRpcProvider.ts
```diff
@@ -0,0 +1,156 @@
...
+        provider,
+        performance: {
+          latency: Number.MAX_SAFE_INTEGER,
+          failureRate: 0,
+          lastEvaluated: 0,
+        },
+      })
+    })
+
+    this.evaluationInterval = evaluationInterval
```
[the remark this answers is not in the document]
JFrankfurt: I don't follow. Why is the eval interval unnecessary?

src/rpc/AppRpcProvider.ts
JFrankfurt: The ultimate goal is that this will be extended to support signers

src/constants/providers.ts
```diff
@@ -1,75 +1,91 @@
...
 import { ChainId } from '@uniswap/sdk-core'
+import AppRpcProvider from 'rpc/AppRpcProvider'
+import AppStaticJsonRpcProvider from 'rpc/StaticJsonRpcProvider'
 
-import { AVERAGE_L1_BLOCK_TIME } from './chainInfo'
-import { CHAIN_IDS_TO_NAMES, SupportedInterfaceChain } from './chains'
+import { SupportedInterfaceChain } from './chains'
 import { RPC_URLS } from './networks'
 
-class AppJsonRpcProvider extends StaticJsonRpcProvider {
```
JFrankfurt, line 12: moved to its own file

src/constants/providers.ts
```diff
@@ -1,75 +1,91 @@
...
+    new AppStaticJsonRpcProvider(ChainId.AVALANCHE, RPC_URLS[ChainId.AVALANCHE][0]),
+    new AppStaticJsonRpcProvider(ChainId.AVALANCHE, RPC_URLS[ChainId.AVALANCHE][1]),
+  ]),
+  [ChainId.BASE]: new AppRpcProvider(ChainId.BASE, [
+    new AppStaticJsonRpcProvider(ChainId.BASE, RPC_URLS[ChainId.BASE][0]),
+    new AppStaticJsonRpcProvider(ChainId.BASE, RPC_URLS[ChainId.BASE][1]),
+  ]),
+}
+
+export const DEPRECATED_RPC_PROVIDERS: { [key in SupportedInterfaceChain]: AppStaticJsonRpcProvider } = {
```
JFrankfurt, line 70: retained for purpose of feature gating

just-toby (approved):

just-toby (approved):

just-toby (approved):

src/rpc/AppRpcProvider.ts
```diff
@@ -0,0 +1,158 @@
...
+    // Periodically evaluate all providers
+    const currentTime = Date.now()
+    // Note that this async action will not affect the current perform call
+    this.providerEvaluations.forEach((providerEval) => {
+      if (currentTime - providerEval.performance.lastEvaluated >= this.evaluationIntervalMs) {
+        this.evaluateProvider(providerEval)
+      }
+    })
+
+    this.providerEvaluations = AppRpcProvider.sortProviders(this.providerEvaluations.slice())
```
zzmp, line 91: sorting should be moved after `evaluateProvider` so it occurs less frequently and does not block `perform` from proceeding.

src/rpc/AppRpcProvider.ts
```diff
@@ -0,0 +1,158 @@
...
+        })
+      )
+
+      // Any success is good enough
+      for (let i = 0; i < results.length; i++) {
+        if (typeof results[i] === 'string') return results[i]
+      }
+
+      // They were all an error; pick the first error
+      throw results[0]
```
zzmp, line 111: I think we should remove this. It is dead code.

PR #7206, opened 2023-08-22
fix: ensure lists are present before running update

JFrankfurt:
## Description

Fix sentry error:
`TypeError: Cannot read properties of undefined (reading 'current')`

will hotfix to prod when this is merged

_Linear ticket:_
_Slack thread:_ https://uniswapteam.slack.com/archives/C03NZGRTS3A/p1692724036927089
_Relevant docs:_

### DIFF
--- a/src/state/lists/updater.ts
+++ b/src/state/lists/updater.ts
@@ -1,7 +1,8 @@
 import { getVersionUpgrade, VersionUpgrade } from '@uniswap/token-lists'
 import { useWeb3React } from '@web3-react/core'
-import { DEFAULT_LIST_OF_LISTS, UNI_EXTENDED_LIST, UNI_LIST, UNSUPPORTED_LIST_URLS } from 'constants/lists'
+import { DEFAULT_LIST_OF_LISTS, UNSUPPORTED_LIST_URLS } from 'constants/lists'
 import TokenSafetyLookupTable from 'constants/tokenSafetyLookup'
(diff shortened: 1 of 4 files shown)


cbachmeier (approved):

src/state/lists/updater.ts
```diff
@@ -23,7 +23,8 @@ export default function Updater(): null {
   const listsState = useAppSelector((state) => state.lists)
 
   useEffect(() => {
-    TokenSafetyLookupTable.update(listsState)
+    const keysArePresent = [UNI_EXTENDED_LIST, UNI_LIST, ...UNSUPPORTED_LIST_URLS].every((key) => listsState.byUrl[key])
```
zzmp, line 26: This is essentially the same as checking for hydration, but doesn't check for type safety at the point of use. Can you amend it in two ways:

- check for type safety at the point of use (ie in `tokenSafetyLookup#update`)
- this check is essentially the same as checking for hydration, but it couples `keysArePresent` to the implementation of `tokenSafetyLookup`. Can you instead add a hook to check if the state is hydrated? (if you search for "hydrated" you'll find similar logic already in the codebase.)

zzmp (changes requested):

src/state/lists/updater.ts
JFrankfurt: 1. is proving difficult while retaining the flexibility of string indexing byUrl

JackShort (approved):

PR #7209, opened 2023-08-22
fix: isolate assertions into single tests

JFrankfurt:
## Description

Cypress was failing on main. This isolates some flakey spy assertions into their own contexts.


JackShort (approved):
ty for this

tinaszheng (approved):

PR #7210, opened 2023-08-22
feat: redirect /address links to the wallet download landing page or app store

natew:
…app store

## Description

going to 

https://app.uniswap.org/#/address/0x50EC05ADe8280758E2077fcBC08D878D4aef79C3

would 404, now it will redirect to our app landing page, or ios app store

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2693/can-we-add-a-redirect-link-for-external-profile-urls-when-a-user-is
_Slack thread:_
_Relevant docs:_

### Before

CleanShot 2023-08-22 at 12 16 23

### After

CleanShot 2023-08-22 at 12 16 32

### Reproducing the error

1. 

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test


src/pages/App.tsx
```diff
@@ -185,6 +185,18 @@ export default function App() {
     [account]
   )
 
+  // redirect address to landing pages until implemented
+  const shouldRedirectToAppInstall = pathname?.startsWith('/address/')
```
cbachmeier, line 189: how do users get to `/address`? are we routing them from the mobile app?

src/pages/App.tsx
```diff
@@ -185,6 +185,18 @@ export default function App() {
...
+  useLayoutEffect(() => {
+    if (shouldRedirectToAppInstall) {
+      window.location.href = getDownloadAppLink()
+    }
+  }, [shouldRedirectToAppInstall])
+
+  if (shouldRedirectToAppInstall) {
```
cbachmeier, line 196: can this be inside of the first `if`?

src/pages/App.tsx
natew: yes the share button from the mobile app profile screen

src/pages/App.tsx
natew: the first effect redirects and the second if just avoids rendering stuff, not sure which you mean specifically. i was worried about redirecting directly in the render as i'm not sure if thats fully synchronous and it feels more like a side-effect to me, but it's possible we could just do this entirely in the single render / if?

natew: the effect could go in the if as well but i guess technically weird looking as it breaks rules of hooks, though in this case i don't think there's any issue i just feel like worth avoiding the weirdness

cbachmeier: this is what I was imagining to reduce redundancy
```
useLayoutEffect(() => {
    if (shouldRedirectToAppInstall) {
      window.location.href = getDownloadAppLink()
      return null
    }
  }, [shouldRedirectToAppInstall]) 
  ```
  And just getting rid of the second if statement

cbachmeier (approved):

src/pages/App.tsx
natew: the return null is in the actual component render though, the intention is to avoid displaying anything / doing extra work while the browser redirects

just-toby (approved):

PR #7211, opened 2023-08-23
chore: remove unused assets via bash script

grabbou:
Fixes https://linear.app/uniswap/issue/WEB-1324/assets-folder-in-interface-has-unused-icons-needs-lint-rule

This is the script that I run to detect unused assets. I will follow-up with removal, but figured that we could either: a) keep this script for future reference, b) add it as a part of "npm scripts", c) plug in with the testing pipeline to fail early if it detects those files.

The initial suggestion was to have an ESLint rule, but I don't think that is possible and intuitive - we don't usually open assets with VSCode, so ESLint wouldn't trigger issues unless run on the CI (or locally, but I guess most of us just rely on ESLint extension).

Current output:
```
Unreferenced files:
./src/components/TokenSafety/verified.svg
./src/components/About/images/About_BG_Dark.jpg
./src/components/About/images/About_BG_Light.jpg
./src/assets/svg/static_route.svg
./src/assets/svg/ethereum_square_logo.svg
./src/assets/svg/fiat_mask.svg
./src/assets/svg/phantom-icon.svg
./src/assets/images/arbitrumCircle.png
./src/assets/images/welcomeModal-light@3x.jpg
./src/assets/images/arrow-right.svg
./src/assets/images/welcomeModal-dark@2x.jpg
./src/assets/images/welcomeModal-dark@3x.jpg
./src/assets/images/welcomeModal-light@2x.jpg
./src/assets/images/welcomeModal-dark.jpg
./src/assets/images/welcomeModal-light.jpg
./src/assets/images/celoCircle.png
./public/nft/svgs/marketplaces/looksrare-grey.svg
./public/nft/svgs/marketplaces/opensea-grey.svg
./public/nft/svgs/marketplaces/x2y2-grey.svg
```


scripts/find-unused-assets.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# We look for assets in the /src and /public and then, check if they are referenced in
+# the code, typically in either /src, /public (manifest.json) or /functions
+while IFS= read -r -d $'\0' file; do
+    name="$(basename "$file")"
+    grep -rn -F -q "$name" ./src ./functions ./public
+    if [ $? -ne 0 ]; then
+      rm $file
+    fi
+done < <(find ./src ./public -type f \( -name "*.png" -o -name "*.svg" -o -name "*.jpg" \) -print0)
```
zzmp, line 14: It would be nice to pull out relevant file extensions into a variable at the top of the file, so it can be expanded without having to read too much code, eg

```
ASSET_EXTENSIONS=["*.png", "*.svg", "*.jpg"]
```

zzmp (changes requested):
i love this - just some minor questions

JFrankfurt (changes requested):
This is great, but we should make sure to preserve cross-platform functionality and write this in JS like the rest of the `scripts` files

scripts/find-unused-assets.sh
```diff
@@ -0,0 +1,26 @@
...
+# Array to hold the list of unreferenced files
+unreferenced_files=()
+
+# We look for assets in the /src and /public and then, check if they are referenced in
+# the code, typically in either /src, /public (manifest.json) or /functions
+while IFS= read -r -d $'\0' file; do
+    name="$(basename "$file")"
+    grep -rn -F -q "$name" ./src ./functions ./public
+    if [ $? -ne 0 ]; then
+      rm $file
```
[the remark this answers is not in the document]
grabbou: Oh, that is unexpected result of my local testing. This should be added to the array <facepalm>.

grabbou:
@JFrankfurt do we have an equivalent of `grep` and `find` in Node.js that delivers results equally fast? I feel like doing this in Node.js will take forever, given the size of the codebase. 

The script was intended to be run in the CI originally (and typically only there), so I didn't consider Windows compatibility important. Otherwise, it should run just fine.

Unfortunately, I am not familiar with the Windows platform and their support for shell commands (PowerShell comes to my mind), so I would consider this unsupported, unless you want to take a stab at it!

scripts/find-unused-assets.sh
grabbou: Yeah, that's an option. Will make this code slightly less readable, but will on the other hand make it more explicit to modify parameters. 

There are few interesting gotchas here, like the fact that instead of piping `find` to `while` loop, we're doing process substitution to avoid spawning a subprocess inside `while` loop (which will make it impossible to access and modify the array).

grabbou: That said, I will send an update!

grabbou:
> i love this - just some minor questions
> 
> we'd also want this to be executable easily - could it be run through a package script: `yarn clean-assets`? we'd also want to execute the package script in the `test.yml`, see `yarn deduplicate` as an example.

Yup, that's the plan - I would run it BY DEFAULT early on in the testing pipeline. Worst case scenario (if you're on Windows), you can always look up the CI results (which will be quite explicit) and just remove the files. Whether that scenario materialises depends on the previous comment.

CC: @JFrankfurt I know you're on Windows, so I appreciate your perspective here! 🙏

JFrankfurt:
here ya go

```js
/* eslint-env node */

const fs = require('fs')
const path = require('path')

async function* findFiles(dir, patterns) {
  const entries = await fs.promises.readdir(dir, { withFileTypes: true })
  for (const entry of entries) {
    const fullPath = path.join(dir, entry.name)
    if (entry.isDirectory()) {
      yield* findFiles(fullPath, patterns)
    } else if (patterns.some((pattern) => pattern.test(entry.name))) {
      yield fullPath
    }
  }
}

async function searchInFiles(file, directories) {
  const name = path.basename(file)
  for (const dir of directories) {
    const files = await fs.promises.readdir(dir, { withFileTypes: true })
    for (const entry of files) {
      const fullPath = path.join(dir, entry.name)
      if (entry.isDirectory()) {
        if (await searchInFiles(file, [fullPath])) return true
      } else {
        const content = await fs.promises.readFile(fullPath, 'utf8')
        if (content.includes(name)) return true
      }
    }
  }
  return false
}

async function main() {
  const unreferencedFiles = []
  const patterns = [/\.png$/, /\.svg$/, /\.jpg$/]
  const searchDirs = ['./src', './functions', './public']

  for await (const file of findFiles('./src', patterns)) {
    unreferencedFiles.push(file)
  }

  for await (const file of findFiles('./public', patterns)) {
    unreferencedFiles.push(file)
  }

  for (const file of unreferencedFiles) {
    if (!(await searchInFiles(file, searchDirs))) {
      console.log('Unreferenced file:', file)
    }
  }

  process.exit(0)
}

main().catch((err) => {
  console.error(err)
  process.exit(1)
})
```

grabbou:
@JFrankfurt thanks for the snippet, but I don't think this was necessary. We were at the point of discussing performance and potential implications of using Node built-ins instead. If we reached a conclusion that it was insignificant or "worth a try", I would've happily re-written it myself.

Now, having said that, I did check your code locally and I think while it runs 2/3x slower, it is still reasonably fast (sub 6s).

Since you literally did the subject of this PR, I will close and re-assign to you. You just need to plug it to the CI / scripts as per @zzmp suggestions and are good to go.

PR #7212, opened 2023-08-23
fix: eliminate flake from account connect race condition

JFrankfurt:
## Description

The settings menu was occasionally opened before account connection in the cypress tests.

~We have an updater that closes modals when accounts change, so when an account would connect and cypress had already opened the settings menu it would stall and be unable to continue. This diff adds an optional argument to the callback returned by `useCloseModal` allowing consumers to specify which modal should be closed. (As before if no option is specified any open modal will be closed.)~

edit:
- updater `useEffect` call to close modals has been removed. 
- ApplicationModal.WALLET has been removed

I also took some time to clean up the code and test a little.


src/state/application/hooks.ts
```diff
@@ -88,9 +88,21 @@ export function useToggleModal(modal: ApplicationModal): () => void {
...
+      if (modal && currentlyOpenModal === modal) {
+        // If a modal is specified, then we only close if that modal is open.
+        dispatch(setOpenModal(null))
+      } else if (!modal) {
+        // If no modal is specified, then we close any open modal.
+        dispatch(setOpenModal(null))
+      }
+    },
+    [currentlyOpenModal, dispatch]
+  )
```
JFrankfurt, line 105: this new callback behavior is what fixes the bug

src/state/application/updater.ts
```diff
@@ -20,7 +20,7 @@ export default function Updater(): null {
   useEffect(() => {
     if (account && account !== previousAccountValue.current) {
       previousAccountValue.current = account
-      closeModal()
+      closeModal(ApplicationModal.WALLET)
```
JFrankfurt, line 23: using the new closeModal behavior

cypress/e2e/swap/settings.test.ts
```diff
@@ -16,14 +16,24 @@ describe('Swap settings', () => {
...
+    // Visit the swap page with specific feature flag
     cy.visit('/swap', { featureFlags: [FeatureFlag.uniswapXEnabled] })
-    cy.get(getTestSelector('open-settings-dialog-button')).click()
-    cy.get(getTestSelector('mobile-settings-menu')).should('exist')
-    cy.contains('Max slippage').should('exist')
-    cy.contains('Transaction deadline').should('exist')
-    cy.contains('UniswapX').should('exist')
-    cy.contains('Local routing').should('exist')
-    cy.get(getTestSelector('mobile-settings-scrim')).click()
+
```
JackShort, line 24: can't you just do this `cy.get(getTestSelector('web3-status-connected'))` to wait for it to be connected?

JFrankfurt: Yeah, but the underlying behavior that was causing this flake is actually an issue in our app. If you move quickly after page load you can hit this bug in prod today.

src/components/Settings/index.tsx
```diff
@@ -136,22 +146,18 @@ export default function SettingsTab({
...
-          onClick={toggleMenu}
-        />
-        {isOpenDesktop && !isMobile && <MenuFlyout>{Settings}</MenuFlyout>}
-      </Menu>
-      {isMobile && (
-        <Portal>
+    <Menu ref={node}>
+      <MenuButton disabled={!isChainSupported || chainId !== connectedChainId} isActive={isOpen} onClick={toggleMenu} />
+      {isOpenDesktop && <MenuFlyout>{Settings}</MenuFlyout>}
+      <Portal>
```
JackShort, line 152: is there a reason you moved portal outside of the mobile check?

src/components/Settings/index.tsx
```diff
@@ -136,22 +146,18 @@ export default function SettingsTab({
...
+        {isOpenMobile && (
           <MobileMenuContainer data-testid="mobile-settings-menu">
-            <Scrim testId="mobile-settings-scrim" onClick={toggleMenu} open={isOpenMobile} />
+            <Scrim onClick={closeMenu} $open={isOpenMobile} />
```
JackShort, line 155: nit: `$open` can just be set to `true` here

src/state/application/hooks.ts
```diff
@@ -88,9 +88,22 @@ export function useToggleModal(modal: ApplicationModal): () => void {
...
   const dispatch = useAppDispatch()
-  return useCallback(() => dispatch(setOpenModal(null)), [dispatch])
+  const currentlyOpenModal = useAppSelector((state: AppState) => state.application.openModal)
+  return useCallback(
+    (modal?: ApplicationModal) => {
+      // Check if modal is defined and either matches the currently open modal
+      // or is not a valid instance of ApplicationModal enum (i.e., if it is passed a click event)
+      if ((modal && currentlyOpenModal === modal) || (modal && !Object.values(ApplicationModal).includes(modal))) {
+        dispatch(setOpenModal(null))
+      } else if (!modal) {
```
JackShort, line 100: nit: make this the first check so you don't have to check in both places in the other branch

src/state/application/hooks.ts
```diff
@@ -88,9 +88,22 @@ export function useToggleModal(modal: ApplicationModal): () => void {
...
-export function useCloseModal(): () => void {
+export function useCloseModal() {
   const dispatch = useAppDispatch()
-  return useCallback(() => dispatch(setOpenModal(null)), [dispatch])
+  const currentlyOpenModal = useAppSelector((state: AppState) => state.application.openModal)
+  return useCallback(
+    (modal?: ApplicationModal) => {
+      // Check if modal is defined and either matches the currently open modal
+      // or is not a valid instance of ApplicationModal enum (i.e., if it is passed a click event)
+      if ((modal && currentlyOpenModal === modal) || (modal && !Object.values(ApplicationModal).includes(modal))) {
```
JackShort, line 98: in what cases would it pass a click event? also the function takes an application modal or undefined as the parameter so how would this happen?

cypress/e2e/swap/settings.test.ts
JFrankfurt: This pr does resolve the bug. It also updates the test to be easier to read and more idiomatic.

src/components/Settings/index.tsx
JFrankfurt: The createPortal api (which Portal uses under the hood) specifies child/container elements in a way that feels static to me. This isn't important to me though. 

https://react.dev/reference/react-dom/createPortal

src/state/application/hooks.ts
JFrankfurt: The `@reach/dialog` `DialogOverlay` component is mistyped internally and claims to not send a click event on `onDismiss`, but it actually does send one. This caused a bug (which e2e tests caught!!) and means we need to be extra defensive here (or could do it in the onDismiss handler too).

JFrankfurt: moving the handling of this to the onDismiss handler

src/components/Settings/index.tsx
JFrankfurt: reverted this

src/components/Settings/index.tsx
```diff
@@ -10,20 +10,29 @@ import useDisableScrolling from 'hooks/useDisableScrolling'
...
-import { CloseIcon, Divider, ThemedText } from 'theme'
+import { Divider, ThemedText } from 'theme'
 import { Z_INDEX } from 'theme/zIndex'
 
 import MaxSlippageSettings from './MaxSlippageSettings'
 import MenuButton from './MenuButton'
 import RouterPreferenceSettings from './RouterPreferenceSettings'
 import TransactionDeadlineSettings from './TransactionDeadlineSettings'
 
+const CloseButton = styled.button`
```
cbachmeier, line 28: Close Button is no longer aligned vertically with Header Title
Screenshot 2023-08-24 at 10 12 52 AM

JFrankfurt: Are you sure? the button on the left is 24px wide, and the `Settings` text wrapper has a 24px padding on the right.

I've got a ruler up on the screen and it stays in the same spot when I flip between prod and this vercel build.

JackShort: charlie is right i am observing the same thing and its different than whats on prod

JFrankfurt: idk how I'm not seeing this

I'll check other browsers

JackShort: oh thats a good point - we both use firefox lol

JFrankfurt: oh lmao I misread--I thought @cbachmeier was saying it's not centered in the screen

ok I see it, will fix.

JFrankfurt: fixed

src/components/Settings/index.tsx
```diff
@@ -26,11 +26,13 @@ import RouterPreferenceSettings from './RouterPreferenceSettings'
 import TransactionDeadlineSettings from './TransactionDeadlineSettings'
 
 const CloseButton = styled.button`
-  cursor: pointer;
-  color: ${({ theme }) => theme.textPrimary};
   background: transparent;
   border: none;
+  color: ${({ theme }) => theme.textPrimary};
```
cbachmeier, line 31: merging in from main with spore update and changing to neutral1 should fix

cbachmeier (approved):
LGTM just need to rebase and fix that 1 style

src/state/application/reducer.ts
```diff
@@ -43,7 +43,6 @@ export enum ApplicationModal {
   TAX_SERVICE,
   TIME_SELECTOR,
   VOTE,
-  WALLET,
```
JackShort, line 46: Is this really necessary?

JFrankfurt: no, but as @cbachmeier pointed out, it's dead code now

JackShort (approved):

PR #7213, opened 2023-08-23
fix: null check on elapsedTime for some browsers

just-toby:
## Description

sentry issue: https://linear.app/uniswap/issue/WEB-2733/elapsedtime-is-undefined-occasionally

on iOS webviews (not safari, embedded browsers in other apps), `performance.mark` does not return  a `PerformanceEntry` object. as a quick fix we should just fail gracefully and not log performance metrics for these browsers.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2733/elapsedtime-is-undefined-occasionally

### Reproducing the error

1. load the app in an embedded browser on iOS (e.g. ...)

### QA (ie manual testing)

- [x] hard-coded `elapsedTime` to be undefined and ensured the app fails gracefully and doesn't log the perf metrics

#### Devices

- Mobile Safari UI/WKWebView

### DIFF
--- a/src/tracing/utils.ts
+++ b/src/tracing/utils.ts
@@ -2,7 +2,8 @@
  * Returns the time elapsed between page load and now.
  * @param markName the identifier for the performance mark to be created and measured.
  */
-export function calculateElapsedTimeWithPerformanceMark(markName: string): number {
+export function calculateElapsedTimeWithPerformanceMark(markName: string): number | undefined {
   const elapsedTime = performance.mark(markName)
+  if (!elapsedTime) return undefined
   return elapsedTime.startTime
 }


JFrankfurt (approved):

src/tracing/utils.ts
```diff
@@ -2,7 +2,8 @@
  * Returns the time elapsed between page load and now.
  * @param markName the identifier for the performance mark to be created and measured.
  */
-export function calculateElapsedTimeWithPerformanceMark(markName: string): number {
+export function calculateElapsedTimeWithPerformanceMark(markName: string): number | undefined {
   const elapsedTime = performance.mark(markName)
+  if (!elapsedTime) return undefined
```
zzmp, line 7: This is fine as a hotfix, but we are cutting a section of browsers out of our metrics by not polyfilling this. Do we want to find some other way to get at an equivalent value?

zzmp (changes requested):

src/tracing/utils.ts
just-toby: @zzmp yes we most likely do. i think it's fine to fix the crash now and follow up with a fix to get the metrics for those iOS browsers. right now, the iOS browsers aren't reporting metrics anyway so this just helps us on the sentry side

zzmp: sgtm - please just make sure the linear ticket stays open :)

zzmp (approved):

PR #7214, opened 2023-08-23
fix: null check on elapsedTime for some browsers

just-toby:
hotfix #7213  to reduce the error counts in prod

### DIFF
--- a/src/tracing/utils.ts
+++ b/src/tracing/utils.ts
@@ -2,7 +2,8 @@
  * Returns the time elapsed between page load and now.
  * @param markName the identifier for the performance mark to be created and measured.
  */
-export function calculateElapsedTimeWithPerformanceMark(markName: string): number {
+export function calculateElapsedTimeWithPerformanceMark(markName: string): number | undefined {
   const elapsedTime = performance.mark(markName)
+  if (!elapsedTime) return undefined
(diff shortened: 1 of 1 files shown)


src/tracing/utils.ts
```diff
@@ -2,7 +2,8 @@
  * Returns the time elapsed between page load and now.
  * @param markName the identifier for the performance mark to be created and measured.
  */
-export function calculateElapsedTimeWithPerformanceMark(markName: string): number {
+export function calculateElapsedTimeWithPerformanceMark(markName: string): number | undefined {
   const elapsedTime = performance.mark(markName)
+  if (!elapsedTime) return undefined
```
zzmp, line 7: This makes sense as a hotfix. For a more permanent fix (to `main`) do we want to find some other way to get at this value, so we're not just cutting some browsers out of our metrics?

zzmp (changes requested):

src/tracing/utils.ts
just-toby: @zzmp yes this PR is going to `releases/prod`, #7213 is the PR to main

zzmp (approved):

JFrankfurt (approved):

PR #7215, opened 2023-08-23
chore: adds e2e test for when usd quote fetch fails

JackShort:
* adds an e2e test for when fetching usd quote fails on swap

_Linear ticket:_https://linear.app/uniswap/issue/WEB-2097/e2e-test-for-inability-to-fetch-usd-quote-fiat-value-for-a-token-in


src/components/CurrencyInputPanel/SwapCurrencyInputPanel.tsx
```diff
@@ -312,7 +312,13 @@ export default function SwapCurrencyInputPanel({
           <FiatRow>
             <RowBetween>
               <LoadingOpacityContainer $loading={loading}>
-                {fiatValue && <FiatValue fiatValue={fiatValue} priceImpact={priceImpact} />}
+                {fiatValue && (
+                  <FiatValue
+                    fiatValue={fiatValue}
+                    priceImpact={priceImpact}
+                    testId={`fiat-value-${id === InterfaceSectionName.CURRENCY_INPUT_PANEL ? 'input' : 'output'}-panel`}
```
gnewfield, line 319: Is it necessary to name the input and output panels differently? An alternative might be to select both and index them:
```
// input panel
cy.get(getTestSelector('fiat-value-panel')).eq(0)
// output panel
cy.get(getTestSelector('fiat-value-panel')).eq(1)
```
I think it's good to limit the business logic in components as much as possible.

gnewfield: Additionally, would this allow us to remove `testId` as a prop and just statically apply it inside of `FiatValue`, instead of passing it down as a dynamic value from `SwapCurrencyInputPanel`?

gnewfield (changes requested):
Sorry meant to start a review on my other comment. Please see for details.

gnewfield (approved):

cypress/e2e/swap/errors.test.ts
```diff
@@ -116,4 +116,15 @@ describe('Swap errors', () => {
       getBalance(DAI).should('be.closeTo', initialBalance + 200, 1)
     })
   })
+
+  it('inability to fetch usd quote', () => {
```
zzmp, line 120:
```suggestion
  it('insufficient liquidity', () => {
```

src/components/CurrencyInputPanel/SwapCurrencyInputPanel.tsx
zzmp:
```suggestion
                    testId={`fiat-value-${id}`}
```

zzmp (approved):
lgtm - a few little nits to simplify things

tinaszheng (approved):

Issue #7083, opened 2023-08-04
Issues on USDT approvals on Ethereum mainnet (incorrect reset step)


Nanyan:
**Bug Description**
For an address which have successfully approved the USDT (on Ethereum miannet) to de `Permit2` contract( 0x000000000022D473030F116dDEE9F6B43aC78BA3) , and then it did not send a swap transaction immediately, for example, stop operation for several hours or close the browser, and then open the uniswap web site, and try to swap the USDT again, then  It requires a reseting of USDT approval , even though the allowance is large enough for the swap!

may be it should be a bug, since I think it's unnecessary to do a re-approval.

**Steps to Reproduce**

I'm not sure the follows steps will absolutely reproduce this issue.

1. For an address do an approval, like this : https://etherscan.io/tx/0xae44fad94ecfb9e50abe65eaa9a3976561655c37bce168d29f6a02de3c5058bd
2. Then close the dapp, or just wait several hours and then refresh the web,
3. Try to swap USDT for others tokens again, then it requires a reseting usdt approval.

**Expected Behavior**

It's no need to reset USDT approval when the allowance is enough.

**Additional Context**
![image](https://github.com/Uniswap/interface/assets/1623062/19b5c841-71aa-4aa1-bb9b-7a7a55aee178)

![image](https://github.com/Uniswap/interface/assets/1623062/15155e79-68d7-4791-93fc-f18924ceacc9)

kevinho:
The (SINOHOPE)[https://sinohope.com] team is developing a crypto wallet based on MPC-TSS technology with multi-signature feature. 
Since its launch 3 months ago, we have nearly 1B USD in AUM and over 100 corporate clients. In line with customer needs, we are one of the few that offer complete DApp browser functionality within our MPC wallet.

Uniswap is foundational to the decentralized world and is the DApp most favored by our users. During the implementation of the DApp browser, we have primarily used the Uniswap DApp for testing.

Due to the existence of multi-signature scenarios, users need to exit the DApp (**close webview**) to carry out the approval process. This behavior results in a life cycle for the DApp within our built-in browser that is very different from that in single-signature wallets (such as Metamask Mobile). It may be this specific scenario difference that has **increased the occurrence of the aforementioned bug.**

The Sinohope team is willing to collaborate with the Uniswap team, offering all kinds of conditions to assist in fixing this problem and helping Uniswap expand to a larger scale of customers.

just-toby:
[...]

PR #7219, opened 2023-08-23
fix: reset USDT logic

just-toby:
## Description

fixes a bug in the Resetting USDT logic - we only want to force the USDT approval if the setup approval step is needed in the first place. In other words, if the USDT setup approval is already large enough, we don't need to reset. 

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2620/investigate-permit2-bug-with-tether-reapproval-after-inactivity

issue #7083 

### Before

https://github.com/Uniswap/interface/assets/66155195/23450fad-5f45-4d7f-86c1-63db2f14048e

### After

https://github.com/Uniswap/interface/assets/66155195/00b77d57-71c6-4e8c-8a4a-5c9470166df5

### Reproducing the error

1. for USDT mainnet, set up your token approval to X
2. try to swap USDT to another token, with input amount <= X
3. notice that the reset step occurs even though you're swapping X or less USDT

### QA (ie manual testing)

- [x] repeat steps 1-2 above
- [x] ensure that the reset step is skipped

### Automated testing

- [x] Unit test n/a
- [x] Integration/E2E test - added

### DIFF
--- a/src/components/swap/ConfirmSwapModal.tsx
+++ b/src/components/swap/ConfirmSwapModal.tsx
@@ -97,6 +97,7 @@ function useConfirmModalState({
     // See the `approve` function here: https://etherscan.io/address/0xdAC17F958D2ee523a2206206994597C13D831ec7#code
     if (
       allowance.state === AllowanceState.REQUIRED &&
+      allowance.needsSetupApproval &&
       allowance.token.equals(USDT_MAINNET) &&
       allowance.allowedAmount.greaterThan(0)
     ) {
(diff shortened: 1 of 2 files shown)


src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -97,6 +97,7 @@ function useConfirmModalState({
     // See the `approve` function here: https://etherscan.io/address/0xdAC17F958D2ee523a2206206994597C13D831ec7#code
```
JackShort, line 97: maybe update the comment here

JackShort (approved):

src/components/swap/ConfirmSwapModal.tsx
just-toby: i think the comment is still accurate

natew (approved):

PR #7220, opened 2023-08-23
fix: zustand, migrate to createWithEqualityFn

just-toby:
## Description

using an equality function in the zustand accessor is now deprecated. the migration is to replace `create` with `createWithEqualityFn` and pass the same `shallow` comparator we were using before.

_Relevant docs:_ https://github.com/pmndrs/zustand/discussions/1937

## Screen capture

warnings in the console

### After

no warnings in the console

## Test plan

let e2e tests run


src/pages/Swap/index.tsx
```diff
@@ -405,6 +405,7 @@ export function Swap({
   )
 
   const handleContinueToReview = useCallback(() => {
+    console.log('handleContinueToReview', trade)
```
JackShort, line 408: i don't think this should be in here?

just-toby: what the heck!

JackShort (approved):

src/nft/components/profile/list/Modal/ListModal.tsx
```diff
@@ -63,8 +62,7 @@ export const ListModal = ({ overlayClick }: { overlayClick: () => void }) => {
         getLooksRareNonce,
         collectionsRequiringApproval,
         listings,
-      }),
-      shallow
```
cbachmeier, line 67: I'm getting a bug with the listing experience where when I have a collection that needs to be approved, the tx auto completes and goes back to the profile page. This isn't happening on main or staging
this branch:
listForSale
staging:
listForSaleStaging

cbachmeier (changes requested):

cbachmeier (approved):

PR #7226, opened 2023-08-24
fix: smart contract wallets unsupported

JackShort:
* fixes a bug that causes smart contract wallets to fail

linear: https://linear.app/uniswap/issue/WEB-2613/smart-contract-wallets-are-unsupported-investigative
slack: https://uniswapteam.slack.com/archives/C047U65H422/p1692894402092459

* smart contracts alter the calldata so checking if they are the same is not correct - better to check if it exists

### DIFF
--- a/src/hooks/useUniversalRouter.ts
+++ b/src/hooks/useUniversalRouter.ts
@@ -117,7 +117,7 @@ export function useUniversalRouterSwapCallback(
               }),
               ...analyticsContext,
             })
-            if (tx.data !== response.data) {
+            if (!response.data || response.data.length === 0 || response.data === '0x') {
               sendAnalyticsEvent(SwapEventName.SWAP_MODIFIED_IN_WALLET, {
                 txHash: response.hash,
                 ...analyticsContext,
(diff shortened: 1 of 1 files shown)


tinaszheng (approved):

src/hooks/useUniversalRouter.ts
```diff
@@ -117,7 +117,7 @@ export function useUniversalRouterSwapCallback(
               }),
               ...analyticsContext,
             })
-            if (tx.data !== response.data) {
+            if (!response.data || response.data.length === 0 || response.data === '0x') {
```
just-toby, line 120: we should still log SWAP_MODIFIED_IN_WALLET based on the original criteria. if you don't want that to error, it's fine. just throw the error if the new check fails

just-toby (changes requested):
see comment

src/hooks/useUniversalRouter.ts
JackShort: cool cool

PR #7234, opened 2023-08-25
fix: remove unused feature flags

just-toby:
## Description

remove dead/unused feature flags

## test plan
automated tests and eyeballs


cypress/e2e/swap/settings.test.ts
```diff
@@ -24,6 +24,6 @@ describe('Swap settings', () => {
     cy.contains('Transaction deadline').should('exist')
     cy.contains('UniswapX').should('exist')
     cy.contains('Local routing').should('exist')
-    cy.get(getTestSelector('mobile-settings-scrim')).click()
+    cy.get(getTestSelector('mobile-settings-scrim')).click({ force: true })
```
cbachmeier, line 27: qq: why was this change needed?

src/state/routing/slice.ts
```diff
@@ -146,8 +138,7 @@ export const routingApi = createApi({
               tokenOut: tokenOutAddress,
               amount,
               type,
-              // if forceUniswapXOn is not ON, then use the backend's default value
-              useUniswapX: forceUniswapXOn || undefined,
+              useUniswapX: true,
```
cbachmeier, line 141: I think this should be undefined here based on the previous comment? `if forceUniswapXOn is not ON, then use the backend's default value` tagging @tinaszheng

cypress/e2e/swap/settings.test.ts
just-toby: the test was failing for me without this :/ i don't think it's related to my changes though

src/state/routing/slice.ts
just-toby: yeah i agree, we should just remove this

cbachmeier (approved):

PR #7237, opened 2023-08-28
feat: disable exact_output FOT swaps

cartcrom:
## Description

Prevents exact_output swaps when the output token has a fee on transfer, since we cannot currently guarantee exact output amounts for these types of swaps.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2762/fot-exact-output-tax-ui-fix

## Screen capture

|    Before    |   After (clicking output selects input)    |
| image | image |

### QA (ie manual testing)

- [ ] Create an exact output swap with two non-FOT tokens, switch to an FOT output token (e.g. BULLET), ensure the trade becomes exact_input 
- [ ] Create an exact input swap from an FOT token to a non-FOT token, press the switch currencies button, ensure that trade becomes exact_input
- [ ] Try to type an output amount when the output token is an FOT token, see tooltip and animation preventing input


src/components/NumericalInput/index.tsx
```diff
@@ -1,9 +1,10 @@
-import React from 'react'
+import React, { useEffect, useRef } from 'react'
 import styled from 'styled-components'
 import { escapeRegExp } from 'utils'
 
-const StyledInput = styled.input<{ error?: boolean; fontSize?: string; align?: string }>`
+const StyledInput = styled.input<{ error?: boolean; fontSize?: string; align?: string; disabled?: boolean }>`
   color: ${({ error, theme }) => (error ? theme.critical : theme.neutral1)};
+  pointer-events: ${({ disabled }) => (disabled ? 'none' : 'auto')};
```
cartcrom, line 7: I set pointer-events to none when the input is disabled so that clicking the input will allow the click event to bubble up to the parent component when the input is disabled.

src/components/CurrencyInputPanel/SwapCurrencyInputPanel.tsx
```diff
@@ -241,8 +278,21 @@ export default function SwapCurrencyInputPanel({
...
+    if (numericalInputSettings?.disabled) {
+      setAnimateShake(true)
+      setTimeout(() => setAnimateShake(false), ms('4s')) // reset shake animation state after 4s
+      numericalInputSettings.onDisabledClick()
+    }
+  }, [numericalInputSettings])
+
   const chainAllowed = isSupportedChain(chainId)
 
+  const [animateShake, setAnimateShake] = useState(false)
```
just-toby, line 291: can you rename this to something like `tooltipVisible` ? naming it after the animation made it harder to understand why we need this in state, as a reader

src/pages/Swap/TaxTooltipBody.tsx
```diff
@@ -0,0 +1,37 @@
...
+          </Trans>
+        ) : (
+          <Trans>
+            Fees on the selected output token don&apos;t allow for accurate exact outputs. Use the `You pay` field
+            instead.
+          </Trans>
+        )}
+        <Trans>
+          {currencySymbol} fees don&apos;t allow for accurate exact outputs. Use the `You pay` field instead.
+        </Trans>
```
just-toby, line 33: remove this?

src/pages/Swap/index.tsx
```diff
@@ -650,6 +665,14 @@ export function Swap({
                 showCommonBases
                 id={InterfaceSectionName.CURRENCY_OUTPUT_PANEL}
                 loading={independentField === Field.INPUT && routeIsSyncing}
+                numericalInputSettings={{
+                  // We disable numerical input here if the selected token has tax, since we cannot guarantee exact_outputs for FOT tokens
+                  disabled: outputTokenHasTax,
+                  // Focus the input currency panel if the user tries to type into the disabled output currency panel
+                  onDisabledClick: () =>
+                    window.dispatchEvent(new Event(`forceInputFocus:${InterfaceSectionName.CURRENCY_INPUT_PANEL}`)),
```
just-toby, line 673: can we just remove this behavior? the input is disabled, and it appears disabled, so i don't think we should be hacking in this reaction. because the Event-sending across different components does feel hacky and easy to break later

just-toby (changes requested):

src/pages/Swap/index.tsx
cartcrom: thx was hoping to get feedback on this pattern, zach said the same, I'm gonna use refs instead.

src/pages/Swap/TaxTooltipBody.tsx
cartcrom: whoops

marktoda (approved):

gnewfield (approved):

src/components/CurrencyInputPanel/SwapCurrencyInputPanel.tsx
```diff
@@ -206,162 +237,204 @@ interface SwapCurrencyInputPanelProps {
...
+              </ThemedText.BodySecondary>
+            </AutoColumn>
+          </FixedContainer>
+        )}
+
+        <Container hideInput={hideInput}>
+          <ThemedText.SubHeaderSmall style={{ userSelect: 'none' }}>{label}</ThemedText.SubHeaderSmall>
+          <InputRow style={hideInput ? { padding: '0', borderRadius: '8px' } : {}}>
+            {!hideInput && (
+              <div style={{ display: 'flex', flexGrow: '1' }} onClick={handleDisabledNumericalInputClick}>
```
cbachmeier, line 316: nit: styled-component

src/pages/Swap/TaxTooltipBody.tsx
```diff
@@ -0,0 +1,34 @@
...
+  border-width: 0;
+  margin: 12px 0;
+  background-color: ${({ theme }) => theme.surface3};
+`
+
+export function OutputTaxTooltipBody({ currencySymbol }: { currencySymbol?: string }) {
+  return (
+    <>
+      <ThemedText.SubHeaderSmall color="textPrimary">
+        <Trans>Exact input only</Trans>
```
cbachmeier, line 17: Is "Exact input" used colloquially? I worry this is only used internally and outside users may get confused. Maybe "Input currency only"  or "Input amount only"

src/pages/Swap/index.tsx
```diff
@@ -279,6 +281,19 @@ export function Swap({
     inputError: swapInputError,
   } = swapInfo
 
+  const [inputTokenHasTax, outputTokenHasTax] = useMemo(
+    () => [
+      !!currencies[Field.INPUT] && !getInputTax(currencies[Field.INPUT]).equalTo(0),
+      !!currencies[Field.OUTPUT] && !getOutputTax(currencies[Field.OUTPUT]).equalTo(0),
```
cbachmeier, line 287: nit: I think a test here would be useful

cbachmeier (approved):
LGTM minor nits and comments

src/pages/Swap/TaxTooltipBody.tsx
cartcrom: Not sure, but this came from cal & fred 🤷

src/pages/Swap/TaxTooltipBody.tsx
```diff
@@ -0,0 +1,34 @@
...
+        {currencySymbol ? (
+          <Trans>
+            {currencySymbol} fees don&apos;t allow for accurate exact outputs. Use the `You pay` field instead.
+          </Trans>
+        ) : (
+          <Trans>
+            Fees on the selected output token don&apos;t allow for accurate exact outputs. Use the `You pay` field
+            instead.
+          </Trans>
+        )}
```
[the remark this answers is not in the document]
cartcrom: I think i.e. 'BULLET fees' reads better than 'Fees on Bullet' hence the separation

PR #7238, opened 2023-08-28
fix: remove base goerli from chain selector

just-toby:
## Description

remove base testnet until routing supports it

_Slack thread:_ https://uniswapteam.slack.com/archives/C03ESADBQPK/p1693248493220419

## Test plan

verify that base goerli no longer appears in the chain selector with testnets enabled (in settings)


zzmp (changes requested):
Why not remove _all_ references to `BASE_GOERLI`? Are we definitely adding back support to it in the future? Wouldn't it still just be added back as a revert of this PR?

just-toby:
> Why not remove _all_ references to `BASE_GOERLI`? Are we definitely adding back support to it in the future? Wouldn't it still just be added back as a revert of this PR?

that's fair, we can remove them all and still just revert this PR when the time is right.

zzmp (approved):

gnewfield (approved):

PR #7239, opened 2023-08-28
chore: updating numberFormat for other currencies and languages

JackShort:
* Updates the number formatting function to accept and format with a locale and/or with a new currency
main changes are in `src/utils/formatNumbers.ts`
* the locale and language is currently not being used yet because all of the other functions still need be transferred over but that is going to be done in separate prs (as you can tell how big this one is)
* this also adds a test for another language and currency for every existing formatNumber based test


src/utils/formatNumbers.ts
```diff
@@ -300,40 +358,65 @@ const TYPE_TO_FORMATTER_RULES = {
...
+  const { hardCodedInput, formatterOptions } = getFormatterRule(input, type)
+
+  if (formatterOptions.currency) {
+    formatterOptions.currency = localCurrency
+    formatterOptions.currencyDisplay = LOCAL_CURRENCY_SYMBOL_DISPLAY_TYPE[localCurrency]
+  }
+
+  if (!hardCodedInput) {
+    return new Intl.NumberFormat(locale, formatterOptions).format(input)
+  }
```
just-toby, line 403: are you introducing the concept of "hardcoded" input/output values in this PR? can you add a comment somewhere explaining why it's needed?

just-toby (approved):

src/utils/formatNumbers.ts
JackShort: i am not really introducing it - it kinda already existed already (I more dissected it out of the current functionality) but I actually explained it above the type definition of it

src/utils/formatNumbers.ts
```diff
@@ -1,244 +1,302 @@
...
 ]
 
 const fiatTokenPricesFormatter: FormatterRule[] = [
-  { exact: 0, formatter: '$0.00' },
-  { upperBound: 0.00000001, formatter: '<$0.00000001' },
-  { upperBound: 1, formatter: THREE_SIG_FIGS_USD },
-  { upperBound: 1e6, formatter: TWO_DECIMALS_USD },
-  { upperBound: 1e16, formatter: SHORTHAND_USD_TWO_DECIMALS },
-  { upperBound: Infinity, formatter: SEVEN_SIG_FIGS__SCI_NOTATION_USD },
+  { exact: 0, hardCodedInput: { input: 0 }, formatterOptions: TWO_DECIMALS_CURRENCY },
```
cbachmeier, line 234: for the formatters on `exact: 0`, do you need the hardcodedInput of `0` as well?

cbachmeier (approved):
LGTM had a couple Qs for my understanding

src/constants/localCurrencies.tsx
```diff
@@ -46,6 +46,27 @@ export type SupportedLocalCurrency = (typeof SUPPORTED_LOCAL_CURRENCIES)[number]
 
 export const DEFAULT_LOCAL_CURRENCY: SupportedLocalCurrency = 'USD'
 
+export const LOCAL_CURRENCY_SYMBOL_DISPLAY_TYPE: Record<SupportedLocalCurrency, 'narrowSymbol' | 'symbol'> = {
```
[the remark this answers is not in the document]
JackShort: ah yeah so some currencies need to be forced to use the narrow symbol when they are displayed and others need to use the normal symbol. so for example if CAD uses the narrow symbol its `$` but when set to `symbol` its `CA$` - so i have to force it to use `symbol`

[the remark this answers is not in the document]
JackShort: others like `THB` need to be set to `narrowSymbol` or they are displayed as `THB` and not their actual currency symbol - which is super annoying

src/utils/formatNumbers.ts
```diff
@@ -1,244 +1,302 @@
...
   minimumFractionDigits: 2,
-})
+}
 
-const TWO_DECIMALS_USD = new Intl.NumberFormat('en-US', {
+const TWO_DECIMALS_CURRENCY: NumberFormatOptions = {
   notation: 'standard',
   maximumFractionDigits: 2,
   minimumFractionDigits: 2,
   currency: 'USD',
```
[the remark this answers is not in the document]
JackShort: it it gets overridden eventually - however i use currency as being set as to tell if it should set its currency. So like if currency is set to 'USD' i know to update the currency however if no currency is set obviously the currency shouldn't be updated

src/utils/formatNumbers.ts
JackShort: lol good point no it doesn't

cbachmeier (approved):

PR #7245, opened 2023-08-30
feat: Add intent to swap quote request

tinaszheng:
## Description
This tags our price quote requests with `intent: 'pricing'` for backend to have better metrics

### QA (ie manual testing)
I looked at the quote requests in the Network tab to make sure the pricing requests had the `intent` tag and regular quote requests dont


cartcrom (approved):

zzmp (approved):
lgtm, but why don't non-pricing quotes have their own distinct intents, so that everything is categorized?

tinaszheng:
@zzmp so true bestie

PR #7250, opened 2023-08-30
feat: [info] Initial Pool Details Page

cbachmeier:
## Description

- Add initial pool details page file
- Add pool data gql query
  - dynamically use ApolloClient/subgraph url based on pool's chain
- Build header for Pool Details Page
- Added tests

_Relevant docs:_ https://www.figma.com/file/2yKuZXmQ2UvgBbXR3txz4O/Dot-Info?type=design&node-id=474%3A1165649&mode=design&t=5TY9uOA19hFqLTHW-1

## Screen capture
`#/pools/ethereum/0x88e6a0c2ddd26feeb64f039a2c41296fcb3f5640`
Screenshot 2023-08-30 at 4 26 09 PM
`#/pools/polygon/0x167384319b41f7094e62f7506409eb38079abff8`
Screenshot 2023-08-30 at 4 27 36 PM

### Automated testing

- [x] Added tests for util function
- [x] Added general PDP tests
- [x] Added PDP Header test


src/graphql/thegraph/PoolData.ts
```diff
@@ -0,0 +1,64 @@
...
+      txCount
+      totalValueLockedToken0
+      totalValueLockedToken1
+      totalValueLockedUSD
+    }
+    bundles(where: { id: "1" }) {
+      ethPriceUSD
+    }
+  }
+`
```
JFrankfurt, line 45: copy/pasta from info. ?

src/graphql/thegraph/PoolData.ts
```diff
@@ -0,0 +1,64 @@
...
+export function usePoolData(poolAddress: string, chainId?: ChainId) {
+  const poolId = [poolAddress]
+  const apolloClient = useMemo(
+    () =>
+      new ApolloClient({
+        cache: new InMemoryCache(),
+        uri: CHAIN_SUBGRAPH_URL[chainId ?? ChainId.MAINNET],
+      }),
+    [chainId]
+  )
```
JFrankfurt, line 56: should the client be instantiated elsewhere and just consumed inside the hook?

src/graphql/thegraph/PoolData.ts
```diff
@@ -0,0 +1,64 @@
...
+  const { data, loading } = usePoolDataQuery({ variables: { poolId }, client: apolloClient })
+  return useMemo(() => {
+    return {
+      data: data?.data[0],
```
JFrankfurt, line 60: this feels a little jank

src/graphql/thegraph/apollo.ts
```diff
@@ -3,7 +3,7 @@ import { ChainId } from '@uniswap/sdk-core'
 
 import store from '../../state/index'
 
-const CHAIN_SUBGRAPH_URL: Record<number, string> = {
+export const CHAIN_SUBGRAPH_URL: Record<number, string> = {
```
JFrankfurt, line 6: maybe this is the right file for apollo client instantiation?

src/pages/App.tsx
```diff
@@ -217,6 +220,7 @@ export default function App() {
                     <Route path=":chainName" />
                   </Route>
                   <Route path="tokens/:chainName/:tokenAddress" element={<TokenDetails />} />
+                  {infoPoolPageEnabled && <Route path="pools/:chainName/:poolAddress" element={<PoolDetails />} />}
```
JFrankfurt, line 223: clean 👌

src/pages/PoolDetails/PoolDetailsHeader.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+    toggleReversed: jest.fn(),
+  }
+
+  it('renders header text correctly', () => {
+    render(<PoolDetailsHeader {...mockProps} />)
+
+    expect(screen.getByText(/Explore/i)).toBeInTheDocument()
+    expect(screen.getByText(/Pool/i)).toBeInTheDocument()
+    expect(screen.getByText(/USDC \/ WETH \(0x88e6...5640\)/i)).toBeInTheDocument()
+    expect(screen.getByText('0.05%')).toBeInTheDocument()
```
JFrankfurt, line 22: you might consider a snapshot test here instead

```
  expect(asFragment()).toMatchSnapshot()
```

src/pages/PoolDetails/PoolDetailsHeader.tsx
```diff
@@ -0,0 +1,209 @@
...
+}
+
+const DoubleLogoContainer = styled.div`
+  display: flex;
+  flex-direction: row;
+  gap: 2px;
+  position: relative;
+  top: 0;
+  left: 0;
+  img:nth-child(n) {
```
JFrankfurt, line 174:
```suggestion
  img {
```

src/pages/PoolDetails/PoolDetailsHeader.tsx
```diff
@@ -0,0 +1,209 @@
...
+    width: 16px;
+    height: 32px;
+    object-fit: cover;
+  }
+  img:nth-child(1) {
```
JFrankfurt, line 179: first-child or first-of-type

src/pages/PoolDetails/PoolDetailsHeader.tsx
```diff
@@ -0,0 +1,209 @@
...
+    border-radius: 16px 0 0 16px;
+    object-position: 0 0;
+  }
+  img:nth-child(2) {
```
JFrankfurt, line 183: last-child or last-of-type

src/pages/PoolDetails/PoolDetailsHeader.tsx
```diff
@@ -0,0 +1,209 @@
...
+  return <DoubleLogo logo1={src} onError1={nextSrc} logo2={src2} onError2={nextSrc2} />
+}
+
+const DoubleLogoContainer = styled.div`
+  display: flex;
+  flex-direction: row;
+  gap: 2px;
+  position: relative;
+  top: 0;
+  left: 0;
```
JFrankfurt, line 173: are these doing anything w/ the relative position?

src/pages/PoolDetails/PoolDetailsHeader.tsx
```diff
@@ -0,0 +1,209 @@
...
+function DoubleCurrencyLogo({ chainId, currencies }: { chainId: number; currencies: Array<Currency | undefined> }) {
+  const [src, nextSrc] = useTokenLogoSource(currencies?.[0]?.wrapped.address, chainId, currencies?.[0]?.isNative)
+  const [src2, nextSrc2] = useTokenLogoSource(currencies?.[1]?.wrapped.address, chainId, currencies?.[1]?.isNative)
+
+  return <DoubleLogo logo1={src} onError1={nextSrc} logo2={src2} onError2={nextSrc2} />
+}
+
+const DoubleLogoContainer = styled.div`
+  display: flex;
+  flex-direction: row;
```
JFrankfurt, line 169:
```suggestion

```

JFrankfurt: nit: row is the default (not important to remove--sometimes I like being explicit about this)

src/pages/PoolDetails/index.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+    volumeUSD: '394920157156.0515346899898790592366',
+    volumeToken0: '394894081779.781168',
+    volumeToken1: '190965971.266407832255075308',
+    txCount: '5406827',
+    totalValueLockedToken0: '180078648.881221',
+    totalValueLockedToken1: '142782.017882048454494774',
+    totalValueLockedUSD: '417233634.1468435997761171520463339',
+  },
+  loading: false,
+}
```
JFrankfurt, line 57: consider starting a mock data directory, since this is a pattern we'll probably have to replicate

src/pages/PoolDetails/index.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+  })
+
+  it('pool header is displayed when data is received from thegraph', () => {
+    render(<PoolDetails />)
+
+    waitFor(() => {
+      expect(screen.getByText(/Explore/i)).toBeInTheDocument()
+      expect(screen.getByText(/Pool/i)).toBeInTheDocument()
+      expect(screen.getByText(/USDC \/ WETH \(0x88e6...5640\)/i)).toBeInTheDocument()
+    })
```
JFrankfurt, line 132: consider a snapshot test on `asFragment()`

src/pages/PoolDetails/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+  }>()
+  const chain = getValidUrlChainName(chainName)
+  const chainId = chain && supportedChainIdFromGQLChain(chain)
+  const { data: poolData, loading } = usePoolData(poolAddress ?? '', chainId)
+  const [isReversed, toggleReversed] = useReducer((x) => !x, false)
+  const token0 = isReversed ? poolData?.token1 : poolData?.token0
+  const token1 = isReversed ? poolData?.token0 : poolData?.token1
+  const isInvalidPool = !chainName || !poolAddress || !getValidUrlChainName(chainName) || !isAddress(poolAddress)
+
+  // TODO: Add skeleton once designed
```
JFrankfurt, line 30: make a linear task for this todo and reference it here

src/pages/PoolDetails/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+  if (loading) return <></>
```
JFrankfurt, line 31:
```suggestion
  if (loading) return null
```

src/pages/PoolDetails/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+  if ((!loading && !poolData) || isInvalidPool) return <NotFound />
```
JFrankfurt, line 32: consider a more descriptive single boolean value

src/graphql/thegraph/PoolData.ts
cbachmeier: Basically, I just haven't integrated using blocks for different time windows yet https://github.com/Uniswap/v3-info/blob/master/src/data/pools/poolData.ts#L18

src/graphql/thegraph/PoolData.ts
cbachmeier: Data for each chain comes from its own subgraph which requires its own apollo client. We only know the client to use at pageload as users can query pools across chains separate from their connected chain so I placed it here at the query level. We could in theory instantiate it on the page and then pass the client to the query?

src/graphql/thegraph/PoolData.ts
cbachmeier: because of the double data or taking the first elem? This does handle invalid poolIds by returning `undefined` for data. ex: https://interface-git-cab-initpdp-uniswap.vercel.app/#/pools/ethereum/invalid

src/graphql/thegraph/apollo.ts
cbachmeier: see my comment on needing dynamic client based on chain to search for https://github.com/Uniswap/interface/pull/7250#discussion_r1318936410

src/pages/PoolDetails/PoolDetailsHeader.test.tsx
cbachmeier: I generally querying for specific elements as opposed to a more broad snapshot match as it's much easier to glance over a breaking change when updating snapshots vs having to make sure the output looks as you intended it.

src/pages/PoolDetails/index.test.tsx
cbachmeier: see comment on general testing preferences https://github.com/Uniswap/interface/pull/7250/files#r1318937970

src/graphql/data/util.test.tsx
```diff
@@ -34,3 +34,25 @@ describe('fromGraphQLChain', () => {
...
+
+  it('should return false for invalid chain name', () => {
+    const invalidChainName = 'invalidchain'
+    expect(getValidUrlChainName(invalidChainName)).toBe(undefined)
+  })
+
+  it('should return false for a misconfigured chain name', () => {
+    const invalidChainName = 'eThErEuM'
+    expect(getValidUrlChainName(invalidChainName)).toBe(undefined)
+  })
```
JFrankfurt, line 57: maybe we should not be case sensitive?

src/graphql/data/util.tsx
```diff
@@ -132,6 +132,15 @@ const URL_CHAIN_PARAM_TO_BACKEND: { [key: string]: InterfaceGqlChain } = {
...
 
+/**
+ * @param chainName parsed in chain name from url query parameter
+ * @returns if chainName is a valid chain name, returns the backend chain name, otherwise returns undefined
+ */
+export function getValidUrlChainName(chainName: string | undefined): Chain | undefined {
+  const validChainName = chainName && URL_CHAIN_PARAM_TO_BACKEND[chainName]
+  return validChainName ? validChainName : undefined
+}
+
```
JFrankfurt, line 143: this seems p similar to the one below `validateUrlChainParam`. Could we reuse the existing code?

src/graphql/thegraph/PoolData.ts
JFrankfurt: I was imagining we instantiate all the apollo clients on page load, then select the appropriate one in the hook when we need to fetch data.

src/graphql/thegraph/PoolData.ts
JFrankfurt: `data: data?.data[0]` 

both reasons I guess. When I try to verbalize this I get something like: "data is data's data array's first element, except if data is undefined, in which case we return undefined data". I think it would be better to assign the query data to a descriptive var name, then handle the undefined case separately before returning the result in the useMemo.

src/graphql/thegraph/PoolData.ts
JFrankfurt: My intuition is that client instantiation is somewhat expensive and we wouldn't want to create/dispose for each data request

src/pages/PoolDetails/PoolDetailsHeader.test.tsx
JFrankfurt: I get that, but I think text presence is much less likely to break here relative to styling. Snapshots get you the style safety *and* the text safety.

JFrankfurt (approved):

src/graphql/data/util.tsx
cbachmeier: I was originally using the function below, however this has the important distinction that it supports url chain params that are supported by subgraph but not necessarily supported by the GQL BE, whereas the fn below requires it is supported by the GQL BE for Explore and TDP

src/graphql/data/util.test.tsx
cbachmeier: Valid point, might be worth discussing with product. Right now for TDP we're pretty forgiving, like this will resolve which I think is a little too far https://app.corn-staging.com/tokens/eTh%20Er%20%20%20%20____%20%20%20%20EuM/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48

src/pages/PoolDetails/PoolDetailsHeader.test.tsx
cbachmeier: I've also added snapshot tests to cover both scenarios

gnewfield (approved):

src/pages/PoolDetails/PoolDetailsHeader.tsx
```diff
@@ -0,0 +1,208 @@
...
+        <ThemedText.BodySecondary>&nbsp;{'>'}&nbsp;</ThemedText.BodySecondary>
+        {/* TODO: When Explore Pool table is added, link directly back to it */}
+        <StyledLink to={origin}>
+          <ThemedText.BodySecondary>
+            <Trans>Pool</Trans>
+          </ThemedText.BodySecondary>
+        </StyledLink>
+        <ThemedText.BodySecondary>&nbsp;{'>'}&nbsp;</ThemedText.BodySecondary>
+        <ThemedText.BodyPrimary>
+          {token0?.symbol} / {token1?.symbol} ({shortenAddress(poolAddress)})
```
natew, line 79: does this have a failure case where a token is undefined? would look odd? jw

src/pages/PoolDetails/PoolDetailsHeader.tsx
```diff
@@ -0,0 +1,208 @@
...
+      <Row gap="18px">
+        <Row gap="8px" width="max-content">
+          {chainId && (
+            <DoubleCurrencyAndChainLogo data-testid="double-token-logo" chainId={chainId} currencies={currencies} />
+          )}
+          <ThemedText.HeadlineSmall>
+            {token0?.symbol} / {token1?.symbol}
+          </ThemedText.HeadlineSmall>
+        </Row>
+        {feeTier && <FeeTier>{feeTier / 10000}%</FeeTier>}
```
natew, line 91:
```suggestion
        {Boolean(feeTier) && <FeeTier>{feeTier / 10000}%</FeeTier>}
```

avoids showing 0 if its 0

src/pages/PoolDetails/icons.tsx
```diff
@@ -0,0 +1,15 @@
+type SVGProps = React.SVGProps<SVGSVGElement> & {
+  fill?: string
```
natew, line 2: isn't fill on svgprops already?

natew: or rather is fill necessary? same with gradientId never seen that

src/pages/PoolDetails/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export default function PoolDetailsPage() {
+  const { poolAddress, chainName } = useParams<{
+    poolAddress: string
+    chainName: string
+  }>()
+  const chain = getValidUrlChainName(chainName)
+  const chainId = chain && supportedChainIdFromGQLChain(chain)
+  const { data: poolData, loading } = usePoolData(poolAddress ?? '', chainId)
+  const [isReversed, toggleReversed] = useReducer((x) => !x, false)
```
natew, line 25: nifty, i never use reducer but for simple toggles makes a lot of sense

natew (approved):

src/pages/PoolDetails/icons.tsx
cbachmeier: that comes from eslint not properly understanding SVGProps https://github.com/Uniswap/interface/blob/main/src/nft/components/icons.tsx#L6C13-L6C13

src/pages/PoolDetails/PoolDetailsHeader.tsx
cbachmeier: For a token to be undefined would mean there is an upstream issue with thegraph where a valid pool address was given and the gql query did not return an error but the token data was undefined.

src/graphql/thegraph/apollo.ts
```diff
@@ -32,3 +32,34 @@ export const apolloClient = new ApolloClient({
   cache: new InMemoryCache(),
   link: concat(authMiddleware, httpLink),
 })
+
+export const chainToApolloClient: Record<number, ApolloClient<NormalizedCacheObject>> = {
```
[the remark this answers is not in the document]
cbachmeier: that should be a separate PR since that would affect live UX outside of the hidden PDP content

natew (approved):

PR #7251, opened 2023-08-31
feat: adding currency conversion hook to useUSDPrice

JackShort:
* adding conversion rate hook where we can pass local currencies into it and get the conversion rate relative to USD
* this also updates the current `useUSDPrice` hooks to use this hook instead since it just wraps it
* this should have no impact (like no extra overhead) on users that are using USD (all our users since the currency selection is behind a feature flag)

### DIFF
--- a/src/graphql/data/ConversionRate.ts
+++ b/src/graphql/data/ConversionRate.ts
@@ -1,0 +1,11 @@
+import gql from 'graphql-tag'
+
+gql`
+  query Convert {
+    convert(fromAmount: { currency: USD, value: 1.0 }, toCurrency: JPY) {
+      id
+      value
+      currency
+    }
+  }
+`
--- a/src/graphql/data/ConversionRate.ts
+++ b/src/graphql/data/ConversionRate.ts
@@ -1,4 +1,4 @@
-import { DEFAULT_LOCAL_CURRENCY, SupportedLocalCurrency } from 'constants/localCurrencies'
+import { SupportedLocalCurrency } from 'constants/localCurrencies'
 import gql from 'graphql-tag'
 import ms from 'ms'
 import { getFetchPolicyForKey } from 'utils/getFetchPolicyForKey'
@@ -15,21 +15,12 @@ gql`
   }
 `
 
-export function useLocalCurrencyConversionRate(localCurrency: SupportedLocalCurrency) {
-  const isDefaultCurrency = localCurrency === DEFAULT_LOCAL_CURRENCY
-
+export function useLocalCurrencyConversionRate(localCurrency: SupportedLocalCurrency, skip?: boolean) {
   const { data, loading } = useConvertQuery({
     variables: { toCurrency: localCurrency },
     fetchPolicy: getFetchPolicyForKey(`convert-${localCurrency}`, ms('5m')),
-    skip: isDefaultCurrency,
+    skip,
   })
-
-  if (isDefaultCurrency) {
-    return {
-      data: 1.0,
-      isLoading: false,
-    }
-  }
 
   return {
     data: data?.convert?.value,
--- a/src/graphql/data/ConversionRate.ts
+++ b/src/graphql/data/ConversionRate.ts
@@ -1,11 +1,25 @@
+import { SupportedLocalCurrency } from 'constants/localCurrencies'
 import gql from 'graphql-tag'
+import ms from 'ms'
+import { getFetchPolicyForKey } from 'utils/getFetchPolicyForKey'
+
+import { useConvertQuery } from './__generated__/types-and-hooks'
 
 gql`
-  query Convert {
-    convert(fromAmount: { currency: USD, value: 1.0 }, toCurrency: JPY) {
+  query Convert($toCurrency: Currency!) {
+    convert(fromAmount: { currency: USD, value: 1.0 }, toCurrency: $toCurrency) {
       id
       value
       currency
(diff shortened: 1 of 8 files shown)


src/utils/getFetchPolicyForKey.ts
```diff
@@ -0,0 +1,16 @@
+import { WatchQueryFetchPolicy } from '@apollo/client'
+
+const keys = new Map<string, number>()
+
```
JackShort, line 4: this implements a time based cache policy. if a key doesn't exist it stores it at the timestamp and the next time it queries it it checks if its within the expiration time. if it is it sets the fetch-policy to use the cache if its outside it sets it to use the network

src/hooks/useLocalCurrencyPrice.ts
```diff
@@ -0,0 +1,27 @@
...
+  const activeLocalCurrency = useActiveLocalCurrency()
+  const activeLocalCurrencyIsUSD = activeLocalCurrency === DEFAULT_LOCAL_CURRENCY
+
+  const { data: usdPrice, isLoading: isUSDPriceLoading } = useUSDPrice(...useUSDPriceParameters)
+  const { data: localCurrencyConversionRate, isLoading: isLocalCurrencyConversionRateLoading } =
+    useLocalCurrencyConversionRate(activeLocalCurrency, activeLocalCurrencyIsUSD)
+
+  if (activeLocalCurrencyIsUSD) {
+    return { data: usdPrice, isLoading: isUSDPriceLoading }
+  }
```
JackShort, line 19: if we are currently in usd no need to fetch conversion rates so we just return what we got from the `useUSDPrice` hook

cbachmeier (approved):
LGTM, minor nits and Qs

src/hooks/useLocalCurrencyPrice.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export function useLocalCurrencyPrice(...useUSDPriceParameters: useUSDPriceParameters) {
+  const activeLocalCurrency = useActiveLocalCurrency()
+  const activeLocalCurrencyIsUSD = activeLocalCurrency === DEFAULT_LOCAL_CURRENCY
+
+  const { data: usdPrice, isLoading: isUSDPriceLoading } = useUSDPrice(...useUSDPriceParameters)
+  const { data: localCurrencyConversionRate, isLoading: isLocalCurrencyConversionRateLoading } =
+    useLocalCurrencyConversionRate(activeLocalCurrency, activeLocalCurrencyIsUSD)
+
+  if (activeLocalCurrencyIsUSD) {
```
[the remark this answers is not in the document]
JackShort: considering we are just using primitives and not destructuring objects I don't think memoization is necessary - every value being compared or returned is a number or a boolean

src/hooks/useLocalCurrencyPrice.ts
```diff
@@ -0,0 +1,27 @@
...
+    return { data: usdPrice, isLoading: isUSDPriceLoading }
+  }
+
+  const isLoading = isUSDPriceLoading || isLocalCurrencyConversionRateLoading
+  if (!usdPrice || !localCurrencyConversionRate) {
```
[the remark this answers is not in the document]
JackShort: yeah thats a good point - eventually something should be implemented when they change localCurrencies to check that the api is up i guess

src/graphql/data/ConversionRate.ts
```diff
@@ -0,0 +1,29 @@
...
+      value
+      currency
+    }
+  }
+`
+
+export function useLocalCurrencyConversionRate(localCurrency: SupportedLocalCurrency, skip?: boolean) {
+  const { data, loading } = useConvertQuery({
+    variables: { toCurrency: localCurrency },
+    fetchPolicy: getFetchPolicyForKey(`convert-${localCurrency}`, ms('5m')),
```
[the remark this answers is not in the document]
JackShort: it was decided in the design doc: https://www.notion.so/uniswaplabs/Currency-Conversion-4b0c301fd52841e1a7b573c8a0a8293f

since the conversion rate isn't that crucial for trading information we decided that 5 minutes is good enough

src/hooks/useLocalCurrencyPrice.ts
```diff
@@ -0,0 +1,27 @@
...
+
+type useUSDPriceParameters = Parameters<typeof useUSDPrice>
+
+export function useLocalCurrencyPrice(...useUSDPriceParameters: useUSDPriceParameters) {
+  const activeLocalCurrency = useActiveLocalCurrency()
+  const activeLocalCurrencyIsUSD = activeLocalCurrency === Currency.Usd
+
+  const { data: usdPrice, isLoading: isUSDPriceLoading } = useUSDPrice(...useUSDPriceParameters)
+  const { data: localCurrencyConversionRate, isLoading: isLocalCurrencyConversionRateLoading } =
+    useLocalCurrencyConversionRate(activeLocalCurrency, activeLocalCurrencyIsUSD)
```
just-toby, line 15: should we skip this fetch if `activeLocalCurrencyIsUSD` is true?

just-toby (approved):

src/hooks/useLocalCurrencyPrice.ts
JackShort: yeah thats exactly what happens

PR #7252, opened 2023-08-31
feat: dynamic FOT tax fetching

cartcrom:
## Description

Adds dynamic fetching for tax-rates for mainnet tokens, replacing previously implemented hardcoded solution.

Tax-fetching is done using async calls and custom caching logic rather than using redux-multicall, in order to prevent wasteful calls in certain scenarios, e.g. the switch currency button is clicked so the same tokens are being checked but in a different order.

Taxes are initially fetched by the swap component so that they can be displayed in the UI, and are then passed into trade objects.

### Manual testing

I looked at the networks tab to ensure that no new quote requests were fired as a result of taxes updating, and made sure that no new rpc calls for tax rates were being fired for tokens that had already been checked for taxes


src/pages/Swap/index.tsx
```diff
@@ -279,14 +278,13 @@ export function Swap({
...
 
   const [inputTokenHasTax, outputTokenHasTax] = useMemo(
-    () => [
-      !!currencies[Field.INPUT] && !getInputTax(currencies[Field.INPUT]).equalTo(0),
-      !!currencies[Field.OUTPUT] && !getOutputTax(currencies[Field.OUTPUT]).equalTo(0),
-    ],
-    [currencies]
+    () => [!inputTax.equalTo(0), !outputTax.equalTo(0)],
+    [inputTax, outputTax]
   )
```
just-toby, line 288: i don't think you need to memoize this anymore, checking equality against zero isn't alot of work.

src/hooks/useSwapTaxes.ts
```diff
@@ -0,0 +1,82 @@
...
+import { useWeb3React } from '@web3-react/core'
+import FOT_DETECTOR_ABI from 'abis/fee-on-transfer-detector.json'
+import { FeeOnTransferDetector } from 'abis/types'
+import { sendAnalyticsEvent } from 'analytics'
+import { ZERO_PERCENT } from 'constants/misc'
+import { useEffect, useState } from 'react'
+
+import { useContract } from './useContract'
+
+const FEE_ON_TRANSFER_DETECTOR_ADDRESS = '0x19C97dc2a25845C7f9d1d519c8C2d4809c58b43f'
```
just-toby, line 13: have you tested this hook while connected to a different chain?  just want to make sure it doesn't cause a crash or something

src/hooks/useSwapTaxes.ts
```diff
@@ -0,0 +1,82 @@
+import { InterfaceEventName } from '@uniswap/analytics-events'
+import { ChainId, Percent } from '@uniswap/sdk-core'
+import { WETH_ADDRESS as getWethAddress } from '@uniswap/universal-router-sdk'
```
just-toby, line 3: 😆 i see you're tired of the weird enum-function syntax. i like this change

src/hooks/useSwapTaxes.ts
```diff
@@ -0,0 +1,82 @@
...
+      addresses.forEach((address, index) => {
+        const { sellFeeBps, buyFeeBps } = data[index]
+        const sellTax = new Percent(sellFeeBps.toNumber(), 10000)
+        const buyTax = new Percent(buyFeeBps.toNumber(), 10000)
+
+        FEE_CACHE[address] = { sellTax, buyTax }
+      })
+    }
+  } catch {
+    console.error('could not fetch fee on transfer rate for tokens:', addresses)
```
just-toby, line 63: is this is expected to happen alot? just wondering if we should be logging to sentry. or if this is kinda expected for many tokens, maybe it should be downgraded to a warning

just-toby (approved):

src/hooks/useSwapTaxes.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+  }, [account, contract])
+  return contract
+}
+
+// TODO(WEB-2787): add tax-fetching for other chains
+const WETH_ADDRESS = getWethAddress(ChainId.MAINNET)
+const AMOUNT_TO_BORROW = 10000 // smallest amount that has full precision over bps
+
+const FEE_CACHE: { [address in string]?: { sellTax?: Percent; buyTax?: Percent } } = {}
```
gnewfield, line 34: I like the idea of a cache, but I'm wondering if the cost / benefit analysis works out in practice. Specifically, there may be a risk with staleness, since it's possible for a token's tax rates to change, in which case the cache would return the wrong values. Also, since the call is made through the user's provider, what is the benefit? Is there a high latency on using the FOT detector? Perhaps the increased speed would be a worthwhile value prop.

I definitely see caching being valuable when FOT detection is done in our backend (i.e. impacting latency of routing API) or if we're using our own providers and will incur the cost of these calls (as is the case for multichain features). 

What do you think?

src/hooks/useSwapTaxes.ts
cartcrom: will log a warning to sentry instead

src/hooks/useSwapTaxes.ts
```diff
@@ -0,0 +1,82 @@
...
+  outputTokenAddress: string | undefined
+) {
+  const addresses = []
+  if (inputTokenAddress && FEE_CACHE[inputTokenAddress] === undefined) {
+    addresses.push(inputTokenAddress)
+  }
+
+  if (outputTokenAddress && FEE_CACHE[outputTokenAddress] === undefined) {
+    addresses.push(outputTokenAddress)
+  }
```
cbachmeier, line 48:
```suggestion
  const addresses = [inputTokenAddress, outputTokenAddress].filter((address) => address && FEE_CACHE[address] === undefined)
```

src/hooks/useSwapTaxes.ts
cbachmeier: +1 seems like a good place for amplitude logging

src/hooks/useSwapTaxes.ts
cartcrom: it doesn't crash, but I'll add a check to return early if not on mainnet

src/hooks/useSwapTaxes.ts
```diff
@@ -0,0 +1,82 @@
...
+
+        FEE_CACHE[address] = { sellTax, buyTax }
+      })
+    }
+  } catch {
+    console.error('could not fetch fee on transfer rate for tokens:', addresses)
+  }
+
+  const inputTax = (inputTokenAddress ? FEE_CACHE[inputTokenAddress]?.sellTax : undefined) ?? ZERO_PERCENT
+  const outputTax = (outputTokenAddress ? FEE_CACHE[outputTokenAddress]?.buyTax : undefined) ?? ZERO_PERCENT
```
cbachmeier, line 67:
```suggestion
  const inputTax = inputTokenAddress ? FEE_CACHE[inputTokenAddress]?.sellTax : ZERO_PERCENT
  const outputTax = outputTokenAddress ? FEE_CACHE[outputTokenAddress]?.buyTax : ZERO_PERCENT
```

src/state/swap/hooks.tsx
```diff
@@ -104,6 +107,12 @@ export function useDerivedSwapInfo(state: SwapState, chainId: ChainId | undefine
 
   const inputCurrency = useCurrency(inputCurrencyId, chainId)
   const outputCurrency = useCurrency(outputCurrencyId, chainId)
+
+  const { inputTax, outputTax } = useSwapTaxes(
+    inputCurrency?.isToken ? inputCurrency.address : undefined,
```
cbachmeier, line 112: can't wait till we have to support a chain that's native has a FOT

src/state/routing/useRoutingAPITrade.ts
```diff
@@ -81,6 +88,8 @@ export function useRoutingAPITrade<TTradeType extends TradeType>(
     amount: amountSpecified,
     tradeType,
     routerPreference,
+    inputTax,
+    outputTax,
```
cbachmeier, line 92: you do a lot of work to pass the input and output tax to `src/state/routing/useRoutingAPITrade.ts` but could you just call `useSwapTaxes` in `useRoutingApiTrade` on `currencyIn` and `currencyOut`?

src/hooks/useSwapTaxes.ts
cartcrom: We're still using our own provider a good portion of the time, e.g. if the user hasn't connected a wallet yet but they're setting up swaps, or if they're connected via WalletConnect.

Staleness is definitely possible here, but uncommon to happen during the lifetime of a swap. I think it's best to conserve resources / performance rather than support tokens that change rates super quickly, and rely on the BE to handle that in the future if it becomes a problem. wdyt?

src/state/routing/useRoutingAPITrade.ts
cartcrom: once the BE handle this for us, we should be able to take all of the work bubbling up the taxes

src/hooks/useSwapTaxes.ts
cartcrom: Yeah I had thought about this, I hate the way the ternary looks, but I'm not sure if there is a better solution other than writing a helper function which seems overboard. Cause this results in `Percent | undefined`:
 
image

src/state/swap/hooks.tsx
cartcrom: 💀💸

src/hooks/useSwapTaxes.ts
gnewfield: I agree that staleness won't be common and definitely think we should prioritize resource conservation - thanks for clarifying those cases in which we're actually using our own provider!

gnewfield (dismissed):

src/hooks/useSwapTaxes.ts
cartcrom: This would result in an array of type `(string | undefined)[]`, so I prefer simple if-statements/readability here rather than a filter and type guard

src/pages/Swap/index.tsx
cartcrom: discussed on slack, gonna keep the memo since sdk-core/Fraction's equalTo is a bit verbose

just-toby (approved):

cbachmeier (approved):

marktoda:
:partying_face: !!!

PR #7265, opened 2023-09-05
refactor: price chart timestamps

cartcrom:
## Description

Refactors price charts tick / crosshair time formatting logic for better decomposition and readability

| Before | After |
| ------ | ------ | 
| image | image |
| image | image |

### QA (ie manual testing)

- [ ] Test that tick location and formatting has not changed across all time periods
- [ ] Test that crosshair formatting has not changed across all time periods


src/components/Charts/PriceChart/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+  const { interval, step } = TIME_PERIOD_INTERVAL_TABLE[timePeriod]
+  const ticks = interval.range(startDate, endDate, step).map((x) => x.valueOf() / 1000) // convert to seconds
+
+  if (ticks.length <= maxTicks) return ticks
+
+  const newTicks = []
+  const tickSpacing = Math.floor(ticks.length / maxTicks)
+  for (let i = 1; i < ticks.length; i += tickSpacing) {
+    newTicks.push(ticks[i])
+  }
```
JackShort, line 40: i don't think this is doing exactly what you want it to do. so you know the `tickSpacing` is actually `1/maxTicks` right. so what you actually want to do is increment by `1/maxTicks` and only choose one tick that is inside that area so something like this:

```
const tickSpacing = 1 / maxTicks
for (let i = 0; i < 1; i += tickSpacing) {
  const maxTick = Math.floor(i * ticks.length)
  newTicks.push(ticks[maxTick])
}
```

JackShort (approved):

src/components/Charts/PriceChart/utils.ts
cartcrom: this code was copied over from the existing solution, I think it's working as originally intended

tickSpacing currently represents how many (original) ticks apart the new ticks should be. what would 1/maxTicks represent? im confused why we would increment by that fraction in the for loop

JackShort: so with the old one if `ticks.length` is 1.9x the size of `maxTicks` the ticksSpacing is `1` so you are just basically getting the first `maxTicks` of `ticks`. If you do it by the way of using `1/maxTicks` thats basically the percentage of the `maxTicks` that you want to map `ticks` onto so you are able to graph the data across the whole subset of `ticks` instead of the first `maxTicks` of `ticks`

JackShort: current way: `maxTicks = 1000` `ticks.length = 1900` `newTicks = ticks[0...999]`
proposed way: `maxTicks = 1000` `ticks.length = 1900` `newTicks = ticks[0...1900]` where we grab every other 1.9 tick

cartcrom: Ahhh okay I think I understand what you're saying. I think the discrete set of ticks.length and maxTicks that this function currently deals with don't produce an edge case like this. I think since this is a refactor PR the functionality shouldn't change for now, but we can improve in the future esp if we have larger charts or more ticks.

One issue tho, I think your solution is more inclusive of all items in the original array, but doesn't produce evenly spaced ticks

image

JackShort: Ah true all good then

cbachmeier (approved):

JackShort (approved):

PR #7266, opened 2023-09-05
refactor: price chart utility functions

cartcrom:
## Description

Breaks price chart utilities into utils file, improves performance of getPriceBounds and getNearestPricePoint, and improves readability / documentation.

### QA (ie manual testing)

- [ ] Ensure that scrubbing through a chart appears the same as before

### DIFF
--- a/src/components/Charts/PriceChart/utils.ts
+++ b/src/components/Charts/PriceChart/utils.ts
@@ -1,5 +1,101 @@
-import { timeDay, timeHour, TimeInterval, timeMinute, timeMonth } from 'd3'
-import { TimePeriod } from 'graphql/data/util'
+import { bisector, ScaleLinear, timeDay, timeHour, TimeInterval, timeMinute, timeMonth } from 'd3'
+import { PricePoint, TimePeriod } from 'graphql/data/util'
+
+/**
+ * Returns the minimum and maximum values in the given array of PricePoints.
+ */
+export function getPriceBounds(prices: PricePoint[]): { min: number; max: number } {
+  if (!prices.length) return { min: 0, max: 0 }
+
+  let min = prices[0].value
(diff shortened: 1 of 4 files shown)


src/components/Charts/PriceChart/utils.ts
```diff
@@ -1,5 +1,101 @@
...
+ * @param prices - The original array of PricePoints
+ * @returns An object containing two arrays: fixedChart and blanks
+ */
+export function cleanUpPricePoints(prices: PricePoint[]) {
+  const fixedChart: PricePoint[] = [] // PricePoint array with 0 values removed
+  const blanks: [PricePoint, PricePoint][] = [] // PricePoint pairs that represent blank spaces in the chart
+  let lastValidPrice: PricePoint | undefined
+
+  prices.forEach((pricePoint, index) => {
+    const isValueNonZero = pricePoint.value !== 0
```
cbachmeier, line 37: nit: this is only used once and the check is pretty self explanatory, does it need its own const?

cbachmeier (approved):
LGTM

src/components/Charts/PriceChart/utils.ts
```diff
@@ -1,5 +1,101 @@
...
+  return { min, max }
+}
+
+/**
+ * Cleans up an array of PricePoints by removing zero values and marking gaps in data as blanks.
+ *
+ * @param prices - The original array of PricePoints
+ * @returns An object containing two arrays: fixedChart and blanks
+ */
+export function cleanUpPricePoints(prices: PricePoint[]) {
```
just-toby, line 31: can we just call this `cleanPricePoints` ? i feel like "clean up" is almost slang

just-toby (approved):

PR #7267, opened 2023-09-05
chore: displaying local currency and language formatting

JackShort:
* Adds a hook that returns the formatting language and currency to use so we can hide displaying the formatting behind the feature flag
* adds this to all formatNumbers calls

turning on the screenshot you should be able to change the lang formatting and the currency formatter
Screen Shot 2023-09-05 at 3 08 16 PM


src/components/AccountDrawer/AuthenticatedHeader.tsx
```diff
@@ -282,6 +283,8 @@ export default function AuthenticatedHeader({ account, openSettings }: { account
               {formatNumber({
                 input: totalBalance,
                 type: NumberType.PortfolioBalance,
+                locale: formatterLocale,
+                localCurrency: formatterLocalCurrency,
```
natew, line 287: feel like this would be nicer with a single hook:

```tsx
const { formatNumber } = useFormatter()

formatNumber({ input, type })
```

natew: or `useFormatNumber()` if we don't need other types of formatters / they have different performance costs

JackShort: that does sound better will impl that instead

src/utils/formatNumbers.ts
```diff
@@ -525,3 +529,42 @@ export function formatReviewSwapCurrencyAmount(amount: CurrencyAmount<Currency>)
...
+    }
+  }
+
+  return {
+    formatterLocale: DEFAULT_LOCALE,
+    formatterLocalCurrency: DEFAULT_LOCAL_CURRENCY,
+  }
+}
+
+// used to get the appropriate formatter functions with the correct locales passed into them
```
JFrankfurt, line 554: nit: use a complete sentence w/ caps and punctuation

src/utils/formatNumbers.ts
```diff
@@ -525,3 +529,42 @@ export function formatReviewSwapCurrencyAmount(amount: CurrencyAmount<Currency>)
...
+      formatNumber({ ...options, locale: formatterLocale, localCurrency: formatterLocalCurrency }),
+    [formatterLocalCurrency, formatterLocale]
+  )
+
+  return useMemo(
+    () => ({
+      formatNumber: formatNumberWithLocales,
+    }),
+    [formatNumberWithLocales]
+  )
```
JFrankfurt, line 569: at a glance this usememo feels like it might be premature optimization

JFrankfurt (approved):

src/utils/formatNumbers.ts
JackShort: this is not going to be the only function that is going to get returned - i have a pr in the stack that adds another function thats already approved

PR #7273, opened 2023-09-06
chore: formatter to use local currency price

JackShort:
* removes `useLocalUsdPrice` and implements similar functionality in the `useFormatter` hook to apply conversion rates to any formatters inside of it


src/utils/formatNumbers.ts
```diff
@@ -569,18 +581,47 @@ export function useFormatterLocales(): {
...
+  const previousSelectedCurrency = usePrevious(activeLocalCurrency)
+  const previousConversionRate = usePrevious(localCurrencyConversionRate)
+
+  const shouldFallbackToPrevious = !localCurrencyConversionRate && localCurrencyConversionRateIsLoading
+  const shouldFallbackToUSD = !localCurrencyConversionRate && !localCurrencyConversionRateIsLoading
+  const currencyToFormatWith = shouldFallbackToUSD
+    ? DEFAULT_LOCAL_CURRENCY
+    : shouldFallbackToPrevious
+    ? previousSelectedCurrency
+    : formatterLocalCurrency
```
JFrankfurt, line 600: nit: prefer not nesting ternary logic

src/nft/hooks/useUsdPrice.ts
```diff
@@ -1,14 +1,14 @@
...
 import tryParseCurrencyAmount from 'lib/utils/tryParseCurrencyAmount'
 import { GenieAsset } from 'nft/types'
 
 export const useNativeUsdPrice = (chainId: number = ChainId.MAINNET): number => {
   const nativeCurrency = useNativeCurrency(chainId)
   const parsedAmount = tryParseCurrencyAmount('1', nativeCurrency)
-  const usdcValue = useLocalCurrencyPrice(parsedAmount)?.data ?? 0
+  const usdcValue = useUSDPrice(parsedAmount)?.data ?? 0
   return usdcValue
 }
```
JFrankfurt, line 13: would it be possible to remove this file and use 'hooks/useUSDPrice' directly instead?

JFrankfurt (approved):

src/utils/formatNumbers.ts
```diff
@@ -362,12 +365,15 @@ const TYPE_TO_FORMATTER_RULES = {
...
   const rules = TYPE_TO_FORMATTER_RULES[type]
   for (const rule of rules) {
+    const shouldConvertInput = rule.formatterOptions.currency && conversionRate
+    const convertedInput = shouldConvertInput ? input * conversionRate : input
+
     if (
-      (rule.exact !== undefined && input === rule.exact) ||
-      (rule.upperBound !== undefined && input < rule.upperBound)
+      (rule.exact !== undefined && convertedInput === rule.exact) ||
+      (rule.upperBound !== undefined && convertedInput < rule.upperBound)
```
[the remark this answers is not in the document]
JackShort: I agree with you however I made this decision because we are formatting so many different number types so the currency conversion would have to deal with all of them so I thought maybe the best thing would to be just go to the source where it's sent as a number. Also dev-x wise I thought another hook might be annoying to be honest when all you really want to do is format the number

src/utils/formatNumbers.ts
JackShort: I also agree with you - I too hate the ternaries, however they let me declare this as a const so to me that was more upside than a mutable variable

src/utils/formatNumbers.ts
```diff
@@ -569,18 +581,47 @@ export function useFormatterLocales(): {
...
   const { formatterLocale, formatterLocalCurrency } = useFormatterLocales()
 
+  const activeLocalCurrencyIsUSD = activeLocalCurrency === GqlCurrency.Usd
+  const { data: localCurrencyConversionRate, isLoading: localCurrencyConversionRateIsLoading } =
+    useLocalCurrencyConversionRate(activeLocalCurrency, activeLocalCurrencyIsUSD)
+
+  const previousSelectedCurrency = usePrevious(activeLocalCurrency)
+  const previousConversionRate = usePrevious(localCurrencyConversionRate)
+
+  const shouldFallbackToPrevious = !localCurrencyConversionRate && localCurrencyConversionRateIsLoading
```
cbachmeier, line 594: can we show a loading state instead of the previous conversion rate? or also show the previous symbol as well? For large diffs such as USD -> JPY the symbol changes and you're momentarily flashed a very innaccurate price. This would be further emphasized on slower internet connections

cbachmeier (dismissed):
LGTM just have a suggestion/design Q

src/utils/formatNumbers.ts
JackShort: I think this is a weird caching thing by vercel because thats how it previously worked so i updated it so it stayed on the previous symbol and currency so there was no flashing and its smooth - you can observe this if you run it locally

src/pages/Swap/index.tsx
```diff
@@ -34,12 +34,12 @@ import { asSupportedChain, isSupportedChain } from 'constants/chains'
 import { getSwapCurrencyId, TOKEN_SHORTHANDS } from 'constants/tokens'
 import { useCurrency, useDefaultActiveTokens } from 'hooks/Tokens'
 import { useIsSwapUnsupported } from 'hooks/useIsSwapUnsupported'
-import { useLocalCurrencyPrice } from 'hooks/useLocalCurrencyPrice'
 import { useMaxAmountIn } from 'hooks/useMaxAmountIn'
 import usePermit2Allowance, { AllowanceState } from 'hooks/usePermit2Allowance'
 import usePrevious from 'hooks/usePrevious'
 import { SwapResult, useSwapCallback } from 'hooks/useSwapCallback'
 import { useSwitchChain } from 'hooks/useSwitchChain'
+import { useUSDPrice } from 'hooks/useUSDPrice'
```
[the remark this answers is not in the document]
JackShort: `useUSDPrice` still returns the raw USD price the function that converts is the `formatNumber` function from `useFormatter`

tinaszheng (approved):

cbachmeier (approved):

PR #7275, opened 2023-09-07
feat: path-based routing

zzmp:
This reverts commit 3f05a88409e86c6c86930c5abab1f6f632ff69b0, which had reverted c9b4016b78e597b824c98e4219867e0af301c755.

## Description

This PR brings in a runtime workaround where router is chosen depending on the origin.

Ultimately we want to use BrowserRouter and path-based routing for Uniswap. Unfortunately, IPFS does not support catch-all, which means all paths different than `/` would not hit our `index.html` and would result in a 404 error as such.

This change is transparent, since all routing happens via `<Link />` abstraction that deals with URLs. 

### Where do we turn this on

Everywhere but on IPFS. We treat IPFS deployment target as every target where `isProductionEnv` is true, but it does not match own origin.


JackShort (approved):
lgtm

src/tracing/errors.test.ts
```diff
@@ -11,17 +11,6 @@ describe('beforeSend', () => {
...
-    test('should remove "/#" from the request URL', () => {
-      const event = {
-        request: {
-          url: 'https://app.uniswap.org/#/example',
-        },
-      } as ErrorEvent
-
-      beforeSend(event, {})
-      expect(event.request?.url).toBe('https://app.uniswap.org/example')
-    })
```
cbachmeier, line 23: are we removing this functionality? ie no longer handling converting from hash to path based urls? This will break a lot of downstream users who have hardcoded Uni urls

cypress/e2e/wallet-dropdown.test.ts
```diff
@@ -153,6 +153,8 @@ describe('Wallet Dropdown', () => {
       cy.contains('USD')
 
       cy.visit('/?cur=AUD', { featureFlags: [FeatureFlag.currencyConversion] })
+      cy.get(getTestSelector('web3-status-connected')).click()
+      cy.get(getTestSelector('wallet-settings')).click()
```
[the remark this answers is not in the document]
zzmp: This was needed for these tests to pass, and the functionality still works (via this and manual testing), so I think it's a necessary addition.

src/tracing/errors.test.ts
zzmp: good point - I added this back in.

cypress/e2e/token-explore.test.ts
```diff
@@ -69,8 +69,5 @@ describe('Token explore', () => {
     cy.get(getTestSelector('tokens-network-filter-selected')).click()
     cy.get(getTestSelector('tokens-network-filter-option-optimism')).click()
     cy.get(getTestSelector('tokens-network-filter-selected')).should('contain', 'Optimism')
-    cy.reload()
-    cy.get(getTestSelector('tokens-network-filter-selected')).should('contain', 'Optimism')
-    cy.get(getTestSelector('chain-selector-logo')).invoke('attr', 'alt').should('eq', 'Ethereum')
```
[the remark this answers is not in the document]
zzmp: I don't know. They were added as part of the last revert. I'll add the last line back in, but I don't think the reload is necessary.

src/utils/env.ts
```diff
@@ -23,6 +23,25 @@ export function isAppUniswapStagingOrg({ hostname }: { hostname: string }): bool
...
+  if (isProductionEnv()) {
+    if (
+      isAppUniswapOrg(window.location) ||
+      isAppUniswapStagingOrg(window.location) ||
+      // Cypress tests
+      isLocalhost(window.location)
+    ) {
+      return true
+    }
+    return false
```
just-toby, line 36: can you add a comment explaining when `false` is expected? it's just IPFS right?

src/pages/App.tsx
```diff
@@ -211,7 +219,16 @@ export default function App() {
             <Suspense fallback={<Loader />}>
               {isLoaded ? (
                 <Routes>
-                  <Route path="/" element={<Landing />} />
+                  <Route
+                    path="/"
+                    element={
+                      // if react-router-dom matches "/" with window.location.hash defined, it means that we're
+                      // using BrowserRouter and can safely redirect to a path route
```
just-toby, line 226: do you think it would be helpful to mention IPFS here as the reason why react-router-dom might not match "/" ?

just-toby: also, should we just use `isBrowserRouterEnabled()` in this file to be more explicit as to when BrowserRouter is in use?

just-toby (approved):

cbachmeier (approved):

just-toby (approved):

PR #7276, opened 2023-09-07
fix: hotfix fot ui bug in swap details dropdown + update flag and sentry

cartcrom:
## Description

Fixes visual bug where FOT details in swap dropdown were wrong color and lacked a loading state

this hotfix is also coupled with a change for which statsig flag is used to gate FOT logic. the current flag used in prod is having issues possibly related to caching

also removes sentry logging since we now expect tax lookups to fail regularly for tokens that have insufficient v2 liquidity

### Before
https://github.com/Uniswap/interface/assets/39385577/278fa5e8-24c2-43eb-b79c-1180fa6a0fc4  

### After
https://github.com/Uniswap/interface/assets/39385577/0f1c1033-4fdd-406b-a81c-4baf64414fbe

### DIFF
--- a/src/components/swap/AdvancedSwapDetails.tsx
+++ b/src/components/swap/AdvancedSwapDetails.tsx
@@ -191,6 +191,8 @@ function TokenTaxLineItem({
   type: 'input' | 'output'
   syncing: boolean
 }) {
+  if (syncing) return null
+
   const [currency, percentage] =
     type === 'input' ? [trade.inputAmount.currency, trade.inputTax] : [trade.outputAmount.currency, trade.outputTax]
 
@@ -213,9 +215,7 @@ function TokenTaxLineItem({
       >
         <ThemedText.BodySmall color="neutral2">{`${currency.symbol} fee`}</ThemedText.BodySmall>
       </MouseoverTooltip>
-      <TextWithLoadingPlaceholder syncing={syncing} width={50}>
-        <ThemedText.BodySmall>{formatPriceImpact(percentage)}</ThemedText.BodySmall>
-      </TextWithLoadingPlaceholder>
+      <ThemedText.BodySmall>{formatPriceImpact(percentage)}</ThemedText.BodySmall>
     </RowBetween>
   )
 }
--- a/src/components/swap/AdvancedSwapDetails.tsx
+++ b/src/components/swap/AdvancedSwapDetails.tsx
@@ -84,8 +84,8 @@ export function AdvancedSwapDetails({ trade, allowedSlippage, syncing = false }:
       )}
       {isClassicTrade(trade) && (
         <>
-          <TokenTaxLineItem trade={trade} type="input" />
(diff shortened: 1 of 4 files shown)


just-toby (approved):

natew (approved):

JFrankfurt (approved):

src/components/swap/AdvancedSwapDetails.tsx
```diff
@@ -203,9 +211,11 @@ function TokenTaxLineItem({ trade, type }: { trade: ClassicTrade; type: 'input'
           </>
         }
       >
-        <ThemedText.BodySmall color="textSecondary">{`${currency.symbol} fee`}</ThemedText.BodySmall>
+        <ThemedText.BodySmall color="neutral2">{`${currency.symbol} fee`}</ThemedText.BodySmall>
```
cbachmeier, line 214: this should also have a syncing/loading skeleton. when switching FOT tokens, the old one is showing while loading

cbachmeier (changes requested):

src/components/swap/AdvancedSwapDetails.tsx
cbachmeier: loading fee

src/featureFlags/index.tsx
```diff
@@ -13,7 +13,7 @@ export enum FeatureFlag {
   uniswapXExactOutputEnabled = 'uniswapx_exact_output_enabled',
   multichainUX = 'multichain_ux',
   currencyConversion = 'currency_conversion',
-  fotAdjustedmentsEnabled = 'fot_adjustments_enabled',
+  fotAdjustedmentsEnabled = 'fot_dynamic_adjustments_enabled',
```
cbachmeier, line 16: super nit and non-blocking but this long name is pushing over the dropdown, may need to add truncation
Screenshot 2023-09-07 at 1 45 43 PM

src/components/swap/AdvancedSwapDetails.tsx
cbachmeier: it also shows when switching to non FOT
loadingFeeNon

cartcrom: I did try this, and it looked visually unbalanced w/ the rest of the rows in loading state. I'm not sure how big of an issue this is, since there's is still an overall indication that that row is about to update.

@bryankuuu @callil thoughts here?

cbachmeier: i'll defer to design but I think it's confusing especially going from a FOT to normal and seeing a fee loading so I would strongly recommend a loader or just hiding the row until a fee has loaded

cartcrom: hiding row makes a lot of sense, just added that ty

cbachmeier (approved):

cbachmeier (approved):

PR #7280, opened 2023-09-07
test: fix cloud tests

zzmp:
## Description
Updates the infrastructure for running cloud tests so that they will run in GitHub actions.
Whether the proxy was running or not was incorrectly detected by `jest-test-server`, so this functionality is being moved to use `start-server-and-test`.


functions/jest.config.json
```diff
@@ -1,6 +1,4 @@
 {
-  "globalSetup": "<rootDir>/global-setup.ts",
-  "globalTeardown": "<rootDir>/global-teardown.ts",
```
cbachmeier, line 3: why are we also removing global jest setup and teardown?

functions/nfts/collection/collection.test.ts
```diff
@@ -16,44 +16,53 @@ const collections = [
...
+    address: '0xed5af388653567af2f388e6224dc7c4b3241c545',
+    collectionName: '0xed5af388653567af2f388e6224dc7c4b3241c545',
+    image: 'http://127.0.0.1:3000/api/image/nfts/collection/0xed5af388653567af2f388e6224dc7c4b3241c545',
+  },
 ]
 
-test.each(invalidCollections)(
-  'should not inject metadata for invalid collection urls',
-  async (url) => {
+test.each([...collections, ...nonexistentCollections])(
```
cbachmeier, line 27: should `nonexistentCollections` be part of the test below `should inject default metadata for nonexistent collections` instead of this test for valid collections?

cbachmeier (approved):
LGTM, have a couple Qs

functions/jest.config.json
zzmp: These have been replaced in ci by start-server-and-test. Manual testing now requires running start:cloud and test:cloud separately. This is a more reliable way of making sure the server is up, and allows for more iterative testing (via fast-refresh server).

functions/nfts/collection/collection.test.ts
zzmp: Good call - the tests just need to be renamed.

patches/@vercel+og+0.5.8.patch
```diff
@@ -13,35 +13,34 @@ index 5187f88..c4a1c41 100644
  
  // src/emoji/index.ts
  var U200D = String.fromCharCode(8205);
-@@ -18809,18 +18809,18 @@ async function render(satori, resvg, opts, defaultFonts, element) {
+@@ -18809,18 +18809,15 @@ async function render(satori, resvg, opts, defaultFonts, element) {
  // src/index.edge.ts
  var initializedResvg = initWasm(resvg_wasm);
  var initializedYoga = initYoga(yoga_wasm).then((yoga2) => Ll(yoga2));
 -var fallbackFont = fetch(new URL("./noto-sans-v27-latin-regular.ttf", import.meta.url)).then((res) => res.arrayBuffer());
-+// var fallbackFont = fetch(new URL("https://fonts.gstatic.com/s/notosans/v28/o-0IIpQlx3QUlC5A4PNr6zRF.ttf", import.meta.url)).then((res) => res.arrayBuffer());
```
just-toby, line 21: do we need to do anything to maintain this patch? if the `@vercel/og` version updates will this patch still be applied, or will we lose it and the test starts failing again?

zzmp: because we have @vercel/og pinned through yarn.lock, we'd need to actively take any new updates, at which point we would need to make a new patch for the new version. Without any patch the tests would fail, so there should be a blocker to anyone just updating the package without also updating the patch.

just-toby (approved):

JFrankfurt (approved):

JFrankfurt:
Should we expect all tests to pass here? Looks like a few are borked

JFrankfurt (approved):

PR #7294, opened 2023-09-11
fix: [Spore] update fonts

callil:
## Description
Updates the font files to use the latest "real" woff files. It's the same font as we use already so we don't expect any visual changes.


natew (approved):

mattbspector:

callil:
I'd like to re-open this please.

natew:
Noticing that `book` went way up in size @callil do we know why? Should we subset these fonts so they don't use as large of a character set? There's obviously some concerns about internationalization, but honestly I'm not the most experienced with how that is typically handled. I'd imagine we could get a subset that covers 90% of people somehow lightweight, and then lazy load a bigger one, but I will research into what the strategies are here.

natew:
Oh this is cool:

https://web.dev/font-best-practices/#subset-fonts

Looks like we can use unicode-range and subsets together so we can serve subsets per-language, seems like the perfect setup. 120kb is right at the level where we'd see noticeable improvements, I think for example just a pure english subset would be closer to 20kb, and fonts are pretty critical in terms of avoiding flashes of bad styling

callil:
@natew can you take this over for me? We still want to use the real font if possible

PR #7295, opened 2023-09-11
chore: remove extra blocknumber from dep array

JackShort:
* `blocknumber` was added twice to the dep array due to merge of #7231


tinaszheng (approved):
OOPS

PR #7300, opened 2023-09-11
feat: move ga events to amplitude

just-toby:
## Description

this is part of #6554 based on the feedback on that PR - instead of completely removing those events we'll log them to amplitude.

in a follow up PR i'll actually delete the dead GA code

## Screen capture

image

image

image

image

image

### DIFF
--- a/src/pages/AddLiquidity/index.tsx
+++ b/src/pages/AddLiquidity/index.tsx
@@ -55,7 +55,7 @@ import useTransactionDeadline from '../../hooks/useTransactionDeadline'
 import { useV3PositionFromTokenId } from '../../hooks/useV3Positions'
 import { Bound, Field } from '../../state/mint/v3/actions'
 import { useTransactionAdder } from '../../state/transactions/hooks'
-import { TransactionType } from '../../state/transactions/types'
+import { TransactionInfo, TransactionType } from '../../state/transactions/types'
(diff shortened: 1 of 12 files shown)


JFrankfurt (approved):

src/pages/AddLiquidity/index.tsx
```diff
@@ -434,11 +433,6 @@ function AddLiquidity() {
     const maxPrice = pricesAtLimit[Bound.UPPER]
     if (maxPrice) searchParams.set('maxPrice', maxPrice.toSignificant(5))
     setSearchParams(searchParams)
-
-    sendEvent({
-      category: 'Liquidity',
-      action: 'Full Range Clicked',
-    })
```
cbachmeier, line 441: are we no longer tracking this one?

cbachmeier (commented):
Looks good, just had a Q about 1 event

src/pages/AddLiquidity/index.tsx
just-toby: no i don't think it's necessary but we should be including more details in the add events... updating

just-toby: also this wasn't on the list of events Blair requested we migrate in #6554

cbachmeier (approved):

PR #7301, opened 2023-09-11
chore: updating all numbers in swap flow to be converted properly

JackShort:
* Updates all numbers in the swap flow to be formatted and converted properly

|  Before   |
| ------------ |
|  Screen Shot 2023-09-12 at 4 13 58 PM Screen Shot 2023-09-12 at 4 16 37 PM |

|   After   |
| ----------- |
| Screen Shot 2023-09-12 at 4 13 36 PM Screen Shot 2023-09-12 at 4 16 43 PM|


src/utils/formatNumbers.ts
```diff
@@ -510,50 +526,15 @@ export const priceToPreciseFloat = (price: Price<Currency, Currency> | undefined
...
-}
-
 const MAX_AMOUNT_STR_LENGTH = 9
 
-export function formatReviewSwapCurrencyAmount(amount: CurrencyAmount<Currency>): string {
-  let formattedAmount = formatCurrencyAmount({ amount, type: NumberType.TokenTx })
+export function formatReviewSwapCurrencyAmount(
+  amount: CurrencyAmount<Currency>,
+  locale: SupportedLocale = DEFAULT_LOCALE
+): string {
```
just-toby, line 534: to avoid confusion i think you should have one entrypoint/interface to the whole suite of formatters. it seems like the hook below is the interface. so we shouldn't export these functions anymore.

if there's a use case for formatting these values with locales that are not the "active" `formatterLocale` then you should build that into the interface below. 

i just think having the interface surface be this wide/unconstrained will make it easy for new users of formatters to use the wrong function and introduce an incorrectly formatted value in the future.

src/components/swap/SwapModalFooter.tsx
```diff
@@ -28,6 +28,8 @@ import { GasBreakdownTooltip } from './GasBreakdownTooltip'
 import { SwapCallbackError, SwapShowAcceptChanges } from './styled'
 import { Label } from './SwapModalHeaderAmount'
 
+const sixFigsFormatterRules: FormatterRule[] = [{ upperBound: Infinity, formatterOptions: SIX_SIG_FIGS_NO_COMMAS }]
```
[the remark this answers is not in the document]
JackShort: i made it so you can make one-off rules

src/utils/formatNumbers.ts
JackShort: agree with this - making this a draft to refactor it to this

JackShort: actually the reason for this is for the tests - in a pr ontop of this i am going to do this refactor

src/components/swap/SwapModalFooter.tsx
```diff
@@ -164,8 +169,14 @@ export default function SwapModalFooter({
...
-                ? `${trade.minimumAmountOut(allowedSlippage).toSignificant(6)} ${trade.outputAmount.currency.symbol}`
-                : `${trade.maximumAmountIn(allowedSlippage).toSignificant(6)} ${trade.inputAmount.currency.symbol}`}
+                ? `${formatCurrencyAmount({
+                    amount: trade.minimumAmountOut(allowedSlippage),
+                    type: sixFigsFormatterRules,
+                  })} ${trade.outputAmount.currency.symbol}`
+                : `${formatCurrencyAmount({
+                    amount: trade.maximumAmountIn(allowedSlippage),
+                    type: sixFigsFormatterRules,
+                  })} ${trade.inputAmount.currency.symbol}`}
```
JFrankfurt, line 179: possible to remove the duplication here?

src/utils/formatNumbers.ts
JackShort: this was done in this pr: #7308

src/components/swap/SwapModalFooter.tsx
JackShort: they aren't the same code - one is `minimumAmountOut` the other is `maximumAmountOut`

src/utils/formatTickPrice.ts
```diff
@@ -20,5 +20,5 @@ export function formatTickPrice({ price, atLimit, direction, placeholder, number
     return placeholder
   }
 
-  return formatPrice(price, numberType ?? NumberType.TokenNonTx)
+  return formatPrice({ price, type: numberType ?? NumberType.TokenNonTx })
```
cbachmeier, line 23: this doesn't take in locale which i believe is why on the TDP the price shown in the chart isn't converted
Screenshot 2023-09-14 at 10 54 55 AM

src/utils/formatNumbers.ts
```diff
@@ -592,14 +573,13 @@ function handleFallbackCurrency(
 
 // Constructs an object that injects the correct locale and local currency into each of the above formatter functions.
 export function useFormatter() {
-  const activeLocalCurrency = useActiveLocalCurrency()
   const { formatterLocale, formatterLocalCurrency } = useFormatterLocales()
 
-  const activeLocalCurrencyIsUSD = activeLocalCurrency === GqlCurrency.Usd
+  const activeLocalCurrencyIsUSD = formatterLocalCurrency === GqlCurrency.Usd
```
cbachmeier, line 578:
```suggestion
  const formatterLocalCurrencyIsUSD = formatterLocalCurrency === GqlCurrency.Usd
```

JFrankfurt (approved):

cbachmeier (approved):

PR #7303, opened 2023-09-12
fix: remove GA code

just-toby:
## Description

removes dead code that was used to log to Google Analytics. the underlying logger implementation was already gutted, so this code wasn't doing anything.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2832/dead-code-remove-google-analytics-code-from-interace

### TEST PLAN

- [x] typecheck
- [x] automated tests

Screenshot 2023-09-12 at 10 01 26 AM


public/index.html
```diff
@@ -36,6 +36,8 @@
     -->
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
 
+    <link rel="preconnect" href="%REACT_APP_AMPLITUDE_PROXY_URL%" />
```
JFrankfurt, line 39: preconnect is new to me--maybe we should add one for infura?

just-toby: same here, i'm just following a suggestion from the original PR #6554

gnewfield (approved):

JFrankfurt (approved):

natew (approved):

public/index.html
zzmp: we shouldn't add too many, and the infura subdomain may not be known at first (the subdomains would each require a distinct connection).

I think it's best to just update this one. right now it includes a path, which is unnecessary; it also needs to be crossorigin. See https://www.debugbear.com/resource-hint-validator?url=app.corn-staging.com.

PR #7308, opened 2023-09-13
chore: only exposing useFormatter

JackShort:
* updates all functions that call functions in `numberFormat` to call the callback function rendered by `useFormatter`
* going to make an eslint rule and move `useFormatter` to its own file in another pr after this is approved because merge conflicts on that would go crazy

__most of the additions/deletions come from updating the formatNumbers tests to deal with the hook version__
_also sorry for doing so many functions at once I was doing one at a time and it was taking too long_


zzmp:
What's the advantage to using a hook here over a utility function?

JackShort:
> What's the advantage to using a hook here over a utility function?

the hook takes into account the currency to convert it into and the locale

just-toby (approved):
code changes lgtm, but i'm open to more discussion about the design for this if needed.

i like how you have to go through the hook to format something now, meaning it's harder to bypass the user's locale.

but it's still not ideal that we're passing around formatting functions. what if these functions suddenly need to format a new number with a different rule,, then we're adding new parameters?

JackShort:
> code changes lgtm, but i'm open to more discussion about the design for this if needed.
> 
> i like how you have to go through the hook to format something now, meaning it's harder to bypass the user's locale.
> 
> but it's still not ideal that we're passing around formatting functions. what if these functions suddenly need to format a new number with a different rule,, then we're adding new parameters?

since everything calls `numberFormat` in the end of the day you could just pass `numberFormat` around and call it with your own rules. I also added the functionality of adding one off rules!

JackShort:
> > code changes lgtm, but i'm open to more discussion about the design for this if needed.
> > i like how you have to go through the hook to format something now, meaning it's harder to bypass the user's locale.
> > but it's still not ideal that we're passing around formatting functions. what if these functions suddenly need to format a new number with a different rule,, then we're adding new parameters?
> 
> since everything calls `numberFormat` in the end of the day you could just pass `numberFormat` around and call it with your own rules. I also added the functionality of adding one off rules!

I am also open to the idea of only just returning `numberFormat` in the hook and changing the other functions to generate inputs to `numberFormat` since they are basically just calling `numberFormat` internally - open to changing designs

src/components/AccountDrawer/MiniPortfolio/Activity/parseLocal.test.ts
```diff
@@ -237,6 +238,8 @@ jest.mock('../../../../state/transactions/hooks', () => {
 
 describe('parseLocalActivity', () => {
   it('returns swap activity fields with known tokens, exact input', () => {
+    const { formatNumber } = renderHook(() => useFormatter()).result.current
```
cbachmeier, line 241: could you define this one globally or beforeEach to avoid the repetition in this file?

src/components/swap/SwapModalFooter.tsx
```diff
@@ -85,7 +84,7 @@ export default function SwapModalFooter({
   const label = `${trade.executionPrice.baseCurrency?.symbol} `
   const labelInverted = `${trade.executionPrice.quoteCurrency?.symbol}`
   const formattedPrice = formatNumber({
```
cbachmeier, line 86: will this overwrite the formatting you added in https://github.com/Uniswap/interface/pull/7301/files#diff-d0161b639d6c25d54a994e66cf2be7521eb5be41d7d8e7c723a07210b5fd9f5fR87 ?

src/utils/formatNumbers.test.ts
```diff
@@ -1,523 +1,461 @@
...
-  expect(
-    formatNumber({ input: 0.00123, type: NumberType.FiatTokenStats, locale: 'ja-JP', localCurrency: Currency.Cad })
-  ).toBe('<CA$0.01')
-  expect(formatNumber({ input: 0, type: NumberType.FiatTokenStats })).toBe('-')
-  expect(
-    formatNumber({ input: 0, type: NumberType.FiatTokenStats, locale: 'ja-JP', localCurrency: Currency.Cad })
-  ).toBe('-')
-})
+  it('formats token reference numbers correctly', () => {
+    const { formatNumber } = renderHook(() => useFormatter()).result.current
```
cbachmeier, line 25: same Q here about mocking `formatNumber` globally

cbachmeier (approved):

src/components/swap/SwapModalFooter.tsx
JackShort: no it shouldn't - i didn't change anything from the base branch and thats the base branch

JFrankfurt (approved):

src/components/swap/SwapModalFooter.tsx
JackShort: oh sorry - yes it should because I am deleting this function. Sorry github was bugging out and couldn't see the blame

src/components/AccountDrawer/MiniPortfolio/Activity/parseLocal.test.ts
JackShort: the hook has to be rendered inside of the test

JFrankfurt (approved):

cbachmeier (approved):

PR #7312, opened 2023-09-14
fix: synchronously load first page

zzmp:
## Description

Loads top-level modals synchronously, as they block the first paint.
This prevents a waterfall, which - on my 500 Mbps connection - improves cold load time by 200ms.

In the following diagnostic screenshots, the red vertical line denotes Load, which correlates to Largest Contentful Paint (LCP).

| Before | After |
| - | - |
| <image src="https://github.com/Uniswap/interface/assets/5403956/5ea85d77-d9d9-41e0-8f6a-5addcfa22320"/> | <image src="https://github.com/Uniswap/interface/assets/5403956/b603db8d-7e03-45c2-a21d-9535a62b7a0a"/> |

### DIFF
--- a/src/components/TopLevelModals/index.tsx
+++ b/src/components/TopLevelModals/index.tsx
@@ -1,18 +1,16 @@
 import { useWeb3React } from '@web3-react/core'
 import { OffchainActivityModal } from 'components/AccountDrawer/MiniPortfolio/Activity/OffchainActivityModal'
 import UniwalletModal from 'components/AccountDrawer/UniwalletModal'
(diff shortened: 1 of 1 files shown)


just-toby (approved):
is there a way to make these components NOT block the first paint?

zzmp:
> is there a way to make these components NOT block the first paint?

Wrapping them in their own `<Suspense>`, but that may risk modals not working on first load. We could potentially lazy-load _all_ modals.

tinaszheng (approved):

PR #7314, opened 2023-09-14
feat: lazy-load MiniPortfolio, avoid eager data fetching

zzmp:
## Description

Does not render the AccountDrawer when it is not open, which avoids eagerly- and over- fetching data.
Lazy-loads the MiniPortfolio to improve pageload.

This will make the AccountDrawer populate slower, but should be a minor change, and will improve pageload and data usage overall. AccountDrawer uses skeleton loaders, so it will not appear broken.

## Testing

To test:
- Navigate to the preview url
- Open your devtools (F12)
- In the Network pane, set throttling to "Fast 3G"
- Open the AccountDrawer
  - Note that the MiniPortfolio takes a second to load, and then the data loads after you navigate to it
    This is in contrast to the current state, where data is pre-loaded even if you don't request it

### DIFF
--- a/src/components/AccountDrawer/AuthenticatedHeader.tsx
+++ b/src/components/AccountDrawer/AuthenticatedHeader.tsx
@@ -17,12 +17,12 @@ import useENSName from 'hooks/useENSName'
 import { useProfilePageState, useSellAsset, useWalletCollections } from 'nft/hooks'
 import { useIsNftClaimAvailable } from 'nft/hooks/useIsNftClaimAvailable'
 import { ProfilePageStateType } from 'nft/types'
-import { useCallback, useState } from 'react'
+import { lazy, Suspense, useCallback, useState } from 'react'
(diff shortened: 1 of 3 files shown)


natew (approved):

cartcrom (changes requested):
This is resulting in an bad transition state:

image

prod vs preview:

https://github.com/Uniswap/interface/assets/39385577/2f073770-81fc-4791-96fb-332b65214207

cartcrom:
The account drawer does _not_ eagerly or overly fetch data. Porfolio queries are fetched by the PrefetchedBalancesWrapper.

View the graphql fetches on prod:

https://github.com/Uniswap/interface/assets/39385577/d74e2c77-33ff-4d05-999e-36d0b98995c1

The data is only fetched once the account button is hovered. See there is a loading state, because the data hasn't come in yet. Other than lazy loading the component and changing the opacity, I'm not sure if this PR does anything?

PR #7317, opened 2023-09-14
feat: lazy-load smart-order-router

zzmp:
## Description
Lazy loads the smart-order router. This significantly reduces the initial bundle required for pageload.
Adds a lint rule to prevent regression, by only allowing types (which are compiled out) to be synchronously loaded.


cbachmeier (approved):
Verified I can still receive client side quotes. MiniPort also looks identical to staging
Screenshot 2023-09-15 at 9 53 40 AM

tinaszheng (approved):

PR #7319, opened 2023-09-15
fix: lazy-load web3-react libs

zzmp:
## Description
Fixes web3-react so that large libs (especially coinbase) are lazy-loaded. This improves pageload, as coinbase is a ~500kb package.
This is an issue because cjs libs use `require`, which is not split into an async package by webpack (see https://webpack.js.org/api/module-methods/#requireensure).

Until a better fix is found - presumably either (a) re-compiling our own packages, or (b) adding a custom loader to translate `__importStar(require('package'))` to `__importStar(await import('package'))` - this is being done as a patch for web3-react, so that the *large* savings of not loading coinbase-wallet on pageload can be realized.

### DIFF
--- a/patches/@web3-react+coinbase-wallet+8.2.2.patch
+++ b/patches/@web3-react+coinbase-wallet+8.2.2.patch
@@ -1,0 +1,13 @@
+diff --git a/node_modules/@web3-react/coinbase-wallet/dist/index.js b/node_modules/@web3-react/coinbase-wallet/dist/index.js
+index f38d06e..4f8fa19 100644
+--- a/node_modules/@web3-react/coinbase-wallet/dist/index.js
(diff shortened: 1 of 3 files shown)


JFrankfurt:
How much work would it take to solve this upstream?

zzmp:
> How much work would it take to solve this upstream?

I think it would need to be published using `import` syntax, as esm, whereas right now it is only published as cjs. My fear is just that that work is very easy to get wrong, which is why I think this hot-fix is appropriate as an interim fix.

cbachmeier (approved):
Vercel preview loaded in 540ms for me down from >1s. Verified I can still connect to MM and WC to sign txs

JFrankfurt (approved):

PR #7321, opened 2023-09-15
fix: fetch balances when opening token selector in LP page

just-toby:
## Description

we're not properly fetching the user's token balances on the LP page. this should happen when hovering over the token input similar to the swap component.

to fix we just want to wrap the token input with a `PrefetchBalancesWrapper` which triggers the fetch

_Slack thread:_ https://uniswapteam.slack.com/archives/C04D07TQBT4/p1694811755326329

### Reproducing the error

1. go to add liquidity, v2 or v3 - notice that the token balances don't load when you open the selector

### QA (ie manual testing)

- [x] repeat step 1 above but notice that the balances load now


src/components/CurrencyInputPanel/index.tsx
```diff
@@ -230,45 +231,50 @@ export default function CurrencyInputPanel({
...
-                            '...' +
-                            currency.symbol.slice(currency.symbol.length - 5, currency.symbol.length)
-                          : currency?.symbol) || <Trans>Select a token</Trans>}
-                      </StyledTokenName>
-                    )}
-                  </RowFixed>
-                  {onCurrencySelect && <StyledDropDown selected={!!currency} />}
-                </Aligner>
-              </CurrencySelect>
+              <PrefetchBalancesWrapper shouldFetchOnAccountUpdate={modalOpen}>
```
just-toby, line 234: just wrapping with this

src/components/CurrencyInputPanel/index.tsx
```diff
@@ -4,6 +4,7 @@ import { Currency, CurrencyAmount } from '@uniswap/sdk-core'
 import { Pair } from '@uniswap/v2-sdk'
 import { useWeb3React } from '@web3-react/core'
 import { TraceEvent } from 'analytics'
+import PrefetchBalancesWrapper from 'components/AccountDrawer/PrefetchBalancesWrapper'
```
JFrankfurt, line 7: path name might lead people to believe this is exclusively an account component--should prob bump it up to a root level component now that it's used here

JFrankfurt (changes requested):
lgtm--1 request

natew (approved):
agree w @JFrankfurt otherwise lgtm

JFrankfurt (approved):

PR #7325, opened 2023-09-16
build: improve tree-shaking

zzmp:
## Description

Reduces the initial payload size by eliminating 0.1 MiB of dead code and, more importantly, delivers the payload in 3 assets (node_modules, source, and css). Currently, the payload is delivered in 8 assets, as node_modules is split amongst 6 files. This improves pageload time by allowing all of the node_modules scripts to be parsed/evaluated at once, shaving off 100ms in local testing.

This is done by compressing the code (using `terser`) before tree-shaking, so that dead code is exposed by the compression (which will remove unused usages of imports), as described in [Extreme Dead Code Elimination](https://medium.com/engineering-housing/dead-code-elimination-and-tree-shaking-at-housing-part-1-307a94b30f23#7e03).

Also adds back CSS compression by removing the custom minimizers and using those provided by create-react-app.

## Tree maps

|    Current   |   With improved tree-shaking    |
|  |  |

### DIFF
--- a/craco.config.cjs
+++ b/craco.config.cjs
@@ -5,7 +5,6 @@ const { execSync } = require('child_process')
 const MiniCssExtractPlugin = require('mini-css-extract-plugin')
 const path = require('path')
 const ModuleScopePlugin = require('react-dev-utils/ModuleScopePlugin')
-const TerserPlugin = require('terser-webpack-plugin')
 const { IgnorePlugin, ProvidePlugin } = require('webpack')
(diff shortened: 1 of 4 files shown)


JFrankfurt (changes requested):
Reading the linked medium post it looks like you're supposed to run terser two times in order to achieve the desired effect. 
image

I don't see a second run happening here--is it missing?

tinaszheng (approved):

PR #7328, opened 2023-09-18
feat: show usd price with no input amount

### LINKED ISSUE #1712: Show spot price on token select

just-toby:
## Description

fixes #1712 by defaulting to an input amount of "1" when calculating the USD value

_Linear ticket:_ https://github.com/Uniswap/interface/issues/1712

### Before
image

### After

image

## Test plan

Screenshot 2023-09-18 at 10 11 57 AM


src/pages/Swap/index.tsx
```diff
@@ -314,10 +314,20 @@ export function Swap({
...
-  const fiatValueInput = useUSDPrice(parsedAmounts[Field.INPUT], currencies[Field.INPUT] ?? undefined)
-  const fiatValueOutput = useUSDPrice(parsedAmounts[Field.OUTPUT], currencies[Field.OUTPUT] ?? undefined)
-  const showFiatValueInput = Boolean(parsedAmounts[Field.INPUT])
-  const showFiatValueOutput = Boolean(parsedAmounts[Field.OUTPUT])
+  const fiatValueInput = useUSDPrice(
+    parsedAmounts[Field.INPUT] ??
+      (currencies[Field.INPUT]
+        ? CurrencyAmount.fromRawAmount(currencies[Field.INPUT], JSBI.BigInt(10 ** currencies[Field.INPUT].decimals))
+        : undefined),
+    currencies[Field.INPUT] ?? undefined
```
natew, line 322: `?? undefined` i think that can just be removed

natew: unless currencies can return `null` and it only wants an undefined type there which would be odd

natew (changes requested):
see comment

src/pages/Swap/index.tsx
```diff
@@ -314,10 +314,20 @@ export function Swap({
...
+        : undefined),
+    currencies[Field.INPUT]
+  )
+  const fiatValueOutput = useUSDPrice(
+    parsedAmounts[Field.OUTPUT] ??
+      (currencies[Field.OUTPUT]
+        ? CurrencyAmount.fromRawAmount(currencies[Field.OUTPUT], JSBI.BigInt(10 ** currencies[Field.OUTPUT].decimals))
+        : undefined),
+    currencies[Field.OUTPUT]
+  )
```
JFrankfurt, line 330: this is muuuuch harder to parse quickly. Can you find a more readable way to break out the same logic?

JFrankfurt (approved):

src/state/swap/hooks.tsx
```diff
@@ -77,7 +77,7 @@ const BAD_RECIPIENT_ADDRESSES: { [address: string]: true } = {
 }
 
 export type SwapInfo = {
-  currencies: { [field in Field]?: Currency | null }
+  currencies: { [field in Field]?: Currency | undefined }
```
natew, line 80: `?:` already means `| undefined` so no need for `| undefined`

src/lib/hooks/useCurrency.ts
```diff
@@ -116,7 +116,7 @@ export function useCurrencyFromMap(
 
   const token = useTokenFromMapOrNetwork(tokens, isNative ? undefined : shorthandMatchAddress ?? currencyId)
 
-  if (currencyId === null || currencyId === undefined || !isSupportedChain(chainId)) return null
+  if (currencyId === null || currencyId === undefined || !isSupportedChain(chainId)) return undefined
```
natew, line 119: can just `return` as it returns undefined

natew (changes requested):
small changes

cbachmeier (approved):

PR #7329, opened 2023-09-18
fix: do not re-init active locale

zzmp:
## Description
Does not re-load and re-activate the locale if it has already been loaded.
This prevents a re-render during initialization for en-US users, improving pageload time.


JFrankfurt (approved):
great little fix

JackShort (approved):

cbachmeier (approved):

PR #7332, opened 2023-09-18
fix: wait to fetch lists until rehydration

zzmp:
## Description

Waits to fetch token lists until state has been rehydrated, as lists should only be fetched if they are not already loaded, and that is impossible to tell without hydrated state. Effectively, this improves pageload by eliminating main thread usage.

## Test plan
Manually verified that token list requests were no longer going out as part of pageload, but were deferred.


just-toby (approved):
nice

cbachmeier (approved):

PR #7334, opened 2023-09-18
feat: eagerly connect outside of react lifecycle

zzmp:
## Description

Connects to gnosis/network/selectedWallet on pageload, instead of waiting for react to bootstrap and the state to rehydrate. selectedWallet is not available until the state is hydrated, which requires react to bootstrap, so it is made available by duplicating it in localStorage.

Effectively, this both speeds up time to connect (to selectedWallet, and to gnosis/network), **and** fixes the jitter that occurred when loading the landing page, because the landing page could only redirect (knowing that there is a selectedWallet) _after_ the state was rehydrated.

Also adds performance measures for `web3:connect`, so that connection timing is apparent in performance profiles.

### Automated testing

- [X] Integration/E2E test


src/connection/eagerlyConnect.ts
```diff
@@ -0,0 +1,36 @@
...
+  } catch (error) {
+    console.debug(`web3-react eager connection error: ${error}`)
+    return false
+  } finally {
+    performance.measure(`web3:connect:${type}`, `web3:connect:${type}:start`)
+  }
+}
+
+connect(gnosisSafeConnection.connector, ConnectionType.GNOSIS_SAFE)
+connect(networkConnection.connector, ConnectionType.NETWORK)
```
just-toby, line 24: is it ok to call these together? will networkConnection overwrite the gnosis connection?

just-toby: nvm, i see this is just copied from `useEagerlyConnect`. if it was working before then it should be good now.

just-toby (approved):
this addresses [this linear ticket](https://linear.app/uniswap/issue/WEB-2706/investigate-moving-persisted-wallet-type-to-jotaiwithstorage-for). thanks!

src/connection/eagerlyConnect.ts
zzmp: they'll both attempt to connect. if both succeed (which is possible), web3-react will use the one that comes first in the ordered list, which should be gnosis safe. as you said, this is code that's already in there.

JFrankfurt (approved):

src/connection/eagerlyConnect.ts
```diff
@@ -0,0 +1,36 @@
+import { Connector } from '@web3-react/types'
+
+import { getConnection, gnosisSafeConnection, networkConnection } from './index'
+import { ConnectionType, selectedWalletKey, toConnectionType } from './types'
+
+async function connect(connector: Connector, type: ConnectionType) {
+  performance.mark(`web3:connect:${type}:start`)
+  try {
+    if (connector.connectEagerly) {
```
natew, line 9: i **think** could be fancy with `await (connector.connectEagerly?.() ?? connector.activate())`

src/connection/types.ts
```diff
@@ -11,6 +13,14 @@ export enum ConnectionType {
   GNOSIS_SAFE = 'GNOSIS_SAFE',
 }
 
+export function toConnectionType(value = ''): ConnectionType | undefined {
+  if (Object.keys(ConnectionType).includes(value)) {
```
natew, line 17: `value in ConnectionType`?

natew (approved):
two nit but not necessary

src/connection/types.ts
zzmp: this is maybe a utility we could write? but there's a lot of this pattern in the codebase. I *do* like your pattern better, though.

src/connection/eagerlyConnect.ts
zzmp: connectEagerly returns *something* asynchronously, and I'm not sure what it is. But I *think* this would work.
Still, KISS.

PR #7336, opened 2023-09-18
chore: updating input currency panel to handle comma separator

JackShort:
* updates swap currency input to use commas for decimals based of the locale
* note this is calling a hook that returns the default locale if the `currencyConversion` flag isn't enabled so you have to enable the feature flag to test

### Before
|   Desktop    |
| Screen Shot 2023-09-18 at 3 55 14 PM|

### After
|    Desktop   |
| Screen Shot 2023-09-18 at 3 55 38 PM |


JackShort (commented):
note: when value is actually processed the commas are replaced to periods no matter what

src/components/NumericalInput/index.tsx
```diff
@@ -39,6 +41,13 @@ const StyledInput = styled.input<{ error?: boolean; fontSize?: string; align?: s
   }
 `
 
+function localeUsesComma(locale: SupportedLocale): boolean {
+  const n = 1.1
```
just-toby, line 45: no need for a variable here

just-toby (approved):
nice!

src/components/NumericalInput/index.tsx
```diff
@@ -39,6 +41,13 @@ const StyledInput = styled.input<{ error?: boolean; fontSize?: string; align?: s
...
+  const decimalSeparator = new Intl.NumberFormat(locale).format(n)[1]
```
zzmp, line 46: What a great hack. Thanks, ECMAScript.

```suggestion
  const decimalSeparator = new Intl.NumberFormat(locale).format(1.1)[1]
```

zzmp (approved):
lgtm but please fix the nit

src/components/NumericalInput/index.tsx
justindomingue: should we document this hack in code comments?

just-toby (approved):

PR #7337, opened 2023-09-18
feat: account suspense

zzmp:
## Description

Shows a connecting state for returning users who have a wallet that should reconnect. This state is based on their last displayed name (address or ENS name), and will persist until the connection is re-established (including ENS, if applicable) or fails.

## Screen capture

https://github.com/Uniswap/interface/assets/5403956/b70d26fe-9cfb-4b20-b909-ecc98965090c

## Test plan
Manually tested with a selectedWallet and without.


src/components/Web3Status/index.tsx
```diff
@@ -128,7 +129,10 @@ const StyledConnectButton = styled.button`
 function Web3StatusInner() {
   const switchingChain = useAppSelector((state) => state.wallets.switchingChain)
   const ignoreWhileSwitchingChain = useCallback(() => !switchingChain, [switchingChain])
-  const { account, connector } = useLast(useWeb3React(), ignoreWhileSwitchingChain)
+  const connectionReady = useConnectionReady()
+  const activeWeb3 = useWeb3React()
+  const lastWeb3 = useLast(useWeb3React(), ignoreWhileSwitchingChain)
```
just-toby, line 134: i know you didn't add `useLast` but the name is a little confusing. i think `useMostRecent` would make more sense

src/connection/eagerlyConnect.ts
```diff
@@ -1,8 +1,23 @@
...
+    (onStoreChange) => {
+      if (connectionReady instanceof Promise) {
+        connectionReady.finally(onStoreChange)
+      }
+      return () => undefined
+    },
+    () => connectionReady === true
+  )
+}
+
```
just-toby, line 20: this is an interesting pattern. is this only needed because the connected state is updating outside the react lifecycle?

just-toby (approved):

src/connection/eagerlyConnect.ts
zzmp: yes, exactly that. `useSyncExternalStore` let's the react lifecycle subscribe to external updates and trigger renders accordingly.

src/components/Web3Status/index.tsx
zzmp: I'd support auditing `useLast` and `usePrevious` for both performance and naming. I think they each only have a couple of uses.

src/components/Web3Status/index.tsx
```diff
@@ -173,6 +177,12 @@ function Web3StatusInner() {
         </Web3StatusConnected>
       </TraceEvent>
     )
+  } else if (!connectionReady) {
+    return (
+      <Web3StatusConnected disabled={true}>
+        <LoaderV3 size="24px" />
```
cbachmeier, line 183: this works well for mobile but we should consider using a pill for screenWidths > 1100px to prevent the trashing of going from the small circular loader to the full wallet pill
thrash from the spinner to full account pill
thrashCircle


Loader pills used on token explore for reference
Screenshot 2023-09-20 at 9 38 21 AM

src/connection/eagerlyConnect.ts
```diff
@@ -26,11 +41,15 @@ const selectedWallet = toConnectionType(localStorage.getItem(selectedWalletKey)
 if (selectedWallet) {
   const selectedConnection = getConnection(selectedWallet)
   if (selectedConnection) {
-    connect(selectedConnection.connector, selectedWallet).then((connected) => {
-      if (!connected) {
-        // only clear the persisted wallet type if it failed to connect.
-        localStorage.removeItem(selectedWalletKey)
-      }
-    })
+    connectionReady = connect(selectedConnection.connector, selectedWallet)
```
cbachmeier, line 44: do you need to set `connectionReady = Promise<void>`? Could you instead `let connectionReady = false` and then just update it to `true` on a successful connect? That would simplify the `if` statement in `useConnectionReady`

src/components/Web3Status/index.tsx
zzmp: A loader pill would not eliminate layout shift / jank / thrashing, because the address is not displayed in a monospaced font, and the width is unknown until connection occurs. we can "cheat", but we can't necessarily get it exactly right.

zzmp: added a "cheat", lmk what you think.

src/components/Web3Status/index.tsx
```diff
@@ -146,6 +146,15 @@ function Web3StatusInner() {
 
   const { hasPendingActivity, pendingActivityCount } = usePendingActivity()
 
+  const displayName = useMemo(() => (account ? ENSName || shortenAddress(account) : null), [account, ENSName])
+  const lastDisplayName = useRef(localStorage.getItem(selectedWalletDisplayKey) ?? null)
+  useEffect(() => {
+    if (displayName) {
+      // It takes a moment to load the ENSName, so set the lastDisplayName (for next time) to the account to reduce layout shift.
+      localStorage.setItem(selectedWalletDisplayKey, shortenAddress(account))
```
cbachmeier, line 154: why is lastDisplayName set to `shortenAddress(account)` and not `displayName`?

src/components/Web3Status/index.tsx
```diff
@@ -181,7 +190,13 @@ function Web3StatusInner() {
   } else if (!connectionReady) {
     return (
       <Web3StatusConnected disabled={true}>
-        <LoaderV3 size="24px" />
+        <IconWrapper size={24}>
+          <LoaderV3 size="24px" />
+        </IconWrapper>
+        <AddressAndChevronContainer>
+          {/* Render the last known wallet address to prevent a layout shift on initial connection. */}
+          <Text style={{ visibility: 'hidden' }}>{lastDisplayName.current}</Text>
```
cbachmeier, line 198: Definitely hacky but effective and looks a lot cleaner loading

src/connection/eagerlyConnect.ts
```diff
@@ -43,9 +45,14 @@ if (selectedWallet) {
...
-          // only clear the persisted wallet type if it failed to connect.
-          localStorage.removeItem(selectedWalletKey)
+        if (!connected) throw new FailedToConnect()
+      })
+      .catch((error) => {
+        // Clear the persisted wallet type if it failed to connect.
+        localStorage.removeItem(selectedWalletKey)
+        // Log it if it threw an unknown error.
+        if (!(error instanceof FailedToConnect)) {
+          console.error(error)
```
cbachmeier, line 55: do we do anything with console logged errors? Should we instead log it to Sentry?

cbachmeier (approved):

src/components/Web3Status/index.tsx
zzmp: Because the account address will load before ens. We could tighten this up more but this was a first attempt. 

See inline comment above this line.

src/connection/eagerlyConnect.ts
zzmp: A user may just be denying a connection. I think this is the same behavior as before.

just-toby (dismissed):

src/components/Web3Status/index.tsx
```diff
@@ -147,13 +148,36 @@ function Web3StatusInner() {
...
+        type: connection.type,
+        display: displayName,
+        displayType: ENSName ? ConnectionDisplayType.ENSName : ConnectionDisplayType.Address,
+      }
+      setConnectionMeta(meta)
     }
-  }, [account, displayName])
+  }, [ENSName, account, connection.type, displayName])
+  const isDisplayNameLoading = Boolean(
+    account && meta.current?.displayType === ConnectionDisplayType.ENSName && ENSLoading
```
cbachmeier: Is there a situation where we would be loading an ENSName without an `account` or `meta.current?.displayType !== ConnectionDisplayType.ENSName`?

cypress/utils/user-state.ts
```diff
@@ -13,7 +14,12 @@ export const DISCONNECTED_WALLET_USER_STATE: Partial<UserState> = { selectedWall
 export function setInitialUserState(win: Cypress.AUTWindow, state: UserState) {
   // Selected wallet should also be reflected in localStorage, so that eager connections work.
   if (state.selectedWallet) {
-    win.localStorage.setItem('selected_wallet', state.selectedWallet)
+    win.localStorage.setItem(
```
cbachmeier, line 17: could we use the `setConnectionMeta` fn here?

zzmp: we can't without passing the window object, which complicates the setter. I chose to just export the key, rather than make the setter more complex.

src/components/Web3Status/index.tsx
zzmp: not that i can think of, no

cbachmeier (approved):

src/components/Web3Status/index.tsx
```diff
@@ -141,6 +153,37 @@ function Web3StatusInner() {
 
   const { hasPendingActivity, pendingActivityCount } = usePendingActivity()
 
+  const displayName = useMemo(() => (account ? ENSName || shortenAddress(account) : null), [account, ENSName])
```
just-toby, line 156: i think we usually use `??` instead of `||`

src/components/Web3Status/index.tsx
```diff
@@ -141,6 +153,37 @@ function Web3StatusInner() {
...
+  const meta = useRef(getConnectionMeta())
```
just-toby, line 157: using a ref for this makes the flow hard to follow. maybe if you wrapped it in a `useCurrentMeta` hook which handled updating when the connection loads, this file would be easier to follow

src/components/Web3Status/index.tsx
just-toby: also we usually use `undefined` instead of `null` now

src/components/Web3Status/index.tsx
```diff
@@ -141,6 +153,37 @@ function Web3StatusInner() {
...
+      const meta: ConnectionMeta = {
+        type: connection.type,
+        address: account,
+        ENSName: ENSName ?? undefined,
+      }
+      setConnectionMeta(meta)
+    }
+  }, [ENSName, account, connection.type, displayName])
+
+  const isENSNameLoading = Boolean(meta.current?.address === account && meta.current?.ENSName && ENSLoading)
```
just-toby, line 169: this variable name is also confusing to me, because it includes more than just the loading state. 

is `meta` just supposed to be the cached wallet information? if you can clarify that with a couple comments or name changes then i think this file will be alot easier to follow

just-toby (changes requested):

just-toby (approved):

cbachmeier (approved):

PR #7338, opened 2023-09-18
fix: quote price in SwapDetailsDropdown should match execution price

### LINKED ISSUE #2648: price in price bar should be the execution price, not price of 1 base
image
example on arbitrum, selling 35k eth but it says 1 eth = 3800 dai even though there's huge price impact

Jdecristi:
fixes #2648 

## Description
The price in SwapDetailsDropdown should match the execution price.

## Screen capture

|    Before   |   After    |
|Screenshot 2023-09-18 at 4 06 23 PM|Screenshot 2023-09-18 at 4 04 51 PM|

### Reproducing the error

1.  Navigate to [uniswap](app.uniswap.org/swap)
2. Enter in a huge number like 1000000
3. Select a output token
4. Look at the ETH to output token ratio

### DIFF
--- a/src/components/swap/TradePrice.tsx
+++ b/src/components/swap/TradePrice.tsx
@@ -32,8 +32,8 @@ export default function TradePrice({ price }: TradePriceProps) {
 
   const [showInverted, setShowInverted] = useState<boolean>(false)
 
-  const { baseCurrency, quoteCurrency } = price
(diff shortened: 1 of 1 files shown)


tinaszheng:
hey! looks like there is a bug since 1 ETH != $15 but this isn't the right fix - the USD price should still be based on whatever's currency on the left - in this case 1 ETH is still ~$1600, so closing this for now, sorry!

PR #7340, opened 2023-09-18
fix: preconnect

zzmp:
## Description

Updates preconnect to the bare subdomains for `api.uniswap.org` and `mainnet.infura.io`, to more quickly load quotes once requested.


public/index.html
```diff
@@ -36,7 +36,8 @@
     -->
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
 
-    <link rel="preconnect" href="%REACT_APP_AMPLITUDE_PROXY_URL%" />
```
just-toby, line 39: why remove amplitude though?

just-toby (changes requested):

public/index.html
just-toby: nevermind, i see that the proxy live at the same domain+subdomain as the quote API

just-toby: maybe comment that is intended to catch  quotes, proxies etc under the umbrella

just-toby (approved):

natew (approved):

PR #7345, opened 2023-09-19
fix: reset LDO approvals

just-toby:
## Description

LDO token has the same `approve` logic as USDT on mainnet which requires you to reset the allowance to zero before increasing the amount.

see the contract source code [here](https://etherscan.io/token/0x5a98fcbea516cf06857215779fd812ca3bef1b32#code)

```
function approve(address _spender, uint256 _amount) public returns (bool success) {
// ........
        // To change the approve amount you first have to reduce the addresses`
        //  allowance to zero by calling `approve(_spender,0)` if it is not
        //  already 0 to mitigate the race condition described here:
        //  https://github.com/ethereum/EIPs/issues/20#issuecomment-263524729
        require((_amount == 0) || (allowed[msg.sender][_spender] == 0));
// ........
}
```

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2857/generalize-reset-usdt-approval-step-to-work-for-ldo-too

## Screen capture

https://github.com/Uniswap/interface/assets/66155195/a77a1dbf-5f1c-4d3f-8700-44bfa3046fe1

### Reproducing the error

1.  set your LDO token allowance to some amount greater than zero, but smaller than the amount you want to trade.
2. try to trade the higher amount and notice that the approval step fails.

### Automated testing

- [x] Integration/E2E test - existing tests for the USDT flow

### DIFF
--- a/src/components/swap/ConfirmSwapModal.tsx
+++ b/src/components/swap/ConfirmSwapModal.tsx
@@ -12,7 +12,7 @@ import Badge from 'components/Badge'
 import Modal, { MODAL_TRANSITION_DURATION } from 'components/Modal'
 import { RowFixed } from 'components/Row'
 import { getChainInfo } from 'constants/chainInfo'
-import { USDT as USDT_MAINNET } from 'constants/tokens'
+import { LDO, USDT as USDT_MAINNET } from 'constants/tokens'
 import { TransactionStatus } from 'graphql/data/__generated__/types-and-hooks'
 import { useMaxAmountIn } from 'hooks/useMaxAmountIn'
 import { Allowance, AllowanceState } from 'hooks/usePermit2Allowance'
@@ -43,7 +43,7 @@ import SwapModalHeader from './SwapModalHeader'
 export enum ConfirmModalState {
   REVIEWING,
   WRAPPING,
-  RESETTING_USDT,
+  RESETTING_TOKEN,
   APPROVING_TOKEN,
   PERMITTING,
   PENDING_CONFIRMATION,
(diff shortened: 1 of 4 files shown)


src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -58,9 +58,13 @@ const StyledL2Logo = styled.img`
   width: 16px;
 `
 
+// Any existing USDT or LDO allowance needs to be reset before we can approve the new amount (mainnet only).
+// See the `approve` function here: https://etherscan.io/address/0xdAC17F958D2ee523a2206206994597C13D831ec7#code
+const RESET_APPROVAL_TOKENS = [USDT_MAINNET, LDO]
```
kristiehuang, line 63: should this be in a constants file instead?

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -43,7 +43,7 @@ import SwapModalHeader from './SwapModalHeader'
 export enum ConfirmModalState {
   REVIEWING,
   WRAPPING,
-  RESETTING_USDT,
+  RESETTING_TOKEN,
```
kristiehuang, line 46: nit: would resetting_token_allowance make more sense?

just-toby: it woudl

src/components/swap/ConfirmSwapModal.tsx
just-toby: it should

cbachmeier (approved):

kristiehuang (approved):

src/components/swap/constants.ts
```diff
@@ -0,0 +1,5 @@
+import { LDO, USDT as USDT_MAINNET } from 'constants/tokens'
+
+// Any existing USDT or LDO allowance needs to be reset before we can approve the new amount (mainnet only).
+// See the `approve` function here: https://etherscan.io/address/0xdAC17F958D2ee523a2206206994597C13D831ec7#code
+export const RESET_APPROVAL_TOKENS = [USDT_MAINNET, LDO]
```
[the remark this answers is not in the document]
just-toby: right now it's specific to the `ConfirmSwapModal` and only used there, so i think it's ok to colocate it here

PR #7347, opened 2023-09-19
feat: Condense color extraction logic and improve fallback

cbachmeier:
## Description

- Move `getColor` logic and consts from `functions` to `src`
- Update `useColor` to use `getColor` fn
- Add `useTokenLogoSource` as the primary token logo source since it is 1st party which avoids CORS issues which would cause old implementation to fail and use default color
- Simplify fallback logic

_Slack thread:_ https://uniswapteam.slack.com/archives/C02T729PMQE/p1695139683408519

### Before
Screenshot 2023-09-19 at 11 14 52 AM
Screenshot 2023-09-19 at 11 15 14 AM
Screenshot 2023-09-19 at 11 15 27 AM
Screenshot 2023-09-19 at 11 15 45 AM
Screenshot 2023-09-19 at 11 16 01 AM

### After
Screenshot 2023-09-19 at 10 52 29 AM
Screenshot 2023-09-19 at 10 52 03 AM
Screenshot 2023-09-19 at 10 52 43 AM
Screenshot 2023-09-19 at 11 05 57 AM
Screenshot 2023-09-19 at 11 06 15 AM

### QA (ie manual testing)

- [x] Tested loading several token pairs on the Add Liquidity page on the Vercel preview (localhost won't show CORS issues)

### Automated testing

- [ ] getColor unit tests pass


natew (approved):

src/hooks/useColor.ts
```diff
@@ -1,80 +1,67 @@
...
-  let detectedHex = palette.Vibrant.hex
-  let AAscore = hex(detectedHex, '#FFF')
-  while (AAscore < 3) {
-    detectedHex = shade(0.005, detectedHex)
-    AAscore = hex(detectedHex, '#FFF')
-  }
-
-  return detectedHex
+function convertColorArrayToString(color: number[]): string {
+  return rgb({ red: color[0], green: color[1], blue: color[2] })
```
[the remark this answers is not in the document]
cbachmeier: can confirm would look nice

functions/utils/getColor.test.ts
```diff
@@ -1,5 +1,5 @@
-import { DEFAULT_COLOR } from '../constants'
-import getColor from './getColor'
+import { DEFAULT_COLOR } from '../../src/constants/tokenColors'
```
zzmp, line 1: This should be moved to src/utils/getColor.test.ts

src/constants/tokenColors.ts
```diff
@@ -0,0 +1,73 @@
+export const DEFAULT_COLOR = [35, 43, 43]
+
+export const predefinedTokenColors: { [key: string]: number[] } = {
```
zzmp, line 3: I think this would be nicer if you find a way to use the address as opposed to the url.

src/utils/getColor.ts
```diff
@@ -2,9 +2,9 @@ import { Buffer } from 'buffer'
 import JPEG from 'jpeg-js'
 import PNG from 'png-ts'
 
-import { DEFAULT_COLOR, predefinedTokenColors } from '../constants'
+import { DEFAULT_COLOR, predefinedTokenColors } from '../constants/tokenColors'
 
-export default async function getColor(image: string | undefined, checkDistance = false) {
+export async function getColor(image: string | undefined, checkDistance = false) {
```
zzmp, line 7: why do you hate my default exports? whyyyy

zzmp (changes requested):
just move the tests and lgtm

src/constants/tokenColors.ts
cbachmeier: I looked into that for a little but you get into an issue with xchain tokens that have different address on different chains. They are currently covered by the current solution as they can have the same url for the image such as the coingecko url regardless of what chain they're on

src/utils/getColor.ts
cbachmeier: I generally prefer named exports as I find it easier to refactor them in a project.

zzmp (approved):

JFrankfurt (approved):

PR #7350, opened 2023-09-19
feat: log chain changed

just-toby:
## Description

1. log CHAIN_CHANGED event when the chain changes, no matter how the change was initiated
2. log element_clicked on chain selectors in the app's menu

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2660/log-chain-changed-events

## Screen capture

CHAIN_CHANGED  when initiated from wallet:
https://github.com/Uniswap/interface/assets/66155195/0fa18d86-9e3d-4b44-b270-810686fda845

element_clicked (AND chain_changed) when switching chains from the app:
Screenshot 2023-09-19 at 2 00 32 PM

### QA (ie manual testing)

- [x] manually verified the logging works as expected.


src/components/Web3Provider/index.tsx
```diff
@@ -46,7 +47,21 @@ function Updater() {
       provider?.off('debug', trace)
       networkProvider?.off('debug', trace)
     }
-  }, [networkProvider, provider, shouldTrace])
+  }, [analyticsContext, networkProvider, provider, shouldTrace])
+
+  const previousConnectedChainId = usePrevious(chainId)
+  useEffect(() => {
+    const chainChanged = previousConnectedChainId && previousConnectedChainId !== chainId
```
kristiehuang, line 54: do we want to track the initial chain that a user connects with on the first page load? Or only on chain change

kristiehuang (approved):

src/components/Web3Provider/index.tsx
just-toby: i think we track that as a user property, and also as a field on the wallet_connected event. so no need to emit this event on page load

src/components/NavBar/ChainSelectorRow.test.tsx
```diff
@@ -0,0 +1,43 @@
+import { ChainId } from '@uniswap/sdk-core'
+import { render } from 'test-utils/render'
+
+import ChainSelectorRow from './ChainSelectorRow'
+
+describe('ChainSelectorRow', () => {
+  it('should match snapshot', () => {
+    const { container } = render(
+      <ChainSelectorRow targetChain={ChainId.OPTIMISM} onSelectChain={jest.fn()} isPending={false} disabled={false} />
```
cbachmeier, line 9: nit: would be good to iterate through all possible targetChains for snapshot matching instead of only Optimism

cbachmeier (approved):
LGTM tested with switching via wallet, query param, and the in-app selector

src/components/Web3Provider/index.tsx
cbachmeier: might be worth adding `previousConnectedChainId` to the event?

cbachmeier (approved):

PR #7352, opened 2023-09-19
fix: ui fixes on the add liquidity flow

just-toby:
## Description

fixes some visual bugs in the LP flow:
- make the token select buttons fill their parent container
- align the items in the header
- removed an unnecessary div and used `gap` instead for spacing the token select elements

the bug described in the ticket isn't happening for me.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2803/clicking-preview-when-creating-an-lp-position-requires-two-clicks

### Before

image

### After

Screenshot 2023-09-19 at 2 40 37 PM

### Reproducing the error

1. go through the add liquidity flow and use eyes

### QA (ie manual testing)

- [x] visual inspection

### DIFF
--- a/src/components/NavigationTabs/index.tsx
+++ b/src/components/NavigationTabs/index.tsx
@@ -21,6 +21,8 @@ const Tabs = styled.div`
 
 const StyledLink = styled(Link)<{ flex?: string }>`
   flex: ${({ flex }) => flex ?? 'none'};
+  display: flex;
+  align-items: center;
 
   ${({ theme }) => theme.deprecated_mediaWidth.deprecated_upToMedium`
     flex: none;
@@ -82,7 +84,7 @@ export function AddRemoveTabs({
 
   return (
     <Tabs>
-      <RowBetween style={{ padding: '1rem 1rem 0 1rem' }}>
+      <RowBetween style={{ padding: '1rem 1rem 0 1rem' }} align="center">
         <StyledLink
           to={poolLink}
           onClick={() => {
--- a/src/components/PrefetchBalancesWrapper/PrefetchBalancesWrapper.tsx
+++ b/src/components/PrefetchBalancesWrapper/PrefetchBalancesWrapper.tsx
@@ -30,7 +30,8 @@ const hasUnfetchedBalancesAtom = atom<boolean>(true)
 export default function PrefetchBalancesWrapper({
   children,
   shouldFetchOnAccountUpdate,
-}: PropsWithChildren<{ shouldFetchOnAccountUpdate: boolean }>) {
+  className,
+}: PropsWithChildren<{ shouldFetchOnAccountUpdate: boolean; className?: string }>) {
   const { account } = useWeb3React()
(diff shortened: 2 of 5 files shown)


src/pages/AddLiquidity/styled.tsx
```diff
@@ -32,10 +31,6 @@ export const DynamicSection = styled(AutoColumn)<{ disabled?: boolean }>`
   pointer-events: ${({ disabled }) => (disabled ? 'none' : 'initial')};
 `
 
-export const CurrencyDropdown = styled(CurrencyInputPanel)`
```
just-toby, line 35: we don't need this - see the `gap` i added above

src/components/PrefetchBalancesWrapper/PrefetchBalancesWrapper.tsx
```diff
@@ -30,7 +30,8 @@ const hasUnfetchedBalancesAtom = atom<boolean>(true)
 export default function PrefetchBalancesWrapper({
   children,
   shouldFetchOnAccountUpdate,
-}: PropsWithChildren<{ shouldFetchOnAccountUpdate: boolean }>) {
+  className,
+}: PropsWithChildren<{ shouldFetchOnAccountUpdate: boolean; className?: string }>) {
```
gnewfield, line 34: Why is classname needed?

just-toby: this lets us make a styled-components wrapper in the other file

kristiehuang:
I'm not able to repro the token select buttons filling their parent containers bug - here's what it looks like on app.uniswap.org rn for me
Screenshot 2023-09-21 at 3 57 48 PM

just-toby:
@kristiehuang  pretty sure i introduced the bug just 3 days ago: https://github.com/Uniswap/interface/commit/cf09e80934afec82f077fef30ec846b02ec15149

src/components/NavigationTabs/index.tsx
```diff
@@ -82,7 +84,7 @@ export function AddRemoveTabs({
 
   return (
     <Tabs>
-      <RowBetween style={{ padding: '1rem 1rem 0 1rem' }}>
+      <RowBetween style={{ padding: '1rem 1rem 0 1rem' }} align="center">
```
kristiehuang, line 87: it may look misaligned bc Settings were removed -- should look proper once settings are back?

src/components/PrefetchBalancesWrapper/PrefetchBalancesWrapper.tsx
zzmp: `className` is what styled components use to target changes, so if you don't pass it through then you can't styled-componentize the component (eg `const StyledWrapper = styled(PrefetchBalancesWrapper)`)

zzmp (approved):

kristiehuang (approved):

PR #7310, opened 2023-09-13
feat: [info] Migrate Historical Pool Data Queries

cbachmeier:
## Description

The purpose of this PR was to be able to match the data shown on the info site for a given pool to what we want to show on the interface on the new Pool Details Page. This required:
- Adding apollo clients for subgraph queries for block numbers
- Add queries and helper functions to get block numbers based on timestamps
- Update PoolData query to pull current, 24H, 48H, and 1Week historical data
- Use the retrieved data to format and calculate stats to match what is shown on the info site
- Also Base was missing a supgraph URL so I added it
- Update tests
- **Note:** Much of this code came directly from the info site
  - https://github.com/Uniswap/v3-info/blob/master/src/data/pools/poolData.ts
  - https://github.com/Uniswap/v3-info/blob/master/src/utils/data.ts
  - https://github.com/Uniswap/v3-info/blob/master/src/utils/queries.ts
  - https://github.com/Uniswap/v3-info/blob/master/src/hooks/useBlocksFromTimestamps.ts
  - https://github.com/Uniswap/v3-info/blob/master/src/constants/index.ts#L52

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2853/query-subgraph-for-blocks-based-on-timestamps https://linear.app/uniswap/issue/WEB-2854/query-subgraph-for-current-24h-48h-and-1week-historical-data

### QA (ie manual testing)

- [x] I tested this by going to such as `/pools/polygon/0x167384319b41f7094e62f7506409eb38079abff8` and making sure the requests to both the normal subgraph url as well as the newly added blocks url returned the correct data
- [x] I manually inspected the returned `poolData` from the `usePoolData` hook for a given pool and made sure the stats such as total and token TVL matched up to what is shown on the info site. For example https://info.uniswap.org/#/polygon/pools/0x167384319b41f7094e62f7506409eb38079abff8
Screenshot 2023-09-13 at 3 27 14 PM

Screenshot 2023-09-13 at 3 27 23 PM

### Automated testing

- [x] Updated unit tests and snapshots

PR #7353, opened 2023-09-19
feat: [info] Add Stats Section to PDP

cbachmeier:
## Description

- Adds Stats Section for Pool Details Page
- Show TVL, 24H Volume, 24H fees, and a graph of the pool's balance between the two tokens
- Supports Mobile
- Leverages #7310 and #7347 for data and color extraction
- Add tests for new section

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2736/stats

## Screen capture
Desktop
Screenshot 2023-09-19 at 1 53 19 PM
Mobile
Screenshot 2023-09-19 at 3 42 08 PM


src/pages/PoolDetails/PoolDetailsStats.tsx
```diff
@@ -0,0 +1,229 @@
...
+  isReversed: boolean
+  chainId?: number
+}
+
+export function PoolDetailsStats({ poolData, isReversed, chainId }: PoolDetailsStatsProps) {
+  const isScreenSize = useScreenSize()
+  const screenIsNotLarge = isScreenSize['lg']
+  const { formatNumber } = useFormatter()
+
+  // We can't wrap this in a useMemo hook because useCurrency is also a hook
```
justindomingue, line 103: is there a reason you can't do this?

```
const currencyA = useCurrency(poolData?.token0?.id, chainId) ?? undefined
const currencyB = useCurrency(poolData?.token1?.id, chainId) ?? undefined
const currencies = useMemo(() => ([currencyA, currencyB]), [currencyA, currencyB])
```

justindomingue: but honestly i don't think you even need to wrap in an array at all here given you access by index later anyway?

src/pages/PoolDetails/PoolDetailsStats.tsx
```diff
@@ -0,0 +1,229 @@
...
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
justindomingue, line 104: is this eslint-disable used?

src/pages/PoolDetails/PoolDetailsStats.tsx
```diff
@@ -0,0 +1,229 @@
...
+  const currencies = [
+    useCurrency(poolData?.token0?.id, chainId) ?? undefined,
+    useCurrency(poolData?.token1?.id, chainId) ?? undefined,
+  ]
+
+  // We can't wrap this in a useMemo hook because useColor is also a hook
```
justindomingue, line 110: ditto, don't think this comment is really necessary

src/pages/PoolDetails/PoolDetailsStats.tsx
```diff
@@ -0,0 +1,229 @@
...
+      <HeaderText>
+        <Trans>Stats</Trans>
+      </HeaderText>
+      <StatItemColumn>
+        <ThemedText.BodySecondary>
+          <Trans>Pool balances</Trans>
+        </ThemedText.BodySecondary>
+        <PoolBalanceSymbols>
+          <PoolBalanceTokenNames>
+            {!screenIsNotLarge && (
```
justindomingue, line 147: could you use one of the higher order components for this ? https://github.com/Uniswap/interface/blob/main/src/theme/components/index.tsx#L457

src/pages/PoolDetails/PoolDetailsStats.tsx
```diff
@@ -0,0 +1,229 @@
...
+import { useColor } from 'hooks/useColor'
+import { useScreenSize } from 'hooks/useScreenSize'
+import { ReactNode, useMemo } from 'react'
+import { Text } from 'rebass'
+import styled, { css } from 'styled-components'
+import { BREAKPOINTS } from 'theme'
+import { ThemedText } from 'theme/components'
+import { NumberType, useFormatter } from 'utils/formatNumbers'
+
+const HeaderText = styled(Text)`
```
justindomingue, line 17: should we leverage `ThemedText` here instead of raw rebass https://github.com/Uniswap/interface/blob/main/src/theme/components/text.tsx#L17

src/pages/PoolDetails/PoolDetailsStats.tsx
```diff
@@ -0,0 +1,229 @@
...
+        {screenIsNotLarge && (
+          <Row data-testid="pool-balance-chart">
+            {token0.color && <BalanceChartSide percent={token0.percent} $color={token0.color} isLeft={true} />}
+            {token1.color && <BalanceChartSide percent={token1.percent} $color={token1.color} isLeft={false} />}
+          </Row>
+        )}
+      </StatItemColumn>
+      <StatItem title={<Trans>TVL</Trans>} value={poolData.tvlUSD} delta={poolData.tvlUSDChange} />
+      <StatItem title={<Trans>24H volume</Trans>} value={poolData.volumeUSD} delta={poolData.volumeUSDChange} />
+      <StatItem title={<Trans>24H fees</Trans>} value={poolData.volumeUSD * (poolData.feeTier / 1000000)} />
```
justindomingue, line 178: double-checking, should we divide by `10000` here?

src/pages/PoolDetails/PoolDetailsStats.tsx
cbachmeier: We currently don't have a ThemedText that matches this criteria. I could add it for this currently one-off?

src/pages/PoolDetails/PoolDetailsStats.tsx
cbachmeier: yes, see the above comment

src/pages/PoolDetails/PoolDetailsStats.tsx
cbachmeier: it's used to explain why the eslint was needed

src/pages/PoolDetails/PoolDetailsStats.tsx
cbachmeier: Yes to get an accurate calculation. This is the same logic currently used on the info site.

src/pages/PoolDetails/PoolDetailsStats.tsx
cbachmeier: For these two logos a new one would need to be added that's displays only on Large and above

src/pages/PoolDetails/PoolDetailsStats.tsx
cbachmeier: I stand corrected, this used to be flagged as needing to be in it's own useMemo hook, but no longer. Not sure what has changed, but I've removed it.

src/pages/PoolDetails/PoolDetailsStats.tsx
cbachmeier: This component has also only be used once, it seems like it has been nearly completely replaced with doing direct css styling or the useScreenSize hook

cbachmeier:
@justindomingue and for any other reviewer who wants to see this page. Go to the [preview](https://interface-git-cab-pdpstats-uniswap.vercel.app/swap), enable the `pool_detail_page` flag, and then you can check out this [URL](https://interface-git-cab-pdpstats-uniswap.vercel.app/pools/ethereum/0x11950d141ecb863f01007add7d1a342041227b58). To see other pools, go to the [info site](https://info.uniswap.org/#/pools)and find a pool. Grab the address and chain and just update the url.

src/pages/PoolDetails/PoolDetailsStats.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+    expect(screen.getByText(/Stats/i)).toBeInTheDocument()
+    expect(screen.getByText('90.93M USDC')).toBeInTheDocument()
+    expect(screen.getByText('82,526.49 WETH')).toBeInTheDocument()
+    expect(screen.getByText(/TVL/i)).toBeInTheDocument()
+    expect(screen.getByText('$223.2M')).toBeInTheDocument()
+    expect(screen.getByTestId('pool-balance-chart')).toBeInTheDocument()
+  })
+
+  it('pool balance chart not visible on mobile', () => {
+    Object.defineProperty(window, 'innerWidth', {
```
justindomingue, line 27: curious what you think of mocking the fn instead of updating window properties? nbd but this feels more fickle and more of an implementation detail

src/pages/PoolDetails/PoolDetailsStats.tsx
justindomingue: ah okay probably fine as is

justindomingue (approved):

src/pages/PoolDetails/PoolDetailsStats.test.tsx
cbachmeier: I could mock the `useScreenSize` hook but we also have some styling in styled component css which wouldn't be affected by the mocked behaviour so this covers both cases.

src/pages/PoolDetails/PoolDetailsStats.tsx
JFrankfurt: @infredible thoughts on one-off? From a maintenance perspective it would be great if this aligned w/ our text theme.

infredible: We shouldn't be using any one-offs on this page. This should be heading3

cbachmeier: @infredible I think heading3 might be out of date since the spore work as in the Figma its showing as fontWeight 500. Do you know if it should match up to one of [these Header styles](https://github.com/Uniswap/interface/blob/main/src/theme/components/text.tsx#L17) we currently have?
Screenshot 2023-09-27 at 11 24 19 AM

cartcrom (approved):

PR #7355, opened 2023-09-20
feat: Add Zeal Wallet integration

masm:


cartcrom:
Hi @masm, we are planning to support new injected wallets via eip6963 rather than through `window.ethereum`, so we are currently not accepting this type of PR.

To have your extension's name/icon displayed, you can implement eip6963 in your extension, and it will automatically be detected once we implement eip6963 in interface.

PR #7356, opened 2023-09-20
feat: [info] add transactions table to token details page

gnewfield:
## Description
The Dot Info project is an effort to integrate new data and visualization features into the Uniswap interface to create a trading experience that is both sophisticated and intuitive. One of the major stories is a redesign of the Token Details Page (TDP), which allows customers to view all available information related to a given token (for a given chain and contract address) and buy / sell it right there on the page.

This PR provides the initial implementation for one of the key features of the page, an Activity section that allows customers to view pool information and transaction data in a tabular format with sorting / searching capabilities. Because tables are present throughout the Dot Info project, including multiple instances on the Explore and Pool Detail Pages (PDP), the approach taken here is to create a generic table implementation that can then be easily customized for different data and contexts.

This is accomplished by reusing the `react-table` library that was used for tables within the NFT platform. This "headless UI" framework provides abstractions for the structure of a table (e.g. rows, columns, cells), as well as common business logic, right out of the box.

Using `react-table`, this PR adds a `Table` component that can be reused throughout the project. The interface for this component is simple, robust, and minimally opinionated about the table's usage. It takes in a definition of the tables columns, as well as an array of data objects that parallel the column definition and will populate the table's rows. 

With a common implementation, this PR adds a transactions table to TDP. The columns of this table (from left to right) are:

- Time, displayed in custom date string format (with link to tx page on etherscan)
- Type, i.e. Bought / Sold
- Amount, quantity of reference token (i.e. token displayed on TDP)
- For, quantity of the other token in the pair
- USD, dollar value of trade
- Maker, address of swapper (with link to address page on etherscan)

The table takes in `Swap` data objects which follow a format that roughly maps to the columns. When designing this interface, it was important to find the right balance between matching the table exactly and having a reusable data object. On one hand, it's beneficial to keep the table display logic simple, i.e. not processing raw data in the table cells. On the other hand, if the data matches the table exactly, the model can't be reused throughout the application in different contexts. The approach taken was to have the data object represent a swap conceptually. Then, a small amount of processing logic can live in the table to customize the display for the precise use-case, but the same processing logic of generic swap events can be reused in other contexts.

Follow-ups to this PR will:

- Replace the mocked swap data with a live V3 subgraph query (as used by the V3 info site)
- Add infinite scroll for loading additional data + loading skeleton
- Add a live V2 subgraph query (as used by the v2 site and adapt V2 transactions to the Swap data format)
- Add searching / sorting functionality into the table

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2745/transactions-table
_Relevant docs:_ [Figma Design](https://www.figma.com/file/2yKuZXmQ2UvgBbXR3txz4O/Dot-Info?type=design&node-id=474-691184&mode=design&t=FZtWis2hzdtbQ2xE-0)

### Before
| Before Mobile | Before Desktop |

### After
|After Mobile| After Desktop |

### QA (ie manual testing)

- [ ] Visit the preview link and enable the feature flag for info detail page
- [ ] Visit the TDP for HarryPotterObamaSonic10Inu (this is simply because the mock data is for BITCOIN swaps) `/tokens/ethereum/0x72e4f9f808c49a2a61de9c5896298920dc4eeea9`
- [ ] Verify that the table appears as expected

#### Devices

Columns are removed from the table for smaller screen sizes, so try viewing the page with multiple window sizes and verify that columns are dynamically shown / hidden as expected.

### Automated testing

- [x] Unit test


src/components/Tokens/TokenDetails/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Displays the time as a human-readable string.
+ *
+ * @param {number} timestamp - Transaction timestamp in milliseconds.
+ * @param {number} locale - BCP 47 language tag (e.g. en-US).
+ * @returns {string} Message to display.
+ */
+export function getLocaleTimeString(timestamp: number, locale = 'en-US') {
+  const now = Date.now()
```
justindomingue, line 11: probably would have been easier to test (no need to manipulate timers) with this api

```
export function getLocalTimeString(timestamp: number, locale = 'en-US', getNow = Date.now) {
  const now = getNow()
```

src/components/Tokens/TokenDetails/utils.ts
```diff
@@ -0,0 +1,52 @@
+import { SwapAction } from './types'
+
+/**
+ * Displays the time as a human-readable string.
+ *
+ * @param {number} timestamp - Transaction timestamp in milliseconds.
+ * @param {number} locale - BCP 47 language tag (e.g. en-US).
+ * @returns {string} Message to display.
+ */
+export function getLocaleTimeString(timestamp: number, locale = 'en-US') {
```
justindomingue, line 10: this util seems useful across the app. can we move it to a common place?

src/components/Tokens/TokenDetails/types.ts
```diff
@@ -0,0 +1,19 @@
...
+  timestamp: number
+  usdValue: number
+  input: SwapInOut
+  output: SwapInOut
+  maker: string
+  transactionHash: string
+}
+
+export interface SwapInOut {
+  contractAddress: string
```
justindomingue, line 11: `Address` not used in codebase?

src/components/Tokens/TokenDetails/TransactionsTable.tsx
```diff
@@ -0,0 +1,118 @@
...
+    ]
+  }, [theme.success, theme.critical, locale, referenceTokenAddress, formatNumber])
+  return (
+    <Table
+      columns={columns}
+      data={mockSwapData}
+      smallHiddenColumns={[ColumnHeader.For, ColumnHeader.USD, ColumnHeader.Maker]}
+      mediumHiddenColumns={[ColumnHeader.For, ColumnHeader.Maker]}
+      largeHiddenColumns={[ColumnHeader.Maker]}
+      extraLargeHiddenColumns={[]}
```
justindomingue, line 114: `[]` not ideal here because it creates a new reference every render.. is there a `const STABLE_ARRAY = []` const somehwere?

src/components/Tokens/TokenDetails/TransactionsTable.tsx
```diff
@@ -0,0 +1,118 @@
+import { Table } from 'components/Table'
+import { LinkCell, TextCell } from 'components/Table/Cells'
+import { useActiveLocale } from 'hooks/useActiveLocale'
+import { useMemo } from 'react'
+import { Column } from 'react-table'
+import { useTheme } from 'styled-components'
+import { shortenAddress } from 'utils/addresses'
+import { NumberType, useFormatter } from 'utils/formatNumbers'
+
+import { mockSwapData } from './mockData'
```
justindomingue, line 10: 😬

src/components/Tokens/TokenDetails/TransactionsTable.tsx
```diff
@@ -0,0 +1,118 @@
...
+  Maker = 'Maker',
+}
+
+export function TransactionsTable({ referenceTokenAddress }: { referenceTokenAddress: string }) {
+  const locale = useActiveLocale()
+  const { formatNumber } = useFormatter()
+  const theme = useTheme()
+
+  const columns: Column<Swap>[] = useMemo(() => {
+    const getColor = (action: SwapAction) => (action === SwapAction.Buy ? theme.success : theme.critical)
```
justindomingue, line 29: does this need to be a fn? it's called 4 times with the same val

src/components/Tokens/TokenDetails/TransactionsTable.tsx
```diff
@@ -0,0 +1,118 @@
...
+    return [
+      {
+        Header: ColumnHeader.Time,
+        accessor: (swap) => swap,
```
justindomingue, line 33: how does this accessor work? looks confusing

src/components/Tokens/TokenDetails/ActivitySection.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export function ActivitySection({ tokenInfo }: { tokenInfo: TokenInfo }) {
+  const [activityInView, setActivityInView] = useState(ActivityTab.Transactions)
+  return (
+    <AutoColumn>
+      <Row gap="24px" marginBottom="12px">
+        {Object.values(ActivityTab).map((activity, index) =>
+          activity === activityInView ? (
+            <ActiveTab key={`activity-${index}`}>
+              <Trans>{activity}</Trans>
```
justindomingue, line 35: don't think you can trans dynamic values like that

src/components/Table/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+                          column.isSortedDesc ? (
+                            <ArrowChangeUp width="16px" height="16px" style={{ position: 'absolute', top: 3 }} />
+                          ) : (
+                            <ArrowChangeDown width="16px" height="16px" style={{ position: 'absolute', top: 3 }} />
+                          )
+                        ) : (
+                          ''
+                        )}
+                      </Box>
+                      <Box as="span" paddingLeft={column.isSorted ? '18' : '0'}>
```
justindomingue, line 137: random padding left here feels hacky.. does sorting add an icon that you're accounting for here?

src/components/Table/Cells.tsx
```diff
@@ -0,0 +1,45 @@
...
+const StyledCell = styled.div<{ justifyContent?: string; color?: string }>`
+  justify-content: ${({ justifyContent }) => justifyContent ?? 'right'};
+  color: ${({ color }) => color};
+  padding: 14px 6px 14px 6px;
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  height: 64px;
+`
+const Link = styled.a`
```
justindomingue, line 14: externalLnik here?

cbachmeier (commented):
TDP page is crashing for native tokens 
- https://interface-git-web-2745-token-details-transactions-table-uniswap.vercel.app/tokens/polygon/NATIVE
- https://interface-git-web-2745-token-details-transactions-table-uniswap.vercel.app/tokens/ethereum/NATIVE

src/components/Tokens/TokenDetails/TransactionsTable.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  it('hides some columns at small breakpoint', () => {
+    window.innerWidth = 640
+
+    render(<TransactionsTable referenceTokenAddress={BITCOIN_ADDRESS} />)
+
+    expect(screen.getByText('Time')).toBeInTheDocument()
+    expect(screen.getByText('Type')).toBeInTheDocument()
+    expect(screen.getByText('Amount')).toBeInTheDocument()
+  })
```
cbachmeier, line 53: these tests are identical to Table/index.test.tsx and don't really add value

src/components/Tokens/TokenDetails/TransactionsTable.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  it('displays swap row', () => {
```
cbachmeier, line 55: add a snapshot test such as https://github.com/Uniswap/interface/blob/main/src/pages/PoolDetails/index.test.tsx#L95 this will help ensure styling is consistent

src/components/Tokens/TokenDetails/TransactionsTable.tsx
```diff
@@ -0,0 +1,118 @@
...
+        Cell: ({ value }: { value: { timestamp: number; transactionHash: string } }) => (
+          <LinkCell url={`https://etherscan.io/tx/${value.transactionHash}`} justifyContent="left">
```
cbachmeier, line 35: use the helper function [getExplorerLink](https://github.com/Uniswap/interface/blob/main/src/utils/getExplorerLink.ts)

src/components/Tokens/TokenDetails/TransactionsTable.tsx
cbachmeier: Add TODOs wherever you're using MockData to be easier to remove later when BE implementation is available.

src/components/Tokens/TokenDetails/TransactionsTable.tsx
```diff
@@ -0,0 +1,118 @@
...
+enum ColumnHeader {
+  Time = 'Time',
+  Type = 'Type',
+  Amount = 'Amount',
+  For = 'For',
+  USD = 'USD',
+  Maker = 'Maker',
+}
+
+export function TransactionsTable({ referenceTokenAddress }: { referenceTokenAddress: string }) {
```
cbachmeier, line 23: The TX Table's colors and columns aren't matching what I'm seeing in the [Figma](https://www.figma.com/file/2yKuZXmQ2UvgBbXR3txz4O/Dot-Info?type=design&node-id=684%3A1452112&mode=dev) should I be looking somewhere else?
Screenshot 2023-09-25 at 1 48 02 PM

src/components/Tokens/TokenDetails/TransactionsTable.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  const columns: Column<Swap>[] = useMemo(() => {
+    const getColor = (action: SwapAction) => (action === SwapAction.Buy ? theme.success : theme.critical)
+    return [
+      {
+        Header: ColumnHeader.Time,
+        accessor: (swap) => swap,
+        Cell: ({ value }: { value: { timestamp: number; transactionHash: string } }) => (
+          <LinkCell url={`https://etherscan.io/tx/${value.transactionHash}`} justifyContent="left">
+            {getLocaleTimeString(value.timestamp, locale ?? 'en-US')}
```
cbachmeier, line 36: use `DEFAULT_LOCALE`

src/components/Tokens/TokenDetails/TransactionsTable.tsx
```diff
@@ -0,0 +1,118 @@
...
+          )
+        },
+        disableSortBy: true,
+        id: ColumnHeader.USD,
+      },
+      {
+        Header: ColumnHeader.Maker,
+        accessor: 'maker',
+        Cell: ({ value }: { value: string }) => (
+          <LinkCell url={`https://etherscan.io/address/${value}`}>{shortenAddress(value)}</LinkCell>
```
cbachmeier, line 100: https://github.com/Uniswap/interface/blob/main/src/utils/getExplorerLink.ts#L31

src/components/Tokens/TokenDetails/index.tsx
```diff
@@ -58,6 +62,11 @@ const TokenTitle = styled.div`
     display: inline;
   }
 `
+const DividerLine = styled(Hr)`
```
cbachmeier, line 65: not seeing a divider
Screenshot 2023-09-25 at 2 10 24 PM

src/components/Tokens/TokenDetails/types.ts
```diff
@@ -0,0 +1,19 @@
+export interface Swap {
+  timestamp: number
+  usdValue: number
+  input: SwapInOut
+  output: SwapInOut
+  maker: string
+  transactionHash: string
+}
```
cbachmeier, line 8: we should have generated graphql types for this

src/components/Tokens/TokenDetails/TransactionsTable.tsx
cbachmeier: Looking at the Figma, the colors for txs are only used for the PDP not TDP

cbachmeier (changes requested):

src/components/Tokens/TokenDetails/utils.ts
cbachmeier, line 10: I believe we already have a function that handles this. See [getTimestampFormatter](https://github.com/Uniswap/interface/blob/main/src/components/Charts/PriceChart/format.ts)

src/components/Tokens/TokenDetails/TransactionsTable.tsx
gnewfield: I've created a reusable `ColorText` component instead. 

There was some back-and-forth regarding the color of the transaction data. For now, it's in the latest mocks, but this may change.

src/components/Tokens/TokenDetails/TransactionsTable.tsx
gnewfield: This accessor function is used internally by `react-table` and its value is made available to the `Cell` component when it's rendered.

src/components/Tokens/TokenDetails/ActivitySection.tsx
gnewfield: Refactored in order to statically translate and simplify the component overall.

src/components/Table/index.tsx
gnewfield: Removed. Will be adding sorting into a follow-up PR.

src/components/Table/Cells.tsx
gnewfield: Using the shared `ExternalLink` component now, thank you for suggesting

.env
```diff
@@ -13,3 +13,4 @@ REACT_APP_STATSIG_PROXY_URL="https://api.uniswap.org/v1/statsig-proxy"
 REACT_APP_TEMP_API_URL="https://temp.api.uniswap.org/v1"
 REACT_APP_UNISWAP_API_URL="https://api.uniswap.org/v2"
 REACT_APP_WALLET_CONNECT_PROJECT_ID="c6c9bacd35afa3eb9e6cccf6d8464395"
+TZ=UTC
```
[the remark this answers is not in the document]
gnewfield: I believe it's necessary for the tests that validate time display. Jest can mock the system timers, but the only way to mock the time zone is through an environment variable. If the `TZ` variable isn't set, then the time zone on the local host machine is used, meaning that the tests will start failing if they run in a different time zone than they're written in. This is from the [Date documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date):

> Note: While the time value at the heart of a Date object is UTC, the basic methods to fetch the date and time or its components all work in the local (i.e. host system) time zone and offset.

Please let me know if you have any other ideas about how to get around this.

src/components/Table/index.test.tsx
```diff
@@ -0,0 +1,169 @@
...
+        smallHiddenColumns={smallHiddenColumns}
+        mediumHiddenColumns={mediumHiddenColumns}
+        largeHiddenColumns={largeHiddenColumns}
+        extraLargeHiddenColumns={extraLargeHiddenColumns}
+      />
+    )
+
+    expect(screen.getByText(ColumnHeader.Column1)).toBeInTheDocument()
+    expect(screen.getByText(ColumnHeader.Column2)).toBeInTheDocument()
+    expect(screen.getByText(ColumnHeader.Column3)).toBeInTheDocument()
```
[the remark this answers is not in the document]
gnewfield: Great suggestion, I've done this now

src/components/Table/index.test.tsx
```diff
@@ -0,0 +1,169 @@
...
+      />
+    )
+
+    expect(screen.getByText('column1-data')).toBeInTheDocument()
+    expect(screen.getByText('column2-data')).toBeInTheDocument()
+    expect(screen.getByText('column3-data')).toBeInTheDocument()
+    expect(screen.getByText('column4-data')).toBeInTheDocument()
+    expect(screen.getByText('column5-data')).toBeInTheDocument()
+  })
+})
```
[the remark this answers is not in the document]
gnewfield: I think it's best for each unit test to test only one thing. This immediately makes it clear what's wrong when a test starts failing, instead of having to look into the test code to understand which part of the test failed.

src/components/Table/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+  max-height: 600px;
+  overflow-y: auto;
+  border: 2px solid ${({ theme }) => theme.surface3};
+  border-radius: 12px;
+  box-shadow: 2% 2% 10px rgba(0, 0, 0, 0.5);
+  scrollbar-width: none;
+  -ms-overflow-style: none;
+  &::-webkit-scrollbar {
+    display: none;
+  }
```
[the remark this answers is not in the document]
gnewfield: These styles are actually meant to disable the scrollbar. I tried out a couple of variations of scrollbar, and I think it looks better without. Also, I don't see the scrollbar in the mocks. I'll bring this up with design and get a definitive answer on how it should look.

src/components/Table/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+import { useEffect } from 'react'
+import { Column, IdType, useSortBy, useTable } from 'react-table'
+import styled, { useTheme } from 'styled-components'
+
+const Container = styled.div`
+  max-height: 600px;
+  overflow-y: auto;
+  border: 2px solid ${({ theme }) => theme.surface3};
+  border-radius: 12px;
+  box-shadow: 2% 2% 10px rgba(0, 0, 0, 0.5);
```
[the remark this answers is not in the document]
gnewfield: Actually I don't think this is needed anymore. I've removed.

src/components/Table/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+  ])
+
+  return (
+    <Container>
+      <StyledTable {...getTableProps()} data-testid={dataTestId}>
+        <thead
+          style={{
+            position: 'sticky',
+            top: '0px',
+            zIndex: 1,
```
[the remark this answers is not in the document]
gnewfield: This should be fixed now. `z-index` property is needed to make the head of the table float on vertical scrolling. Did we remove this view in which the swap button is stretched across the screen? I can't seem to recreate it on any screen size. In any case though, the header now appears to scroll underneath the `MobileBalanceSummaryFooter` and `MobileBottomBar` components as expected.

src/components/Table/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+    theme.breakpoint,
+    extraLargeHiddenColumns,
+  ])
+
+  return (
+    <Container>
+      <StyledTable {...getTableProps()} data-testid={dataTestId}>
+        <thead
+          style={{
+            position: 'sticky',
```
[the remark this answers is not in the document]
gnewfield: Fixed now

src/components/Table/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+            background: `${theme.surface1}`,
+          }}
+        >
+          {headerGroups.map((headerGroup, headerGroupIndex) => (
+            <tr {...headerGroup.getHeaderGroupProps()} key={`header-group-${headerGroupIndex}`}>
+              {headerGroup.headers.map((column, index) => {
+                return (
+                  <th {...column.getHeaderProps(column.getSortByToggleProps())} key={column.id}>
+                    <StyledHeaderContent leftMost={index === 0} disabled={column.disableSortBy}>
+                      <Box as="span" color="neutral2" position="relative">
```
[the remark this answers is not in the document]
gnewfield: Removed in favor of styled components

src/components/Table/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+                        {column.isSorted ? (
+                          column.isSortedDesc ? (
+                            <ArrowChangeUp width="16px" height="16px" style={{ position: 'absolute', top: 3 }} />
```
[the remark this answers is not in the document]
gnewfield: Removed all inline styling. Planning to add sort functionality and necessary UI elements in a follow-up.

src/components/Tokens/TokenDetails/ActivitySection.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+  it('has Pools and Transactions tabs', () => {
+    render(<ActivitySection tokenInfo={mockTokenInfo} />)
+
+    expect(screen.getByText('Pools')).toBeInTheDocument()
+    expect(screen.getByText('Transactions')).toBeInTheDocument()
+  })
+
+  it('lets user navigate between tabs', () => {
+    render(<ActivitySection tokenInfo={mockTokenInfo} />)
+    // expect(screen.getByTestId('pools-table')).toBeInTheDocument()
```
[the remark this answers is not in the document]
gnewfield: Was planning to uncomment this assertion after the Pools table is added, but I refactored the test so that's not necessary anymore.

src/components/Tokens/TokenDetails/ActivitySection.tsx
```diff
@@ -0,0 +1,51 @@
...
+const ActiveTab = styled(ThemedText.BodyPrimary)`
+  cursor: pointer;
+  font-size: 24px !important;
+  line-height: 32px !important;
+`
+const InactiveTab = styled(ThemedText.BodySecondary)`
+  cursor: pointer;
+  font-size: 24px !important;
+  line-height: 32px !important;
+`
```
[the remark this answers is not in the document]
gnewfield: Combined into a single component with `isActive` prop. This was also better for including the translations.

src/components/Table/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+  dataTestId?: string
+}
+
+export function Table<Data extends object>({
+  columns,
+  data,
+  smallHiddenColumns,
+  mediumHiddenColumns,
+  largeHiddenColumns,
+  extraLargeHiddenColumns,
```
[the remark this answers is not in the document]
gnewfield: Implemented horizontal scrolling. Let me know what you think now. One thought I have is that there should be some indication to the user that they should scroll to see more. Maybe a faint arrow or a light shadow?

src/components/Tokens/TokenDetails/TransactionsTable.test.tsx
gnewfield: After implementing horizontal scrolling instead of this old scheme of hiding/showing columns, I was able to significantly simplify the tests and remove a lot of duplication.

src/components/Tokens/TokenDetails/TransactionsTable.test.tsx
gnewfield: Added snapshot test

src/components/Tokens/TokenDetails/TransactionsTable.tsx
gnewfield: Nice, using this now

src/components/Tokens/TokenDetails/TransactionsTable.tsx
gnewfield: Added

src/components/Tokens/TokenDetails/TransactionsTable.tsx
gnewfield: I've removed the coloring based on the latest designs. TDP will be a neutral view in order to reduce the cognitive load of exploring all this data.

src/components/Tokens/TokenDetails/TransactionsTable.tsx
gnewfield: Changed

src/components/Tokens/TokenDetails/TransactionsTable.tsx
gnewfield: Removed address link

src/components/Tokens/TokenDetails/index.tsx
gnewfield: Added the divider with the spacing requested by design

src/components/Tokens/TokenDetails/types.ts
gnewfield: My understanding after reviewing the info repos is that V1, V2, and V3 transactions all have different schemas that are currently being queried, and none of them has a raw format that's ready to display (i.e. they all require some pre-processing). With that in mind, I thought it would be cleanest to define my own interface and implement adapters that can be paired with the queries in order to prepare the data before rendering it. Ideally, this would all be handled by the backend, and we could directly use the schema they settle on, but until then, this option should suffice. What do you think?

src/components/Tokens/TokenDetails/utils.ts
gnewfield: I tried these and found that they didn't match up with the display styles in Figma. I can ask design if they prefer any of the existing formats, but otherwise a new formatter is needed. What is your preferred date string representation?

cbachmeier: You can define your own format as well, ask @JackShort for how to implement

src/components/Table/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+  border: 1px solid ${({ theme }) => theme.surface3};
+  border-radius: 20px;
+  overflow: hidden;
+`
+const TableContainer = styled.div`
+  overflow: auto; // Enable horizontal scroll on full table
+`
+const TableHeadContainer = styled.div`
+  // Enable horizontal scroll on the header but hide scrollbar
+  overflow-x: auto;
```
cbachmeier, line 15: is this not inherited by the `TableContainer` overflow?

src/components/Table/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+  ::-webkit-scrollbar {
+    display: none;
+  }
+  -ms-overflow-style: none;
+  scrollbar-width: none;
+`
+const TableBodyContainer = styled.div`
+  max-height: 550px;
```
cbachmeier, line 23: when we add pagination will more rows be added and height expanded or will we just scroll?

src/components/Table/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+  width: 100%;
+`
+const TableBody = styled.tbody`
+  display: table;
+  width: 100%;
+`
+const TableRow = styled.tr`
+  :hover {
+    background: ${({ theme }) => theme.surface3};
+  }
```
cbachmeier, line 41: will the rows be clickable? If so you can import `${ClickableStyles}`

src/components/Table/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+    setDataWidths(Array.from(tds).map((td) => (td as HTMLElement).offsetWidth))
+  }, [])
+
+  const syncScroll = (e: React.UIEvent<HTMLDivElement>) => {
+    const tableElement = e.target as HTMLTableElement
+    if (headRef.current && bodyRef.current) {
+      headRef.current.scrollLeft = tableElement.scrollLeft
+      bodyRef.current.scrollLeft = tableElement.scrollLeft
+    }
+  }
```
cbachmeier, line 79: an alternative approach to avoid this custom scroll logic is to handle the header row as just another row similar to the data rows, this is currently what's done for the explore table. This would allow for the headers and data rows to be scroll in sync

cbachmeier: this would also simplify the scrollbar styling logic above

src/components/Tokens/TokenDetails/ActivitySection.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { TransactionsTable } from './TransactionsTable'
+
+const Container = styled.div`
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+`
+const Tab = styled(ThemedText.BodyPrimary)<{ isActive?: boolean }>`
+  cursor: pointer;
+  font-size: 24px !important;
```
cbachmeier, line 18: in these cases where you end up overwriting the main styles of the `ThemedText` I find it preferable to extend the base `Text` class to avoid `!important` override

src/components/Tokens/TokenDetails/TransactionsTable.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+    1,
+    '0x72e4f9f808c49a2a61de9c5896298920dc4eeea9',
+    18,
+    'BITCOIN',
+    'HarryPotterObamaSonic10Inu'
+  )
+
+  it('displays swap transaction', () => {
+    window.innerWidth = 1280
+    const txHash = '0xc3e8878f48b4c5048fef4988136b1cad4401b77f36f0e034e4e97929df85fb5e'
```
cbachmeier, line 18: should be exported from the `mockData`

src/components/Tokens/TokenDetails/TransactionsTable.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const mockToken = new Token(
+    1,
+    '0x72e4f9f808c49a2a61de9c5896298920dc4eeea9',
+    18,
+    'BITCOIN',
+    'HarryPotterObamaSonic10Inu'
+  )
+
+  it('displays swap transaction', () => {
+    window.innerWidth = 1280
```
cbachmeier, line 17: does this need to be set? Wouldn't the elements still be in the document since you can scroll?

src/components/Tokens/TokenDetails/TransactionsTable.tsx
```diff
@@ -0,0 +1,175 @@
...
+            <Trans>Type</Trans>
+          </ThemedText.BodySecondary>
+        </TableCell>
+      ),
+      cell: (outputTokenAddress) => (
+        <TableCell>
+          <ThemedText.BodyPrimary>
+            {String(outputTokenAddress.getValue()).toLowerCase() === referenceToken.address.toLowerCase()
+              ? 'Buy'
+              : 'Sell'}
```
cbachmeier, line 74: `Trans`

src/components/Tokens/TokenDetails/TransactionsTable.tsx
```diff
@@ -0,0 +1,175 @@
...
+        </TableCell>
+      ),
+      cell: (makerAddress) => (
+        <TableCell alignRight>
+          <ThemedText.BodyPrimary>{shortenAddress(makerAddress.getValue(), 0)}</ThemedText.BodyPrimary>
+        </TableCell>
+      ),
+    }),
+    columnHelper.accessor((row) => getExplorerLink(chainId, row.hash, ExplorerDataType.TRANSACTION), {
+      id: 'etherscan-link',
```
cbachmeier, line 156:
```suggestion
      id: 'explorer-link',
```

cbachmeier: for non-mainnet

src/components/Tokens/TokenDetails/TransactionsTable.tsx
```diff
@@ -0,0 +1,175 @@
...
+      cell: (explorerLink) => (
+        <TableCell alignRight>
+          <StyledExternalLink href={explorerLink.getValue()} data-testid={explorerLink.getValue()}>
+            <ExternalLinkIcon size="16px" stroke={theme.neutral1} />
+          </StyledExternalLink>
+        </TableCell>
+      ),
+    }),
+  ]
+  // TODO: use live data
```
cbachmeier, line 173: add Linear story to comment, alson in `#proj-interface-info` I added the base for adding token tx query

src/components/Tokens/TokenDetails/TransactionsTable.tsx
```diff
@@ -0,0 +1,175 @@
...
+  symbol: string
+  amount: number
+}
+
+export function TransactionsTable({ chainId, referenceToken }: { chainId: ChainId; referenceToken: Token }) {
+  const theme = useTheme()
+  const locale = useActiveLocale()
+  const { formatNumber } = useFormatter()
+
+  const columnHelper = createColumnHelper<SwapTransaction>()
```
cbachmeier, line 41: this is pretty cool

cbachmeier (changes requested):
overall looks good, mostly small nits

src/components/Table/index.tsx
cbachmeier: after going through the whole PR, not sure how this would affect the columnHelper in TokenDetails/TransactionsTable

src/components/Table/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+  -ms-overflow-style: none;
+  scrollbar-width: none;
+`
+const TableBodyContainer = styled.div`
+  max-height: 550px;
+  overflow-y: auto; // Enable vertical scroll only on table body
+`
+const TableSection = styled.table`
+  border-collapse: collapse;
```
cbachmeier, line 27: this also needs `width: 100%;` to fit parent container
For example when this table is used elsewhere
Screenshot 2023-11-09 at 11 51 16 AM
Screenshot 2023-11-09 at 11 51 30 AM

src/components/Table/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const headRef = useRef<HTMLTableElement>(null)
+  const bodyRef = useRef<HTMLTableElement>(null)
+  const [headerWidths, setHeaderWidths] = useState<number[]>(new Array(columns.length).fill(0))
+  const [dataWidths, setDataWidths] = useState<number[]>(new Array(columns.length).fill(0))
+
+  useLayoutEffect(() => {
+    const ths = document.querySelectorAll('thead th')
+    const tds = document.querySelectorAll('tbody tr:first-child td')
+    setHeaderWidths(Array.from(ths).map((th) => (th as HTMLElement).offsetWidth))
+    setDataWidths(Array.from(tds).map((td) => (td as HTMLElement).offsetWidth))
```
cbachmeier, line 70: the logic for determining width seems off as on a fullscreen macbook I believe I should be able to see all the columns?
Screenshot 2023-11-09 at 12 05 25 PM

And when made fullscreen, the cells still seem too large and overflow
Screenshot 2023-11-09 at 12 06 55 PM

src/components/Table/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+  padding: 12px 16px;
+  white-space: nowrap;
+  justify-content: ${({ alignRight }) => (alignRight ? 'flex-end' : 'flex-start')};
+`
+const TableHeader = styled.th<{ width: number }>`
+  min-width: ${({ width }) => width}px;
+  padding: 8px;
+  border-bottom: 1px solid ${({ theme }) => theme.surface3};
+`
+const TableData = styled.td<{ width: number }>`
```
cbachmeier, line 55: all inputs that share the name of a style should be prefixed with `$` ie `$width` and `$alignRight`

src/components/Table/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+  overflow: auto; // Enable horizontal scroll on full table
+`
+const TableHeadContainer = styled.div`
+  // Enable horizontal scroll on the header but hide scrollbar
+  overflow-x: auto;
+  ::-webkit-scrollbar {
+    display: none;
+  }
+  -ms-overflow-style: none;
+  scrollbar-width: none;
```
cbachmeier, line 20: adding `overscroll-behavior: none;` will stop the bouncing on the header when overscrolled

src/components/Table/index.tsx
cbachmeier: this actually runs into an issue where the `TableCells` offsetWidth is too large and basically always gets an overflow

PR #7358, opened 2023-09-20
chore: currency percentages

JackShort:
* updates percentages to use the numberFormatter functions (not including anything on NFT pages)


just-toby (approved):
pls add unit tests for formatPercent

zzmp (changes requested):

src/utils/formatNumbers.test.ts
```diff
@@ -460,3 +460,37 @@ describe('formatReviewSwapCurrencyAmount', () => {
...
+  beforeEach(() => {
+    mocked(useLocalCurrencyConversionRate).mockReturnValue({ data: 1.0, isLoading: false })
+    mocked(useCurrencyConversionFlagEnabled).mockReturnValue(true)
+  })
+
+  it('should correctly format undefined', () => {
+    const { formatPercent } = renderHook(() => useFormatter()).result.current
+
+    expect(formatPercent(undefined)).toBe('-')
+  })
```
zzmp, line 474: you should include all of the special cases here: `null`, `undefined`, `Infinity`, `NaN`.
you can do it most elegantly with `it.each`:

```suggestion
  it.each([null, undefined, Infinity, NaN])('should correctly format %p', ([value]) => {
    const { formatPercent } = renderHook(() => useFormatter()).result.current

    expect(formatPercent(value)).toBe('-')
  })
```

zzmp (changes requested):

src/utils/formatNumbers.test.ts
JackShort: oh cool - thats good to know ty

zzmp (approved):

PR #7359, opened 2023-09-20
test: unit tests for parseRemote

just-toby:
## Description

increasing test coverage for this file

_Relevant docs:_ https://app.codecov.io/gh/Uniswap/interface/blob/main/src%2Fcomponents%2FAccountDrawer%2FMiniPortfolio%2FActivity%2FparseRemote.tsx

## Test plan
- [x] Unit tests


src/components/AccountDrawer/MiniPortfolio/Activity/parseRemote.test.tsx
```diff
@@ -0,0 +1,392 @@
...
+    jest.useFakeTimers()
+  })
+  describe('parseRemoteActivities', () => {
+    it('should not parse open UniswapX order', () => {
+      const result = parseRemoteActivities(jest.fn(), [MockOpenUniswapXOrder])
+      expect(result).toEqual({})
+    })
+    it('should parse closed UniswapX order', () => {
+      const result = parseRemoteActivities(jest.fn(), [MockClosedUniswapXOrder])
+      expect(result?.['someHash']).toEqual({
```
zzmp, line 38: These would be excellent candidates for snapshot testing, because they verify that the parsed object doesn't change. Instead of manually entering what it should equal, why not just `expect(result?.['someHash']).toMatchSnapshot()`?

I think this whole file could be simplified with snapshot testing, as you're testing that the parsing isn't changing over time - it's less relevant exactly how it is parsed.

zzmp (changes requested):
I think this'd be better as snapshot tests, but wdyt?

src/components/AccountDrawer/MiniPortfolio/Activity/parseRemote.test.tsx
just-toby: yeah i agree

cartcrom (approved):

natew (approved):

PR #7360, opened 2023-09-20
feat: lazy-load safe only if iframed

zzmp:
## Description

Lazy-loads Gnosis Safe, and only does so if we are on an iframe. This allows us to defer loading viem (used by Safe) unless we need to, and reduces the main bundle by 10%.

### DIFF
--- a/patches/@web3-react+gnosis-safe+8.2.3.patch
+++ b/patches/@web3-react+gnosis-safe+8.2.3.patch
@@ -1,0 +1,15 @@
+diff --git a/node_modules/@web3-react/gnosis-safe/dist/index.js b/node_modules/@web3-react/gnosis-safe/dist/index.js
+index 015a33c..4cd7cde 100644
+--- a/node_modules/@web3-react/gnosis-safe/dist/index.js
++++ b/node_modules/@web3-react/gnosis-safe/dist/index.js
+@@ -68,8 +68,8 @@ class GnosisSafe extends types_1.Connector {
+             if (this.eagerConnection)
+                 return;
+             // kick off import early to minimize waterfalls
+-            const SafeAppProviderPromise = Promise.resolve().then(() => __importStar(require('@safe-global/safe-apps-provider'))).then(({ SafeAppProvider }) => SafeAppProvider);
+-            yield (this.eagerConnection = Promise.resolve().then(() => __importStar(require('@safe-global/safe-apps-sdk'))).then((m) => __awaiter(this, void 0, void 0, function* () {
++            const SafeAppProviderPromise = Promise.resolve().then(async () => __importStar(await import('@safe-global/safe-apps-provider'))).then(({ SafeAppProvider }) => SafeAppProvider);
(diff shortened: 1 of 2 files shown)


src/connection/eagerlyConnect.ts
```diff
@@ -20,8 +20,14 @@ async function connect(connector: Connector, type: ConnectionType) {
   }
 }
 
-connect(gnosisSafeConnection.connector, ConnectionType.GNOSIS_SAFE)
+// The Safe connector will only work from safe.global, which iframes;
```
JFrankfurt, line 23: can you give me some more context on this?

patches/@web3-react+gnosis-safe+8.2.3.patch
```diff
@@ -0,0 +1,15 @@
...
+             if (this.eagerConnection)
+                 return;
+             // kick off import early to minimize waterfalls
+-            const SafeAppProviderPromise = Promise.resolve().then(() => __importStar(require('@safe-global/safe-apps-provider'))).then(({ SafeAppProvider }) => SafeAppProvider);
+-            yield (this.eagerConnection = Promise.resolve().then(() => __importStar(require('@safe-global/safe-apps-sdk'))).then((m) => __awaiter(this, void 0, void 0, function* () {
++            const SafeAppProviderPromise = Promise.resolve().then(async () => __importStar(await import('@safe-global/safe-apps-provider'))).then(({ SafeAppProvider }) => SafeAppProvider);
++            yield (this.eagerConnection = Promise.resolve().then(async () => __importStar(await import('@safe-global/safe-apps-sdk'))).then((m) => __awaiter(this, void 0, void 0, function* () {
+                 this.sdk = new m.default(this.options);
+                 const safe = yield Promise.race([
+                     this.sdk.safe.getInfo(),
```
JFrankfurt, line 15: I'm a little worried about the frequency with which I'm starting to see patches

JFrankfurt (approved):

patches/@web3-react+gnosis-safe+8.2.3.patch
zzmp: it's only for web3-react, because it's easier to do this (for a large saving) than to publish an esm version of web3-react...for now.

tinaszheng (approved):

patches/@web3-react+gnosis-safe+8.2.3.patch
tinaszheng: can you document a followup to patch this in the main package :O

src/connection/eagerlyConnect.ts
zzmp: [safe app](https://app.safe.global/welcome) works by iframing dapps, and this is a special connector that proxies RPCs over the iframe messaging API (`sendMessage`). If we're not in an iframe, we're also not in the safe app, so this connector (and its baggage) are unnecessary.

PR #7361, opened 2023-09-20
fix: lazy load nft profile

zzmp:
## Description

Removes unused parts of the lazy-loading flow for the nft profile.
Suspense was used on a component defined in the same file, so that component was always already loaded.


cbachmeier (approved):
Profile page skeleton loads as expected

JackShort (approved):

PR #7362, opened 2023-09-20
chore: prohibit barrels

zzmp:
## Description

Lints against the typescript barrel pattern (`export *`) because it prevents checks for unused code and it reduces tree-shaking (because the barrel is considered a side effect and always bundled).

This eliminates some unused nft code from first load, including the entirety of the @opensea dependency, altogether reducing the initial bundle by about 12.5%.


src/utils/index.ts
```diff
@@ -1,3 +1,3 @@
-export * from './addresses'
-export * from './escapeRegExp'
-export * from './getContract'
+export { isAddress, shortenAddress } from './addresses'
```
cbachmeier, line 1: this feels like a good barrel compromise

cbachmeier (approved):
Good night sweet prince 🫡

.eslintrc.js
```diff
@@ -70,6 +70,13 @@ module.exports = {
             ],
           },
         ],
+        'no-restricted-syntax': [
+          'error',
+          {
+            selector: ':matches(ExportAllDeclaration)',
+            message: 'Barrel exports bloat the bundle size.',
```
cbachmeier, line 77: nit: could also add a note about tree-shaking

JackShort (approved):
actually viewed every file

PR #7368, opened 2023-09-21
feat: uniswapX opt-out update

just-toby:
## Description

updates the code for transitioning to UniswapX by default, or opt-out, phase. This change is needed because we want to opt-in users who only saw but dismissed the opt-in flow, but unfortunately right now we're treating them the same as users who actually tried uniswapX and then disabled it - both types of users would have `user.disabledUniswapX=true`. since we can't differentiate between those two groups of users, we are deciding to opt-in both groups when we roll out the on-by-default change.

now, when we flip on `isUniswapXDefaultEnabled`, we want to treat all users the same and opt them into uniswapX until they explicitly opt out again. 

as a reminder (and see the linked doc for more details), we are rolling this out gradually via the feature flag which is why we need these temporary flags. after the transition we'll be able to migrate all users to the proper RouterPreference setting via a redux migration and delete all these boolean flags

why add a new user state field? we could implement a small redux migration now to reset all users' `disabledUniswapX` values to false, but migrations only run once per user and happen in the first session after the release. this would force the timing of the rollout, and it's not compatible with a gradual rollout. using a new field decouples the app release from the feature rollout in this case.

here are the important places in code implementing the transition

- Router Preference toggle
  - This needs to *appear* enabled even if the user's routerPreference is `API` 
  - if the user disables the toggle again, after the on-by-default period has started, then it should be disabled.
  - we use the new field, `user.optedOutOfUniswapX`, to track if it's been disabled
  - we now disregard the old field, `user.disabledUniswapX` since it was too general to be useful here
- Quote request - [slice.ts](https://github.com/Uniswap/interface/blob/main/src/state/routing/slice.ts#L76)
  - this is where we actually implement whether a user has turned OFF uniswapX during this period.
  - using the new field, if the user has disabled the toggle in settings and has not re-enabled it, then we'll just request classic quotes
- Quote response - [transformRoutesToTrade](https://github.com/Uniswap/interface/blob/main/src/state/routing/utils.ts#L187)
  - this is where we decide to show the UniswapX quote instead of classic, if the price is better.
  - no changes here in this PR - we still check the featureFlag and, if true, show the better X quote to all users regardless of persisted RouterPreference.

_Relevant docs:_ https://www.notion.so/uniswaplabs/Web-UniswapX-Opt-Out-d7d5d33b3e2c447ab43d40cc92e427ae

### QA (ie manual testing)

- [x] verified that the toggle is working as expected, and updating the persisted fields correctly

### DIFF
--- a/src/components/Settings/RouterPreferenceSettings/index.tsx
+++ b/src/components/Settings/RouterPreferenceSettings/index.tsx
@@ -8,8 +8,8 @@ import { isUniswapXSupportedChain } from 'constants/chains'
 import { useUniswapXDefaultEnabled } from 'featureFlags/flags/uniswapXDefault'
 import { useAppDispatch } from 'state/hooks'
 import { RouterPreference } from 'state/routing/types'
-import { useRouterPreference, useUserDisabledUniswapX } from 'state/user/hooks'
-import { updateDisabledUniswapX } from 'state/user/reducer'
+import { useRouterPreference, useUserOptedOutOfUniswapX } from 'state/user/hooks'
+import { updateDisabledUniswapX, updateOptedOutOfUniswapX } from 'state/user/reducer'
 import styled from 'styled-components'
 import { Divider, ExternalLink, ThemedText } from 'theme'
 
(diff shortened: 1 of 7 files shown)


tinaszheng (approved):
Nice!

cartcrom (approved):

PR #7369, opened 2023-09-22
fix: Remove e2e test

tinaszheng:
It looks like backend stopped returning stats data for tokens with no prices, so this test is no longer true


zzmp (approved):

cypress/e2e/token-details.test.ts
```diff
@@ -55,14 +55,6 @@ describe('Token details', () => {
     if (cy.get('[data-cy="chart-header"]').contains('Price Unavailable')) {
       cy.get('[data-cy="missing-chart"]').should('exist')
     }
```
cbachmeier, line 57: If this is expected we could add a test to confirm
```suggestion
    }
    // Stats should not exist
    cy.get(getTestSelector('token-details-stats')).should('not.exist')
```

just-toby (approved):

PR #7376, opened 2023-09-26
feat: uniswapx deadline

just-toby:
## Description

problem: right now, some fillers are receiving uniswapx orders after startTime has already passed so they no longer have an exclusivity window. we think it's because users are taking too long to sign the permit signature in their wallet. the interface currently only re-prompts a signature if the submission happens after deadline
solution: change the re-prompt logic to happen after startTime instead (so moving up the time requirement), and also add copy in the swap submission modal to say that their signature has expired 

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2898/uniswapx-change-time-to-sign
_Slack thread:_ https://uniswapteam.slack.com/archives/C053G2YEK5M/p1695663790335929

## Screen capture

https://github.com/Uniswap/interface/assets/66155195/18987195-1465-49cb-b287-4a43d0edce41

### QA (ie manual testing)

- [x] manual testing - wait 30s before signing and ensure the new text appears, transitions, and re-requests the signature
- [x] make sure that this happens on subsequent timed-out signatures too (see video w/ two in a row)

### DIFF
--- a/src/components/swap/ConfirmSwapModal.tsx
+++ b/src/components/swap/ConfirmSwapModal.tsx
@@ -20,7 +20,7 @@ import { SwapResult } from 'hooks/useSwapCallback'
 import useWrapCallback from 'hooks/useWrapCallback'
 import useNativeCurrency from 'lib/hooks/useNativeCurrency'
 import { getPriceUpdateBasisPoints } from 'lib/utils/analytics'
-import { useCallback, useEffect, useState } from 'react'
+import { useCallback, useEffect, useMemo, useState } from 'react'
 import { InterfaceTrade, TradeFillType } from 'state/routing/types'
 import { Field } from 'state/swap/actions'
 import { useIsTransactionConfirmed, useSwapTransactionStatus } from 'state/transactions/hooks'
@@ -28,6 +28,7 @@ import styled from 'styled-components'
 import { ThemedText } from 'theme/components'
 import invariant from 'tiny-invariant'
 import { isL2ChainId } from 'utils/chains'
+import { SignatureExpiredError } from 'utils/errors'
 import { NumberType, useFormatter } from 'utils/formatNumbers'
 import { formatSwapPriceUpdatedEventProperties } from 'utils/loggingFormatters'
 import { didUserReject } from 'utils/swapErrorToUserReadableMessage'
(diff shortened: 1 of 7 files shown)


src/components/swap/PendingModalContent/TransitionText.tsx
```diff
@@ -0,0 +1,81 @@
...
+  height: 100%;
+  position: absolute;
+  transition: display ${({ theme }) => `${theme.transition.duration.fast} ${theme.transition.timing.inOut}`};
+  ${slideInAnimation}
+  &.${AnimationType.EXITING} {
+    ${slideOutAnimation}
+  }
+`
+
+export function TransitionText({
```
tinaszheng, line 44: oh nice shared component here! maybe move this to `src/components/TransitionText` so it can be used in the future? nbd tho

tinaszheng (approved):

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -286,6 +288,10 @@ export default function ConfirmSwapModal({
   onCurrencySelection: (field: Field, currency: Currency) => void
   fiatValueInput: { data?: number; isLoading: boolean }
   fiatValueOutput: { data?: number; isLoading: boolean }
+  // onConfirm triggers the same callback, but it also clears other swap state.
+  // for the UniswapX signature expiration logic, we need a callback which only
+  // requests the signature without clearing swap state.
+  onRetryUniswapXSignature?: () => void
```
justindomingue, line 294: possibly more usable/readable/less hacky alternative would be to provide `onconfirm` and `onclear` and call both for no uniswapx signature retry

src/components/swap/ConfirmSwapModal.tsx
```diff
@@ -410,14 +419,19 @@ export default function ConfirmSwapModal({
     return undefined
   }
 
+  const errorType = useMemo(() => {
```
justindomingue, line 422: what's the goal of memo here?

just-toby: actually i don't need to memoize here, i just wanted the logic to be in a function for readability. i'll update this

src/components/swap/ConfirmSwapModal.tsx
just-toby: i like it

cartcrom (approved):

PR #7379, opened 2023-09-26
fix: hide price when no amount input

just-toby:
## Description

addressing design feedback here - don't show the spot price when there's no input amount.

note: we're still fetching the USD price when there's no input, to preload the price which helps later price checks load faster

_Slack thread:_ https://uniswapteam.slack.com/archives/C047U65H422/p1695760705173299

## Test plan

image


tinaszheng (approved):
nice

justindomingue (approved):

PR #7383, opened 2023-09-27
feat: dynamic defaultCurrencyCode for moonpay

just-toby:
## Description

when opening moonpay from TDP, if the detailed token is one of the few supported by moonpay for direct purchase, we will auto-populate the moonpay output currency.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-1971/[tdp]-pre-populate-buy-button-on-swap-box-to-tokens-tdp

### QA (ie manual testing)

- [x] navigate to the TDP for one of the supported tokens, click "buy", verify the token is auto-selected

### Automated testing

- [x] Unit test


JFrankfurt:
> if the detailed token

what is the "detailed" token?

src/components/FiatOnrampModal/util.ts
```diff
@@ -0,0 +1,75 @@
...
+  'weth',
+  'wbtc',
+  'matic_polygon',
+  'polygon',
+  'usdc_arbitrum',
+  'usdc_optimism',
+  'usdc_polygon',
+  'usdc',
+  'usdt',
+] as const
```
JFrankfurt, line 21: nit: make this a `constants` file, or add it to an existing one.
nit2: util.ts is too general imo--should have a single file for specific util functions, or define them where they are used

src/constants/tokens.ts
```diff
@@ -96,6 +96,13 @@ export const DAI_OPTIMISM = new Token(
   'DAI',
   'Dai stable coin'
 )
+export const MATIC = new Token(
+  ChainId.MAINNET,
+  '0x7d1afa7b718fb893db30a3abc0cfc608aacfebb0',
+  18,
+  'MATIC',
+  'Polygon Matic'
+)
```
JFrankfurt, line 105: v surprised this was not already defined

src/components/FiatOnrampModal/util.ts
```diff
@@ -0,0 +1,75 @@
...
+  return 'eth'
+}
+
+export function parsePathParts(pathname: string): { network?: string; tokenAddress?: string } {
+  const pathParts = pathname.split('/')
+  // Matches the /tokens/<network>/<tokenAddress> path.
+  const network = pathParts.length > 2 ? pathParts[pathParts.length - 2] : undefined
+  const tokenAddress = pathParts.length > 2 ? pathParts[pathParts.length - 1] : undefined
+  return { network, tokenAddress }
+}
```
JFrankfurt, line 75: Is there an easier way to do this w/ the existing hooks from react-router?

JFrankfurt (approved):

src/components/FiatOnrampModal/util.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (isSameAddress(address, '0x7ceb23fd6bc0add59e62ac25578270cff1b9f619') && chain === Chain.Polygon) {
+    return 'eth_polygon'
+  }
+  if (isSameAddress(address, WETH9[ChainId.MAINNET]?.address) && chain === Chain.Ethereum) {
+    return 'weth'
+  }
+  if (isSameAddress(address, WBTC.address) && chain === Chain.Ethereum) {
+    return 'wbtc'
+  }
+  if (isSameAddress(address, 'NATIVE') && chain === Chain.Polygon) {
```
justindomingue, line 45: is there a reason you aren't using `isSameAddress` to compare `NATIVE` above but you are here?

src/components/FiatOnrampModal/util.ts
```diff
@@ -0,0 +1,75 @@
...
+  chainName: string | undefined
+): MoonpaySupportedCurrencyCode {
+  const chain = validateUrlChainParam(chainName)
+  if (address === 'NATIVE' && chain === Chain.Arbitrum) {
+    return 'eth_arbitrum'
+  }
+  if (address === 'NATIVE' && chain === Chain.Optimism) {
+    return 'eth_optimism'
+  }
+  if (isSameAddress(address, '0x7ceb23fd6bc0add59e62ac25578270cff1b9f619') && chain === Chain.Polygon) {
```
justindomingue, line 36:
```suggestion
  if (isSameAddress(address, WETH_POLYGON.address) && chain === Chain.Polygon) {
```

src/utils/addresses.ts
```diff
@@ -11,6 +11,10 @@ export function isAddress(value: any): string | false {
   }
 }
 
+export function isSameAddress(a?: string, b?: string) {
```
justindomingue, line 14: duped here now https://github.com/Uniswap/interface/blob/main/src/components/AccountDrawer/MiniPortfolio/Activity/parseRemote.tsx#L80

src/components/FiatOnrampModal/util.ts
```diff
@@ -0,0 +1,75 @@
+import { ChainId, WETH9 } from '@uniswap/sdk-core'
+import { MATIC, USDC_ARBITRUM, USDC_MAINNET, USDC_OPTIMISM, USDC_POLYGON, USDT, WBTC } from 'constants/tokens'
+import { Chain } from 'graphql/data/__generated__/types-and-hooks'
+import { validateUrlChainParam } from 'graphql/data/util'
+import { isSameAddress } from 'utils/addresses'
+
+export const MOONPAY_SUPPORTED_CURRENCY_CODES = [
```
cbachmeier, line 7: you can convert this to a map of addresses and chains which would greatly reduce the amount of copy and pasted code needed in the `getDefaultCurrencyCode` function

cbachmeier (changes requested):
Should this also affect the moonpay modal from the mini portfolio
buyFromMiniPort

src/utils/addresses.ts
just-toby: i removed this in this PR already

src/components/FiatOnrampModal/getDefaultCurrencyCode.ts
```diff
@@ -0,0 +1,76 @@
...
+  [Chain.Optimism]: {
+    [USDC_OPTIMISM.address]: 'usdc_optimism',
+    NATIVE: 'eth_optimism',
+  },
+  [Chain.Polygon]: {
+    [USDC_POLYGON.address]: 'usdc_polygon',
+    [WETH_POLYGON.address]: 'eth_polygon',
+    NATIVE: 'matic_polygon',
+  },
+  // Defaults for unsupported chains, needed to satisfy type checker.
```
cbachmeier, line 44: to work around this you can just define a type for the subset of Chain for only the Moonpay supported ones

cbachmeier (changes requested):
Currently getting eth on moonpay on the deploy for tdp pages like usdc and wbtc
Screenshot 2023-09-28 at 12 58 07 PM
Screenshot 2023-09-28 at 12 58 48 PM

just-toby:
@cbachmeier @JFrankfurt 

i tried using the `useParams` hook at first, but that breaks because the modal is not in the Router context. it's at the top level in `AppChrome`. i don't think we need to refactor that when this function works.

as for the default token when opening from Mini Portfolio - i think you could make a valid argument for either ETH or the detailed Token, but for now i think we should use the detailed token since it's A) simpler to implement and B) the user can still change the token in the moonpay modal at any point.

just-toby:
also, apologies if i push a broken commit - i can only test this change in the preview URL. i can't test locally before pushing

JFrankfurt:
can you add a comment on the paeans function explaining it shouldn't be preferred, and the reader probably wants the router hook instead?

kristiehuang (approved):

src/components/FiatOnrampModal/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+} from 'constants/tokens'
+import { Chain } from 'graphql/data/__generated__/types-and-hooks'
+import { validateUrlChainParam } from 'graphql/data/util'
+
+import { MoonpaySupportedCurrencyCode } from './constants'
+
+type MoonpaySupportedChain = Chain.Ethereum | Chain.Polygon | Chain.Arbitrum | Chain.Optimism
+const moonPaySupportedChains = [Chain.Ethereum, Chain.Polygon, Chain.Arbitrum, Chain.Optimism]
+
+const CURRENCY_CODES: {
```
kristiehuang, line 21: kind of got lost with MoonpaySupportedChain, moonPaySupportedChains, CURRENCY_CODES, MOONPAY_SUPPORTED_CURRENCY_CODES all being separate things. think a comment would help with readability! but otherwise lgtm

kristiehuang (approved):

justindomingue (approved):

PR #7384, opened 2023-09-27
fix: wrap/unwrap activity parsing

just-toby:
## Description

fix activity parsing for wrap/unwrap transactions that are labelled as "LEND" by zerion

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2721/[web][metamask]-inaccurate-title-and-transaction-hash-paragraph-in

### Before

Screenshot 2023-09-27 at 3 39 09 PM

### After

Screenshot 2023-09-27 at 3 38 47 PM

### Reproducing the error

1. wrap ETH or unwrap WETH, let the transaction confirm, then open the Mini Portfolio activity history and see "LEND"

### QA (ie manual testing)

- [x] repeat above process, verify that "wrapped" or "unwrapped" is shown instead w/ amounts

### Automated testing

- [x] Unit test

### DIFF
--- a/src/components/AccountDrawer/MiniPortfolio/Activity/parseRemote.tsx
+++ b/src/components/AccountDrawer/MiniPortfolio/Activity/parseRemote.tsx
@@ -187,38 +187,12 @@ function parseSwap(changes: TransactionChanges, formatNumberOrString: FormatNumb
  * This function parses the transaction changes to determine if the transaction is a wrap/unwrap transaction.
  */
 function parseLend(changes: TransactionChanges, formatNumberOrString: FormatNumberOrStringFunctionType) {
-  const received = changes.TokenTransfer.find((t) => t.direction === 'IN')
-  const sent = changes.TokenTransfer.find((t) => t.direction === 'OUT')
-  if (!received || !sent) {
-    return {}
+  const native = changes.TokenTransfer.find((t) => t.tokenStandard === 'NATIVE')?.asset
+  const erc20 = changes.TokenTransfer.find((t) => t.tokenStandard === 'ERC20')?.asset
+  if (native && erc20 && gqlToCurrency(native)?.wrapped.address === gqlToCurrency(erc20)?.wrapped.address) {
+    return parseSwap(changes, formatNumberOrString)
   }
-  const supportedSentChain = supportedChainIdFromGQLChain(sent?.asset.chain)
-  const supportedReceivedChain = supportedChainIdFromGQLChain(received?.asset.chain)
-  if (!supportedSentChain || !supportedReceivedChain) {
-    return {}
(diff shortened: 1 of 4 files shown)


justindomingue:
could this be a backend fix?

src/components/AccountDrawer/MiniPortfolio/Activity/parseRemote.tsx
```diff
@@ -182,6 +182,45 @@ function parseSwap(changes: TransactionChanges, formatNumberOrString: FormatNumb
...
+    isSameAddress(nativeOnChain(supportedReceivedChain).wrapped.address, received.asset.address)
+  ) {
+    return {
+      title: getSwapTitle(sent, received),
+      descriptor: getSwapDescriptor({ tokenIn: sent.asset, inputAmount, tokenOut: received.asset, outputAmount }),
+      currencies: [gqlToCurrency(sent.asset), gqlToCurrency(received.asset)],
+    }
+  }
+  return {} // Not a wrap/unwrap transaction, so don't overwrite the default "Lend" title
+}
```
cartcrom, line 222: reuse swap logic:
```suggestion
/**
 * Some unrap/unwrap transactions are erroneously labelled as lend transactions on the backend.
 * This function parses the transaction changes to determine if the transaction is a wrap/unwrap transaction.
 */
function parseLend(changes: TransactionChanges, formatNumberOrString: FormatNumberOrStringFunctionType) {
  const native = changes.TokenTransfer.find((t) => t.tokenStandard === 'NATIVE')?.asset
  const erc20 = changes.TokenTransfer.find((t) => t.tokenStandard === 'ERC20')?.asset

  if (native && erc20 && gqlToCurrency(native)?.wrapped.address === gqlToCurrency(erc20)?.wrapped.address) {
    return parseSwap(changes, formatNumberOrString)
  }

  return { title: t`Unknown Lend` }
}
```

cartcrom (changes requested):

just-toby:
> could this be a backend fix?

i think this is coming from zerion - i can notify the backend team just in case this is not desired behavior. but don't see a reason why we shouldn't update this logic in the meantime.

i was also torn on this because i think wrapping is technically lending

natew (approved):

cartcrom (approved):

cartcrom:
> > could this be a backend fix?
> 
> i think this is coming from zerion - i can notify the backend team just in case this is not desired behavior. but don't see a reason why we shouldn't update this logic in the meantime.
> 
> i was also torn on this because i think wrapping is technically lending

I have reported this to BE, it should be fixed on either zerion or the BE's end eventually because we're _sometimes_ getting SWAP and other times getting LEND

PR #7386, opened 2023-09-28
feat: remove buy button and landing terminology for uk

JackShort:
We were requested to remove certain items and verbiage regarding UK regulation - you can see here: https://www.notion.so/uniswaplabs/UK-Regulation-Marketing-Updates-0061115831f947a7bd573e86c2f16f3a

The solution involves updating the analytics package to return the origin country and that is used appropriately to display the correct information.

__caching will be implemented in a subsequent pr__
__will also write tests when caching is implemented__


src/analytics/index.tsx
```diff
@@ -52,3 +52,8 @@ export const sendAnalyticsEvent: typeof sendAnalyticsTraceEvent = (event, proper
     sendAnalyticsTraceEvent(event, properties)
   }
 }
+
+// This is only used for initial page load so we can get the user's country
+export const sendAnalyticsEventIgnoreTracking: typeof sendAnalyticsTraceEvent = (event, properties) => {
+  sendAnalyticsTraceEvent(event, properties)
```
just-toby, line 58: how is this function ignoring tracking? it's just calling the logging function in the same way as `sendAnalyticsEvent` right?

JackShort: `sendAnalyticsEvent` checks to make sure the analytics ignore key is not set - this does not

src/pages/Landing/index.tsx
```diff
@@ -124,6 +141,12 @@ const TitleText = styled.h1<{ isDarkMode: boolean }>`
   background-clip: text;
   -webkit-background-clip: text;
 
+  ${({ $visible }) =>
```
just-toby, line 144: should there be CSS which hides this component when `$visible` is false?

src/pages/Landing/index.tsx
```diff
@@ -150,9 +173,15 @@ const SubText = styled.div`
   }
 `
 
-const SubTextContainer = styled.div`
+const SubTextContainer = styled.div<{ $visible: boolean }>`
   display: flex;
   justify-content: center;
+
+  ${({ $visible }) =>
```
just-toby, line 180: should there be CSS which hides this component when $visible is false?

src/pages/Landing/index.tsx
```diff
@@ -320,6 +376,34 @@ export default function Landing() {
 
   const location = useLocation()
   const queryParams = parse(location.search, { ignoreQueryPrefix: true })
+
+  const Titles = useMemo(() => {
```
just-toby, line 380: i think this should be. lowercase

src/analytics/index.tsx
```diff
@@ -52,3 +52,8 @@ export const sendAnalyticsEvent: typeof sendAnalyticsTraceEvent = (event, proper
     sendAnalyticsTraceEvent(event, properties)
   }
 }
+
+// This is only used for initial page load so we can get the user's country
```
just-toby, line 56: if you don't want this function to be used in the future (it probably shouldn't), we might want to rename to something like "sendInitializationEvent"

src/pages/Landing/index.tsx
JackShort: nothing gets displayed anyway but i shall just add one since it won't hurt

src/pages/Landing/index.tsx
```diff
@@ -402,8 +479,8 @@ export default function Landing() {
               />
             ))}
           </CardGrid>
-          <CardGrid cols={3}>
-            {MORE_CARDS.map(({ darkIcon, lightIcon, ...card }) => (
+          <CardGrid cols={renderUkSpecificText ? 2 : 3}>
```
just-toby, line 482: can this be less specific? like can we use `extraCards.length` to determine this value? 

it's fine with one parameter but i'm just imagining the cards logic getting more complex

just-toby (approved):
lgtm

src/pages/Landing/index.tsx
JackShort: based

src/pages/Landing/index.test.tsx
```diff
@@ -7,7 +9,12 @@ import Landing from '.'
 jest.mock('hooks/useDisableNFTRoutes')
 
 describe('disable nft on landing page', () => {
+  beforeAll(() => {
+    store.dispatch(setOriginCountry('US'))
+  })
+
   it('renders nft information and card', () => {
+    store.dispatch(setOriginCountry('US'))
```
gnewfield, line 17: Why do this again if already done in `beforeAll`?

src/pages/Landing/index.tsx
```diff
@@ -303,9 +334,36 @@ export default function Landing() {
   const cardsRef = useRef<HTMLDivElement>(null)
   const selectedWallet = useAppSelector((state) => state.user.selectedWallet)
   const shouldDisableNFTRoutes = useDisableNFTRoutes()
-  const cards = useMemo(
-    () => MAIN_CARDS.filter((card) => !(shouldDisableNFTRoutes && card.to.startsWith('/nft'))),
-    [shouldDisableNFTRoutes]
+  const originCountry = useAppSelector((state: AppState) => state.user.originCountry)
+  const renderUkSpecificText = Boolean(originCountry) && originCountry === 'GB'
```
gnewfield, line 338: Just confirming - will origin country be `GB` or `UK`?

src/components/swap/SwapBuyFiatButton.test.tsx
```diff
@@ -45,6 +47,7 @@ describe('SwapBuyFiatButton.tsx', () => {
   let useOpenModal: jest.Mock<any, any>
 
   beforeAll(() => {
+    store.dispatch(setOriginCountry('US'))
```
gnewfield, line 50: Can you also add tests for when origin country is `GB` and `null`?

gnewfield (changes requested):

src/pages/Landing/index.tsx
JackShort: GB is the two character ISO code for UK

src/components/swap/SwapBuyFiatButton.test.tsx
JackShort: I'm adding tests when I implement caching tomorrow because it's just annoying to test with the opacity stuff right now and when caching is implemented it will be easier

src/pages/Landing/index.test.tsx
JackShort: oops forgot to delete

src/analytics/index.tsx
justindomingue: ^ probably worth putting this as a comment in code

tinaszheng (approved):

gnewfield (approved):

PR #7389, opened 2023-09-28
feat: add chains dynamicconfig for feature flags

kristiehuang:
Using statsig Dynamic Config to store list of chains that we want to support feature-flagging for.

Linear ticket: https://linear.app/uniswap/issue/WEB-2912/set-up-statsig-dynamic-config-for-the-interface-app
Statsig dynamic config: https://console.statsig.com/5M2TFMQiHkbY9RML95FAEa/dynamic_configs/quick_route_chains

### Before

### After

### Reproducing the error

1. 

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test

### DIFF
--- a/src/featureFlags/index.tsx
+++ b/src/featureFlags/index.tsx
@@ -100,6 +100,7 @@ export function useBaseFlag(flag: string, defaultValue = BaseVariant.Control): B
   }
 }
 
+// eslint-disable-next-line import/no-unused-modules
 export function useFeatureFlagsChains(): ChainId[] {
   const { config } = useFeatureFlagsConfig()
   const chains = config.get(FeatureFlagConfig.chains, [])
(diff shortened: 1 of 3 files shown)


just-toby (approved):

tinaszheng:
ahh my bad for not explaining this dynamic config - it should be specific to the `quickroutes` project, i.e. we want to support all dynamic configs but the one we want to add is just for storing chainId for `quickroutes` in particular, not for all feature flags!

src/featureFlags/index.tsx
```diff
@@ -38,6 +43,10 @@ function useFeatureFlagsContext(): FeatureFlagsContextType {
   }
 }
 
+function useFeatureFlagsConfig(): ConfigResult {
+  return useConfig('feature_flags')
```
tinaszheng, line 47: so we you should write a generic `useDynamicConfig` getter similar to `useBaseFlag` that accepts the config name, and the config names should also be an enum type similar to `FeatureFlag` 

so something like:

```
export enum DynamicConfig {
  quickRouteChains = 'quick_route_chains'
}

export function useDynamicConfig(config: DynamicConfig, defaultValue = '') {
    return useConfig(config)
}
```

something like this

tinaszheng (changes requested):

tinaszheng (approved):
💯 🔥 ⭐

PR #7393, opened 2023-09-29
chore: add caching to origin country and tests

JackShort:
## Description

_Linear ticket:_
_Slack thread:_
_Relevant docs:_

### Before

### After

### Reproducing the error

1. 

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test

### DIFF
--- a/src/components/swap/SwapBuyFiatButton.tsx
+++ b/src/components/swap/SwapBuyFiatButton.tsx
@@ -5,9 +5,10 @@ import { TraceEvent } from 'analytics'
 import { useAccountDrawer } from 'components/AccountDrawer'
 import { ButtonText } from 'components/Button'
 import { MouseoverTooltip } from 'components/Tooltip'
-import { useIsNotOriginCountry } from 'hooks/useIsNotOriginCountry'
 import { useCallback, useEffect, useState } from 'react'
-import styled from 'styled-components'
+import { useAppSelector } from 'state/hooks'
+import { AppState } from 'state/reducer'
+import styled, { css } from 'styled-components'
 import { ExternalLink } from 'theme/components'
 import { textFadeIn } from 'theme/styles'
 
@@ -26,8 +27,7 @@ enum BuyFiatFlowState {
   ACTIVE_NEEDS_ACCOUNT,
 }
 
-const StyledTextButton = styled(ButtonText)`
-  ${textFadeIn}
+const StyledTextButton = styled(ButtonText)<{ shouldFadeIn: boolean }>`
   color: ${({ theme }) => theme.neutral2};
   gap: 4px;
   font-weight: 485;
@@ -37,12 +37,19 @@ const StyledTextButton = styled(ButtonText)`
   &:active {
     text-decoration: none;
   }
+  ${({ shouldFadeIn }) =>
+    shouldFadeIn &&
+    css`
+      ${textFadeIn}
+    `}
 `
 
 export default function SwapBuyFiatButton() {
   const { account } = useWeb3React()
   const openFiatOnRampModal = useOpenModal(ApplicationModal.FIAT_ONRAMP)
-  const shouldShowBuyFiatButton = useIsNotOriginCountry('GB')
+  const originCountry = useAppSelector((state: AppState) => state.user.originCountry)
+  const shouldShowBuyFiatButton = originCountry && originCountry !== 'GB'
+  const [shouldFadeIn] = useState(!originCountry)
   const [checkFiatRegionAvailability, setCheckFiatRegionAvailability] = useState(false)
   const {
     available: fiatOnrampAvailable,
@@ -130,7 +137,12 @@ export default function SwapBuyFiatButton() {
         element={InterfaceElementName.FIAT_ON_RAMP_BUY_BUTTON}
         properties={{ account_connected: !!account }}
       >
-        <StyledTextButton onClick={handleBuyCrypto} disabled={buyCryptoButtonDisabled} data-testid="buy-fiat-button">
+        <StyledTextButton
+          onClick={handleBuyCrypto}
+          disabled={buyCryptoButtonDisabled}
+          data-testid="buy-fiat-button"
+          shouldFadeIn={shouldFadeIn}
+        >
           <Trans>Buy</Trans>
         </StyledTextButton>
       </TraceEvent>
--- a/src/pages/Landing/index.tsx
+++ b/src/pages/Landing/index.tsx
@@ -11,7 +11,7 @@ import { AppleLogo } from 'components/Logo/AppleLogo'
 import { useDisableNFTRoutes } from 'hooks/useDisableNFTRoutes'
 import Swap from 'pages/Swap'
 import { parse } from 'qs'
-import { useEffect, useMemo, useRef } from 'react'
+import { useEffect, useMemo, useRef, useState } from 'react'
 import { ArrowDownCircle } from 'react-feather'
 import { Navigate, useLocation, useNavigate } from 'react-router-dom'
 import { Link as NativeLink } from 'react-router-dom'
@@ -123,7 +123,7 @@ const DownloadWalletLink = styled.a`
   }
 `
 
-const TitleText = styled.h1<{ isDarkMode: boolean; $visible: boolean }>`
+const TitleText = styled.h1<{ isDarkMode: boolean; applyFadeIn: boolean; initiallyVisible: boolean }>`
   color: transparent;
   font-size: 36px;
   line-height: 44px;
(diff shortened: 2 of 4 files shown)


src/state/user/reducer.ts
```diff
@@ -137,6 +143,12 @@ const userSlice = createSlice({
       state.timestamp = currentTimestamp()
     },
     setOriginCountry(state, { payload: country }) {
```
justindomingue, line 145: isn't this slice already persisted through redux-persist?

justindomingue: redux should await store rehydration before rendering

src/pages/Landing/index.tsx
```diff
@@ -140,13 +140,12 @@ const TitleText = styled.h1<{ isDarkMode: boolean; $visible: boolean }>`
...
   background-clip: text;
   -webkit-background-clip: text;
-
-  ${({ $visible }) =>
-    $visible
-      ? css`
-          ${textFadeIn}
-        `
-      : 'opacity: 0;'}
+  ${({ initiallyVisible }) => (initiallyVisible ? `opacity: 1.0;` : 'opacity: 0.0')}
```
justindomingue, line 143:
```suggestion
  ${({ initiallyVisible }) => (initiallyVisible ? `opacity: 1.0;` : 'opacity: 0.0;')}
```

src/pages/Landing/index.tsx
```diff
@@ -174,16 +173,15 @@ const SubText = styled.div`
...
   display: flex;
   justify-content: center;
-
-  ${({ $visible }) =>
-    $visible
-      ? css`
-          ${textFadeIn}
-        `
-      : 'opacity: 0;'}
+  ${({ initiallyVisible }) => (initiallyVisible ? `opacity: 1.0;` : 'opacity: 0.0')}
```
justindomingue, line 179:
```suggestion
  ${({ initiallyVisible }) => (initiallyVisible ? `opacity: 1.0;` : 'opacity: 0.0;')}
```

src/pages/Landing/index.tsx
```diff
@@ -336,6 +334,7 @@ export default function Landing() {
   const shouldDisableNFTRoutes = useDisableNFTRoutes()
   const originCountry = useAppSelector((state: AppState) => state.user.originCountry)
   const renderUkSpecificText = Boolean(originCountry) && originCountry === 'GB'
+  const [initiallyVisible] = useState(!!originCountry)
```
justindomingue, line 337: probably better to use a ref here? generally feels like a code smell though that you need to do this gymnastic

src/state/user/reducer.ts
JackShort: yeah this is persisted but we really need the info on first render - theoretically we could wait for rehydration before rendering everything but that is really out of scope for this pr

just-toby (approved):

src/state/user/reducer.ts
```diff
@@ -5,9 +5,15 @@ import { ConnectionType } from '../../connection/types'
...
 import { RouterPreference } from '../../state/routing/types'
+import {
+  deletePersistedOriginCountry,
+  getPersistedOriginCountry,
+  setPersistedOriginCountry,
+} from './originCountryStorage'
 import { SerializedPair, SerializedToken, SlippageTolerance } from './types'
 
 const selectedWallet = getPersistedConnectionMeta()?.type
+const originCountry = getPersistedOriginCountry()
```
zzmp, line 16: I'd like to see a follow up where we group these localStorage-duplicated data into their own folder, with some shared infra between them so we're not re-writing it every time we need something accessible on load.

src/state/user/reducer.ts
zzmp: we don't wait for rehydration before rendering because it would substantially delay the first frame.

src/state/user/originCountryStorage.ts
```diff
@@ -0,0 +1,34 @@
+import ms from 'ms'
+
+const originCountryKey = 'origin-country'
```
zzmp, line 3: Why do we need the timestamp here? Can't we just have the country?

src/pages/Landing/index.tsx
```diff
@@ -140,13 +140,13 @@ const TitleText = styled.h1<{ isDarkMode: boolean; $visible: boolean }>`
...
-          ${textFadeIn}
-        `
-      : 'opacity: 0;'}
+  ${({ initiallyVisible }) => (initiallyVisible ? `opacity: 1.0;` : 'opacity: 0.0;')}
+  ${({ applyFadeIn }) =>
+    applyFadeIn &&
+    css`
+      ${textFadeIn}
+      opacity: 1.0;
+    `}
```
zzmp, line 149: nit: this could be a css fragment shared between the two styled components.

src/pages/Landing/index.tsx
```diff
@@ -174,16 +174,16 @@ const SubText = styled.div`
...
-    $visible
-      ? css`
-          ${textFadeIn}
-        `
-      : 'opacity: 0;'}
+  ${({ initiallyVisible }) => (initiallyVisible ? `opacity: 1.0;` : 'opacity: 0.0;')}
+  ${({ applyFadeIn }) =>
+    applyFadeIn &&
+    css`
+      ${textFadeIn}
```
zzmp, line 184: won't the fade in only occur if the opacity was not initially 1.0? in this case, is it necessary for applyFadeIn to be used? Couldn't it just always be applied?

src/components/swap/SwapBuyFiatButton.tsx
```diff
@@ -130,7 +137,12 @@ export default function SwapBuyFiatButton() {
         element={InterfaceElementName.FIAT_ON_RAMP_BUY_BUTTON}
         properties={{ account_connected: !!account }}
       >
-        <StyledTextButton onClick={handleBuyCrypto} disabled={buyCryptoButtonDisabled} data-testid="buy-fiat-button">
+        <StyledTextButton
+          onClick={handleBuyCrypto}
+          disabled={buyCryptoButtonDisabled}
+          data-testid="buy-fiat-button"
+          shouldFadeIn={shouldFadeIn.current}
```
zzmp, line 144: won't the fade in only occur if the opacity was not initially 1.0? in this case, is it necessary for shouldFadeIn to be used? Couldn't it just always be applied?

zzmp (changes requested):

src/state/user/originCountryStorage.ts
zzmp: if you _do_ remove the timestamp, keep the stored data as an object, so it is extendable in the future.

JackShort: the timestamp is for cache invalidation

src/components/swap/SwapBuyFiatButton.tsx
tinaszheng, line 144: 👀 why is this using a ref instead of state?

zzmp: using a ref means it will retain the value from initial render, which is the intention. It should not use state, as it won't be updated after it is initially set. See https://github.com/Uniswap/interface/pull/7393#discussion_r1342779363.

src/state/user/reducer.ts
JackShort: by follow up, you mean follow up pr right? or do you expect this in this pr?

src/pages/Landing/index.tsx
JackShort: the fade in should not occur when the origin country is available on the first render only if it is being fetched later. 
so if its available on the first render i set the opacity to 1.0 and the apply fade in effect will not happen

src/components/swap/SwapBuyFiatButton.tsx
JackShort: I answered this above, but shouldFadeIn is for when the country is not in cache and it will fade in later. however, if it is cached then it will never fade in and the opacity will always be set to 1.0

src/state/user/reducer.ts
zzmp: Follow up PR, yeah

zzmp (changes requested):
Instead of accomplishing this through cached origin country, can we piggy-back on the `x-blocked-paths` header that we are sending for GB? That way all of the logic is in one place (our CF configuration), and we can update through configuration without requiring code updates.

You can re-use [this selector](https://github.com/Uniswap/interface/blob/main/src/pages/App.tsx#L155-L156) (move it into utils) and check to see if `/buy` is blocked to decide whether to render buy or not, without checking the origin country at all. More context in this exceedingly long [slack thread](https://uniswapteam.slack.com/archives/C02T729PMQE/p1696431788008969).

`originCountry` is also set through Amplitude, so this seems like a faster way of doing that - `x-blocked-paths` will always be available synchronously on pageload.

PR #7394, opened 2023-09-29
test: mock http requests in jest

just-toby:
speeds up unit tests by mocking http requests using  the `nock`  package and by explicitly setting the `maxWorkers` jest option to 2 (the max CPU cores available on a Github action).

### test plan

tests are passing and the unit test check seems to be running about a minute faster.

to verify that nock is doing what i expect, i used another package `[jest-offline](https://github.com/palmerj3/jest-offline)` which causes a test to fail if it sends a network request. without nock, it caused many tests to fail. with nock, the tests passed.


.github/workflows/test.yml
```diff
@@ -69,7 +69,7 @@ jobs:
           path: node_modules/.cache
           key: ${{ runner.os }}-jest-${{ github.run_id }}
           restore-keys: ${{ runner.os }}-jest-
-      - run: yarn test --coverage --maxWorkers=100%
+      - run: yarn test --coverage --maxWorkers=2
```
zzmp, line 72: `100%` should be equivalent to `2`, and is more future-proof. Can this line be reverted, or is there a reason they are not equivalent?

zzmp (changes requested):
I love this, I just have a concern over `100% == 2`.

zzmp:
Is there a reason you didn't include jest-offline in this PR? It would help guard against regressions (where someone disabled nock to get a test to pass by using real and flakey network requests).

.github/workflows/test.yml
just-toby: i'll revert to compare the run times

zzmp (approved):

.github/workflows/test.yml
zzmp: looks similar - a lot of the difference is from caching, which should be discounted.

just-toby: yep, that sounds right. the actual unit test step is taking the same amount of time

tinaszheng (approved):

PR #7395, opened 2023-09-29
feat: add dynamicconfig modal setting

kristiehuang:
enable seeing a dynamicconfig value / editing the value locally through our FeatureFlagsModal

### Before
Screenshot 2023-09-29 at 5 17 24 PM

### After
Screenshot 2023-09-29 at 5 16 47 PM


src/featureFlags/dynamicConfig/quickRouteChains.ts
```diff
@@ -1,13 +1,20 @@
...
 // eslint-disable-next-line import/no-unused-modules
 export function useQuickRouteChains(): ChainId[] {
   const config = useDynamicConfig(DynamicConfigName.quickRouteChains)
-  const chains = config.get('chains', [])
+  const featureFlagsContext = useFeatureFlagsContext()
+  let chains = config.get('chains', []) as ChainId[]
+
+  const modalSetChains = featureFlagsContext.flags[DynamicConfigName.quickRouteChains]
+  if (Array.isArray(modalSetChains) && modalSetChains !== chains) {
+    chains = modalSetChains
```
just-toby, line 14: this line is for overwriting the remote values with local overrides right? is it possible that the two objects diverge so we lose some fields when assigning like this?
we could also do `chains = { ...chains, ...modalSetChains }`

src/featureFlags/index.tsx
```diff
@@ -25,12 +25,12 @@ export enum FeatureFlag {
 
 interface FeatureFlagsContextType {
   isLoaded: boolean
-  flags: Record<string, string>
+  flags: Record<string, any> // includes flags & configs
```
just-toby, line 28: why `any`? can a config be typed as `Record<string, string>` or something similar?
```suggestion
  flags: Record<string, string | Record<string, string>> // includes flags & configs
```

just-toby (changes requested):
RC to see if we can avoid using `any`

src/featureFlags/index.tsx
```diff
@@ -56,6 +56,20 @@ export function useUpdateFlag() {
   )
 }
 
+export function useUpdateConfig() {
```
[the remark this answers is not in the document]
tinaszheng: hmm actually on second thought, another alternative is to store the dynamic configs in a separate redux data structure than `featureFlags`, maybe add a new state atom for `dynamicConfigSettings`? Benefit here is that you'll still be able to strictly type feature flags while adding special types (or `any`) for just dynamic configs

src/featureFlags/dynamicConfig/quickRouteChains.ts
kristiehuang: hmm, it is for overwriting so no -- if statsig chains are set to [1], but we want to test locally on [3, 10], `useQuickRouteChains()` should be [3, 10]

src/featureFlags/index.tsx
kristiehuang: hmm config is a json object so hard to say what it would be used for -- in this case we're storing something like `{ chains: [1, 3, 10] }` which is Record<string, ChainId[]>

kristiehuang:
going to try to merge this PR first & then look at the changes lynn was supposed to make in separate PRs

just-toby (approved):

src/components/FeatureFlagModal/FeatureFlagModal.tsx
```diff
@@ -218,16 +221,47 @@ function FeatureFlagOption({ value, variant, featureFlag, label }: FeatureFlagPr
...
+  configName: DynamicConfigName
+  label: string
+  options: any[]
+  selected: any[]
+}
+
+function DynamicConfigDropdown({ configName, label, options, selected }: DynamicConfigDropdownProps) {
+  const updateConfig = useUpdateConfig()
+  const handleSelectChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    const selectedValues = Array.from(e.target.selectedOptions, (opt) => Number.parseInt(opt.value))
```
tinaszheng, line 234: so the values are not always going to be integers 🤔 so the default convention is that `select` values are generally strings and its up to the consumer to parse the string into the selected format correctly, maybe allow there to be a `prop` like `parser` that defaults to `Number.parseInt`?

tinaszheng (approved):

src/featureFlags/dynamicConfig/quickRouteChains.ts
```diff
@@ -1,13 +1,20 @@
...
-  const chains = config.get('chains', [])
+  const statsigConfig = useDynamicConfig(DynamicConfigName.quickRouteChains)
+  const featureFlagsContext = useFeatureFlagsContext()
+  let chains = statsigConfig.get(DynamicConfigName.quickRouteChains, []) as ChainId[]
+
+  const modalSettings = featureFlagsContext.configs[DynamicConfigName.quickRouteChains]
+  if (modalSettings) {
+    const modalSetChains = modalSettings[DynamicConfigName.quickRouteChains]
+    if (Array.isArray(modalSetChains) && modalSetChains !== chains) chains = modalSetChains
+  }
```
tinaszheng, line 15: i think this block can be clearer without a `let` statement and to be more descriptive of whats going on 🤔 

```
const remoteConfigChains = statsigConfig.get(DynamicConfigName.quickRouteChains, []) as ChainId[]
const localConfigChains = featureFlagsContext.configs[DynamicConfigName.quickRouteChains]?.[DynamicConfigName.quickRouteChains] 

const chains = Array.isArray(localConfigChains) ? localConfigChains : remoteConfigChains
```

kristiehuang: ah this is much clearer thank you

src/dev/DevFlagsBox.tsx
```diff
@@ -40,11 +46,31 @@ const Gate = (flagName: string, featureFlagSettings: Record<string, string>) =>
...
+}
+
 export default function DevFlagsBox() {
   const featureFlagsAtom = useAtomValue(featureFlagSettingsAtom)
   const featureFlags = useMemo(() => Object.values(FeatureFlag), [])
+  const dynamicConfigsAtom = useAtomValue(dynamicConfigSettingsAtom)
+  const dynamicConfigs = useMemo(() => Object.values(DynamicConfigName), [])
 
   const overrides = featureFlags.map((flagName) => Gate(flagName, featureFlagsAtom))
+  dynamicConfigs.forEach((configName) => overrides.push(Config(configName, dynamicConfigsAtom)))
```
tinaszheng, line 72: think you should memoize these components so they're not new elements on every render

```
const overrides = useMemo(() => {
  const featureFlagOverrides = featureFlags.map((flagName) => Gate(flagName, featureFlagsAtom))
  const dynamicConfigOverrides = dynamicConfigs.map((configName) => Config(configName, dynamicConfigsAtom))

return [...dynamicConfigOverrides, ...featureFlagOverrides].filter(Boolean) // filters out `null` items
}, [dynamicConfigsAtom, dynamicConfigs, featureFlags])
```

tinaszheng: and then your `hasOverrides` check can just be on the length of `overrides`

tinaszheng (approved):

src/dev/DevFlagsBox.tsx
kristiehuang: thought about this but is hard to memoize bc Gate & Config call hooks

PR #7398, opened 2023-10-02
chore: update @web3-react/*

JFrankfurt:
There were some `patches/` files to async import a few of these. I've ported those to the new versions correctly, I think.


cbachmeier (approved):
LGTM tested with uni wallet

zzmp (approved):
at some point we should update the web3-react packages to be accessible as esm, and then remove the patch files, but this lgtm in the interim :)

PR #7402, opened 2023-10-03
feat: Move UniswapX signature expiry back to deadline

tinaszheng:
## Description
We just switched UniswapX signature deadlines to expire if the message gets signed past `startTime`, but it's a bit too strict, so move it back to `deadline`

## Screen capture
Showing that it still submits after waiting for ~30s:

https://github.com/Uniswap/interface/assets/59578595/eb2dfd64-bd41-41d1-9a59-6423bc1d6075

### QA (ie manual testing)
- Wait in the signature screen between 30s and 2 minutes, and the signature should still submit.
- If you wait over 2 minutes, you should see the regular expiration message.


just-toby (approved):
🧐

zzmp (approved):
should there be an in-between, or a warning?
like, if the deadline is less than X seconds because you took so long to sign, then show a tooltip that you may want to re-submit or something?

PR #7403, opened 2023-10-03
fix: no stale trade when otherCurrency is missing

just-toby:
## Description

fixes a bug where we incorrectly used a stale quote when the `otherCurrency` was not defined.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2885/swap-shows-incorrect-quote-after-changing-chains-with-no-output

### Before

https://github.com/Uniswap/interface/assets/66155195/b2d83e31-6c61-45a3-aaaf-c25410f65c66

### After

https://github.com/Uniswap/interface/assets/66155195/43b985a0-b8d3-4f98-9198-21ed42556d23

### Reproducing the error

1. on chain A, select output token (leave input as ETH), and enter an amount to load a quote
2. switch to chain B. verify that the swap inputs are cleared but ETH is still the input token.
3. enter an input amount for ETH - note that a quote is incorrectly shown

### QA (ie manual testing)

- [x] followed the above steps to verify the fix worked

### DIFF
--- a/src/hooks/useDebouncedTrade.ts
+++ b/src/hooks/useDebouncedTrade.ts
@@ -96,6 +96,7 @@ export function useDebouncedTrade(
 
   const skipBothFetches = !autoRouterSupported || !isWindowVisible || isWrap
   const skipRoutingFetch = skipBothFetches || isDebouncing
+
   const skipPreviewTradeFetch =
     skipBothFetches || routerPreference === RouterPreference.CLIENT || isPreviewTradeDebouncing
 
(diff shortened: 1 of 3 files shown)


src/hooks/useDebouncedTrade.ts
```diff
@@ -96,6 +96,7 @@ export function useDebouncedTrade(
 
   const skipBothFetches = !autoRouterSupported || !isWindowVisible || isWrap
   const skipRoutingFetch = skipBothFetches || isDebouncing
+
```
JackShort, line 99: is this intentional

JackShort (approved):

src/hooks/useDebouncedTrade.ts
just-toby: no

PR #7406, opened 2023-10-03
feat: add feature flags settings overrides box

kristiehuang:
Adds box only visible in dev/staging to make clear which flag settings are overriding statsig defaults

_Slack thread:_ https://uniswapteam.slack.com/archives/C047U65H422/p1680039362943649

## Screen capture
Screenshot 2023-10-04 at 2 02 30 PM

### Reproducing the error

1. 

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test


src/components/FeatureFlagModal/DevFlagsBox.tsx
```diff
@@ -0,0 +1,88 @@
...
+  border: 1px solid ${({ theme }) => theme.accent1};
+  z-index: 1000;
+  width: 300px;
+`
+
+export default function DevFlagsBox() {
+  const shouldShow = isDevelopmentEnv() || isStagingEnv()
+  const featureFlags = Object.entries(useAtomValue(featureFlagSettings))
+
+  const statsigValues: { [key: string]: GateResult } = {
```
kristiehuang, line 24: hooks cannot be called in loop :(

src/components/FeatureFlagModal/DevFlagsBox.tsx
```diff
@@ -0,0 +1,88 @@
...
+        <div
+          style={{
+            marginBottom: '8px',
+            display: 'flex',
+            justifyContent: 'space-between',
+          }}
+          onClick={toggleCollapse}
+        >
+          <ThemedText.SubHeader>Feature flag setting overrides</ThemedText.SubHeader>
+          {isCollapsed ? '😿☝️' : '😺👇'}
```
kristiehuang, line 73: by popular demand

src/components/FeatureFlagModal/DevFlagsBox.tsx
```diff
@@ -0,0 +1,88 @@
...
+      }
+    }
+    return overrides
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [featureFlags])
+
+  const [isCollapsed, setIsCollapsed] = useState(false)
+  const toggleCollapse = () => setIsCollapsed(!isCollapsed)
+
+  if (shouldShow) {
```
justindomingue, line 61: would switch the conditions for readability

```
if (!shouldShow) return null

return (<box>...)
```

src/components/FeatureFlagModal/DevFlagsBox.tsx
```diff
@@ -0,0 +1,88 @@
...
+    return (
+      <Box>
+        <div
+          style={{
```
justindomingue, line 65: weird mix of styled components and `style`...?

src/components/FeatureFlagModal/DevFlagsBox.tsx
```diff
@@ -0,0 +1,88 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [featureFlags])
+
+  const [isCollapsed, setIsCollapsed] = useState(false)
+  const toggleCollapse = () => setIsCollapsed(!isCollapsed)
+
+  if (shouldShow) {
+    return (
+      <Box>
+        <div
```
justindomingue, line 64: nbd in this case but for a11y should probably have a button role (or be a button)

src/components/FeatureFlagModal/DevFlagsBox.tsx
justindomingue: one way around this which might be cleaner too is to use a component per gate

```
function Gate(flag: FeatureFlag) {
  const value = useGate(flag)
  const override = statsigValues[flag]
  ...
}

function DevFlagsBox() {
    const featureFlags = Object.entries(useAtomValue(featureFlagSettings))
    return <>{featureFlags.map(ff => <Gate flag={ff} key />)}
```

src/components/FeatureFlagModal/DevFlagsBox.tsx
```diff
@@ -0,0 +1,88 @@
...
+  background-color: ${({ theme }) => theme.surface1};
+  padding: 10px;
+  border: 1px solid ${({ theme }) => theme.accent1};
+  z-index: 1000;
+  width: 300px;
+`
+
+export default function DevFlagsBox() {
+  const shouldShow = isDevelopmentEnv() || isStagingEnv()
+  const featureFlags = Object.entries(useAtomValue(featureFlagSettings))
```
justindomingue, line 22: `Object.entries` will return a new object every render.. better to do that inside the memo below

src/components/FeatureFlagModal/DevFlagsBox.tsx
```diff
@@ -0,0 +1,88 @@
...
+            display: 'flex',
+            justifyContent: 'space-between',
+          }}
+          onClick={toggleCollapse}
+        >
+          <ThemedText.SubHeader>Feature flag setting overrides</ThemedText.SubHeader>
+          {isCollapsed ? '😿☝️' : '😺👇'}
+        </div>
+        {!isCollapsed &&
+          overrides.length &&
```
justindomingue, line 76: don't think you need to check length here given `[].map()` returns `[]`

src/components/FeatureFlagModal/DevFlagsBox.tsx
zzmp: you can also use something like `Object.values(FeatureFlag).filter(isNaN)`

zzmp (changes requested):

kristiehuang:
When you try to disable a flag via the Modal that is currently enabled via Statsig, the toggle button does not flip back to "Enabled". Unrelated issue, will fix in another PR

src/components/FeatureFlagModal/DevFlagsBox.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <ThemedText.LabelSmall key={flagName}>
+          {flagName}: {setting}
+        </ThemedText.LabelSmall>
+      )
+    }
+  }
+  return null
+}
+
+export default function DevFlagsBox() {
```
just-toby, line 41: can you exclude this whole file from the code coverage calculations please: https://github.com/Uniswap/interface/blob/main/codecov.yml

just-toby (approved):
nice

codecov.yml
```diff
@@ -9,6 +9,7 @@ ignore:
   - "**/styled.tsx"
   - "**/constants/**/*"
   - "constants/**/*"
+  - "src/components/FeatureFlagModal/DevFlagsBox.tsx"
```
zzmp, line 12: instead of one-offing a file in here, can you mv it to a dev-specific folder? eg `src/components/dev/...` or `src/dev/DevFlagsBox`?

src/components/FeatureFlagModal/DevFlagsBox.tsx
```diff
@@ -0,0 +1,64 @@
...
+  z-index: 1000;
+  width: 300px;
+`
+const TopBar = styled.div`
+  margin-bottom: 8px;
+  display: flex;
+  justify-content: space-between;
+`
+const Gate = ({ flagSetting }: { flagSetting: [string, string] }) => {
+  const [flagName, setting] = flagSetting
```
zzmp, line 25:
```suggestion
const Gate = ([flagName, flagValue]: [string, string]) => {
```

src/components/FeatureFlagModal/DevFlagsBox.tsx
```diff
@@ -0,0 +1,64 @@
...
+  if (gateResult) {
+    const { value: statsigValue } = gateResult
+    const settingValue = setting === BaseVariant.Enabled
+    if (statsigValue !== settingValue) {
+      return (
+        <ThemedText.LabelSmall key={flagName}>
+          {flagName}: {setting}
+        </ThemedText.LabelSmall>
+      )
+    }
```
zzmp, line 36: Should we also display the values set by StatSig? Why not display them in eg blue if they are set by statsig or red if they are overridden?

If we do do it this way, then we'd want to check everything, and we'd need a separate construct, like `Object.values(FeatureFlags).filter(isNaN).map(Gate)`.

src/components/FeatureFlagModal/DevFlagsBox.tsx
```diff
@@ -0,0 +1,64 @@
...
+    <Box>
+      <TopBar onClick={toggleCollapse}>
+        <ThemedText.SubHeader>
+          {isStagingEnv() && 'Staging build overrides'}
+          {isDevelopmentEnv() && 'Development build overrides'}
+        </ThemedText.SubHeader>
+        {isCollapsed ? '😿☝️' : '😺👇'}
+      </TopBar>
+      {!isCollapsed && overrides}
+      {!isCollapsed && hasNoOverrides && <ThemedText.LabelSmall>No overrides</ThemedText.LabelSmall>}
```
zzmp, line 61:
```suggestion
      {!isCollapsed && (
        hasNoOverrides ? <ThemedText.LabelSmall>No overrides</ThemedText.LabelSmall> : overrides
      )}
```

src/components/FeatureFlagModal/DevFlagsBox.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export default function DevFlagsBox() {
+  const featureFlagsAtom = useAtomValue(featureFlagSettings)
+  const featureFlags = useMemo(() => Object.entries(featureFlagsAtom), [featureFlagsAtom])
+
+  const [isCollapsed, setIsCollapsed] = useState(false)
+  const toggleCollapse = () => setIsCollapsed(!isCollapsed)
+
+  const overrides = featureFlags.map((flag) => Gate({ flagSetting: flag }))
+  const hasNoOverrides = useMemo(() => overrides.every((g) => g === null), [overrides])
```
zzmp, line 49: nit: use the positive case: `hasOverrides = ...overrides.some...`

zzmp (changes requested):

src/components/FeatureFlagModal/DevFlagsBox.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  return (
+    <Box>
+      <TopBar onClick={toggleCollapse}>
+        <ThemedText.SubHeader>
+          {isStagingEnv() && 'Staging build overrides'}
+          {isDevelopmentEnv() && 'Development build overrides'}
+        </ThemedText.SubHeader>
+        {isCollapsed ? '😿☝️' : '😺👇'}
```
zzmp, line 58: If you move this around you can have it collapse much smaller for those that don't want to see it:

```suggestion

        {isCollapsed ? '😿☝️' : '😺👇'}
        {!isCollapsed && <ThemedText.SubHeader>
          {isStagingEnv() && 'Staging build overrides'}
          {isDevelopmentEnv() && 'Development build overrides'}
        </ThemedText.SubHeader>}
```

src/components/FeatureFlagModal/DevFlagsBox.tsx
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+  return null
+}
+
+export default function DevFlagsBox() {
+  const featureFlagsAtom = useAtomValue(featureFlagSettings)
+  const featureFlags = useMemo(() => Object.entries(featureFlagsAtom), [featureFlagsAtom])
+
+  const [isCollapsed, setIsCollapsed] = useState(false)
```
zzmp, line 45: nit: use the positive case: `[isOpen, setIsOpen]` and `toggleOpen`

src/components/FeatureFlagModal/DevFlagsBox.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const toggleCollapse = () => setIsCollapsed(!isCollapsed)
```
zzmp, line 46: This is based on current state, which technically could cause issues when clicking faster than rendering. Instead, this should use the old state for toggling:

```suggestion
  const toggleCollapse = () => setIsCollapsed((value) => !value)
```

src/components/FeatureFlagModal/DevFlagsBox.tsx
```diff
@@ -0,0 +1,64 @@
...
+  position: fixed;
+  bottom: 20px;
+  left: 20px;
+  background-color: ${({ theme }) => theme.surface1};
+  padding: 10px;
+  border: 1px solid ${({ theme }) => theme.accent1};
+  z-index: 1000;
+  width: 300px;
+`
+const TopBar = styled.div`
```
zzmp, line 19: The top bar adds a margin which makes the box look unbalanced when it is collapsed.

zzmp (changes requested):

src/components/FeatureFlagModal/DevFlagsBox.tsx
kristiehuang: hm, do we necessarily care about Statsig values that we haven't overridden?

src/dev/DevFlagsBox.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const [isOpen, setIsOpen] = useState(true)
+  const toggleOpen = () => setIsOpen((open) => !open)
+
+  const overrides = featureFlags.map((flag) => Gate(flag))
+  const hasOverrides = useMemo(() => overrides.some((g) => g !== null), [overrides])
+
+  return (
+    <Box>
+      <TopBar onClick={toggleOpen}>
+        {!isOpen ? '😿☝️' : '😺👇'}
```
zzmp, line 51: nit: this is opposite logic of others

```suggestion
        {isOpen ? '😺👇' : '😿☝️'}
```

zzmp (approved):

zzmp (changes requested):
This is unusable on mobile. It should either be removed from mobile (totally fine solution for now) or made to work there as well.

zzmp (approved):

src/dev/DevFlagsBox.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export default function DevFlagsBox() {
+  const featureFlagsAtom = useAtomValue(featureFlagSettings)
+  const featureFlags = useMemo(() => Object.entries(featureFlagsAtom), [featureFlagsAtom])
+
+  const [isOpen, setIsOpen] = useState(true)
+  const toggleOpen = () => setIsOpen((open) => !open)
+
+  const overrides = featureFlags.map((flag) => Gate(flag))
+  const hasOverrides = useMemo(() => overrides.some((g) => g !== null), [overrides])
```
justindomingue, line 50: this memo isn't doing anything i think (actually hurting perf) because `overrides` will change every render

justindomingue: i would change to this

```
  const featureFlagsAtom = useAtomValue(featureFlagSettings)

  const hasOverrides = useMemo(() => Object.entries(featureFlagsAtom)..map((flag) => Gate(flag)).some((g) => g !== null), [overrides])

```

justindomingue (approved):
some nits but lgtm

src/dev/DevFlagsBox.tsx
```diff
@@ -40,15 +41,15 @@ const Gate = ([flagName, flagSetting]: [string, string]) => {
 }
 
 export default function DevFlagsBox() {
-  const featureFlagsAtom = useAtomValue(featureFlagSettings)
-  const featureFlags = useMemo(() => Object.entries(featureFlagsAtom), [featureFlagsAtom])
+  const featureFlagsAtom = useAtomValue(featureFlagSettingsAtom)
+  const featureFlags = useMemo(() => Object.values(FeatureFlag), [])
```
kristiehuang, line 45: needed to replace this with the enum values instead of the atom values -- because atom is initialized empty

zzmp (approved):

PR #7408, opened 2023-10-03
feat: add sitemap for app.uniswap.org

just-toby:
## Description

adds a sitemap.xml for the app - right now this requires some manual maintenance to keep up to date, because there are no good sitemap generators for CRA apps (i sank many hours into trying to get one working). instead i refactored the route definitions a bit so we can use unit tests to help us stay up to date

_Slack thread:_ https://uniswapteam.slack.com/archives/C04LM5E8KT2/p1696300551701749

### QA (ie manual testing)

- [x] manually tested every url 

### Automated testing

- [x] Unit test - passing / added
- [x] Integration/E2E test - passing


src/pages/routes.test.ts
```diff
@@ -0,0 +1,44 @@
...
+import parser from 'xml2json'
+
+import { routes } from './RouteDefinitions'
+
+describe('Routes', () => {
+  it('sitemap URLs should exist as Router paths', async () => {
+    const pathNames: string[] = routes.map((routeDef) => routeDef.path)
+    const result = await new Promise<boolean>((resolve) => {
+      fs.readFile('./public/sitemap.xml', 'utf8', (err, data) => {
+        try {
```
zzmp, line 11: You can just throw errors to fail the test, instead of `resolve(false)`, and then you can avoid the expectation at the end.

src/pages/routes.test.ts
```diff
@@ -0,0 +1,44 @@
...
+          const sitemap = parser.toJson(data)
+          const sitemapPaths = JSON.parse(sitemap).urlset.url.map((url: any) => new URL(url.loc).pathname)
+
+          sitemapPaths.forEach((path: string) => {
+            if (!pathNames.includes(path)) {
```
zzmp, line 16: This can be expressed as an expectation.

src/pages/routes.test.ts
```diff
@@ -0,0 +1,44 @@
...
+   */
+
+  it('sitemap should match snapshot', () => {
+    const result = fs.readFileSync('./public/sitemap.xml', 'utf8')
+    expect(result).toMatchSnapshot()
+  })
+
+  it('router definition should match snapshot', () => {
+    expect(routes).toMatchSnapshot()
+  })
```
zzmp, line 43: Can a sitemap be generated from the JSON in the RouteDefinitions file, or must it be done manually? It'd be nice if you could write a short utility script to do this for us, especially as it seems like the timestamps will go out of date very quickly.

src/pages/App.tsx
```diff
@@ -5,51 +5,28 @@ import ErrorBoundary from 'components/ErrorBoundary'
...
 import { flexRowNoWrap } from 'theme/styles'
 import { Z_INDEX } from 'theme/zIndex'
 import { STATSIG_DUMMY_KEY } from 'tracing'
-import { getEnvName, isBrowserRouterEnabled } from 'utils/env'
+import { getEnvName } from 'utils/env'
 import { getDownloadAppLink } from 'utils/openDownloadApp'
 import { getCurrentPageFromLocation } from 'utils/urlRoutes'
 import { getCLS, getFCP, getFID, getLCP, Metric } from 'web-vitals'
 
 // High-traffic pages (index and /swap) should not be lazy-loaded.
```
zzmp, line 26: This comment should be moved to the other file.

zzmp (changes requested):

src/pages/routes.test.ts
just-toby: i'm board with this idea in theory, but it's hard to share code between the app .tsx files and the scripts. i think the only way to do this would be to have a standalone, static JSON file with just the path names. we could use the file in both the script and the `RouteDefinitions.tsx` app code. The problem with this approach is just verbosity and having to update multiple files to add a new path - the JSON file and the App code which turns the pathname into a React element. i'm not sure if that extra friction is worth it, vs just accepting some latency in updating the sitemap. what do you think?

just-toby: oh and i added a script which keeps the timestamps up to date

public/sitemap.xml
```diff
@@ -0,0 +1,20 @@
...
+  <url loc="https://app.uniswap.org/pools/v2/find" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
+  <url loc="https://app.uniswap.org/pools/v2" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
+  <url loc="https://app.uniswap.org/pools" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
+  <url loc="https://app.uniswap.org/add/v2" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
+  <url loc="https://app.uniswap.org/add" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
+  <url loc="https://app.uniswap.org/increase" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
+  <url loc="https://app.uniswap.org/migrate/v2" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
+  <url loc="https://app.uniswap.org/nfts" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
+  <url loc="https://app.uniswap.org/nfts/profile" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
+  <url loc="https://app.uniswap.org/not-found" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
```
zzmp, line 19: this does not need to be in the sitemap

public/sitemap.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.sitemaps.org/schemas/sitemap/0.9 http://www.sitemaps.org/schemas/sitemap/0.9/sitemap.xsd">
+  <url loc="https://app.uniswap.org/" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
```
zzmp, line 3: changefreq can be weekly - we do not typically update more than that

src/pages/routes.test.ts
```diff
@@ -0,0 +1,41 @@
...
+          }
+          const sitemap = await parseStringPromise(data)
+
+          const sitemapPaths = sitemap.urlset.url.map((url: any) => new URL(url['$'].loc).pathname)
+
+          sitemapPaths.forEach((path: string) => {
+            expect(pathNames).toContain(path)
+            if (!pathNames.includes(path)) {
+              throw new Error(`${path} is missing from Routes`)
+            }
```
zzmp, line 23:
```suggestion
```

You should test this, but I'm fairly sure that if the expectation fails, an error will be thrown immediately.

src/pages/routes.test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+describe('Routes', () => {
+  it('sitemap URLs should exist as Router paths', async () => {
+    const pathNames: string[] = routes.map((routeDef) => routeDef.path)
+    await new Promise<boolean>((resolve, reject) => {
+      try {
+        fs.readFile('./public/sitemap.xml', 'utf8', async (err, data) => {
+          if (err) {
+            reject(err)
+          }
```
zzmp, line 14: nit: you could simplify this, you have a try-catch around it
```suggestion
          if (err) throw err
```

src/pages/routes.test.ts
```diff
@@ -0,0 +1,41 @@
...
+            if (!pathNames.includes(path)) {
+              throw new Error(`${path} is missing from Routes`)
+            }
+          })
+
+          resolve(true)
+        })
+      } catch (err) {
+        reject(err)
+      }
```
zzmp, line 30: nit: considering the try-catch, you could just resolve at the end:

```suggestion
        })
      } catch (err) {
        reject(err)
      }
      resolve(true)
```

src/pages/routes.test.ts
```diff
@@ -0,0 +1,41 @@
+import fs from 'fs'
+import { parseStringPromise } from 'xml2js'
```
zzmp, line 2: nit: considering this is a test, you could just use the sync versions of readFile and parseString, and avoid any async

zzmp (changes requested):
little nits, mostly there

public/sitemap.xml
zzmp, line 3: should the priority of some of these pages be higher? i feel like /, /swap, /tokens, /nfts, /pools; in that order.

public/sitemap.xml
```diff
@@ -1,20 +1,19 @@
...
-  <url loc="https://app.uniswap.org/pools/v2" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
-  <url loc="https://app.uniswap.org/pools" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
-  <url loc="https://app.uniswap.org/add/v2" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
-  <url loc="https://app.uniswap.org/add" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
-  <url loc="https://app.uniswap.org/increase" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
-  <url loc="https://app.uniswap.org/migrate/v2" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
-  <url loc="https://app.uniswap.org/nfts" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
-  <url loc="https://app.uniswap.org/nfts/profile" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
-  <url loc="https://app.uniswap.org/not-found" lastmod="2023-10-04T16:33:05.134Z" changefreq="daily" priority="0.7"/>
+  <url loc="https://app.uniswap.org/" lastmod="2023-10-05T17:48:32.538Z" changefreq="weekly" priority="1"/>
```
zzmp, line 3: nit: it'd be easier to maintain this if you ordered it by priority

public/sitemap.xml
```diff
@@ -1,20 +1,19 @@
...
+  <url loc="https://app.uniswap.org/pool/v2" lastmod="2023-10-05T17:48:32.538Z" changefreq="weekly" priority="0.6"/>
+  <url loc="https://app.uniswap.org/pool" lastmod="2023-10-05T17:48:32.538Z" changefreq="weekly" priority="0.6"/>
+  <url loc="https://app.uniswap.org/pools/v2/find" lastmod="2023-10-05T17:48:32.538Z" changefreq="weekly" priority="0.6"/>
+  <url loc="https://app.uniswap.org/pools/v2" lastmod="2023-10-05T17:48:32.538Z" changefreq="weekly" priority="0.6"/>
+  <url loc="https://app.uniswap.org/pools" lastmod="2023-10-05T17:48:32.538Z" changefreq="weekly" priority="0.7"/>
+  <url loc="https://app.uniswap.org/add/v2" lastmod="2023-10-05T17:48:32.538Z" changefreq="weekly" priority="0.6"/>
+  <url loc="https://app.uniswap.org/add" lastmod="2023-10-05T17:48:32.538Z" changefreq="weekly" priority="0.6"/>
+  <url loc="https://app.uniswap.org/increase" lastmod="2023-10-05T17:48:32.538Z" changefreq="weekly" priority="0.6"/>
+  <url loc="https://app.uniswap.org/migrate/v2" lastmod="2023-10-05T17:48:32.538Z" changefreq="weekly" priority="0.6"/>
+  <url loc="https://app.uniswap.org/nfts" lastmod="2023-10-05T17:48:32.538Z" changefreq="weekly" priority="0.6"/>
```
zzmp, line 17: nit: should nfts be higher?

src/pages/routes.test.ts
```diff
@@ -6,28 +6,13 @@ import { routes } from './RouteDefinitions'
...
-            }
-          })
-
-          resolve(true)
-        })
-      } catch (err) {
-        reject(err)
-      }
+    sitemapPaths.forEach((path: string) => {
+      expect(pathNames).toContain(path)
```
zzmp, line 15: won't this err with not-found?

zzmp (approved):

tinaszheng (approved):

src/pages/routes.test.ts
just-toby: no, i removed not-found from the sitemap

PR #7411, opened 2023-10-04
feat: support redirects for a list of header paths

cmcewen:
## Description
Support adding a list of blocked paths in cloudflare headers

### QA (ie manual testing)

Add a meta tag `<meta name="x:blocked-paths" content="/,/nfts" />` and see that the pages redirect

### Automated testing
- [X] Unit test
- [ ] Integration/E2E test

### DIFF
--- a/functions/components/metaTagInjector.ts
+++ b/functions/components/metaTagInjector.ts
@@ -39,7 +39,7 @@ export class MetaTagInjector implements HTMLRewriterElementContentHandlers {
       this.append(element, 'twitter:image:alt', this.input.title)
     }
 
-    const blockedPaths = request.headers.get('x-blocked-paths-list')
+    const blockedPaths = this.request.headers.get('x-blocked-paths')
     if (blockedPaths) {
       this.append(element, 'x:blocked-paths', blockedPaths)
     }
--- a/functions/components/metaTagInjector.ts
+++ b/functions/components/metaTagInjector.ts
@@ -39,8 +39,9 @@ export class MetaTagInjector implements HTMLRewriterElementContentHandlers {
       this.append(element, 'twitter:image:alt', this.input.title)
     }
 
-    if (this.blockedPathsList) {
-      this.append(element, 'cf:blockedpaths', this.blockedPathsList)
+    const blockedPaths = request.headers.get('x-blocked-paths-list')
+    if (blockedPaths) {
+      this.append(element, 'x:blocked-paths', blockedPaths)
     }
   }
 }
--- a/functions/components/metaTagInjector.ts
+++ b/functions/components/metaTagInjector.ts
@@ -3,6 +3,7 @@ type MetaTagInjectorInput = {
   image?: string
   url: string
   description?: string
+  country: string | null
 }
 
 /**
@@ -37,6 +38,10 @@ export class MetaTagInjector implements HTMLRewriterElementContentHandlers {
     if (this.input.image) {
       this.append(element, 'twitter:image', this.input.image)
       this.append(element, 'twitter:image:alt', this.input.title)
+    }
+
+    if (this.input.country) {
+      this.append(element, 'cf:country', this.input.country)
     }
   }
 }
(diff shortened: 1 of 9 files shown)


functions/components/metaTagInjector.ts
```diff
@@ -10,7 +10,11 @@ type MetaTagInjectorInput = {
  * to inject meta tags into the <head> of an HTML document.
  */
 export class MetaTagInjector implements HTMLRewriterElementContentHandlers {
-  constructor(private input: MetaTagInjectorInput) {}
+  blockedPathsList: string | null
+
+  constructor(private input: MetaTagInjectorInput, request: Request) {
+    this.blockedPathsList = request.headers.get('cf-blocked-paths-list')
+  }
```
zzmp, line 17: This can be simplified if you're willing to move `request.headers.get(...)` to the `append` body:

```suggestion
  constructor(private input: MetaTagInjectorInput, private request: Request) {}
```

functions/components/metaTagInjector.ts
```diff
@@ -38,5 +42,9 @@ export class MetaTagInjector implements HTMLRewriterElementContentHandlers {
       this.append(element, 'twitter:image', this.input.image)
       this.append(element, 'twitter:image:alt', this.input.title)
     }
+
+    if (this.blockedPathsList) {
+      this.append(element, 'cf:blockedpaths', this.blockedPathsList)
+    }
```
zzmp, line 48: custom headers should begin with `x-`. I'd do the same with meta tags (`x:`) for consistency.

```suggestion
    const blockedPaths = request.headers.get('x-blocked-paths-list')
    if (blockedPaths) {
      this.append(element, 'x:blocked-paths', blockedPaths)
    }
```

functions/components/metaTagInjector.ts
```diff
@@ -10,7 +10,11 @@ type MetaTagInjectorInput = {
  * to inject meta tags into the <head> of an HTML document.
  */
 export class MetaTagInjector implements HTMLRewriterElementContentHandlers {
-  constructor(private input: MetaTagInjectorInput) {}
+  blockedPathsList: string | null
+
+  constructor(private input: MetaTagInjectorInput, request: Request) {
+    this.blockedPathsList = request.headers.get('cf-blocked-paths-list')
```
zzmp, line 16: nit: no need to have the `-list` suffix here.

src/pages/App.tsx
```diff
@@ -200,6 +200,12 @@ export default function App() {
     return null
   }
 
+  // redirect from blocked pages if needed
+  const blockedPathsList = document.querySelector('meta[name="cf:blockedpaths"]')?.getAttribute('content')?.split(',')
+  if (blockedPathsList && blockedPathsList?.includes(pathname)) {
+    return <Navigate to="/swap" replace />
+  }
+
```
zzmp, line 208: I think this will loop. Instead, can you set this as a variable and use it on L231 (before your edits) with the other redirect?

```suggestion
  const blockedPaths = document.querySelector('meta[name="x:blocked-paths"]')?.getAttribute('content')?.split(',')
  const shouldBlockPath = blockedPaths?.includes(pathname) ?? false
```

cmcewen: i added a check to make sure it won't redirect loop - but i don't think your suggestion will work, because it doesn't account for all of the other potential blocked paths - would have to add a unique bool for each possible route

src/pages/App.tsx
```diff
@@ -200,6 +200,12 @@ export default function App() {
     return null
   }
 
+  const blockedPaths = document.querySelector('meta[name="x:blocked-paths"]')?.getAttribute('content')?.split(',')
+  const shouldBlockPath = blockedPaths?.includes(pathname) ?? false
+  if (shouldBlockPath && pathname !== '/swap') {
```
zzmp, line 205: idk, I feel like you should include some weird snarky comment about how you can't block /swap:

```suggestion
  // Uniswap, the unstoppable swappable
  if (shouldBlockPath && pathname !== '/swap') {
```

zzmp (approved):

tinaszheng (approved):

AndrewMohawk (approved):

zzmp (approved):
lgtm, but my last comment stands - it'd be pretty cool to have a snarky comment in there

tinaszheng (approved):

PR #7413, opened 2023-10-04
fix: update description truncation from TokenDescription

just-toby:
## Description

to improve SEO , we should not hide any content behind a user interaction (like a click). web crawlers don't click so they don't see this content. the new solution keeps both strings in the DOM all the time, and just updates the CSS display property to toggle between them

_Relevant docs:_ https://www.notion.so/uniswaplabs/Web-SEO-Improvements-1de35e2d684648efac0ca36055a6f8ad#6aba3250e8d140cbbfac0ba4731c6a29

## Test plan

verified that truncation is still working on the pools page

image

### DIFF
--- a/src/components/Tokens/TokenDetails/TokenDescription.tsx
+++ b/src/components/Tokens/TokenDetails/TokenDescription.tsx
@@ -64,6 +64,10 @@ const TokenDescriptionContainer = styled(ThemedText.BodyPrimary)`
   white-space: pre-wrap;
 `
 
+const DescriptionVisibilityWrapper = styled.span<{ isVisible: boolean }>`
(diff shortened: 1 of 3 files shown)


src/components/Tokens/TokenDetails/TokenDescription.tsx
```diff
@@ -64,6 +64,10 @@ const TokenDescriptionContainer = styled(ThemedText.BodyPrimary)`
   white-space: pre-wrap;
 `
 
+const DescriptionVisibilityWrapper = styled.span<{ isVisible: boolean }>`
+  display: ${({ isVisible }) => (isVisible ? 'inline' : 'none')};
```
JackShort, line 68: nit could just use `$visible`

JackShort (approved):

PR #7416, opened 2023-10-04
feat: pool page currency conversion

JackShort:
* updating v3 pool page to use the appropriate formatting functions
Screen Shot 2023-10-18 at 2 26 53 PM
Screen Shot 2023-10-18 at 2 27 01 PM
Screen Shot 2023-10-18 at 2 32 36 PM
Screen Shot 2023-10-18 at 2 32 43 PM
Screen Shot 2023-10-18 at 2 33 03 PM


src/components/LiquidityChartRangeInput/index.tsx
```diff
@@ -151,9 +152,9 @@ export default function LiquidityChartRangeInput({
 
       const percent = (x < price ? -1 : 1) * ((Math.max(x, price) - Math.min(x, price)) / price) * 100
 
-      return price ? `${format(Math.abs(percent) > 1 ? '.2~s' : '.2~f')(percent)}%` : ''
+      return price ? `${(Math.sign(percent) < 0 ? '-' : '') + formatPercent(percent)}` : ''
```
zzmp, line 155: What was the previous logic here?

zzmp (commented):
lgtm but I don't understand the LiquidityChartRangeInput

src/components/LiquidityChartRangeInput/index.tsx
JackShort: it switched to two fixed two decimal points when the percent was between -1-0-1, and two sig figs when it was greater than 1. I thought it was just best to keep it to how we currently handle percentages

zzmp (approved):

tinaszheng (approved):

PR #7419, opened 2023-10-05
fix: Remove Minus Sign for FOT Display

mattbspector:
## Description
Removing the minus sign to come to parity with mobile

Ticket here: https://linear.app/uniswap/issue/WEB-2917/remove-in-front-of-fot-fee

image

### Before

### After

### Reproducing the error

1. 

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test

### DIFF
--- a/src/components/swap/SwapLineItem.tsx
+++ b/src/components/swap/SwapLineItem.tsx
@@ -71,8 +71,8 @@ function ExchangeRateRow({ trade }: { trade: InterfaceTrade }) {
 }
 
 function ColoredPercentRow({ percent }: { percent: Percent }) {
-  const { formatPriceImpact } = useFormatter()
-  return <ColorWrapper textColor={getPriceImpactColor(percent)}>{formatPriceImpact(percent)}</ColorWrapper>
+  const { formatSlippage } = useFormatter()
+  return <ColorWrapper textColor={getPriceImpactColor(percent)}>{formatSlippage(percent)}</ColorWrapper>
 }
 
 function CurrencyAmountRow({ amount }: { amount: CurrencyAmount<Currency> }) {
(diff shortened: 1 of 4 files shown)


zzmp (approved):

JackShort (approved):

src/components/swap/SwapLineItem.tsx
```diff
@@ -71,8 +71,8 @@ function ExchangeRateRow({ trade }: { trade: InterfaceTrade }) {
 }
 
 function ColoredPercentRow({ percent }: { percent: Percent }) {
-  const { formatPriceImpact } = useFormatter()
-  return <ColorWrapper textColor={getPriceImpactColor(percent)}>{formatPriceImpact(percent)}</ColorWrapper>
+  const { formatSlippage } = useFormatter()
+  return <ColorWrapper textColor={getPriceImpactColor(percent)}>{formatSlippage(percent)}</ColorWrapper>
```
tinaszheng, line 75: so there are other instances of price impact having the negative sign, including [here](https://github.com/Uniswap/interface/blob/main/src/components/swap/PriceImpactWarning.tsx#L44) 

i think we should just fix all price impact to be not negative sign since the negative sign never made sense to me lol

theres one place (the USD value change) where it would probably still make sense to keep, but we should special case that one

tinaszheng (changes requested):

src/components/swap/SwapLineItem.tsx
mattbspector: Makes sense, I'll double check with product to make sure that they are aligned

mattbspector: @tinaszheng Spoke to product, sounds like this is the only place we want to remove the minus sign

mattbspector:
Only getting rid of the - sign in this one spot for now

tinaszheng (approved):

PR #7420, opened 2023-10-05
fix: override user pref in analytics

just-toby:
## Description

_Slack thread:_ https://uniswapteam.slack.com/archives/C052XQ3EUTT/p1696466852040699
_Relevant docs:_ https://www.notion.so/Web-UniswapX-Opt-Out-d7d5d33b3e2c447ab43d40cc92e427ae

## Test plan

logged the value i'm setting to the user preference in the following cases and manually verified it's correct:
1. stored preference = API and has not opted out => X
2. stored preference = client => client
3. stored preference = X => X
4. stored preference = API after user opted out => X

### DIFF
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -6,12 +6,14 @@ import Loader from 'components/Icons/LoadingSpinner'
 import NavBar, { PageTabs } from 'components/NavBar'
 import { useFeatureFlagsIsLoaded } from 'featureFlags'
 import { useInfoPoolPageEnabled } from 'featureFlags/flags/infoPoolPage'
+import { useUniswapXDefaultEnabled } from 'featureFlags/flags/uniswapXDefault'
 import { useAtom } from 'jotai'
 import { useBag } from 'nft/hooks/useBag'
 import { lazy, Suspense, useEffect, useLayoutEffect, useMemo, useState } from 'react'
 import { Navigate, Route, Routes, useLocation, useSearchParams } from 'react-router-dom'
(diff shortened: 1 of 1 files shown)


src/pages/App.tsx
```diff
@@ -165,8 +169,21 @@ export default function App() {
...
   useEffect(() => {
+    // If we're not in the transition period to UniswapX opt-out, set the router preference to whatever is specified.
+    if (!isUniswapXDefaultEnabled) {
+      user.set(CustomUserProperties.ROUTER_PREFERENCE, routerPreference)
+      return
+    }
+
+    // In the transition period, override the stored API preference to UniswapX if the user hasn't opted out.
+    if (routerPreference === RouterPreference.API && !userOptedOutOfUniswapX) {
+      user.set(CustomUserProperties.ROUTER_PREFERENCE, RouterPreference.X)
```
tinaszheng, line 180: oh is this because we will eventually pass through UniswapX as the preference? would it make more sense to just override their actual router preference in this case

tinaszheng (approved):

src/pages/App.tsx
just-toby: yeah, we eventually need to override the stored preference. i'm doing it this way because i wanted to write a single redux migration to migrate everyone's preferences at once, but since we're doing a partial rollout for a couple weeks i need to wait until it's at 100% to enact the migration

cbachmeier (approved):

src/pages/App.tsx
tinaszheng: smart!

PR #7422, opened 2023-10-05
fix: test if description is hidden or not

cbachmeier:
- tests if the element is hidden or not
- updated the id


just-toby (approved):
thanks!

PR #7425, opened 2023-10-05
fix: ignore large slices in immutable check

just-toby:
## Description

our large redux slices are causing console warnings (and unit test failures) in the immutable check middleware, so we should ignore them in this check.

these checks don't happen in prod, so this won't have any effect on users


tinaszheng (approved):
ok nice

JackShort (approved):

zzmp:
Is there a reason they are so large though?

just-toby:
> Is there a reason they are so large though?

yes, they're storing the full token lists

PR #100, opened 2018-10-28
Add Warning Messages and various fixes

chikeichan:

PR #7429, opened 2023-10-05
feat: add top token urls to sitemap and improve script

just-toby:
## Description

1. fixes the script to only update the sitemap's `lastmod` values if they are more than a week old
2. adds a step to the script for fetching the top 100 tokens for each chain and adding them to the sitemap if not already there

## Test plan

ran the script and inspected the results. ran it again to make sure it didn't change the sitemap


package.json
```diff
@@ -256,6 +256,7 @@
     "multicodec": "^3.0.1",
     "multihashes": "^4.0.2",
     "nock": "^13.3.3",
+    "node-fetch": "^3.3.2",
```
just-toby, line 259: this was already in the project as a 2nd order dependency

scripts/generate-sitemap.js
```diff
@@ -3,23 +3,67 @@
...
+const nowISO = new Date().toISOString()
+
+const getQuery = (chain) => `
+  query {
+    topTokens(pageSize: 100, page: 1, chain: ${chain}, orderBy: VOLUME) {
+      address
+    }
+  }
+`
+const chains = ['ETHEREUM', 'ARBITRUM', 'OPTIMISM', 'POLYGON', 'BASE', 'BNB', 'CELO']
```
cbachmeier, line 16: can we import this from src? I think we already have an array of supported chain names and this will be easy to forget to update

scripts/generate-sitemap.js
```diff
@@ -3,23 +3,67 @@
...
+
+    for (const chainName of chains) {
+      const response = await fetch('https://api.uniswap.org/v1/graphql', {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          Origin: 'https://app.uniswap.org',
+          'User-Agent':
+            'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.5735.110 Safari/537.36',
+        },
```
cbachmeier, line 40: are these headers needed?

scripts/generate-sitemap.js
```diff
@@ -3,23 +3,67 @@
...
+        body: JSON.stringify({ query: getQuery(chainName) }),
+      })
+      const tokensJSON = await response.json()
+      const tokenAddresses = tokensJSON.data.topTokens.map((token) => token.address.toLowerCase())
```
kristiehuang, line 44: do we want all 100 top tokens in the sitemap?  + if this array is changing on a frequent basis (say hourly), does this mean our sitemap is also changing hourly?

kristiehuang (changes requested):

scripts/generate-sitemap.js
kristiehuang: i.e. top token #100 rn is Alien Milady Fumo lol

scripts/generate-sitemap.js
just-toby: nope, removing

scripts/generate-sitemap.js
just-toby: i think we do, the sitemap is meant to be an index of as many pages as possible. and i think we do want the "token of the month" in the sitemap since more people will be searching for it while it's trending, and we want our TDP for that token to rank highly

kristiehuang (approved):

scripts/generate-sitemap.js
just-toby: no, this is a generated gQL type and i didn't find such an array already existing

cbachmeier: okay so you didn't need to manually add these, these were auto-generated? we may want to not check these in then similar to how we don't check in the other GQL schema

cbachmeier: This also seems identical to `BACKEND_SUPPORTED_CHAINS` in `src/graphql/data/util.tsx`

cbachmeier (commented):
see comments here https://github.com/Uniswap/interface/pull/7429#discussion_r1349274440

scripts/generate-sitemap.js
just-toby: oh sorry, i meant that the `Chain` type accepted in the query was a gQL definition and i didn't see an array of that type in `src/constants/chains.ts`.

`BACKEND_SUPPORTED_CHAINS` definitely works though

just-toby: actually, i don't think we can import product code into these .js scripts. i always get this error when i try to do it:

`SyntaxError: Cannot use import statement outside a module`

cbachmeier: you'll need to use require syntax, try: `const { BACKEND_SUPPORTED_CHAINS: chains } = require('../src/graphql/data/util.tsx')`

just-toby: that's what i did! i think these scripts aren't compatible with our current build setup

cbachmeier: okay I think we're stuck unless we can get this const in the sdk. Can you update the chain migration information to also include updating this constant?

cbachmeier (approved):

PR #7432, opened 2023-10-06
feat: [info] update color extract default and have PDP fallback

cbachmeier:
## Description

- Updates the color extract default to be accent1 to fit with spore
- In the PoolDetailsStats graph, fallback to blue for the second token in the case that both colors default to accent1

_Slack thread:_ https://uniswapteam.slack.com/archives/C05NXMASXNZ/p1696614227484049?thread_ts=1696355275.788919&cid=C05NXMASXNZ

### Before
Screenshot 2023-10-06 at 10 34 35 AM

### After
Screenshot 2023-10-06 at 10 52 10 AM

### QA (ie manual testing)

- [x] manually cause color extraction to fallback to default and show new pink and blue fallback

### DIFF
--- a/src/hooks/useColor.ts
+++ b/src/hooks/useColor.ts
@@ -62,13 +62,13 @@ export function useColor(token?: Token) {
   useEffect(() => {
     let stale = false
 
-    // if (token) {
-    //   getColorFromToken(token, src).then((tokenColor) => {
-    //     if (!stale && tokenColor !== null) {
-    //       setColor(tokenColor)
-    //     }
-    //   })
-    // }
+    if (token) {
+      getColorFromToken(token, src).then((tokenColor) => {
+        if (!stale && tokenColor !== null) {
+          setColor(tokenColor)
+        }
+      })
+    }
 
     return () => {
       stale = true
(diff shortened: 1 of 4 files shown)


src/hooks/useColor.ts
```diff
@@ -70,9 +72,9 @@ export function useColor(token?: Token) {
 
     return () => {
       stale = true
-      setColor('#2172E5')
+      setColor(theme.accent1)
     }
-  }, [src, token])
+  }, [src, theme.accent1, token])
```
natew, line 77: i guess just on reading this on dark/light switch this will revert the color to accent1 until it re-runs the color get and sets it again, but probably not a big deal and maybe even fast enough to not notice

natew (approved):

mattbspector (approved):

src/hooks/useColor.ts
cbachmeier: No new query occurs as the token's image's cached and the math's quick to calculate so changing the theme looks instant on the page. Thanks for the callout!

kristiehuang (approved):

PR #7442, opened 2023-10-10
fix: move user prop updater into statsigProvider

just-toby:
## Description

moves all the `useEffect` hooks which set user properties into their own component.  this is important because it puts the logic within StatsigProvider and DarkModeQueryParamProvider, which can affect the values we set.

_Slack thread:_ https://uniswapteam.slack.com/archives/C052XQ3EUTT/p1696973999717179

## Test plan

manually verified that `isUniswapXDefaultEnabled` correctly updates to `true` at this callsite now - before it was never updating to the correct value

### DIFF
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -97,12 +97,10 @@ export default function App() {
   const location = useLocation()
   const { pathname } = location
   const currentPage = getCurrentPageFromLocation(pathname)
-  const isDarkMode = useIsDarkMode()
-  const [routerPreference] = useRouterPreference()
+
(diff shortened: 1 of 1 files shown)


src/pages/App.tsx
```diff
@@ -122,53 +120,6 @@ export default function App() {
     }
   }, [searchParams, setShouldDisableNFTRoutes])
 
-  useEffect(() => {
```
just-toby, line 125: all this code is just relocated and unchanged

mattbspector (approved):

tinaszheng (approved):

PR #7443, opened 2023-10-10
fix: user prop updater staging

just-toby:
staging hotfix for #7442

### DIFF
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -97,12 +97,10 @@ export default function App() {
   const location = useLocation()
   const { pathname } = location
   const currentPage = getCurrentPageFromLocation(pathname)
-  const isDarkMode = useIsDarkMode()
-  const [routerPreference] = useRouterPreference()
+
   const [scrollY, setScrollY] = useState(0)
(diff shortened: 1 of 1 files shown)


src/pages/App.tsx
```diff
@@ -255,3 +207,59 @@ export default function App() {
...
+  const isDarkMode = useIsDarkMode()
+
+  const [routerPreference] = useRouterPreference()
+  const userOptedOutOfUniswapX = useUserOptedOutOfUniswapX()
+  const isUniswapXDefaultEnabled = useUniswapXDefaultEnabled()
+
+  useEffect(() => {
+    // User properties *must* be set before sending corresponding event properties,
+    // so that the event contains the correct and up-to-date user properties.
+    user.set(CustomUserProperties.USER_AGENT, navigator.userAgent)
```
tinaszheng, line 221: are we sure these are also ok to move into the statsig provider? 👀 might run into the issue of these use effects running at the same time as other ones that send events I think? nbd would just keep an eye on it

tinaszheng (approved):

mattbspector (approved):

PR #7446, opened 2023-10-11
fix: remove bridge usdc arbitrum

just-toby:
## Description

moonpay will no longer support bridged USDC on arbitrum, which means we don't have a reason to keep this token in the codebase anymore

_Slack thread:_ https://uniswapteam.slack.com/archives/C047U65H422/p1697038790829849


cbachmeier (approved):

kristiehuang (approved):
just wanna make sure that whatever token list we use to pull arbitrum token info from is updated with the new Arbitrum native USDC

(wasn't the case with the Optimism/Base token list for Base native USDC, had to make PR on op's tokenlist repo to update)

PR #7451, opened 2023-10-11
fix: change default tx deadline to 10m

just-toby:
## Description

updates the default deadline from 30 minutes (1800 seconds) to 10 minutes (600 seconds). this includes changing the constant value used to seed the default state, but also includes a migration for all users who had the old default value persisted. 

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2938/30m-feels-high-for-classic-swaps-xinwan-and-austinadams-can-weigh-in
_Relevant docs:_ https://www.notion.so/Redux-migrations-in-interface-30aca899626a4ac4af7b070f92ce976d

### QA (ie manual testing)

- [x] manually loaded the app and verified that `user.userDeadline` had the new default value saved in my IndexedDB

### Automated testing

- [x] Unit test

### DIFF
--- a/src/state/migrations/1.test.ts
+++ b/src/state/migrations/1.test.ts
@@ -1,5 +1,5 @@
+import { createMigrate } from 'redux-persist'
 import { RouterPreference } from 'state/routing/types'
-import { UserState } from 'state/user/reducer'
 import { SlippageTolerance } from 'state/user/types'
 
 import { migration1, PersistAppStateV1 } from './1'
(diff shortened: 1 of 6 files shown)


src/state/migrations/1.ts
```diff
@@ -0,0 +1,24 @@
...
+export type PersistAppStateV1 = {
+  _persist: PersistState
+} & { user?: UserState }
+
+/**
+ * Migration to change the default user deadline from 30 minutes to 10 minutes.
+ * We only migrate if the saved deadline is the old default.
+ */
+export const migration1 = (state: PersistAppStateV1 | undefined) => {
+  if (state?.user && state.user?.userDeadline === 1800) {
```
cbachmeier, line 14: qq: we don't track if a user has set their deadline to be 1800 manually right?

src/state/migrations/1.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    userSlippageTolerance: SlippageTolerance.Auto,
+    userSlippageToleranceHasBeenMigratedToAuto: true,
+    userDeadline: 1800,
+    tokens: {},
+    pairs: {},
+    timestamp: Date.now(),
+    hideBaseWalletBanner: false,
+  },
+  _persist: {
+    version: 0,
```
cbachmeier, line 21: nit: would this version be updated after running `migration1`? Can you check if it's updated.

src/state/migrations/1.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    rehydrated: true,
+  },
+}
+
+describe('migration to v1', () => {
+  it('should migrate the default deadline', () => {
+    const result = migration1(previousState)
+    expect(result?.user?.userDeadline).toEqual(600)
```
cbachmeier, line 29: nit: can you also verify no other states have been changed?

cbachmeier (approved):
LGTM minor nits and qs

src/state/migrations/1.ts
just-toby: no, we just persist the deadline all the time and 1800 is the initial/default value.

src/state/migrations/1.test.ts
just-toby: yes

mattbspector (approved):
LGTM

PR #7453, opened 2023-10-11
fix: sitemap format

just-toby:
## Description

the old format was valid according to free sitemap validators, but Google Search Console seems to prefer this format.


kristiehuang (approved):

zzmp (approved):
Is there a line limit for sitemaps?

just-toby:
> Is there a line limit for sitemaps?

the limit is 50,000 URLs or 50MB, whichever comes first

PR #7455, opened 2023-10-11
fix: specify canonical URLs

just-toby:
## Description

declares the canonical for each page/route in the app in a way that removes query parameters, to better inform google.

right now, google is selecting the canonical automatically which is causing us to have a bunch of "not indexed" URLs - if we specify the canonical then google will know that this isn't an error and should just prioritize the URL we tell it to.

### QA (ie manual testing)

- [x] just logged the URL to make sure it's what we expect, and inspected the page content to make sure the tag appears


mattbspector (approved):
this should help a lot

src/index.tsx
```diff
@@ -35,8 +36,13 @@ if (window.ethereum) {
 }
 
 function Updaters() {
+  const location = useLocation()
```
cbachmeier, line 39: is there any issue with using a hook here?

src/index.tsx
```diff
@@ -35,8 +36,13 @@ if (window.ethereum) {
...
+  const baseUrl = `${window.location.protocol}//${window.location.host}${location.pathname}`
```
cbachmeier, line 40: could we use `window.location.pathname` instead of using the hook?

cbachmeier: you can also use window.location.origin instead of both protocol and host https://developer.mozilla.org/en-US/docs/Web/API/Location

src/index.tsx
just-toby: no, `<Updaters />` is a child of  `<Router>`. but let's switch just in case someone moves this code in the future

just-toby: sorry i spoke too soon ... when i tested without `useLocation` it breaks. because react-router handles navigations, window.location doesn't update the way we'd need it to

cbachmeier (approved):

PR #7456, opened 2023-10-11
feat: nfts sitemap

just-toby:
## Description

add nft collection pages to the sitemap, so google indexing has a better chance of surfacing them in search results.

### test plan
ran the script and checked the output


scripts/generate-sitemap.js
```diff
@@ -53,10 +67,39 @@ fs.readFile('./public/sitemap.xml', 'utf8', async (err, data) => {
...
-    fs.writeFile('./public/sitemap.xml', xml, (error) => {
+    const path = './public/sitemap.xml'
+    fs.writeFile(path, xml, (error) => {
       if (error) throw error
+      const stats = fs.statSync(path)
+      const fileSizeBytes = stats.size
+      const fileSizeMegabytes = fileSizeBytes / (1024 * 1024)
+
+      if (fileSizeMegabytes > 50) {
+        throw new Error('File size exceeds 50MB')
```
cbachmeier, line 101: qq: how often does this script run? on build or regularly? What's the effect when this errors and is our recourse to get the BE to return less data?

cbachmeier, line 101: nit: to help oncall when tracking down the issue. Should we also add the `sitemap` to help with debugging?
```suggestion
        throw new Error('Generated sitemap file size exceeds 50MB')
```

cbachmeier (approved):
non-blocking nits and Qs

scripts/generate-sitemap.js
```diff
@@ -53,10 +67,39 @@ fs.readFile('./public/sitemap.xml', 'utf8', async (err, data) => {
...
+      headers: {
+        'Content-Type': 'application/json',
+        Origin: 'https://app.uniswap.org',
+      },
+      body: JSON.stringify({ query: nftQuery }),
+    })
+    const nftJSON = await nftResponse.json()
+    const collectionAddresses = nftJSON.data.topCollections.edges.map((edge) => edge.node.nftContracts[0].address)
+    collectionAddresses.forEach((address) => {
+      const collectionURL = `https://app.uniswap.org/nfts/collection/${address}`
```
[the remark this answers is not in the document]
just-toby: what do you mean by this?

scripts/generate-sitemap.js
just-toby: it runs when we push main -> staging, and it would interrupt the process if the script fails. 

not sure if we'd want to just fail silently and leave the sitemap unchanged, or just generate the large sitemap and be alerted to the error on the google search console side.

i think if we reach this point, our recourse would be to prune the sitemap of old/irrelevant TDP URLs. 

the current size is about 1/666 of the size limit. ....

mattbspector (approved):

PR #7460, opened 2023-10-13
fix: Autofocus 'You pay' field when Swap navigation item clicked

darkwing:
## Description

There's a very high likelihood the user intends to use those swap inputs due to clicking the Swap navigation item.  To improve the speed with which a user can go from navigation to executing a swap, this PR autofocuses the "You pay" field immediately after clicking the "Swap" navigation item.

## Screen capture

https://github.com/Uniswap/interface/assets/46655/36aea2fc-d6d3-41a1-8483-8e242720d4dd

## Test plan

1. Click the "Swap" navigation item
2. See the cursor in the "You pay" input
3. Immediately be able to start typing your "you pay" number

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test

### DIFF
--- a/src/components/CurrencyInputPanel/SwapCurrencyInputPanel.tsx
+++ b/src/components/CurrencyInputPanel/SwapCurrencyInputPanel.tsx
@@ -242,6 +242,7 @@ interface SwapCurrencyInputPanelProps {
     onDisabledClick?: () => void
     disabledTooltipBody?: ReactNode
   }
+  autoFocus?: boolean
 }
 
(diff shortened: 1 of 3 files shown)


zzmp (changes requested):
I agree with the UX choice that "You pay" should be auto-focused, but it should not be auto-focused when on the landing page, which requires a clickthrough (you can force the landing page to show by appending `?intro=true`: https://app.uniswap.org/?intro=true).

Can you update this so that it will not autofocus on the landing page? It may be a bit more convoluted, unfortunately :(

tinaszheng:
closing because this PR is stale!

PR #7461, opened 2023-10-13
fix: Make Swap to / from toggle keyboard accessible

darkwing:
## Description
At present the down/up-arrow button to switch the "You pay" and "You receive" value is not keyboard accessible.  Simply switching to using a `button` element, with some helpful CSS, makes this element more accessible.

## Screen capture

https://github.com/Uniswap/interface/assets/46655/a61871d2-8c48-4cb4-9880-62dcf1e41f9e

### QA (ie manual testing)

1.  Open the Swaps page
2. Tab until you get to the Down/Up toggle
3. Press enter a bunch
4. See the values change

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test

### DIFF
--- a/src/pages/Swap/index.tsx
+++ b/src/pages/Swap/index.tsx
@@ -69,10 +69,13 @@ import { useIsDarkMode } from '../../theme/components/ThemeToggle'
 import { OutputTaxTooltipBody } from './TaxTooltipBody'
 import { UniswapXOptIn } from './UniswapXOptIn'
 
-export const ArrowContainer = styled.div`
+export const ArrowContainer = styled.button`
   display: inline-flex;
   align-items: center;
   justify-content: center;
+  background: transparent;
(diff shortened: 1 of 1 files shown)


src/pages/Swap/index.tsx
```diff
@@ -69,10 +69,13 @@ import { useIsDarkMode } from '../../theme/components/ThemeToggle'
 import { OutputTaxTooltipBody } from './TaxTooltipBody'
 import { UniswapXOptIn } from './UniswapXOptIn'
 
-export const ArrowContainer = styled.div`
+export const ArrowContainer = styled.button`
```
zzmp, line 72: Rather than change this to a button, which requires additional styles, would it be feasible to render the actual component using `as="button"` ([docs](https://styled-components.com/docs/api#as-polymorphic-prop))?

PR #7462, opened 2023-10-13
feat: Remove local routing setting

tinaszheng:
## Description
This PR removes the ability to force using client-side SOR. Users will now only use SOR in the case where the routing-api request fails.

### Web 
| Mainnet | L2s | Polygon |
|---------|-----|---------|
|image|image|image|

### Mobile 
| Mainnet | L2s | Polygon |
|---------|-----|---------|
|image|image|image|

## Test plan
I manually overrode my client preference to be `client`, then updated the redux-persist version to be 2, then refreshed the page, and then checked that the setting in IndexedDB did indeed change to `api`.

### QA (ie manual testing)
Use your eyes

### Automated testing
- [x] Unit test

### DIFF
--- a/src/state/migrations/2.test.ts
+++ b/src/state/migrations/2.test.ts
@@ -8,7 +8,7 @@ import { migration2, PersistAppStateV2 } from './2'
 const previousState: PersistAppStateV2 = {
   user: {
     userLocale: null,
-    // @ts-ignore, this is a deprecated router preference
+    // @ts-ignore this is intentionally a string and not the `RouterPreference` enum because `client` is a deprecated option
     userRouterPreference: 'client',
     userHideClosedPositions: false,
     userSlippageTolerance: SlippageTolerance.Auto,
@@ -20,7 +20,7 @@ const previousState: PersistAppStateV2 = {
     hideBaseWalletBanner: false,
   },
   _persist: {
-    version: 0,
+    version: 1,
     rehydrated: true,
   },
 }
--- a/src/state/migrations/2.test.ts
+++ b/src/state/migrations/2.test.ts
@@ -1,0 +1,62 @@
+import { createMigrate } from 'redux-persist'
+import { RouterPreference } from 'state/routing/types'
+import { SlippageTolerance } from 'state/user/types'
+
+import { migration1 } from './1'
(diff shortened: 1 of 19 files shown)


gnewfield (approved):

mattbspector (approved):

just-toby (approved):

src/state/migrations/2.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    rehydrated: true,
+  },
+}
+
+describe('migration to v2', () => {
+  it('should migrate users who currently have `client` router preference', async () => {
+    const migrator = createMigrate(
+      {
+        1: migration1,
+        2: migration2,
```
cbachmeier, line 33: qq: do you need all previous migrations in the migrator or can you just use the relevant one ie 2?

src/state/migrations/2.ts
```diff
@@ -0,0 +1,29 @@
...
+import { RouterPreference } from 'state/routing/types'
+import { UserState } from 'state/user/reducer'
+
+export type PersistAppStateV2 = {
+  _persist: PersistState
+} & { user?: UserState }
+
+/**
+ * Migration to change the default user deadline from 30 minutes to 10 minutes.
+ * We only migrate if the saved deadline is the old default.
```
cbachmeier, line 11: update migration description

cbachmeier (commented):
LGTM other than the migration description. Once that's updated ping me for approval

cbachmeier:
Is it intentional that having the X toggle on only hides the extra settings when there's an actual quote?
Screenshot 2023-10-13 at 9 31 16 AM
Screenshot 2023-10-13 at 9 31 10 AM

tinaszheng:
@cbachmeier yes, since deadline and slippage don't apply to UniswapX trades

src/state/migrations/2.test.ts
tinaszheng: technically only need `2` but its more similar to the behavior on prod if you pass in all :D

src/state/migrations/2.ts
tinaszheng: oops nice catch!

cbachmeier (approved):

PR #7467, opened 2023-10-13
refactor: consolidate percent formatting util

cartcrom:
## Description

Renames existing `formatPercent` utility to `formatDelta`, since that util is used for formatting floats rather than Objects of type `Percent`

Combines formatPriceImpact and formatSlippage into a new `formatPercent` utility, since priceImpact is no longer displayed as negative and both should now follow the same formatting rules (no trailing zeros). 

prepends tilda ~ to places in code where price impact is displayed


src/utils/formatNumbers.ts
```diff
@@ -458,31 +458,21 @@ function formatCurrencyAmount({
...
-
-  return `${Number(slippage.toFixed(3)).toLocaleString(locale, {
-    maximumFractionDigits: 3,
-    useGrouping: false,
-  })}%`
-}
-
-function formatPercent(percent: Nullish<number>, locale: SupportedLocale = DEFAULT_LOCALE) {
-  if (percent === null || percent === undefined || percent === Infinity || isNaN(percent)) {
+function formatDelta(delta: Nullish<number>, locale: SupportedLocale = DEFAULT_LOCALE) {
```
tinaszheng, line 470: hmm the variable name `delta` is a bit vague, maybe we can call it something like `percentString`? either that or just leave a comment above describing what this one is for haha

tinaszheng (approved):
:catjam:

mattbspector (approved):
agreed w tina's comment but otherwise LGTM

mattbspector (approved):
restamping

PR #7471, opened 2023-10-14
fix: currency balance for search token

### LINKED ISSUE #7463: Token Balance doesn't load when searching in token selector
**Bug Description**
if i paste in the token address the balance is loading, but not if its just from the default list view

**Steps to Reproduce**

1. Go to /swap page, open token selector, and paste a token address in the search box for a token which you have a balance for
2. Notice the balance never loads

**Expected Behavior**
The balance should load and match the value shown in the default list view

ksvat:
## Description
fixes #7463

## Screen capture

|    Before  |  After  |
| Screenshot 2023-10-13 at 9 45 09 PM | Screenshot 2023-10-13 at 9 45 20 PM |

### Reproducing the error

1. Search for any token in token selector
2. Observe infinite loading


just-toby (approved):

zzmp (approved):

zzmp:
I think if I close and re-open the PR it may cause tests to run...

just-toby (approved):

PR #7476, opened 2023-10-16
feat: pools sitemap

just-toby:
## Description

adds a sitemap file for top token pairs, linking to the new pools page.

### test plan
verify the format of the new sitemap, and verified the new URLs are valid with the pools page feature flag enabled


scripts/generate-sitemap.js
kristiehuang, line 1: is there a better way to reduce the duplicate code in this file? slightly hard to read

PR #7479, opened 2023-10-16
fix: merge portfolio tokens w/ default token list in Currency Selector

just-toby:
## Description

fixes a bug where we don't show a user's portfolio token (w/ balance) at the top of the token selector if it's neither on the default list or in the local user-added tokens store.

this solution just merges the portfolio tokens into the default list before the sorting step.

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2710/include-all-tokens-w-balances-in-token-selector

### Before
image

### After
image

### Reproducing the error

1. own a token that isn't on the default token list, and make sure it's not in your local "user-added tokens" store
  a. if you need to obtain one, swap for a random token on the app that isn't on the list, then just clear your user state
  b. to clear your user state, in devtools go to application > indexed DB > redux > keyvaluepairs and delete the `persist:interface` entry
2. note that the token doesn't appear at the top of the token selector currency list

### QA (ie manual testing)

- [x] performed the above steps and verified this fixes the issue

### DIFF
--- a/src/components/SearchModal/CurrencySearch.tsx
+++ b/src/components/SearchModal/CurrencySearch.tsx
@@ -1,7 +1,7 @@
 // eslint-disable-next-line no-restricted-imports
 import { t, Trans } from '@lingui/macro'
 import { InterfaceEventName, InterfaceModalName } from '@uniswap/analytics-events'
-import { Currency, Token } from '@uniswap/sdk-core'
+import { ChainId, Currency, Token } from '@uniswap/sdk-core'
 import { useWeb3React } from '@web3-react/core'
 import { Trace } from 'analytics'
 import { useCachedPortfolioBalancesQuery } from 'components/PrefetchBalancesWrapper/PrefetchBalancesWrapper'
@@ -76,9 +76,6 @@ export function CurrencySearch({
   const searchTokenIsAdded = useIsUserAddedToken(searchToken)
 
   const defaultTokens = useDefaultActiveTokens(chainId)
-  const filteredTokens: Token[] = useMemo(() => {
-    return Object.values(defaultTokens).filter(getTokenFilter(debouncedQuery))
-  }, [defaultTokens, debouncedQuery])
 
   const { data, loading: balancesAreLoading } = useCachedPortfolioBalancesQuery({ account })
   const balances: TokenBalances = useMemo(() => {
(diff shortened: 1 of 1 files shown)


src/components/SearchModal/CurrencySearch.tsx
```diff
@@ -76,9 +76,6 @@ export function CurrencySearch({
   const searchTokenIsAdded = useIsUserAddedToken(searchToken)
 
   const defaultTokens = useDefaultActiveTokens(chainId)
-  const filteredTokens: Token[] = useMemo(() => {
```
just-toby, line 79: i just moved this into the `sortedTokens` memo function since it's only used once.

src/components/SearchModal/CurrencySearch.tsx
```diff
@@ -205,7 +225,7 @@ export function CurrencySearch({
 
   // if no results on main list, show option to expand into inactive
   const filteredInactiveTokens = useSearchInactiveTokenLists(
-    !onlyShowCurrenciesWithBalance && (filteredTokens.length === 0 || (debouncedQuery.length > 2 && !isAddressSearch))
+    !onlyShowCurrenciesWithBalance && (sortedTokens.length === 0 || (debouncedQuery.length > 2 && !isAddressSearch))
```
just-toby, line 228: for this check using `sortedTokens` achieves the same thing

src/components/SearchModal/CurrencySearch.tsx
```diff
@@ -100,34 +97,57 @@ export function CurrencySearch({
...
+
+    const filteredTokens = Object.values(defaultTokens)
+      .filter(getTokenFilter(debouncedQuery))
+      // Filter out tokens with balances so they aren't duplicated when we merge below.
+      .filter((token) => !(token.address?.toLowerCase() in balances))
+    const mergedTokens = [...(portfolioTokens ?? []), ...filteredTokens]
+
+    if (balancesAreLoading) {
+      return mergedTokens
+    }
```
zzmp, line 124: Moving this to the bottom is a behavior change. Should this be the first line of the `useMemo` callback?

src/components/SearchModal/CurrencySearch.tsx
```diff
@@ -100,34 +97,57 @@ export function CurrencySearch({
...
+
+    return mergedTokens
+      .filter((token) => {
+        if (onlyShowCurrenciesWithBalance) {
+          return balances[token.address?.toLowerCase()]?.usdValue > 0
+        }
+
+        // If there is no query, filter out unselected user-added tokens with no balance.
+        if (!debouncedQuery && token instanceof UserAddedToken) {
```
zzmp, line 133: I think that `UserAddedToken`s are only introduced through `defaultTokens`, so de-duping in L119 may cause them to continue to display with no balance.

zzmp (changes requested):

src/components/SearchModal/CurrencySearch.tsx
just-toby: no because now we're using `mergedTokens` as the base list. look at this as if i added a bunch of logic before, and switched from a ternary to an early return , not as if i moved it.

src/components/SearchModal/CurrencySearch.tsx
just-toby: it is true that UserAddedTokens are introduced through defaultTokens, but i tested with a user-added token too and it's appearing correctly

zzmp: this is saying it should *not* appear if it has no balance. if that's still functional then lgtm.

zzmp (approved):

JackShort (approved):

PR #7485, opened 2023-10-17
fix: dedupe matic native token

kristiehuang:
### Before
`/tokens/polygon/NATIVE` was displaying WMATIC
Screenshot 2023-10-17 at 6 04 17 PM

### After
`/tokens/polygon/NATIVE` is the same as `/tokens/polygon/0x0000000000000000000000000000000000001010`
Screenshot 2023-10-17 at 6 04 38 PM

`/tokens/polygon/0x0d500b1d8e8ef31e21c99d1db9a6444d3adf1270` is WMATIC

### Context
Polygon has a precompile contract for its native token @ 0x0000000000000000000000000000000000001010
& there was confusion in the codebase around the "wrapped" token usage
[Here are the addresses](https://github.com/Uniswap/data-api-graphql/blob/4d06f6d825caa17e985778815b80cfb40fed161a/lib/lambda/graphql/data-processors/constants.ts#L17) we use for native tokens on the BE, we only use 0x0...1010 for the native token on POLYGON.

Worth noting, though - we cannot query Uniswap V2/V3 subgraphs for Polygon (and most other) native token addresses, so we use WMATIC as proxy to calculate prices/stats. See [NATIVE_PRICE_PROXY_MAP in data api](https://github.com/Uniswap/data-api-graphql/blob/4d06f6d825caa17e985778815b80cfb40fed161a/lib/lambda/graphql/data-processors/constants.ts#L6).

* _Slack thread:_ https://uniswapteam.slack.com/archives/C04D07TQBT4/p1697561344028839


src/constants/tokens.ts
```diff
@@ -353,25 +360,18 @@ function getCeloNativeCurrency(chainId: number) {
   }
 }
 
-export function isMatic(chainId: number): chainId is ChainId.POLYGON | ChainId.POLYGON_MUMBAI {
+export function isPolygon(chainId: number): chainId is ChainId.POLYGON | ChainId.POLYGON_MUMBAI {
```
kristiehuang, line 363: pedantically (after the rename of 2021) polygon is the chain & Matic is the token

zzmp (approved):

src/pages/Swap/index.tsx
```diff
@@ -193,8 +193,8 @@ export function Swap({
   const trace = useTrace()
 
   // token warning stuff
-  const prefilledInputCurrency = useCurrency(initialInputCurrencyId)
-  const prefilledOutputCurrency = useCurrency(initialOutputCurrencyId)
+  const prefilledInputCurrency = useCurrency(initialInputCurrencyId, chainId)
+  const prefilledOutputCurrency = useCurrency(initialOutputCurrencyId, chainId)
```
kristiehuang, line 197: "In most cases, chainId should refer to the connected chain, i.e. `useWeb3React().chainId`. However if this component is being used in a context that displays information from a different, unconnected chain (e.g. the TDP), then chainId should refer to the unconnected chain."

mattbspector (approved):

just-toby (approved):
nice

PR #7489, opened 2023-10-18
chore: cleaning up previous uk blocking functionality

JackShort:
* since we are redirecting the landing page to /swap for uk users, this removes the changes done in #7386 for the landing page for uk users
* it also changes the buy button to check for blocked paths if it should be rendered or not


zzmp (changes requested):
You can rm `src/hooks/useIsNotOriginCountry.ts`.
You _could_ ticket a follow-up to also include country in the headers added by CF, so that we don't need to rely on Amplitude for that (which is a misuse of Amplitude).

src/pages/Landing/index.tsx
```diff
@@ -479,8 +419,8 @@ export default function Landing() {
               />
             ))}
           </CardGrid>
-          <CardGrid cols={extraCards.length}>
-            {extraCards.map(({ darkIcon, lightIcon, ...card }) => (
+          <CardGrid cols={3}>
```
zzmp, line 422:
```suggestion
          <CardGrid cols={MORE_CARDS.length}>
```

src/pages/Landing/index.tsx
```diff
@@ -379,34 +337,6 @@ export default function Landing() {
...
 
-  const titles = useMemo(() => {
-    if (!originCountry) {
-      return {
-        header: null,
-        subHeader: null,
-      }
-    }
-
-    if (renderUkSpecificText) {
```
zzmp, line 390: So we are now blocking nft routes completely, instead of just rendering different text for them. Is that right?

zzmp (changes requested):

src/pages/Landing/index.tsx
JackShort: we are still rendering different text for them its just done where we were previously doing it ( in the render function)

JackShort: https://github.com/Uniswap/interface/pull/7489/files/f8ec4a9c481b0ef5f26a905deb41869ed8c61988#diff-f091cac6919a486258fcacdf80a2b93b0d4219d1aea881a10ad046b91b8c0be9R364-R367

cypress/e2e/landing.test.ts
```diff
@@ -40,16 +40,23 @@ describe('Landing Page', () => {
...
+        const parser = new DOMParser()
+        const doc = parser.parseFromString(res.body, 'text/html')
+        const meta = document.createElement('meta')
+        meta.setAttribute('property', 'x:blocked-paths')
+        meta.setAttribute('content', '/,/buy')
+        doc.head.appendChild(meta)
+
+        res.body = doc.documentElement.outerHTML
+      })
     })
```
zzmp, line 55: nit: this would be a good candidate for a cypress test utility function, as it is re-used across tests. not necessary, though

zzmp (approved):

zzmp (approved):

tinaszheng (approved):

PR #7492, opened 2023-10-18
fix: Changing some language around fees

mattbspector:
## Description
Per product + design, changing some language related to fees

### DIFF
--- a/src/components/swap/GasBreakdownTooltip.tsx
+++ b/src/components/swap/GasBreakdownTooltip.tsx
@@ -4,6 +4,7 @@ import { AutoColumn } from 'components/Column'
 import UniswapXRouterLabel, { UniswapXGradient } from 'components/RouterLabel/UniswapXRouterLabel'
 import Row from 'components/Row'
 import { nativeOnChain } from 'constants/tokens'
+import { chainIdToBackendName } from 'graphql/data/util'
 import { ReactNode } from 'react'
 import { InterfaceTrade } from 'state/routing/types'
 import { isPreviewTrade, isUniswapXTrade } from 'state/routing/utils'
@@ -70,9 +71,13 @@ export function GasBreakdownTooltip({ trade, hideUniswapXDescription }: GasBreak
 }
 
 function NetworkCostDescription({ native }: { native: Currency }) {
+  const chainName = chainIdToBackendName(native.chainId)
+
   return (
     <ThemedText.LabelMicro>
-      <Trans>The cost to transact on the Ethereum network. This must be paid in {native.symbol}.</Trans>{' '}
+      <Trans>
+        Network cost is paid in {native.symbol} on the {chainName} Network in order to transact.
+      </Trans>{' '}
       <ExternalLink href="https://support.uniswap.org/hc/en-us/articles/8370337377805-What-is-a-network-fee-">
         <Trans>Learn more</Trans>
       </ExternalLink>
--- a/src/components/swap/GasBreakdownTooltip.tsx
+++ b/src/components/swap/GasBreakdownTooltip.tsx
@@ -43,7 +43,7 @@ export function GasBreakdownTooltip({ trade, hideUniswapXDescription }: GasBreak
   const inputCurrency = trade.inputAmount.currency
   const native = nativeOnChain(inputCurrency.chainId)
 
-  if (isPreviewTrade(trade)) return <NetworkFeesDescription native={native} />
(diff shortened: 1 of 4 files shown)


src/components/swap/GasBreakdownTooltip.tsx
```diff
@@ -69,12 +69,10 @@ export function GasBreakdownTooltip({ trade, hideUniswapXDescription }: GasBreak
...
 }
 
-function NetworkFeesDescription({ native }: { native: Currency }) {
+function NetworkCostDescription({ native }: { native: Currency }) {
   return (
     <ThemedText.LabelMicro>
-      <Trans>
-        The fee paid to the Ethereum network to process your transaction. This must be paid in {native.symbol}.
-      </Trans>{' '}
+      <Trans>The cost to transact on the Ethereum network. This must be paid in {native.symbol}.</Trans>{' '}
```
cbachmeier, line 75: blocking: we should update this to the name of the network, in the screenshot below I would be paying the fee to the Polygon network not Ethereum
Screenshot 2023-10-18 at 12 14 17 PM

non-blocking: Also I would personally word this as `The cost of this transaction on the X network`. The current wording makes it seem like this is a flat fee per network

cbachmeier (changes requested):

src/components/swap/GasBreakdownTooltip.tsx
mattbspector: @bryankuuu just want to get your 👀 here as well to sign off on the above

gnewfield:
Looks like the unit tests in `src/components/swap/SwapModalFooter.test.tsx` and `src/components/swap/SwapLineItem.test.tsx`, as well as the corresponding snapshots, are failing because of the new message. We should update these.

gnewfield (changes requested):
Looks like the unit tests in `src/components/swap/SwapModalFooter.test.tsx` and `src/components/swap/SwapLineItem.test.tsx`, as well as the corresponding snapshots, are failing because of the new message. We should update these.

src/components/swap/GasBreakdownTooltip.tsx
infredible: Agree, this the name should be variable to match the chain.

I hear your suggestion, but the intent behind the copy was to more clearly attribute the cost to the chain and disambiguate from ourselves. All of the values here are variable, so I'm personally less concerned with it being mistaken as a flat fee. wdyt @bryankuuu ?

bryankuuu: Let's go with this
"Network cost is paid in Y to the X network in order to transact. Learn more"

Thank you!

mattbspector:
Screenshot 2023-10-30 at 12 13 14 PM


@cbachmeier @bryankuuu @infredible pushed up the requested change here

mattbspector:
@gnewfield just pushed up fixes for those tests 🙏

src/components/swap/GasBreakdownTooltip.tsx
mattbspector: pushed this change!

src/components/swap/GasBreakdownTooltip.tsx
```diff
@@ -69,11 +70,13 @@ export function GasBreakdownTooltip({ trade, hideUniswapXDescription }: GasBreak
...
 
-function NetworkFeesDescription({ native }: { native: Currency }) {
+function NetworkCostDescription({ native }: { native: Currency }) {
+  const chainName = chainIdToBackendName(native.chainId)
+
   return (
     <ThemedText.LabelMicro>
       <Trans>
-        The fee paid to the Ethereum network to process your transaction. This must be paid in {native.symbol}.
+        Network cost is paid in {native.symbol} on the {chainName} network in order to transact.
```
natew, line 79: I actually don't know how this works but just curious - i imagine `<Trans>` relies on the inner string contents to do the translation, but here we have multiple slots that are filled in eg `{native.symbol}`, does Trans still work?

natew: not sure who is best suited to answer maybe @zzmp

tinaszheng: so lingui is smart enough to extrapolate variables inside jsx correctly, see the example on their home page :D 
https://lingui.dev/introduction

cbachmeier (approved):

gnewfield (approved):

PR #7494, opened 2023-10-18
feat: [info] add PDP loading skeleton

cbachmeier:
## Description

- Adds loading skeletons for PDP page
- Skeleton adjusts to different screen widths
- While waiting for tx table and graph implementation, always show loading skeleton for those components
- Add tests for loading skeletons
- Fix regression bug from fees work where Pool fee was incorrectly using BIPS instead of flat 10k
- Note: the token description designs have been updated so the skeleton does not currently match the component but will be updated as part of [WEB-2985](https://linear.app/uniswap/issue/WEB-2985/update-token-links)

_Linear ticket:_ https://linear.app/uniswap/issue/WEB-2814/add-loading-skeleton
_Slack thread:_ https://uniswapteam.slack.com/archives/C05NXMASXNZ/p1697663576036959

## Screen capture

Screenshot 2023-10-18 at 3 01 47 PM

### Mobile
| Mobile no-scroll | Mobile mid-scroll | Mobile full-scroll |
| ------------ | ----------- | ----------- |
| Screenshot 2023-10-18 at 3 01 57 PM | Screenshot 2023-10-18 at 3 02 03 PM | Screenshot 2023-10-18 at 3 02 10 PM |

### QA (ie manual testing)

- [ ] Try this [link](https://interface-ipdeif6g3-uniswap.vercel.app/pools/ethereum/0xa06591252908576a0f64a8560ec0c8051782377c) which always show loading as true
- [ ] Test on this [preview](https://interface-git-cab-web-2814-add-loading-skeleton-uniswap.vercel.app/explore/pools/ethereum/0xa06591252908576a0f64a8560ec0c8051782377c) with info feature flags onto make sure it loads before items populate

### Automated testing

- [ ] Unit tests added for existing components


src/pages/PoolDetails/PoolDetailsHeader.tsx
```diff
@@ -89,7 +112,7 @@ export function PoolDetailsHeader({
             {token0?.symbol} / {token1?.symbol}
           </ThemedText.HeadlineSmall>
         </Row>
-        {!!feeTier && <FeeTier>{feeTier / BIPS_BASE}%</FeeTier>}
+        {!!feeTier && <FeeTier>{feeTier / 10000}%</FeeTier>}
```
kristiehuang, line 115: what's the difference between `10000` and `10_000`?

src/pages/RouteDefinitions.tsx
```diff
@@ -139,7 +139,11 @@ export const routes: RouteDefinition[] = [
   }),
   createRouteDefinition({
     path: 'explore/pools/:chainName/:poolAddress',
-    getElement: () => <PoolDetails />,
+    getElement: () => (
+      <Suspense fallback={null}>
```
kristiehuang, line 143: can this Suspense wrapper go inside PoolDetails?

kristiehuang:
Screenshot 2023-10-19 at 4 04 37 PM

should only some of the column names load in?

src/pages/PoolDetails/PoolDetailsHeader.tsx
cbachmeier: This factor is unrelated to the `BIPS_BASE` field and I didn't want to correlate the two

cbachmeier:
> Screenshot 2023-10-19 at 4 04 37 PM
> 
> should only some of the column names load in?

@kristiehuang Yes! We're trying out a new table pattern for smaller screen widths where instead of hiding columns we do a horizontal scroll. see https://uniswapteam.slack.com/archives/C05NXMASXNZ/p1697664257279839?thread_ts=1697663576.036959&cid=C05NXMASXNZ

src/pages/RouteDefinitions.tsx
cbachmeier: No it needs to be top level or else Suspense will show the default pink loading circle. You can see this pattern used throughout the RouteDefinitions

kristiehuang (approved):

gnewfield (approved):
How do you feel about all the `Skeleton` components being in the same file, similar to [this](https://github.com/Uniswap/interface/blob/8734ee5986d9eebf1956b144ce3d086023e461f3/src/components/Tokens/TokenDetails/Skeleton.tsx)? Not a blocker; I see benefits both ways. Putting all the skeleton components in a single file makes the other component files more focused on the loaded content, and it allows the skeleton components to be reused more easily. On the other hand, having a component and its skeleton in the same file is nice because it makes it clear that structural changes to the loaded component should also be reflected in the skeleton.

gnewfield:
> Screenshot 2023-10-19 at 4 04 37 PM
> should only some of the column names load in?

Yep most of the column headers are static, except for the two shown as bubbles, which are supposed to be the symbols of the tokens in the pool, e.g. if this table was loading for the DAI/ETH pool, those two column headers would eventually render as "$DAI" and "$ETH" once the data loads.

cbachmeier:
> How do you feel about all the `Skeleton` components being in the same file, similar to [this](https://github.com/Uniswap/interface/blob/8734ee5986d9eebf1956b144ce3d086023e461f3/src/components/Tokens/TokenDetails/Skeleton.tsx)? Not a blocker; I see benefits both ways. Putting all the skeleton components in a single file makes the other component files more focused on the loaded content, and it allows the skeleton components to be reused more easily. On the other hand, having a component and its skeleton in the same file is nice because it makes it clear that structural changes to the loaded component should also be reflected in the skeleton.

When I original built this I had it into its own file but I think the best pattern is to have each skeleton be directly tied to the component it's representing and updated to its current design. Like you said this allows the two to share styles and if you update the component you don't have to worry about updating the skeleton separately or for the skeleton to become out of date. Additionally this allows the individual components to be loaded separately if they have different data requirements such as the graph vs the stats.

src/pages/PoolDetails/index.tsx
```diff
@@ -79,31 +112,55 @@ export default function PoolDetailsPage() {
...
-        )}
+        <PoolDetailsStatsButtons
+          chainId={chainId}
+          token0={token0}
+          token1={token1}
+          feeTier={poolData?.feeTier}
+          loading={loading}
+        />
+        <PoolDetailsStats poolData={poolData} isReversed={isReversed} chainId={chainId} loading={loading} />
+        {(token0 || token1 || loading) &&
```
just-toby, line 144: rather than nesting these ternaries, we could separate the `loading` case from the loaded case. maybe these lines deserve their own function component at this point

just-toby (approved):

src/pages/PoolDetails/PoolDetailsHeader.tsx
zzmp: it looks like feeTier *is* specified in bips, so it *would* be appropriate to use `BIPS_BASE`. Is that wrong?

zzmp:
> > How do you feel about all the `Skeleton` components being in the same file, similar to [this](https://github.com/Uniswap/interface/blob/8734ee5986d9eebf1956b144ce3d086023e461f3/src/components/Tokens/TokenDetails/Skeleton.tsx)? Not a blocker; I see benefits both ways. Putting all the skeleton components in a single file makes the other component files more focused on the loaded content, and it allows the skeleton components to be reused more easily. On the other hand, having a component and its skeleton in the same file is nice because it makes it clear that structural changes to the loaded component should also be reflected in the skeleton.
> 
> When I original built this I had it into its own file but I think the best pattern is to have each skeleton be directly tied to the component it's representing and updated to its current design. Like you said this allows the two to share styles and if you update the component you don't have to worry about updating the skeleton separately or for the skeleton to become out of date. Additionally this allows the individual components to be loaded separately if they have different data requirements such as the graph vs the stats.

If the intention of a skeleton is ever to be shown while a component is lazy-loading, then it must be decoupled from the component it loads. They can share base definitions, but if the skeleton loads any of the logic, then it cannot be loaded separately, and you'll end up eagerly loading the component and gumming up your package size despite using React's `lazy` mechanisms.

cbachmeier:
> > > How do you feel about all the `Skeleton` components being in the same file, similar to [this](https://github.com/Uniswap/interface/blob/8734ee5986d9eebf1956b144ce3d086023e461f3/src/components/Tokens/TokenDetails/Skeleton.tsx)? Not a blocker; I see benefits both ways. Putting all the skeleton components in a single file makes the other component files more focused on the loaded content, and it allows the skeleton components to be reused more easily. On the other hand, having a component and its skeleton in the same file is nice because it makes it clear that structural changes to the loaded component should also be reflected in the skeleton.
> > 
> > 
> > When I original built this I had it into its own file but I think the best pattern is to have each skeleton be directly tied to the component it's representing and updated to its current design. Like you said this allows the two to share styles and if you update the component you don't have to worry about updating the skeleton separately or for the skeleton to become out of date. Additionally this allows the individual components to be loaded separately if they have different data requirements such as the graph vs the stats.
> 
> If the intention of a skeleton is ever to be shown while a component is lazy-loading, then it must be decoupled from the component it loads. They can share base definitions, but if the skeleton loads any of the logic, then it cannot be loaded separately, and you'll end up eagerly loading the component and gumming up your package size despite using React's `lazy` mechanisms.

@zzmp none of the sub-components here plan to be lazy-loaded outside of the page level lazy-load done in `src/pages/RouteDefinitions.tsx`

src/pages/PoolDetails/index.tsx
cbachmeier: we're in a weird intermediary state where TokenDescription will be replaced with Links which is what the skeleton is built to support, I have a separate PR for that [here](https://github.com/Uniswap/interface/pull/7505), and once either this one or that one are merged the other PR can be updated to simplify this logic

zzmp (approved):
lgtm - I'd like to see the two comments (@kristiehuang's BIPS_BASE callout and @just-toby's early return callout) resolved, but would consider them non-blocking if left as-is.

src/pages/PoolDetails/PoolDetailsHeader.tsx
cbachmeier: Looking back it does seem like this has no impact. I originally thought this was a regression from the fee work as this change occurred there but PDP was separate from adding fees and was just caused by confusion between uni fees and feeTier

zzmp (approved):

PR #7495, opened 2023-10-18
fix: rename WrappedTokenInfo class

kristiehuang:
Misleading class name, thought it was for WETH/WBTC/etc. Proposing a rename to TokenFromList?

Also replace use of this class for Gql token queries- we have an existing `gqlToCurrency` function for this.


src/nft/components/profile/list/NFTListRow.tsx
```diff
@@ -65,7 +65,7 @@ const HideTextOverflow = css`
   white-space: nowrap;
 `
 
-const TokenInfoWrapper = styled(Column)`
+const TokenFromList = styled(Column)`
```
cartcrom, line 68: I think this one is accidental

cartcrom (changes requested):

src/nft/components/profile/list/NFTListRow.tsx
kristiehuang: oops good catch thank you

cartcrom (approved):

mattbspector (approved):

cartcrom (approved):

PR #7500, opened 2023-10-19
fix: add native USDC to Polygon quick-select

kristiehuang:
Add native USDC to Polygon quick-select, replace old bridge USDC.e (which is not issued by circle + will be deprecating Circle Mint API support for): https://www.circle.com/blog/native-usdc-now-available-on-polygon-pos

Slack thread: https://uniswapteam.slack.com/archives/C023ZR8Q1DE/p1697652827963319


src/constants/tokens.ts
```diff
@@ -47,17 +47,17 @@ export const USDC_ARBITRUM_GOERLI = new Token(
 )
 export const USDC_POLYGON = new Token(
   ChainId.POLYGON,
-  '0x2791bca1f2de4661ed88a30c99a7a9449aa84174',
+  '0x3c499c542cef5e3811e1192ce70d8cc03d5c3359',
   6,
   'USDC',
   'USD//C'
```
cbachmeier, line 53: should this name be updated like with the mumbai one?

src/constants/routing.ts
```diff
@@ -102,12 +103,11 @@ export const COMMON_BASES: ChainCurrencyList = {
   [ChainId.POLYGON_MUMBAI]: [
     nativeOnChain(ChainId.POLYGON_MUMBAI),
     WRAPPED_NATIVE_CURRENCY[ChainId.POLYGON_MUMBAI] as Token,
+    USDC_POLYGON_MUMBAI,
     WETH_POLYGON_MUMBAI,
   ],
-
   [ChainId.CELO]: [nativeOnChain(ChainId.CELO), CEUR_CELO, CUSD_CELO, PORTAL_ETH_CELO, PORTAL_USDC_CELO, WBTC_CELO],
   [ChainId.CELO_ALFAJORES]: [nativeOnChain(ChainId.CELO_ALFAJORES), CUSD_CELO_ALFAJORES, CEUR_CELO_ALFAJORES],
-
```
cbachmeier, line 110: nit: I think it's easier to read if the chains are separated by empty lines

mattbspector (approved):

src/constants/tokens.ts
kristiehuang: I am not actually sure if the name here matters for anything - only the chainId, address, decimals, and symbol get used afai can tell. In theory though USDC's full name should be `USD Coin`

cbachmeier (approved):

PR #7507, opened 2023-10-26
fix: display chart when tickCurrent is zero

maliabr:
## Description
Updated the getActiveTick condition to recognize 0 as an acceptable tickAmount.
This state can occur frequently when stable pools have a currentTick of zero.

### Before
Screenshot from 2023-10-26 12-51-45

### After
Screenshot from 2023-10-26 12-51-22


just-toby (approved):
makes sense to me. i think @cartcrom  should also take a look

natew:
`!= null` is a good way for us to standardize on as a check if something is anything except null/undefined (but wont remove 0, empty string etc).

PR #7520, opened 2023-10-30
ci: basic semgrep configuration

mr-uniswap:
## Description
Testing out Semgrep on interface.  This adds a semgrep github action.

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -5,7 +5,6 @@ on:
   push:
     branches:
       - main
-      - master
     paths:
       - .github/workflows/semgrep.yml
   schedule:
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -1,0 +1,25 @@
+name: Semgrep
+on:
+  workflow_dispatch: {}
+  pull_request: {}
+  push:
+    branches:
+      - main
+      - master
+    paths:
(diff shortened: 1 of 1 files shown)


.github/workflows/semgrep.yml
```diff
@@ -0,0 +1,25 @@
+name: Semgrep
+on:
+  workflow_dispatch: {}
+  pull_request: {}
+  push:
+    branches:
+      - main
+      - master
```
cartcrom, line 8: prob don't need master here?

cartcrom (approved):

.github/workflows/semgrep.yml
mr-uniswap: I see, I'll remove that.

.github/workflows/semgrep.yml
```diff
@@ -0,0 +1,24 @@
+name: Semgrep
+on:
+  workflow_dispatch: {}
+  pull_request: {}
+  push:
+    branches:
+      - main
+    paths:
+      - .github/workflows/semgrep.yml
```
kristiehuang, line 9: is `paths` used correctly here? https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths

mr-uniswap: Yeah, this is correct.

kristiehuang (approved):

PR #7529, opened 2023-10-31
fix: remove bridged USDC from quickselect

kristiehuang:
Context: https://uniswapteam.slack.com/archives/C047U65H422/p1698780250493989 
Remove from Base and Optimism

### Before
Screenshot 2023-10-31 at 4 10 06 PM

### After
Screenshot 2023-10-31 at 4 09 47 PM


src/state/migrations/3.ts
```diff
@@ -0,0 +1,29 @@
+import { PersistState } from 'redux-persist'
```
tinaszheng, line 1: did u mean to add this XD

kristiehuang: no oops i removed it aha

mattbspector (approved):

PR #7536, opened 2023-11-01
fix: basel font using wrong variation settings

natew:
## Description

Fix font variation settings and use latest font file

_Linear ticket:_ WEB-3044
_Slack thread:_
_Relevant docs:_

## Screen capture

Just manually editing to show the diff (g and a):

### Before

CleanShot 2023-11-01 at 09 42 00@2x

### After

CleanShot 2023-11-01 at 09 42 04@2x

### Reproducing the error

1. 

### QA (ie manual testing)

- [ ] N/A

### Automated testing

- [ ] Unit test
- [ ] Integration/E2E test


cbachmeier (approved):

src/pages/AppBody.tsx
```diff
@@ -19,7 +19,6 @@ export const BodyWrapper = styled.main<BodyWrapperProps>`
   margin-left: auto;
   margin-right: auto;
   z-index: ${Z_INDEX.default};
-  font-feature-settings: 'ss01' on, 'ss02' on, 'cv01' on, 'cv03' on;
```
zzmp, line 22: Why are these being turned off?

public/index.html
```diff
@@ -57,9 +58,10 @@
         font-weight: 535;
         font-style: normal;
         font-display: block;
-        font-named-instance: 'Book';
+        font-named-instance: 'Medium';
```
zzmp, line 61: What is the difference between `Medium` and `Book`? Are they *both* in use?

src/pages/AppBody.tsx
zzmp: this makes sense to turn off. they are tweaks for the font:
- ss = stylistic set
- cv = character variant

if we want the font to look as expected - the default variants - we should omit this line.

public/index.html
zzmp: It looks like these instance names are never actually referenced, but these *are* disambiguated by their font weights, and selected via sprinkles' fontWeight vars.

src/pages/AppBody.tsx
natew: they were causing the wrong looking `g`

public/index.html
```diff
@@ -57,25 +58,24 @@
...
-        font-named-instance: 'Book';
         src: 
-          url(%PUBLIC_URL%/fonts/Basel-Book.woff) format('woff');
+          url('%PUBLIC_URL%/fonts/Basel-Book.woff') format('woff2'),
+          url('%PUBLIC_URL%/fonts/Basel-Book.woff') format('woff');
       }
 
       @supports (font-variation-settings: normal) {
         * {
           font-family: 'Basel', sans-serif;
```
zzmp, line 78: Why don't we pre-load a variable font for Basel? Is this not available? Are we trying to save bandwidth?

natew: Variable font is much higher weight yea thats basically the reason, there was also a weird issue with variable font having a different `$` sign but not as big a reason

zzmp (approved):

PR #7539, opened 2023-11-01
feat: add redux migration to clear lists state

kristiehuang:
Circle launching multichain native USDC caused breaking changes in some token lists. But due to some atm-unknown bug in our redux code, not all users are getting the updated token lists. This is a bandaid to force everyone's redux to re-fetch token lists.


tinaszheng (changes requested):
in order to register this migration with the root redux logic, you'll have to add it to the migrations list here: https://github.com/Uniswap/interface/blob/main/src/state/migrations.ts#L21

and then also update the latest version here to 3: https://github.com/Uniswap/interface/blob/1c924828554a990b63813b1d19b77c964267d2e2/src/state/reducer.ts#L47

PR #7540, opened 2023-11-02
feat: Remove logging whole error object for analytics

tinaszheng:
## Description
We weren't using this `error` field, and it was logging every individual object property as a separate field, so let's remove it from logging here.


src/hooks/useUniversalRouter.ts
```diff
@@ -104,7 +104,6 @@ export function useUniversalRouterSwapCallback(
             ...analyticsContext,
             client_block_number: blockNumber,
             tx,
-            error: gasError,
```
cbachmeier, line 107: Do we have any desire to log the error message or code?

cbachmeier (approved):

PR #7543, opened 2023-11-02
chore: converting useStablecoinPrice to useUsdPrice in autoslippage

JackShort:
* implements tina's idea to reuse the same routing-api result from the previous useUsdPrice call as explained in slack here: https://uniswapteam.slack.com/archives/C047U65H422/p1698868768687079

### DIFF
--- a/src/hooks/useAutoSlippageTolerance.ts
+++ b/src/hooks/useAutoSlippageTolerance.ts
@@ -11,7 +11,8 @@ import { useMemo } from 'react'
 import { ClassicTrade } from 'state/routing/types'
 
 import useGasPrice from './useGasPrice'
-import useStablecoinPrice, { useStablecoinAmountFromFiatValue, useStablecoinValue } from './useStablecoinPrice'
+import { useStablecoinAmountFromFiatValue, useStablecoinValue } from './useStablecoinPrice'
+import { useUSDPrice } from './useUSDPrice'
 
 const DEFAULT_AUTO_SLIPPAGE = new Percent(5, 1000) // 0.5%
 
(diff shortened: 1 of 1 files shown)


src/hooks/useAutoSlippageTolerance.ts
```diff
@@ -77,30 +78,31 @@ export default function useClassicAutoSlippageTolerance(trade?: ClassicTrade): P
...
-  const nativeCurrencyPrice = useStablecoinPrice((trade && nativeCurrency) ?? undefined)
+
+  const nativeGasCost =
+    nativeGasPrice && typeof gasEstimate === 'number'
+      ? JSBI.multiply(nativeGasPrice, JSBI.BigInt(gasEstimate))
+      : undefined
+  const gasCostUSD = useUSDPrice(
+    trade && nativeCurrency && nativeGasCost ? CurrencyAmount.fromRawAmount(nativeCurrency, nativeGasCost) : undefined
+  )
+  const gasCostStablecoinAmount = useStablecoinAmountFromFiatValue(gasCostUSD.data)
```
tinaszheng, line 94: so dont htink you have to convert this to a stablecoin value, its actually more accurate for autoslippage to just do USD 

so the calculation is like, we want the ratio of gas to trade size, and trade size is currently in terms of USD, so gas in USD is technically better

tinaszheng: oh hmm i see this just converts it to a `CurrencyAmount`, wondering if you can avoid doing conversions at all here

tinaszheng (approved):

mattbspector (approved):

PR #7544, opened 2023-11-02
fix: remove version check in redux lists update + migrate outdated USDC saved tokens

kristiehuang:
Circle launching multichain native USDC caused breaking changes on tokenlists. TLDR, what used to be called "USDC" on many chains was actually non-native USDC bridged from ethereum. With the launch of native USDC, these tokens were renamed to USDC.e / 'Bridged USDC'.
Some users however are not seeing the updated tokenlists in their redux stores, causing some confusing breakage in our UI because of how we use tokenlists.

This PR attempts to fix this.
* Remove versioning check in redux lists -- should update redux lists state on every semver change
* Migrate outdated USDC saved tokens in a users' self-imported tokens store (edge case)

## Test plan
- unit testing


src/state/migrations/3.ts
```diff
@@ -0,0 +1,67 @@
...
+} & { user?: UserState }
+
+/**
+ * Migration to clear users' imported token lists, after
+ * breaking changes to token info for multichain native USDC.
+ */
+export const migration3 = (state: PersistAppStateV3 | undefined) => {
+  if (state?.user) {
+    let tokens = { ...state.user.tokens }
+    const tokenAddresses: { [key in ChainId]?: string } = {
```
zzmp, line 18:
```suggestion
    const USDCe_ADDRESSES: { [key in ChainId]?: string } = {
```

zzmp, line 18: Group your logic, and add comments describing what is happening:

```
let tokens ...

// Update USDC.e tokens to use the the new USDC.e symbol (from USDC)
...

// Update USDbC token to use the new USDbC symbol (from USDC)
...

return {
...
}

File: src/state/migrations/3.ts [hunk 2]
```diff
@@ -0,0 +1,67 @@
...
+  _persist: PersistState
+} & { user?: UserState }
+
+/**
+ * Migration to clear users' imported token lists, after
+ * breaking changes to token info for multichain native USDC.
+ */
+export const migration3 = (state: PersistAppStateV3 | undefined) => {
+  if (state?.user) {
+    let tokens = { ...state.user.tokens }
```
zzmp, line 17: This can be mutable, no? Why not just mutate it, and then always return state?

zzmp (changes requested):

mattbspector (approved):

cartcrom (approved):
LGTM. hope we can gut most of this tokenlist code soon

zzmp (approved):

PR #7545, opened 2023-11-06
fix: spam tokens showing up in currency selector

JackShort:
* spam tokens were showing up in token selector
https://uniswapteam.slack.com/archives/C047U65H422/p1699294205232069

### DIFF
--- a/src/components/SearchModal/CurrencySearch.tsx
+++ b/src/components/SearchModal/CurrencySearch.tsx
@@ -5,6 +5,7 @@ import { ChainId, Currency, CurrencyAmount, Token } from '@uniswap/sdk-core'
 import { useWeb3React } from '@web3-react/core'
 import { Trace } from 'analytics'
 import { useCachedPortfolioBalancesQuery } from 'components/PrefetchBalancesWrapper/PrefetchBalancesWrapper'
+import { TokenBalance } from 'graphql/data/__generated__/types-and-hooks'
 import { supportedChainIdFromGQLChain } from 'graphql/data/util'
 import useDebounce from 'hooks/useDebounce'
 import { useOnClickOutside } from 'hooks/useOnClickOutside'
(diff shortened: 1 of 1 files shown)


src/components/SearchModal/CurrencySearch.tsx
```diff
@@ -99,18 +101,26 @@ export function CurrencySearch({
   }, [chainId, data?.portfolios])
 
   const sortedTokens: Token[] = useMemo(() => {
-    const portfolioTokens = data?.portfolios?.[0].tokenBalances
-      ?.map((tokenBalance) => {
+    const portfolioTokenBalances = data?.portfolios?.[0].tokenBalances as TokenBalance[] | undefined
+    const portfolioTokens = splitHiddenTokens(portfolioTokenBalances ?? [])
```
just-toby, line 105: we might want to hide small balances here - it would match the logic in the mini portfolio

src/components/SearchModal/CurrencySearch.tsx
```diff
@@ -99,18 +101,26 @@ export function CurrencySearch({
...
           supportedChainIdFromGQLChain(tokenBalance.token?.chain) ?? ChainId.MAINNET,
           tokenBalance.token?.address,
           tokenBalance.token?.decimals,
           tokenBalance.token?.symbol,
           tokenBalance.token?.name
         )
+
+        if (portfolioToken.chainId !== chainId) {
+          return undefined
+        }
```
just-toby, line 121: maybe move this check above the `Token` instantiation?

just-toby (approved):

src/components/SearchModal/CurrencySearch.tsx
JackShort: the default param is true for checking for small balances

mattbspector (approved):

PR #7550, opened 2023-11-07
chore: disabling exact output for fot tokens

JackShort:
* disable exact output for FOT tokens

Screen Shot 2023-11-07 at 4 22 46 PM


tinaszheng (approved):
lit

mattbspector (approved):

PR #7552, opened 2023-11-08
chore: add arbitrum fallback endpoint

JFrankfurt:
Adds a fallback endpoint via quicknode for arbitrum.

image


just-toby (approved):

JFrankfurt:
I'm noticing that a subset of requests to infura are not being re-tried with the fallback rpc url. Some are clearly being swapped over when the client blocks an endpoint, but some continue to target the blocked url. Need to investigate that issue before this is ready for review.

JFrankfurt:
closing because stale

